PR #10, opened 2025-05-21
feat(test-suite): add test for two public decrypt requests made in sa…

jatZama:
…me block

### DIFF
--- a/test-suite/e2e/test/gatewayDecrypt/testAsyncDecrypt.ts
+++ b/test-suite/e2e/test/gatewayDecrypt/testAsyncDecrypt.ts
@@ -192,6 +192,16 @@ describe('TestAsyncDecrypt', function () {
     expect(y).to.equal(1267650600228229401496703205443n);
   });
 
+  it('test async decrypt uint128 - two requests in same tx', async function () {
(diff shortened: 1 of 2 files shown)


test-suite/e2e/test/gatewayDecrypt/testAsyncDecrypt.ts
```diff
@@ -132,6 +132,31 @@ describe('TestAsyncDecrypt', function () {
...
+      }),
+      this.contract.requestUint32_3({
+        nonce: pendingNonce+1,
+      }),
+    ]);
+
+    const [receiptA, receiptB] = await Promise.all([txA.wait(), txB.wait()]);
+
+    console.log("txA block:", receiptA.blockNumber);
+    console.log("txB block:", receiptB.blockNumber);
```
rudy-6-4, line 150: can you assert/expect the block is the same ?

jatZama: strictly speaking, depending on the network status, there is no guarantee here because we are racing to try to include both tx in same block, so i would prefer not to, especially since ZWS env is different and they want to use the exact same tests in the real testnet. Should I resolve this? wdyt?

PR #12, opened 2025-05-21
fix(coprocessor): fhevm-listener, build contracts in build.rs

rudy-6-4:
move workaround in CI to cargo

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-listener/build.rs
+++ b/coprocessor/fhevm-engine/fhevm-listener/build.rs
@@ -4,10 +4,68 @@ use foundry_compilers::{
     Project, ProjectPathsConfig,
 };
 use semver::Version;
-use std::path::Path;
+use std::{env, fs, path::Path, process::Command};
+
+fn build_contracts() {
+    println!("cargo:rerun-if-changed=../../../host-contracts");
+    // Step 1: Copy ../../contracts/.env.example to ../../contracts/.env
+    let env_example = Path::new("../../../host-contracts/.env.example");
+    let env_dest = Path::new("../../../host-contracts/.env");
+    let artefacts = Path::new("../../../host-contracts/artifacts");
+    if env_example.exists() {
+        // CI build
+        if !env_dest.exists() {
+            fs::copy(env_example, env_dest).expect("Failed to copy .env.example to .env");
+            println!("Copied .env.example to .env");
+        }
+    } else if artefacts.exists() {
+        // Docker build
+        println!("Assuming artefacts are up to date.");
+        return;
+    } else {
+        panic!("Error: .env.example not found in contracts directory");
+    }
+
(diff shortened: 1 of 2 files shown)


coprocessor/fhevm-engine/fhevm-listener/build.rs
```diff
@@ -4,10 +4,68 @@ use foundry_compilers::{
...
+    }
+    println!("Ran npx hardhat compile successfully");
+}
 
 fn main() {
     println!("cargo::warning=build.rs run ...");
+    build_contracts();
     // build tests contracts
     let paths =
         ProjectPathsConfig::hardhat(Path::new(env!("CARGO_MANIFEST_DIR")))
```
dartdart26, line 71: Do we need the code that installs Solc if we are compiling them via hardhat?

coprocessor/fhevm-engine/fhevm-listener/build.rs
```diff
@@ -4,10 +4,68 @@ use foundry_compilers::{
...
+    }
+
+    // Change to the contracts directory for npm commands.
+    let contracts_dir = Path::new("../../../host-contracts");
+    if !contracts_dir.exists() {
+        panic!("Error: contracts directory not found");
+    }
+    env::set_current_dir(contracts_dir).expect("Failed to change to contracts directory");
+
+    // Step 2: Run `npm ci run `npm ci --include=optional` in ../../contracts
```
dartdart26:
```suggestion
    // Step 2: Run `npm ci --include=optional` in ../../contracts
```

coprocessor/fhevm-engine/fhevm-listener/build.rs
dartdart26: In general, is it possible to use Solc and not run external npm?

dartdart26 (approved):
Looks good! I just added some comments that might allow us to delete code.

Also, can we use Solc instead of external `npm` and avoid JS? Or that would be too complicated with libraries, dependencies, etc?

coprocessor/fhevm-engine/fhevm-listener/build.rs
rudy-6-4: > Do we need the code that installs Solc if we are compiling them via hardhat?

I am gonna try without solc.

For using Solc only, I have tried in the past and failed. It could probably work but I didn't manage to configure it to support all our contract dependencies.

rudy-6-4: Not possible because there is still the test contract which is compile via solc and not hardhat.

rudy-6-4: https://github.com/zama-ai/fhevm-test/issues/56

PR #51, opened 2025-05-22
chore(common): add initial issue templates to guide devs use Conventional Commits titles

isaacdecoded:
Closes zama-ai/fhevm-internal#155


leventdem (approved):
Good to me, having  some examples is also good for general issue.

PR #52, opened 2025-05-22
chore(sdk): first version of rust sdk

leventdem:
For now only the encrypted inputs has been implemented as a first version.


sdk/rust-sdk/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+
+
+[package]
+name = "relayer-sdk-rust"
```
fd0r: Shouldn't it be fhevm-js? I expect relayer-sdk to be implemented in https://github.com/zama-ai/relayer-sdk

sdk/rust-sdk/src/logging.rs
```diff
@@ -0,0 +1,63 @@
+use log::{LevelFilter, info};
```
fd0r: Any reason to use `log` instead of `tracing`?

All other rust components in the company use `tracing` so just for consistency purposes it should imho be `tracing` here.

fd0r (commented):
minor comments

sdk/rust-sdk/LICENSE
```diff
@@ -0,0 +1,28 @@
+BSD 3-Clause Clear License
+
+Copyright © 2024 ZAMA.
```
jatZama: 2025 not 2024

sdk/rust-sdk/Cargo.toml
jatZama: wdym this is a rust crate. fhevm-rs would make sense, not fhevm-js.

sdk/rust-sdk/README.md
```diff
@@ -0,0 +1,213 @@
...
+## Configuration
+
+The SDK can be configured using the builder pattern:
+
+```rust
+let sdk_builder = FhevmSdkBuilder::new()
+    .with_keys_directory(PathBuf::from("./keys"))
+    .with_gateway_chain_id(43113)
+    .with_host_chain_id(11155111)
+    .with_gateway_contract("Decryption", "0x1234567890123456789012345678901234567890")
```
fd0r: Why the choice of `with_gateway_contract(contract-name, contract-address)` instead of `with_contract-name(contract-address)` ?

sdk/rust-sdk/examples/input.rs
```diff
@@ -0,0 +1,180 @@
...
+        Arc::new(public_key.clone()),
+        Arc::new(crs.clone()),
+        chain_id,
+    );
+
+    // Add a boolean value
+    input_builder.add_bool(true)?;
+
+    // Generate the encrypted value with proof
+    let encrypted_value = input_builder.encrypt_for(contract_address, user_address)?;
```
fd0r: `encrypt_for` -> `encrypt_and_prove_for` ?

sdk/rust-sdk/Cargo.toml
leventdem: in sdk folder we will have a rust sdk and a js sdk

sdk/rust-sdk/README.md
leventdem: they are stored in two different structure fields

```
            gateway_contracts: HashMap::new(),
            host_contracts: HashMap::new(),
```

I think it is easier to follow when calling the builder, i.e knowing for each blockchain it is added.

sdk/rust-sdk/examples/input.rs
leventdem: yes, make sense 👍

fd0r:
I really like the examples! 👍🏼

sdk/rust-sdk/README.md
fd0r: My main concern here is that a developer can easily make a typo making it hard to debug or to even know the name of contracts that are expected.

Maybe having something like

```rust
struct GatewayContracts{
    contract_name: Option<Address>,
    ...
}
```

Would make sense as we have a finite number of contracts associated with the fhevm?

sdk/rust-sdk/Cargo.toml
fd0r: @jatZama it was a typo, I meant fhevm-rs indeed

fd0r: What I mean is that it doesn't make sense to mention the relayer here imo

sdk/rust-sdk/src/decryption/user.rs
```diff
@@ -0,0 +1,1014 @@
...
+    pub request_validity: RequestValidity,
+    pub contracts_chain_id: u64,
+    pub contract_addresses: Vec<Address>,
+    pub user_address: Address,
+    pub signature: Bytes,
+    pub public_key: Bytes,
+}
+
+/// Builder pattern for creating UserDecryptRequest instances
+pub struct UserDecryptRequestBuilder {
```
leventdem: This module is starting to get large, so I will split it and separate the builders into their own files.

sdk/rust-sdk/src/logging.rs
leventdem: I will change it 👍

sdk/rust-sdk/artifacts/contract-abis/Decryption.json
jatZama, line 1: Do you really need all those artifacts? in the js sdk we dont use them to make the package significantly lighter and easier to maintain. We only use hardcoded interfaces as array of strings there. Could you not do the same here?

sdk/rust-sdk/src/blockchain/bindings.rs
```diff
@@ -0,0 +1,71 @@
...
+);
+
+// Define the Transfer event structure using alloy_sol_types
+sol! {
+    #[derive(Debug)]
+    event Transfer(address indexed from, address indexed to, uint256 value);
+}
+
+#[cfg(test)]
+mod tests {
```
jatZama: those tests seem useless for me, did I miss something?

sdk/rust-sdk/src/main.rs
```diff
@@ -0,0 +1,204 @@
...
+    FhevmSdkBuilder::new()
+        .with_keys_directory(path::PathBuf::from("./keys"))
+        .with_gateway_chain_id(43113)
+        .with_host_chain_id(11155111) // Example: Ethereum Sepolia
+        .with_gateway_contract("Decryption", "0x1234567890123456789012345678901234567bbb")
+        .with_gateway_contract(
+            "input-verifier",
+            "0x1234567890123456789012345678901234567aaa",
+        )
+        .with_host_contract("acl", "0x0987654321098765432109876543210987654321")
```
jatZama: Can you please replace all instances of "acl" with "ACL" to keep same conventions as in host-contracts?

sdk/rust-sdk/src/decryption/user.rs
```diff
@@ -0,0 +1,1014 @@
...
+        // Validate timestamp is not in the future (with 5 minute tolerance)
+        let now = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap()
+            .as_secs();
+
+        if start_timestamp > now + 300 {
+            // 5 minute tolerance
+            return Err(FhevmError::InvalidParams(
+                "Start timestamp cannot be more than 5 minutes in the future".to_string(),
```
jatZama: Why did you add this check? As far as I remember it is not present in the js sdk, and maybe there are legitimate use cases to prepare a request for later, for instance in the case of delegation to a smart contract account?

sdk/rust-sdk/Cargo.toml
immortal-tofu: Should be fhevm-rs, definitely

sdk/rust-sdk/src/decryption/user.rs
```diff
@@ -0,0 +1,1014 @@
...
+            ));
+        }
+
+        self.start_timestamp = Some(start_timestamp);
+        self.duration_days = Some(duration_days);
+        Ok(self)
+    }
```
jatZama: I would have kept at least the optional possibility to add acl checks (those are present by default always in the js sdk) maybe by adding a boolean argument "check_acl", to avoid the user to potentially waste gas. This would do an optional eth_call on isAllowed on all the handles requested for decryption. Wdyt?

sdk/rust-sdk/artifacts/contract-abis/Decryption.json
leventdem: Difference here is that we are preparing calldata also in this crate, in fhevm-js, abi was used only to check if handle is allowed

sdk/rust-sdk/src/blockchain/bindings.rs
leventdem: copy paste from relayer, it was just to check the event signature IIRC

sdk/rust-sdk/src/decryption/user.rs
leventdem: ok, is there any specific check to do in this case or not at all ?

sdk/rust-sdk/src/decryption/user.rs
leventdem: I will keep this for later. I will create an issue.

leventdem: Here is the [issue](https://github.com/zama-ai/fhevm/issues/216)

sdk/rust-sdk/README.md
leventdem: Implemented

sdk/rust-sdk/Cargo.toml
leventdem: package name will be now gateway-sdk; basically we are preparing everything for request and response to gateway in this lib

sdk/rust-sdk/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+tfhe = { version = "~1.1.3", features = ["boolean", "integer", "zk-pok"] }
+hex = "0.4.3"
+alloy = "1.0.5"
+rand = "0.9.1"
+serde_json = "1.0.140"
+futures-util = "0.3.31"
+tokio = "1.44.2"
+crypto_box = "0.9.1"
+kms = { git = "ssh://git@github.com/zama-ai/kms-core.git", tag = "v0.11.0-rc15", default-features = false, features = ["insecure"] }
+kms-grpc = { git = "ssh://git@github.com/zama-ai/kms-core.git", tag = "v0.11.0-rc15", default-features = false, features = ["insecure"] }
```
fd0r, line 34: Do we know when these will be open source?

fd0r: Also are we sure we need the `insecure` feature ?

sdk/rust-sdk/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Features
+
+- **Encrypted Input Creation**: Generate encrypted inputs with zero-knowledge proofs
+- **Multiple Data Types**: Support for various encrypted types (ebool, euint8, euint16, euint32, euint64, euint128, eaddress, euint256)
+- **Chain Support**: Compatible with multiple EVM blockchains
+- **Flexible Configuration**: Builder pattern for easy SDK configuration
+- **EIP-712 Support**: Tools for generating and verifying signatures
+
+## Quick Start
```
fd0r: worth a small line about needing the access to some currently private repos?

sdk/rust-sdk/README.md
fd0r: Is it used under the hood?
Or is the documentation outdated?

sdk/rust-sdk/examples/input.rs
```diff
@@ -0,0 +1,181 @@
...
+        Arc::new(public_key.clone()),
+        Arc::new(crs.clone()),
+        chain_id,
+    );
+
+    // Add a boolean value
+    input_builder.add_bool(true)?;
+
+    // Generate the encrypted value with proof
+    let encrypted_value = input_builder.encrypt_and_prove_for(contract_address, user_address)?;
```
fd0r, line 42: nice!

sdk/rust-sdk/README.md
leventdem: Maybe on usage you were thinking differently, I created structures, but still keep the old way of adding contract name with addresses. Code is [here ](https://github.com/zama-ai/fhevm/pull/52/files#diff-d004a61ce37e517e7b6a025587dd421b530f57bb953233e6c2419a394fb48a8e)

Structs are defined [here ](https://github.com/zama-ai/fhevm/pull/52/files#diff-d004a61ce37e517e7b6a025587dd421b530f57bb953233e6c2419a394fb48a8e)

I have added checks and meaningfull messages. 

However, you are right, let me also add clear setters.

fd0r (approved):

sdk/rust-sdk/Cargo.toml
leventdem: End of this or next month. for insecure I will verify it.

PR #55, opened 2025-05-22
fix(gateway-contracts): add support to ID counters and function params propagation in mock contracts

isaacdecoded:
Closes zama-ai/fhevm-internal#154


fd0r (approved):
lgtm

PR #58, opened 2025-05-23
build(kms-connector): solve submodule paths issue

isaacdecoded:
This PR solves the issue with the gateway contracts and the kms-core dependencies.


dartdart26 (approved):
Looks good, thank you!

PR #59, opened 2025-05-23
test(coprocessor): fhevm-listner, test ACL events

rudy-6-4:


PacificYield (approved):
LGTM. We should adjust to have a similar split for all events in all Solidity from host-contracts for consistency purposes.

PR #60, opened 2025-05-23
chore(charts): add charts workflows and cleanup

tawadaa:


charts/fhevm-coprocessor/templates/coprocessor-httpz-listener-deployment.yaml
isaacdecoded, line 1: Not sure but maybe we'd want to replace "httpz" with "fhevm" in this file name and in the below one.

tawadaa: good catch, thanks!

isaacdecoded (approved):
LGTM! 👏

PR #67, opened 2025-05-23
test(gateway-contracts): add malicious scenario on public decryption responses

isaacdecoded:
Closes zama-ai/fhevm-internal#176 

This PR tests the situations in which a KMS signer with a valid signature but a malicious public decryption share is "isolated" given the usage of the EIP712 message's digest.

### DIFF
--- a/gateway-contracts/test/Decryption.ts
+++ b/gateway-contracts/test/Decryption.ts
@@ -526,6 +526,62 @@ describe("Decryption", function () {
       ).to.be.revertedWithCustomError(decryption, "EnforcedPause");
     });
 
+    it("Should public decrypt with 3 valid and 1 malicious signatures", async function () {
+      // Request public decryption
+      await decryption.publicDecryptionRequest(ctHandles);
+
+      const decryptionAddress = await decryption.getAddress();
+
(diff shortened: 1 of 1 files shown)


gateway-contracts/test/Decryption.ts
```diff
@@ -1276,6 +1333,73 @@ describe("Decryption", function () {
       ).to.be.revertedWithCustomError(decryption, "EnforcedPause");
     });
 
+    /*
```
isaacdecoded: For user decryption response consensus, we don't consider the "digest" as the `userDecryptedShares` are expected to be different for each KMS node, hence we can't "isolate" malicious responses like we do for `input verification responses` and `public decryption responses`.

Do you think this is a concern we should address in Gateway? 🤔 cc @RomanBredehoft @jatZama

melanciani: yes you are right there isn't a real notion of "malicious" got user decryption within the gateway, only the user is able to check if the shares are right when reconstructing

melanciani (changes requested):
all is good, you can removed the commented part for user decryption

melanciani (approved):
thanks !

PR #72, opened 2025-05-26
chore(gateway-contracts): remove proxy addresses directory

isaacdecoded:
Closes zama-ai/fhevm-internal#160


gateway-contracts/tasks/addHostChains.ts
```diff
@@ -1,50 +1,40 @@
...
+  }
 
-    let proxyAddress: string;
-    if (taskArgs.useInternalGatewayConfigAddress) {
-      const parsedEnvGatewayConfig = dotenv.parse(fs.readFileSync("addresses/.env.gateway_config"));
-      proxyAddress = parsedEnvGatewayConfig.GATEWAY_CONFIG_ADDRESS;
-    } else {
-      proxyAddress = getRequiredEnvVar("GATEWAY_CONFIG_ADDRESS");
-    }
+  const parsedEnvGatewayConfig = dotenv.parse(fs.readFileSync("addresses/.env.gateway_config"));
```
melanciani: I would check with @tawadaa to make sure it's ok to revert this change : as the associated comments explain (in docs/env var file/task file), the fact that we read the env var by default was to make deployment tasks only consider env vars handled by ZWS 

I agree that it is now safe(r) to consider the addresses in `addresses/` since there are no default values anymore, but still please sync with her on the subject

melanciani (changes requested):
please with with @tawadaa regarding env vars, the rest looks ok

gateway-contracts/tasks/addHostChains.ts
tawadaa: @RomanBredehoft @isaacdecoded I confirm, we still need the env var in stagings, as it is the safe and solid way to get the address using a configMap in k8s cluster.
We can keep the env var, and add some script that will get value from /addresses and put it in the `GATEWAY_CONFIG_ADDRESS` (similar approach as configMap that we fill from the /addresses during contracts deployment)

isaacdecoded: Thanks @tawadaa, I'll rollback the env variable and the conditional above then 👍

gateway-contracts/.env.example
```diff
@@ -105,12 +105,3 @@ HOST_CHAIN_WEBSITE_3="https://host-chain-2028.com" # (string)
...
 FHE_PARAMS_DIGEST="0xd61fa44c57b6e7526a46e1072d63332ab4eb38e050fe54dd51fe995f9055c354" # (bytes32)
-
-# GatewayConfig Address 
-# This (static) address is needed for properly running the docker-compose locally, in particular 
-# when adding host chains to the GatewayConfig contract as a separate service
-# In a proper staging/deployment setting, this env var will be dynamically set after running the 
-# service that deploys the contracts
-# For tests, this address is read directly from the `addresses/.env.gateway_config` file, so the value in
-# this file is ignored
-GATEWAY_CONFIG_ADDRESS="0xC7D45661a345eC5cA0e8521CFEF7e32FDA0Daa68" # (address)
```
isaacdecoded, line 116: @RomanBredehoft I'm not sure about putting back or removing this env variable 🤔 I mean, this is no longer required for running the docker-compose locally (because it now uses a docker volume) but at the same time it's an env variable used for deployment so maybe could be good to have as a reference? wdyt?

melanciani: I'm not sure I get the question, we still use it in 'addHostChainsToGatewayConfig' 🤔

isaacdecoded: It's used only when `taskArgs.useInternalGatewayConfigAddress = false`, i.e. during deployment stage, not for running the docker-compose locally anymore. But yes, right, maybe better to keep it but adjust a little the comment just to explain this is used/required for deployment only (by ZWS, for example)

gateway-contracts/.env.example
```diff
@@ -106,11 +106,7 @@ HOST_CHAIN_WEBSITE_3="https://host-chain-2028.com" # (string)
...
-# GatewayConfig Address 
-# This (static) address is needed for properly running the docker-compose locally, in particular 
-# when adding host chains to the GatewayConfig contract as a separate service
-# In a proper staging/deployment setting, this env var will be dynamically set after running the 
-# service that deploys the contracts
-# For tests, this address is read directly from the `addresses/.env.gateway_config` file, so the value in
-# this file is ignored
+# GatewayConfig contract address
+# This address is required during the production deployment, at which point this environment variable
+# is dynamically set after the service that deploys the contracts runs.
```
melanciani: small things to improve in this comment :
- we want to avoid using the notion of "service" : https://github.com/zama-ai/fhevm/issues/112. Maybe just say `after contracts are deployed`
- I would prefer if you keep the last sentence : 
> For tests, this address is read directly from the `addresses/.env.gateway_config` file, so the value in this file is ignored

gateway-contracts/docs/getting-started/deployment/env_variables.md
```diff
@@ -234,16 +234,3 @@ RPC_URL="http://127.0.0.1:8757" # (string)
 ```
 
 This is the URL of the RPC node for the fhevm gateway network.
-
-#### Local docker-compose deployment only
-
-Some additional settings are needed when deploying the contracts through docker-compose locally.
```
melanciani, line 240: actually would be great to keep this section for in the deployment section, but by re-wording it a bit : 
```
#### After deploying contracts in production

Some additional settings are needed after deploying the contracts in a production setting.

- GatewayConfig address

# ```bash
# ```

This (static) address is needed for adding host chains to the GatewayConfig contract separately. In a proper production setting, this environment variable needs to be dynamically set after deploying the contracts.
```

melanciani (changes requested):
sorry, a few things to update still !

melanciani (approved):
thanks !

PR #73, opened 2025-05-26
chore(test-suite): add fhevm/kms-connector to the stack

tawadaa:


.crates2.json
isaacdecoded, line 1: What are these `crates.toml` and `crates2.json` files for? 🤔

isaacdecoded (approved):
LGTM 👍

.crates2.json
tawadaa: they are for kms-connector

PR #80, opened 2025-05-27
chore(kms-connector): remove kms-core submodule

eudelins-zama:
Closes zama-ai/fhevm-internal#147 

Note: The `clap` and `tonic` versions have been downgraded to match the ones used by `kms-grpc`, otherwise the code cannot be built.


leventdem (approved):
Ok for me, if we opensource, just remind us that this component will fail at build until kms core is opensource.

tawadaa (approved):
LGTM

PR #82, opened 2025-05-27
fix(kms-connector): move dockerfile in kms-connector

fegmorte:


toolchain.txt
```diff
@@ -0,0 +1 @@
+1.87.0
```
tawadaa, line 1: Is it mandatory to have this file at root?
Reason I'm asking is because there is another rust project coprocessor, so we might have conflict

tawadaa (approved):
lgtm, thanks 🙏🏼

PR #83, opened 2025-05-27
feat(coprocessor): handle S3 unavailability and defer uploads jobs

goshawk-3:
Resolves: https://github.com/zama-ai/fhevm-backend/issues/413

The sns-worker now validates the S3 client by checking the following:

- S3 connectivity, auth, policy
- Required buckets availability

If any of these checks fail, the worker enters a fallback mode where:

1.  **squash_noise** computation continues to run
2.  The resulting **ct128** blobs are stored in the DB
3.  S3 upload job is deferred, enqueued in the DB
4.  Relevant alarms are triggered (logs with "action = review" )

While in this mode, the worker continuously retries to recover S3 client. Once a valid S3 client is available, it refetches pending S3 jobs from the DB and resumes the upload process.

_Also handles edge case:_
If the S3 upload succeeds but the sns-worker fails to update the DB accordingly, it results in a state inconsistency.
This commit addresses that edge case by ensuring consistency.


tawadaa (approved):

coprocessor/fhevm-engine/sns-executor/src/aws_upload.rs
```diff
@@ -70,14 +124,24 @@ pub(crate) async fn process_s3_uploads(
...
 
                 // Spawn a new task to upload the ciphertexts
                 let h = tokio::spawn(async move {
-                        if let Err(err) = upload_ciphertexts(task, &client, &pool, &conf).await {
-                            error!("Failed to upload ciphertexts: {}", err);
-                            // TODO: Implement retry-mechanism.
+                    let s = task.otel.child_span("upload_s3");
+                        if let Err(err) = upload_ciphertexts(trx_lock, task, &client, &conf).await {
+                            if let ExecutionError::S3TransientError(_) = err {
+                                ready_flag.store(false, Ordering::Release );
```
dartdart26, line 135:
```suggestion
                                ready_flag.store(false, Ordering::Release);
```

dartdart26: Maybe run cargo fmt before merging.

coprocessor/fhevm-engine/sns-executor/src/aws_upload.rs
```diff
@@ -110,170 +179,421 @@ pub(crate) async fn process_s3_uploads(
...
+                        try_resubmit(&pool, is_ready.clone(), tasks.clone(), token.clone()).await
+                            .unwrap_or_else(|err| {
+                                error!("Failed to resubmit tasks: {}", err);
+                            });
+                    }
+                }
+            }
+
+            // A regular resubmit to ensure there no left over tasks
+
```
dartdart26, line 452:
```suggestion
            // A regular resubmit to ensure there no left over tasks
```

coprocessor/fhevm-engine/sns-executor/src/executor.rs
```diff
@@ -309,11 +321,20 @@ async fn update_ciphertext128(
...
                 "
                 UPDATE ciphertexts
                 SET ciphertext128 = $1
-                WHERE handle = $2;",
+                WHERE handle = $2
+                AND EXISTS (
+                    SELECT 1
+                    FROM ciphertext_digest
+                    WHERE handle = $2
+                        AND ciphertext128 IS NULL
```
dartdart26, line 336:
```suggestion
                    AND ciphertext128 IS NULL
```

dartdart26 (approved):
Looks good!
 
Just have a question regarding:
> Also handles edge case:
> If the S3 upload succeeds but the sns-worker fails to update the DB accordingly, it results in a state inconsistency.
> This commit addresses that edge case by ensuring consistency.

By failure to update the DB, do you mean failed to enqueue the pending S3 upload in the DB? I was thinking another way would be to mark everything as pending in the DB and only mark as non pending if it succeeds. But I guess that's more load on the DB.

Would your solution work if the DB fails and **then** the process terminates (before the DB being back up again)? Would that have side effects on the system state?

goshawk-3:
> By failure to update the DB, do you mean failed to enqueue the pending S3 upload in the DB? I was thinking another way would be to mark everything as pending in the DB and only mark as non pending if it succeeds. But I guess that's more load on the DB.
> 

The current approach guarantees that if a ct128 is computed and inserted into the DB then an uploading task is enqueued (inserted) in the DB too ( ref `task.enqueue_upload_task(..)` calls).  If the S3 upload is successful only then we try to mark it as succeeded in the DB.
`By failure to update the DB`, I meant that the action `mark as succeeded in the DB` might be missed due to any type of service/DB failure.

> Would your solution work if the DB fails and **then** the process terminates (before the DB being back up again)? Would that have side effects on the system state?

Yes, that's one of the scenarios taken into account. At any intermediate state of the DB, we have sufficient information to either recompute the ct128 (if it's not been inserted yet) or re-upload the ct128 (in case it's been computed but not uploaded).

coprocessor/fhevm-engine/sns-executor/src/aws_upload.rs
goshawk-3: Strange that the IDE didn't fmt. I'll fix this with the next PR to avoid waiting for the CI.

PR #88, opened 2025-05-27
chore(coprocessor): share data (cts) blobs between tokio tasks with Arc

goshawk-3:
In case of FheBytes256 type the `uncompressed_ct128` is 32.0 MiB and the `compressed_ct64` is 8 KiB.

This PR avoids cloning these large blobs when sharing them between the execute task and the uploader.


dartdart26 (approved):
Looks good!

PR #100, opened 2025-05-28
chore(gateway-contracts): upgrade forge version and bindings

eudelins-zama:
Upgrade forge version to the latest stable one (v1.2.1).

This is useful to improve the Gateway Rust bindings, because they now automatically derive the `serde::Serialize` and `serde::Deserialize` traits.

Note: one should ensure to run `foundryup` on his local machine if he wants to be able to generate the same bindings (but that should not be an issue right now as we do not have pre-commit check anymore)


isaacdecoded (approved):
LGTM! 👍

melanciani (approved):

melanciani:
> (but that should not be an issue right now as we do not have pre-commit check anymore)

can we have pre-commit checks for specific directories ?

eudelins-zama:
> > (but that should not be an issue right now as we do not have pre-commit check anymore)
> 
> can we have pre-commit checks for specific directories ?

There are existing (but not widely used) tools to achieve this like [gookme](https://lmaxence.github.io/gookme/), but I do not know if we plan to use such tools

PR #110, opened 2025-05-29
ci(gateway-contracts): add solidity coverage

isaacdecoded:
This PR enables the solidity coverage in the CI workflow, but also intends to increase the test coverage by including missed test cases and removing some unused methods.

Closes zama-ai/fhevm-internal#174


.github/workflows/gateway-contracts-hardhat-tests.yml
```diff
@@ -53,3 +53,7 @@ jobs:
       - name: Run hardhat tests
         working-directory: gateway-contracts
         run: make test
+
+      - name: Run coverage
+        working-directory: gateway-contracts
+        run: npm run coverage
```
isaacdecoded: I think we'd like to upload the coverage result somewhere (e.g. OpenZeppelin uses [codecov](https://about.codecov.io/)), right? And then add the coverage badge in the README.md 🤔

cc @RomanBredehoft

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded, line 1: @RomanBredehoft, please note that I'm removing a few unused methods because the coverage result was being impacted since we don't have tests for them, but I'd like to double-check with you if one of them is planned to be used in an upcoming implementation/used from external use cases so we need to include the tests instead.

gateway-contracts/Makefile
```diff
@@ -75,7 +78,7 @@ conformance: prettier update-bindings update-mocks update-selectors
 
 # Make sure we only use allowed licenses for dependencies
 check-licenses:
-	output=$$(npx license-checker --onlyAllow '0BSD; Apache-2.0; BSD-2-Clause; BSD-3-Clause; CC-BY-3.0; CC0-1.0; ISC; MIT; MPL-2.0; Python-2.0; WTFPL' 2>&1); \
+	output=$$(npx license-checker --onlyAllow '0BSD; BSD; Apache-2.0; BSD-2-Clause; BSD-3-Clause; CC-BY-3.0; CC0-1.0; ISC; MIT; MPL-2.0; Python-2.0; WTFPL; PSF; LGPL-3.0' 2>&1); \
```
isaacdecoded: I had to add these three new licenses because of some `solidity-coverage` dependencies:

```
Package "web3-utils@1.10.4" is licensed under "LGPL-3.0"
Package "difflib@0.2.4" is licensed under "PSF"
Package "estraverse@1.9.3" is licensed under "BSD"
```

I've been reading about them and it seems they are compatible with GNU General Public License (GPL). @RomanBredehoft do you know if we can use these licenses or better to ask somebody else?

melanciani: we cannot use `Package "web3-utils@1.10.4" is licensed under "LGPL-3.0"` no, but since this is for tooling, better to ask in working tools

also `BSD` is not really a license iirc so you need to check the exact one (probably BSD-2-Clause) to make sure it's ok

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -312,12 +304,6 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         return $.coprocessorSignerAddresses;
     }
 
-    /// @dev See {IGatewayConfig-getHostChain}.
-    function getHostChain(uint256 index) external view virtual returns (HostChain memory) {
```
melanciani: we should keep this one, it's the regular getter as if hostChains was a a public state var

.github/workflows/gateway-contracts-hardhat-tests.yml
melanciani: I would ask @PacificYield about what they do for coverage in host contracts and @aquint-zama if the coverage badge is something we want !

gateway-contracts/contracts/GatewayConfig.sol
melanciani: I answered as a comment below yes

.github/workflows/gateway-contracts-hardhat-tests.yml
aquint-zama: badge is up to you but I think it's a great addition to the repo.
I would setup the coverage so that it's tracked in PR, goal being making sure every PR increases coverage (to make sure tests are added).

gateway-contracts/Makefile
isaacdecoded: True 🤔 [the estraverse repo](https://github.com/estools/estraverse?tab=BSD-2-Clause-1-ov-file#readme) says it's `BSD-2-Clause` but when running the `npx license-checker` it shows just `BSD`:

```
├─ estraverse@1.9.3
│  ├─ licenses: BSD
│  ├─ repository: https://github.com/estools/estraverse
│  ├─ path: /Users/iherrera/Zama/fhevm/gateway-contracts/node_modules/estraverse
│  └─ licenseFile: /Users/iherrera/Zama/fhevm/gateway-contracts/node_modules/estraverse/LICENSE.BSD
```

Anyway, I think this is not an issue for this `estraverse` package then but for the other two. I'll ask in the working-tools channel.

melanciani: yes i've seen several license like this. I would say, the same for gpl, like Daniel said, better to set `estraverse` as an exeception then puting BSD as a wildcard

isaacdecoded: Adding [this thread](https://zama-ai.slack.com/archives/C01022E8SRJ/p1748943082850819) where we asked Rand about the licenses above.

.github/workflows/gateway-contracts-hardhat-tests.yml
PacificYield: Most of the unit tests are written using foundry for host contracts so it would be a bit different but badges would be a really good addition

Jermy0817:
that‘s useful for me！thank you

gateway-contracts/Makefile
melanciani: coming back to this , even if their use is fine for tools, we should avoid having these licenses as wildcards like this, and instead add the dependencies as exceptions one by one. Else we won't catch issues with future licenses. So : 
- `PSF` is fine (if I recall)
- `BSD` : as mentioned above, it's not a real license, so would be better to add the associated library as an exception
- `LGPL-3.0`: it's the most important one, so same as for `BSD``

however I know that sometimes these licenses come from several sub-dependencies and it can be annoying to have to add them all one by one. If that's the case here (like more than 5/10), tell me and maybe we can think of another approach

gateway-contracts/test/GatewayConfig.ts
```diff
@@ -425,30 +423,68 @@ describe("GatewayConfig", function () {
         }
       });
 
+      it("Should not be registered as coprocessors transaction senders", async function () {
+        await expect(gatewayConfig.checkIsCoprocessorTxSender(hre.ethers.ZeroAddress))
```
melanciani: would prefer that you define and use a `fakeTxSender` at the top and re-use it below, like in other tests. I believe using these kind of common variables makes debugging easier in the future

melanciani (changes requested):
looks good ! except: would be much better to avoid license wildcards for some !

also, how long does `make coverage` take ?

.github/workflows/gateway-contracts-hardhat-tests.yml
melanciani: regarding the badge maube create an issue somwhere and we'll see later !

isaacdecoded: Here's the tracking issue https://github.com/zama-ai/fhevm-internal/issues/194

gateway-contracts/Makefile
isaacdecoded: Great, added as excluded packages in [the last commit](https://github.com/zama-ai/fhevm/pull/110/commits/55c2acf2c03a151a8a16c4becea11250c76a00af#diff-c7090c99029bd5f438e9accb9a7b4b8905da84c0001a9f5d0aec640d28b18c50R92-R97).

isaacdecoded:
> also, how long does `make coverage` take ?

It takes as long as all the tests take, basically. I mean, the coverage executes all the tests to measure the total lines of code covered.

gateway-contracts/Makefile
```diff
@@ -89,8 +89,12 @@ ensure-addresses:
 	ENV_PATH=$(ENV_PATH) npx ts-node scripts/ensure_proxy_addresses.ts
 
 # Make sure we only use allowed licenses for dependencies
+# The following packages are excluded based on their licenses incompatibility:
```
melanciani: I would prefer something more specific, like : 
```
# Make sure we only use allowed licenses for dependencies
# Full list of SPDX identifiers can be found here: https://spdx.org/licenses/
# The following packages are exceptionally excluded from the check:
# - estraverse@1.9.3: invalid SPDX identifier (`BSD`) but licensed under `BSD-2-Clause` (https://www.npmjs.com/package/estraverse)
# - source-map@0.2.0: invalid SPDX identifier (`BSD`) but licensed under `BSD-3-Clause` (https://www.npmjs.com/package/source-map)
# - web3-utils@1.10.4: licensed under `LGPL-3.0 license` (https://www.npmjs.com/package/web3-utils), which we only allow for tooling
```

melanciani (approved):
awesome thanks a lot !

PR #116, opened 2025-06-02
chore(coprocessor): support for global localstack in txn-sender

dartdart26:
Ability to avoid creating a localstack container for every test. Instead, an already running container is used, which is started either manually (for local development) or by CI.

This makes tests run faster.

### DIFF
--- a/coprocessor/fhevm-engine/transaction-sender/tests/allow_handle_tests.rs
+++ b/coprocessor/fhevm-engine/transaction-sender/tests/allow_handle_tests.rs
@@ -198,7 +198,10 @@ async fn retry_on_transport_error(#[case] signer_type: SignerType) -> anyhow::Re
         ..Default::default()
     };
 
-    let mut env = TestEnvironment::new_with_config(signer_type, conf.clone()).await?;
+    let force_per_test_localstack: bool = false;
+    let mut env =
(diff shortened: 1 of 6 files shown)


coprocessor/fhevm-engine/transaction-sender/tests/allow_handle_tests.rs
```diff
@@ -294,7 +297,10 @@ async fn retry_on_aws_kms_error(#[case] signer_type: SignerType) -> anyhow::Resu
         ..Default::default()
     };
 
-    let mut env = TestEnvironment::new_with_config(signer_type, conf.clone()).await?;
+    let force_per_test_localstack = true;
```
goshawk-3, line 300: For my benefit, why do we need to enforce the `localstack` here instead of using the global one?

goshawk-3 (approved):
LGTM

coprocessor/fhevm-engine/transaction-sender/tests/allow_handle_tests.rs
dartdart26: Because these tests actually stop the localstack to test what happens sender behaves properly. when AWS is down (or networking is down).

I made it such that they stop their own instance and not the global one as we don't have a handle to the global one. And it looks clearer this way in terms of possible issues.

PR #117, opened 2025-06-02
docs(common): add project structure

tawadaa:


leventdem (approved):
You can also add sdk folder.

- A multi-language SDK for interacting with the host chain and gateway.

tawadaa:
> You can also add sdk folder.
> 
> * A multi-language SDK for interacting with the host chain and gateway.

Let's add it when the PR for sdk is merged

PR #122, opened 2025-06-02
feat(gateway-contracts): introduce the custodians as part of the backup feature

isaacdecoded:
This PR only introduces the custodians' metadata and initialization. As part of the backup feature, there are other incoming implementations regarding the different phases of the backup logic:

1. Setup/Update custodians phase (implies a KMS context refresh)
2. Key backup phase (receives signatures and emit backup completed event)
3. Recover shares phase (requires multi sig)
4. Dry run phase

Closes zama-ai/fhevm-internal#137


gateway-contracts/contracts/shared/Structs.sol
```diff
@@ -29,6 +29,18 @@ struct Coprocessor {
...
+/// @notice Struct that represents a custodian
+struct Custodian {
+    /// @notice ECDSA public key for verifying EIP712 signatures.
+    bytes verificationKey;
+    /// @notice Post-quantum secure public key used for encrypting symmetric key shares during backup.
+    bytes encryptionKey;
+    /// @notice Address of the custodian's transaction sender.
+    address txSenderAddress;
+    /// @notice Address of the custodian's signer (used for signing inputs with EIP712 signatures).
+    address signerAddress;
```
isaacdecoded, line 37: This is not clear if we should have a `signerAddress` (I asked Tore also) but reading the "[Recover shares](https://github.com/zama-ai/tech-spec/blob/51f7f8e0c5e3b7efd3d0dfbf89dc83405175b411/architecture/backup_keygen.md#recover-shares)" section from the tech spec it seems we should:
> We assume there is an authenticated channel between the, potentially new, KMS cores (i.e. the people/companies running them) and the custodians, __either through signatures on the request__

So, I think we'll need it to validate EIP712 signatures at the "Recover shares phase"

wdyt @RomanBredehoft?

melanciani: I don't think so, the paragraph you refer to is talking about off-chain communication. But better to let @dd23 confirm this !

dd23: We need to make sure that recovery request come from the correct and known Custodians.

I think there are (at least) two options for this, if we go via gateway: 
1. We look for a transaction from a known Custodian address (which implies the actual Custodian made the tx), so something like a "Custodian white list"
2. we have a dedicated EIP712 signature on the recovery request

Maybe even both. Not sure what's easier/preferable.

An alternative to that (that we might want to do at first, before the way via gateway is ready) could be a manual process, where the Custodians talk directly to the MPC parties. If we trust the channel, we would not need signatures here, but we do if the gateway forwards things.

isaacdecoded: I see 🤔 So, it seems we don't need a `signerAddress` for the initial backup logic in this PR, since the communication from custodians should be off-chain (via trusted channel). Thanks @dd23

gateway-contracts/contracts/shared/Structs.sol
```diff
@@ -29,6 +29,16 @@ struct Coprocessor {
     string s3BucketUrl;
 }
 
+/// @notice Struct that represents a custodian
+struct Custodian {
+    /// @notice ECDSA public key for verifying EIP712 signatures.
+    bytes verificationKey;
```
melanciani: couldn't this be an address instead @jot2re @dd23 ? 

since I believe we'll use this to verify eip712 signatures on-chain, it's much easier and robust for us to consider a public address directly (we do this for kms nodes and coprocessor as well, and call them "signerAddress"

melanciani (approved):
looks great ! just have one question for the MPC team to clarify before we can merge this

melanciani (changes requested):
actually, regarding what we've discussed in https://github.com/zama-ai/fhevm/pull/142, we now would need to consider versioning the contracts and anticipate upgrading them

gateway-contracts/contracts/shared/Structs.sol
jot2re: So for now I believe an address is enough since the public key can be derived from the signature for ecdsa. 

If/when we end up using PQ signature this will not be the case. But this is not planned for now :)

isaacdecoded: Great, so let's align it with the `signerAddress`. Thanks both of you! 🙏

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -22,7 +22,7 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
     /// @dev they can still define their own private constants with the same name.
     string private constant CONTRACT_NAME = "GatewayConfig";
     uint256 private constant MAJOR_VERSION = 0;
-    uint256 private constant MINOR_VERSION = 1;
+    uint256 private constant MINOR_VERSION = 2;
```
isaacdecoded, line 25: @RomanBredehoft, here's the minor version bump. But given [the ongoing thread](https://zama-ai.slack.com/archives/C088TGFV264/p1749649264576049), maybe we should wait for a decision to be made? 🤔

gateway-contracts/contracts/shared/Structs.sol
```diff
@@ -29,6 +29,16 @@ struct Coprocessor {
...
 }
 
+/// @notice Struct that represents a custodian
+struct Custodian {
+    /// @notice Post-quantum secure public key used for encrypting symmetric key shares during backup.
+    bytes encryptionKey;
+    /// @notice Address of the custodian's transaction sender.
+    address txSenderAddress;
+    /// @notice Address of the custodian's signer (used for signing inputs with EIP712 signatures)
+    address signerAddress;
```
isaacdecoded, line 37: @RomanBredehoft `verificationKey` replaced by a `signerAddress`

gateway-contracts/hardhat.config.ts
```diff
@@ -19,7 +19,7 @@ import "./tasks/upgradeProxy";
 const dotenvConfigPath: string = process.env.DOTENV_CONFIG_PATH || "./.env";
 dotenv.config({ path: resolve(__dirname, dotenvConfigPath) });
 
-const NUM_ACCOUNTS = 20;
+export const NUM_ACCOUNTS = 30;
```
isaacdecoded, line 22: @RomanBredehoft note that this number needed to be increased due to the introduced custodian tx senders and signers.

gateway-contracts/contracts/GatewayConfig.sol
melanciani: yes let's wait for a conclusion here indeed

melanciani (changes requested):
sorry, a few things has changed, in the end : 
- https://github.com/zama-ai/fhevm/issues/247 will need to be done before 
- as updated in https://github.com/zama-ai/fhevm/issues/222, we actually won't consider multiple files for different versions, only the latest one

melanciani (changes requested):
also please update the test suite and make sure sure e2e tests pass with your changes

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -139,7 +147,39 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
+        if (custodians.length == 0) {
+            revert EmptyCustodians();
+        }
+
+        /// @dev Register the custodians
+        for (uint256 i = 0; i < custodians.length; i++) {
+            $.custodians[custodians[i].txSenderAddress] = custodians[i];
+        }
+
+        emit Reinitialization(custodians);
```
isaacdecoded: @RomanBredehoft, do you think we should emit such an event on reinitialization? 🤔

gateway-contracts/test/GatewayConfig.ts
```diff
@@ -422,6 +471,18 @@ describe("GatewayConfig", function () {
       });
     });
 
+    describe("Reinitialization", function () {
```
isaacdecoded: I'm not completely sure about adding these test cases. I mean, it means we will need to always change them with every new version we implement, right? 🤔

wdyt? @RomanBredehoft

gateway-contracts/contracts/GatewayConfig.sol
melanciani: I think it makes sense yes, generally any time we modify the state it's better to emit an event for better tracking I would say

gateway-contracts/contracts/examples/GatewayConfigUpgradedExample.sol
```diff
@@ -10,7 +10,7 @@ contract GatewayConfigUpgradedExample is GatewayConfig {
 
     /// @notice Version of the contract
     uint256 private constant MAJOR_VERSION = 0;
-    uint256 private constant MINOR_VERSION = 2;
+    uint256 private constant MINOR_VERSION = 3;
```
melanciani: this will be annoying to update everytime we update the initial contract 🤔  wdyt ?

maybe we could create these examples by using `MAJOR_VERSION = 1000` for V2, 1001 for V3 ect. Could you create an issue to do somethign about this in a following PR please ?

gateway-contracts/contracts/GatewayConfig.sol
melanciani: but I would instead name it `ReinitializationV2`

melanciani: actually, something I thought recently : 
- maybe we should rename our `Initialization` events to `InitializeGatewayConfig`, ect, in order to make it more clear
- and in that case, rename the above to `ReinitializeGatewayConfigV2`

wdyt ? if you agree, could you please create an issue for the initi part so that we do it in a following PR ?

gateway-contracts/test/GatewayConfig.ts
melanciani: hmm, I wouldn't expect this to be possible actually, we probably want to add a check that forbids calling the the `reinitializeV{i}` on a contract that was initialized using `initializeFromEmptyProxy` for the same version : https://github.com/zama-ai/fhevm/issues/261

but down the line, yes imo it still makes sense but more as part of https://github.com/zama-ai/fhevm/issues/222. So probably in a separate file in a following PR 

the reason is : 
- reinitializers (may) have errors , so better to test these
- throughtout the next weeks, V2 will be improved and might contain even more logic, so better to make sure it still works

and yes indeed we'll update them with new versions this is not an issue imo, we know upgrades are pretty custom

melanciani (changes requested):
sorry, realized we are missing a safeguard regarding updates, better to consider it before continuing this PR imo : https://github.com/zama-ai/fhevm/issues/261

wdyt ?

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Agree, better to be more specific. Here's the follow-up issue #265.

gateway-contracts/contracts/examples/GatewayConfigUpgradedExample.sol
isaacdecoded: Yes, indeed, is it ok if I consider this change in the current WIP PR #142 instead? @RomanBredehoft

melanciani: yes great !

melanciani: you probably want to merge https://github.com/zama-ai/fhevm/pull/142 and rebase on main then !

gateway-contracts/test/GatewayConfig.ts
melanciani: done in https://github.com/zama-ai/fhevm/pull/279

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -331,6 +361,12 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         return $.hostChains;
     }
 
+    /// @dev See {IGatewayConfig-getCustodian}.
+    function getCustodian(address custodianTxSenderAddress) external view virtual returns (Custodian memory) {
```
melanciani, line 397: I think we also want a `getAllCustodians` method like for other operators. You can add a new state var for that

gateway-contracts/test/GatewayConfig.ts
```diff
@@ -330,6 +379,7 @@ describe("GatewayConfig", function () {
               userDecryptionThreshold,
               kmsNodes,
               coprocessors,
+              custodians,
             ],
           },
         }),
```
melanciani, line 388: you also need a test for `getCustodian` + the upcoming `getCustodians` !

melanciani (changes requested):
thanks ! I spotted a few missing details + you might want to merge the upgrade PR + rebase

isaacdecoded:
> thanks ! I spotted a few missing details + you might want to merge the upgrade PR + rebase

Missed methods and tests for custodians pushed in [the last commit](https://github.com/zama-ai/fhevm/pull/122/commits/54ad98b4e6bf428713a86eed0af81561f874c10c) 🙏

gateway-contracts/contracts/GatewayConfig.sol
melanciani: @isaacdecoded great for the other methods, but I think having a `getAllCustodians()` would also be great (and will follow what we do with other components) !

isaacdecoded: Ah, I thought you referred to the dedicated methods for txSenders, signers, etc.

> will follow what we do with other components

What do you mean by this? I mean, KMS nodes and Coprocessors don't have getters for the full lists of metadata 🤔

melanciani: @isaacdecoded ah ! my bad I must have added it by error in the context PRs , then alright no need to do it then should be fine sorry

melanciani (approved):
alright, thanks a lot !!

PR #142, opened 2025-06-03
test(gateway-contracts): include upgrade test for enhanced storage layout case

### LINKED ISSUE #124: test(gateway-contracts): improve upgrade tests
### Describe the issue

we need to improve tests to make sure state is persisted between upgrades

we can also maybe consider custom logic in the upgrade func and/or task for better coverage

### Context

_No response_

_No response_

isaacdecoded:
This PR intends not only to include the case when an upgraded contract contains storage state changes, but to refactor some global variables to gain type-checking. Using the "this" property caused these variables to lose their type.

Note that the `hardhat.upgrades` library doesn't allow for changes on the ERC-7201 storage variable names or types between versions, failing, for example, with:

When modifying the variable type:
```
Error: New storage layout is incompatible
contracts/examples/GatewayConfigV3Example.sol:38: Upgraded `userDecryptionThreshold` to an incompatible type
  - Bad upgrade from uint256 to string
```

When modifying  the variable name:
```
Error: New storage layout is incompatible
contracts/examples/GatewayConfigUpgradedExample.sol:33: Renamed `protocolMetadata` to `protocolMetadataV2`
```

When modifying the fields of a struct:
```
Error: New storage layout is incompatible
contracts/examples/GatewayConfigUpgradedExample.sol:33: Upgraded `protocolMetadata` to an incompatible type
  - Bad upgrade from struct ProtocolMetadata to struct GatewayConfigUpgradedExample.ProtocolMetadataV2
  - In struct GatewayConfigUpgradedExample.ProtocolMetadataV2
    - Added `newField`
```

Closes zama-ai/fhevm-internal#131


gateway-contracts/test/upgrades/upgrades.ts
```diff
@@ -1,135 +1,166 @@
...
-    this.inputVerificationFactory = await ethers.getContractFactory("InputVerification", this.owner);
-    this.inputVerificationFactoryUpgraded = await ethers.getContractFactory(
-      "InputVerificationUpgradedExample",
-      this.owner,
-    );
+    owner = new Wallet(process.env.DEPLOYER_PRIVATE_KEY!).connect(ethers.provider);
+    emptyUUPSFactory = await ethers.getContractFactory("EmptyUUPSProxy", owner);
+
+    ciphertextCommitsFactoryV1 = await ethers.getContractFactory("CiphertextCommits", owner);
+    ciphertextCommitsFactoryV2 = await ethers.getContractFactory("CiphertextCommitsUpgradedExample", owner);
```
isaacdecoded: Maybe we can rename the current `/contracts/examples/...UpgradedExample` contracts too, for example:

`CiphertextCommitsUpgradedExample` -> `CiphertextCommitsV2Example`

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
```diff
@@ -0,0 +1,108 @@
...
+    constructor() {
+        _disableInitializers();
+    }
+
+    function initialize(string calldata newField) public virtual reinitializer(3) {
+        __Ownable_init(owner());
+        __Pausable_init();
+
+        // Execute the migration logic to set the new field in the protocol metadata
+        _migrate(newField);
```
melanciani, line 61: are you sure the _migrate is no already called by default when upgrading the contract ?

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    function getProtocolMetadata() external view virtual returns (ProtocolMetadataV2 memory) {
```
melanciani, line 64: the initial goal of this test is also to make sure the state properly remains the same : ie for counters, the v2/c3 contracts does not start the counter at 0 (here we're not talking about changing the var's name of whatever, just making sure we do not lose the state) 

btw if you prefer and if it's not too annoying to do , you could define a completely new contract as a v1 with simple logic and states and run your test(s) with it (by defining a V2 based on it)

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
```diff
@@ -0,0 +1,108 @@
...
+        address[] kmsSignerAddresses;
+        uint256 mpcThreshold;
+        uint256 publicDecryptionThreshold;
+        uint256 userDecryptionThreshold;
+        mapping(address coprocessorTxSenderAddress => Coprocessor coprocessor) coprocessors;
+        address[] coprocessorTxSenderAddresses;
+        address[] coprocessorSignerAddresses;
+        HostChain[] hostChains;
+        // New state variables added in the upgraded version
+        ProtocolMetadataV2 protocolMetadataV2;
```
melanciani, line 44: I've read your comment in the PR but I'm surprised it's not possible to remove / update `ProtocolMetadata` ... because that means we'll need to keep all the old structs defined and as a state var even though they'll never be used buy the new versions 🤔

gateway-contracts/test/upgrades/upgrades.ts
melanciani: yes that makes sense to me

melanciani (changes requested):
a few comments, but also could you maybe check with @PacificYield (and maybe @jatZama) if this is the only way we can upgrade storage in contracts ?

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
jatZama: Of course you cannot simply change the old ProtocolMetadata and swap it, this would wreck storage layout in an upgrade. Only solution is to append new variables like what was done by Isaac. I already mentionned the Rareskills blog for everything related to proxies and delegatecall to understand why btw.

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
jatZama: This seems correct for me, not sure what you mean.

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
isaacdecoded: Not sure if I understand well, but I'd say that this "state remaining" test you mention is already covered when we execute the `_migrate` function. I mean, the test is checking that, after the upgrade, the `ProtocolMetadataV2` (contract V3) remains the same as the `ProtocolMetadata` (contract V2) plus the new field added in V3, i.e. we are not losing the previous `name` and `website` values from `ProtocolMetadata` 🤔

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
isaacdecoded: AFAIU, the `hardhat.upgrades()` doesn't call any method by default during the upgrade, but only if you specify it in the `opts` argument.

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
isaacdecoded: The behavior I tested is what Joseph-André confirms. And yes, I'm also concerned about "accumulating" old unused variables in storage but it seems to be common in upgradeable contracts.

There's another pattern I've reading about that defines separated storage layouts per contract version, e.g. `erc7201:fhevm_gateway.storage.CiphertextCommitsV1`, `erc7201:fhevm_gateway.storage.CiphertextCommitsV2`, etc., but that means we need to execute a "hard" migration from one storage layout to the other, and I haven't tested it tbh 🤔

melanciani: > There's another pattern I've reading about that defines separated storage layouts per contract version, e.g. erc7201:fhevm_gateway.storage.CiphertextCommitsV1, erc7201:fhevm_gateway.storage.CiphertextCommitsV2, etc., but that means we need to execute a "hard" migration from one storage layout to the other, and I haven't tested it tbh 🤔

this is similar to what I had in mind when I was mentioning "copying" data from the old to the new storage yes, imo we would need to do this at one point, ex : when we'll add contexts, we'll add and/or remove _a lot_ of state vars from the gateway config contract so I really think this could be a great situation to have that

jatZama: As I said on slack, normally you should use same first slot during upgrade (meaning same namespace).
if you need additional new state you append them at the end, but previous state should still be present after upgrade. You cannot migrate potentially millions of data in a mapping for instance, so it does not make sense to change namespace at upgrade.

melanciani: @jatZama yes ok that's what I feared, but this is really annoying then : for ex, with the new context feature in the gateway, we won't use many state var in the gatewayConfig anymore , but that means we'll still keep them forever even if we'll never use them .. ?

jatZama: Exactly, the old variables would still exist in the namespace mapping, even if unused, the new ones will appear at the end of the namespace mapping, this is the widely adopted convention.

jatZama: Note that it won't impact gas price, so it is not so much annoying, and you can rename the old state variables, for example by appending _old suffix or something.

melanciani: alright well I guess we have no choice and we'll go with that, ie : 
- append new state vars at the end
- keep and rename the old ones (with `_old` at the beginning)
- migrate data with a dedicated logic if needed and possible

isaacdecoded: Ok, let's go with that then. Although about renaming state variables, I'm not sure it's possible, as mentioned in the PR description, the `hre.upgrade` library throws an error when a state variable is renamed, this is another example I just tested:
```
Error: New storage layout is incompatible

contracts/examples/GatewayConfigV3Example.sol:26: Renamed `pauser` to `old_pauser`
```

So I think we can't deprecate states by renaming them, but adding NatSpec comments like: 

`/// @custom:deprecated`

jatZama: Adding natspecs could silent some errors in the upgrade plugin. The important thing to keep in mind is to keep both same order and types for old variables, names change should have no impact if you know what you are doing.

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
```diff
@@ -0,0 +1,108 @@
...
+
+    /// @custom:storage-location erc7201:fhevm_gateway.storage.GatewayConfig
+    struct GatewayConfigStorage {
+        address pauser;
+        mapping(address kmsTxSenderAddress => bool isKmsTxSender) _isKmsTxSender;
+        mapping(address kmsSignerAddress => bool isKmsSigner) _isKmsSigner;
+        mapping(address coprocessorTxSenderAddress => bool isCoprocessorTxSender) _isCoprocessorTxSender;
+        mapping(address coprocessorSignerAddress => bool isCoprocessorSigner) _isCoprocessorSigner;
+        mapping(uint256 chainId => bool isRegistered) _isHostChainRegistered;
+        ProtocolMetadata protocolMetadata;
```
melanciani: I think to better reflect what we'll do, you should rename this state var by `_oldProtocolMetadata`

melanciani: btw it would be much clearer if these old state vars are the only one using underscores at the beginning : https://github.com/zama-ai/fhevm/issues/239

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
```diff
@@ -0,0 +1,108 @@
...
+        // Execute the migration logic to set the new field in the protocol metadata
+        _migrate(newField);
+    }
+
+    function getProtocolMetadata() external view virtual returns (ProtocolMetadataV2 memory) {
+        GatewayConfigStorage storage $ = _getGatewayConfigStorage();
+        return $.protocolMetadataV2;
+    }
+
+    function getCoprocessor(address coprocessorTxSenderAddress) external view virtual returns (Coprocessor memory) {
```
melanciani: not sure why this is needed 🤔

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
```diff
@@ -0,0 +1,108 @@
...
+        GatewayConfigStorage storage $ = _getGatewayConfigStorage();
+        return $.coprocessors[coprocessorTxSenderAddress];
+    }
+
+    /// @notice Getter for the name and version of the contract
+    /// @return string representing the name and the version of the contract
+    function getVersion() external pure virtual returns (string memory) {
```
melanciani, line 71: is it really needed to re-define this ? if not, we might want to remove it from the other implementations as well

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
```diff
@@ -0,0 +1,108 @@
...
+        GatewayConfigStorage storage $ = _getGatewayConfigStorage();
+        $.protocolMetadataV2 = ProtocolMetadataV2({
+            name: $.protocolMetadata.name,
+            website: $.protocolMetadata.website,
+            newField: newField
+        });
+    }
+
+    // solhint-disable-next-line no-empty-blocks
+    function _authorizeUpgrade(address _newImplementation) internal virtual override {}
```
melanciani, line 96: same, not sure if we need to re-define it

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
```diff
@@ -0,0 +1,108 @@
...
+
+    function _getGatewayConfigStorage() internal pure returns (GatewayConfigStorage storage $) {
```
melanciani, line 98: same, not sure if we need to re-define it

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
melanciani: yes ok I think it's fine then, thanks !

gateway-contracts/test/upgrades/upgrades.ts
```diff
@@ -141,16 +173,40 @@ describe("Upgrades", function () {
...
+  it("Should maintain state consistency after upgrade", async function () {
+    const fixtureData = await loadFixture(loadTestVariablesFixture);
+    const { gatewayConfig } = fixtureData;
+
+    // Check that GatewayConfig is at version 0.1.0
+    expect(await gatewayConfig.getVersion()).to.equal("GatewayConfig v0.1.0");
+
+    // Check that the protocol metadata is correct
+    const metadata = await gatewayConfig.getProtocolMetadata();
+    const protocolMetadata = { name: "Protocol", website: "https://protocol.com" };
```
melanciani: nit for being a bit more future proof : would be slightly better to define a 
```
const name = "Protocol"
const website = "https://protocol.com"
```
at the top of this test, and use them to avoid having to rewrite them twice

gateway-contracts/test/upgrades/upgrades.ts
```diff
@@ -141,16 +173,40 @@ describe("Upgrades", function () {
...
+    expect(metadata).to.deep.equal(toValues(protocolMetadata));
+
+    // Upgrade the GatewayConfig contract to version 0.3.0
+    const protocolNewField = "Protocol new field";
+    const gatewayConfigV2 = await upgrades.upgradeProxy(gatewayConfig, gatewayConfigFactoryV3, {
```
melanciani: I think for clarity, either we : 
- also upgrade to the v2 implementation before upgrade to v3 (to confirm that double upgrades work right)
- or we simply do no define a v3 and replace the actual v2 with your current v3, and update the `Should deploy upgradable GatewayConfig` test instead of creating a new one

I'm more leaning towards the first one but as you wish

melanciani (changes requested):
overall looks great, just have a few comments to clarify !

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
isaacdecoded: Yes, this is needed in order to properly reflect the new version, i.e. if we re-use the parent contract `getVersion()` method it will return that parent version instead of the new upgraded version.

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
isaacdecoded: For this case (the `GatewayConfigV3Example` contract), it's needed to define these methods since this contract is not inheriting from the original `GatewayConfig` as the other examples do. 

And the reason why it's not inheriting is because this `GatewayConfigV3Example` contains the storage layout change from `ProtocolMetadata` to `ProtocolMetadataV2` which impacts the `getProtocolMetadata` method that cannot be overridden because the original returns `ProtocolMetadata` while the `GatewayConfigV3Example` returns `ProtocolMetadataV2`.

Same happen with the `_getGatewayConfigStorage()`, `_authorizeUpgrade()` methods.

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
isaacdecoded: Removed in last commit, I indeed forgot to remove it, thanks!

gateway-contracts/test/upgrades/upgrades.ts
isaacdecoded: Added the intermediate upgrade to V2 in [last commit](https://github.com/zama-ai/fhevm/pull/142/files#diff-0d1ca356f8adcf4d83c111268dae544a56c9548f089c412ad7c7108b2165af73R207-R219) 👍

gateway-contracts/contracts/examples/GatewayConfigV3Example.sol
melanciani: in the end we can't really rename them so we'll just add a `// deprecated` comment like you did

gateway-contracts/test/upgrades/upgrades.ts
melanciani: in the end we'll start to use the `initFromMeptyProxy` + a single upgrade in tests, but I think it's fine to keep at least an example of double upgrade in here, so all good

melanciani (approved):
great, thanks a lot !

PR #149, opened 2025-06-04
fix(coprocessor): commit atomic txn to update db on zkproof execution

goshawk-3:
fixes: zama-ai/fhevm-internal#115


dartdart26 (approved):
Looks good!

PR #152, opened 2025-06-04
chore(gateway-contracts): fix contract deployment with empty addresses

melanciani:
closes https://github.com/zama-ai/fhevm-internal/issues/114


isaacdecoded (approved):
LGTM! 👍

melanciani:
@tawadaa are you ok with this PR ?

tawadaa (approved):
LGTM, thanks!
Note that this only persists `/addresses` from container in a volume instead of versioning them in the repo, the behavior of addresses being generated and stored in `/addresses` remains the same inside the container.

PR #154, opened 2025-06-04
feat(host-contracts): ACL pausable

PacificYield:
This PR makes the ACL pausable (similar to how it is managed with the Gateway L2 logic)


host-contracts/test/acl/acl.ts
```diff
@@ -1,57 +0,0 @@
-import { expect } from 'chai';
```
PacificYield: I removed these unit tests since the foundry tests already replicate these

host-contracts/contracts/ACL.sol
```diff
@@ -32,42 +32,53 @@ contract ACL is UUPSUpgradeable, Ownable2StepUpgradeable {
...
+
+    /// @notice Returned if the sender address is not allowed for allow operations.
+    /// @param sender Sender address.
     error SenderNotAllowed(address sender);
 
-    /// @notice         Emitted when a handle is allowed.
-    /// @param caller   account calling the allow function.
-    /// @param account  account being allowed for the handle.
-    /// @param handle   handle being allowed.
+    /// @notice Returned if the handle is not allowed for decryption.
```
jatZama: This natspec seems to not make much sense, did I miss something?

PacificYield: Fixed

host-contracts/test/acl/acl.ts
jatZama: I am against removal of those, since more tests is better in general... and I expect hardhat should still be used because we use hardhat for deployments and tasks, so in the future same test could interact easily with those parts which are already (and remain) in typescript.

jatZama: Another reason to keep those is that the gateway contracts team used tests only in hardhat for the moment, so for them it is surely easier to keep the hardhat version as well.

host-contracts/contracts/ACL.sol
```diff
@@ -97,20 +108,28 @@ contract ACL is UUPSUpgradeable, Ownable2StepUpgradeable {
     }
 
     /**
-     * @notice  Re-initializes the contract.
+     * @notice Re-initializes the contract.
      */
     /// @custom:oz-upgrades-validate-as-initializer
-    function reinitialize() public virtual reinitializer(2) {
+    function reinitialize(address initialPauser) public virtual reinitializer(2) {
```
melanciani, line 114: small question : will only the ACL contract be pausable ? I ask because in the gateway we define and set the pauser address one time in the GatewayConfig contract, and then we made all contracts pausable using this address

melanciani (approved):
looks good if only the ACL should be pausable !

host-contracts/contracts/ACL.sol
PacificYield: I think there was discussion with making the FHEVMExecutor pausable but since ACL is called in all functions of the FHEVMExecutor, it is sufficient to only make the ACL pausable.

host-contracts/test/acl/acl.ts
PacificYield: I'll put it back

host-contracts/contracts/ACL.sol
melanciani: makes sense then, thanks !

PR #159, opened 2025-06-04
feat(coprocessor): add healthcheck to gw-listener in new docker image

tawadaa:


coprocessor/fhevm-engine/gw-listener/Cargo.toml
```diff
@@ -16,6 +16,10 @@ tokio = { workspace = true }
 tokio-util = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
+axum = "0.7"
+tower-http = { version = "0.5", features = ["trace"] }
+serde = { version = "1.0", features = ["derive"] }
```
dartdart26: Could we use serde and serde_json dependencies from workspace?

coprocessor/fhevm-engine/gw-listener/Cargo.toml
```diff
@@ -16,6 +16,10 @@ tokio = { workspace = true }
 tokio-util = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
+axum = "0.7"
```
dartdart26: We do use actix_web in coprocessor. Maybe worth putting it as a workspace dependency and using that in both places? Or use axum in both places? Just to avoid having a lot of dependencies for the same thing.

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
```diff
@@ -29,6 +31,10 @@ struct Conf {
 
     #[arg(long, default_value = "10")]
     error_sleep_max_secs: u16,
+
+    /// HTTP server port for health checks
+    #[arg(long, default_value_t = 8080)]
+    http_port: u16,
```
dartdart26: Nit: Maybe can name it health_check_port or similar to be explicit.

dartdart26 (approved):
Looks good!

Added some comments about dependencies.

coprocessor/fhevm-engine/gw-listener/Cargo.toml
tawadaa: I noticed actix_web is only used in metrics.rs in coprocessor (tfhe-worker), I decided to switch to axum, as it's simpler and minilamist

Issue #132, opened 2025-06-03
chore(kms-connector): update docker config


eudelins-zama:
### Describe the issue

Following the significant architecture change of the `kms-connector` proposed in #90, the docker configuration of the connector must be updated

PR #161, opened 2025-06-05
feat(kms-connector): database design

eudelins-zama:
Closes zama-ai/fhevm-internal#128

### Note

I reorganized the kms-connector folder into a cargo workspace for the future components and kept the "old" connector codebase into the "simple-connector" package. I also adapted the Dockerfile so we can still pass the CI using the "old" connector codebase.

### DB design choice

I decided to create a table/notification for each kind of event. Some tables could be merged (user and public decryptions for example) if it is needed, but I was not sure it would be useful, let me know what you think @maksymsur.


kms-connector/connector-db/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Use the Rust image as the base
+FROM rust:1.87.0-bookworm
+
+# Install required runtime dependencies only
+RUN cargo install sqlx-cli --version 0.8.6 --no-default-features --features postgres --locked && \
+    groupadd -r zama && useradd -r -g zama zama
+
+COPY connector-db/init_db.sh /init_db.sh
+COPY connector-db/migrations /migrations
```
maksymsur, line 9: I think it may be optimized with caching. Perhaps ask @fegmorte for guidance as we have more or less standardized approach to this

kms-connector/connector-db/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      context: ..
+      dockerfile: connector-db/Dockerfile
+    environment:
+      DATABASE_URL: postgresql://postgres:postgres@kms-connector-db:5432/kms-connector
+    depends_on:
+      kms-connector-db:
+        condition: service_healthy
+
+volumes:
+  db:
```
maksymsur: NL

kms-connector/connector-db/migrations/20250604084502_gw_events.sql
```diff
@@ -0,0 +1,75 @@
...
+    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (pre_ksk_id)
+);
+
+CREATE TABLE IF NOT EXISTS crsgen_requests (
+    crsgen_request_id BYTEA NOT NULL,
+    fhe_params_digest BYTEA NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (crsgen_request_id)
+);
```
maksymsur: NL

maksymsur (approved):
LGTM as soon as you handle compilation errors!
Let's proceed and see how far we get with this approach given io frequencies.

kms-connector/connector-db/Dockerfile
eudelins-zama: I relied on the [Dockerfile](https://github.com/zama-ai/fhevm/blob/main/coprocessor/fhevm-engine/fhevm-db/Dockerfile) of the coprocessor DB

Should we update both then?

maksymsur: I'd suggest towards **Yes** as it improves build time

eudelins-zama: I suggest we tackle in a branch resolving this issue https://github.com/zama-ai/fhevm/issues/132 instead of this one. Wdyt?

maksymsur: agreed

PR #106, opened 2025-05-29
feat(gateway-contracts): implement KMS context

melanciani:
still a WIP 

done : 
- new `KmsContexts` contract with full logic (except resharing : https://github.com/zama-ai/fhevm-internal/issues/151)
- compatible with new migration mechanism

to be done : 
- update tests
- final alignements with coprocessor contexts
- clean and comment code
- update doc

closes https://github.com/zama-ai/fhevm-internal/issues/171

PR #163, opened 2025-06-05
chore(gateway-contracts): update mock script, makefile and forge version

melanciani:
Small PR that extracts general fixes from https://github.com/zama-ai/fhevm/pull/106 

In particular for the mock script : 
- allows a function to emit several events in mock contracts
- anticipate the use of a new shared file (`enums.sol`) for contexts

Also, in scripts : 
- I needed to introduce a new script that makes sure contract addresses are the matching the default ones we have in the dev env. Else, following fixes introduced in https://github.com/zama-ai/fhevm-internal/issues/117, the rust binding check could fail if the local addresses used at compilation (ex: in `make conformance) are different from the ones generated in the CI (the default ones)
- I updated forge version to `1.2.2`, which can be updated locally by running `foundryup`

Finally in the makefile : 
- `.prettierignore` makes `prettier` ignore the rust bindings, making `make conformance` much quicker
- add a `make lint-binding` for better local checks
- add a `deploy-empty-proxies` for locally create the address files
- add a `get-contracts-sized` to get the contracts' size easily

### DIFF
--- a/.github/workflows/gateway-contracts-integrity-checks.yml
+++ b/.github/workflows/gateway-contracts-integrity-checks.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@de808b1eea699e761c404bda44ba8f21aba30b2c # v1.3.1
         with:
-          version: v1.2.1
+          version: v1.2.2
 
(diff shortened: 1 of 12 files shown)


eudelins-zama (approved):
LGTM

PR #164, opened 2025-06-05
feat(host-contracts): add support for msgValue in FHE.requestDecryption

PacificYield:


immortal-tofu (approved):

jatZama:
I think test with payable variant is missing and ideally should be added at some point with a payable variant of oracle contract (or open an issue if you dont want to add this test now).
Also I would prefer a refactor of current code, instead of having almost duplicated code for both `requestDecryption` functions, I would prefer to see `requestDecryption(
        bytes32[] memory ctsHandles,
        bytes4 callbackSelector )` that should just be calling `requestDecryption(
        bytes32[] memory ctsHandles,
        bytes4 callbackSelector,
        0
    )` wdyt?

PacificYield:
> I think test with payable variant is missing and ideally should be added at some point with a payable variant of oracle contract (or open an issue if you dont want to add this test now). Also I would prefer a refactor of current code, instead of having almost duplicated code for both `requestDecryption` functions, I would prefer to see `requestDecryption( bytes32[] memory ctsHandles, bytes4 callbackSelector )` that should just be calling `requestDecryption( bytes32[] memory ctsHandles, bytes4 callbackSelector, 0 )` wdyt?

Yes, it would be much cleaner indeed. I adjusted accordingly! Thanks for the feedback. For the tests, I created an issue to do it in a later PR

jatZama (approved):

PR #178, opened 2025-06-05
chore(coprocessor): bump alloy version to 1.0.9

goshawk-3:


rudy-6-4:
This review could be a lot shorter if not including auto-formatting.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -192,9 +235,10 @@ impl InfiniteLogIter {
         let mut retry = 20;
         loop {
             let ws = WsConnect::new(&self.url);
-            match ProviderBuilder::new().on_ws(ws).await {
+            match ProviderBuilder::new().connect_ws(ws).await {
```
rudy-6-4, line 253: why not adding retry configuration here ?

rudy-6-4 (approved):

goshawk-3:
> This review could be a lot shorter if not including auto-formatting.

Indeed. Maybe we should  remove rustfmt.toml from the listener then? What would you suggest @rudy-6-4 ?

ref: https://github.com/zama-ai/fhevm/blob/main/coprocessor/fhevm-engine/fhevm-listener/rustfmt.toml

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
goshawk-3: Done

rudy-6-4:
> > This review could be a lot shorter if not including auto-formatting.
> 
> Indeed. Maybe we should remove rustfmt.toml from the listener then? What would you suggest @rudy-6-4 ?
> 
> ref: https://github.com/zama-ai/fhevm/blob/main/coprocessor/fhevm-engine/fhevm-listener/rustfmt.toml

Because you want shorter line ? At least it could be made consistent in coprocessor, yes.
You can use "git add -p" to avoid committing unrelated stuff and having formatting in separate commit.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -47,28 +54,53 @@ pub struct Args {
...
+        default_value = "5",
+        help = "Initial block time, refined on each block"
+    )]
     pub initial_block_time: u64,
+
+    #[arg(long, default_value = "1000000")]
+    pub provider_max_retries: u32,
+
+    #[arg(long, default_value = "4s", value_parser = parse_duration)]
+    pub provider_retry_interval: Duration,
```
dartdart26, line 82: Just to note here that when reconnection happens for the WebSocket via `eth_subscribe`, the `fromBlock` is ignored by geth and most other nodes and instead it will start streaming the latest live events.

@rudy-6-4 Will that mess up catching up? I mean **after** the catchup via `get_logs`.

dartdart26 (approved):
Great job!

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
dartdart26: We discussed that offline, all good, it is taken into consideration.

dartdart26:
Resolves https://github.com/zama-ai/fhevm-internal/issues/108

PR #196, opened 2025-06-06
docs(kms-connector): update design_docs to new architecture

eudelins-zama:
Closes zama-ai/fhevm-internal#121


kms-connector/design_docs/architecture.md
```diff
@@ -0,0 +1,159 @@
...
+- it never misses any Gateway events
+- it never misses any KMS Core responses
+- it is able to catch up events/responses in case it has been down for some time
+
+## Architecture overview
+
+In order to achieve this, the KMS Connector has been divided into 3 components:
+- **GatewayListener**
+  - Multiple listeners, so we do not miss Gateway events if one is down
+  - Each listener listens to a RPC node of the Gateway
```
maksymsur, line 19: A couple of questions here:
- what is the rationale of this, isn't this an overkill?
- how many RPC nodes will be there?
- why can't it be `N` listeners to minimize io traffic, where `N == N' needed for Arbitrum consensus`? 
- maybe it's better for the listener to be able to have a simple connection logic, like if `N1` RPC node is unhealthy - it connects to `N2/3/4/5` ... until reliable connection is established?

kms-connector/design_docs/architecture.md
```diff
@@ -0,0 +1,159 @@
...
+  - Each listener tries to write the events it catches in a Postgres DB (only 1 will succeed, the other will ignore the duplicate key error)
+ 
+- **KmsWorker**
+  - One or multiple workers
```
maksymsur, line 25: one or multiple? as it will influence the event-processing logic and related risks

kms-connector/design_docs/architecture.md
```diff
@@ -0,0 +1,159 @@
...
+  - Get notified by the Postgres DB when new events are stored
+  - Forward the events' requests to the KMS Core, and store its responses to DB
+  - Remove the events from the DB once handled
```
maksymsur, line 30: shall it be in batches? as adding/removing high sub second frequency events is not optimal for DB operational/efficiency capabilities

kms-connector/design_docs/architecture.md
```diff
@@ -0,0 +1,159 @@
...
+
+- **TransactionSender**
+  - Only one tx sender
+  - Gets notified by the Postgres DB when new KMS Core responses are stored
+  - Forwards the KMS Core response to the Gateway by submitting transactions
+  - Removes the responses from the DB once forwarded
```
maksymsur, line 36: same as https://github.com/zama-ai/fhevm/pull/196/files#r2137668482

kms-connector/design_docs/architecture.md
eudelins-zama: 1. Hmm it is what we discussed in this issue (https://github.com/zama-ai/fhevm/issues/90), not sure how we could do something more simple to avoid missing events...
2. RPC nodes -> I don't know but do not think it impacts the architecture of the kms-connector
3. It could be N listeners or it could be less or more, I do not think it is important to know exactly how many listeners we will have for now
4. It could have backup RPC nodes, but if it crashes for some reason, we must rely on other listeners to ensure we do not miss events, which is the whole point of the new architecture

kms-connector/design_docs/architecture.md
eudelins-zama: The architecture must be designed in order to support multiple worker. But I think in practice we'll have only one worker in the first place

kms-connector/design_docs/architecture.md
eudelins-zama: I think in the first place it will not be in batches to keep it simple, when the volume of decryption increase, we could optimize with batching or using Kafka or something else

kms-connector/design_docs/architecture.md
maksymsur: Ok, let's progress with this design and see it in action. Perhaps add a note to return to this design element in future when we may support scale up/down of listeners depending on event hits success rate not to overspend resources.

kms-connector/design_docs/architecture.md
maksymsur: ok, than perhaps worth mentioning that we start with one? So that it was clear that there is no config to multiply workers for now

kms-connector/design_docs/architecture.md
maksymsur: ok

kms-connector/design_docs/architecture.md
```diff
@@ -0,0 +1,159 @@
...
+
+CREATE OR REPLACE FUNCTION notify_public_decryption_response()
+    RETURNS trigger AS $$
+BEGIN
+    NOTIFY public_decryption_response_available;
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trigger_from_public_decryption_responses_insertions
```
maksymsur: `CREATE OR REPLACE TRIGGER` ?

maksymsur:
Can you describe logic about removing data from DBs? It's briefly mentioned in **Architecture overview** but not detailed. Perhaps also put some SQL queries

maksymsur (approved):
LGTM! But for small additions to description as mentioned in review

kms-connector/design_docs/architecture.md
eudelins-zama: Done in f26cbfed913451a731cb3c8ffc141a89e236fc0c

kms-connector/design_docs/architecture.md
eudelins-zama: Done in 78f38a875a7def27ca4df6921bf7649a0213c1c1

kms-connector/design_docs/architecture.md
eudelins-zama: Done in 733a3ab84c38b22ce0b87cdb61304d1e0c5bdda9

eudelins-zama:
> Can you describe logic about removing data from DBs? It's briefly mentioned in **Architecture overview** but not detailed. Perhaps also put some SQL queries

I added more queries in the **Database design** section (see 733a3ab84c38b22ce0b87cdb61304d1e0c5bdda9)

Let me know wdyt @maksymsur

kms-connector/design_docs/architecture.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+When a `KmsWorker` picks events in the DB to process them, it uses a database transaction and
+`FOR UPDATE SKIP LOCKED` in its query to lock these events from other workers. Then it can remove
+them when processed without concurrency issue:
+
+```rust
+let mut db_tx = db_pool.begin().await?;
+let event = sqlx::query!(
+    "SELECT * FROM public_decryption_requests WHERE decryption_id = $1 FOR UPDATE SKIP LOCKED",
```
maksymsur: I'd avoid using `SELECT *` here and in other places that may be a performance killer and prevents query optimizations. Also showing columns is better for self-explainability

eudelins-zama: It was just as a simple example for the documentation :wink:

maksymsur: ah... I thought it'll be used further in the codebase. If Not - ignore this comment

eudelins-zama: I'll update so we do not forget

eudelins-zama: Done!

PR #201, opened 2025-06-09
chore(test-suite): pass properly channels to sns-worker

goshawk-3:
fixes https://github.com/zama-ai/fhevm-internal/issues/96


rudy-6-4 (approved):
pg-listen-channels taking 2 strings is error prone, maybe it could be made foolproof

goshawk-3:
> foolproof

`pg-listen-channels` accepts an array of strings and that is a trivial config.

PR #205, opened 2025-06-10
feat(coprocessor): fhevm-listner, also recheck unseen blocks

rudy-6-4:


coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -200,46 +200,55 @@ impl InfiniteLogIter {
...
             return false;
         };
         let Some(event) = &self.prev_event else {
             return false;
         };
-        let Some(block) = event.block_number else {
+        let Some(from_block) = event.block_number else {
             return false;
         };
+        let to_block_inclusive = match &self.current_event {
```
rudy-6-4, line 214: due to timeout this will not work

PR #211, opened 2025-06-10
feat(gateway-contracts): implement coprocessor contexts

melanciani:
todo: 
- put back contextId in extraData

what have been done : 
- main logic (cleaned, with code comments)
    - add context ID in `VerifyProofRequest` event as a new field 
    - add context ID in `CiphertextVerification` struct' as a new field
    - add contexId in `VerifyProofResponse` event as a new field 
        - also add missing `extraData` field (same for `RejectProofResponse`)
- tests
- documentation
- compatible with new migration logic
- add verify task for new contract
- add contract upgrade tests in CI for updated contracts
- add `getConsensusS3BucketUrls` in `CiphertextCommits` 
- include coprocessor s3 bucket URLs in decryption request event's extraData field
- clean some eip712 structs from contracts (refs https://github.com/zama-ai/fhevm-internal/issues/403)

also updated spec : https://github.com/zama-ai/tech-spec/pull/292

closes https://github.com/zama-ai/fhevm-internal/issues/439
closes https://github.com/zama-ai/fhevm-internal/issues/172


gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -1,17 +1,20 @@
 // SPDX-License-Identifier: BSD-3-Clause-Clear
 pragma solidity ^0.8.24;
 import { gatewayConfigAddress } from "../addresses/GatewayConfigAddress.sol";
+import { coprocessorContextsAddress } from "../addresses/CoprocessorContextsAddress.sol";
 import { kmsManagementAddress } from "../addresses/KmsManagementAddress.sol";
 import { Ownable2StepUpgradeable } from "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 import { Strings } from "@openzeppelin/contracts/utils/Strings.sol";
 import "./interfaces/ICiphertextCommits.sol";
 import "./interfaces/IGatewayConfig.sol";
+import { ICoprocessorContexts } from "./interfaces/ICoprocessorContexts.sol";
```
isaacdecoded, line 12: Why not import like interfaces above? i.e. `import "./interfaces/ICoprocessorContexts.sol";`

gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -1,17 +1,20 @@
...
 import "./interfaces/IKmsManagement.sol";
 import "./shared/UUPSUpgradeableEmptyProxy.sol";
 import "./shared/GatewayConfigChecks.sol";
 import "./shared/Pausable.sol";
 import "./libraries/HandleOps.sol";
-
+import { CoprocessorContexts } from "./CoprocessorContexts.sol";
```
isaacdecoded: Is this import needed if already importing its interface `ICoprocessorContexts` above?

gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -1,17 +1,20 @@
...
+import { ContextChecks } from "./shared/ContextChecks.sol";
```
isaacdecoded, line 18: Or import it the same as others above from "/shared", i.e. `import "./shared/ContextChecks.sol";`

gateway-contracts/contracts/CoprocessorContexts.sol
```diff
@@ -0,0 +1,539 @@
...
+     * @dev See {ICoprocessorContexts-refreshCoprocessorContextStatuses}.
+     */
+    function refreshCoprocessorContextStatuses() external virtual whenNotPaused {
+        CoprocessorContextsStorage storage $ = _getCoprocessorContextsStorage();
+
+        // If there is a pre-activation coprocessor context, then check if it is time to activate it
+        // and thus suspend the current active coprocessor context
+        uint256 preActivationContextId = $.coprocessorContextLifecycle.preActivationContextId;
+        if (preActivationContextId != 0) {
+            if (block.number >= $.coprocessorContextActivationBlockNumber[preActivationContextId]) {
```
isaacdecoded: Maybe these two conditionals can be merged into a single one with `&&`

gateway-contracts/contracts/CoprocessorContexts.sol
```diff
@@ -0,0 +1,539 @@
...
+                // Set the new active coprocessor context
+                ContextLifecycle.setActive($.coprocessorContextLifecycle, preActivationContextId);
+                emit ActivateCoprocessorContext(preActivationContextId);
+            }
+        }
+
+        // If there is a suspended coprocessor context, then check if it is time to deactivate it
+        uint256 suspendedContextId = _getSuspendedCoprocessorContextId();
+        if (suspendedContextId != 0) {
+            if (block.number >= $.coprocessorContextDeactivatedBlockNumber[suspendedContextId]) {
```
isaacdecoded: Same as the comment above.

gateway-contracts/contracts/CiphertextCommits.sol
melanciani: I think we should start doing like this instead because it often starts to get hard to know from which files/libraries which objects are coming from

gateway-contracts/contracts/CiphertextCommits.sol
melanciani: nope good catch

gateway-contracts/contracts/CiphertextCommits.sol
melanciani: same as above, I think we should prefer this

gateway-contracts/contracts/CoprocessorContexts.sol
melanciani: indeed !

isaacdecoded (approved):
What a great job! 👏

Do you think we should open an issue to unify the `imports` at some point? 🤔

melanciani:
> Do you think we should open an issue to unify the imports at some point? 🤔

I'll let you create it yes, it's not urgent at all though no worries

dartdart26:
@RomanBredehoft Should the `CiphertextVerification` struct include a `context_id` now?

gateway-contracts/contracts/libraries/ContextLifecycle.sol
```diff
@@ -0,0 +1,403 @@
...
+        // Only an active context ID can be set as suspended
+        if (!isActive($, contextId)) {
+            revert ContextNotActive(contextId);
+        }
+
+        $.contextStatuses[contextId] = ContextStatus.Suspended;
+        $.suspendedContextId = contextId;
+
+        // Reset the active context ID as it is now suspended
+        $.activeContextId = 0;
```
melanciani, line 238: open question : in practice, we should always suspend a context while activating a new one, so I hesitated to remove the `setActive` and `setSuspended` and replace it by a single `setActiveAndSuspend` (for ex) to remove the possibility to (internally) be able to ever have an activeContextId set to 0 ! This is because this rule will most likely always be true for any kind of contexts (KMS, copro, custodians, even keys)

note that the invariant of always having an active key remains true in the `CoprocessorContext` contract because we call `setActive` and `setSuspended` at the same time

gateway-contracts/contracts/CoprocessorContexts.sol
```diff
@@ -0,0 +1,561 @@
...
+        CoprocessorContextsStorage storage $ = _getCoprocessorContextsStorage();
+        if (status == ContextStatus.Active) {
+            ContextLifecycle.setActive($.coprocessorContextLifecycle, contextId);
+            emit ActivateCoprocessorContext(contextId);
+
+            // Set the suspended block timestamp to the current block timestamp (i.e., the status update
+            // is immediate)
+            // ⚠️ This should be used with caution as it will create a state where no context is active,
+            // which should never happen. Please make sure to activate a context right after. ⚠️
+        } else if (status == ContextStatus.Suspended) {
```
melanciani: this is the only way we can break the invariant of always having an active context, but I think we should still allow this in case of emergency (to avoid blocking any unwanted behaviors). Wdyt ? cc @jot2re @jatZama

jot2re: Won't this functionally be the same a pausing? 
Basically in my view, simply setting a context to suspended is in my mind kinda what we want to do in order to selectively pause. So it looks fine to me as long as it can always be put back into active

melanciani: I guess so, I made sure that a almost anythign will be reverted if active context ID is null

> So it looks fine to me as long as it can always be put back into active

this would be diectly possible through the `moveSuspendedCoprocessorContextToActive` method, where we specifically make sure to allow an active context Id to be null (to not block unexpected behaviors): https://github.com/zama-ai/fhevm/pull/211/files#diff-5314455b58acef26c2dcf572ee6eed263edc104eaba6aa20e66a653621c7574dR308

still, maybe we could find a better way idk

gateway-contracts/contracts/libraries/ContextLifecycle.sol
jatZama: I agree for security it is better to merge them

gateway-contracts/contracts/libraries/ContextLifecycle.sol
```diff
@@ -0,0 +1,403 @@
...
+library ContextLifecycle {
+    /**
+     * @notice The storage struct for the context lifecycle.
+     * This struct keeps track of all context statuses, as well as some of the IDs per status
+     */
+    struct ContextLifecycleStorage {
+        /// @notice The status of each context
+        mapping(uint256 => ContextStatus) contextStatuses;
+        /// @notice The ID of the context that is generating (at most one)
+        uint256 generatingContextId;
```
jatZama, line 20: do you really need uint256 variables here for all the ids? you could have saved gas by using storage packing with smaller types maybe.

gateway-contracts/contracts/interfaces/IInputVerification.sol
```diff
@@ -27,10 +29,34 @@ interface IInputVerification {
     /**
      * @notice Emitted once a correct ZK Proof verification is completed.
      * @param zkProofId The ID of the ZK Proof.
+     * @param coprocessorContextId The ID of the coprocessor context.
      * @param ctHandles The coprocessor's computed ciphertext handles.
      * @param signatures The coprocessor's signature.
      */
-    event VerifyProofResponse(uint256 indexed zkProofId, bytes32[] ctHandles, bytes[] signatures);
+    event VerifyProofResponse(
```
melanciani, line 40: I'm surprised there is no extraData here, although it's listened by the relayer cc @isaacdecoded @jatZama

gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -144,11 +180,19 @@ contract CiphertextCommits is
         // sender address will still be added in the list
         $._coprocessorTxSenderAddresses[addCiphertextHash].push(msg.sender);
 
+        // Get and store the coprocessor's S3 bucket URL for the ciphertext material addition
+        // The comment above regarding the coprocessor transaction sender addresses also applies here.
+        string memory coprocessorS3BucketUrl = COPROCESSOR_CONTEXTS
+            .getCoprocessorFromContext(contextId, msg.sender)
+            .s3BucketUrl;
+        $.consensusCoprocessorS3BucketUrls[addCiphertextHash].push(coprocessorS3BucketUrl);
```
jatZama: why do you push this to storage each time? seems to cost a lot of gas

melanciani: we need to: each url is different, copro should be aware of all of them who participated in the consensus (ex in `getConsensusS3BucketUrls`method)

gateway-contracts/contracts/libraries/ContextLifecycle.sol
melanciani: no indeed, but this won't happen very often so I think it's fine

gateway-contracts/contracts/CiphertextCommits.sol
jatZama: but this url is supposed to be constant and proper to the context, so no need to push it again for each handle, this seems really wasteful.

gateway-contracts/contracts/interfaces/ICoprocessorContexts.sol
```diff
@@ -0,0 +1,339 @@
...
+     * @param contextId The ID of the coprocessor context.
+     * @param deactivatedBlockTimestamp The block timestamp at which the coprocessor context will be deactivated.
+     */
+    event SuspendCoprocessorContext(uint256 contextId, uint256 deactivatedBlockTimestamp);
+
+    /**
+     * @notice Emitted when a coprocessor context gets activated.
+     * @param contextId The ID of the coprocessor context.
+     */
+    event ActivateCoprocessorContext(uint256 contextId);
```
dartdart26, line 56: How do `PreActivateCoprocessorContext` and `ActivateCoprocessorContext` interact with each other? If the block timestamp is reached, do we need `ActivateCoprocessorContext`? What if `ActivateCoprocessorContext` is received before the block timestamp?

gateway-contracts/contracts/shared/Structs.sol
```diff
@@ -20,7 +20,9 @@ struct KmsNode {
 }
 
 /// @notice Struct that represents a coprocessor
-struct Coprocessor {
+struct CoprocessorV2 {
+    /// @notice Name of the coprocessor, as a human-readable identifier
+    string name;
     /// @notice Address of the coprocessor's transaction sender
     address txSenderAddress;
```
dartdart26, line 35: Is that the unique identifier that a coprocessor needs to look for? I mean if a coprocessor finds its `txnSenderAddress`, it would update other data based on it?

gateway-contracts/contracts/interfaces/IInputVerification.sol
```diff
@@ -27,10 +29,34 @@ interface IInputVerification {
...
+    error InvalidCoprocessorContextProofVerification(uint256 zkProofId, uint256 contextId, ContextStatus contextStatus);
+
+    /**
+     * @notice Error indicating that the coprocessor context is no longer valid for rejecting the ZK Proof.
+     * A context is valid if it is active or suspended.
+     * @param zkProofId The ID of the ZK Proof.
+     * @param contextId The context ID of the coprocessor.
+     * @param contextStatus The status of the coprocessor context.
+     */
+    error InvalidCoprocessorContextProofRejection(uint256 zkProofId, uint256 contextId, ContextStatus contextStatus);
```
dartdart26, line 71: @isaacdecoded I guess we need to consider these as non-errors and not retry them in the txn-sender as they won't ever succeed, right?

dartdart26: We had an offline discussion with @isaacdecoded and a context can be re-enabled later. In theory, if we do retry these errors, they might succeed. But it makes more sense to reject and don't retry as at the time of the txn initially, the context was not active. That's IMO, at least.

gateway-contracts/contracts/shared/Structs.sol
```diff
@@ -29,6 +31,26 @@ struct Coprocessor {
     string s3BucketUrl;
```
dartdart26: Does it make sense to not call this s3 bucket URL, but`ciphertextUrl` or something like that? That makes it more generic, should we want to use Azure or whatever. Suggested by @antoniupop 

Btw, we could use the `featureSet` and know that from a specific feature set onwards, we start using say a different storage provider and the `ciphertextUrl` means something different.

gateway-contracts/contracts/shared/Structs.sol
dartdart26: What would happen if `signerAddress` is no longer correct in relation to the secret key of the given coprocessor?

gateway-contracts/.env.example
```diff
@@ -52,21 +52,27 @@ KMS_TX_SENDER_ADDRESS_3="0xc45994e4098271c3140117ebD5c74C70dd56D9cd" # accounts[
...
+
 # Coprocessors
 # The number of coprocessors must be lower or equal to the number of coprocessors' metadata defined below
 NUM_COPROCESSORS="3" # (number)
 
 # Coprocessor 1
+COPROCESSOR_NAME_0="Coprocessor 1" # (string)
 COPROCESSOR_TX_SENDER_ADDRESS_0="0x6518D50aDc9036Df37119eA465a8159E34417E2E" # accounts[11] (address)
 COPROCESSOR_SIGNER_ADDRESS_0="0xa5eE8292dA52d8234248709F3E217ffEBA5E8312" # accounts[12] (address)
 COPROCESSOR_S3_BUCKET_URL_0="s3://bucket-1" # (string)
```
jatZama: I think we should remove all mention of S3 with something else and more neutral, in case we want to use another Cloud service. For instance, Petar was suggesting to replace every mention of `S3BucketUrls` by `CiphertextsBaseUrl`. See https://zama-ai.slack.com/archives/C0320EL7Y4U/p1757432197810599?thread_ts=1757075669.170999&cid=C0320EL7Y4U

gateway-contracts/contracts/CiphertextCommits.sol
isaacdecoded: This URL duplication storage was removed here in [this PR](https://github.com/zama-ai/fhevm/pull/869).

gateway-contracts/contracts/interfaces/ICoprocessorContexts.sol
melanciani: @dartdart26 order is : 
- emit `PreActivateCoprocessorContext` 
- wait (at least) activation block timestamp 
- emit `ActivateCoprocessorContext` 

the activation event is emitted following this logic: https://github.com/zama-ai/fhevm/pull/211/files#diff-5314455b58acef26c2dcf572ee6eed263edc104eaba6aa20e66a653621c7574dR221

(note that the event won't be emitted until a addCt/allow/inputProof request/response is called, so in practice it could be emitted some time after the block timestamp)

> What if ActivateCoprocessorContext is received before the block timestamp?

this should not be possible, unless you're referring to how [arbitrum timestamps work](https://docs.arbitrum.io/build-decentralized-apps/arbitrum-vs-ethereum/block-numbers-and-time#block-timestamps-arbitrum-vs-ethereum) ?

gateway-contracts/contracts/shared/Structs.sol
melanciani: I believe we should trust the operators and consider that they provide the right information. fwiu, the (new) coprocessor(s) should already be up and running when we add a next context in pre-activation mode (the block timestamp is then mostly used for syncronizing everything)

so I don't think there's anything to verify/get in the coprocessor when adding a new copro context in the gateway

as for the identifier, I think it's best that we consider `signerAddress` instead, in case we handle multiple `txnSenderAddress` one day (not a big issue though since both addresses are the same for copros currently )

gateway-contracts/contracts/interfaces/IInputVerification.sol
melanciani: yes most of the time the context won't be re-enabled, so there it does not make sense to retry on this error 

we only re-enable a context in case of emergency, but this should be very rare

gateway-contracts/contracts/interfaces/IInputVerification.sol
melanciani: added the field (same for RejectProofResponse)

gateway-contracts/contracts/interfaces/ICoprocessorContexts.sol
dartdart26: So before `ActivateCoprocessorContext` is emitted, the previous context ID is valid? Meaning addCt/allow/inputProof would succeed with the previous context ID?

gateway-contracts/contracts/shared/Structs.sol
dartdart26: OK, so `signerAddress` is the way to identify that this is for a particular coprocessor as it updates its context.

gateway-contracts/contracts/interfaces/IInputVerification.sol
dartdart26: Yeah, there are other errors too for CiphertextCommits that we don't need to retry on, but I will add handling for them in separate PR in transaction-sender in the coprocessor.

gateway-contracts/contracts/shared/Structs.sol
```diff
@@ -95,3 +121,13 @@ struct DelegationAccounts {
     /// @notice The address of the account that receives the delegation.
     address delegatedAddress;
 }
+
+// ----------------------------------------------------------------------------------------------
+// DEPRECATED
```
isaacdecoded, line 136: Should we definitely remove this regarding the state reset? @RomanBredehoft

gateway-contracts/contracts/CiphertextCommits.sol
isaacdecoded: Should we also remove these deprecated states? Similar to GatewayConfig [here](https://github.com/zama-ai/fhevm/pull/211/commits/f768df3edfe9ad6bbe7a2ea2b3d8959f1c07a1f3#diff-4139d8055b558914fe9fc072c5a6a27496114f8e6bc3e081b09d321f2528dd59L88).

gateway-contracts/contracts/KmsManagement.sol
isaacdecoded, line 1: Maybe these changes could be avoided since this contract is rebranded in the [keygen feature branch](https://github.com/zama-ai/fhevm/blob/feat/keygen_via_gateway/gateway-contracts/contracts/KMSManagement.sol).

isaacdecoded (approved):
LGTM! 💯

gateway-contracts/contracts/KmsManagement.sol
melanciani: I'll rebase/update the feature branch once keygen is merged anyway yes

gateway-contracts/contracts/CiphertextCommits.sol
melanciani: since copro context is for fhevm 0.10, I think we need to keep it backward compat

gateway-contracts/contracts/mocks/KMSGenerationMock.sol
```diff
@@ -40,8 +40,7 @@ contract KMSGenerationMock {
     }
 
     function prepKeygenResponse(uint256 prepKeygenId, bytes calldata signature) external {
-        keyCounter++;
-        uint256 keyId = keyCounter;
+        uint256 keyId;
```
eudelins-zama, line 43: Why this has changed? (and real KMSGeneration contract has not changed)

It breaks one of the connector test because the id is 0 instead of the incremented key counter. (Easily fixable, but want to understand why this has changed since I already adapted my test in the other way in the past)

melanciani: basically `keyCounter++` shouldn't have been,  in `prepKeygenResponse` in the first place, the script shold only use a counter when it's defined in the very function. In our case it's in `keygen` , not `prepKeygenResponse`, hence the change 

so I just fixed this as it was created issues with the coprocessor context contract

melanciani: so yes this change is expected !

melanciani:
closing this for now as we'll continue work on this later !

PR #224, opened 2025-06-11
docs(common): introduction

immortal-tofu:


yuxizama (approved):
Hey @immortal-tofu 
Thanks for the update! 
I think there are some content duplications for architecture. I would break them down and regroup these info in a better way! And there are some types/links to fix
Let's merge this and I'll polish this part in another PR.

NijasR:
Thanks for the update!

PR #231, opened 2025-06-11
fix(kms-connector): fix build & lint errors

eudelins-zama:
According to `reqwest` [docs](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.timeout), there is no timeout by default, so I just remove the problematic line.

Also, clippy was raising some errors due to [large enums](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant) so I fix that.


maksymsur (approved):
LGTM!

PR #238, opened 2025-06-12
refactor(library-solidity): config addresses

PacificYield:


immortal-tofu (approved):

library-solidity/config/ZamaConfig.sol
```diff
@@ -0,0 +1,72 @@
...
+        return
+            FHEVMConfigStruct({
+                ACLAddress: 0x687820221192C5B662b25367F70076A37bc79b6c,
+                FHEVMExecutorAddress: 0x848B0066793BcC60346Da1F49049357399B8D595,
+                KMSVerifierAddress: 0x1364cBBf2cDF5032C47d8226a6f6FBD2AFCDacAC,
+                InputVerifierAddress: 0xbc91f3daD1A5F19F8390c400196e58073B6a0BC4
+            });
+    }
+
+    function getSepoliaDecryptionOracleAddress() internal pure returns (address) {
```
jatZama: This is wrong. Oracle contract address should be imported from oracle package, and updated here https://github.com/zama-ai/fhevm-relayer/blob/main/oracle/address/ZamaOracleAddress.sol

jatZama:
I noticed an issue related to contract address, se my comment.
Next time you do a change on library-solidity, can you please assign me as a reviewer please? Because I stumbled on this issue by luck here.

library-solidity/config/ZamaConfig.sol
immortal-tofu: We've decided to simplify address to avoid importing a new package. Still, we could import the address from `@zama-fhe/oracle-solidity` 👍

jatZama: I think it is cleaner to import it from `@zama-fhe/oracle-solidity` because it is also necessary to update it there.

jatZama: When I look at the package.json in library-solidity, the oracle is already in the dependencies btw https://github.com/zama-ai/fhevm/blob/main/library-solidity/package.json#L59 so not sure current version simplifies

PacificYield: Adjusted!

jatZama (approved):
I approved for now, but please dont forget to update the zama-fhe/oracle package with correct address as well, and change the dependency here once this will be done...

0xalexbel (approved):

jatZama (approved):

PR #250, opened 2025-06-12
feat(gateway-contracts): refactor and add new argument for using env variables for proxy addresses

isaacdecoded:
Closes zama-ai/fhevm-internal#81


gateway-contracts/tasks/upgradeProxy.ts
```diff
@@ -20,40 +21,92 @@ function stripContractName(input: string): string {
...
+  await hre.run("compile:specific", { contract: getImplementationDirectory(newImplementation) });
+}
+
+async function checkImplementationArtifacts(
+  expectedArtifactName: string,
+  currentImplementation: string,
+  newImplementation: string,
+  hre: HardhatRuntimeEnvironment,
+): Promise<void> {
+  const currentImplementationArtifact = await hre.artifacts.readArtifact(currentImplementation);
```
melanciani: I'm not too familiar with `hre.artifacts.readArtifact(newImplementation);`, although the logic looks good, could you please check with @jatZama that this will still work in ZWS

maybe so, it just that iiuc, artifacts are not handle the same way there for the "old" implem

gateway-contracts/tasks/upgradeProxy.ts
```diff
@@ -20,40 +21,92 @@ function stripContractName(input: string): string {
...
+  if (currentImplementationArtifact.contractName !== expectedArtifactName) {
+    throw new Error(
+      `The current implementation artifact does not match the expected contract name "${expectedArtifactName}". Found: ${currentImplementationArtifact.contractName}`,
+    );
+  }
+
+  const newImplementationArtifact = await hre.artifacts.readArtifact(newImplementation);
```
melanciani: same for the new implementation

melanciani (changes requested):
please check with Jat that this will work for ZWS ! if so, I'm ok with these changes

gateway-contracts/tasks/upgradeProxy.ts
isaacdecoded: I'll check it with Joseph-André, but I think it should work since it's a requirement for the upgrade plugin to have the contract factories (which implies compiling both, the current and the new implementations that will generate the artifacts ofc). Otherwise, there's no way to succeed when executing `hre.upgrades.forceImport` and `hre.upgrades.upgradeProxy` since both need the contract implementation factories

Btw, this compilation is done by the `host-contracts` as well, but yes, let's double check this. Thanks for pointing it out! 👍

melanciani (approved):
ok looks good, worth adding these checks in host contracts + also, I think we should rename the file to`upgrade.ts` or `upgradeContracts.ts` since we should use these tasks on proxies !

gateway-contracts/tasks/upgradeProxy.ts
```diff
@@ -63,23 +116,53 @@ task("task:upgradeMultichainAcl")
...
+  .addOptionalParam(
+    "useInternalProxyAddress",
+    "If proxy address from the /addresses directory should be used",
+    false,
+    types.boolean,
+  )
   .addOptionalParam(
     "verifyContract",
     "Verify new implementation on Etherscan (for eg if deploying on Sepolia or Mainnet)",
     false,
```
jatZama: change this to true, to be in sync with host.

isaacdecoded: Set to true as default in [last commit](https://github.com/zama-ai/fhevm/pull/250/commits/2f29f777568eacb994b3d610468e926f34ab92e2) 👍

PR #253, opened 2025-06-12
refactor(gateway-contracts): rename initialize method for deployments from empty proxies

### LINKED ISSUE #247: feat(gateway-contracts): introduce `initializeFromEmptyProxy`
### Describe the issue

After discussion upgrades with @jatZama, we've realized that that there was no way to initialize a contract version "from scratch", which could be useful for external developers + would simplify testing 

in summary, each contract will provide : 
- a `initializeFromEmptyProxy` function : 
    - used for deploying the contract (ie, in the deployment tasks)
    - it can evolve in different versions to better suite the current state
    - it will _always_ use a `reinitializer(2)` modifier + keep the same name
- a `reinitializeV{i}` function (ex : `reinitializeV3`) :
    - starting from contract v2 (ie, the initial contract (v1) will not provide a `reinitializeV1`)
    - used for upgrading from version `{i-1}` to version `i` (ie, in upgrade tasks)
    - the name will be updated at each version, but a contract will have a single one of them (ex: contract v2 will have `reinitializeV2`, contract v3 will have `reininializeV3`). This is mostly for clarity
    - it will increment the counter in the `reinitializer` modifier each time, starting from 3 (ex: contract v2 will use `reinitialize(3)`, contract v3 will use `reinitialize(4)` 
    - it will be independent of the major/minor/patch version numbers in state, although whenever one of them is bumped, we should expect the `reinitializeV{i}` to be updated

Note: in unit tests (except the upgrade ones) we will directly init all contracts through `initializeFromEmptyProxy`

### Context

_No response_

_No response_

isaacdecoded:
Note that the renaming of the `initialize` method required for the custom NatSpec comment `@custom:oz-upgrades-validate-as-initializer` since "Reinitializers are not included in validations by default". Otherwise, the hardhat upgrade plugin fails with, for example:

```
An unexpected error occurred:

Error: Contract `contracts/GatewayConfig.sol:GatewayConfig` is not upgrade safe

contracts/GatewayConfig.sol:16: Missing initializer
    Define an initializer function and use it to call the initializers of parent contracts
    https://zpl.in/upgrades/error-001
```
For more info please refer to [this OZ docs](https://docs.openzeppelin.com/upgrades-plugins/writing-upgradeable#validating_initializers).

Closes zama-ai/fhevm-internal#78 
closes https://github.com/zama-ai/fhevm-internal/issues/84


melanciani:
we already had somekind of warning with the init anyway, but I guess it was done by default : https://github.com/zama-ai/fhevm-internal/issues/84 (i modified you PR to also close this issue as well)

melanciani (approved):
looks good to me, thanks !

PR #258, opened 2025-06-13
fix(kms-connector): increase gas_limit

eudelins-zama:


kms-connector/README.md
```diff
@@ -174,6 +174,7 @@ The KMS Connector supports flexible configuration through both TOML files and en
    export KMS_CONNECTOR_PUBLIC_DECRYPTION_TIMEOUT_SECS="300"
    export KMS_CONNECTOR_USER_DECRYPTION_TIMEOUT_SECS="300"
    export KMS_CONNECTOR_RETRY_INTERVAL_SECS="5"
+   export KMS_CONNECTOR_GAS_LIMIT = "6000"
```
maksymsur: how do you come up with exactly this number?

maksymsur:
I have doubts about the reasoning behind this PR. Setting a manual gas limit provides a strict upper bound, ensuring trxs do not exceed predictable cost boundaries. Currently, we automatically estimate gas based on trx complexity and size. By explicitly setting `KMS_CONNECTOR_GAS_LIMIT`, we’re introducing a safeguard rather than fully overriding automation. If our goal differs - like dynamically adjusting based on transaction specifics - we should explicitly derive gas limit logic based on transaction size/bytes, computational steps, or other criteria to ensure reliability and optimize costs.

maksymsur (approved):

maksymsur:
LGTM! 
Although, can you please:

1. Add notes (TODO) clearly marking this MAX buffer as a temporary workaround, explaining that the current automatic gas estimation logic is failing under our conditions. Also, you may reference the Slack conversation with gas consumption screenshots for context.
2. Insert more detailed logging (TODO to replace with proper metrics later), showing clearly:

- estimated gas before submitting the transaction,
- call-data length,
- actual gas consumed as per the transaction receipt, other relative data as needed.

This will allow us to gather accurate statistics and later refine our gas estimation logic based on real Arbitrum network behavior in our scenarios.

maksymsur (approved):
GREAT! Thanks!

PR #262, opened 2025-06-13
fix(coprocessor): fhevm-listener, unused variable

rudy-6-4:


coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -375,10 +375,9 @@ impl InfiniteLogIter {
                 }
             }
         }
-        let Some(current_event) = &self.current_event else {
-            return None;
+        if self.current_event.is_some() else {
+            self.last_block_event_count += 1;
         };
-        self.last_block_event_count += 1;
         return self.current_event.clone();
```
dartdart26: If self.current_event is none, existing code returns None without incrementing last_block_event_count and new code does. Does that have any side effects?

dartdart26 (commented):
Added a single comment, because not sure of the implications.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
rudy-6-4: +1 is only done if not None in both cases

dartdart26 (approved):

PR #263, opened 2025-06-13
feat(coprocessor): add log level configurations

dartdart26:
Add the `log-level` cmd line option to configure the log level.

Also, make sure gw-listener and transaction-sender retry connecting to the node infinitely at startup as the `WsConnect` options don't apply to connect during provider creation.

Finally, add structured logging to fhevm-listener instead of println!s. Also, added some TODOs to remove panics and, instead, propagate errors. Will be done in a separate PR.

As a side note, using structured logging in all other services where it is applicable will be done in a separate PR as well (as we usually only utilize the `message` field in the log macros).

Resolves https://github.com/zama-ai/fhevm-internal/issues/88

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
+++ b/coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
@@ -7,6 +7,7 @@ use sqlx::types::Uuid;
 use std::collections::VecDeque;
 use std::str::FromStr;
 use std::time::Duration;
+use tracing::{error, info, warn, Level};
 
 use alloy::primitives::Address;
 use alloy::providers::{Provider, ProviderBuilder, RootProvider, WsConnect};
@@ -75,6 +76,12 @@ pub struct Args {
         help = "Initial block time, refined on each block"
     )]
     pub initial_block_time: u64,
+
+    #[arg(
+        long,
+        value_parser = clap::value_parser!(Level),
+        default_value_t = Level::INFO)]
+    pub log_level: Level,
 }
 
 type RProvider = FillProvider<
@@ -143,6 +150,7 @@ impl InfiniteLogIter {
     }
 
     async fn get_chain_id_or_panic(&self) -> ChainId {
+        // TODO: remove expect and, instead, propagate the error
         let ws = WsConnect::new(&self.url);
         let provider = ProviderBuilder::new()
             .connect_ws(ws)
@@ -192,7 +200,7 @@ impl InfiniteLogIter {
 
     async fn recheck_prev_block(&mut self) -> bool {
         let Some(provider) = &self.provider else {
-            eprintln!("No provider, inconsistent state");
+            error!("No provider, inconsistent state");
             return false;
         };
         let Some(event) = &self.prev_event else {
@@ -220,10 +228,11 @@ impl InfiniteLogIter {
         if logs.len() as u64 == last_block_event_count {
             return false;
         }
-        eprintln!(
-            "Replaying Block {block} with {} events (vs {})",
-            logs.len(),
-            last_block_event_count
+        info!(
+            block = block,
+            events_count = logs.len(),
+            last_block_event_count = last_block_event_count,
+            "Replaying Block"
         );
         self.catchup_logs.extend(logs);
         if let Some(event) = self.current_event.take() {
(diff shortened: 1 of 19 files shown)


coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -143,6 +150,7 @@ impl InfiniteLogIter {
     }
 
     async fn get_chain_id_or_panic(&self) -> ChainId {
+        // TODO: remove expect and, instead, propagate the error
```
rudy-6-4, line 153: It's suppose to fail, since it's only called at start, hence the name or_panic

dartdart26: Yes, it will still fail, but it will propagate the error to main (will make main return anyhow::Result<()>).

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -272,6 +280,12 @@ impl InfiniteLogIter {
                 Err(err) => {
                     let delay = if not_initialized {
                         if retry == 0 {
+                            // TODO: remove panic and, instead, propagate the error
```
rudy-6-4, line 283: Same here, at start after 20 retry failed we want a real failure. Error could be bubble up for display but it will end the service.

dartdart26: Yeah, for sure, but better to not do it via a panic and log and exit with an error (not a panic).

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
```diff
@@ -64,24 +70,43 @@ fn install_signal_handlers(cancel_token: CancellationToken) -> anyhow::Result<()
...
+                    .with_max_retries(conf.provider_max_retries)
+                    .with_retry_interval(conf.provider_retry_interval),
+            )
+            .await
+        {
+            Ok(provider) => {
+                info!("Connected to Gateway at {}", conf.gw_url);
+                break provider;
+            }
+            Err(e) => {
```
rudy-6-4, line 101: infinite loop at start ? if the url is bad don't you want the service to stop ?

coprocessor/fhevm-engine/transaction-sender/src/lib.rs
```diff
@@ -89,17 +90,25 @@ where
...
-    let provider = ProviderBuilder::new()
-        .connect_ws(
-            WsConnect::new(ws_url)
-                .with_max_retries(max_retries)
-                .with_retry_interval(retry_interval),
-        )
-        .await?;
+// Gets the chain ID from the given WebSocket URL.
+// This is a utility function that will try to connect until it succeeds.
+pub async fn get_chain_id(ws_url: Url, retry_interval: Duration) -> anyhow::Result<u64> {
```
rudy-6-4: infinite loop if bad url ?

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
dartdart26: How do we know if the URL is bad? If you mean bad format, it won't parse at clap level as the type is `Url`.

coprocessor/fhevm-engine/transaction-sender/src/lib.rs
dartdart26: Same as above - the data type for the field in the clap data structure is `Url`, if it is bad it won't parse at all.

rudy-6-4: how can it fail with infinite retry ? (return a Result<>)

rudy-6-4: ah ok at the last line.

rudy-6-4 (approved):

coprocessor/fhevm-engine/transaction-sender/src/lib.rs
dartdart26: Good point, I will redo that.

dartdart26: No, you are right - provider.get_chain_id() can fail too, put that into the loop too.

dartdart26: As we discussed offline, I am thinking also of the case if the URL is good, but the GW is down - do we want to stop the process?

PR #276, opened 2025-06-13
fix(coprocessor): change common-docker template

fegmorte:


tawadaa (approved):
lgtm, thanks @fegmorte

PR #277, opened 2025-06-13
feat(test-suite): use latest solidity library and host-contracts

jatZama:
This PR updates the test-suite to use latest version of the `FHE.sol` library (0.7.0-5) with removal of ebytesXX types. It also uses latest version of host contracts, with updated FHEVMEExecutor (removal of ebytesXX) as well as corresponding updated HCULimit contract, and updated ACL contract, which adds an env variable `PAUSER_ADRESS` inside the `.env.host` file. Note this last remark is important for ZWS setup.
For the upgrade of testnet, we still need to follow same conventions as agreed with gateway team, as current contracts still use `reinitializer(2)` with a singlie insitilizer, instead of 2 initializer functions which shoudl be protected via `reinitializer(3)` instead, this will be done in a futrue PR, according to this issue https://github.com/zama-ai/fhevm-internal/issues/68.


test-suite/e2e/contracts/TestAsyncDecrypt.sol
```diff
@@ -126,6 +112,15 @@ contract TestAsyncDecrypt is E2EFHEVMConfig {
         return yBool;
     }
 
+    /// @notice Callback function for 4-bit unsigned integer decryption
+    /// @param decryptedInput The decrypted 4-bit unsigned integer
+    /// @return The decrypted value
+    function callbackUint4(uint256 requestID, uint8 decryptedInput, bytes[] memory signatures) public returns (uint8) {
```
PacificYield: Maybe remove this one?

PacificYield (approved):

test-suite/e2e/contracts/TestAsyncDecrypt.sol
jatZama: Will be done later because it is also present in other subdirs I oepend an issue https://github.com/zama-ai/fhevm/issues/280

PR #279, opened 2025-06-13
feat(gateway-contracts): introduce EmptyProxyUpgradeable for deployment behind empty proxies

### LINKED ISSUE #261: chore(gateway-contracts): prevent calls to `reinitializeV{i}` after `initialize` for the same version
### Describe the issue

recently in https://github.com/zama-ai/fhevm/issues/247 we've introduced `initializeFromEmptyProxy` to allow a contract to be init directly from the latest version 

however, nothing prevents someone to call `reinitializeV2` on a contract that have been directly initialized to V2 using `initializeFromEmptyProxy`. Imo both should not be possible as contracts should be init with only one of them

a way to do so might be something like this : 
- do not change anything with reinitializers : 
- update how we use the `reinitializer modifer for `InitializeFromEmptyProxy` by : 
    - make it match the version of the reinitializer from `reinitializeV{i}`, ie both `reinitializer` modifiers will have the same number in it for the same contract version
    - add the following modifier to put **before** the `reinitializer` modifier : 
```
modifier onlyFromEmptyProxy() {
    if(_getInitializedVersion() != 1):
        revert NotInitializingFromEmptyProxy();
    _;
}
```

### Context

Following this comment : https://github.com/zama-ai/fhevm/pull/122/files#r2144606807

_No response_

isaacdecoded:
Closes zama-ai/fhevm-internal#74


gateway-contracts/contracts/shared/EmptyProxyUpgradeable.sol
```diff
@@ -0,0 +1,25 @@
...
+ * @title EmptyProxyUpgradeable
+ * @dev Abstract base contract for upgradeable contracts that are intended to be deployed behind
+ * empty proxies. This contract provides a modifier that ensures functions can only be called
+ * during the first initialization phase (i.e., when initialized version is 1), enforcing
+ * correct deployment from an empty proxy using the UUPSUpgradeable pattern.
+ *
+ * Inheriting contracts should use the `onlyFromEmptyProxy` modifier to protect initialization logic
+ * that must not run on upgrades or reinitializations.
+ */
+abstract contract EmptyProxyUpgradeable is UUPSUpgradeable {
```
melanciani: I think would be best to keep `UUPSUpgradeable` in the name to clarify this is inheriting from the OZ contract, maybe like `UUPSUpgradeableEmptyProxy`

melanciani (approved):
awesome thanks !

PR #281, opened 2025-06-13
refactor(gateway-contracts): rename GatewayConfig initialization event to InitializeGatewayConfig

isaacdecoded:
Closes zama-ai/fhevm-internal#72


isaacdecoded:
@RomanBredehoft do you think we must sync with the Coprocessor Team about this event renaming or better to ask them directly?

melanciani (approved):
great !

melanciani:
@isaacdecoded not sure why coprocessor would need to be updated, are they listening to the init event ?

isaacdecoded:
> are they listening to the init event?

That's the actual question haha I'd say not but wanted to confirm with you 👍

melanciani:
you could simply check in the whole fhevm repo if they listen to it, or directly ask @dartdart26 , but I really don't think they are

PR #283, opened 2025-06-16
fix(kms-connector): change dockerfile for security

fegmorte:


kms-connector/Dockerfile
```diff
@@ -74,4 +69,9 @@ RUN groupadd -g 10002 kms && \
 RUN chown -R kms:kms /app/kms-connector
 USER kms
 
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
+    CMD pgrep -f "dist/src/main" > /dev/null || exit 1
+
+CMD ["/app/kms-connector/bin/kms-connector", "start", "-c", "/var/lib/kms-connector/config.toml"]
```
eudelins-zama: I am not sure to understand where this `/var/lib/kms-connector/config.toml` file comes from...

Anyway, the config file is optional (you can use env variables only)  and not used in staging I believe (do you confirm @tawadaa ?) so I think we could just use `CMD ["/app/kms-connector/bin/kms-connector", "start"]`

fegmorte: @eudelins-zama the config is passed through env var and config file during the deployemnt:
```
      containers:
        - name: kms-connector
          command:
            - "/bin/sh"
            - "-c"
          args:
            - |
              kms-connector validate -c /var/lib/kms-connector/config.toml
              kms-connector start -c /var/lib/kms-connector/config.toml
```
And if you use this image locally you have to mount a volume with this file.
Are you sure you can start without `-c` args ?

eudelins-zama: Hmm, I see.  Then maybe we can delete `COPY ./kms-connector/config/environments/config-base.toml /app/kms-connector/config/environments/config-base.toml` at least?

And yes you can start without the `-c` flag **if required config fields are provided via env vars** which I believe was the cas from my last conversation with @tawadaa who suggested to remove the config file in the future. But I think we can leave the Dockerfile like this for now

kms-connector/Dockerfile
```diff
@@ -74,4 +69,9 @@ RUN groupadd -g 10002 kms && \
 RUN chown -R kms:kms /app/kms-connector
 USER kms
 
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
+    CMD pgrep -f "dist/src/main" > /dev/null || exit 1
```
eudelins-zama: Are you sure the `pgrep -f "dist/src/main"` is a valid healthcheck command?

Sould it be something like `pgrep -f "kms-connector"` as it is the name of the binary?

eudelins-zama: I tested locally and as expected the `pgrep -f "dist/src/main"` command does not find any process, while `pgrep -f "kms-connector"` does!

eudelins-zama (approved):
LGTM!

@tawadaa could you just confirm that the config file is not used in staging nor by zws ?

PR #288, opened 2025-06-16
docs(common): Add links to HH template

0xalexbel:


PacificYield (approved):
Looks good for now. It will be necessary to port the whole guide from the hardhat repo to this folder in the future.

PR #314, opened 2025-06-17
feat(coprocessor): fhevm-listener, health check

rudy-6-4:


coprocessor/fhevm-engine/fhevm-listener/src/health_ckeck.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+        let timestamp_now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
+        let elapsed_time = timestamp_now - self.last_tick;
+        if elapsed_time > 30 {
+            self.message += &format!("Last tick is too old {elapsed_time}\n");
+            self.unhealthy();
+        }
+    }
+
+    pub async fn check_database_connected(&mut self) {
```
rudy-6-4, line 109: could be replace by database tick ?

coprocessor/fhevm-engine/.cargo/deny.toml
```diff
@@ -96,6 +96,7 @@ allow = [
     "BSD-3-Clause-Clear",
     "BSL-1.0",
     "CC0-1.0",
+    "CDLA-Permissive-2.0",
```
dartdart26, line 99: Not qualified on it myself, but just wanted to understand why we allow it?

coprocessor/fhevm-engine/fhevm-listener/Cargo.toml
```diff
@@ -18,14 +18,19 @@ alloy-rpc-types = "1.0.9"
 alloy-sol-types = "1.2.0"
 
 # workspace dependencies
+anyhow = { workspace = true }
 alloy = { workspace = true, features = ["contract", "json", "providers", "provider-ws", "pubsub", "rpc-types", "sol-types"] }
 alloy-primitives = { workspace = true }
+axum = { workspace = true }
 clap = { workspace = true }
 futures-util = { workspace = true }
+http-serde = "2.1.1"
```
dartdart26: This crate doesn't seem super well-maintained, is there a way to not add it as a dependency? AFAICT, we only need it to serialize the status code, can we just convert it to an int?

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -20,13 +20,17 @@ use clap::Parser;
...
 use crate::contracts::{AclContract, TfheContract};
 use crate::database::tfhe_event_propagate::{ChainId, Database};
+use crate::health_ckeck::{HealthCheck, HealthState};
+
 
 #[derive(Parser, Debug, Clone)]
 #[command(version, about, long_about = None)]
 pub struct Args {
-    #[arg(long, default_value = "ws://0.0.0.0:8746")]
+    #[arg(long, default_value = "ws://0.0.0.0:8545")]
```
dartdart26, line 33: As a side note, maybe this and other arguments shouldn't have default values and be mandatory, e.g. also the database_url, etc.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -535,4 +561,5 @@ pub async fn main(args: Args) {
             }
         }
     }
+    health_check_service.abort();
```
dartdart26: Why do we need to abort here, don't we want to use the cancellation token?

dartdart26 (commented):
Looks good overall. I added some comments that, I think, are worth considering.

coprocessor/fhevm-engine/.cargo/deny.toml
rudy-6-4: it's for test with reqwest.
I ll change for tokio since we already have it. I wasn't aware of the new  licence before the CI.

coprocessor/fhevm-engine/fhevm-listener/Cargo.toml
rudy-6-4: removing it and replaced with u16

coprocessor/fhevm-engine/.cargo/deny.toml
rudy-6-4: finally I ll keep it, it's only for tests and the licence looks ok. (used in relayer and kms-core)

coprocessor/fhevm-engine/fhevm-listener/src/health_check.rs
```diff
@@ -0,0 +1,238 @@
...
+            self.database_connected = false;
+            self.unhealthy();
+            return;
+        };
+        self.database_connected = true;
+    }
+
+    pub async fn check_blockchain_connected(&mut self, blockchain_url: &str) {
+        let ws = WsConnect::new(blockchain_url);
+        let provider = ProviderBuilder::new()
```
goshawk-3, line 122: We should find a way to check the actual provider in use instead of creating a new one.

It might happen this returns `blockchain_connected` but the provider in use is faulty.


Also, keep in mind that we need to ensure it's not too expensive to recreate a provider on each request/check.

coprocessor/fhevm-engine/fhevm-listener/src/health_check.rs
```diff
@@ -0,0 +1,238 @@
...
+        let timestamp_now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
+        let elapsed_time = timestamp_now - self.last_tick;
+        if elapsed_time > 30 {
+            self.message += &format!("Last tick is too old {elapsed_time}.\n");
+            self.unhealthy();
+        }
+    }
+
+    pub async fn check_database_connected(&mut self, database_url: &str) {
+        let Ok(options) = database_url.parse::<PgConnectOptions>() else {
```
goshawk-3, line 104: Can you reuse the DB pool that is in use?

goshawk-3 (changes requested):

coprocessor/fhevm-engine/fhevm-listener/src/health_check.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+async fn health_handler(
+    State(state_health): State<Arc<RwLock<HealthStateContent>>>,
+) -> impl IntoResponse {
+    let mut health_state = state_health.read().await.clone();
+    health_state.status.check_last_tick();
+    health_state.status.check_database_connected(&health_state.database_url).await;
+    health_state.status.check_blockchain_connected(&health_state.blockchain_url).await;
+    let health = health_state.status;
```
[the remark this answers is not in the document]
rudy-6-4: We only need to read from the other service, local update is ok

coprocessor/fhevm-engine/fhevm-listener/src/health_check.rs
rudy-6-4: Yes. This check for reachability more than actual connection from the original service.

rudy-6-4: I am not sure about sharing the real component, if health check is requested many time I don't want to saturate the provider or the pool.

coprocessor/fhevm-engine/fhevm-listener/src/health_check.rs
rudy-6-4: Same answer as provider.

coprocessor/fhevm-engine/fhevm-listener/src/health_check.rs
goshawk-3: As per the spec, the health-checking must be lightweight call.

Wrt DoS attack concerns, we can have a rate-limiter inside or outside of the service.

Let's resolve this with the next PR, if you decide to use the common crate.

goshawk-3 (approved):

dartdart26 (approved):

PR #315, opened 2025-06-17
Multi-GPU scheduling

antoniupop:


coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -105,59 +113,16 @@ async fn tfhe_worker_cycle(
         s.end();
         // This query locks our work items so other worker doesn't select them.
         let mut s = tracer.start_with_context("query_work_items", &loop_ctx);
+        #[cfg(feature = "bench")]
+        let now = std::time::SystemTime::now();
         let the_work = query!(
             "
-            WITH RECURSIVE dependent_computations(tenant_id, output_handle, dependencies, fhe_operation, is_scalar, produced_handles) AS (
```
rudy-6-4, line 110: Wow, can we have that for bench only ?
The previous request was doing a lot of checks. If events are in the wrong order, is probably going to break ?

antoniupop: If we trust the blockchain node to not introduce unschedulable computations, the checks are redundant. The worker already checks if computations were able to complete and leave them unchanged in the DB if they are not currently schedulable, so as soon as the inputs are available they can run. This is only a problem if there can be invalid computations inserted.

rudy-6-4: Another possibility is to go the optimistic route and switch to the previous in case of error.

antoniupop: The previous query just filters out any computation that is not computable (doesn't have imputs either directly or from the result of another computation that itself is computable and fetched in the same batch). The new one optimistically assumes that all ocmputations are ultimately computable and just doesn't do anything for those that can't be computed.
The main issue we could get is if the listener inserts lots of computations that can never be computed. If that can't happen we should be fine. 
We'll soon change this again to just select computations needed to compute "allowed" handles - if you think this is not mergeable I can work on that and postpone merging once that's also included as that means we would just not need to worry about any uncomputable bits until they're needed for an allowed handle in which case we have a meaningful error to send out.

rudy-6-4: "The new one optimistically assumes that all ocmputations are ultimately computable and just doesn't do anything for those that can't be computed."
If for some reason events are out of order either by the subscribe or by insertion, it will only delay the compulation ?

rudy-6-4 (approved):

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: Yes, once a batch completes, the computations that couldn't be done are just unlocked and left in the DB, so next batch will get them and if their inputs are available they complete.

PR #322, opened 2025-06-17
feat(test-suite): fhevm-listener, health check

rudy-6-4:
Done separately because the test-suite has fixed versions for components


rudy-6-4:
Waiting test-suite version bump of coprocessor to 0.7.3

PR #328, opened 2025-06-17
ci(common): fix name and make string values uniform

aquint-zama:
- renaming missed one file in rebase
- revert permissions value to quoted string (to make it uniform)


fegmorte (approved):
👍

PR #332, opened 2025-06-18
feat(charts): add probes to coprocessor and bump to 0.4.2

tawadaa:


fegmorte (commented):
I'm not 100% sure we can define a probe ports without defining them at container level.
@piizama or @maxnovawind your advice on that please ?

fegmorte (approved):
LGTM ! 👍

PR #336, opened 2025-06-18
chore(kms-connector): create tx-sender crate structure

eudelins-zama:


kms-connector/crates/tx-sender/src/core/config/raw.rs
dartdart26, line 1: Do we need the separate `RawConfig` data structure? Can we make `Config` itself se/deserializable?

kms-connector/crates/tx-sender/src/core/config/wallet.rs
dartdart26, line 1: Do we need to have the concrete types (i.e. PrivateKeySigner and AwsSigner) instead of an abstract signer? Code might simpler with an abstract one, without explicit matches and also will force us to use the signers in the same way, without specifics. Wdyt?

dartdart26 (approved):
Looks good! I posted some minor comments that we might or might not address.

kms-connector/crates/tx-sender/src/core/config/raw.rs
eudelins-zama: I do not think that `Config` can be deserializable, as AWS sdk use async function to initialize the `KmsWallet`.

Even if it was not the case, if we want `Config` to be deserializable but with custom logic to parse and validate the config, we would need to implement the `Deserialize` trait manually (see serde [docs](https://serde.rs/impl-deserialize.html)) which can lead to a lot of code. Thus I think the easiest way is to have this `RawConfig` already deserializable and then adding our parsing/validation logic on top of it

kms-connector/crates/tx-sender/src/core/config/wallet.rs
eudelins-zama: I do not have a strong opinion on this. I imported the `KmsWallet` from the previous kms-connector codebase.

As we discussed today, my priority is to have e2e tests passing with the new architecture ASAP with as little code modification as possible to validate the new architecture is working. Once that milestone is reached, there would be a lot of room for improvements and new features.

PR #351, opened 2025-06-19
ci(gateway-contracts): introduce CI workflow for upgrade tests

isaacdecoded:
CI workflow success can be checked [here](https://github.com/zama-ai/fhevm/actions/runs/15757653522/job/44416622656?pr=351#step:9:278).

Closes zama-ai/fhevm-internal#87


.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -0,0 +1,132 @@
...
+      - name: Upgrade GatewayConfig contract
+        working-directory: current-fhevm/gateway-contracts
+        env:
+          DOTENV_CONFIG_PATH: .env
+          HARDHAT_NETWORK: staging
+          CHAIN_ID_GATEWAY: 54321
+          RPC_URL: http://localhost:8546
+        run: |
+          cp .env.example .env
+          cp -r ../../previous-fhevm/gateway-contracts/contracts ./previous-contracts
```
melanciani: I'm not sure I get the reason why we need to deploy the contracts in docker and so on if we simply need to copy the old implementation 🤔

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -0,0 +1,132 @@
...
+          npx hardhat task:deployEmptyUUPSProxies
```
melanciani: why is this step necessary ?

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -0,0 +1,132 @@
...
+          npx hardhat task:upgradeGatewayConfig \
```
melanciani, line 125: sorry, I did not realie this in your last PR, but the `reinitialize-args` should only be part of `upgradeCurrentToNew` , the tasks tehmsevles like `task:upgradeGatewayConfig` should provide the necessary upgrade logic from the previous version to the new , and loading the env vars these directly 

it it somewhat what is done in host contracts : https://github.com/zama-ai/fhevm/blob/main/host-contracts/tasks/upgradeProxy.ts#L64 (although here they define a new common tak, and I think it can be avoided with `reinitialize-args`

I really don't think we should make these inputs part of the CLI like here, this will make ZWS work harder

.github/workflows/gateway-contracts-upgrade-tests.yml
melanciani: or more like : I don't get how the upgrade tasks work if ou run it outstide of the docker 🤔

melanciani (commented):
I won't put a request change because I'm off tomorrow and I don't want to block, but I think we can improve how upgrade tasks currently are

.github/workflows/gateway-contracts-upgrade-tests.yml
isaacdecoded: The upgrade task works because the anvil node is running locally from the previous step: `Build and start Docker services from previous release`. Hence, in this step, we execute the upgrade over those already deployed contracts that are supposed to contain the previous/latest release version. Furthermore, with the env variable `RPC_URL` above, it makes the task communicate with the anvil node running locally.

.github/workflows/gateway-contracts-upgrade-tests.yml
isaacdecoded: It's necessary because the upgrade tasks run the `compile:specific` for both the current and the new implementations, and the compilation of our contracts requires the `/addresses` directory.

.github/workflows/gateway-contracts-upgrade-tests.yml
isaacdecoded: Change done in [this PR](https://github.com/zama-ai/fhevm/pull/378).

melanciani: great you can merge it and rebase this PR on it then !

.github/workflows/gateway-contracts-upgrade-tests.yml
melanciani: ok the same issue as https://github.com/zama-ai/fhevm/issues/345 

could you please add a comment (in-line I think) to briefly say we have this here  please ? thanks 1

.github/workflows/gateway-contracts-upgrade-tests.yml
melanciani: ok thanks for the explanation

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -0,0 +1,132 @@
...
+  sc-upgrade:
+    name: gateway-contracts-upgrade-tests/sc-upgrade (bpr)
+    needs: check-changes
+    if: ${{ needs.check-changes.outputs.changes-gw-contracts == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout previous release code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: v0.7.0
```
melanciani: would be great to avoid hardcoding this and instead get the latest release/tag automatically ! 

else we'll for sure forget to update this whenever we create a btw release (already the case btw : we should consider 0.7.2 now)

melanciani (changes requested):
nice ! please incldue the other PR + try to get the latest tag instead !

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -34,11 +34,17 @@ jobs:
     if: ${{ needs.check-changes.outputs.changes-gw-contracts == 'true' }}
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout previous release code
+      - name: Get latest tag
+        id: get-latest-tag
+        run: |
+          git fetch --tags
+          echo "tag=$(git describe --tags `git rev-list --tags --max-count=1`)" >> $GITHUB_OUTPUT
```
melanciani: quick question here : what is the notion of "latest" using this comment ? I mean, is it : 
- time-wise (ie: the latest tag created: not ideal, we sometimes have new tags for rcs that target an older version)
- semver wise (ie: the latest release in terms of versioning: ideal)

also, maybe there is a way directly in github actions to see if we can get the `latest` release (which is what we want to target imo) , since github has this defined somwhere (like in this display : https://github.com/zama-ai/fhevm/releases/tag/v0.7.2)

I can see that there are already some gh actions made for this by people : 
- latest release : https://github.com/marketplace/actions/get-latest-release
- latest tag : https://github.com/marketplace/actions/get-latest-tag

but not sure what they are worth + we should ask in working tools/ @tawadaa  if it's fine to allow these small actions running in the repo

isaacdecoded: The current approach is "time-wise": the latest tag by commit date. Let me take a look at the gh actions you mentioned 👍

isaacdecoded: The step [is now changed](https://github.com/zama-ai/fhevm/pull/351/files#diff-439c64678a6191c111c4a14343dd38cf212877d607101d59609fb8746eefa58eR42-R47) to get the latest semver while using regexp to filter rcs.

I was looking at the gh actions above, but not sure if this is worth adding that dependency for this, that seems pretty simple. Btw, looking at the [get-latest-tag source code](https://github.com/WyriHaximus/github-action-get-previous-tag/blob/master/main.js) it is ordering by `creatordate` which is still "time-wise". For the other one [get-latest-release](https://github.com/pozetroninc/github-action-get-latest-release/blob/master/main.js) it doesn't seem that useful.

Another option could be using the `actions/github-script` and place our own script to get the latest semver tag.

melanciani: > The step [is now changed](https://github.com/zama-ai/fhevm/pull/351/files#diff-439c64678a6191c111c4a14343dd38cf212877d607101d59609fb8746eefa58eR42-R47) to get the latest semver while using regexp to filter rcs.

rcs were an example, we already had an issue in fhevm-gateway because Amina created golden images with tags that were considered latest time-wise but were not semver-wise, hence my point above 😅

melanciani: actually we would not want the latest semver-wise but rather the "previous" semver-wise. for ex if we do a pr for 0.7.3 it should target 0.7.2 although we might already have released 0.8.1 

worth asking in working tools if someone has an idea on how to do this maybe !

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -0,0 +1,176 @@
...
+  sc-upgrade:
+    name: gateway-contracts-upgrade-tests/sc-upgrade (bpr)
+    needs: check-changes
+    if: ${{ needs.check-changes.outputs.changes-gw-contracts == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install semver
+        run: npm install npm/node-semver#281055e7716ef0415a8826972471331989ede58c # v7.7.2
+
+      - name: Get previous release tag
```
isaacdecoded: @RomanBredehoft, I think I'm getting close (with the help of GPT ofc) to what we are seeking here with the "previous semver-wise" logic, but need your review, please. Briefly, it's using the semver package installed above in separate step. Then, it's getting the current target branch of the PR which could be a `release/x.x.x` or `main`:
- If the target branch is a release, it should get the latest tag relative to this release branch, i.e. `release/0.7.x` <- `my-pr-branch` then it should return its relative latest tag, e.g. `v0.7.4`
- If the target branch is main, it should get the absolute latest tag including pre-releases, i.e. `main` <- `my-pr-branch` then it should return the latest tag, e.g. `v0.8.0` or `v0.8.1-rc`, etc.

Wdyt? 🤔

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -0,0 +1,176 @@
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: 'false'
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36  # v3.0.2
+        id: filter
+        with:
+          filters: |
+            gw-contracts:
+              - .github/workflows/gateway-contracts-upgrade-tests.yml
+              - gateway-contracts/contracts/**
```
isaacdecoded: Btw, I changed the filter to consider modifications on the contracts only, since I think if nothing has changed in the contracts then we shouldn't test any upgrade. Wdyt? @RomanBredehoft

melanciani: I don't fully agree, first thing I thought was if we remove/modfy some env vars (only), it could lead to failed upgrades but the CI will be green since we haven't checked this workflow 

down the line, unless a workflow takes too much time, I think it's always safer to include a large filter

.github/workflows/gateway-contracts-upgrade-tests.yml
melanciani: it starts to get too complex for such a small thing, let's not waste too much time here

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -0,0 +1,136 @@
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: 'false'
+          path: current-fhevm
+
+      - name: Install dependencies
+        working-directory: current-fhevm/gateway-contracts
+        run: npm ci
+
+      - name: Upgrade GatewayConfig contract
```
melanciani, line 117: just realized : the decryption contract also needs to be tested as it's been migrated lately

melanciani (changes requested):
great, almost done : we need to test the decryption contract as well

.github/workflows/gateway-contracts-upgrade-tests.yml
isaacdecoded: Ah! We indeed had an upgrade during the weekend. Thanks for the catch!

isaacdecoded:
> great, almost done : we need to test the decryption contract as well

Added the Decryption contract upgrade [here](https://github.com/zama-ai/fhevm/pull/351/commits/b17eeed9f40ecac4e40dc7fd0590dc66e992854f) but also note that it was necessary [to include the Decryption contract address](https://github.com/zama-ai/fhevm/pull/351/commits/ce5935bf1f9c30c62f451ef1c1bf2a322027b1aa) in the `.env.example` file (and the rest of the other contract addresses)

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -0,0 +1,152 @@
...
+      - name: Install dependencies
+        working-directory: current-fhevm/gateway-contracts
+        run: npm ci
+
+      - name: Prepare and copy previous contract into current directory
+        working-directory: current-fhevm/gateway-contracts
+        run: |
+          cp .env.example .env
+          cp -r ../../previous-fhevm/gateway-contracts/contracts ./previous-contracts
+          # Executing task:deployEmptyUUPSProxies is required to generate the proxy contract addresses, which are then used
```
melanciani: nit: you could move the comment right above the step so that we can see it a bit more clear !

gateway-contracts/.env.example
```diff
@@ -125,8 +125,13 @@ HOST_CHAIN_WEBSITE_3="https://host-chain-2028.com" # (string)
...
 
-# GatewayConfig contract address
-# This address is required during the production deployment, at which point this environment variable
-# is dynamically set after contracts are deployed.
-# For tests, this address is read directly from the addresses/.env.gateway_config file, so the value in this file is ignored.
+# Contract addresses
+# These addresses are required during the production deployment, at which point these environment variables
+# are dynamically set after contracts are deployed.
+# For tests, these addresses are read directly from the /addresses directory, so the values in this file are ignored.
+CIPHERTEXT_COMMITS_ADDRESS="0xD582Ec82a1758322907DF80dA8A754e12A5acB95" # (address)
```
melanciani: I don't think you need them, couldn't you use the `useInternalProxyAddress` flag from `task:deployEmptyUUPSProxies` instead ?

melanciani (approved):
thanks a lot, this is a great thing to have !

PR #358, opened 2025-06-19
chore(sdk): change mutability

ppoliani:
Usage of the SDK has two stages:

1. The build (through the builder patter)
2. Usage

The code currently imposes sdk instances to be mutable when encrypting data. More specifically:

```rust
let mut builder = sdk.create_input_builder()?;
builder.add_u8(42)?;
```

Most client would initialize the sdk once and then use that same instance multiple times to call `create_input_builder` to encrypt new data. This function though expect a `&mut self`. This introduces a bottleneck when using the sdk instance in a concurrent scenario. More specifically, it would require wrapping the instance in the slow `Arc<Mutex<FhevmSdk>>`.

We can avoid this by changing `create_input_builder` into a simple `&self` and move the logic that requires mutability into the first stage i.e. the build stage.


leventdem:
It seems good to me, can you rebase it and I will approve.

ppoliani:
> It seems good to me, can you rebase it and I will approve.

Done

leventdem (approved):
Ok 👍

aquint-zama:
@cla-bot check

aquint-zama:
This PR should be redone from a branch now that you're part of the team 🎉

ppoliani:
Closing this in favour of https://github.com/zama-ai/fhevm/pull/499

PR #363, opened 2025-06-19
refactor(host-contracts): add artifacts checks and new param for reinitialize args in upgrade tasks

### LINKED ISSUE #356: refactor(host-contracts): include checks and new arguments for upgrade tasks
### Describe the issue

We want to align the host-contracts' upgrade tasks with the recent changes introduced in the gateway-contracts from PR https://github.com/zama-ai/fhevm/pull/250

cc @jatZama @RomanBredehoft 

### Context

_No response_

_No response_

isaacdecoded:
Closes zama-ai/fhevm-internal#34


host-contracts/tasks/upgradeContracts.ts
```diff
@@ -0,0 +1,474 @@
...
+  .addParam(
+    'currentImplementation',
+    'The currently deployed implementation solidity contract path and name, eg: contracts/ACL.sol:ACL',
+  )
+  .addParam(
+    'newImplementation',
+    'The new implementation solidity contract path and name, eg: examples/ACLUpgradedExample.sol:ACLUpgradedExample',
+  )
+  .addOptionalParam(
+    'reinitializeArgs',
```
melanciani: I missed this in your last PR, but the `reinitializeArgs` should only be part of the common function, not the dedicated tasks imo

see my comment here : https://github.com/zama-ai/fhevm/pull/351#discussion_r2157346776

( I won't do a request changes because I'm off tomorrow and I don't want to block)

host-contracts/tasks/upgradeContracts.ts
```diff
@@ -0,0 +1,474 @@
...
+      proxyAddress,
+      currentImplementation,
+      newImplementation,
+      verifyContract,
+      hre,
+      reinitializeArgs,
+    );
+  });
+
+task('task:test').setAction(async function (_, hre) {
```
jatZama: Sorry but what is this task needed? Did you forget to remove it?

host-contracts/tasks/upgradeContracts.ts
```diff
@@ -0,0 +1,474 @@
...
+      proxyAddress,
+      currentImplementation,
+      newImplementation,
+      verifyContract,
+      hre,
+      reinitializeArgs,
+    );
+  });
+
+task('task:upgradeDecryptionOracleContract')
```
jatZama: Currently the DecryptionOracle does not follow same conventions as host/gateway contracts, ie still a unique reinitialize function, so this won't work fyi.

jatZama: Maybe this task should be removed and upgrade task for oracle handled by ZWS. cc: @fd0r

host-contracts/tasks/upgradeContracts.ts
jatZama: Not sure what you mean here, having an optional argument for args is fine for me. However, I don't like current solution, i think the case when the argument is not given should read from env variables the PAUSER_ADDRESS like what was done before, this seems less error prone. And this is why I had a specific function for ACL.

host-contracts/tasks/upgradeContracts.ts
fd0r: Thanks for the review @jatZama , I'll remove it for now and create an issue to handle it in the future.

host-contracts/tasks/upgradeContracts.ts
isaacdecoded: I see. So you think it's a better approach to implement the tasks in a more specific way for each reinitialize version (e.g. `reinitializeVX()`)? In such a case, maybe this optional `reinitializeArgs` parameter is redundant if we take them from environment variables. @jatZama @RomanBredehoft 

Initially, I thought about the tasks to be more reusable, so we don't need to modify them every time we want to upgrade to a new version but just passing the proper `reinitializeArgs` for the given new version. If the reinitialize args are wrong, then the `upgradeProxy()` transaction should fail on calling the `reinitializeVX()` method.

host-contracts/tasks/upgradeContracts.ts
isaacdecoded: 🤦‍♂️ yes, sorry.

host-contracts/tasks/upgradeContracts.ts
fd0r: Yes definitely, now that we have a docker image in the fhevm-relayer with the Decryption Oracle all mention of it can be removed from here.

isaacdecoded: Upgrade DecryptionOracle task removed in [last commit](https://github.com/zama-ai/fhevm/pull/363/commits/b516bcc161b9d93b1b1c55ca5ecf2593e50ad7ee#diff-2d3b1642261372dd8e4536c2c4014b88917ad2ec8db6705115dc5260f322115eL405) 👍

host-contracts/tasks/upgradeContracts.ts
isaacdecoded: Missed task removed in the [last commit](https://github.com/zama-ai/fhevm/pull/363/commits/b516bcc161b9d93b1b1c55ca5ecf2593e50ad7ee#diff-2d3b1642261372dd8e4536c2c4014b88917ad2ec8db6705115dc5260f322115eL464).

host-contracts/tasks/upgradeContracts.ts
isaacdecoded: `PAUSER_ADDRESS` rolled back to be gotten from env variables and removed the optional `reinitializeArgs` parameter in [the last commit](https://github.com/zama-ai/fhevm/pull/363/commits/b516bcc161b9d93b1b1c55ca5ecf2593e50ad7ee#diff-2d3b1642261372dd8e4536c2c4014b88917ad2ec8db6705115dc5260f322115eR147).

jatZama (approved):

PR #375, opened 2025-06-20
coprocessor: Enable basic health-checking logic for sns-worker 

goshawk-3:
- Expose s3 and database health checks.
- Ensure the liveness status (alive/not_responding) are never false positive.
- Simplify the main loop in the executor.

### DIFF
--- a/coprocessor/fhevm-engine/sns-executor/src/executor.rs
+++ b/coprocessor/fhevm-engine/sns-executor/src/executor.rs
@@ -1,122 +1,210 @@
+use crate::aws_upload::check_is_ready;
 use crate::keyset::fetch_keyset;
 use crate::squash_noise::SquashNoiseCiphertext;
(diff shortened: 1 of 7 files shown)


coprocessor/fhevm-engine/sns-executor/src/executor.rs
```diff
@@ -158,7 +251,10 @@ async fn fetch_and_execute_sns_tasks(
 
     let trx = &mut db_txn;
 
+    let mut maybe_remaining = false;
```
rudy-6-4, line 254: could be returned in each branch instead of mut. either ia a return or a let bla = if

rudy-6-4 (approved):

tawadaa (approved):
LGTM, thanks!

dartdart26 (approved):
Looks good!

PR #378, opened 2025-06-20
refactor(gateway-contracts): get reinitialize args from env variables

isaacdecoded:
This PR refactors the upgrade tasks to get the reinitialize arguments from env variables instead of receiving them from an optional parameter.

More info in discussion from [PR #351](https://github.com/zama-ai/fhevm/pull/351#discussion_r2157346776).


melanciani (approved):
great thanks !

PR #383, opened 2025-06-24
feat(coprocessor): gas limit oveprovision in txn-sender

dartdart26:
Add a config option to overprovision gas limit in transaction sender (in percentage).

Reason is that sometimes we need more gas on rollups such as Arbitrum, due to the way gas is calculater there. Intention is to reduce the likelihood of transaction failure due to insufficient gas.

Resolves https://github.com/zama-ai/fhevm-internal/issues/56

### DIFF
--- a/coprocessor/fhevm-engine/transaction-sender/src/overprovision_gas_limit.rs
+++ b/coprocessor/fhevm-engine/transaction-sender/src/overprovision_gas_limit.rs
@@ -1,0 +1,42 @@
+use alloy::network::{Ethereum, TransactionBuilder};
+use alloy_provider::Provider;
+use alloy_rpc_types::TransactionRequest;
+use tracing::{debug, warn};
+
+// If `txn_request.gas` is set, overprovision it by the given percent.
+// If `txn_request.gas` is not set, estimate the gas limit and then overprovision it by the given percent.
+// If the percent is less than 100, code will assert.
+// If the gas estimation fails, it will not set the gas limit for overprovisioning and will log a warning.
+pub async fn try_overprovision_gas_limit<T: Provider<Ethereum>>(
+    txn_request: impl Into<TransactionRequest>,
+    provider: &T,
+    percent: u32,
+) -> TransactionRequest {
+    assert!(percent >= 100, "Overprovision percent must be at least 100");
+
+    let overprovision = |gas: u64| (gas as u128 * percent as u128 / 100) as u64;
+
+    let mut txn: TransactionRequest = txn_request.into();
+
+    let new_gas = match txn.gas {
+        Some(existing_gas) => Some(existing_gas),
+        None => match provider.estimate_gas(txn.clone()).await {
+            Ok(estimated_gas) => Some(estimated_gas),
+            Err(err) => {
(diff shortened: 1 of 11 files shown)


coprocessor/fhevm-engine/transaction-sender/src/overprovision_gas_limit.rs
```diff
@@ -0,0 +1,52 @@
...
+// If `txn_request.gas` is set, overprovision it by the given percent.
+// If `txn_request.gas` is not set, estimate the gas limit and then overprovision it by the given percent.
+// If the percent is less than 100, it will be set to 100%.
+// If the gas estimation fails, it will not set the gas limit for overprovisioning and will log a warning.
+pub async fn try_overprovision_gas_limit<T: Provider<Ethereum>>(
+    txn_request: impl Into<TransactionRequest>,
+    provider: &T,
+    percent: u32,
+) -> TransactionRequest {
+    if percent < 100 {
```
rudy-6-4: The warn could be done at object creation not at runtime.

dartdart26: In practice, I've made it such that clap parses and checks it is >= 100, maybe I can make it an assert then and remove the test and the warn?

coprocessor/fhevm-engine/transaction-sender/src/overprovision_gas_limit.rs
```diff
@@ -0,0 +1,52 @@
...
+    let percent = percent.max(100);
+
+    let overprovision = |gas: u64| (gas as u128 * percent as u128 / 100) as u64;
+
+    let mut txn: TransactionRequest = txn_request.into();
+
+    let new_gas = async {
+        txn.gas
+            .or({
+                match provider.estimate_gas(txn.clone()).await {
```
rudy-6-4: or is eager !? (not sure with all these async block :) )
Maybe you want or_else ?

coprocessor/fhevm-engine/transaction-sender/tests/add_ciphertext_tests.rs
```diff
@@ -349,7 +349,12 @@ async fn retry_on_transport_error(#[case] signer_type: SignerType) -> anyhow::Re
         ProviderBuilder::default()
             .filler(FillersWithoutNonceManagement::default())
             .wallet(env.wallet.clone())
-            .connect_ws(WsConnect::new(env.ws_endpoint_url()))
+            .connect_ws(
```
rudy-6-4, line 352: Is it really  part of this PR ?

rudy-6-4 (approved):

coprocessor/fhevm-engine/transaction-sender/tests/add_ciphertext_tests.rs
dartdart26: No, it isn't, just noticed it taking more time than I would expect. I can put it in a separate PR?

coprocessor/fhevm-engine/transaction-sender/src/overprovision_gas_limit.rs
dartdart26: Wow, good catch! Let me redo that!

PR #394, opened 2025-06-26
fix(coprocessor): cargo fmt

rudy-6-4:


.github/workflows/coprocessor-cargo-fmt.yml
dartdart26, line 1: We discussed offline whether to have 2 workflows, 2 jobs or 2 run commands in the same jobs for clippy and fmt. We discussed visibility, performance (what is in parallel, etc.) Will let you decide, I am fine either way.

dartdart26 (approved):
Nice!

rudy-6-4:
https://github.com/zama-ai/fhevm-internal/issues/30

.github/workflows/coprocessor-cargo-fmt.yml
rudy-6-4: Chosed 2 workflow before reworking how it's displayed in github UI.

PR #406, opened 2025-06-28
Add Chainid host in user decrypt request 

immortal-tofu:
Fix https://github.com/zama-ai/fhevm-internal/issues/23


fd0r (approved):

gateway-contracts/contracts/Decryption.sol
```diff
@@ -6,6 +6,7 @@ import { ciphertextCommitsAddress } from "../addresses/CiphertextCommitsAddress.
 import { gatewayConfigAddress } from "../addresses/GatewayConfigAddress.sol";
 import { ECDSA } from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import { EIP712Upgradeable } from "@openzeppelin/contracts-upgradeable/utils/cryptography/EIP712Upgradeable.sol";
+import { MessageHashUtils } from "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
```
jatZama, line 9: I don't like this low level approach, which is very error-prone, I would rather use the EIP712UpgradeableCrossChain contract on gateway which is currently used on host, and use there regular EIP712Upgradeable. I can't approve this PR as it is now.

immortal-tofu: Ok good to know! I'll do the change :)

EDIT: Sadly EIP712UpgradeableCrossChain require to have a fixed chainId, but in our case, it can change depending of the network. We can't use it.

immortal-tofu: Sadly EIP712UpgradeableCrossChain require to have a fixed chainId, but in our case, it can change depending of the network. We can't use it.

immortal-tofu: This "low-level" approach still use OZ functions. Except if you know a OZ native way to do it, I don't see how we can do it differently :(

jatZama: ok then forget my comment, we can still use low level code exceptionally here, and refactor later.

gateway-contracts/contracts/Decryption.sol
```diff
@@ -640,19 +641,30 @@ contract Decryption is
...
-                        EIP712_USER_DECRYPT_REQUEST_TYPE_HASH,
-                        keccak256(userDecryptRequestVerification.publicKey),
-                        keccak256(abi.encodePacked(userDecryptRequestVerification.contractAddresses)),
-                        userDecryptRequestVerification.contractsChainId,
-                        userDecryptRequestVerification.startTimestamp,
-                        userDecryptRequestVerification.durationDays
-                    )
-                )
-            );
+        // Could be a constant
```
jatZama: Can you remove this comment and declare `bytes32 private constant TYPE_HASH = keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");` at the start of the contract instead please?

immortal-tofu: Would it work in an upgradable contract?

gateway-contracts/scripts/bindings_update.py
```diff
@@ -16,7 +16,7 @@
 GW_CONTRACTS_DIR = GW_ROOT_DIR.joinpath("contracts")
 GW_MOCKS_DIR = GW_CONTRACTS_DIR.joinpath("mocks")
 
-ALLOWED_FORGE_VERSIONS = ["1.2.2-v1.2.2", "1.2.2-stable"]
+ALLOWED_FORGE_VERSIONS = ["1.2.2-v1.2.2", "1.2.2-stable", "1.2.3-stable"]
```
jatZama, line 19: not sure about this change?

immortal-tofu: New version has been published and I use it. That's all :)

melanciani: yes but we should keep only a single version between all users, so either 2.2 or 2.3 but not both !! : https://github.com/zama-ai/fhevm/issues/423

gateway-contracts/contracts/Decryption.sol
melanciani: I created https://github.com/zama-ai/fhevm/issues/424 just in case we can improve this in the future

PR #427, opened 2025-06-30
fix(gateway-contracts): use contractsChainId in delegate request eip712 verification

melanciani:
idk why some readmes got updated by my `conformance` locally 

note that I'll tackle https://github.com/zama-ai/fhevm-internal/issues/16 just after

also note that the relayer sdk does not need to be updated as delegate tests have already been update alon the user decrypt changes : https://github.com/zama-ai/relayer-sdk/commit/e73e60fe24127162a764a7e093d96c312183a29f

closes https://github.com/zama-ai/fhevm-internal/issues/19


isaacdecoded (approved):
LGTM!

PR #428, opened 2025-06-30
fix(coprocessor): use safe_deserialize_conformant in zkproof worker

dartdart26:
Instead of using just `safe_deserialize`, use the `safe_deserialize_conformant` function to ensure that the deserialization is conformant with the expected parameters.


goshawk-3 (approved):
LGTM

PR #439, opened 2025-06-30
chore(gateway-contracts): add reinitializer version constant in contracts

melanciani:
since we've done several upgrades lately, better be as future proof as possible

closes https://github.com/zama-ai/fhevm-internal/issues/32


isaacdecoded (approved):
LGTM! 👍

PR #442, opened 2025-06-30
fix(gateway-contracts): make the verify subtask call more specific on the verifying contract

### LINKED ISSUE #415: feat(gateway-contracts): fix verification step on blockscoute in upgrade script
### Describe the issue

We noticed this last step is currently failing with current upgrade script with @piizama today. 
I will let @isaacdecoded investigate by deploying and upgrading contracts on blockscout. 

### Context

_No response_

_No response_

isaacdecoded:
The hardhat verify plugin was failing because of bytecode conflicts for the `Decryption.sol` contract. In order to prevent this ambiguity, there's an attribute `contract` that allows to be more specific about the bytecode implementation to be verified. The specific error was:

```
Error in plugin @nomicfoundation/hardhat-verify: More than one contract was found to match the deployed bytecode.
Please use the contract parameter with one of the following contracts:
  * contracts/Decryption.sol:Decryption
  * contracts/examples/DecryptionUpgradedExample.sol:DecryptionUpgradedExample

For example:

hardhat verify --contract contracts/Example.sol:ExampleContract <other args>

If you are running the verify subtask from within Hardhat instead:

await run("verify:verify", {
<other args>,
contract: "contracts/Example.sol:ExampleContract"
};

For more info run Hardhat with --show-stack-traces
```
Closes zama-ai/fhevm-internal#20


gateway-contracts/tasks/upgradeContracts.ts
```diff
@@ -63,6 +63,7 @@ async function upgradeCurrentToNew(
     const implementationAddress = await hre.upgrades.erc1967.getImplementationAddress(proxyAddress);
     await hre.run("verify:verify", {
       address: implementationAddress,
+      contract: newImplementation,
```
isaacdecoded, line 66: Do you want me to add this in the [host-contract](https://github.com/zama-ai/fhevm/blob/main/host-contracts/tasks/upgradeContracts.ts#L64-L67) as well? 🤔

melanciani: I think we should yes cc @jatZama

melanciani (approved):
nice ! could you explain a bit why this closes the issue ?

isaacdecoded:
> nice ! could you explain a bit why this closes the issue ?

Sure, I forgot to add more info indeed, sorry. Posted in the PR description up there 👍

gateway-contracts/tasks/upgradeContracts.ts
jatZama: I dont understand the error, it seems the bytecodes are not the same when looking at the code.

jatZama: The error is very weird, because bytecodes should be different contrarily to what is logged in error message. Maybe an `npx hardhat clean` could have fixed the issue without needing to change the command? Did you try it @isaacdecoded ?

jatZama: Otherwise, yes we need to change it also on host-contracts, and let ZWS team know that command for upgrading contracts have changed.

jatZama: AH no this does not change the actual command, because this task is internally called, sorry i was confused. So No need to tell ZWS there is a change to take into account on their side! So I am ok to update host-contracts anyways. I am just not understanding the error message though. My suspicion is there is maybe some clashing in compiled artifacts, somehow, but it is not clear at all...

isaacdecoded: @jatZama yes, I tried cleaning the artifacts and still have the error. I don't understand why the conflict is happening on the bytecodes but Pierre [confirmed it's the same error he had](https://zama-ai.slack.com/archives/C0915Q59CKG/p1751300535496819?thread_ts=1751299008.763139&cid=C0915Q59CKG).

PR #459, opened 2025-07-01
fix(kms-connector): fix s3 hash check + add retry

### LINKED ISSUE #206: fix(kms-connector): retry S3 ciphertext retrieval
### Describe the issue

When too many requests are sent to the S3 buckets to retrieve ciphertexts, some of these requests can fail even if the URL is correct and the ciphertext is stored in the S3 bucket.

So the KMS Connector should retry a few time to retrieve a ciphertext from S3 before considering it is not in the S3 bucket.

### Context

_No response_

_No response_

eudelins-zama:
Closes zama-ai/fhevm-internal#93 

Also fix the ciphertext hash check we discussed this morning

### DIFF
--- a/kms-connector/crates/kms-worker/src/core/config/raw.rs
+++ b/kms-connector/crates/kms-worker/src/core/config/raw.rs
@@ -33,10 +33,14 @@ pub struct RawConfig {
     pub public_decryption_timeout_secs: u64,
     #[serde(default = "default_user_decryption_timeout")]
     pub user_decryption_timeout_secs: u64,
-    #[serde(default = "default_retry_interval")]
-    pub retry_interval_secs: u64,
+    #[serde(default = "default_grpc_retry_interval")]
+    pub grpc_retry_interval_secs: u64,
     #[serde(default)]
     pub s3_config: Option<S3Config>,
+    #[serde(default = "default_s3_ciphertext_retrieval_retries")]
+    pub s3_ciphertext_retrieval_retries: u8,
+    #[serde(default = "default_s3_connect_timeout")]
+    pub s3_connect_timeout: u64,
     #[serde(default = "default_verify_coprocessors")]
     pub verify_coprocessors: bool,
 }
@@ -57,8 +61,16 @@ fn default_user_decryption_timeout() -> u64 {
     300 // 5 minutes
 }
 
-fn default_retry_interval() -> u64 {
+fn default_grpc_retry_interval() -> u64 {
     5 // 5 seconds
+}
+
+fn default_s3_ciphertext_retrieval_retries() -> u8 {
+    3 // 3 seconds
+}
+
+fn default_s3_connect_timeout() -> u64 {
+    2 // 2 seconds
 }
 
 fn default_verify_coprocessors() -> bool {
@@ -89,7 +101,9 @@ impl Default for RawConfig {
             service_name: "kms-connector".to_string(),
             public_decryption_timeout_secs: 300,
             user_decryption_timeout_secs: 300,
-            retry_interval_secs: 5,
+            grpc_retry_interval_secs: 5,
+            s3_ciphertext_retrieval_retries: 3,
+            s3_connect_timeout: 2,
             s3_config: None,
             verify_coprocessors: false,
         }
--- a/kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
@@ -1,5 +1,5 @@
 use crate::core::config::{Config, S3Config};
-use alloy::{hex::encode, primitives::Address, providers::Provider, transports::http::reqwest};
+use alloy::{hex, primitives::Address, providers::Provider, transports::http::reqwest};
 use anyhow::anyhow;
 use dashmap::DashMap;
 use fhevm_gateway_rust_bindings::{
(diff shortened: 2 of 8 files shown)


kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
```diff
@@ -211,70 +219,137 @@ where
...
+            match self
+                .retrieve_s3_ciphertext_with_retry(s3_urls, ct_digest, &ct_digest_hex)
                 .await
+            {
+                Some(ciphertext) => {
+                    sns_ciphertext_materials.push((sns_material.ctHandle.to_vec(), ciphertext));
+                }
+                None => error!(
+                    "Failed to retrieve ciphertext for digest {ct_digest_hex} from any S3 URL"
+                ),
```
maksymsur, line 250: Is this a recoverable error, or will it cause the app to crash? If Yes - are you sure it's a desired behavior that cannot be exploited?

kms-connector/crates/kms-worker/src/core/config/raw.rs
```diff
@@ -57,10 +61,18 @@ fn default_user_decryption_timeout() -> u64 {
     300 // 5 minutes
 }
 
-fn default_retry_interval() -> u64 {
+fn default_grpc_retry_interval() -> u64 {
     5 // 5 seconds
 }
 
+fn default_s3_ciphertext_retrieval_retries() -> u8 {
+    3 // 5 seconds
```
maksymsur: Inconsistency with comments

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
```diff
@@ -211,70 +219,137 @@ where
...
+        if calculated_digest != ciphertext_digest {
+            let calculated_digest_hex = hex::encode(&calculated_digest);
+            return Err(anyhow!(
+                "DIGEST MISMATCH: Expected: {digest_hex}, Got: {calculated_digest_hex}",
+            ));
         }
+        info!("DIRECT HTTP RETRIEVAL COMPLETE: Successfully verified ciphertext digest");
 
-        sns_ciphertext_materials
+        // Return data even with digest mismatch for non-failability
```
maksymsur: Shall this be removed?

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
```diff
@@ -211,70 +219,137 @@ where
...
+            ciphertext.len()
+        );
+
+        // Verify digest
+        let calculated_digest = compute_digest(&ciphertext);
+        if calculated_digest != ciphertext_digest {
+            let calculated_digest_hex = hex::encode(&calculated_digest);
+            return Err(anyhow!(
+                "DIGEST MISMATCH: Expected: {digest_hex}, Got: {calculated_digest_hex}",
+            ));
```
maksymsur, line 313: Is this a recoverable error, or will it cause the app to crash? If Yes - are you sure it's a desired behavior that cannot be exploited?

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
```diff
@@ -211,70 +219,137 @@ where
...
+                            attempt = i,
+                            "Successfully retrieved ciphertext for digest {digest_hex} from S3 URL {s3_url}"
                         );
-                        // Continue to the next URL
+                        return Some(ciphertext);
                     }
+                    Err(e) => warn!(
+                        attempt = i,
+                        "Failed to retrieve ciphertext for digest {digest_hex} from S3 URL {s3_url}: {e}"
+                    ),
```
maksymsur, line 280: Is this a recoverable error, or will it cause the app to crash? If Yes - are you sure it's a desired behavior that cannot be exploited?

maksymsur (changes requested):
Looks good overall, but I’m concerned about error handling here—could this lead to an app crash or expose a vulnerability? Please reassess whether this failure mode is appropriate for the Connector, given its importance to system reliability.

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
eudelins-zama: `error!` is just a tracing macro, very different from `panic!`, so I do not see why a crash would happened here.

If this error log happens, it means that none of the S3 buckets contain the ciphertext we are looking for, and therefore we'll ignore this one and continue fetching the other sns_ciphertext_material from s3

kms-connector/crates/kms-worker/src/core/config/raw.rs
eudelins-zama: Good catch!

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
eudelins-zama: Yes!

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
eudelins-zama: This error is recovered in `retrieve_s3_ciphertext_with_retry`, so no crash. I do not see an attack vector from this as the only consequence will be warning/error log(s)

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
eudelins-zama: Same as above, only logging

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
```diff
@@ -211,70 +219,136 @@ where
...
+        digest_hex: &str,
+    ) -> Option<Vec<u8>> {
+        for i in 1..=self.s3_ciphertext_retrieval_retries {
+            for s3_url in s3_urls.iter() {
+                match self
+                    .retrieve_s3_ciphertext(s3_url, ciphertext_digest, digest_hex)
+                    .await
                 {
                     Ok(ciphertext) => {
                         info!(
```
maksymsur, line 271: Let’s consider whether it’s worth retrying a potentially dead URL, or if we’d be better off moving on to other URLs that might be healthy. May be spend minimal time and retries on each as your defaults are a bit time consuming?

maksymsur (approved):
LGTM! Thanks for clarifications!

PR #464, opened 2025-07-01
feat(host-contracts): make the verify subtask call more specific on the upgrade task

isaacdecoded:
Refs zama-ai/fhevm-internal#20 
Refs https://github.com/zama-ai/fhevm/pull/442


jatZama (approved):
LGTM

PR #469, opened 2025-07-01
chore(host-contracts): add reinitializer version constant in contracts

melanciani:
similarly to https://github.com/zama-ai/fhevm/pull/439, let's do the same for host contracts


jatZama (approved):
good

QdanF:
nice ser

PR #499, opened 2025-07-02
chore(sdk): change mutability

ppoliani:
Usage of the SDK has two stages:

The build (through the builder pattern)
Usage
The code currently imposes sdk instances to be mutable when encrypting data. More specifically:

```rust
let mut builder = sdk.create_input_builder()?;
builder.add_u8(42)?;
```

Most clients would initialize the sdk once and then use that same instance multiple times to call `create_input_builder` to encrypt new data. This function though expects a `&mut self.` This introduces a bottleneck when using the sdk instance in a concurrent scenario. More specifically, it would require wrapping the instance in the slow `Arc<Mutex<FhevmSdk>>`.

We can avoid this by changing `create_input_builder` into a simple &self and move the logic that requires mutability into the first stage i.e. the build stage.

### DIFF
--- a/sdk/rust-sdk/README.md
+++ b/sdk/rust-sdk/README.md
@@ -34,7 +34,7 @@ use std::path::PathBuf;
 
 fn main() -> Result<()> {
     // Initialize SDK
-    let mut sdk = FhevmSdkBuilder::new()
+    let sdk = FhevmSdkBuilder::new()
         .with_keys_directory(PathBuf::from("./keys"))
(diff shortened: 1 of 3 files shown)


leventdem (approved):
Good catch!

aquint-zama:
@cla-bot check

aquint-zama:
@cla-bot check

aquint-zama:
@cla-bot check

PR #500, opened 2025-07-02
fix(sdk): use the correct values for eip712 domain

ppoliani:
In the JS SDK the EIP712 has slightly different values. More specifically as shown [here](https://github.com/zama-ai/relayer-sdk/blob/v0.1.0/src/index.ts#L178)  the two main differences are:

1. `chain_id` is `contracts_chain_id` and not the `gateway_id`
2. `verifying_contract` is the `verifyingContractAddressDecryption` rather than `inputVerifierContractAddress`

After making those changes the signature generated is the one that `v1/user-decrypt` verifies correctly.

### DIFF
--- a/sdk/rust-sdk/src/lib.rs
+++ b/sdk/rust-sdk/src/lib.rs
@@ -237,7 +237,7 @@ impl FhevmSdk {
         let verifying_contract = self
             .config
             .gateway_contracts
-            .input_verification
+            .decryption
             .unwrap_or_else(|| {
                 warn!("Input verification contract not set, using zero address");
                 Address::ZERO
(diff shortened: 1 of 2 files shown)


sdk/rust-sdk/src/lib.rs
```diff
@@ -237,7 +237,7 @@ impl FhevmSdk {
         let verifying_contract = self
             .config
             .gateway_contracts
-            .input_verification
+            .decryption
             .unwrap_or_else(|| {
                 warn!("Input verification contract not set, using zero address");
```
leventdem: Can we update the message also ?

aquint-zama:
@cla-bot check

leventdem (approved):
+1

aquint-zama:
@cla-bot check

PR #504, opened 2025-07-03
Preserve  the creation timestamp for computations

antoniupop:


coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -111,7 +111,7 @@ async fn tfhe_worker_cycle(
             FROM computations
             WHERE is_completed = false
             AND is_error = false
-            ORDER BY created_at
+            ORDER BY schedule_order
```
goshawk-3, line 114: Should we set the `schedule_order` in the host-listener when inserting into computations?

coprocessor/fhevm-engine/db-migration/migrations/20250703000000_add_schedule_order_column.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE computations
+ADD COLUMN IF NOT EXISTS schedule_order TIMESTAMP NOT NULL DEFAULT NOW();
```
goshawk-3, line 2: Should we  set any value to `schedule_order` for the existing computations?

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: It's got a default, so should be set.

coprocessor/fhevm-engine/db-migration/migrations/20250703000000_add_schedule_order_column.sql
antoniupop: I think it will be set automatically during migration, no?

goshawk-3 (approved):
LGTM

PR #511, opened 2025-07-03
feat(gateway-contracts): implement dedicated contract verification tasks

isaacdecoded:
Closes [#35](https://github.com/zama-ai/fhevm-internal/issues/35)

### DIFF
--- a/gateway-contracts/tasks/blockExplorerVerify.ts
+++ b/gateway-contracts/tasks/blockExplorerVerify.ts
@@ -1,0 +1,171 @@
+import dotenv from "dotenv";
+import fs from "fs";
+import { task, types } from "hardhat/config";
+
+import { getRequiredEnvVar } from "./utils/loadVariables";
+
+task("task:verifyCiphertextCommits")
+  .addOptionalParam(
+    "useInternalProxyAddress",
+    "If proxy address from the /addresses directory should be used",
+    false,
+    types.boolean,
+  )
+  .setAction(async function ({ useInternalProxyAddress }, { upgrades, run }) {
+    let proxyAddress: string;
+    if (useInternalProxyAddress) {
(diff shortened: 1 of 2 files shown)


gateway-contracts/tasks/blockExplorerVerify.ts
isaacdecoded, line 1: @jatZama do you think it's ok to refactor the [etherscanVerify.ts](https://github.com/zama-ai/fhevm/blob/main/host-contracts/tasks/etherscanVerify.ts) from host-contracts to make it similar to this? i.e:
- Rename to `blockExplorerVerify.ts`
- Add the `useInternalProxyAddress` parameter
- Remove the current verification of `DecryptionOracle`

jatZama (approved):

gateway-contracts/tasks/blockExplorerVerify.ts
jatZama: Ok for the first two points, but removing verification of DecryptionOracle is too early right now, because no one is handling this on ZWS team for the moment as far as i know. Anyways, it should first be added in fhevm-relayer, and deployment process updated (at least on e2e test suite and staging) to take into account this, wdyt?

isaacdecoded: Great, I'll create the issue to tackle this then, thanks!

melanciani: yes thanks @isaacdecoded , don't forget to create the issue (and link it here for convenience) !

melanciani (approved):
thanks !

isaacdecoded:
Follow-up issue for aligning verification tasks in `host-contracts` here https://github.com/zama-ai/fhevm-internal/issues/196

PR #519, opened 2025-07-04
fix: wrong return with `error!` macro

mdqst:
turns out I was doing `return error!(...)`, which doesn’t make sense - the macro just returns `()`, and the function expected something else.

replaced it with a proper `error!` call + return.


harakakiri:
thank you

mdqst:
[harakakiri](https://github.com/harakakiri), please merge it.

eudelins-zama:
Closing as it concerns old version of connector and is not a relevant fix

PR #523, opened 2025-07-04
docs(common): add devnet EOL notice

yuxizama:


docs/protocol/README.md
```diff
@@ -17,12 +17,20 @@ layout:
 **Welcome to the Zama Confidential Blockchain Protocol Docs.**\
 The docs aim to guide you to build confidential dApps on top of any L1 or L2 using Fully Homomorphic Encryption (FHE).
 
-{% hint style="warning" %}
+{% hint style="info" %}
 
-The Zama Protocol Testnet is not audited and is not intended for production use. **Do not publish any critical or sensitive data**. For production workloads, please wait for the Mainnet release.
```
bcm-at-zama, line 22: it's too bad to keep on top of the page? it's really important so I would keep on top

bcm-at-zama (approved):
Looks good to me but maybe we also need to do it in the release branch, for it to appear in the https://docs.zama.ai/fhevm/0.6? That GB is, to me, the most important part where to have the warning

docs/protocol/README.md
yuxizama: It's better to bring it back after the EOL. 
If we put both on top, it takes up a very large space on the landing page -> IMO it doesn't leave users a good impression to arrive on a page with many warnings and notices.  wdyt?

immortal-tofu (approved):

NijasR:
bot

PR #532, opened 2025-07-06
chore(sdk): upgrade to kms v0.11.0-rc17

ppoliani:
- upgraded to kms v0.11.0-rc17 which is compatible with the latest testnet
- fixed issues caused by this upgrade

### DIFF
--- a/sdk/rust-sdk/Cargo.lock
+++ b/sdk/rust-sdk/Cargo.lock
@@ -134,9 +134,9 @@ dependencies = [
 
 [[package]]
 name = "alloy"
-version = "1.0.8"
+version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ppoliani:
@leventdem [this](https://github.com/zama-ai/fhevm/blob/b7b2485161993cb7cd16d824d97d6d901879da3a/sdk/rust-sdk/src/decryption/user/response.rs#L447) test currently fails presumably because the public key with the latest changes is way longer than the one in this test.

The issue is that we're using the `js_api` from `kms-core` which targets wasm. Currently when a `JsError` is created we get a panic. 

> running 1 test
panicked at /wasm-bindgen-0.2.100/src/lib.rs:1062:1:
function not implemented on non-wasm32 targets
panicked at library/core/src/panicking.rs:218:5:


Should we try avoid using the `wasm` interface within the `gateway-sdk`?

ppoliani:
Closing in favour of https://github.com/zama-ai/fhevm/pull/590

PR #534, opened 2025-07-07
chore(test-suite): update keygen command to new KMS API

dd23:
Bring test-suite up to date with KMS `main`. 
This is WIP until we have a new release that we can test.


bossr1:
cool

dd23:
Has made its way to main via #593 already, so we can close this one

PR #537, opened 2025-07-07
chore(test-suite): bump images and add missing workflow inputs

melanciani:
- added missing coprocessor components's versions as inputs to workflow dispatch
- bumped fhevm default image versions to 0.7.5
- bumped the relayer version to rc14, and included necessary changes


tawadaa (approved):
lgtm, thanks!

PR #538, opened 2025-07-07
feat(charts): add hpa to coprocessor workers

tawadaa:


charts/charts/coprocessor/templates/coprocessor-zkproof-worker-hpa.yaml
```diff
@@ -0,0 +1,42 @@
...
+  maxReplicas: {{ .Values.zkProofWorker.hpa.maxReplicas }}
+  metrics:
+  - type: Resource
+    resource:
+      name: cpu
+      target:
+        type: Utilization
+        averageUtilization: {{ .Values.zkProofWorker.hpa.targetCPUUtilizationPercentage }}
+  behavior:
+    scaleUp:
```
maxnovawind: We can simplify the behavior section by using toYaml.

charts/charts/coprocessor/templates/coprocessor-tfhe-worker-hpa.yaml
```diff
@@ -0,0 +1,42 @@
...
+  maxReplicas: {{ .Values.tfheWorker.hpa.maxReplicas }}
+  metrics:
+  - type: Resource
+    resource:
+      name: cpu
+      target:
+        type: Utilization
+        averageUtilization: {{ .Values.tfheWorker.hpa.targetCPUUtilizationPercentage }}
+  behavior:
+    scaleUp:
```
maxnovawind: Same thing here.

piizama:
Would it be possible to move the `charts/tests` files to a `tests/` folder within each charts folder (eg. `charts/coprocessor/tests`). This can then be added to .helmignore to prevent being packaged with the chart. This would avoid, having to create a nested `charts/charts` folder and have to move everything.

.github/config/ct.yaml
```diff
@@ -3,7 +3,7 @@
 remote: origin
 target-branch: main
 chart-dirs:
-  - charts
+  - charts/charts
```
maxnovawind: Why do we need to do this?

charts/charts/coprocessor/templates/coprocessor-sns-worker-hpa.yaml
```diff
@@ -0,0 +1,42 @@
...
+  maxReplicas: {{ .Values.snsWorker.hpa.maxReplicas }}
+  metrics:
+  - type: Resource
+    resource:
+      name: cpu
+      target:
+        type: Utilization
+        averageUtilization: {{ .Values.snsWorker.hpa.targetCPUUtilizationPercentage }}
+  behavior:
+    scaleUp:
```
maxnovawind: same thing here

maxnovawind (commented):
LGTM for the HPA! 👍 I just added a few comments regarding some renaming suggestions. Also, having charts/charts feels a bit odd, might be worth reconsidering the folder structure.

.github/config/ct.yaml
NekiMax: interesting

tawadaa: The reason was that I wanted to include helmfile deployments for local testing in `charts` folder, but that's no longer needed, and also not the recommended way to do so

charts/coprocessor/values.yaml
```diff
@@ -260,6 +260,29 @@ tfheWorker:
...
+          periodSeconds: 15
+        - type: Pods
+          value: 2
+          periodSeconds: 60
+      scaleDown:
+        stabilizationWindowSeconds: 300
+        selectPolicy: Min
+        policies:
+        - type: Percent
+          value: 50
```
antoniupop: I think that this may need taking down to 40% otherwise we would downscale and hit 100% immediately again if we halve resources.

antoniupop (approved):
Looks good - just one comment, I'l leave it up to you if you want to fix or test first and see if it's an issue!

charts/coprocessor/values.yaml
tawadaa: These default values aren't important, they are never used because `enabled: false` it's just here for the example, but I'll make it 40% for the exmaple

PR #539, opened 2025-07-07
feat(gateway-contracts): implement multi-sig contracts based on Safe Smart Account

isaacdecoded:
The following PR introduces two Smart Accounts (ERC-4337) for the currently Pauser and the Owner roles, based on the [Safe Smart Account](https://docs.safe.global/advanced/smart-account-overview).

This PR also centralizes the Gateway ownership on the `GatewayConfig` contract instead of having separated owners per contracts, i.e. all contracts must check ownership by calling the `GatewayConfig` contract.

Closes [fhevm-internal/issues/8](https://github.com/zama-ai/fhevm-internal/issues/8)


gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -32,7 +36,7 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
     /// @notice The contract's variable storage struct (@dev see ERC-7201)
     /// @custom:storage-location erc7201:fhevm_gateway.storage.GatewayConfig
     struct GatewayConfigStorage {
-        /// @notice The pauser's address
+        /// @notice The pauser's address - deprecated, use `CIRCUIT_BREAKER_SAFE_ACCOUNT_ADDRESS` instead.
         address pauser;
```
melanciani: we started to put deprecated on the same line as the field itself, like : 
```
address pauser;  // deprecated, use `CIRCUIT_BREAKER_SAFE_ACCOUNT_ADDRESS` instead.
```

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -101,7 +104,6 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
     /// @param initialCustodians List of custodians
     /// @custom:oz-upgrades-validate-as-initializer
     function initializeFromEmptyProxy(
-        address initialPauser,
```
melanciani: don't forge to bump the contract's version and reinitlaizer btw

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -251,8 +245,7 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
 
     /// @dev See {IGatewayConfig-checkIsPauser}.
     function checkIsPauser(address pauserAddress) external view virtual {
```
melanciani: I think we don't need `checkIsPauser` and the `GatewayConfigChecks`'s `onlyPauser` modifier anymore btw

gateway-contracts/contracts/safeModules/CircuitBreakerModule.sol
```diff
@@ -0,0 +1,65 @@
...
+            abi.encode(PAUSE_CONTRACT_TYPE_HASH, contractAddress, msg.sender, nonces[msg.sender]++, deadline)
+        );
+
+        bytes32 hash = keccak256(abi.encodePacked("\x19\x01", getDomainSeparator(), signatureData));
+
+        Safe(payable(safeAddress)).checkSignatures(hash, abi.encodePacked(signatureData), signature);
+
+        bytes memory data = abi.encodeWithSignature("pause()");
+
+        require(
```
melanciani: I think it's best practice to instead use if + revert statement instead of require (like mentioned [here](https://ethereum.stackexchange.com/a/119381))

gateway-contracts/tasks/deployment/contracts.ts
```diff
@@ -165,11 +160,62 @@ task("task:deployDecryption").setAction(async function (_, hre) {
   await deployContractImplementation("Decryption", hre);
 });
 
+task("task:deploySafeSmartAccounts").setAction(async function (_, { ethers, run }) {
```
melanciani: the task should use the same name as the contract !

gateway-contracts/tasks/deployment/contracts.ts
```diff
@@ -165,11 +160,62 @@ task("task:deployDecryption").setAction(async function (_, hre) {
...
+  const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
+  const deployer = new Wallet(deployerPrivateKey).connect(ethers.provider);
+
+  const safeFactory = await ethers.getContractFactory("Safe", deployer);
+  const masterCopy = await safeFactory.deploy();
+
+  console.log("Deploy Safe Smart Account contracts:");
+
+  // Deploy a new SafeProxyFactory contract
+  const safeProxyFactoryFactory = await ethers.getContractFactory("SafeProxyFactory", deployer);
```
melanciani: I think this proxy should be deployed separately and its address will be constant, so I think we should use a similar pattern than empty proxies (ie, get the address from the env addresses ect)

jatZama: EDIT: My first comment was wrong, I thought you deployed an UUPS Proxy for upgradeability, but this is obviously not the case here, you are just using the proxy from Safe default implem, because they do this to save gas at deployment, this is not an upgradeable proxy but a clone proxy. So all good.

gateway-contracts/contracts/GatewayConfig.sol
jatZama: why do you need a new state variable? i would not deprecate pauser, and just reuse it for the safe multisig. this is what i had in mind in the beginning. wdyt?

gateway-contracts/contracts/safeModules/CircuitBreakerModule.sol
jatZama: Actually it is best practise to use custom errors instead of string. You already use custom errors in all other contracts, so why use a string here?
If+revert pattern was used everywhere with custom errors actually because solidity 0.8.24 did not support require with custom errors, but with newer versions of the compiler it is now possible to use require with custom errors, which is more readable but we could change this later in another PR (also on host contracts maybe). Let's use current pattern everywhere for the moment: if+revert with custom errors.

gateway-contracts/hardhat.config.ts
```diff
@@ -156,6 +157,9 @@ const config: HardhatUserConfig = {
       viaIR: false,
     },
   },
+  dependencyCompiler: {
```
jatZama, line 165: why do you need this?

gateway-contracts/hardhat.config.ts
```diff
@@ -4,6 +4,7 @@ import "@nomicfoundation/hardhat-verify";
 import "@openzeppelin/hardhat-upgrades";
 import "@typechain/hardhat";
 import dotenv from "dotenv";
+import "hardhat-dependency-compiler";
```
jatZama, line 7: why do you need this? if possible let's avoid adding new dependencies unless absolutely necessary.

gateway-contracts/package.json
```diff
@@ -48,5 +50,10 @@
   "dependencies": {
     "@openzeppelin/contracts": "^5.2.0",
     "@openzeppelin/contracts-upgradeable": "^5.2.0"
+  },
+  "overrides": {
+    "@safe-global/safe-contracts": {
+      "ethers": "^6.13.5"
```
jatZama, line 58: why is this needed?

gateway-contracts/package.json
```diff
@@ -34,6 +35,7 @@
     "dotenv": "^16.4.7",
     "ethers": "^6.4.0",
     "hardhat": "^2.22.18",
+    "hardhat-dependency-compiler": "^1.2.1",
```
jatZama, line 39: why is this needed?

gateway-contracts/contracts/safeModules/CircuitBreakerModule.sol
```diff
@@ -0,0 +1,65 @@
...
+    constructor(address _safeAddress) {
+        safeAddress = _safeAddress;
+    }
+
+    /**
+     * @dev Generates the EIP-712 domain separator for the contract.
+     *
+     * @return The EIP-712 domain separator.
+     */
+    function getDomainSeparator() private view returns (bytes32) {
```
jatZama: Please dont use this low level approach here, it is very error prone. We used it as a quick fix recently for Decryption because we had no other clear easy choice to use a non standard EIP712 with a different chainID than gateway chain there, but here we are using the regular EIP712 related to gateway, so you should NOT use this low level approach. Please use the `_hashTypedDataV4` function from EIP712.sol instead.

gateway-contracts/hardhat.config.ts
isaacdecoded: This is necessary to have the SafeProxyFactory and Safe artifacts available during the deployment tasks.

gateway-contracts/hardhat.config.ts
isaacdecoded: See comment below. But yes, agree with avoiding the addition of superfluous dependencies - it's not the case, this dependency allows hardhat to have the `dependencyCompiler` configs.

gateway-contracts/package.json
isaacdecoded: Reason here: https://github.com/zama-ai/fhevm/pull/539/files#r2195357487

gateway-contracts/package.json
isaacdecoded: This is required to solve a dependency conflict between `@safe-global/safe-contracts` and `@nomicfoundation/hardhat-ethers`. More precisely:

```
npm error code ERESOLVE
npm error ERESOLVE could not resolve
npm error
npm error While resolving: @safe-global/safe-contracts@1.4.1-2
npm error Found: ethers@6.14.3
npm error node_modules/ethers
npm error   dev ethers@"^6.4.0" from the root project
npm error   peer ethers@"^6.14.0" from @nomicfoundation/hardhat-chai-matchers@2.0.9
npm error   node_modules/@nomicfoundation/hardhat-chai-matchers
npm error     dev @nomicfoundation/hardhat-chai-matchers@"^2.0.8" from the root project
npm error   4 more (@nomicfoundation/hardhat-ethers, ...)
npm error
npm error Could not resolve dependency:
npm error peer ethers@"5.4.0" from @safe-global/safe-contracts@1.4.1-2
npm error node_modules/@safe-global/safe-contracts
npm error   dev @safe-global/safe-contracts@"^1.4.1-2" from the root project
npm error
npm error Conflicting peer dependency: ethers@5.4.0
npm error node_modules/ethers
npm error   peer ethers@"5.4.0" from @safe-global/safe-contracts@1.4.1-2
npm error   node_modules/@safe-global/safe-contracts
npm error     dev @safe-global/safe-contracts@"^1.4.1-2" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
```

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Change reverted in [last commit](https://github.com/zama-ai/fhevm/pull/539/commits/13397f53f37f0f4ce39b70c2b49a672dfe4d0885) in favor of keeping the initializable/configurable pauser address.

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Bump not required as we are keeping the current initialization pauser input.

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Method is not present anymore.

gateway-contracts/contracts/safeModules/CircuitBreakerModule.sol
isaacdecoded: Yes, those `requires` were illustrative only during the discussion about the approach of the early draft PR. Thanks anyway for the careful review. 👍 This `CircuitBreakerModule.sol` was removed in [the last commit](https://github.com/zama-ai/fhevm/pull/539/commits/13397f53f37f0f4ce39b70c2b49a672dfe4d0885).

gateway-contracts/Makefile
```diff
@@ -3,6 +3,9 @@ include .env.example
 LOCAL_NETWORK_NAME=localGateway
 ENV_PATH=.env.example
 FORGE_DAPP_OUT=artifacts
+# This is needed because the Safe library implements CREATE2 opcode which calculates addresses based on bytecode hash.
+# The bytecode is altered by the Solidity coverage tool, making the addresses different on such environment.
```
melanciani: I really don't get why the coverage tool would alter the contract's bytecode 🤔  maybe @jatZama knows ?

down the line, I really think it would be better to find a way to avoid having hardcoded values in the makefile and only use the address from `.env.example`, there must be a way

gateway-contracts/tasks/deployment/contracts.ts
```diff
@@ -62,6 +63,75 @@ async function deployContractImplementation(
...
+  const safeFactory = await ethers.getContractFactory("Safe", deployer);
+  const safe = await safeFactory.deploy();
+  const safeAddress = await safe.getAddress();
+
+  // Deploy a new SafeProxyFactory contract
+  const safeProxyFactoryFactory = await ethers.getContractFactory("SafeProxyFactory", deployer);
+  const safeProxyFactory = await safeProxyFactoryFactory.deploy();
+
+  // Prepare the setup transaction data
+  if (owners.length === 0) {
```
melanciani: we should avoid having default logic like this in order to prevent unwanted behaviors in case of misuse

so : 
- if the list it empty, the task should fail (you can throw an error here if you want but the contract should also revert)
- if someone wants to use the deployer as the owner, than whoever runs the task should provide the deployer in the list
- if having the deployer as the single owner really makes the task more convenient to use (for ZWS, for test purposes, ..) , then better to add an optional boolean in the tasks with default to false that will do this

gateway-contracts/tasks/deployment/contracts.ts
```diff
@@ -165,6 +235,161 @@ task("task:deployDecryption").setAction(async function (_, hre) {
...
+  .setAction(async function ({ owners, threshold }, hre) {
+    // Compile contracts from external dependencies (e.g., Safe Smart Account).
+    // These are temporarily stored by `hardhat-dependency-compiler`.
+    // See the `dependencyCompiler` field in `hardhat.config.ts` for configuration details.
+    await hre.run("compile:specific", { contract: "hardhat-dependency-compiler" });
+
+    await deployMultisigSmartAccount("PauserSmartAccount", hre, owners, threshold);
+  });
+
+task("task:transferOwnershipsToOwnerSmartAccount")
```
melanciani: are you sure we need to provide such a task ? I initially thought it would be done manually, but if ZWS asked for it could you share the discussion about it ?

besides: 
- since transfer ownership is a 2-step process, I believe this should at least be 2 tasks (one to transfer, 1 to accept) + see below
- this task is not very general, it looks like it just transfers from the deployer to the owner safe contract: the 2 tasks' name should highlight this

but before making changes make sure these are really necessary to provide

melanciani (commented):
I have several observations and questions, but won't request a change and let @jatZama be the one to approve these changes

gateway-contracts/Makefile
isaacdecoded: AFAIU, the coverage library injects some tracking events into the contract in order to check the coverage on calling its methods. These injected events plus the usage of CREATE2 opcode, are the ones that makes the contract bytecode change, affecting the contract address on deployment.

gateway-contracts/tasks/deployment/contracts.ts
isaacdecoded: Yes, this is not required by anyone. Initially, I implemented it to test the transfer ownership and just leave it there but let's remove it then. 👍

gateway-contracts/tasks/deployment/contracts.ts
isaacdecoded: Yes, I agree, thanks! 👍

gateway-contracts/Makefile
jatZama: Isaac is right, it adds events.

gateway-contracts/tasks/deployment/contracts.ts
jatZama: I think this task will be needed by ZWS. I would not remove it personally, just add another task for approcing ownership instead maybe?

gateway-contracts/tasks/deployment/contracts.ts
```diff
@@ -165,6 +235,161 @@ task("task:deployDecryption").setAction(async function (_, hre) {
...
+    // are sorted by owner address. This is required to easily validate no confirmation duplicates exist.
+    const sortedSigners = signers.sort((a, b) => {
+      const addressA = addresses[signers.indexOf(a)];
+      const addressB = addresses[signers.indexOf(b)];
+      return addressA.localeCompare(addressB, "en", { sensitivity: "base" });
+    });
+
+    console.log(`Transferring ownerships to OwnerSmartAccount at address: ${ownerSmartAccountAddress}`);
+
+    for (const contractName of contracts) {
```
jatZama: I dont really like the fact that you have to update the owner for each contract in a separate tx, i think there should be always a unique owner for all the protocol contracts don't you agree?

We could solve this by adding a level of indirection: just add a GatewayOwner contract which will be set in bytecode as a constant in all gateway contracts, and this GatewayOwner will have the multisig as a owner, and just redirect calls from the multisig to any of the protocol contracts needed, wdyt about this approach?

gateway-contracts/Makefile
melanciani: this is annoying 🤔  but is there really no other way to handle this except by hardcoding this address here ? maybe in the way we run coverage / read the addresses ? 

worst worst case I think would be better to put this address in the env.example (under a `# Test` section with some comments about it), so that all env vars are there

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -301,6 +301,18 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
 
+    /// @dev See {IGatewayConfig-checkIsGatewayConfigOwner}.
+    function checkIsGatewayConfigOwner(address sender) external view virtual {
+        if (sender != owner()) {
+            revert NotGatewayConfigOwner(sender);
+        }
+    }
+
+    /// @dev See {IGatewayConfig-getOwner}.
+    function getOwner() external view virtual returns (address) {
```
melanciani: `owner()` is already a public function from the abstract, I don't think we need this getter (and its associated test)

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -301,6 +301,18 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         }
     }
 
+    /// @dev See {IGatewayConfig-checkIsGatewayConfigOwner}.
+    function checkIsGatewayConfigOwner(address sender) external view virtual {
```
melanciani: since `owner` is already a public method, wdyt of moving this check + error directly in the GatewayConfigChecks abstract contract ? I don't think we'll need it elsewhere

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Yes, I this method should become unneeded and disappear once [this PR](https://github.com/zama-ai/fhevm/pull/621) got merged

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Yes, that could also work. But in that case, we need to add the `owner() public view` method in the `IGatewayConfig` contract for the `GatewayConfigChecks` to be able to call it.

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -301,6 +301,18 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         }
```
melanciani: I believe we decided to add the owner address as an input field in the initializer right ?

and maybe consider a different account than deployer for it as it'll represent the expected use (ie, the owner will ultimately not be the deployer in test/mainnet)

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+  .setAction(async function ({ owners, threshold }, hre) {
+    // Compile contracts from external dependencies (e.g., Safe Smart Account).
+    // These are temporarily stored by `hardhat-dependency-compiler`.
+    // See the `dependencyCompiler` field in `hardhat.config.ts` for configuration details.
+    await hre.run("compile:specific", { contract: "hardhat-dependency-compiler" });
+
+    await deployMultisigSmartAccount("PauserSmartAccount", hre, owners, threshold);
+  });
+
+task("task:transferOwnershipToOwnerSmartAccount")
```
isaacdecoded: Thinking in this task again, it seems that it should be separated in two tasks:
1. task:transferGatewayConfigOwnership
2. task:acceptGatewayConfigOwnership

Wdyt? @RomanBredehoft @jatZama

gateway-contracts/test/CiphertextCommits.ts
```diff
@@ -330,8 +331,14 @@ describe("CiphertextCommits", function () {
       // Check that the contract is not paused
       expect(await ciphertextCommits.paused()).to.be.false;
 
-      // Pause the contract with the pauser address
-      await expect(ciphertextCommits.connect(pauser).pause()).to.emit(ciphertextCommits, "Paused").withArgs(pauser);
+      // Get the target contract address and the data to call the pause function.
+      const to = await ciphertextCommits.getAddress();
+      const data = ciphertextCommits.interface.encodeFunctionData("pause");
```
melanciani: btw I'm not sure we need to test the multi-sig logic in unit tests, maybe only e2e tests, as the safe contract is not a logic coming from us. wdyt @leventdem @jatZama ?

gateway-contracts/.env.example
```diff
@@ -18,7 +18,7 @@ PROTOCOL_NAME="Protocol" # (string)
 PROTOCOL_WEBSITE="https://protocol.com" # (string)
 
 # Pauser
-PAUSER_ADDRESS="0xa44366bAA26296c1409AD1e284264212029F02f1" # accounts[2] (address)
+PAUSER_ADDRESS="0x7F4518163836161e5316ddddE3f9e5bE112c749c" # safe smart account (address)
```
melanciani: I guess you;ll also need a owner safe smart account in relation to https://github.com/zama-ai/fhevm/pull/539/files#r2236744041

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
+import { OperationType } from "@safe-global/types-kit";
```
melanciani: I think this file should be named `safe_smart_accounts.ts`

gateway-contracts/contracts/shared/GatewayConfigChecks.sol
```diff
@@ -36,4 +36,10 @@ abstract contract GatewayConfigChecks {
         _GATEWAY_CONFIG.checkHostChainIsRegistered(HandleOps.extractChainId(handle));
         _;
     }
+
+    /// @dev Check that the sender is the owner of the GatewayConfig contract.
+    modifier onlyGatewayConfigOwner() {
```
melanciani: maybe `onlyGatewayOwner` is enough ? the modifier already comes from `GatewayConfigChecks` and the owner is essentially the owner for the whole gateway

gateway-contracts/hardhat.config.ts
```diff
@@ -52,10 +55,24 @@ task("compile:specific", "Compiles only the specified contract")
...
 task("test", "Runs the test suite, optionally skipping setup tasks")
   .addOptionalParam("skipSetup", "Set to true to skip setup tasks", false, types.boolean)
   .setAction(async ({ skipSetup }, hre, runSuper) => {
     if (!skipSetup) {
+      // Deploy Smart Accounts with the deployer account as the owner
+      const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
+      const deployer = new hre.ethers.Wallet(deployerPrivateKey).connect(hre.ethers.provider);
+
+      await hre.run("task:deployOwnerSmartAccount", { owners: [deployer.address], threshold: 1 });
+      await hre.run("task:deployPauserSmartAccount", { owners: [deployer.address], threshold: 1 });
```
melanciani: depending what we decide in https://github.com/zama-ai/fhevm/pull/539/files#r2236746157, either : 
- remove the use of safe smart account in tests
- or keep them, but in that case the owners should be the pauser address (already exists as env var) and owner address (see https://github.com/zama-ai/fhevm/pull/539/files#r2236757398)

gateway-contracts/hardhat.config.ts
```diff
@@ -52,10 +55,24 @@ task("compile:specific", "Compiles only the specified contract")
     await hre.run("compile");
   });
 
+task("coverage", "Runs the Solidity Coverage suite").setAction(async (_, hre, runSuper) => {
```
melanciani: great solution ! maybe add a comment saying why we need this !

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: > I believe we decided to add the owner address as an input field in the initializer right ?

So we have another `owner` state variable different from the inherited one from `Ownable2StepUpgradeable`? I'm asking mostly because this `owner` from `Ownable2StepUpgradeable` is initialized in the `EmptyUUPSProxy` during deployment, and to transfer it we need to follow the 2-step flow `transfer` -> `accept`.

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+  const safeFactory = await ethers.getContractFactory("Safe", deployer);
+  const safe = await safeFactory.deploy();
+  const safeAddress = await safe.getAddress();
+
+  // Deploy a new SafeProxyFactory contract
+  const safeProxyFactoryFactory = await ethers.getContractFactory("SafeProxyFactory", deployer);
+  const safeProxyFactory = await safeProxyFactoryFactory.deploy();
+
+  // Prepare the setup transaction data
+  const to = ethers.ZeroAddress; // Contract address for optional delegate call.
```
melanciani: maybe dum question but are these inputs useful for ZWS ? maybe @jatZama or @piizama knows

because if so, they might need to be inputs to the `deployMultisigSmartAccount` function as well as evn vars for the tasks

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+  if (!event) {
+    throw new Error("ProxyCreation event not found in transaction receipt");
+  }
+  const safeProxyAddress = event.args.proxy;
+
+  if (safeProxyAddress === ethers.ZeroAddress) {
+    throw new Error("Safe proxy address not found");
+  }
+
+  const nameSnakeCase = pascalCaseToSnakeCase(name);
```
melanciani: is this part still relevant ? since we now have a single env file

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+  const envContent = `${nameSnakeCase.toUpperCase()}_ADDRESS=${safeProxyAddress}\n`;
+
+  // Ensure the ADDRESSES_DIR exists or create it
+  fs.mkdirSync(ADDRESSES_DIR, { recursive: true });
+
+  // Write the contract's address in the envFilePath file
+  fs.writeFileSync(envFilePath, envContent, { encoding: "utf8", flag: "w" });
+}
+
+task("task:deployOwnerSmartAccount")
```
melanciani: maybe `deployOwnerSafeSmartAccount` instead

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+  .setAction(async function ({ owners, threshold }, hre) {
+    // Compile contracts from external dependencies (e.g., Safe Smart Account).
+    // These are temporarily stored by `hardhat-dependency-compiler`.
+    // See the `dependencyCompiler` field in `hardhat.config.ts` for configuration details.
+    await hre.run("compile:specific", { contract: "hardhat-dependency-compiler" });
+
+    await deployMultisigSmartAccount("OwnerSmartAccount", hre, owners, threshold);
+  });
+
+task("task:deployPauserSmartAccount")
```
melanciani: same, maybe `deployPauserSafeSmartAccount`

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+import fs from "fs";
+import { task, types } from "hardhat/config";
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+import path from "path";
+
+import { ADDRESSES_DIR } from "../../hardhat.config";
+import { getRequiredEnvVar } from "../utils/loadVariables";
+import { pascalCaseToSnakeCase } from "../utils/stringOps";
+
+async function deployMultisigSmartAccount(
```
melanciani: maybe `deploySafeSmartAccount` instead to stay coherent

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+  .addParam("signerPrivateKey", "Private key of one of the owners of the OwnerSmartAccount", undefined, types.string)
+  .addOptionalParam(
+    "useInternalProxyAddress",
+    "If proxy address from the /addresses directory should be used",
+    false,
+    types.boolean,
+  )
+  .setAction(async function ({ currentOwnerPrivateKey, signerPrivateKey, useInternalProxyAddress }, { ethers }) {
+    // Get a currentOwner wallet.
+    // const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
```
melanciani: to remove

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+        throw new Error(`Environment file not found: ${ownerSmartAccountEnvFilePath}`);
+      }
+      dotenv.config({ path: [gatewayEnvFilePath, ownerSmartAccountEnvFilePath], override: true });
+    }
+
+    // Get the GatewayConfig contract from the Ownable2StepUpgradeable factory
+    const gatewayConfigSnakeCase = pascalCaseToSnakeCase("GatewayConfig");
+    const gatewayConfigAddressEnvVarName = `${gatewayConfigSnakeCase.toUpperCase()}_ADDRESS`;
+    const gatewayConfigContractAddress = getRequiredEnvVar(gatewayConfigAddressEnvVarName);
+    const gatewayConfigContract = await ethers.getContractAt("Ownable2StepUpgradeable", gatewayConfigContractAddress);
```
melanciani: why is the contract name `Ownable2StepUpgradeable` ?

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+  .setAction(async function ({ currentOwnerPrivateKey, signerPrivateKey, useInternalProxyAddress }, { ethers }) {
+    // Get a currentOwner wallet.
+    // const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
+    const currentOwner = new Wallet(currentOwnerPrivateKey).connect(ethers.provider);
+    const signer = new Wallet(signerPrivateKey).connect(ethers.provider);
+
+    const ownerSmartAccountSnakeCase = pascalCaseToSnakeCase("OwnerSmartAccount");
+    if (useInternalProxyAddress) {
+      const gatewayEnvFilePath = path.join(ADDRESSES_DIR, `.env.gateway`);
+      const ownerSmartAccountEnvFilePath = path.join(ADDRESSES_DIR, `.env.${ownerSmartAccountSnakeCase}`);
```
melanciani: so we need another env file ?

gateway-contracts/tasks/deployment/smart_accounts.ts
isaacdecoded: This is relevant because the owner and pauser smart account addresses are written in different files. Initially, I thought that way to not mix Gateway contracts addresses with these Safe Smart Accounts ones, but maybe we'd like to keep all in the single `.env.gateway` and `GatewayAddresses.sol` files? 🤔

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+  )
+  .setAction(async function ({ currentOwnerPrivateKey, signerPrivateKey, useInternalProxyAddress }, { ethers }) {
+    // Get a currentOwner wallet.
+    // const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
+    const currentOwner = new Wallet(currentOwnerPrivateKey).connect(ethers.provider);
+    const signer = new Wallet(signerPrivateKey).connect(ethers.provider);
+
+    const ownerSmartAccountSnakeCase = pascalCaseToSnakeCase("OwnerSmartAccount");
+    if (useInternalProxyAddress) {
+      const gatewayEnvFilePath = path.join(ADDRESSES_DIR, `.env.gateway`);
```
melanciani: we should have a utility function for this + the error : https://github.com/zama-ai/fhevm-internal/issues/274

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+        throw new Error(`Environment file not found: ${gatewayEnvFilePath}`);
+      }
+      if (!fs.existsSync(ownerSmartAccountEnvFilePath)) {
+        throw new Error(`Environment file not found: ${ownerSmartAccountEnvFilePath}`);
+      }
+      dotenv.config({ path: [gatewayEnvFilePath, ownerSmartAccountEnvFilePath], override: true });
+    }
+
+    // Get the GatewayConfig contract from the Ownable2StepUpgradeable factory
+    const gatewayConfigSnakeCase = pascalCaseToSnakeCase("GatewayConfig");
```
melanciani: could be a utility function : https://github.com/zama-ai/fhevm-internal/issues/275

gateway-contracts/tasks/deployment/smart_accounts.ts
```diff
@@ -0,0 +1,201 @@
...
+      baseGas,
+      gasPrice,
+      gasToken,
+      refundReceiver,
+      nonce,
+    );
+    const bytesDataHash = ethers.getBytes(transactionHash);
+
+    // Sign the transaction hash with the signer account.
+    const signedMessage = await signer.signMessage(bytesDataHash);
```
melanciani: I'm surprised a single signer is needed here 🤔  should we ask a list of signers ? I imagine it needs at least the threshold number of signers to exec the tx

gateway-contracts/tasks/deployment/smart_accounts.ts
isaacdecoded: Because is the `Ownable2StepUpgradeable` artifact that defines the methods we need for transfer and accept ownership. But maybe this is kind of confusing now that you're asking, so better to get the `GatewayConfig` contract instead.

gateway-contracts/tasks/deployment/smart_accounts.ts
melanciani: shouldn't we do the same for the pauser ? less urgent so maybe this could be done later, and in that case let's create an issue for it

melanciani (changes requested):
great work, I however have several questions !

gateway-contracts/contracts/GatewayConfig.sol
melanciani: I see, then let me continue with this PR sothat i merge it fast and you can rebase on it

gateway-contracts/contracts/GatewayConfig.sol
melanciani: I think it's better if you store the contract in GatewayConfigChecks instead of the interface then !

gateway-contracts/contracts/GatewayConfig.sol
melanciani: iiuc that's the plan yes

gateway-contracts/tasks/deployment/smart_accounts.ts
melanciani: yes I think it makes sense, I suggested this as well last week : https://github.com/zama-ai/fhevm/pull/539#discussion_r2207054980

gateway-contracts/tasks/deployment/smart_accounts.ts
melanciani: yes that would be better I think

gateway-contracts/contracts/GatewayConfig.sol
melanciani: sorry  I think this won't work because it adds a circular import, I think you instead should do something like 
```
Ownable2StepUpgradeable(gatewayConfigAddress).owner()
```

gateway-contracts/hardhat.config.ts
isaacdecoded: The final decision was to remove all Safe multisignature logic from the tests, thanks.

gateway-contracts/Makefile
isaacdecoded: This is no longer an issue since [we decided to remove the Safe multisignature logic](https://github.com/zama-ai/fhevm/pull/539#discussion_r2238791133) from the tests.

gateway-contracts/tasks/deployment/smart_accounts.ts
isaacdecoded: Here's the follow-up issue https://github.com/zama-ai/fhevm-internal/issues/278

gateway-contracts/tasks/deployment/smart_accounts.ts
melanciani: ah I see, maybe a separated `env.safe_smart_accounts` makes sense yes

I don't think we need a `.sol` file for them though

gateway-contracts/tasks/deployment/smart_accounts.ts
isaacdecoded: Yes, multiple signers added [in this commit](https://github.com/zama-ai/fhevm/pull/539/commits/443d933e71f7167232a80bae33bf7c84f7d9f67e#diff-23d5b6bfddd17b25ce0ed45abdc0f875f273416c3deb85226da46bbd78788e13R165).

TL;DR: I don't remember why I thought the Safe contract aggregated the signatures. Many signatures as threshold should be sent at once.

gateway-contracts/tasks/deployment/smart_accounts.ts
isaacdecoded: In the end, I think we don't need to save the safe smart account address since we don't need them for tests anymore and they could be received as input in the tasks like [here](https://github.com/zama-ai/fhevm/pull/539/commits/443d933e71f7167232a80bae33bf7c84f7d9f67e#diff-23d5b6bfddd17b25ce0ed45abdc0f875f273416c3deb85226da46bbd78788e13R130) 🤔 wdyt? @RomanBredehoft

melanciani: yes indeed I think you're right

gateway-contracts/contracts/shared/GatewayConfigChecks.sol
```diff
@@ -36,4 +43,12 @@ abstract contract GatewayConfigChecks {
         _GATEWAY_CONFIG.checkHostChainIsRegistered(HandleOps.extractChainId(handle));
         _;
     }
+
+    /// @dev Check that the sender is the owner of the GatewayConfig contract.
+    modifier onlyGatewayOwner() {
+        if (msg.sender != Ownable2StepUpgradeable(gatewayConfigAddress).owner()) {
```
melanciani, line 54: maybe add a small comment saying why we need to cast to `Ownable2StepUpgradeable` here !

gateway-contracts/contracts/shared/Pausable.sol
```diff
@@ -64,7 +64,7 @@ abstract contract Pausable is Ownable2StepUpgradeable, PausableUpgradeable {
      * - The contract must be paused.
      */
     function unpause() external virtual {
-        if (msg.sender != owner() && msg.sender != gatewayConfigAddress) {
+        if (msg.sender != Ownable2StepUpgradeable(gatewayConfigAddress).owner() && msg.sender != gatewayConfigAddress) {
```
melanciani, line 72: same !

gateway-contracts/hardhat.config.ts
melanciani: maybe add a small comment saying this then !

gateway-contracts/test/Decryption.ts
```diff
@@ -745,6 +745,7 @@ describe("Decryption", function () {
       ciphertextCommits = fixtureData.ciphertextCommits;
       decryption = fixtureData.decryption;
       owner = fixtureData.owner;
+      pauser = fixtureData.pauser;
```
melanciani, line 841: I don't get how the test was working all along without this 🤔

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
```diff
@@ -0,0 +1,250 @@
...
+  { ethers }: HardhatRuntimeEnvironment,
+  owners: string[],
+  threshold: number,
+) {
+  // Get a deployer wallet
+  const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
+  const deployer = new Wallet(deployerPrivateKey).connect(ethers.provider);
+
+  // Deploy a new Safe contract
+  const safeFactory = await ethers.getContractFactory("Safe", deployer);
```
melanciani: after a more careful review, I think there is a mistake here : we shouldn't deploy a safe contract for each type. Instead, we only need to deploy one implementation contract , and then deploy multiple proxies with different owners/thresholds

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
```diff
@@ -0,0 +1,250 @@
...
+  "task:transferGatewayOwnership",
+  "Transfers ownership of the GatewayConfig contract to the passed Safe Smart Account address.",
+)
+  .addParam(
+    "currentOwnerPrivateKey",
+    "Private key of the current owner of the GatewayConfig contract.",
+    undefined,
+    types.string,
+  )
+  .addParam("smartAccountAddress", "Address of the Safe Smart Account receiving ownership", undefined, types.string)
```
melanciani: maybe rename this input `safeSmartAccountAddress`

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
```diff
@@ -0,0 +1,250 @@
...
+  .addParam("ownerPrivateKeys", "List of private keys of the owners of the Safe Smart Account.", undefined, types.json)
+  .addParam("smartAccountAddress", "Address of the Safe Smart Account accepting ownership.", undefined, types.string)
+  .addOptionalParam(
+    "useInternalProxyAddress",
+    "If proxy address from the /addresses directory should be used.",
+    false,
+    types.boolean,
+  )
+  .setAction(async function ({ ownerPrivateKeys, smartAccountAddress, useInternalProxyAddress }, { ethers }) {
+    // Get the signer wallet.
```
melanciani: -> `// Get the signers' wallets.`

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
```diff
@@ -0,0 +1,250 @@
...
+      gasPrice,
+      gasToken,
+      refundReceiver,
+      signatureBytes,
+    );
+    await execTransactionResponse.wait();
+    console.log(
+      `Ownership of Gateway at address ${gatewayConfigContractAddress} successfully accepted by the Safe Smart Account at address: ${smartAccountAddress}`,
+    );
+  });
```
melanciani: I think we should also have a task to update the pauser (through the owner safe contract) !

melanciani (changes requested):
almost there, but I realize there is a small issue in the task !

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
isaacdecoded: @RomanBredehoft not sure to understand the "deploy multiple proxies with different owners/thresholds" part. The `setup` method [is part of the Safe contract](https://github.com/safe-global/safe-smart-account/blob/main/contracts/Safe.sol#L106-L132) not the [Proxy factory contract](https://github.com/safe-global/safe-smart-account/blob/main/contracts/proxies/SafeProxyFactory.sol#L86-L91), so is the implementation contract which handles the owners and threshold 🤔

gateway-contracts/test/Decryption.ts
isaacdecoded: It was working because most of variables are global to all Describe suites, and this `pauser` in concrete is initialized in the previous `describe("Public Decryption")` 😅 Maybe this is kind of error prone using such a global variables instead of having dedicated ones per Describe suite 🤔

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
melanciani: yes but the implementation contract is just here to give the implementation, there is not need to deploy 2 of them, only 1 is enough. Only the proxies store values like thresholds and owners. So I'm pretty sure the goal is to 
- deploy the safe contract once
- deploy proxies with different inputs **and Nonce** 

so something like : 
```
  const safeFactory = await ethers.getContractFactory("Safe", deployer);
  const safe = await safeFactory.deploy();
  const safeAddress = await safe.getAddress();

  // Proxy 1
  const safeProxyFactoryFactory1 = await ethers.getContractFactory("SafeProxyFactory", deployer);
  const safeProxyFactory1 = await safeProxyFactoryFactory1.deploy();

  const saltNonce1 = 0n;
  await safeProxyFactory1.createProxyWithNonce(safeAddress, safeData1, saltNonce1);

  // Proxy 2
  const safeProxyFactoryFactory2 = await ethers.getContractFactory("SafeProxyFactory", deployer);
  const safeProxyFactory2 = await safeProxyFactoryFactory2.deploy();

  const saltNonce2 = 1n;
  await safeProxyFactory2.createProxyWithNonce(safeAddress, safeData2, saltNonce2);
```

to be refactored into 2 or 3 functions ofc

I realized this while looking at the doc for SafeProxyFactory, in particular [the code](https://github.com/safe-global/safe-smart-account/blob/main/contracts/proxies/SafeProxyFactory.sol#L82): the "implementation contract address" is called a `singleton` specifically for this reason. The Nonce is here for this as well, because the proxy address is not generated based on the msg.sender's address (something made on purpose, like explained [here](https://github.com/safe-global/safe-smart-account/issues/321#issuecomment-865609483)) so we need to be able to build different addresses

gateway-contracts/test/Decryption.ts
melanciani: I thought `beforeEach`'s scope were only for the `describe` section they were defined to 🤔  if that's not the case then yes we should maybe see again how we do our tests. could you please create an issue for this ?

gateway-contracts/hardhat.config.ts
isaacdecoded: Comment added in last commit [here](https://github.com/zama-ai/fhevm/pull/539/commits/017acf719134e2b33de4dee16455779515146f43#diff-4174d32e7c30d8e24f30e119f6bf99a004d8cf98648f139900be43df1d159242R164).

gateway-contracts/test/Decryption.ts
isaacdecoded: Issue opened [here](https://github.com/zama-ai/fhevm-internal/issues/322).

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
isaacdecoded: All right, PR updated to follow the singleton pattern. Now there are two tasks: [one for the deployment of the Safe implementation](https://github.com/zama-ai/fhevm/pull/539/commits/017acf719134e2b33de4dee16455779515146f43#diff-09ccc6aff4469c4273678f21c01d97e8dad9797e4cb467310593098fcf9a4f0eR39-R42) and [other for the Safe proxy deployment](https://github.com/zama-ai/fhevm/pull/539/commits/017acf719134e2b33de4dee16455779515146f43#diff-09ccc6aff4469c4273678f21c01d97e8dad9797e4cb467310593098fcf9a4f0eR60).

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
isaacdecoded: New task for updating the pauser added [here](https://github.com/zama-ai/fhevm/pull/539/commits/017acf719134e2b33de4dee16455779515146f43#diff-09ccc6aff4469c4273678f21c01d97e8dad9797e4cb467310593098fcf9a4f0eR271). Note that I've moved (and renamed) the `/tasks/deployment/safe_smart_accounts.ts` to `/tasks/safeSmartAccount.ts` 🤔 I mean, now it seems to me that they are not "deployment" tasks but Gateway + Safe Smart Account utilities.

gateway-contracts/tasks/safeSmartAccounts.ts
```diff
@@ -0,0 +1,365 @@
...
+  // Get a deployer wallet
+  const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
+  const deployer = new Wallet(deployerPrivateKey).connect(ethers.provider);
+
+  // Deploy a new Safe implementation contract
+  const safeFactory = await ethers.getContractFactory("Safe", deployer);
+  const safe = await safeFactory.deploy();
+  const safeAddress = await safe.getAddress();
+
+  console.log(`Safe Smart Account implementation deployed at address: ${safeAddress}`);
```
melanciani: I think it still makes sense to write the address to an env file so you can load the address directly in the proxy deployment tasks

gateway-contracts/tasks/safeSmartAccounts.ts
```diff
@@ -0,0 +1,365 @@
...
+    }
+    const safeProxyAddress = event.args.proxy;
+
+    if (safeProxyAddress === ethers.ZeroAddress) {
+      throw new Error("Safe proxy address not found");
+    }
+
+    console.log(`Safe Smart Account proxy successfully deployed at address: ${safeProxyAddress}\n`);
+  });
+
```
melanciani, line 211: I think it still makes sense to keep 2 dedicated tasks for owner and pauser that calls this `deploySafeSmartAccountProxy` task (with an additional name input like before). The name would then be used to write the address in an env file ! without them, ZWS has no "computational way" of knowing the deployed addresses (as its only logs)

melanciani (changes requested):
sorry, almost there ! I think we can still improve it a bit

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
melanciani: I mean, deployment tasks are essentially gateway utilities imo haha it's to make their life much simpler by providing the deployment scripts ourselves, instead of letting them dig into our structure !

gateway-contracts/test/CiphertextCommits.ts
melanciani: confirmed by Levent : we don't need them

gateway-contracts/tasks/deployment/smart_accounts.ts
melanciani: they don't need them for now

gateway-contracts/tasks/deployment/safe_smart_accounts.ts
isaacdecoded: Yes, they're all utilities, is true. Imo, these "safeSmartAccount" ones are not related to the deployment itself but more for pre- and post-deployment stuff, similar to "addHostChains", "upgradeContracts", etc.

gateway-contracts/tasks/safeSmartAccounts.ts
isaacdecoded: Task separated into two dedicated ones in [the last commit here](https://github.com/zama-ai/fhevm/pull/539/commits/7efcf1536c24fd01ba3826e56d912cdd4570f421#diff-09ccc6aff4469c4273678f21c01d97e8dad9797e4cb467310593098fcf9a4f0eR163). This also introduces the new .env files to store the deployed Safe smart account addresses and their usage in the different tasks.

gateway-contracts/tasks/safeSmartAccounts.ts
```diff
@@ -36,6 +41,94 @@ async function getSortedSignatures(signers: Wallet[], transactionHash: string):
...
+    throw new Error("ProxyCreation event not found in transaction receipt");
+  }
+  const safeProxyAddress = event.args.proxy;
+
+  if (safeProxyAddress === ethers.ZeroAddress) {
+    throw new Error("Safe proxy address not found");
+  }
+
+  const nameSnakeCase = pascalCaseToSnakeCase(name);
+  const envFilePath = path.join(ADDRESSES_DIR, `.env.${nameSnakeCase}`);
```
melanciani: like for gateway contracts, I'm not sure I see the point of having one env file per type, I think you can simply put all env vars in a single `.env.safe_smart_contracts` file 

you can put the implementation contract + the proxy contracts in it imo

melanciani (changes requested):
I think you can simplify and only have one additional env file, but apart from that looks good !

gateway-contracts/tasks/safeSmartAccounts.ts
isaacdecoded: Simplification included [here](https://github.com/zama-ai/fhevm/pull/539/commits/e0fcb9327e94dca6741255c9db7fb2913a6d5da9), thanks! 🙏

melanciani (approved):
great, thanks a lot for the huge work, these are very nice tools that will be helpful for ZWS !

note that we should probably consider these follow-up issues soon (this quarter) for improving our use of safe smart contracts : 
- https://github.com/zama-ai/fhevm-internal/issues/241
- https://github.com/zama-ai/fhevm-internal/issues/327

+ I noticed we could most likely refactor some parts (with other tasks), which we can track through this epic : https://github.com/zama-ai/fhevm-internal/issues/328

PR #544, opened 2025-07-08
feat(coprocessor): add optional LIFO processing mode to SNS worker

goshawk-3:
fixes https://github.com/zama-ai/fhevm-internal/issues/202


dartdart26 (approved):
Nice!

PR #550, opened 2025-07-10
feat(kms-connector): use new kms-core api

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/12


tawadaa (approved):
LGTM, thanks!

PR #551, opened 2025-07-10
Rudy/release/reorgs

rudy-6-4:


coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -310,7 +335,160 @@ impl InfiniteLogIter {
...
+            );
+            let Ok(logs) = self.get_logs_at_hash(missing_block.hash).await
+            else {
+                error!(
+                    block_summary = ?missing_block,
+                    "Reorg. Cannot get logs for missing block, skipping it.",
+                );
+                continue; // skip this block
+            };
+            self.catchup_logs.extend(logs);
```
rudy-6-4, line 484: add to history

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -310,7 +335,160 @@ impl InfiniteLogIter {
...
+        }
+    }
+
+    async fn check_reorgs(&mut self) -> bool {
+        if !self.block_history.is_ready_to_detect_reorg() {
```
rudy-6-4: to do after inserting current block in history otherwise block history is never ready

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -310,7 +335,160 @@ impl InfiniteLogIter {
...
+
+    async fn get_missings_ancestors(
+        &self,
+        mut current_block: BlockSummary,
+    ) -> Vec<BlockSummary> {
+        // iter on current block ancestors to collect missing blocks
+        let mut missing_blocks: Vec<BlockSummary> = Vec::new();
+        for _ in 0..self.reorg_maximum_duration_in_blocks {
+            let parent_block_hash = current_block.parent_hash;
+            if self.block_history.is_known(&parent_block_hash) {
```
rudy-6-4, line 434: faling is_known is O(N) in history size, probably not ok in that loop.

rudy-6-4: but reorg are a dozain block so it's ok

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -310,7 +330,180 @@ impl InfiniteLogIter {
...
+        mut current_block: BlockSummary,
+    ) -> Vec<BlockSummary> {
+        // iter on current block ancestors to collect missing blocks
+        let mut missing_blocks: Vec<BlockSummary> = Vec::new();
+        for _ in 0..self.reorg_maximum_duration_in_blocks {
+            let parent_block_hash = current_block.parent_hash;
+            let is_known = if cfg!(feature="test-simulate-reorg") {
+                let is_known = parent_block_hash.get(0).unwrap_or(&0) % 3 == 0;
+                if is_known != self.block_history.is_known(&parent_block_hash) {
+                    eprintln!("Simulating reorg");
```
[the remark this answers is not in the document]
rudy-6-4: This is visible in test only.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -310,7 +330,180 @@ impl InfiniteLogIter {
...
+            );
+            let Ok(logs) = self.get_logs_at_hash(missing_block.hash).await
+            else {
+                error!(
+                    block_summary = ?missing_block,
+                    "Cannot get logs for missing block, skipping it.",
+                );
+                continue; // skip this block
+            };
+            self.catchup_logs.extend(logs);
```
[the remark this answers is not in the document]
rudy-6-4: Good catch

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
```diff
@@ -0,0 +1,78 @@
...
+            hash: block.header.hash,
+            parent_hash: block.header.parent_hash,
+            timestamp: block.header.timestamp,
+        }
+    }
+}
+
+pub struct BlockHistory {
+    ordered_blocks: VecDeque<BlockSummary>,
+    capacity: usize,
```
goshawk-3: You can use `order_blocks.capacity()` instead

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const MAXIMUM_NUMBER_OF_COMPETING_CHAIN: usize = 5;
+const MINIMUM_HISTORY_SIZE: usize = 2; // current block + at least old block
+
+impl BlockHistory {
+    pub fn new(expected_reorg_duration: usize) -> Self {
+        // we take extra margin for history
+        let capacity =
+            expected_reorg_duration * 2 * MAXIMUM_NUMBER_OF_COMPETING_CHAIN;
```
goshawk-3, line 38: Isn't that a bit overcomplicated?

I'd simply add a CLI param to explicitly define  the **MAX_HISTORY_SIZE** instead

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
```diff
@@ -0,0 +1,78 @@
...
+        Self {
+            ordered_blocks: VecDeque::with_capacity(capacity),
+            capacity,
+        }
+    }
+
+    pub fn is_ready_to_detect_reorg(&self) -> bool {
```
goshawk-3, line 48: Can you explain ( add comments)  to clarify further?

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
```diff
@@ -0,0 +1,78 @@
...
+        self.ordered_blocks.len() >= MINIMUM_HISTORY_SIZE
+    }
+
+    pub fn is_known(&self, block_hash: &BlockHash) -> bool {
+        // we process the history in reverse to have O(1) on no reorg
+        let slices = self.ordered_blocks.as_slices();
+        for history_slice in [slices.1, slices.0].iter() {
+            for historic_block in history_slice.iter().rev() {
+                if historic_block.hash == *block_hash {
```
goshawk-3, line 59: You can use here more idiomatic approach
```
 let slices = self.ordered_blocks.as_slices();
        [slices.1, slices.0]
            .iter()
            .flat_map(|slice| slice.iter().rev())
            .any(|block| block.hash == *block_hash)
  ```

File: .github/workflows/coprocessor-cargo-tests.yml [hunk 10]
```diff
@@ -95,3 +95,8 @@ jobs:
       run: |
         DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor TXN_SENDER_TEST_GLOBAL_LOCALSTACK=1 cargo test --release -- --test-threads=1
       working-directory: coprocessor/fhevm-engine
+
+    - name: Run tests with incidents
+      run: |
+        DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor TXN_SENDER_TEST_GLOBAL_LOCALSTACK=1 cargo test -p fhevm-listener --release --features test-simulate-reorg -- test_listener_restart
```
goshawk-3: As discussed, we should use a test db instead of relying on global ENV in CI.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+        false
+    }
+
+    pub fn block_did_not_changed(&self, block_hash: &BlockHash) -> bool {
+        self.ordered_blocks.back().map(|b| &b.hash) == Some(block_hash)
+    }
+
+    pub fn last_block(&self) -> Option<BlockSummary> {
```
goshawk-3: This can be renamed to local_tip()/local_head() or just `tip()` to refer to the latest block received locally.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
```diff
@@ -0,0 +1,78 @@
...
+        self.ordered_blocks.back().copied()
+    }
+
+    pub fn add_block(&mut self, block: BlockSummary) {
+        while self.ordered_blocks.len() >= self.capacity {
+            self.ordered_blocks.pop_front();
+        }
+        self.ordered_blocks.push_back(block);
```
goshawk-3, line 79: The assumption here is that these are ordered blocks but you don't check if the height of the new block is higher that the tip height.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -20,8 +23,13 @@ use clap::Parser;
 
 use rustls;
 
+pub mod block_history;
 use crate::contracts::{AclContract, TfheContract};
 use crate::database::tfhe_event_propagate::{ChainId, Database};
+use block_history::{BlockHash, BlockHistory, BlockSummary};
+
+const REORG_RETRY_GET_LOGS: u64 = 10; // retry 10 times to get logs for a block
```
goshawk-3, line 31: Shouldn't this be unlimited retry?

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -273,35 +300,28 @@ impl InfiniteLogIter {
...
+        let Ok(logs) = self.get_logs_at_hash(block.hash).await else {
             return false;
         };
         if logs.is_empty() {
             return false;
         }
         info!(
-            block = block,
+            block = ?block.number,
+            block_hash = ?block.hash,
```
goshawk-3, line 326: Does this display the hash as hex?

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -310,7 +330,182 @@ impl InfiniteLogIter {
...
+            missing_blocks.push(current_block);
+            if parent_block_hash == BlockHash::ZERO {
+                // can happen in tests
+                break;
+            }
+            let Some(parent_block) = self.get_block(parent_block_hash).await
+            else {
+                error!(
+                    parent_block_hash = ?parent_block_hash,
+                    "Reorg chaining stopped. Cannot get parent block.",
```
goshawk-3, line 446: Should we trigger some alarm or update any health-check status here?

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -310,7 +330,182 @@ impl InfiniteLogIter {
...
+        for missing_block in missing_blocks {
+            info!(
+                block_summary = ?missing_block,
+                "Missing block detected. Trying to get block",
+            );
+            let Ok(logs) = self.get_logs_at_hash(missing_block.hash).await
+            else {
+                error!(
+                    block_summary = ?missing_block,
+                    "Cannot get logs for missing block, skipping it.",
```
goshawk-3, line 475: If a missing block cannot be retrieved, it should raise an alert or update the system’s health-check status.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -261,7 +287,8 @@ impl InfiniteLogIter {
             if paging_to_block + 1 > to_block {
                 self.catchup_blocks = None;
             }
-        } else if nb_logs == 0 { // either empty or futur block
+        } else if nb_logs == 0 {
+            // either empty or futur block
```
dartdart26:
```suggestion
            // either empty or future block
```

.github/workflows/coprocessor-cargo-tests.yml
rudy-6-4: This is on release branch.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
rudy-6-4: The issue is that it's not obvious to link history size to reorg dynamic (size + number of competing chain).
Why is  expected_reorg_duration too complicate ?

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
rudy-6-4: I think loops are easier to read in that case even it's more verbose.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
rudy-6-4: Yes. It's done on reorg check so either the last block when no reorg or a past block (inserted from older to newer).
But thanks it shows on reorg we should maintain that invariant which is broken now. (add the block post reorg at the end just after reorg catchup)

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
rudy-6-4: I am not sure we want to stuck the listener. Wdyt ? @dd23 @antoniupop 
But yes on main we should report something somewhere.
Unlimited means the health check will be unhealthy after some time because the listener is stuck. (So it will be restarted after some time).

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
rudy-6-4: I think yes.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
rudy-6-4: what do you want as alert ?

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
rudy-6-4: Added
// it needs to have some data before using it to detect reorg
// e.g. at start, an unknown ancestor in history is considered a reorg bloc

.github/workflows/coprocessor-cargo-tests.yml
rudy-6-4: You fixed it in main branch.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
rudy-6-4: It's not easy to maintain the invariant in this function. It would require linear search for each insert.
That linear search is done in the check reorg function via is_known. We only insert in case the block is unkown, so it can only come from a reorg or it's a new block.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
rudy-6-4: Done

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
rudy-6-4: Done

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
```diff
@@ -0,0 +1,82 @@
...
+            for historic_block in history_slice.iter().rev() {
+                if historic_block.hash == *block_hash {
+                    return true;
+                }
+            }
+        }
+        false
+    }
+
+    pub fn block_did_not_changed(&self, block_hash: &BlockHash) -> bool {
```
dartdart26:
```suggestion
    pub fn block_has_not_changed(&self, block_hash: &BlockHash) -> bool {
```

coprocessor/fhevm-engine/fhevm-listener/src/cmd/block_history.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    pub fn tip(&self) -> Option<BlockSummary> {
+        self.ordered_blocks.back().copied()
+    }
+
+    pub fn add_block(&mut self, block: BlockSummary) {
+        if self.ordered_blocks.len() == self.ordered_blocks.capacity() {
+            self.ordered_blocks.pop_front();
+        }
```
dartdart26, line 78: Why do we have to check against .capacity() here?

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
antoniupop: I think it's better to have this error than to block the catchup mechanism in case of irretrievable block, but do agree this should update the health status and probably automatically retry running the catchup. If we don't block we can make some progress up to the missing block at least.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
antoniupop: See comment below - I think it's better not to block on this.

antoniupop (approved):
LGTM! If we can have a retry mechanism, for the catchup in case of missing block I think it's fine.

dartdart26 (approved):
Code looks OK, but I cannot verify if it works correctly just by looking at it, so it is hard for me to give good feedback. Nevertheless, I will approve and we can try test and see if it works in real conditions.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
rudy-6-4: https://github.com/zama-ai/fhevm-internal/issues/286

rudy-6-4:
I have added retry for get_block and get_current_block

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
rudy-6-4: https://github.com/zama-ai/fhevm-internal/issues/286

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
rudy-6-4: This will be done on main.

PR #553, opened 2025-07-10
ci(kms-connector): add deps analysis

eudelins-zama:
Add dependency analysis in CI for the `kms-connector` with the same rules as for the `coprocessor`

Closes https://github.com/zama-ai/fhevm-internal/issues/272


dartdart26 (approved):
Nice!

PR #561, opened 2025-07-11
feat(test-suite): add an optional api key for relayer sdk

avitali-zama:
It adds an optional API Key field when calling the relayer using the `relayer-sdk` in tests


jatZama:
I don't think this would work unless you bump the version of relayer SDK in package.json

avitali-zama:
> I don't think this would work unless you bump the version of relayer SDK in package.json

I'll move this as a draft until we release the next `relayer-sdk` version

eudelins-zama:
Closing as discussed on Slack :slightly_smiling_face:

PR #567, opened 2025-07-14
Pre-compute dependence chains and filter on allowed

antoniupop:


coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -46,22 +46,31 @@ pub struct Database {
     pool: sqlx::Pool<Postgres>,
     tenant_id: TenantId,
     chain_id: ChainId,
+    bucket_cache: std::sync::Arc<
```
rudy-6-4: Why is the ARC needed ? The cache is private to the struct !? Is it shared somewhere ?

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -46,22 +46,31 @@ pub struct Database {
...
         coprocessor_api_key: &CoprocessorApiKey,
         chain_id: ChainId,
+        bucket_cache_size: u16,
     ) -> Self {
         let pool = Self::new_pool(url).await;
         let tenant_id =
             Self::find_tenant_id_or_panic(&pool, coprocessor_api_key).await;
+        let bucket_cache =
+            std::sync::Arc::new(tokio::sync::RwLock::new(lru::LruCache::new(
+                std::num::NonZeroU16::new(bucket_cache_size).unwrap().into(),
```
rudy-6-4: Maybe have minimal safe size instead of an error (could be 1).
e.g. `bucket_cache_size.max(MINIMAL_BUCKET_CACHE_SIZE)`
Also bad size like 0 could be checked by command line, so the error is clearer.
How do you disable the cache ? (I think not being able to disable is ok but you know better)

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
antoniupop: Good point - it's not needed.

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
antoniupop: I guess we could allow 0 and that disables it - the problem is if we don't use it at all, performance would be bad with the new query for work. 
I think having a forced minimum size is best.

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -165,75 +193,122 @@ impl Database {
...
+        query.execute(&self.pool).await.map(|_| ())
+    }
+
+    async fn sort_computation_into_bucket(
+        &self,
+        output: &Handle,
+        dependencies: &[&Handle],
+        transaction_hash: &Option<Handle>,
+    ) -> PrimitiveDateTime {
+        let mut bucket_cache = self.bucket_cache.write().await;
```
rudy-6-4: Just to think about it. Could lead to contention when this will be made concurrent. But I think It's too early to make it future proof (could do read + write lock for optimistic path when the bucket already exist, and redo a full write lock for a recheck/write or create)

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -165,75 +193,122 @@ impl Database {
...
+                    bucket_cache.put(*txh, ce);
                 }
+                return ce;
             }
         }
+        // If this computation is not linked to any others, assign it
+        // to a new empty bucket and add output handle and transaction
+        // hash where relevant
+        let t = OffsetDateTime::now_utc();
+        let t = PrimitiveDateTime::new(t.date(), t.time());
```
rudy-6-4: Why using time here ? Do you sort bucket in workers ? Could be the root handle otherwise ?

rudy-6-4 (commented):
LGTM. Reviewed listener part only.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
```diff
@@ -93,6 +93,13 @@ pub struct Args {
 
     #[arg(long, default_value = "8080", help = "Health check port")]
     pub health_port: u16,
+
+    #[arg(
+        long,
+        default_value = "128",
```
rudy-6-4, line 94: Is the default based on real need ? On high load it could be just one block size.

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
antoniupop: It could be, yes. The main reason was to be able to sort the buckets, but that turned out not to be viable because they can live for a long time and the oldest bucket shouldn't be prioritised. 
I think using the root handle is reasonable - I'll switch then.

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
antoniupop: Updated to be more fine grained.

coprocessor/fhevm-engine/fhevm-listener/src/cmd/mod.rs
antoniupop: As discussed we keep a small default - rationale is that this only helps with very active contracts where all transactions are dependent. It is unlikely that a high number would be very active concurrently. This avoids high overhead in the listener.

coprocessor/fhevm-engine/coprocessor/src/daemon_cli.rs
```diff
@@ -29,9 +29,13 @@ pub struct Args {
     pub server_maximum_ciphertexts_to_get: usize,
 
     /// Work items batch size
-    #[arg(long, default_value_t = 10)]
+    #[arg(long, default_value_t = 50)]
     pub work_items_batch_size: i32,
 
+    /// Number of dependence chains to fetch per worker
+    #[arg(long, default_value_t = 10)]
+    pub dependence_chains_per_worker: i32,
```
rudy-6-4: Does it means if there is only lots of short chains we don't go up to work_items_batch_size ?
Also in the name not clear that it's a maximum for each batch or until the chain is terminated.
Maybe max_bucket_per_batch or maximum_dependence_chain_per_batch or dependence_chain_per_batch.

coprocessor/fhevm-engine/coprocessor/src/server.rs
```diff
@@ -895,3 +901,27 @@ impl CoprocessorService {
         Ok(tonic::Response::new(result))
     }
 }
+
+async fn sort_computation_into_bucket(
+    computations: &[&crate::server::coprocessor::AsyncComputation],
+) -> Vec<Vec<u8>> {
+    let mut res: Vec<Vec<u8>> = vec![vec![0]; computations.len()];
+    let mut cache: HashMap<Vec<u8>, Vec<u8>> = HashMap::with_capacity(computations.len());
```
rudy-6-4: Could be named handle_to_bucket.

coprocessor/fhevm-engine/coprocessor/src/server.rs
```diff
@@ -895,3 +901,27 @@ impl CoprocessorService {
...
+    'comps: for (idx, comp) in computations.iter().enumerate() {
+        let output = &comp.output_handle;
+        for ih in comp.inputs.iter() {
+            if let Some(Input::InputHandle(input)) = &ih.input {
```
rudy-6-4: Could be denested, if you like.
```
let Some(Input::InputHandle(input)) = &ih.input else {
     continue;
}
let Some(ec) = cache.get(input).cloned() else {
    continue;
}
cache.insert(output.to_owned(), ce.to_owned());
res[idx] = ce;
continue 'comps;
```

coprocessor/fhevm-engine/coprocessor/src/server.rs
```diff
@@ -895,3 +901,27 @@ impl CoprocessorService {
         Ok(tonic::Response::new(result))
     }
 }
+
+async fn sort_computation_into_bucket(
+    computations: &[&crate::server::coprocessor::AsyncComputation],
+) -> Vec<Vec<u8>> {
```
rudy-6-4: Maybe you declare `type Handle = Vec<u8>; type BucketId = Vec<u8>;` to make it easier to read.

coprocessor/fhevm-engine/coprocessor/src/daemon_cli.rs
antoniupop: Yes that's correct - we don't want to acquire from too many separate dependence chains or we'd starve other workers. You're right dependence_chains_per_batch is probably better - will update.

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -117,15 +117,45 @@ async fn tfhe_worker_cycle(
...
+              -- Get set of complete transactions
+              SELECT c.tenant_id, c.output_handle, c.transaction_id, ah.handle
+              FROM computations c LEFT JOIN allowed_handles ah ON c.output_handle = ah.handle
+              WHERE ( c.is_completed = false
+              AND c.is_error = false
+              AND c.transaction_id IS NULL )
+              OR c.transaction_id IN (
+                -- Select full transactions out of the buckets we identified below
+                SELECT transaction_id FROM computations
+                WHERE is_completed = false
```
rudy-6-4: Maybe a flag for untouched computations would be faster ?
But since is_completed = false and is_error = false both select very few rows, probably not worth it.

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -140,22 +170,32 @@ async fn tfhe_worker_cycle(
...
                 }
+                // If this operation is not part of a labelled
+                // transaction, we need to treat it as if its output
+                // is allowed (meaning that we will consider it needed
+                // and will save its output in the DB) as otherwise we
+                // cannot reasonably keep track of all needed
+                // operations within the set of computations that are
+                // not part of a transaction
+                if w.transaction_id.is_none() {
+                    w.is_allowed = Some(true);
```
rudy-6-4, line 262: Is it required ? line 319 says it default to true unless it can be Some(false).

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -365,18 +407,25 @@ async fn tfhe_worker_cycle(
...
                     continue;
                 }
-                let r = &mut res
-                    .iter_mut()
-                    .find(|(h, _)| *h == w.output_handle)
-                    .unwrap()
-                    .1;
+                // If no result or error is present, this computation
+                // was either not needed or not allowed
+                let Some(r) = &mut res.iter_mut().find(|(h, _)| *h == w.output_handle) else {
```
rudy-6-4: Not related to review but does it linear search res ? Isn't it quadratic then ? Maybe a set or something in res could make it faster, also the graph structure probably has something to do that quickly if from a graph library.

coprocessor/fhevm-engine/db-migration/migrations/20250708100500_add_transaction_id_for_computations.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE computations
```
rudy-6-4: No index ?
But a sql query has a where clause on it. I am checking with grok since my postgresql skills are on holidays.
It says it need to perform a linear scan but you may know better. Can't the OR clause be completed with something reducing the number of element to check ?

coprocessor/fhevm-engine/scheduler/src/dfg.rs
```diff
@@ -88,4 +101,51 @@ impl DFGraph {
...
         res
     }
+
+    fn is_needed(&self, index: usize) -> bool {
+        let node_index = NodeIndex::new(index);
+        let node = self.graph.node_weight(node_index).unwrap();
+        if node.is_allowed || node.is_needed {
+            true
+        } else {
+            for edge in self.graph.edges_directed(node_index, Direction::Outgoing) {
```
rudy-6-4, line 132: maybe something similar to replace the find call in a previous comment

coprocessor/proto/coprocessor.proto
```diff
@@ -48,6 +48,7 @@ message TrivialEncryptRequestSingle {
 
 message AsyncComputation {
   fhevm.common.FheOperation operation = 1;
+  bytes transaction_id = 2;
```
rudy-6-4, line 51: Is it still needed ? Can't be deleted ?

rudy-6-4 (commented):
Looks good, but the efficiency of the main sql request need to be verified, maybe postgresql can explain how it will run it to be sure there is not perf issue.

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: Yes, the case transaction_id == None and  is_allowed == Some(false) occurs either in legagy elements in the DB (where transaction ID wasn't set) or if for some reason we have operations inserted without a transaction ID later.

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: Good point - I'll change this to traverse the graph directly instead.

coprocessor/proto/coprocessor.proto
antoniupop: We've decided to keep it just for the CI tests for now, but indeed it is going to be removed.

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -165,75 +195,125 @@ impl Database {
...
+                bucket_cache_write.put(*output, ce);
+                // As the transaction hash was not in the cache, add
+                // it to this bucket as well
+                if let Some(txh) = transaction_hash {
+                    bucket_cache_write.put(*txh, ce);
                 }
+                return ce;
             }
         }
+        drop(bucket_cache_read);
```
rudy-6-4, line 278: Why not also dropping just before line 271 (for consistency at least) vs not at all ? That looks like the same extra lock time.

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -165,75 +195,125 @@ impl Database {
...
-                }
-                Err(sqlx_err) => {
-                    return Err(sqlx_err);
+            tenant_id as i32,
+            output_handle,
+            &dependencies,
+            fhe_operation as i16,
+            is_scalar,
+            bucket.to_vec(),
+            if let Some(txh) = log.transaction_hash {
```
rudy-6-4: log.transaction.map(|txh| txh.to_vec())
and or could be done on a separate line.

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -124,16 +135,25 @@ impl Database {
...
         tenant_id: TenantId,
         result: &Handle,
         dependencies_handles: &[&Handle],
         dependencies_bytes: &[Vec<u8>], /* always added after
                                          * dependencies_handles */
         fhe_operation: FheOperation,
         scalar_byte: &FixedBytes<1>,
+        log: &alloy::rpc::types::Log<TfheContractEvents>,
     ) -> Result<(), SqlxError> {
+        let bucket = self
```
rudy-6-4, line 150: this is done twice (at source level, also in insert_computation), since the log is also passed, it could be done once in insert_computation_inner

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -95,14 +95,15 @@ async fn tfhe_worker_cycle(
             tokio::select! {
                 _ = listener.try_recv() => {
                     WORK_ITEMS_NOTIFICATIONS_COUNTER.inc();
-                    info!(target: "tfhe_worker", "Received work_available notification from postgres");
+                    info!(target: "tfhe_worker", "Received event_allowed_handle notification from postgres");
```
rudy-6-4, line 98: we don't do anything until something is allowed ?

rudy-6-4: Maybe there is come clean up to do (ie the auto notify in sql and in server.rs).
git grep work_available
coprocessor/.sqlx/query-be2b10355e815489836278ccb55110baa4dae151e8118da3cc58654f46123f97.json:  "query": "NOTIFY work_available",
coprocessor/src/server.rs:            query!("NOTIFY work_available")
db-migration/migrations/20250303135355_fhevm_listner_auto_notify.sql:CREATE OR REPLACE FUNCTION notify_work_available()
db-migration/migrations/20250303135355_fhevm_listner_auto_notify.sql:    NOTIFY work_available;
db-migration/migrations/20250303135355_fhevm_listner_auto_notify.sql:    EXECUTE FUNCTION notify_work_available();

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -117,21 +118,67 @@ async fn tfhe_worker_cycle(
         let now = std::time::SystemTime::now();
         let the_work = query!(
```
rudy-6-4, line 119: Indentation in SQL is not helping, Is the indentation broken by cargo fmt ? (can be locally disabled)

rudy-6-4: E.g. from copilot
you can even dedent inside the string to have start more on the left.

```sql

WITH selected_computations AS (
    -- Get all computations from such transactions
    (
        SELECT 
            c.tenant_id, 
            c.output_handle, 
            ah.handle
        FROM computations c
        LEFT JOIN allowed_handles ah ON c.output_handle = ah.handle
        WHERE c.transaction_id IN (
            -- Select transaction IDs with uncomputed handles out of the dependence buckets
            SELECT transaction_id
            FROM computations
            WHERE is_completed = false
              AND is_error = false
              AND dependence_chain_id IN (
                  WITH dep_chains AS (
                      -- Find oldest uncomputed allowed handles and get their dependence buckets
                      SELECT dependence_chain_id
                      FROM computations
                      WHERE (tenant_id, output_handle) IN (
                          SELECT tenant_id, handle
                          FROM allowed_handles
                          WHERE is_computed = false
                          ORDER BY allowed_at
                          LIMIT $2
                      )
                  )
                  SELECT DISTINCT dependence_chain_id FROM dep_chains
              )
            ORDER BY schedule_order
            LIMIT $1
        )
    )
    -- For legacy reasons (or whenever no transaction ID is available) we need to also select unsorted computations
    UNION ALL
    (
        SELECT 
            tenant_id, 
            output_handle, 
            NULL
        FROM computations 
        WHERE is_completed = false
          AND is_error = false
          AND transaction_id IS NULL  
        ORDER BY schedule_order
        LIMIT $1
    )
)
-- Acquire all computations from this transaction set
SELECT 
    c.tenant_id, 
    c.output_handle, 
    c.dependencies, 
    c.fhe_operation, 
    c.is_scalar,
    sc.handle IS NOT NULL AS is_allowed, 
    c.dependence_chain_id,
    COALESCE(c.transaction_id) AS transaction_id, 
    c.is_completed
FROM computations c
JOIN selected_computations sc
    ON c.tenant_id = sc.tenant_id
   AND c.output_handle = sc.output_handle
FOR UPDATE SKIP LOCKED;

```

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -117,21 +118,67 @@ async fn tfhe_worker_cycle(
...
+            args.work_items_batch_size as i32,
+            args.dependence_chains_per_batch as i32,
         )
         .fetch_all(trx.as_mut())
         .await?;
         s.set_attribute(KeyValue::new("count", the_work.len() as i64));
         s.end();
-        immedially_poll_more_work = !the_work.is_empty();
         if the_work.is_empty() {
             continue;
```
rudy-6-4, line 214: maybe it could log as before that despite the event_allowed_handle (or timeout) there is nothing to do (as before)?
or not if this happens too frequently due to worker competition.

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -117,21 +118,67 @@ async fn tfhe_worker_cycle(
...
+		  LEFT JOIN allowed_handles ah
+		  ON c.output_handle = ah.handle
+                WHERE c.transaction_id IN (
+                  -- Select transaction IDs with uncomputed handles
+                  -- out of the dependence buckets
+                  SELECT transaction_id FROM computations
+                  WHERE is_completed = false
+                  AND is_error = false
+                  AND dependence_chain_id IN (
+                    WITH dep_chains as (
```
rudy-6-4: Is this WITH required ?
dep_chains is immediately in a SELECT and not reused

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
```diff
@@ -117,21 +118,67 @@ async fn tfhe_worker_cycle(
...
+                      	LIMIT $2
+		      )
+                    )
+                    SELECT DISTINCT dependence_chain_id FROM dep_chains
+                  )
+                  ORDER BY schedule_order
+                  LIMIT $1
+                )
+              )
+	      -- For legacy reasons (or whenever no transaction ID is
```
rudy-6-4: What prevent us from making it mandatory after some time to compute all legacy computations ?

rudy-6-4 (approved):

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: Yes, distinct and order by don't mix (need to be on the same columns).

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: It still does, just the condition is updated to be whether we've computed something rather whether the query returned something. In most cases this should be the same.

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: Oh yes indeed! I forgot to clean this up.

coprocessor/fhevm-engine/fhevm-listener/src/database/tfhe_event_propagate.rs
antoniupop: Indeed, but insert_computation_bytes has the scalars that are added to dependencies, which complicates the sorting into buckets since those scalars shouldn't be included.

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: That's a good question - I would think that this would would be removable as soon as all outstanding computations are guaranteed to have transaction IDs... But I'm not sure if how we can synchronize this on deployment to testnet. I fully agree that this will be removable soon after we deploy and run through any backlog.

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: Done

coprocessor/fhevm-engine/coprocessor/src/tfhe_worker.rs
antoniupop: We'll re-visit this in a subsequent PR.

PR #569, opened 2025-07-14
feat(charts): add affinity and tolerations to anvil-node

tawadaa:


charts/anvil-node/values.yaml
```diff
@@ -2,19 +2,19 @@ nameOverride:
 
 image:
   name: ghcr.io/foundry-rs/foundry
-  tag: stable
+  tag: latest
```
dartdart26: Is it important that we change to latest from stable?

dartdart26 (approved):
Looks good! Just one comment re latest tag.

charts/anvil-node/values.yaml
tawadaa: Fixed, we keep `stable`

PR #571, opened 2025-07-14
feat(common): add all contracts verification task

isaacdecoded:
Closes [#200](https://github.com/zama-ai/fhevm-internal/issues/200)

### DIFF
--- a/gateway-contracts/tasks/blockExplorerVerify.ts
+++ b/gateway-contracts/tasks/blockExplorerVerify.ts
@@ -24,12 +24,10 @@ task("task:verifyCiphertextCommits")
     await run("verify:verify", {
       address: proxyAddress,
       constructorArguments: [],
-      force: true,
     });
     await run("verify:verify", {
       address: implementationAddress,
       constructorArguments: [],
-      force: true,
     });
   });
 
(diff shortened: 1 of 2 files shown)


gateway-contracts/tasks/blockExplorerVerify.ts
```diff
@@ -24,12 +24,10 @@ task("task:verifyCiphertextCommits")
     await run("verify:verify", {
       address: proxyAddress,
       constructorArguments: [],
-      force: true,
```
melanciani, line 27: what is the impact of such change ? what was it used for and why is it not needed ?

melanciani (approved):
looks good thanks, I just have a small question !

gateway-contracts/tasks/blockExplorerVerify.ts
isaacdecoded: This `force` option was introduced [in this PR](https://github.com/zama-ai/fhevm/commit/5457c42e51c20113400001b780e407f8faeced3e#diff-349040b943f377ddd77d9fbb359d37dfbf03005238b139f5f2112dedf5f703c0R27) (e.g. line 27) while I was testing the verify execution, but I forgot to remove it since it's not really needed.

jatZama (approved):
LGTM

PR #572, opened 2025-07-14
refactor(gateway-contracts): consolidate proxy contract addresses

isaacdecoded:
Closes https://github.com/zama-ai/fhevm-internal/issues/195

### DIFF
--- a/.github/workflows/gateway-contracts-upgrade-tests.yml
+++ b/.github/workflows/gateway-contracts-upgrade-tests.yml
@@ -107,12 +107,14 @@ jobs:
 
       # Executing task:deployEmptyUUPSProxies is required to generate the proxy contract addresses, which are then used
       # for compiling the current and new implementations during the upgrade task.
+      # The docker cp command is required until previous-contracts uses separated .env and addresses files.
       - name: Prepare and copy previous contract into current directory
         working-directory: current-fhevm/gateway-contracts
         run: |
           cp .env.example .env
           cp -r ../../previous-fhevm/gateway-contracts/contracts ./previous-contracts
           npx hardhat task:deployEmptyUUPSProxies
+          docker cp deploy-gateway-contracts:/app/addresses ./
 
       - name: Upgrade GatewayConfig contract
         working-directory: current-fhevm/gateway-contracts
--- a/.github/workflows/gateway-contracts-upgrade-tests.yml
+++ b/.github/workflows/gateway-contracts-upgrade-tests.yml
@@ -107,7 +107,8 @@ jobs:
 
       # Executing task:deployEmptyUUPSProxies is required to generate the proxy contract addresses, which are then used
       # for compiling the current and new implementations during the upgrade task.
-      # The docker cp command is required until previous-contracts uses separated .env and addresses files.
+      # The docker cp command is required until previous-contracts uses separated .env and addresses files,
+      # see https://github.com/zama-ai/fhevm-internal/issues/217.
       - name: Prepare and copy previous contract into current directory
         working-directory: current-fhevm/gateway-contracts
         run: |
--- a/gateway-contracts/scripts/ensure_proxy_addresses.ts
+++ b/gateway-contracts/scripts/ensure_proxy_addresses.ts
@@ -21,7 +21,7 @@ function main(): void {
     const addressesDir = "addresses";
 
     // Get gateway config address env file
-    const gatewayConfigAddressEnv = path.join(addressesDir, ".env.gateway_config");
+    const gatewayConfigAddressEnv = path.join(addressesDir, ".env.gateway");
 
     let shouldGenerateAddresses = false;
 
--- a/gateway-contracts/tasks/addHostChains.ts
+++ b/gateway-contracts/tasks/addHostChains.ts
@@ -1,7 +1,8 @@
 import dotenv from "dotenv";
-import fs from "fs";
 import { task, types } from "hardhat/config";
+import path from "path";
 
+import { ADDRESSES_DIR } from "../hardhat.config";
 import { getRequiredEnvVar } from "./utils/loadVariables";
 
 // Add host chains metadata to the GatewayConfig contract
@@ -29,13 +30,10 @@ task("task:addHostChainsToGatewayConfig")
       });
     }
 
-    let proxyAddress: string;
     if (taskArgs.useInternalGatewayConfigAddress) {
-      const parsedEnvGatewayConfig = dotenv.parse(fs.readFileSync("addresses/.env.gateway_config"));
-      proxyAddress = parsedEnvGatewayConfig.GATEWAY_CONFIG_ADDRESS;
-    } else {
-      proxyAddress = getRequiredEnvVar("GATEWAY_CONFIG_ADDRESS");
+      dotenv.config({ path: path.join(ADDRESSES_DIR, ".env.gateway") });
     }
+    const proxyAddress = getRequiredEnvVar("GATEWAY_CONFIG_ADDRESS");
 
     // Add host chains
     const gatewayConfig = await hre.ethers.getContractAt("GatewayConfig", proxyAddress, deployer);
--- a/gateway-contracts/tasks/addHostChains.ts
+++ b/gateway-contracts/tasks/addHostChains.ts
@@ -10,7 +10,7 @@ import { getRequiredEnvVar } from "./utils/loadVariables";
 // for local testing. By default, we use the GATEWAY_CONFIG_ADDRESS env var, as done in deployment
 task("task:addHostChainsToGatewayConfig")
   .addParam("useInternalGatewayConfigAddress", "If internal GatewayConfig address should be used", false, types.boolean)
-  .setAction(async function (taskArgs, hre) {
+  .setAction(async function ({ useInternalGatewayConfigAddress }, hre) {
     await hre.run("compile:specific", { contract: "contracts" });
     console.log("Register host chains to GatewayConfig contract");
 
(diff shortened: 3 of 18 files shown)


gateway-contracts/scripts/ensure_proxy_addresses.ts
isaacdecoded, line 1: @RomanBredehoft do you know what this script is for? I couldn't see where we are using it. Maybe by ZWS?

melanciani: it is used in the `ensure-addresses` make command, which is used in several conformance make commands

as stated in the comment above the make command : 
```
# Ensure that the empty proxy addresses exists as these are required for contract compilation.
# In addition, ensure that the addresses match the ones used in local development in order to
# avoid discrepancies between local and CI (e.g., when generating and checking rust bindings).
```

gateway-contracts/tasks/addHostChains.ts
```diff
@@ -29,13 +30,10 @@ task("task:addHostChainsToGatewayConfig")
...
     }
 
-    let proxyAddress: string;
-    if (taskArgs.useInternalGatewayConfigAddress) {
-      const parsedEnvGatewayConfig = dotenv.parse(fs.readFileSync("addresses/.env.gateway_config"));
-      proxyAddress = parsedEnvGatewayConfig.GATEWAY_CONFIG_ADDRESS;
-    } else {
-      proxyAddress = getRequiredEnvVar("GATEWAY_CONFIG_ADDRESS");
+    if (useInternalGatewayConfigAddress) {
+      dotenv.config({ path: path.join(ADDRESSES_DIR, ".env.gateway"), override: true });
```
melanciani, line 34: could you explain how this line works ?  it's overriding the env vars directly ?

if so maybe add a comment explaining that, although I'm not sure if it's a good thing to do (I'm thinking : if we use the option by mistake, it overrides all contract addresses in the env .. seems a bit dangerous, especially compared to before where we would only read the address ... unless it was also loading them in env ?)

gateway-contracts/tasks/deployment/empty_proxies.ts
```diff
@@ -33,39 +32,45 @@ task("task:setContractAddress")
...
     }
 
-    const solidityFilePath = path.join(ADDRESSES_DIR, `${name}Address.sol`);
-    const solidityTemplate =
-      `// SPDX-License-Identifier: BSD-3-Clause-Clear\n\n` +
-      `pragma solidity ^0.8.24;\n\n` +
-      `address constant ${pascalCaseToCamelCase(name)}Address = ${address};\n`;
+    const solidityFilePath = path.join(ADDRESSES_DIR, "GatewayAddresses.sol");
+    let solidityHeader = "";
+    if (!fs.existsSync(solidityFilePath)) {
```
melanciani: I'm not sure I get this part ? should we  simply raise an error if there is no `GatewayAddresses.sol` ? why do we simply avoid having the license header instead ? 🤔

gateway-contracts/tasks/deployment/empty_proxies.ts
```diff
@@ -33,39 +32,45 @@ task("task:setContractAddress")
...
     }
+
+    console.log(`${name} address ${address} written successfully!`);
   });
 
 // Deploy all the EmptyUUPS proxy contracts
 task("task:deployEmptyUUPSProxies").setAction(async function (_, { ethers, upgrades, run }) {
+  const envFilePath = path.join(ADDRESSES_DIR, ".env.gateway");
+  const solidityFilePath = path.join(ADDRESSES_DIR, "GatewayAddresses.sol");
+  fs.rmSync(envFilePath, { recursive: true, force: true });
```
melanciani: why do need to rm the files now  ? would be great to avoid having tasks that remove files, seems a bit dangerous 🤔

melanciani (changes requested):
great, although I think we should avoid a few things here

gateway-contracts/tasks/addHostChains.ts
isaacdecoded: Yes, this is overriding values from the environment, not by removing any env variable but replacing already set values if present in the new `.env` file. This is the exact doc: `Override any environment variables that have already been set on your machine with values from your .env file.`

I'm not sure of it being dangerous, I think in terms of code it's more clean and the behavior should be the same as before: if the user use the option by mistake it will use different addresses anyway, independently if it's overriding or not. 🤔

gateway-contracts/tasks/deployment/empty_proxies.ts
isaacdecoded: I agree that executing `rm` seems kind of alerting, but maybe in this case is not that much since it's very specific in the files being removed. The thing is that with the consolidation of the addresses in a single file, we need to execute `fs.appendFileSync` so if we execute the `deployEmptyUUPSProxies` multiple times, the generated files will contains the content repeatedly. Hence, the "cleaning" of these files on each execution of this task.

gateway-contracts/tasks/deployment/empty_proxies.ts
isaacdecoded: The intention is to check if the `GatewayAddresses.sol` already exists so we should only append the `address` line, otherwise it's being created and we should append the "Solidity header".

gateway-contracts/tasks/deployment/empty_proxies.ts
isaacdecoded: Maybe another approach could be initializing/cleaning these files instead of removing them:
- Leaving empty the `.env.gateway`
- Writing just the solidity header in the `GatewayAddresses.sol`

jatZama: Not sure why you need to do this here, I did not have to remove files on host contrats.

gateway-contracts/tasks/deployment/empty_proxies.ts
jatZama: Same here, in host contracts I assumed the order was always the same, since everything is deployed in correct order in deployAllHostContracts task. I assume you could do the same here.

gateway-contracts/tasks/addHostChains.ts
melanciani: ok sure i guess it's fine then

gateway-contracts/tasks/deployment/empty_proxies.ts
melanciani: hmm but why do we have these 2 scenarios ? unless I'm missing a big problem here I think you should keep it simple : just (re)write the file each time like before

having these kind of checks is not very future proof imo (eg what if we change the license one day ?)

gateway-contracts/tasks/deployment/empty_proxies.ts
melanciani: I'm not sure I get why we start to have these issues, so yes please keep it as close as what have been done in host contracts , unless there is a big problem / under-optimized code and in that case better to also update the host tasks

gateway-contracts/tasks/deployment/empty_proxies.ts
isaacdecoded: Yes, sticking to a fixed order of deployment solve this problem since the first contract will include the solidity header and the rest just the address line. This is how is done in `host-contracts`, but I followed with this other approach trying to avoid having this mentioned deployment order dependency. I'll update the PR to follow same approach than host-contracts 👍

gateway-contracts/tasks/deployment/empty_proxies.ts
isaacdecoded: Yes, will update to follow the solution from host-contracts, i.e. having a fixed order of deployment.

isaacdecoded: Ok, since we have a generic `setContractAddress` task instead of dedicated ones like `host-contracts` this is not possible to follow the same approach, but there's another solution in the [last commit](https://github.com/zama-ai/fhevm/pull/572/commits/3ee729c79d5a2358adbb209e69275722ccff25f4) where the files are initialized at the beginning of the `deployEmptyUUPSProxies` task and the `fs.rmSync` execution is no longer needed/present.

cc @RomanBredehoft @jatZama

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -107,12 +107,14 @@ jobs:
 
       # Executing task:deployEmptyUUPSProxies is required to generate the proxy contract addresses, which are then used
       # for compiling the current and new implementations during the upgrade task.
+      # The docker cp command is required until previous-contracts uses separated .env and addresses files.
```
melanciani: btw, worth creating an issue (and put it in the comment here) for this if we'll change this in the next version

gateway-contracts/tasks/deployment/empty_proxies.ts
```diff
@@ -72,6 +63,19 @@ task("task:deployEmptyUUPSProxies").setAction(async function (_, { ethers, upgra
   const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
   const deployer = new Wallet(deployerPrivateKey).connect(ethers.provider);
 
+  // Ensure the ADDRESSES_DIR exists or create it
+  fs.mkdirSync(ADDRESSES_DIR, { recursive: true });
+
+  const envFilePath = path.join(ADDRESSES_DIR, ".env.gateway");
```
melanciani, line 65: would add a comment here saying why you need to do this

also, it seems host tasks don't do this, so would probably be interesting to change this there as well if possible ! seems a better way to handle this file

melanciani (approved):
a few comments, but unblocking this PR since it looks almost ready to be merged !

.github/workflows/gateway-contracts-upgrade-tests.yml
isaacdecoded: Issue created here https://github.com/zama-ai/fhevm-internal/issues/217.

jatZama (approved):

PR #575, opened 2025-07-15
chore(kms-connector): only trigger test workflow if changes

melanciani:
to avoid running connector tests in all PRs


eudelins-zama (approved):
LGTM!

PR #577, opened 2025-07-15
chore(kms-connector): skip stop-runner if start-runner is skipped in test workflow

melanciani:
else the workflow fails : https://github.com/zama-ai/fhevm/actions/runs/16291747882?pr=211


eudelins-zama (approved):
LGTM

PR #581, opened 2025-07-15
ci(common): bump common-docker to support latest and provenance only …

fegmorte:
…for release


eudelins-zama:
No more relevant I guess

PR #586, opened 2025-07-16
docs(common): txn-manager high-level design

dartdart26:
The Transaction Manager set of libraries and service supports reliable transaction sending to a blockchain. This commit proposes a high-level design.

### DIFF
--- a/transaction-manager/ARCHITECTURE.md
+++ b/transaction-manager/ARCHITECTURE.md
@@ -43,6 +43,17 @@ Since there might be failures in sending, we require persistence of the transact
 
 Sending an async transaction returns an unique identifier to the user that can later be used to query the state of the transaction, cancel it if not already sent, etc.
 
+### transaction-sdk
+
+The transaction-sdk is a high-level library that abstracts most of the details of how sending works under the hood. For example, wallet configuration would happen on the transaction-manager side and users won't be concerned with it. Also, nonce management is transparent to transaction-sdk users.
+
+At a minumum, the user supplies the following pieces of data when sending:
+
+- destination contract address (and potentially chain)
+- calldata
+- optional retry settings
+- optional error handling settings
+
 #### Error Handling
 
 When sending async transactions, some errors returned from the destination smart contract can be treated as non-errors. For example, if an action has already been done, there's no need to retry the transaction anymore. For such cases, the transaction-sdk allows users to specify such errors when sending the transaction.
--- a/transaction-manager/ARCHITECTURE.md
+++ b/transaction-manager/ARCHITECTURE.md
@@ -1,0 +1,70 @@
+# Transaction Manager
+
+The Transaction Manager is a set of services and libraries that reliably send blockchain transactions. It is used by all FHEVM components that send transactions. It consists of three components:
+
+ - `transaction-manager` - a service (binary) that runs in the background, accepts transaction requests and sends them to the blockchain
+ - `transaction-sdk` - a Rust library to interact with the transaction-manager
+ - `fhevm-contracts-sdk` - a convenience Rust library to create calldata for calls to FHEVM contracts
+
+## Architecture
+
+```mermaid
+graph LR;
+    user[user]
+    tsdk(transaction-sdk)
+    csdk(fhevm-contracts-sdk)
+    mng(transaction-manager)
+    db[(DB)]
+    bc(((Blockchain)))
+
+    user--- |txn request/response| tsdk
+    user--- |calldata| csdk
+    mng ---|queue, poll| db
+    tsdk ---|send_sync over RPC| mng
+    tsdk ---|send_async| db
+    mng --- bc
+```
+
+Transaction Manager supports two types of trasaction sending - `sync` and `async`.
+
+### Sync Transaction Sending
+
+Some use cases require that a transaction is sent and the outcome is known immediately. For such cases, sync sending works by the transaction-sdk sending a RPC call to the transaction-manager and waiting for the response. On the transaction-manager side, gas estimation will be done and the transacttion will be broadcast to the blockchain. Once the receipt is received, the response will be returned over RPC to the user.
+
+Sync sending doesn't require persistence and, therefore, no interaction with the DB is needed.
+
+In terms of retries, the user might optionally specify a retry logic when sending the transaction. Users might also select how many confirmations to wait for.
+
+### Async Transaction Sending
+
+Other use cases require that a transaction is eventually sent or, optionally, is eventually sent with guarantees. For such cases, async sending works by the transaction-sdk inserting the transaction to the DB and notifying the transaction-manager. On the transaction-manager side, the transaction will be fetched and sent to the blockchain.
+
+Since there might be failures in sending, we require persistence of the transaction in the DB, such that retries can be attempted. That also includes cases of the transaction-manager crashing or being stopped.
+
+Sending an async transaction returns an unique identifier to the user that can later be used to query the state of the transaction, cancel it if not already sent, etc.
+
+#### Error Handling
+
+When sending async transactions, some errors returned from the destination smart contract can be treated as non-errors. For example, if an action has already been done, there's no need to retry the transaction anymore. For such cases, the transaction-sdk allows users to specify such errors when sending the transaction.
+
+## Data Format
+
+The transaction-sdk and transaction-manager **are not aware of the contents** of the transactions being sent. Instead, they only care about the destination chain + contract and the calldata.
+
+The fhevm-contracts-sdk can be used to create FHEVM-specific contract calldata.
+
+Such distinction makes the transaction-manager and transaction-sdk independent of the use case. Furthermore, they don't have to be restarted or updated in case of contract changes.
+
+## Reorg/Fork Handling for Async Transactions
+
+Users require that some transaction types be included in the blockchain with certainty. For these, the transaction-manager needs to be aware of reorgs/forks on the blockchain. If one is detected, transactions that could possibly be affected need to be replayed.
+
+## Secret Key and Nonce Management
+
+The transaction-manager will maintain a set of keys used for sending transactions. The keys could be kept in plaintext files or AWS KMS. Other methods can be added in the future.
+
+In terms of nonce management, the transaction-manager will send transactions in parallel from different keys (wallets) and will also cache nonces locally. If a nonce mismatch is found, it will fetch a fresh nonce from the blockchain and go from there.
+
+## Manual Resend for Async Transactions by Operations
+
+Since data is persisted to the DB when sending async transactions, operations teams can easily replay any set of transactions that have already been sent. That can be triggered through the DB.
--- a/transaction-manager/ARCHITECTURE.md
+++ b/transaction-manager/ARCHITECTURE.md
@@ -25,7 +25,7 @@ graph LR;
     mng --- bc
 ```
 
-Transaction Manager supports two types of trasaction sending - `sync` and `async`.
+Transaction Manager supports two types of transaction sending - `sync` and `async`.
 
 ### Sync Transaction Sending
 
--- a/transaction-manager/ARCHITECTURE.md
+++ b/transaction-manager/ARCHITECTURE.md
@@ -29,7 +29,7 @@ Transaction Manager supports two types of transaction sending - `sync` and `asyn
 
 ### Sync Transaction Sending
 
-Some use cases require that a transaction is sent and the outcome is known immediately. For such cases, sync sending works by the transaction-sdk sending a RPC call to the transaction-manager and waiting for the response. On the transaction-manager side, gas estimation will be done and the transacttion will be broadcast to the blockchain. Once the receipt is received, the response will be returned over RPC to the user.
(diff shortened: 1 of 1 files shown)


transaction-manager/ARCHITECTURE.md
```diff
@@ -0,0 +1,70 @@
...
+
+Transaction Manager supports two types of trasaction sending - `sync` and `async`.
+
+### Sync Transaction Sending
+
+Some use cases require that a transaction is sent and the outcome is known immediately. For such cases, sync sending works by the transaction-sdk sending a RPC call to the transaction-manager and waiting for the response. On the transaction-manager side, gas estimation will be done and the transacttion will be broadcast to the blockchain. Once the receipt is received, the response will be returned over RPC to the user.
+
+Sync sending doesn't require persistence and, therefore, no interaction with the DB is needed.
+
+In terms of retries, the user might optionally specify a retry logic when sending the transaction. Users might also select how many confirmations to wait for.
```
leventdem, line 36: Here retry implies also timeout not only failure when sending tx ? 
We are sending a transaction and agree to wait specific amount of time then retry.

dartdart26: Yeah, everything until it succeeds or fails with a definitive error. Could be network issue, timeout, etc.

leventdem:
Suppose we prioritize reliability and success of transaction over performance. 
Is this scenario relevant for this design: 

1. I create or already have a poll of 50 wallets
2. I decided to only send sync transaction using the 50 wallets
3. If I receive more transactions, I just wait until a wallet is available (meaning received required number of confirmation)

dartdart26:
> Suppose we prioritize reliability and success of transaction over performance. Is this scenario relevant for this design:
> 
> 1. I create or already have a poll of 50 wallets
> 2. I decided to only send sync transaction using the 50 wallets
> 3. If I receive more transactions, I just wait until a wallet is available (meaning received required number of confirmation)

My thinking is that the user doesn't know about wallets. If you send more txns than wallets, we would reuse wallets. If something fails, we either retry or return early to the user so that they can decide what to do (depending on whether they requested retries).

Let me add info on the `transaction-sdk` to be clear that it is high-level and doesn't require users to care about wallets.

eudelins-zama:
Regarding the `fhevm-contracts-sdk`, how is it different than the already existing rust bindings of the Gateway contract?

Btw, do we have bindings for the host contracts ? In that case, `fhevm-contracts-sdk` = `gateway_bindings` + `host_bindings` ?

eudelins-zama:
Is this architecture compatible with the draft architecture we imagine for the future communication between kms-core <-> kms-connector? (see https://github.com/zama-ai/kms-core/issues/2654)

It seems that the kms-core would need to directly use the transaction-manager with this architecture, so maybe we should ping them in that case

dartdart26:
> Regarding the `fhevm-contracts-sdk`, how is it different than the already existing rust bindings of the Gateway contract?
> 
> Btw, do we have bindings for the host contracts ? In that case, `fhevm-contracts-sdk` = `gateway_bindings` + `host_bindings` ?

The intention is that `rust-sdk` can be seen as user-facing, whereas this one is for internal protocol stuff. I discussed with @immortal-tofu and we decided to keep these outside of `rust-sdk`. I am open to more discussion on that, but yeah, the equation you give is what we had in mind.

dartdart26:
> Is this architecture compatible with the draft architecture we imagine for the future communication between kms-core <-> kms-connector? (see [zama-ai/kms-core#2654](https://github.com/zama-ai/kms-core/issues/2654))
> 
> It seems that the kms-core would need to directly use the transaction-manager with this architecture, so maybe we should ping them in that case

Good point! I think we should, yes. cc @maksymsur @dd23

dd23:
General question: 
There's probably a good reason for it, but why is it just a tx sender (and does not have receiving capabilities)? Just curious to hear the motivation for this.

>It seems that the kms-core would need to directly use the transaction-manager with this architecture, so maybe we should ping them in that case

Re integration with KMS core:
The perspective from kms-core side so far was that we would offload all blockchain interfaces (sending and receiving) to the kms-connector. That's why this component was created in the first place.
That doesn't mean it has to stay like this forever, but having receiving from BC done by the connector (as it is currently) and moving sending responses to BC to the cores as proposed here, seems a bit odd to me, tbh.

eudelins-zama:
> Re integration with KMS core:
The perspective from kms-core side so far was that we would offload all blockchain interfaces (sending and receiving) to the kms-connector. That's why this component was created in the first place.
That doesn't mean it has to stay like this forever, but having receiving from BC done by the connector (as it is currently) and moving sending responses to BC to the cores as proposed here, seems a bit odd to me, tbh.

I also agree it feels odd...

maksymsur:
I like the advanced logic for managing transactions - it's really needed, but I have several questions / concerns:
- How exactly will client services interact with the TrxManager—via gRPC or direct DB access is also an option? If the envisioned workflow is `Connector → Core → TrxManager` (scenario-1), does this imply the Connector’s role would be limited primarily to a parser of blockchain events and forwarding prepared decryption requests to the Core? This clear separation of concerns might indeed be effective. On the other hand, if the intended flow is `Connector → Core → Connector → TrxManager` (scenario-2), I’m concerned about potential latency introduced by additional message passing + flooding the networking. If we follow scenario-1, then we need to know if the TrxManager could actively query the Core for results? ... which is leading us to the next question..
- Is TrxManager following Passive or optionally Active model for queueing transactions? Should other components always explicitly trigger TrxManager, or can the TrxManager deterministically query components to gather transaction requests? For example, once the Core processes a request and generates a response to be retrieved (a responsibility currently managed by the Connector), could the TrxManager autonomously retrieve these results? (perhaps we may think if it's possible to send light notification about `request-xxx` is ready for retrieval.. but really doubt about its viability). However, I assume that adding such "active" logic might overly complicate the TrxManager, pushing us back towards the `Connector → Core → Connector → TrxManager` flow, which again raises latency concerns. Are we prepared to accept this latency trade-off, or might there be alternative solutions worth exploring? Given the Core’s critical status and associated security guarantees, I doubt that increasing its network interaction or exposure is a good idea. CC @mkmks 
- On the other hand, all our components sending transactions need to have this unified extended logic for managing transactions (their resending, validation of acceptance given finality rules, etc).

transaction-manager/ARCHITECTURE.md
manoranjith: i guess we include Gateway being "paused" as a definitive error ?

manoranjith (commented):
@dartdart26 I was wondering if we need a mechanism to cancel a transaction ?

- **Context 1**: Relayer (for oracle public decrypt) or Copro (for ACL) triggered a transaction to be sent on Gateway. But noticed a reorg on the host chain and hence would like to cancel the transaction.

- **Context 2**: A reorg on Gateway chain leading to relayer wanting to cancel a transaction on host chain.

- Maybe we don't need cancellation:
  - Because the time to detect reorg can be much longer than time to send a transaction in sync/async mode.
  - In this case, we could make an explicit design choice to not support. 

- Maybe we need cancellation:
  - Some errors take longer time to resolve. 
  - Eg: Node handling transactions is down for minutes/hours.

manoranjith:
A question regarding the DB, is the motivation to use it as both purposes: persistence and communication channel (for async requests) ?

When we say eventually, does it mean the caller  not concerned about the result of the transaction ?

manoranjith:
a point regarding error handling. A special case could be "gateway paused". 

At this time, all transaction would fail (or no?).


Should we handle these cases at 
1. service level - monitor the gateway and return error for all transactions without trying
2. transaction level - let each transaction be tried and fail.

maksymsur:
The cancellation scenario and Arbitrum’s extended finality rules are indeed critical. Consider a **high-value transaction** that is executed, communicated to the user application as successful, and subsequently reverted after, say, a day. If the affected party initiates an audit, our processes wouldn’t pass scrutiny if this case were disregarded. Therefore, I believe it’s essential to clearly define what we consider "final" and precisely outline the boundaries of our legal responsibility.

dartdart26:
> General question: There's probably a good reason for it, but why is it just a tx sender (and does not have receiving capabilities)? Just curious to hear the motivation for this.
> 
> Re integration with KMS core: The perspective from kms-core side so far was that we would offload all blockchain interfaces (sending and receiving) to the kms-connector. That's why this component was created in the first place. That doesn't mean it has to stay like this forever, but having receiving from BC done by the connector (as it is currently) and moving sending responses to BC to the cores as proposed here, seems a bit odd to me, tbh.

We will have a separate listener implementation. Is there a reason to bundle these together? Or maybe I miss something?

Re kms-connector - I am not proposing that the kms-core does the sending. This PR is a proposal for a common way of sending transactions, but the sender can be anything, i.e. kms-connector. I meant that we should sync on how the kms-connector is implemented, but not that it should be kms-core that does the sending @eudelins-zama @dd23 .

dartdart26:
> I like the advanced logic for managing transactions - it's really needed, but I have several questions / concerns:
> 
> * How exactly will client services interact with the TrxManager—via gRPC or direct DB access is also an option? If the envisioned workflow is `Connector → Core → TrxManager` (scenario-1), does this imply the Connector’s role would be limited primarily to a parser of blockchain events and forwarding prepared decryption requests to the Core? This clear separation of concerns might indeed be effective. On the other hand, if the intended flow is `Connector → Core → Connector → TrxManager` (scenario-2), I’m concerned about potential latency introduced by additional message passing + flooding the networking. If we follow scenario-1, then we need to know if the TrxManager could actively query the Core for results? ... which is leading us to the next question..
> * Is TrxManager following Passive or optionally Active model for queueing transactions? Should other components always explicitly trigger TrxManager, or can the TrxManager deterministically query components to gather transaction requests? For example, once the Core processes a request and generates a response to be retrieved (a responsibility currently managed by the Connector), could the TrxManager autonomously retrieve these results? (perhaps we may think if it's possible to send light notification about `request-xxx` is ready for retrieval.. but really doubt about its viability). However, I assume that adding such "active" logic might overly complicate the TrxManager, pushing us back towards the `Connector → Core → Connector → TrxManager` flow, which again raises latency concerns. Are we prepared to accept this latency trade-off, or might there be alternative solutions worth exploring? Given the Core’s critical status and associated security guarantees, I doubt that increasing its network interaction or exposure is a good idea. CC @mkmks
> * On the other hand, all our components sending transactions need to have this unified extended logic for managing transactions (their resending, validation of acceptance given finality rules, etc).

1. The way it interacts with kms-connector/core is maybe something we can discuss.
2. Currently, my proposal is for the txn-manager to only send transactions that have been sent to it over the SDK (either RPC or inserted into the DB).
3. Yeah, my intention was to have a simple service that can be used by other components and that is as dummy as possible. If we feel this is not good for the kms-connector, let's discuss. What would the best be for the kms-connector?

dartdart26:
> @dartdart26 I was wondering if we need a mechanism to cancel a transaction ?
> 
> * **Context 1**: Relayer (for oracle public decrypt) or Copro (for ACL) triggered a transaction to be sent on Gateway. But noticed a reorg on the host chain and hence would like to cancel the transaction.
> * **Context 2**: A reorg on Gateway chain leading to relayer wanting to cancel a transaction on host chain.
> * Maybe we don't need cancellation:
>   
>   * Because the time to detect reorg can be much longer than time to send a transaction in sync/async mode.
>   * In this case, we could make an explicit design choice to not support.
> * Maybe we need cancellation:
>   
>   * Some errors take longer time to resolve.
>   * Eg: Node handling transactions is down for minutes/hours.

Yeah, one of the reasons I suggest that send_async() returns an unique handle is for that - to be able to cancel a txn later (if not already added to the chain).

dartdart26:
> A question regarding the DB, is the motivation to use it as both purposes: persistence and communication channel (for async requests) ?
> 
> When we say eventually, does it mean the caller not concerned about the result of the transaction ?

Main motivation is to use it for persistence.

Re eventually, the caller might not be concerned about the result. But they can still get the result via the unique handle they get for the txn when they send it via send_async().

dartdart26:
> a point regarding error handling. A special case could be "gateway paused".
> 
> At this time, all transaction would fail (or no?).
> 
> Should we handle these cases at
> 
> 1. service level - monitor the gateway and return error for all transactions without trying
> 2. transaction level - let each transaction be tried and fail.

If we consider the txn-manager as dummy and generic (e.g. also sending to fhevm chains for public decrypt callback), I'd say it has to be somehow configurable by the user as this feature of pausing is specific to the GW.

dartdart26:
> The cancellation scenario and Arbitrum’s extended finality rules are indeed critical. Consider a **high-value transaction** that is executed, communicated to the user application as successful, and subsequently reverted after, say, a day. If the affected party initiates an audit, our processes wouldn’t pass scrutiny if this case were disregarded. Therefore, I believe it’s essential to clearly define what we consider "final" and precisely outline the boundaries of our legal responsibility.

Yes, I agree, it needs to be up to the user to decide what they want for the txn. One simple solution would be to manually replay potentially affected txns after a reorg - would be easy since we have persistence.

We discussed that for the GW in particular and since a reorg there is very unlikely and close to impossible (as mentioned by Conduit), we might rely on manual.

For fhevm chains, that's a bit more involved and we need to think about it more.

transaction-manager/ARCHITECTURE.md
NekiMax: i`ll see

poppyseedDev:
Are you still working on this PR @dartdart26 ?

dartdart26:
> Are you still working on this PR @dartdart26 ?

It is on hold for now, but we might eventually work on it. Not sure which quarter.

dartdart26:
Closing as decisions re above are pending.

PR #590, opened 2025-07-16
feat(sdk): upgrade kms core dependencies and update user key pair algorithm accordingly

leventdem:
This PR introduces an upgrade of the user decryption key pair generation algorithm. 

Initially cryptobox was used, but now ml-kem is used to enable post-quantum security.

The API changes with the new kms crate version:
- KMS signers are given with the corresponding IDs
- Direct call to client is done instead of calling a function and passing a mutable client

Others changes:
- keys are longer
- use separate json file to test ( a test generated on kms core has been used to verify behavior)


ppoliani (approved):

ppoliani:
@leventdem [this](https://github.com/zama-ai/fhevm/blob/b7b2485161993cb7cd16d824d97d6d901879da3a/sdk/rust-sdk/src/decryption/user/response.rs#L447) test was failing when I applied a similar change to the.

The issue is that we're using the j`s_api` from kms-core which targets wasm. Currently when a JsError is created we get a panic.

> running 1 test
panicked at /wasm-bindgen-0.2.100/src/lib.rs:1062:1:
function not implemented on non-wasm32 targets
panicked at library/core/src/panicking.rs:218:5:

Should we try avoid using the wasm interface within the `gateway-sdk`?

leventdem:
>  Should we try avoid using the wasm interface within the gateway-sdk?

For now, we have to use some wasm compatible method, because initially the crate was implemented for wasm. 
A more full-rust version will be available in kms.

PR #593, opened 2025-07-17
feat(coprocessor):  Enable compression for big ciphertexts

goshawk-3:
fixes: https://github.com/zama-ai/fhevm-internal/issues/235


antoniupop (approved):

coprocessor/fhevm-engine/sns-executor/src/executor.rs
```diff
@@ -560,3 +562,26 @@ fn decompress_ct(
...
+    {
+        if let Some(client_key) = &keys.client_key {
+            let ct = if is_compressed {
+                ct.decrypt_squash_noise_compressed(client_key, squashed_noise_serialized)
+            } else {
+                ct.decrypt_squash_noise(client_key, squashed_noise_serialized)
+            }
+            .expect("Failed to decrypt");
+
+            info!(target: "sns", "Decrypted plaintext: {:?}, handle: {}", ct, compact_hex(handle));
```
kc1212, line 584: I don't really know coprocessor code so maybe I missed it, but do you assert the plaintext against the true plaintext anywhere?

(also, `ct` in this scope should probably be called `pt` as it's a plaintext).

goshawk-3: There is a unit test that does assert the plaintext.

 This snippet is just for debugging purposes - tracing plaintext values of the cts in a local e2e testing.

rudy-6-4 (approved):

coprocessor/fhevm-engine/sns-executor/src/bin/utils/daemon_cli.rs
```diff
@@ -105,6 +105,10 @@ pub struct Args {
     /// if false, default FIFO (First In, First Out) processing is used
     #[arg(long, default_value_t = false)]
     pub lifo: bool,
+
+    /// Enable compression of big ciphertexts before uploading to S3
+    #[arg(long, default_value_t = false)]
```
dartdart26: Wondering whether the default shouldn't be true?

dartdart26 (approved):
Looks good!

coprocessor/fhevm-engine/sns-executor/src/bin/utils/daemon_cli.rs
eudelins-zama: Fixed in 4b120061f96f3721319f5b2c73c40df9f2bd9002

PR #594, opened 2025-07-17
feat(gateway-contracts): introduce the generic bytes metadata

isaacdecoded:
Closes [#209](https://github.com/zama-ai/fhevm-internal/issues/209)


isaacdecoded:
At this moment, this draft PR introduces the `extraData` bytes field only in the `InputVerification` and `Decryption` because these are the contracts that will have breaking changes with the implementation of context IDs. Do you think we should add this `extraData` field in the rest of the contracts for all functions, events, EIP712 signatures at this time? It seems a bit YAGNI but just to double-check that we want to introduce it at once in this PR. Same for host-`contracts`?

cc @RomanBredehoft @jatZama

gateway-contracts/contracts/libraries/HandleOps.sol
```diff
@@ -46,4 +49,36 @@ library HandleOps {
...
+        }
+
+        // The first byte of extraData is the version
+        version = uint8(extraData[0]);
+
+        // Allocate a new bytes array with everything after the first byte (version byte)
+        uint256 payloadLength = extraData.length - 1;
+        payload = new bytes(payloadLength);
+
+        assembly {
```
isaacdecoded: @jatZama I'd like you to check this Yul snippet here, please. Is this a good approach for separating the version and the rest of the payload from the `extraData` bytes, or is there a better solution for this? 🙏

jatZama:
> At this moment, this draft PR introduces the `extraData` bytes field only in the `InputVerification` and `Decryption` because these are the contracts that will have breaking changes with the implementation of context IDs. Do you think we should add this `extraData` field in the rest of the contracts for all functions, events, EIP712 signatures at this time? It seems a bit YAGNI but just to double-check that we want to introduce it at once in this PR. Same for host-`contracts`?
> 
> cc @RomanBredehoft @jatZama

I dont know the answer to this.

Can you remind me why would it be needed for Decryption and InputVerification already? why would it be needed elsewhere? which other contracts?

gateway-contracts/contracts/Decryption.sol
```diff
@@ -885,4 +907,24 @@ contract Decryption is
...
     }
+
+    /**
+     * @dev Returns the extra data to be used in the Public Decryption request.
+     * This is used to ensure that the request is versioned and can be extended in the future.
+     * The first byte is the version, and the rest can be used for additional data.
+     * @return The extra data for the Public Decryption request.
+     */
+    function _getPublicDecryptionRequestExtraData() internal pure returns (bytes memory) {
+        return abi.encodePacked(uint8(1));
```
jatZama: why 1 and not 0?

gateway-contracts/contracts/Decryption.sol
```diff
@@ -885,4 +907,24 @@ contract Decryption is
...
+    }
+
+    /**
+     * @dev Returns the extra data to be used in the User Decryption request.
+     * This is used to ensure that the request is versioned and can be extended in the future.
+     * The first byte is the version, and the rest can be used for additional data.
+     * @return The extra data for the User Decryption request.
+     */
+    function _getUserDecryptionRequestExtraData() internal pure returns (bytes memory) {
+        return abi.encodePacked(uint8(1));
```
jatZama: same question here

gateway-contracts/contracts/libraries/HandleOps.sol
jatZama: this `parseExtraData` library function does not seem to be used anywhere yet , right?

isaacdecoded:
> Can you remind me why would it be needed for Decryption and InputVerification already? why would it be needed elsewhere? which other contracts?

For example, there's an ongoing change that adds a new `contextId` field in the InputVerification contract for the `CiphertextVerification` EIP712 signature [here in this PR](https://github.com/zama-ai/fhevm/pull/211/files#diff-b9560385508b39fa02a1d6a41cba7948e7602970fc9824b944049dc587ee6ea0R62) (line 62). This change means a breaking change for our coprocessors, but with the generic `extraData` field, we can prevent them from failing.

In the case of Decryption, it should be the same when we move forward with the KMS contexts as well, afaiu.

The thing is, these two contracts are the ones that, as I understand, will introduce breaking changes in the short term, but not sure about the rest of the contracts. The intention of the issue is to introduce this `extraData` field, although it doesn't specify on which functions/events/etc we should start with.

gateway-contracts/contracts/libraries/HandleOps.sol
jatZama: did you use Yul for gas optimization? this is not really needed on gateway AFAIK
you were not obliged to use yul
i would rewrite it with regular loop for readability

isaacdecoded: That's right, it's because we don't have any specific logic for any version yet. But the idea is to use it at some point like this:
```
(uint8 version, bytes payload) = parseExtraData(extraData)
if (version == 1) {
     */ logic over the payload, e.g. abi.encode(payload, [address, uint256, etc]) */
}
```

jatZama:
> > Can you remind me why would it be needed for Decryption and InputVerification already? why would it be needed elsewhere? which other contracts?
> 
> For example, there's an ongoing change that adds a new `contextId` field in the InputVerification contract for the `CiphertextVerification` EIP712 signature [here in this PR](https://github.com/zama-ai/fhevm/pull/211/files#diff-b9560385508b39fa02a1d6a41cba7948e7602970fc9824b944049dc587ee6ea0R62) (line 62). This change means a breaking change for our coprocessors, but with the generic `extraData` field, we can prevent them from failing.
> 
> In the case of Decryption, it should be the same when we move forward with the KMS contexts as well, afaiu.
> 
> The thing is, these two contracts are the ones that, as I understand, will introduce breaking changes in the short term, but not sure about the rest of the contracts. The intention of the issue is to introduce this `extraData` field, although it doesn't specify on which functions/events/etc we should start with.

Two things: i thought contexts was not going to be part of extraData, but was really going to be part as a fully defined field as an array of addresses, was it decided to be not the case finally? why? is the way we are defining contexts going to be changed very probably again?
I thought the extraData was something else, to be future proof for OTHER new future fields we would like to add, but after the contexts one, is this not the case?

gateway-contracts/contracts/Decryption.sol
isaacdecoded: No strong opinion, let's start with 0 then 👍

gateway-contracts/contracts/libraries/HandleOps.sol
isaacdecoded: > did you use Yul for gas optimization? this is not really needed on gateway AFAIK you were not obliged to use yul i would rewrite it with regular loop for readability

Ok, I see 👍 I'll change to regular loop then

isaacdecoded:
> Two things: i thought contexts was not going to be part of extraData, but was really going to be part as a fully defined field as an array of addresses, was it decided to be not the case finally? why? is the way we are defining contexts going to be changed very probably again? I thought the extraData was something else, to be future proof for OTHER new future fields we would like to add, but after the contexts one, is this not the case?

According to [this discussion in Slack](https://zama-ai.slack.com/archives/C065F16D5U3/p1752047526280779), it seems to me that the change from the Coprocessor contexts is a use case for this `extraData` generic field, at least particularly referring to the new `contextId` in the EIP712 signature and the `VerifyProofRequest ` event.

But maybe worth it to align with @RomanBredehoft because maybe I'm missing something 🤔

jatZama:
> > Two things: i thought contexts was not going to be part of extraData, but was really going to be part as a fully defined field as an array of addresses, was it decided to be not the case finally? why? is the way we are defining contexts going to be changed very probably again? I thought the extraData was something else, to be future proof for OTHER new future fields we would like to add, but after the contexts one, is this not the case?
> 
> According to [this discussion in Slack](https://zama-ai.slack.com/archives/C065F16D5U3/p1752047526280779), it seems to me that the change from the Coprocessor contexts is a use case for this `extraData` generic field, at least particularly referring to the new `contextId` in the EIP712 signature and the `VerifyProofRequest ` event.
> 
> But maybe worth it to align with @RomanBredehoft because maybe I'm missing something 🤔

So did we decide to use this new extraField already for contexts? if this was decided then its ok for me, i just wanted to clarify correct order. I thought contexts was going to be added as a separate field before the extraBytes field, but apparently it is the oposite, so we do add extraBytes first, then we add contexts inside of it, it seems, right?

host-contracts/contracts/InputVerifier.sol
```diff
@@ -238,6 +241,7 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
             ctVerif.ctHandles = listHandles;
             ctVerif.userAddress = context.userAddress;
             ctVerif.contractAddress = context.contractAddress;
+            ctVerif.extraData = extraData;
```
jatZama: ctVerif.extraData = hex'00';

jatZama:
Split this PR in 3 PRs:
Keep 1 PR just for gateway
Then for host do two PRs:
One for InputVerifier: this is the easiest one because function signature of verifyCiphertext is not needed to be modified (since the context extraData will be, long term, incuded inside the inputProof bytes array). And library FHE.sol not needed to be changed.
The 3rd PR will be for KMSVerifier (meaning decryption and mpc context) : this is the more complex part. Probably we will need at least to change how callback in dapps for decryption are handled. But maybe its possible not to change callback, for eg if we are sure that at most, only two contexts maximum will be valid, maybe we can branch on the two valid contexts. To be discussed. But to be as future proof as possible, I think the most flexible solution, unfortunately, will involve changing the callback in dapps, by adding some extraData bytes field even there (to avoid branching undefinitely if we have a lot of upgrades for instance).

isaacdecoded:
> Split this PR in 3 PRs

Opened the two other draft PRs here:
1. [InputVerifier](https://github.com/zama-ai/fhevm/pull/599)
2. [KMSVerifier](https://github.com/zama-ai/fhevm/pull/598)

gateway-contracts/contracts/Decryption.sol
```diff
@@ -138,7 +142,7 @@ contract Decryption is
 
     /// @notice The definition of the UserDecryptResponseVerification structure typed data.
     string private constant EIP712_USER_DECRYPT_RESPONSE_TYPE =
-        "UserDecryptResponseVerification(bytes publicKey,bytes32[] ctHandles,bytes userDecryptedShare)";
+        "UserDecryptResponseVerification(bytes publicKey,bytes32[] ctHandles,bytes userDecryptedShare,bytes extraData)";
```
isaacdecoded, line 151: Do you think we need to add the `extraData` field to the `UserDecryptRequestVerification` as well? I mean, if there's a breaking change in the user decryption request signature, for example, do we want to continue accepting the old version format or fail-fast so the user has to upgrade to the new format?

cc @RomanBredehoft

melanciani:
> At this moment, this draft PR introduces the extraData bytes field only in the InputVerification and Decryption because these are the contracts that will have breaking changes with the implementation of context IDs. Do you think we should add this extraData field in the rest of the contracts for all functions, events, EIP712 signatures at this time? It seems a bit YAGNI but just to double-check that we want to introduce it at once in this PR. Same for host-contracts?

@isaacdecoded the issue is about anticipating any further breaking changes , especially for mainnet, so that we avoid having to update the ABI if we introduce new fields. This will be case for context feature 

so I would say : all (external) functions, events and signatures that are called, listened or used by off-chain components (coprocessor components, host contracts, relayer, kms connector / core). Only exception is the KmsManagement contract imo, since we don't use it yet and it'll be severely updated

one thing I'm still unsure: should we also consider view functions ? functions like `checkPublicDecryptionReady` are also called by off-chain components  . wdyt @jatZama ?

this also means other off-chain components will need to be updated as well, and we'll need to update staging (and then testnet) using the pausing mechanism

gateway-contracts/contracts/Decryption.sol
melanciani: I think we should probably yes

gateway-contracts/contracts/Decryption.sol
```diff
@@ -885,4 +907,26 @@ contract Decryption is
...
+
+    /**
+     * @dev Returns the extra data to be used in the Public Decryption request.
+     * This is used to ensure that the request is versioned and can be extended in the future.
+     * The first byte is the version, and the rest can be used for additional data.
+     * @return The extra data for the Public Decryption request.
+     */
+    function _getPublicDecryptionRequestExtraData() internal pure returns (bytes memory) {
+        // Current version is 0, and no additional data is used.
+        return abi.encodePacked(uint8(0));
```
melanciani: imo the version should be a constant, common to all extra data ... and maybe the `MAJOR_VERSION` number itself actually if we make the contract version follow semver standards (ie, a breaking change = bump the major version) 

wdyt @jatZama ?

jatZama: I am not sure to understand the question. For now, this value of 0x00 is just a temporary placeholder, it will be replaced by whatever will be needed to represent a context, for instance a bytes32 (casted to a bytes of course) being the hash([ the array of mpc nodes + threshold ] ).
I am not sure why you would need to inject the MAJOR_VERSION here.

melanciani: don't we need the first byte of the extra data being a representation of the contract's version ? else I don't get how the internal (and off-chain) logic will be able to differentiate the contract versions in order to decide which logic to apply 🤔

jatZama: I am still not sure it would be a very wise idea to couple the MAJOR of the contract with the extraData version. Because if we do this, it would imply a strong coupling between MAJOR version of gateway contracts with the host contracts, which adds friction. It is more flexible to keep the extraData version unrelated to the contract version. Wdyt?

jatZama: So we would have maybe extraData= extraDataVesionByte || hash([ the array of mpc nodes + threshold ] )
But extraDataVesionByte is not necessarily equal to MAJOR_VERSION.

jatZama:
> one thing I'm still unsure: should we also consider view functions ? functions like `checkPublicDecryptionReady` are also called by off-chain components . wdyt @jatZama ?

Can you clarify what you mean by `one thing I'm still unsure: should we also consider view functions ? functions like checkPublicDecryptionReady are also called by off-chain components . wdyt @jatZama ?` ? It doesn't seem something need to be changed here?
Can you give a concrete example on why this function could need additional arguments?

melanciani:
> Can you give a concrete example on why this function could need additional arguments?

@jatZama well, isn't tje whole point of this PR to anticipate any potential breaking changes in the future ? not just contexts, just in general. So I think we should be able to anticipate any changes anywhere no ? If not, why ? and should we restrict this PR only to external/public non-view functions ?

gateway-contracts/contracts/Decryption.sol
melanciani: ok fine for me then, but then should this "version" number still be a constant state var, common to all functions ? or should it be decoupled for all ? 

here I'm trying to think about how to make off-chain components or host contracts' life easier once they'll need to handle multiple versions

jatZama:
> > Can you give a concrete example on why this function could need additional arguments?
> 
> @jatZama well, isn't tje whole point of this PR to anticipate any potential breaking changes in the future ? not just contexts, just in general. So I think we should be able to anticipate any changes anywhere no ? If not, why ? and should we restrict this PR only to external/public non-view functions ?

But to me it seems more disruptive to add an additional field in the view function if we are almost sure it would never be needed, so this is why i am asking for a concrete example to understand if it makes any sense to add one.

isaacdecoded:
The PR is now updated to introduce the `extraData` in more methods, events and EIP712 signatures according to the following list of external components interaction:

```
KMS Connector:
  Listens to events:
    - PublicDecryptionRequest
    - UserDecryptionRequest
    - PreprocessKeygenRequest
    - PreprocessKskgenRequest
    - KeygenRequest
    - KskgenRequest
    - CrsgenRequest

  Calls to methods:
    - publicDecryptionResponse
      - EIP: PublicDecryptVerification
    - userDecryptionResponse
      - EIP: UserDecryptResponseVerification

Coprocessor:
  Listens to events:
    - VerifyProofRequest

  Calls to methods:
    - allowPublicDecrypt
    - allowAccount
    - verifyProofResponse
      - EIP: CiphertextVerification
    - rejectProofResponse

Relayer:
  Listens to events:
    - UserDecryptionRequest
    - UserDecryptionResponse
    - PublicDecryptionRequest
    - PublicDecryptionResponse

  Calls to methods:
    - verifyProofRequest
    - publicDecryptionRequest
    - userDecryptionRequest
      - EIP: UserDecryptRequestVerification
    - checkPublicDecryptionReady
    - checkUserDecryptionReady
```

cc @RomanBredehoft @jatZama

gateway-contracts/contracts/Decryption.sol
```diff
@@ -492,12 +521,17 @@ contract Decryption is
...
+                $.userDecryptedShares[decryptionId],
+                verifiedSignatures,
+                extraData
+            );
         }
     }
 
     /// @dev See {IDecryption-checkPublicDecryptionReady}.
-    function checkPublicDecryptionReady(bytes32[] calldata ctHandles) external view virtual {
+    function checkPublicDecryptionReady(bytes32[] calldata ctHandles, bytes calldata extraData) external view virtual {
```
jatZama: I am sorry but this seems not good since you decided to add extraData field in the view function in Decryption, but this argument is not used anywhere, i guess it gives you warning at compilation? we will surely get some warnings and questions by ZWS team at deployment here. At least please silence the warning by adding /* */ around extraData in function declaration here.

isaacdecoded: Warnings silenced in [this commit](https://github.com/zama-ai/fhevm/pull/594/commits/db46e019c1213c641324b5d5eea81aab7338fb56), thanks! ✊

gateway-contracts/contracts/mocks/InputVerificationMock.sol
```diff
@@ -48,7 +48,7 @@ contract InputVerificationMock {
         emit VerifyProofResponse(zkProofId, ctHandles, signatures);
     }
 
-    function rejectProofResponse(uint256 zkProofId, bytes calldata extraData) external {
+    function rejectProofResponse(uint256 zkProofId, bytes calldata) external {
```
jatZama: I prefer you comment the argument name as suggested yesterday, instead of not putting any name.
Meaning `function rejectProofResponse(uint256 zkProofId, bytes calldata /* extraData*/) `
For the sake of readability.

isaacdecoded: Ah yes, but these are the mocked contracts. Please [see here](https://github.com/zama-ai/fhevm/pull/594/commits/db46e019c1213c641324b5d5eea81aab7338fb56#diff-11acb2d556c4a634fdde4f84cc97005bbfe910dfb55f1e811ab08856ba711a2aR536) where are the contracts with commented arguments.

jatZama (approved):

PR #598, opened 2025-07-18
feat(host-contracts): start introducing the extraData field at KMSVerifier

isaacdecoded:
Refs [#209](https://github.com/zama-ai/fhevm-internal/issues/209)

### DIFF
--- a/host-contracts/codegen/templates.ts
+++ b/host-contracts/codegen/templates.ts
@@ -406,7 +406,8 @@ function generateKMSVerifierInterface(): string {
     function verifyDecryptionEIP712KMSSignatures(
         bytes32[] memory handlesList,
         bytes memory decryptedResult,
-        bytes[] memory signatures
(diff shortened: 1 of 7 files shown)


jatZama:
My last commit also vastly simplified the mocked relayer, which was possible thanks to deprecation of ebytesXXX types.
Note: this has implications on:
- relayer which needs to be changed to take into account new extraData argument.
- relayer-sdk which needs to be changed to take into account simplified encoding for checking signature in public decrypt http endpoint.
- KMS connector which needs to take into account simplified encoding thanks to deprecation of ebytesXXX types.

melanciani (approved):
thanks ! ~~but probably better not to merge this before considering the changes in impacted off-chain components like relayer~~ EDIT: we agreed on merging it , as relayer i aware of the change and we'll consider it once the gateway is ready

PR #599, opened 2025-07-18
feat(host-contracts): start introducing the extraData field at InputVerifier

isaacdecoded:
Refs https://github.com/zama-ai/fhevm-internal/issues/209

### DIFF
--- a/host-contracts/contracts/InputVerifier.sol
+++ b/host-contracts/contracts/InputVerifier.sol
@@ -83,11 +83,13 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
         address contractAddress;
         /// @notice The chainId of the contract requiring the ZK Proof verification.
         uint256 contractChainId;
+        /// @notice Generic bytes metadata for versioned payloads.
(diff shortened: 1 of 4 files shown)


jatZama (approved):
LGTM

host-contracts/contracts/InputVerifier.sol
```diff
@@ -388,7 +393,8 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
                         keccak256(abi.encodePacked(ctVerification.ctHandles)),
                         ctVerification.userAddress,
                         ctVerification.contractAddress,
-                        block.chainid
+                        block.chainid,
```
jatZama, line 396: maybe you can change this by putting the correct value in the struct, as discussed on friday, wdyt? it would make the code more readable and less error prone in case of a future update. since this part was legacy code from the time we weren't cross-chain i think.

isaacdecoded: To be solved here: https://github.com/zama-ai/fhevm-internal/issues/224

PR #605, opened 2025-07-21
chore(kms-connector): guaranteed seq blocks polling and trx gas management

maksymsur:
WIP


kms-connector/simple-connector/src/core/utils/nonce_manager.rs
```diff
@@ -360,7 +360,7 @@ where
             // Use aggressive gas price from first attempt
             if attempt == 0 {
                 if let Some(gas_price) = tx.gas_price {
-                    tx.gas_price = Some(gas_price * 700 / 100); // 600% boost for fast processing
+                    tx.gas_price = Some(gas_price * 150 / 100); // 50% boost for fast processing
```
jatZama: We need to put a buffer on gas (ie gasLimit), not necessarily gas price, as we noticed issue was coming from gas under-estimation, not from gasPrice estimation.
can you try something like this?
```
    let est = provider.estimate_gas(tx.clone()).await?;
    let buffered = est * 120 / 100
    tx = tx.gas_limit(buffered); 
 ``` 
This will add 20% buffer to gas (not gas_price)

maksymsur: thank you! this is a very valuable observation!

eudelins-zama (approved):

PR #606, opened 2025-07-21
feat(kms-connector): healthcheck implementation

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/211

Some queries to the healthcheck endpoints before/after shutting down some of the external services used by the connector (kms-core, then anvil node, then database):

Screenshot From 2025-07-22 09-22-00

Note: e2e tests are good: https://github.com/zama-ai/fhevm/actions/runs/16437631242/job/46450769613


kms-connector/crates/utils/src/monitoring/health.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// Performs the database healthcheck.
+///
+/// Stores the potential error in the `errors` vector.
+pub async fn database_healthcheck(
+    db_pool: &Pool<Postgres>,
+    timeout: Duration,
+    errors: &mut Vec<String>,
+) -> bool {
+    match tokio::time::timeout(timeout, db_pool.acquire()).await {
```
dartdart26, line 23: Here, do we rely on acquire() calling ping?

dartdart26 (approved):
Nice!

kms-connector/crates/utils/src/monitoring/health.rs
eudelins-zama: I'm not aware of such ping when calling `acquire`, but according to the docs, `acquire` "Retrieves a connection from the pool". So I think that is what we want to try for a healthcheck.
We could do a dumb query like "SELECT 1", but it would call acquire under the hood first I believe, and then we would check a dummy result for nothing?

tawadaa (approved):
Excellent, thanks!

PR #608, opened 2025-07-21
feat(charts): improve kms-connector chart for better usability

piizama:
* Add resources block
* Explicitly set connector config values in `kmsConnector.config` rather than rely on environment variables
* Remove unused config file
* Don't set any default values so the container will fail fast if those are not properly set


tawadaa (approved):
Nice, thanks!

PR #615, opened 2025-07-23
fix(coprocessor): fhevm-listener, retry on sql timeout

rudy-6-4:
fix timeout statement detection
always retry even if the error looks non transient, could be a detection bug


dartdart26 (approved):
Looks good!

PR #616, opened 2025-07-23
feat(kms-connector): make test setup podman-friendly

dartdart26:
Do not use host networking.

Use port mapping for the Postgres container.

Run Anvil inside a container and also use port mapping for it. Make sure the gateway-contracts container can connect to Anvil to deploy.

### DIFF
--- a/kms-connector/crates/utils/src/tests/setup/db.rs
+++ b/kms-connector/crates/utils/src/tests/setup/db.rs
@@ -2,6 +2,10 @@ use sqlx::{Pool, Postgres};
 use testcontainers::{ContainerAsync, GenericImage, ImageExt, core::WaitFor, runners::AsyncRunner};
 use tracing::info;
 
+use crate::tests::setup::pick_free_port;
+
+const POSTGRES_PORT: u16 = 5432;
+
 pub struct DbInstance {
     /// Use to keep the database container running during the tests.
     _db_container: ContainerAsync<GenericImage>,
@@ -11,11 +15,13 @@ pub struct DbInstance {
 
 impl DbInstance {
     pub async fn setup() -> anyhow::Result<Self> {
+        let host_port = pick_free_port();
         info!("Starting Postgres container...");
         let container = GenericImage::new("postgres", "17.5")
             .with_wait_for(WaitFor::message_on_stderr(
                 "database system is ready to accept connections",
             ))
+            .with_mapped_port(host_port, POSTGRES_PORT.into())
             .with_env_var("POSTGRES_USER", "postgres")
             .with_env_var("POSTGRES_PASSWORD", "postgres")
             .start()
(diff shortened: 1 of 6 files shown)


kms-connector/crates/utils/src/tests/setup/db.rs
```diff
@@ -2,6 +2,10 @@ use sqlx::{Pool, Postgres};
 use testcontainers::{ContainerAsync, GenericImage, ImageExt, core::WaitFor, runners::AsyncRunner};
 use tracing::info;
 
+use crate::tests::setup::pick_free_port;
+
+const POSTGRE_PORT: u16 = 5432;
```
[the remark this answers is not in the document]
eudelins-zama: Nit: indeed

kms-connector/crates/utils/src/tests/setup/db.rs
```diff
@@ -11,23 +15,25 @@ pub struct DbInstance {
...
+        let host_port = pick_free_port();
         info!("Starting Postgres container...");
         let container = GenericImage::new("postgres", "17.5")
             .with_wait_for(WaitFor::message_on_stderr(
                 "database system is ready to accept connections",
             ))
+            .with_mapped_port(host_port, POSTGRE_PORT.into())
             .with_env_var("POSTGRES_USER", "postgres")
             .with_env_var("POSTGRES_PASSWORD", "postgres")
+            .with_container_name("postgres")
```
eudelins-zama: We should avoid setting the container name. If we want to run tests in parallel at some points, there will be conflict

kms-connector/crates/utils/src/tests/setup/gw.rs
```diff
@@ -57,36 +65,67 @@ impl GatewayInstance {
...
 
-pub async fn setup_anvil_gateway() -> anyhow::Result<AnvilInstance> {
+pub async fn setup_anvil_gateway(host_port: u16) -> anyhow::Result<ContainerAsync<GenericImage>> {
+    let docker = docker_client_instance().await?;
+    let _network = docker
+        .create_network(CreateNetworkOptions {
+            name: ANVIL_NETWORK,
+            ..Default::default()
+        })
+        .await?;
```
eudelins-zama: Could we do the same thing as for DB and KMS Core instance, ie do not rely on a dedicated network?

Once again the issue is parallel testing: this will cause conflict. Moreover, network are not automatically deleted on drop, so either we have more overhead to clean up the test setup, or we'll manual chore before running tests. (I used dedicated network for KMS Core setup at first, and it was painful to handle in the end so I used host network)

kms-connector/crates/utils/src/tests/setup/gw.rs
```diff
@@ -57,36 +65,67 @@ impl GatewayInstance {
...
+            "--port",
+            ANVIL_PORT.to_string().as_str(),
+            "--chain-id",
+            CHAIN_ID.to_string().as_str(),
+            "--mnemonic",
+            TEST_MNEMONIC,
+            "--block-time",
+            "1",
+        ])
+        .with_container_name(ANVIL_CONTAINER)
```
eudelins-zama: Same, no container name if possible

eudelins-zama (changes requested):
Few details to discuss, otherwise it looks good! :slightly_smiling_face:

kms-connector/crates/utils/src/tests/setup/gw.rs
dartdart26: Re-creating the network if it exists doesn't happen, AFAIK. So wouldn't be a problem.

But I think I can remove it, because I initially wanted to use it to use DNS. But since I don't use DNS, I think I can remove it. Let me do that.

kms-connector/crates/utils/src/tests/setup/db.rs
dartdart26: Yes, let me do that.

kms-connector/crates/utils/src/tests/setup/gw.rs
dartdart26: Will remove it!

eudelins-zama (approved):
Thx a lot! LGTM

PR #619, opened 2025-07-23
chore(gateway-contracts): add pausing tasks for gateway and host contracts

melanciani:
closes https://github.com/zama-ai/fhevm-internal/issues/208


gateway-contracts/tasks/pauseContracts.ts
```diff
@@ -0,0 +1,172 @@
...
+    "If proxy address from the /addresses directory should be used",
+    false,
+    types.boolean,
+  )
+  .setAction(async function ({ useInternalProxyAddress }, { ethers }) {
+    await unpauseContract("Decryption", ethers, useInternalProxyAddress);
+  });
+
+// Pause all the contracts
+// The following contracts are pausable but don't have pausable functions yet, so we don't need to
```
jatZama: i think that the pausing will be centralized from gatewayConfig, meaning pausing of GatewayCOnfig will pause all other contracts at once in a single tx, right?

jatZama:
Important comment to be confirmed before I could approve and then you can merge:
https://github.com/zama-ai/fhevm/pull/619/files#r2226079350
Currently you have one tx on gateway per contract. This task should be changed soon to be atomic, right?

gateway-contracts/.env.example
```diff
@@ -7,8 +7,9 @@ CHAIN_ID_GATEWAY="31337" # (uint256)
 # Mnemonic
 MNEMONIC="adapt mosquito move limb mobile illegal tree voyage juice mosquito burger raise father hope layer" # (string)
 
-# Deployer
+# Private keys
 DEPLOYER_PRIVATE_KEY="0x7136d8dc72f873124f4eded25f3525a20f6cee4296564c76b44f1d582c57640f" # accounts[1], private key (bytes32)
+PAUSER_PRIVATE_KEY="0x7ae52cf0d3011ef7fecbe22d9537aeda1a9e42a0596e8def5d49970eb59e7a40" # accounts[2], private key (bytes32)
```
isaacdecoded, line 12: In the end, this should be a smart account address as [requested here](https://github.com/zama-ai/fhevm/pull/539/files#diff-57fe50f20e41f3acf58a40f3dc59f9260e4d99bba297bc65af0d3bd641451ea2R21).

Also, these pause/unpause tasks will change once the multisig feature is introduced. Same for the upgrade tasks that will be changed to use the owner smart account.

gateway-contracts/tasks/pauseContracts.ts
isaacdecoded: I think we discussed about centralizing just the contract's ownership (owner role), but it could make sense to centralize the pauser role as well 🤔 Meaning we should have modifiers: `onlyGatewayConfigOwner` and `whenGatewayConfigNotPaused` I guess.

Although not sure if at some point we'd like to have more granularity for the pausing mechanism, e.g., pause decryptions but not input verifications.

gateway-contracts/.env.example
melanciani: yes ofc but better do it in several steps

gateway-contracts/tasks/pauseContracts.ts
melanciani: hmm I'm not sure about this though, I think we still want to be able to pause only the input verificaiton without decryption, and vice versa for example

melanciani: let me ask

gateway-contracts/.env.example
melanciani: basically: depending on which PR is merged first, the other will need to be rebased

gateway-contracts/tasks/pauseContracts.ts
```diff
@@ -0,0 +1,172 @@
...
+    dotenv.config({ path: envFilePath, override: true });
+  }
+
+  // Determine env variable name for the proxy contract address
+  const nameSnakeCase = pascalCaseToSnakeCase(name);
+  const addressEnvVarName = `${nameSnakeCase.toUpperCase()}_ADDRESS`;
+
+  // Get the proxy address
+  const proxyAddress = getRequiredEnvVar(addressEnvVarName);
+  if (!proxyAddress) {
```
isaacdecoded: @RomanBredehoft, I just realized that these validations (`!proxyAddress`) are not necessary since the `getRequiredEnvVar` already throws an error if the value is not present in `process.env`

melanciani: ah yes you're right, the function is made for that

gateway-contracts/tasks/pauseContracts.ts
melanciani: decision is to be able to do both : 
- pause all contracts in a single tx
- pause a specific contract in a single tx

let me update the PR

gateway-contracts/.env.example
melanciani: ~~we decided that we first merge this PR and then https://github.com/zama-ai/fhevm/pull/539 will be rebased~~

it was for another PR, maybe I'll need to rebase this one here instead we'll see

gateway-contracts/tasks/pauseContracts.ts
melanciani: done in last commit !

gateway-contracts/contracts/interfaces/IGatewayConfig.sol
```diff
@@ -108,6 +108,16 @@ interface IGatewayConfig {
     /// @param nKmsNodes The number of KMS nodes.
     error InvalidHighUserDecryptionThreshold(uint256 userDecryptionThreshold, uint256 nKmsNodes);
 
+    /**
+     * @notice Emitted when all the pausable gateway contracts are paused.
+     */
+    event AllGatewayContractsPaused();
```
isaacdecoded: nit: we decided long ago not to use past tense for the events, so maybe we should rename these events to: `PauseAllGatewayContracts` and `UnpauseAllGatewayContracts`

gateway-contracts/tasks/pauseContracts.ts
```diff
@@ -0,0 +1,162 @@
...
+    "If proxy address from the /addresses directory should be used",
+    false,
+    types.boolean,
+  )
+  .setAction(async function ({ useInternalProxyAddress }, { ethers }) {
+    await unpauseSingleContract("Decryption", ethers, useInternalProxyAddress);
+  });
+
+// Pause all the contracts
+// The following contracts are pausable but don't have pausable functions yet, so they are
```
isaacdecoded, line 114: nit: maybe we could be more specific, something like: The following contracts are pausable but don't apply the `whenNotPaused` modifier to any of their functions.

isaacdecoded (approved):
LGTM! 👍 

Btw, maybe for mainnet we'd like to clean the Pausable inheritance from those contracts not using it? Similar to the `Ownership2StepUpgradeable` that should be removed after centralizing the ownership in the GatewayConfig contract 🤔

melanciani:
> Btw, maybe for mainnet we'd like to clean the Pausable inheritance from those contracts not using it? Similar to the Ownership2StepUpgradeable that should be removed after centralizing the ownership in the GatewayConfig contract 🤔

yes you're right, I created an issue to track this down : https://github.com/zama-ai/fhevm-internal/issues/279

jatZama (approved):

PR #620, opened 2025-07-23
chore(kms-connector): static test setup

eudelins-zama:
* enables shared setup for tests
* fix flaky tests
* use test timeout for integration tests

Note: I still have issues for tx manager integration tests: the events filter I used to ensure tx are emitted to the Gateway sometimes miss the events, even if the tx manager claims the tx was sent... If you have any idea @dartdart26, feel free to share it

### DIFF
--- a/kms-connector/crates/tx-sender/tests/integration_tests.rs
+++ b/kms-connector/crates/tx-sender/tests/integration_tests.rs
@@ -5,10 +5,15 @@ use alloy::primitives::U256;
 use anyhow::anyhow;
 use common::insert_rand_public_decrypt_response;
 use connector_utils::{
-    tests::setup::{TestInstance, TestInstanceBuilder},
+    tests::setup::{
+        TestInstance,
+        shared::{clean_test_instance, run_with_shared_db_gw_setup},
+    },
     types::KmsResponse,
 };
+use ctor::dtor;
 use rstest::rstest;
+use serial_test::serial;
 use std::time::Duration;
 use tokio::task::JoinHandle;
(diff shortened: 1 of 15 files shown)


kms-connector/crates/tx-sender/tests/integration_tests.rs
```diff
@@ -18,148 +23,158 @@ use tx_sender::core::{
     DbKmsResponsePicker, DbKmsResponseRemover, TransactionSender, tx_sender::TransactionSenderInner,
 };
 
+#[dtor]
```
dartdart26, line 26: My concern with using `dtor` (albeit it being super useful on a high-level) is that the behavior is not well-defined, it is platform-specific and has races. For example, I tried to use it for test cleanup in the coprocessor and the issue was that thread-local storage was no longer available in the dtor and sometimes it worked, sometimes not.

See the Warnings section here: https://crates.io/crates/ctor

I don't know of a better solution to be honest, but to run tests one at a time and constructing/destructing the test instance in a well-defined manner in each test.

dartdart26 (commented):
Good idea! I have some concerns about `dtor`, let's chat offline first to see if we can solve it or be certain it works.

dartdart26 (approved):
As we discussed offline:
 * try if it works on MacOS
 * put a comment to be explicit we know about the limitations and possible issues with `dtor`
 * cleanup won't have an impact on whether test is successful or not (unless dtor panics)
 * merge and revise later if we see issues
 
 Thanks!

eudelins-zama:
Aborted

PR #623, opened 2025-07-24
feat(host-contracts): Remove dead code in FHEVMExecutor due to legacy…

jatZama:
… ebytesXXX types.
Ths should close issue https://github.com/zama-ai/fhevm-internal/issues/249


rudy-6-4 (approved):
Ok the right check is already done in _verityAndReturnType

PR #625, opened 2025-07-24
feat(kms-connector): recoverable nonce manager

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/247

- I had to add an `under_process` field to KMS response in DB to avoid picking responses multiple times, now that we pick them by batch
- Implemented a `NonceManagedProvider`: heavily inspired by the coprocessor one... :slightly_smiling_face: . Just tried to avoid the `Deref` [antipattern](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html), so it is more verbose...

Note: e2e tests https://github.com/zama-ai/fhevm/actions/runs/16518946299


kms-connector/crates/utils/src/provider.rs
dartdart26, line 1: Do we really need all this boilerplate? I get your concern, but it is a lot of code to maintain :) Is `inner()` enough if we remove `deref()`?

dartdart26 (approved):
Looks good!

kms-connector/crates/utils/src/provider.rs
dartdart26: We discussed offline - will fix it in the coprocessor by removing deref.

eudelins-zama: Discussed offline

PR #626, opened 2025-07-24
feat(host-contracts): initialize chainId in ctVerif and use it in has…

jatZama:
…h method
This should close this issue : https://github.com/zama-ai/fhevm-internal/issues/224


isaacdecoded (approved):
Great! 👍

PR #629, opened 2025-07-25
feat(gateway-contracts): add requestId validation in responses

jatZama:
This should fix https://github.com/zama-ai/fhevm-internal/issues/254


isaacdecoded (approved):
LGTM 💯 Great catch!

PR #631, opened 2025-07-25
chore(host-contracts): use shared constant for handle version

jatZama:
This should close https://github.com/zama-ai/fhevm-internal/issues/252


isaacdecoded (approved):
LGTM 👍

PR #641, opened 2025-07-29
chore(kms-connector): use public kms repo

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/271

Note: contains some crates updates too. This is mandatory because kms repo pin their dependencies, and as kms-grpc is a dependency of the connector, we must align with the version they are using for now


dartdart26 (approved):
Nice!

PR #644, opened 2025-07-29
fix(coprocessor): fhevm-listener, address pre-filter event decoding

rudy-6-4:


rudy-6-4:
Note: this also remove the useless feature of not handling some event ACL or TFHE. This can still be done in test by providing a fake address.

antoniupop (approved):

jonik570771279:
what it?

PR #647, opened 2025-07-29
feat(test-suite): lean uint64 user-decrypt decryption benchmark

fd0r:


jatZama (approved):

fd0r:
Current limitation with this new approach is that FHE-rand costs 100_000 HCU and the current limit per block is 20_000_000, thus limiting to max 200 new created value.

TODO(in-follow-up PR): paginate/batch refresh of values

PR #649, opened 2025-07-29
feat(gateway-contracts): add getters to get txSenders involved in consensus

melanciani:
this will help testing multi-coprocessor : https://github.com/zama-ai/fhevm-internal/issues/296 cc @leventdem 

done: 
- Decryption 
- InputVerification
- MultichainAcl
- CiphertextCommits

while working on it, I thought of maybe adding errors if the consensus hasn't been reached yet, but in the end I think it's fine if we let without : 
- some functions don't require signature verification, like allow calls, so we can track down which tx sender answered even before the threshold is reached, which looks like a nice tracking feature
- other functions (public decrypt, add ct, verify proof) will simply return an empty list if the getter is called before consensus, which seems expected 
    - note that add ct does not have signature verificaiton, but still has an actual consensus on the ct digests, hence why it has a similar behavior here

closes https://github.com/zama-ai/fhevm-internal/issues/219


isaacdecoded (approved):
LGTM!

PR #651, opened 2025-07-30
fix(kms-connector): tracing on config error

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/255


dartdart26 (approved):
Looks good!

PR #657, opened 2025-07-30
chore(kms-connector): improve integration tests

eudelins-zama:
Rely on custom `wait_for_log` function instead of `sleep` to avoid race condition in tests.

### DIFF
--- a/kms-connector/Cargo.lock
+++ b/kms-connector/Cargo.lock
@@ -3516,6 +3516,7 @@ dependencies = [
  "connector-utils",
  "fhevm_gateway_rust_bindings",
  "prometheus",
+ "rstest",
  "serde",
  "serial_test",
  "sqlx",
--- a/kms-connector/crates/gw-listener/Cargo.toml
+++ b/kms-connector/crates/gw-listener/Cargo.toml
@@ -24,6 +24,7 @@ tracing-subscriber.workspace = true
 
 [dev-dependencies]
 connector-utils = { workspace = true, features = ["tests"] }
+rstest.workspace = true
 serial_test.workspace = true
 tempfile.workspace = true
 testcontainers.workspace = true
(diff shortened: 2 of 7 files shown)


dartdart26 (commented):
I am wondering if waiting for specific log lines doesn't make it more brittle in the sense now there is a dependency between logs and tests.

Also, `wait_for_log()` is also an infinite loop basically, comparable to polling. But that's not a big issue, IMO.

It's just that I am not convinced the dependency we introduce is worth it. But it is a test after all, so might as well live with it, not a big deal. What are your thoughts on that?

dartdart26 (approved):
I put a comment about logs, but now that I think more, I think it is fine, so approving it.

eudelins-zama:
@dartdart26 agree with you that it creates a dependency with the logs. However, IMO it is a tradeoff between:
- relying on `tokio::time::sleep`, which lead to race conditions in tests
- messing the code API, to somehow pass a dummy channel that will send empty value to tell the test it is ready to continue (hate this solution tbh)
- or depending between on logs

I am more comfortable with this last solution. Plus, we can choose in which test we use this `wait_for_log` or not, this is not mandatory. Also, if tests break because logs changed, it would be an easy fix I guess

dartdart26:
> @dartdart26 agree with you that it creates a dependency with the logs. However, IMO it is a tradeoff between:
> 
> * relying on `tokio::time::sleep`, which lead to race conditions in tests
> * messing the code API, to somehow pass a dummy channel that will send empty value to tell the test it is ready to continue (hate this solution tbh)
> * or depending between on logs
> 
> I am more comfortable with this last solution. Plus, we can choose in which test we use this `wait_for_log` or not, this is not mandatory. Also, if tests break because logs changed, it would be an easy fix I guess

I was thinking of just polling, but I think it is fine to merge as is!

PR #663, opened 2025-07-31
test(gateway-contracts): introduce gas estimation tests

isaacdecoded:
Closes https://github.com/zama-ai/fhevm-internal/issues/287

### DIFF
--- a/gateway-contracts/Makefile
+++ b/gateway-contracts/Makefile
@@ -21,6 +21,9 @@ test: clean
 coverage: clean
 	DOTENV_CONFIG_PATH=$(ENV_PATH) npx hardhat coverage
 
+report-gas: clean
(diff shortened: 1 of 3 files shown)


melanciani:
ideally for these gas consumption to be relevant we would want to consider an equivalent env as the test net one : 
- 13 KMS nodes
- 1 copro (for now)
- for user decryption: user reencrypted shares of the same size (roughly) : 
    - for this, you can ask the MPC team: normally the size depends on the fhe type of the decrypted ct 
    - the sizes is going to change soon btw so maybe take this one

maybe we would need a dedicated env var for these tests only ? not sure what's the best option here

also, do you think we could easily generate a table in a md file (like we do for selectors) based on these tests ? maybe check with @maksymsur what would be the best format here

isaacdecoded:
Closed in favor of the stress test tooling [here](https://github.com/zama-ai/fhevm/pull/702).

PR #664, opened 2025-07-31
refactor(charts): rename coprocessor binaries

tawadaa:


tawadaa:
Closing in favor of #679

PR #666, opened 2025-07-31
feat(host-contracts): introduce the extraData field at InputVerifier

isaacdecoded:
Closes https://github.com/zama-ai/fhevm-internal/issues/298

### DIFF
--- a/host-contracts/contracts/InputVerifier.sol
+++ b/host-contracts/contracts/InputVerifier.sol
@@ -204,7 +204,7 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
         if (!isProofCached) {
             /// @dev bundleCiphertext is compressedPackedCT+ZKPOK
             ///      inputHandle is keccak256(keccak256(bundleCiphertext)+index)[0:20] + index[21] + chainId[22:29] + type[30] + version[31]
-            ///      and inputProof is numHandles + numSigners + handles + coprocessorSignatures (1 + 1 + numHandles*32 + 65*numSigners)
+            ///      and inputProof is numHandles + numSigners + handles + coprocessorSignatures (1 + 1 + 32*numHandles + 65*numSigners + extraData bytes)
 
             uint256 inputProofLen = inputProof.length;
             if (inputProofLen == 0) revert EmptyInputProof();
--- a/host-contracts/contracts/InputVerifier.sol
+++ b/host-contracts/contracts/InputVerifier.sol
@@ -214,8 +214,11 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
             /// @dev This checks in particular that the list is non-empty.
             if (numHandles <= indexHandle || indexHandle > 254) revert InvalidIndex();
 
+            /// @dev The extraData is the rest of the inputProof bytes after the numHandles + numSigners + handles + coprocessorSignatures.
+            uint256 extraDataOffset = 2 + 32 * numHandles + 65 * numSigners;
+
             /// @dev Check that the inputProof is long enough to contain at least the numHandles + numSigners + handles + coprocessorSignatures
-            if (inputProofLen < 2 + 32 * numHandles + 65 * numSigners) revert DeserializingInputProofFail();
+            if (inputProofLen < extraDataOffset) revert DeserializingInputProofFail();
 
             /// @dev Deserialize handle and check that they are from the correct version.
             bytes32[] memory listHandles = new bytes32[](numHandles);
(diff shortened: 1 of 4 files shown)


host-contracts/contracts/InputVerifier.sol
```diff
@@ -211,7 +213,9 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
 
             /// @dev This checks in particular that the list is non-empty.
             if (numHandles <= indexHandle || indexHandle > 254) revert InvalidIndex();
-            if (inputProofLen != 2 + 32 * numHandles + 65 * numSigners) revert DeserializingInputProofFail();
+
+            /// @dev Check that the inputProof is long enough to contain at least the numHandles + numSigners + handles + coprocessorSignatures
+            if (inputProofLen < 2 + 32 * numHandles + 65 * numSigners) revert DeserializingInputProofFail();
```
isaacdecoded: Just to double-check the impact: is this correct to make this change in the evaluation of the `inputProofLen`? I mean, now that we have "remaining" bytes in the `inputProof` for `extraData`, we should check `inputProofLen < minimum expected size` instead of `!=`.

cc @immortal-tofu

melanciani: I agree , imo this change makes a lot of sense

host-contracts/contracts/InputVerifier.sol
```diff
@@ -359,6 +365,24 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
             );
     }
 
+    function _extractExtraData(
```
melanciani: ~~can you please add some natspec doc for this internal func ?~~

actually if you consider the factorization I mention, I don't think this internal function has much use in the end so maybe you can remove it (pretty much like you do in the KmsVerifier PR)

host-contracts/contracts/InputVerifier.sol
```diff
@@ -359,6 +365,24 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+        uint256 numSigners
+    ) internal pure virtual returns (bytes memory extraData) {
+        uint256 handlesSize = numHandles * 32;
+        uint256 signaturesSize = numSigners * 65;
+
+        /// @dev The extraData is the rest of the inputProof bytes after the numHandles + numSigners + handles + coprocessorSignatures.
+        uint256 extraDataOffset = 2 + handlesSize + signaturesSize;
+        uint256 extraDataSize = inputProof.length - extraDataOffset;
+
+        extraData = new bytes(extraDataSize);
```
melanciani: I think you can just return `inputProof[extraDataOffset:extraDataOffset+extraDataSize]`, sliced arrays are supported from solidity 0.6 : https://soliditylang.org/blog/2020/05/26/array-slices/

host-contracts/contracts/InputVerifier.sol
```diff
@@ -232,11 +236,13 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
                     signatures[j][i] = inputProof[2 + 32 * numHandles + 65 * j + i];
                 }
             }
+
             CiphertextVerification memory ctVerif;
             ctVerif.ctHandles = listHandles;
             ctVerif.userAddress = context.userAddress;
             ctVerif.contractAddress = context.contractAddress;
             ctVerif.contractChainId = block.chainid;
+            ctVerif.extraData = _extractExtraData(inputProof, numHandles, numSigners);
```
melanciani: ~~I'm not sure it's relevant to use this function until we actually use extraData , no ? it adds gas cost for nothing for now. Maybe keep the func's definition but don't use it (and thus add a warning silencer) ?~~

sorry, it's needed for the struct, my bad

melanciani (changes requested):
thanks, just a small change request, besides I don't know the tests too well but it looks fine !

host-contracts/contracts/InputVerifier.sol
```diff
@@ -359,6 +365,24 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+    function _extractExtraData(
+        bytes memory inputProof,
+        uint256 numHandles,
+        uint256 numSigners
+    ) internal pure virtual returns (bytes memory extraData) {
+        uint256 handlesSize = numHandles * 32;
+        uint256 signaturesSize = numSigners * 65;
+
+        /// @dev The extraData is the rest of the inputProof bytes after the numHandles + numSigners + handles + coprocessorSignatures.
+        uint256 extraDataOffset = 2 + handlesSize + signaturesSize;
```
melanciani: this offset ha already been computed in line 218 for checking the size, maybe you can factorize both

melanciani (changes requested):
~~thanks, a few changes to request !~~

sorry for the double request changes, it was meant for the other PR 😅

host-contracts/contracts/InputVerifier.sol
isaacdecoded: > For now, array slices are only supported for calldata arrays.

Unfortunately, it seems to be only for `calldata` arrays. In the IDE the error is: `Index range access is only supported for dynamic calldata arrays.`

melanciani: that's a shame, but ok thanks for pointing this out !

host-contracts/contracts/InputVerifier.sol
isaacdecoded: Additional `_extractExtraData` and suggested refactorization applied [here](https://github.com/zama-ai/fhevm/pull/666/commits/b092022b75068e5ddee5735e4c0509fe6a8c9912#diff-cfdd06859b5fbcd907a11252dc94ef7ffb38c243eeecfcf094deab1541a50b92R249). Thanks a lot for the improvement ✊

melanciani (approved):
looks good to me !

PR #669, opened 2025-07-31
feat(gateway-contracts): add getters for decryption consensus materials

melanciani:
based on https://github.com/zama-ai/fhevm/pull/649 since I needed to re-use some parts of it ! once it's merged, I'll rebase on main 

however: maybe we don't want to merge this yet if we don't update the relayer flow, because it adds an SSTORE in the public decryption consensus step

closes https://github.com/zama-ai/fhevm-internal/issues/285


gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -154,10 +156,9 @@ contract CiphertextCommits is
...
-            // As explained above, a "late" valid coprocessor could still see its transaction sender
-            // address be added to the list after consensus. This var is here to make the
-            // `getCiphertextMaterials` and `getSneCiphertextMaterials` methods retrieve this list
-            // more easily
+            // A "late" valid coprocessor could still see its transaction sender address be added to
+            // the list after consensus. This variable is here to be able to retrieve this list later
+            // by only knowing the handle, since a consensus can only happen once per handle
             $._ctHandleConsensusHash[ctHandle] = addCiphertextHash;
 
             emit AddCiphertextMaterial(
```
isaacdecoded: Now that we have the `getAddCiphertextMaterialConsensusTxSenders`, shouldn't we remove the list of `coprocessorTxSenderAddresses` from the event to reduce gas? Maybe also the ciphertext digests? 🤔

gateway-contracts/contracts/Decryption.sol
```diff
@@ -295,11 +318,26 @@ contract Decryption is
...
+            // after consensus. This storage variable is here to be able to retrieve this list later
+            // by only knowing the decryption ID, since a consensus can only happen once per decryption
+            // request, independently of the decryption response type (public or user).
+            $.decryptionConsensusDigest[decryptionId] = digest;
+
+            // Store the decrypted result for the public decryption response in order to be able to
+            // retrieve it later
+            $.publicDecryptedResult[decryptionId] = decryptedResult;
+
             emit PublicDecryptionResponse(decryptionId, decryptedResult, verifiedSignatures);
```
isaacdecoded, line 341: Same question here, or is this gas optimization on the events for another issue?

isaacdecoded (approved):
Great! 👍

gateway-contracts/contracts/CiphertextCommits.sol
melanciani: this PR is supposed to be a first step , we have other issues for simplifying the events : https://github.com/zama-ai/fhevm-internal/issues/292

but we don't do it yet because it impacts the relayer and we don't plan to do this in fhevm 0.8

gateway-contracts/contracts/Decryption.sol
melanciani: same as above

melanciani:
let's postpone the merge of this PR for now, we'll do it after fhevm 0.8 most likely

melanciani:
we'll most likely go with https://github.com/zama-ai/fhevm-internal/issues/331 instead. We can re-open if needed later

PR #670, opened 2025-07-31
chore(sdk): make keys directory optional

eudelins-zama:
Make the `keys_directory` optional if the SDK user does not want to create any inputs.


leventdem (approved):
+1

PR #671, opened 2025-07-31
feat(gateway-contracts): add getters for input proof consensus materials

melanciani:
like https://github.com/zama-ai/fhevm/pull/669https://github.com/zama-ai/fhevm/pull/669, it is based on https://github.com/zama-ai/fhevm/pull/649 since I needed to re-use some parts of it ! once it's merged, I'll rebase on main 

however: maybe we don't want to merge this yet if we don't update the relayer flow, because it adds an SSTORE in the consensus step

closes https://github.com/zama-ai/fhevm-internal/issues/294


melanciani:
let's postpone the merge of this PR for now, we'll do it after fhevm 0.8 most likely

melanciani:
we'll most likely go with https://github.com/zama-ai/fhevm-internal/issues/331 instead. We can re-open if needed later

PR #672, opened 2025-07-31
feat(host-contracts): introduce the extraData and decryptionProof fields at KMSVerifier

isaacdecoded:
Note that this PR not only introduces the `extraData` field at the `KMSVerifier` but also refactors its `verifyDecryptionEIP712KMSSignatures` method to receive a `bytes decryptionProof` argument containing `number of signers + KMS signatures + extra data`. This refactor is based on the changes requested in the [Add ERC-7995 PR](https://github.com/ethereum/ERCs/pull/1143/files#diff-05bcded41894aaf6a46d15840fd9d378934b820bda8e59c22b0dcfec9fe667a2R342)

Closes https://github.com/zama-ai/fhevm-internal/issues/298

PR #677, opened 2025-08-01
chore(library-solidity): reflect changes for new decryptionProof containing signatures and extraData

isaacdecoded:
Note that this PR also includes the refactor on contract addresses, which are now consolidated in a single file.

Closes https://github.com/zama-ai/fhevm-internal/issues/297


library-solidity/hardhat.config.ts
```diff
@@ -51,12 +51,11 @@ task('test', async (_taskArgs, hre, runSuper) => {
     fs.mkdirSync('node_modules/@fhevm/core-contracts/addresses');
   }
 
+  // TODO: dependency conflict with @fhevm/core-contracts which comes from the old version from github.com/zama-ai/fhevm-backend
```
isaacdecoded: This is currently using an [old released version of the host-contracts](https://github.com/zama-ai/fhevm/blob/main/library-solidity/package.json#L40) that does not contain the new `extraData` field in the KMSVerifier contract. Hence, the tests will fail until we make a new release containing the `extraData` field and update the package version. I guess the plan should be first finish and merge the current PR on `host-contracts` for the [KMSVerifier](https://github.com/zama-ai/fhevm/pull/672) and then make a new release?

cc @RomanBredehoft @leventdem @immortal-tofu

isaacdecoded: We solved it by using npm workspaces. Thus, this is not a requirement anymore to make a release of the core-contracts for the library Solidity to use them, but importing them directly from the monorepo `@fhevm/core-contracts` workspace

library-solidity/hardhat.config.ts
```diff
@@ -47,16 +47,10 @@ task('coverage').setAction(async (taskArgs, hre, runSuper) => {
 });
 
 task('test', async (_taskArgs, hre, runSuper) => {
-  if (!fs.existsSync('node_modules/@fhevm/core-contracts/addresses')) {
-    fs.mkdirSync('node_modules/@fhevm/core-contracts/addresses');
-  }
-
-  const sourceDir = path.resolve(__dirname, 'node_modules/@fhevm/core-contracts/contracts');
+  const sourceDir = path.resolve(__dirname, '../node_modules/@fhevm/core-contracts/contracts');
```
melanciani, line 50: why has the path changed ? 🤔

library-solidity/hardhat.config.ts
```diff
@@ -66,7 +60,6 @@ task('test', async (_taskArgs, hre, runSuper) => {
     await hre.run('task:deployAllHostContracts');
   }
 
-  await hre.run('compile:specific', { contract: 'examples' });
```
melanciani, line 69: are you sure we don't need this anymore ?

library-solidity/lib/FHE.sol
```diff
@@ -8952,7 +8960,7 @@ library FHE {
             IKMSVerifier($.KMSVerifierAddress).verifyDecryptionEIP712KMSSignatures(
```
melanciani, line 8960: so in the end, we do not need to change the `calldatacopy` part ? 🤔

library-solidity/hardhat.config.ts
melanciani: ah I guess it's because it's already in `deployAllHostContracts`

library-solidity/test/asyncDecrypt.ts
```diff
@@ -161,19 +162,31 @@ const fulfillAllPastRequestsIds = async (mocked: boolean) => {
       );
 
       const abiCoder = new ethers.AbiCoder();
-      let encodedData;
-      let decryptedResult;
-
-      encodedData = abiCoder.encode(['uint256', ...types, 'bytes[]'], [31, ...valuesFormatted4, []]); // 31 is just a dummy uint256 requestID to get correct abi encoding for the remaining arguments (i.e everything except the requestID)
+      const encodedData = abiCoder.encode(
+        ['uint256', ...Array(valuesFormatted4.length).fill('uint256'), 'bytes[]'],
```
melanciani: not sure I get the change here

library-solidity/test/asyncDecrypt.ts
```diff
@@ -161,19 +162,31 @@ const fulfillAllPastRequestsIds = async (mocked: boolean) => {
...
+        decryptResultsEIP712signatures,
+      );
+      const decryptionProof = ethers.concat([packedNumSigners, packedSignatures, extraDataV0]);
 
       const calldata =
         callbackSelector +
         abiCoder
-          .encode(['uint256', ...types, 'bytes[]'], [requestID, ...valuesFormatted4, decryptResultsEIP712signatures])
+          .encode(
+            ['uint256', ...Array(valuesFormatted4.length).fill('uint256'), 'bytes'],
```
melanciani: same

melanciani (commented):
a few questions, but overall looks good !

library-solidity/hardhat.config.ts
isaacdecoded: This is because of the npm workspaces config, which installs the `@fhevm/core-contracts` library at the root level of the mono repo.

library-solidity/test/asyncDecrypt.ts
melanciani: ah maybe worth adjusting the comment like it was done in Jat's PR initially ? : https://github.com/zama-ai/fhevm/pull/598/files#diff-140d5a0b10d15f64ba3619a12de0397d55b3d2df6ffd9b080b575c3f0775e672R134

melanciani: ah maybe not, it does not seem you do the same thing

isaacdecoded: It's almost the same, yes, I mean, we are handling all arguments as `uint256`, but we need to keep the last `bytes[]` because this is included in the signatures sent by the KMS. Btw, removing this `bytes[]` caused failures during the e2e tests due to `KMSInvalidSigner` because of the mismatch with the signatures mentioned before. I think we need to tackle this separately in other PR + double-check it with Jat.

library-solidity/lib/FHE.sol
isaacdecoded: I think we don't need to change it because the `calldata` still contains the same information at this level, i.e. `callbackSelector + requestId + decryptedResult + decryptionProof` -> we are copying only the `decryptedResult` bytes 🤔 

cc @immortal-tofu

library-solidity/test/asyncDecrypt.ts
isaacdecoded: We now handle all arguments as `uint256`

library-solidity/lib/FHE.sol
melanciani: ~~I just remember @immortal-tofu saying that we might had to add an abi decoded on `decryptedResult` , don't remember the exact reason~~

it's a change we want to include in fhevm 0.8, I pinged you on slack

library-solidity/test/asyncDecrypt.ts
melanciani: in the end this change was reverted

library-solidity/lib/FHE.sol
melanciani: we'll do this in a follow-up PR for fhevm 0.8 : https://github.com/zama-ai/fhevm-internal/issues/346

melanciani (approved):
thanks !!

PR #685, opened 2025-08-04
fix(test-suite): log level cfg for relayer in e2e test

manoranjith:
- The relayer does not read from configuration file, instead it should be set using RUST_LOG env variable.


antoniupop (approved):
LGTM

PR #694, opened 2025-08-06
fix(coprocessor): host-listener, retry on sql timeout

rudy-6-4:
fix timeout statement detection
always retry even if the error looks non transient, could be a detection bug


rudy-6-4:
Note: this is a backport from release branch

antoniupop (approved):

PR #698, opened 2025-08-07
fix(coprocessor): do not expand  an empty compact ct list

dartdart26:
Current version of tfhe-rs panics when trying to expand and verify an empty compact ciphertext list. We add a check to avoid this if the list is empty.

We consider an empty list to be a valid empty input (with the ZK proof verified in the future when tfhe-rs is updated).

Future versions of tfhe-rs may handle this gracefully, in which case we can remove this check.

Add a test to verify that an empty compact ciphertext list does not cause a panic.

Thanks to @IceTDrinker for proposing this fix.

### DIFF
--- a/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
+++ b/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
@@ -372,6 +372,12 @@ fn try_verify_and_expand_ciphertext_list(
             keys.pks.parameters(), &keys.public_params,
         ))?;
 
+    // TODO: Make sure we don't try to verify and expand an empty list as it would panic with the current version of tfhe-rs.
+    // Could be removed in the future if tfhe-rs is updated to handle empty lists gracefully.
+    if the_list.is_empty() {
+        return Ok(vec![]);
+    }
+
     let expanded: tfhe::CompactCiphertextListExpander = the_list
         .verify_and_expand(&keys.public_params, &keys.pks, &aux_data_bytes)
         .map_err(|err| ExecutionError::InvalidProof(request_id, err.to_string()))?;
--- a/coprocessor/fhevm-engine/Cargo.lock
+++ b/coprocessor/fhevm-engine/Cargo.lock
@@ -8776,6 +8776,7 @@ dependencies = [
  "hex",
  "lru 0.13.0",
  "rand 0.9.1",
+ "serial_test",
  "sha3",
  "sqlx",
  "test-harness",
--- a/coprocessor/fhevm-engine/transaction-sender/src/ops/verify_proof.rs
+++ b/coprocessor/fhevm-engine/transaction-sender/src/ops/verify_proof.rs
@@ -256,10 +256,10 @@ where
                     let handles = row
                         .handles
                         .ok_or(anyhow::anyhow!("handles field is None"))?;
-                    if handles.is_empty() || handles.len() % 32 != 0 {
+                    if handles.len() % 32 != 0 {
                         error!(
                             handles_len = handles.len(),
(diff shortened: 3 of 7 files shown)


coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
```diff
@@ -372,6 +372,12 @@ fn try_verify_and_expand_ciphertext_list(
             keys.pks.parameters(), &keys.public_params,
         ))?;
 
+    // TODO: Make sure we don't try to verify and expand an empty list as it would panic with the current version of tfhe-rs.
+    // Could be removed in the future if tfhe-rs is updated to handle empty lists gracefully.
+    if the_list.is_empty() {
+        return Ok(vec![]);
+    }
```
antoniupop, line 379: I think we still need to update the DB to set the outcome to be verified=false - shouldn't we return a specific error we can then use to update?

antoniupop (changes requested):

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
dartdart26: I am considering an empty list as a valid one, IMO it makes everything easier to reason about. But you are right that the ZK proof is not verified in this case (which would be fixed once tfhe-rs is updated). On the other hand, no handle would be created that can be abused.

In general, I think an empty list should be allowed, it's just that we don't verify any proof now.

dartdart26: What's the benefit of returning an error?

antoniupop: The essential point is that we update the DB so the workers don't get stuck by un-handled proofs that are still selected every time.

dartdart26: My reasoning is that we want to eventually support empty input lists, but if that's not the case, I can change as you suggest. Wdyt?

antoniupop: We basically need to ensure the query doesn't return this anymore. Either have a separate flag for invalid requests or set verified to false, etc.

dartdart26: But the worker itself won't get stuck, it will update verified = true and move on. We never go to the point of the panic this way (expand_and_verify()).

antoniupop: Ok, didn't consider it going true would be fine, but you're right!

dartdart26: > We basically need to ensure the query doesn't return this anymore. Either have a separate flag for invalid requests or set verified to false, etc.

The query won't return it anymore as everything will succeed.

antoniupop (approved):

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
dartdart26: To me the point is whether we want to semantically support empty input lists.

antoniupop: That's probably not very relevant since there's no side effects (that I know of at least) to accepting proof on empty lists. Am I missing something?

antoniupop: Semantically I'd say any proof is valid for proving "nothing" :)

PR #700, opened 2025-08-08
chore(common): update latest rust toolchain

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/338


.github/workflows/golden-container-images-docker-build-rust.yml
```diff
@@ -40,23 +40,34 @@ jobs:
...
+              - toolchain.txt
+  determine-tag:
     permissions:
       actions: 'read'
       contents: 'read'
       pull-requests: 'read'
     runs-on: ubuntu-latest
-    if: github.event_name == 'workflow_dispatch'
     outputs:
-      DOCKER_TAG_IMAGE: ${{ steps.set-tag.outputs.tag }}
```
tawadaa, line 51: I would keep this env var, we don't know where it is used, at least until @fegmorte is back to confirm

.github/workflows/golden-container-images-docker-build-rust.yml
```diff
@@ -40,23 +40,34 @@ jobs:
...
               - 'golden-container-images/rust-glibc/**'
-  set-custom-tag:
+              - toolchain.txt
+  determine-tag:
     permissions:
       actions: 'read'
       contents: 'read'
       pull-requests: 'read'
     runs-on: ubuntu-latest
-    if: github.event_name == 'workflow_dispatch'
```
tawadaa, line 49: I would keep also this one, as it will be launched manually, we need to disable frequent builds, no need,a s we don't change the version that often, wdyt @aquint-zama  @chilcano  ?

tawadaa (changes requested):
Few comments

.github/workflows/golden-container-images-docker-build-rust.yml
eudelins-zama: It is not an env var, it is an output var which was not (or incorrectly) used (I double checked, and I am pretty sure it has no impact to remove it, and I'm in favor of cleaning stuff when we can :slightly_smiling_face:)

.github/workflows/golden-container-images-docker-build-rust.yml
eudelins-zama: Keeping this `if` won't disable the build, it will just build the image with the commit hash instead...

If we want to only enable this workflow to be triggered manually, we should put this `if` on the check-changes job, let me know what you want

.github/workflows/golden-container-images-docker-build-rust.yml
tawadaa: still I'd wait for @fegmorte confirmation

.github/workflows/golden-container-images-docker-build-rust.yml
tawadaa: We want to have control over image built, so not a CI style, I would say, when triggered by:
- manually
- by change in toolchain.txt

tawadaa: @eudelins-zama  let's start by only manual trigger, then we see how to automate, the image upgrade isn't that frequent

.github/workflows/golden-container-images-docker-build-rust.yml
fegmorte: @tawadaa @eudelins-zama we can remove the job `set-custom-tag` because the output is never used. In fact we determine the tag in the common-docker workflow afterward.

.github/workflows/golden-container-images-docker-build-rust.yml
eudelins-zama: I think we are fine. The workflow is only triggered if either the toolchain.txt, the workflow itself or the Dockerfile of the golden image is modified.

We can still disable automatic build and enforce manual build later if really we are facing issues with this workflow

tawadaa (approved):
LGTM, thanks!

PR #701, opened 2025-08-08
feat(kms-connector): healtcheck subcommand

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/339


dartdart26 (approved):
Looks good!

Though I was wondering if we can't use a tool that parses the json and checks the status code and not implement it ourselves? Something like jq?

Or we could put the health check request in a separate binary instead of in all the binaries?

Wdyt?

PR #709, opened 2025-08-12
fix(coprocessor): limit the count of inputs in a single zkproof batch

goshawk-3:
Since the input index is stored as a u8 in the handle bytes, and we reserve 0xFF to mark handles originating from a FHE operation, the maximum valid input index is 254. This means a single batch can contain up to 255 inputs.

ref: https://github.com/zama-ai/fhevm-internal/issues/266


antoniupop (approved):
LGTM!

PR #710, opened 2025-08-12
feat(kms-connector): add db polling backup

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/341

Note: E2E are passing locally, but I can't show it on CI as the input of the workflow does not allow different versions for the connector services, and only two of them have been modified


dartdart26 (approved):
Nice!

PR #716, opened 2025-08-12
feat(coprocessor): add metrics in txn-sender

dartdart26:
Wire the metrics handler into the existing HTTP server under `/metrics`.

Rename the `health-check-port` argument to `http-server-port`, but keep `health-check-port` as an alias for backwards compatibility.


eudelins-zama (approved):
LGTM!

Just some typo Copilot found

antoniupop (approved):

PR #722, opened 2025-08-13
docs(common): changed definitions to public decrypt and user decrypt

poppyseedDev:
It's important that we emphasized in our documentation the differentiation between public decrypt and user decrypt, so that developers know the difference and don't confuse the two concepts together.

### DIFF
--- a/docs/examples/SUMMARY.md
+++ b/docs/examples/SUMMARY.md
@@ -7,12 +7,12 @@
 - Encryption
   - [Encrypt single value](fhe-encrypt-single-value.md)
   - [Encrypt multiple values](fhe-encrypt-multiple-values.md)
-- Decryption
-  - [Decrypt single value](fhe-decrypt-single-value.md)
-  - [Decrypt multiple values](fhe-decrypt-multiple-values.md)
-- Decryption in Solidity
-  - [Decrypt single value in Solidity](fhe-decrypt-single-value-in-solidity.md)
-  - [Decrypt multiple values in Solidity](fhe-decrypt-multiple-values-in-solidity.md)
+- User Decryption
+  - [User decrypt single value](fhe-user-decrypt-single-value.md)
+  - [User decrypt multiple values](fhe-user-decrypt-multiple-values.md)
+- Public Decryption
+  - [Public Decrypt single value](fhe-public-decrypt-single-value.md)
+  - [Public Decrypt multiple values](fhe-public-decrypt-multiple-values.md)
 
 ## Advanced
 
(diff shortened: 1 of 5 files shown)


docs/examples/SUMMARY.md
```diff
@@ -7,12 +7,12 @@
...
-  - [Decrypt multiple values](fhe-decrypt-multiple-values.md)
-- Decryption in Solidity
-  - [Decrypt single value in Solidity](fhe-decrypt-single-value-in-solidity.md)
-  - [Decrypt multiple values in Solidity](fhe-decrypt-multiple-values-in-solidity.md)
+- User Decryption
+  - [User decrypt single value](fhe-user-decrypt-single-value.md)
+  - [User decrypt multiple values](fhe-user-decrypt-multiple-values.md)
+- Public Decryption
+  - [Public Decrypt single value](fhe-public-decrypt-single-value.md)
+  - [Public Decrypt multiple values](fhe-public-decrypt-multiple-values.md)
```
yuxizama, line 14: I would suggest wrapping "User Decryption" and "Public Decryption" under a section called "Decryption" to be aligned with "Encryption" on the side menu.

docs/examples/fhe-public-decrypt-multiple-values.md
```diff
@@ -1,4 +1,6 @@
-This example demonstrates the FHE decryption mechanism in Solidity with multiple values.
+This example demonstrates the FHE public decryption mechanism with multiple value.
+
+Public decryption is a mechanism that makes encrypted values visible to everyone once decrypted. Unlike user decryption where values remain private to authorized users, public decryption makes the data permanently visible to all participants. The public decryption call occurs on-chain through smart contracts, making the decrypted value part of the blockchain's public state.
```
yuxizama: Be sure to write "onchain" instead of "on-chain" in our docs.

yuxizama (changes requested):

docs/examples/SUMMARY.md
poppyseedDev: Did you have something like this in mind? :
 - Decryption
    - Public Decryption
       - [Public Decrypt single value]
       - [Public Decrypt single value]
    - User Decryption
       - [User Decrypt single value]
       - [Public Decrypt single value]

poppyseedDev: Or:

 - Decryption
   - [Public Decrypt single value]
   - [Public Decrypt single value]
   - [User Decrypt single value]
   - [Public Decrypt single value]

yuxizama: Since there are not so many pages, let's not wrap another layer. So like:
Decryption
- [Public Decrypt single value]
- [Public Decrypt multiple values]
- [User Decrypt single value]
- [User Decrypt multiple values]
would be fine!

yuxizama (approved):

PR #731, opened 2025-08-14
chore(library-solidity): consider ABI encoded results in callback inputs

melanciani:
docker images : 
- test-suite: `357a40a` from [this build](https://github.com/zama-ai/fhevm/actions/runs/17074921971) from this current PR (previous build)
- relayer: `ec8eda4` from [this build](https://github.com/zama-ai/console/actions/runs/17075000917) from [this PR](https://github.com/zama-ai/console/pull/524)

e2e : https://github.com/zama-ai/fhevm/actions/runs/17075729809/job/48416522304?pr=731

closes https://github.com/zama-ai/fhevm-internal/issues/346


docs/examples/sealed-bid-auction-tutorial.md
```diff
@@ -232,7 +232,7 @@ function decryptWinningAddress() public onlyAfterEnd {
 }
 ```
 
-Here, we are requesting to decrypt a single parameter for the `winningAddress`. However, you can request multiple ones by increasing the `cts` array and adding other parameters.
+Here, we are requesting to decrypt a single parameter for the `winningAddress`, ABI encoded as a byte array (using `abi.encode`). You can request multiple ones by including more parameters in the ABI encoding.
```
jatZama: I think this comment part is false "You can request multiple ones by including more parameters in the ABI encoding."I think original sentence was better "ou can request multiple ones by increasing the `cts` array and adding other parameters."

docs/solidity-guides/decryption/oracle.md
```diff
@@ -83,7 +85,7 @@ The first argument, `ctsHandles`, should be an array of ciphertexts handles whic
 `callbackSelector` is the function selector of the callback function, which will be called once the relayer fulfils the decryption request.
 
 ```solidity
-function [callbackName](uint256 requestID, XXX x_0, XXX x_1, ..., XXX x_N-1, bytes[] memory signatures) external;
+function [callbackName](uint256 requestID, bytes memory cleartexts, bytes memory decryptionProof) external;
 ```
 
 Notice that `XXX` should be the decrypted type, which is a native Solidity type corresponding to the original ciphertext type, following this table of conventions:
```
jatZama: This sentence is obsolete, maybe replace "Notice that `XXX` should be the decrypted type, which is a native Solidity type corresponding to the original ciphertext type, following this table of conventions:" by something like `"cleartexts" is the bytes array corresponding to the abi encoding of all requested decrypted values.`

host-contracts/codegen/templates.ts
```diff
@@ -1127,39 +1127,75 @@ function generateSolidityDecryptionOracleMethods(fheTypes: AdjustedFheType[]): s
...
+                decryptionProof
+            );
     }
 
     /**
-     * @dev Private low-level function used to compute the length of the decryptedResult bytes array.
+     * @dev Private low-level function used to verify the handle types.
      */
-    function getSignedDataLength(bytes32[] memory handlesList) private pure returns (uint256) {
+    function verifyHandleTypes(bytes32[] memory handlesList) private pure {
```
jatZama: I think this function could be removed.

jatZama: meaning the custom error UnsupportedHandleType could also be removed. This is mainly due to ebytes legacy.

host-contracts/test/asyncDecrypt.ts
```diff
@@ -124,16 +125,30 @@ const fulfillAllPastRequestsIds = async (mocked: boolean) => {
...
 
+      // ABI encode the decryptedResult as done in the KMS, following the format:
+      // - requestId (32 bytes)
+      // - all inputs
+      // - list of signatures (list of bytes)
+      // For this we use the following values for getting the correct abi encoding (in particular for
+      // getting the right signatures offset right after):
+      // - requestId: a dummy uint256
+      // - signatures: a dummy empty array of bytes
       const encodedData = abiCoder.encode(
```
jatZama, line 140: you can update this part by removing the dummy first and last values. but this could wait removal of signatures offset as well.

docs/examples/sealed-bid-auction-tutorial.md
melanciani: I wanted to focus on the fact that it's not just having an array, you have to ABI encode this array

host-contracts/codegen/templates.ts
melanciani: ah so we don't need to check the fhe type anymore ?

host-contracts/test/asyncDecrypt.ts
melanciani: yes let's do this along https://github.com/zama-ai/fhevm-internal/issues/345 , because for now we need to consider the offset if we want e2e tests to pass with current kms core release version

docs/examples/sealed-bid-auction-tutorial.md
melanciani: ah I get it, this comment refers to the function above my bad I revert

jatZama (approved):
Everything looks good here.

PR #736, opened 2025-08-15
chore(coprocessor): remove Deref in NonceManagedProvider

dartdart26:
We don't actually need the Deref implementation, as we can just call `self.provider.inner()` directly. That way we don't introduce unclear semantics.


eudelins-zama (approved):
Approving the PR, but warning: everywhere we provide the inner provider, the custom `send_transaction` method of the `NonceManagedProvider` with nonce recovery won't be used. I think it is OK right now, but might be harder to maintain and confusing for future devs, to understand when to use the inner or not.

Maybe we could implement the `Provider` trait manually for the `NonceManagedProvider` like in the kms-connector, and pass the provider itself instead of the inner one everywhere (could be another PR, or we can discuss it ofc)

PR #743, opened 2025-08-18
feat(library-solidity): add support for ERC-7995

0xalexbel:
closes https://github.com/zama-ai/fhevm-internal/issues/283
closes https://github.com/zama-ai/fhevm-internal/issues/284

e2e: https://github.com/zama-ai/fhevm/actions/runs/17074365466/job/48411662857?pr=743

docker images: 
- test-suite: `2b51479` from [this build](https://github.com/zama-ai/fhevm/actions/runs/17073333611) from this current PR (previous build)


host-contracts/lib/CoprocessorConfig.sol
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * @notice This function returns a struct containing all contract addresses.
+     * @dev    It returns an immutable struct.
+     */
+    function defaultConfig() internal pure returns (CoprocessorConfigStruct memory) {
+        return
+            CoprocessorConfigStruct({
+                ACLAddress: aclAdd,
+                CoprocessorAddress: fhevmExecutorAdd,
+                DecryptionOracleAddress: DECRYPTION_ORACLE_ADDRESS,
```
jatZama: is it possible please to rename this DECRYPTION_ORACLE_ADDRESS variable with same convention as the other ones? i.e decryptionOracleAdd

host-contracts/lib/Impl.sol
```diff
@@ -350,19 +354,20 @@ interface IInputVerifier {
...
  */
 library Impl {
-    /// keccak256(abi.encode(uint256(keccak256("fhevm.storage.FHEVMConfig")) - 1)) & ~bytes32(uint256(0xff))
-    bytes32 private constant FHEVMConfigLocation = 0xed8d60e34876f751cc8b014c560745351147d9de11b9347c854e881b128ea600;
+    /// keccak256(abi.encode(uint256(keccak256("confidential.storage.config")) - 1)) & ~bytes32(uint256(0xff))
+    bytes32 private constant CoprocessorConfigLocation =
+        0x9e7b61f58c47dc699ac88507c4f5bb9f121c03808c5676a8078fe583e4649700;
 
     /// keccak256(abi.encode(uint256(keccak256("fhevm.storage.DecryptionRequests")) - 1)) & ~bytes32(uint256(0xff))
     bytes32 private constant DecryptionRequestsStorageLocation =
```
jatZama, line 362: why did you keep this constant? I think you should remove it.

host-contracts/lib/Impl.sol
```diff
@@ -4,24 +4,22 @@ pragma solidity ^0.8.24;
 import {FheType} from "../contracts/shared/FheType.sol";
 
 /**
- * @title   FHEVMConfigStruct
+ * @title   CoprocessorConfigStruct
  * @notice  This struct contains all addresses of core contracts, which are needed in a typical dApp.
  */
-struct FHEVMConfigStruct {
```
jatZama, line 10: Do we really need to keep the Struct in the name of the struct? I think renaming `CoprocessorConfigStruct` to `CoprocessorConfig` is better, wdyt?

host-contracts/examples/tests/FHEVMManualTestSuite.sol
```diff
@@ -15,7 +15,7 @@ contract FHEVMManualTestSuite {
     eaddress public resAdd;
 
     constructor() {
-        FHE.setCoprocessor(FHEVMConfig.defaultConfig()); // Set up the FHEVM configuration for this contract
+        FHE.setCoprocessor(CoprocessorConfig.defaultConfig()); // Set up the FHEVM configuration for this contract
```
jatZama: Can you update the comment as well by replacing FHEVM to Coprocessor?

host-contracts/lib/Impl.sol
```diff
@@ -278,6 +276,12 @@ interface IFHEVMExecutor {
      * @return result       Result.
      */
     function fheRandBounded(uint256 upperBound, FheType randType) external returns (bytes32 result);
+
+    /**
+     * @notice                      Returns the address of the InputVerifier contract used by the coprocessor.
+     * @return inputVerifierAddress Address of the InputVerifier.
+     */
+    function getInputVerifierAddress() external view returns (address);
```
jatZama, line 284: I am not sure why this getter is added, maybe you could remove it if not used anywhere?

host-contracts/codegen/templates.ts
```diff
@@ -134,18 +134,18 @@ ${generateInputVerifierInterface()}
  * @notice  This library is the core implementation for computing FHE operations (e.g. add, sub, xor).
  */
 library Impl {
-  /// keccak256(abi.encode(uint256(keccak256("fhevm.storage.FHEVMConfig")) - 1)) & ~bytes32(uint256(0xff))
-  bytes32 private constant FHEVMConfigLocation = 0xed8d60e34876f751cc8b014c560745351147d9de11b9347c854e881b128ea600;
+  /// keccak256(abi.encode(uint256(keccak256("confidential.storage.config")) - 1)) & ~bytes32(uint256(0xff))
+  bytes32 private constant CoprocessorConfigLocation = 0x9e7b61f58c47dc699ac88507c4f5bb9f121c03808c5676a8078fe583e4649700;
 
   /// keccak256(abi.encode(uint256(keccak256("fhevm.storage.DecryptionRequests")) - 1)) & ~bytes32(uint256(0xff))
```
jatZama: Please remove this constant as it is no longer needed

host-contracts/codegen/templates.ts
```diff
@@ -203,24 +194,22 @@ function generateImplCoprocessorInterface(operators: Operator[]): string {
...
     }
 
     /**
      * @title   DecryptionRequestsStruct
-     * @notice  This struct contains the address of the decryption oracle contract, 
-     *          the internal counter for requestIDs generated by the dapp, 
+     * @notice  This struct contains the internal counter for requestIDs generated by the dapp, 
      *          and the mapping from internal requestIDs to list of handles requested for decryption.
      */
     struct DecryptionRequestsStruct {
```
jatZama: Maybe here also remove the Struct from the name : i.e rename it to DecryptionRequests

library-solidity/lib/FHE.sol
```diff
@@ -8978,7 +8962,7 @@ library FHE {
                 revert UnsupportedHandleType();
             }
         }
-        signedDataLength += 32; // add offset of signatures
+        signedDataLength += 32; // add offset of the first parameter in the ABI-encoded data
```
jatZama: this comment is incorrect, keep the original one (although it should be deprecated soon)

library-solidity/config/ZamaConfig.sol
```diff
@@ -3,38 +3,39 @@ pragma solidity ^0.8.24;
...
  * @notice  This library returns the FHEVM config for different networks
- *          with the contract addresses for (1) ACL, (2) FHEVMExecutor, (3) KMSVerifier, (4) InputVerifier
+ *          with the contract addresses for (1) ACL, (2) CoprocessorAddress, (3) DecryptionOracleAddress, (4) KMSVerifier,
  *          which are deployed & maintained by Zama. It also returns the address of the decryption oracle.
  */
 library ZamaConfig {
-    function getSepoliaConfig() internal pure returns (FHEVMConfigStruct memory) {
+    function getProtocolId() internal pure returns (uint256) {
+        /// @note '1' is Zama's ERC7995 protocol id
+        return 1;
```
jatZama: I really think this is dangerous, the `1` should be reserved to mainnet, not sepolia imo.

jatZama: not sure which value should be used for sepolia

unless we consider that protocol id is independent from the chain? is this relative only to one protocol on many chains?

maybe i misunderstood the definition of protocol id

immortal-tofu: `protocolId` is not related to the host chain. eg Zama Protocol contracts would return `1`, INCO would return `2`, ...
No matter the chain, it must returns `1` if the contract is using the Zama Protocol.

jatZama: ok but 1 should be used for mainnet, are we sure it makes sense to use 1 on devnet as well?

immortal-tofu: No strong opinion on this. We can use a different for testnet.

host-contracts/lib/CoprocessorConfig.sol
0xalexbel: Good idea!

host-contracts/lib/Impl.sol
0xalexbel: Why not!

host-contracts/examples/tests/FHEVMManualTestSuite.sol
0xalexbel: Good catch!

host-contracts/lib/Impl.sol
0xalexbel: originally used by Impl.cleanTransientStorageInputVerifier(). I had to find a way to port the function. @immortal-tofu  suggested to remove it. I was not sure. WDYT ?

host-contracts/codegen/templates.ts
0xalexbel: Do I miss something ? I simply renamed existing constants. 

constant is used here in `Impl.sol`
```
    function getCoprocessorConfig() internal pure returns (CoprocessorConfigStruct storage $) {
        assembly {
            $.slot := CoprocessorConfigLocation
        }
    }
```

host-contracts/codegen/templates.ts
0xalexbel: ok

library-solidity/lib/FHE.sol
0xalexbel: ok, may be you should also notify @isaacdecoded and @manoranjith since this comment is coming from their code

host-contracts/codegen/templates.ts
jatZama: I am talking about the obsolete constant `DecryptionRequestsStorageLocation` which you should remove.

library-solidity/lib/FHE.sol
jatZama: If you want then you can keep this for now, but this will be outdated very soon anyways.

host-contracts/lib/Impl.sol
0xalexbel: Still required

host-contracts/codegen/templates.ts
jatZama: My bad, i thought this was just used for the oracle contract, then you can keep it. But in this case I would like you maybe to change the preimage to remove mention of fhevm? since you remove fhevm from the coprocessor config. Maybe something like `"confidential.storage.decryptionRequests"`

0xalexbel: ok

library-solidity/lib/FHE.sol
0xalexbel: I finally kept the original one

host-contracts/codegen/templates.ts
0xalexbel: All Struct prefixes have been removed!

host-contracts/lib/Impl.sol
0xalexbel: All Struct prefixes have been removed

host-contracts/lib/CoprocessorConfig.sol
0xalexbel: DECRYPTION_ORACLE_ADDRESS renamed to decryptionOracleAdd

host-contracts/lib/Impl.sol
jatZama: I unresolved this one because you forgot to remove this function, it should be deleted because it is not used anywhere.

jatZama (approved):
I just unresolved a minor issue but because this is urgent to be merged you can merge as is and remove the unused function from the interface later if you prefer.

host-contracts/lib/Impl.sol
melanciani: since it's still used in `cleanTransientStorageInputVerifier`, we don't remove it

PR #744, opened 2025-08-18
fix(coprocessor): host-listener, more reliable block time estimation

rudy-6-4:


coprocessor/fhevm-engine/host-listener/src/cmd/block_history.rs
```diff
@@ -78,4 +79,20 @@ impl BlockHistory {
         }
         self.ordered_blocks.push_back(block);
     }
+
+    pub fn estimated_block_time(&self) -> u64 {
```
rudy-6-4: use Option to keep the previous value if not available

coprocessor/fhevm-engine/host-listener/src/cmd/block_history.rs
```diff
@@ -78,4 +79,112 @@ impl BlockHistory {
...
+            self.ordered_blocks.get(self.ordered_blocks.len() - 2)?;
+        if last.timestamp <= second_last.timestamp {
+            return None;
+        }
+        if last.number <= second_last.number {
+            return None;
+        }
+        let estimation = (last.timestamp - second_last.timestamp) as f64
+            / (last.number - second_last.number) as f64;
+        Some((estimation.round() as u64).max(MINIMUM_BLOCK_TIME_SECONDS))
```
dartdart26, line 98: Nit: Just wondering if we can avoid the f64 arithmetic altogether?

dartdart26 (approved):
Looks good!

coprocessor/fhevm-engine/host-listener/src/cmd/block_history.rs
rudy-6-4: Probably by adding 1/2 before division.
e.g.  (last.number - second_last.number) / 2
What would be the benefit ? It's probably not faster.

PR #750, opened 2025-08-19
fix(coprocessor): host-listener, missings blocks, missed the last one

rudy-6-4:


rudy-6-4:
Note: it's expected to not put the first current block as it's the current one and is not missing

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -462,6 +461,7 @@ impl InfiniteLogIter {
                 break;
             };
             current_block = parent_block.into();
+            missing_blocks.push(current_block);
```
goshawk-3, line 464: Now, the missing_blocks will include all ancestors starting from the first parent of the initial current block. However, the initial current block itself won’t be added — is that the intended behavior?

goshawk-3: Ok, I just saw your post here https://github.com/zama-ai/fhevm/pull/750#issuecomment-3200920441

goshawk-3 (approved):
LGTM

PR #752, opened 2025-08-19
fix(coprocessor): host-listener, recheck on prev block broken by reorg handling

rudy-6-4:
this could prevent the recheck when no events

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
@@ -307,12 +307,13 @@ impl InfiniteLogIter {
         }
     }
 
-    async fn recheck_block(&mut self, block: Option<BlockSummary>) -> bool {
+    async fn recheck_tip_block(&mut self) {
+        let block = self.block_history.tip();
         if self.no_block_immediate_recheck {
-            return false;
+            return;
         }
         let Some(block) = block else {
-            return false;
+            return;
         };
         let logs = match self.get_logs_at_hash(block.hash).await {
             Ok(logs) => logs,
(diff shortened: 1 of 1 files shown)


coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -714,9 +714,10 @@ impl InfiniteLogIter {
                     if is_first_of_block {
                         self.tick_block.update();
                     }
+                    // get prev_block before it's updated by check_missing_ancestors
+                    let prev_block = self.block_history.tip();
```
goshawk-3: If so, why don't you do the `self.recheck_block(prev_block)` immediately after this?

rudy-6-4: It was to retain order of event in the catchup vec but you are righ, let's put it before.

rudy-6-4: Done, also make explicit that check_missing_ancestor is only worth it in case of new block.

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -706,37 +704,33 @@ impl InfiniteLogIter {
...
                     self.current_event = Some(log);
                     if is_first_of_block {
                         self.tick_block.update();
+                        self.recheck_tip_block().await;
+                        self.check_missing_ancestors(block_hash).await;
+                        if self.current_event.is_none() {
+                            // current log has been pushed in catchup_logs
+                            // after events from previous block (recheck or reorg)
+                            continue; // jump to process events from catchup_logs
+                        }
```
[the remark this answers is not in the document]
rudy-6-4: modified by called function

goshawk-3 (approved):
LGTM

PR #759, opened 2025-08-20
fix(sdk): add extra_data for eip712

eudelins-zama:


leventdem (approved):
We needed to add extraData for the next release, well done!

PR #766, opened 2025-08-21
chore(coprocessor): aws-lc-rs in txn-sender and gw-listener

dartdart26:
Use the `aws-lc-rs` rustls provider in txn-sender and gw-listener.


goshawk-3 (approved):
LGTM

PR #771, opened 2025-08-22
feat(host-contracts): update operatorsPrices and HCULimit contract

jatZama:


bcm-at-zama (commented):
I am just curious: (speaking with my C language, sorry, I don't know the wording in the blockchain world)

> would it be a bad idea to have all those constants in a kind of .h header file, with #define's?

Then, in the future:
- we might easily just update the magic values in the .h
- and we could even imagine this .h is automatically generated / updated from some benchmark scripts that we have

jatZama:
> I am just curious: (speaking with my C language, sorry, I don't know the wording in the blockchain world)
> 
> > would it be a bad idea to have all those constants in a kind of .h header file, with #define's?
> 
> Then, in the future:
> 
> * we might easily just update the magic values in the .h
> * and we could even imagine this .h is automatically generated / updated from some benchmark scripts that we have

There is no header file in solidity, but we could put constants in a dedicated solidity file, but the issue is that we also need the json file for the utility functions for the mock mode, currently the codegen is doing the automatic conversion. Not sure what you want to change here.

bcm-at-zama:
> codegen is doing the automatic conversion

Maybe it was what I was missed here. If you are able to change at a single place and there is something which populates everything as a .h would do, then yes, all good.

jatZama:
> > codegen is doing the automatic conversion
> 
> Maybe it was what I was missed here. If you are able to change at a single place and there is something which populates everything as a .h would do, then yes, all good.

Indeed the HCULimit contract is not updated by hand, since this would be extremely error prone.

melanciani (approved):
thanks !

PR #772, opened 2025-08-22
feat(gateway-contracts): implement single keygen

melanciani:
this PR properly implements a way to trigger keygen through the gateway. It includes: 
- new keygen spec: https://github.com/zama-ai/fhevm-internal/issues/304  
    - + add s3BucketUrl for KMS nodes
- new FHE params spec: https://github.com/zama-ai/fhevm-internal/issues/325
- add CRS params type: https://github.com/zama-ai/fhevm-internal/issues/177
- generate requestIds in gateway: https://github.com/zama-ai/fhevm-internal/issues/186
- make pre-process call keygen request directly: https://github.com/zama-ai/fhevm-internal/issues/182
- making the new requestIds unique for KMS (following a specific map: https://github.com/zama-ai/fhevm-internal/issues/199
- use 2/3+1 threshold for consensus : https://github.com/zama-ai/fhevm-internal/issues/158

TODO: 
- update KmsManagement tests
- maybe: add hardhat task to trigger keygen from owner
- add `extraData` field in functions, structs and events

also created these issues for later  : 
- https://github.com/zama-ai/fhevm-internal/issues/365 (along key lifecycles)
- https://github.com/zama-ai/fhevm-internal/issues/377
- https://github.com/zama-ai/fhevm-internal/issues/376

closes https://github.com/zama-ai/fhevm-internal/issues/304  
closes https://github.com/zama-ai/fhevm-internal/issues/325
closes https://github.com/zama-ai/fhevm-internal/issues/177
closes https://github.com/zama-ai/fhevm-internal/issues/186
closes https://github.com/zama-ai/fhevm-internal/issues/182
closes https://github.com/zama-ai/fhevm-internal/issues/199
closes https://github.com/zama-ai/fhevm-internal/issues/158
closes https://github.com/zama-ai/fhevm-internal/issues/180


gateway-contracts/contracts/interfaces/IKmsManagement.sol
```diff
@@ -3,372 +3,191 @@ pragma solidity ^0.8.24;
...
-     * @notice Emitted when the key generation preprocessing is completed.
-     * @param preKeygenRequestId The ID of the preprocessed key generation request.
-     * @param preKeyId The ID of the preprocessed key.
+     * @notice The type of the generated key.
      */
-    event PreprocessKeygenResponse(uint256 preKeygenRequestId, uint256 preKeyId);
+    enum KeyType {
+        SERVER, // 0
+        PUBLIC // 1
+    }
```
eudelins-zama, line 25: Does not respect solidity style : https://docs.soliditylang.org/en/latest/style-guide.html#enums

Should be `Server` and `Public`

melanciani: of course ! good catch

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -185,26 +216,48 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
+                v3UpgradeInputs[i].s3BucketUrl
+            );
+
+            $.kmsNodesV2[kmsTxSenderAddress] = kmsNodeV2;
+            kmsNodesV1[i] = kmsNodeV1;
+            kmsNodesV2[i] = kmsNodeV2;
         }
 
-        emit ReinitializeGatewayConfigV2(custodians);
+        emit ReinitializeGatewayConfigV3(kmsNodesV1, kmsNodesV2);
```
isaacdecoded, line 269: Is this important to include the `kmsNodesV1` in the event? I mean, the `kmsNodesV2` already had the information plus the new field `s3BucketUrl`

gateway-contracts/contracts/KmsManagement.sol
```diff
@@ -543,26 +583,103 @@ contract KmsManagement is
...
+     * @return The hash of the KeygenVerification struct
+     */
+    function _hashKeygenVerification(
+        uint256 prepKeygenId,
+        uint256 keyId,
+        KeyDigest[] calldata keyDigests
+    ) internal view virtual returns (bytes32) {
+        return
+            _hashTypedDataV4(
+                keccak256(abi.encode(EIP712_KEYGEN_TYPE_HASH, prepKeygenId, keyId, keccak256(abi.encode(keyDigests))))
```
isaacdecoded: I think for arrays and bytes we should use `abi.encodePacked` instead to get the packed value without padding 🤔 i.e. `keccak256(abi.encodePacked(keyDigests)) `

gateway-contracts/contracts/KmsManagement.sol
```diff
@@ -543,26 +583,103 @@ contract KmsManagement is
...
+     * @param crsDigest The digest of the generated CRS.
+     * @return The hash of the CrsgenVerification struct
+     */
+    function _hashCrsgenVerification(
+        uint256 crsId,
+        uint256 maxBitLength,
+        bytes calldata crsDigest
+    ) internal view virtual returns (bytes32) {
+        return
+            _hashTypedDataV4(keccak256(abi.encode(EIP712_CRSGEN_TYPE_HASH, crsId, maxBitLength, keccak256(crsDigest))));
```
isaacdecoded: Same as [here](https://github.com/zama-ai/fhevm/pull/772/files#r2314367492) for `crsDigest`, i.e. `keccak256(abi.encodePacked(crsDigest))`

gateway-contracts/contracts/KmsManagement.sol
```diff
@@ -1,126 +1,237 @@
...
     IKmsManagement,
+    EIP712Upgradeable,
     Ownable2StepUpgradeable,
     UUPSUpgradeableEmptyProxy,
     GatewayConfigChecks,
     Pausable
 {
-    /// @notice The address of the GatewayConfig contract for protocol state calls.
+    // ----------------------------------------------------------------------------------------------
+    // EIP712 utility constants:
```
isaacdecoded, line 30: Looking at this, I'm wondering if we should "simplify" the other contracts using EIP712 by removing the struct definitions based on each EIP712 payload, e.g., [UserDecryptResponseVerification](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L66-L75), [DelegatedUserDecryptRequestVerification](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L80-L95), etc. cc @RomanBredehoft

gateway-contracts/contracts/GatewayConfig.sol
melanciani: I included it mostly for better tracking, as we should be 100% sure we haven't misconfigured the new nodes vs the old ones

gateway-contracts/contracts/KmsManagement.sol
melanciani: yes most likely, feel free to update this then !

gateway-contracts/contracts/KmsManagement.sol
melanciani: same

gateway-contracts/contracts/KmsManagement.sol
melanciani: yes, I've already done it for decryption contract in the copro PR: https://github.com/zama-ai/fhevm/pull/211/files#diff-11acb2d556c4a634fdde4f84cc97005bbfe910dfb55f1e811ab08856ba711a2a

(mainly because I needed to remove some local vars following a stack too deep error)

let's do this for all contracts:  https://github.com/zama-ai/fhevm-internal/issues/403

melanciani: actually context pr will also remove the input verification one : https://github.com/zama-ai/fhevm/pull/211/files#diff-b9560385508b39fa02a1d6a41cba7948e7602970fc9824b944049dc587ee6ea0

so in the end not sure the issue is relevant, but let's keep it open until we merge both PRs

gateway-contracts/contracts/KmsManagement.sol
```diff
@@ -41,14 +41,25 @@ contract KmsManagement is
...
     bytes32 private constant EIP712_PREP_KEYGEN_TYPE_HASH = keccak256(bytes(EIP712_PREP_KEYGEN_TYPE));
 
+    /**
+     * @notice The EIP-712 type definition for the KeyDigest struct.
+     * @dev Required because EIP-712 mandates that each nested struct type
+     *      used in a primary type (e.g. KeygenVerification) must be explicitly
+     *      declared with its own type string and type hash.
+     *      These constants are used when computing the struct hash of each
+     *      KeyDigest element inside the keyDigests[] array.
+     */
```
melanciani, line 53: could you add comment for fields like done for other structs ? ex below : 
```
    /**
     * @notice The KeygenVerification typed definition.
     * @dev prepKeygenId: The ID of the preprocessing keygen request.
     * @dev keyId: The ID of the generated key.
     * @dev keyDigests: The digests of the generated key.
     */
```

gateway-contracts/contracts/KmsManagement.sol
```diff
@@ -41,14 +41,25 @@ contract KmsManagement is
...
+
     /**
      * @notice The KeygenVerification typed definition.
      * @dev prepKeygenId: The ID of the preprocessing keygen request.
      * @dev keyId: The ID of the generated key.
      * @dev keyDigests: The digests of the generated key.
      */
     string private constant EIP712_KEYGEN_TYPE =
-        "KeygenVerification(uint256 prepKeygenId,uint256 keyId,(uint8,bytes)[] keyDigests)";
+        "KeygenVerification(uint256 prepKeygenId,uint256 keyId,KeyDigest[] keyDigests)KeyDigest(uint8 keyType,bytes digest)";
```
melanciani, line 64: still surprised by the format of the type, make sure this aligns with KMS core !

gateway-contracts/contracts/Decryption.sol
```diff
@@ -293,8 +319,15 @@ contract Decryption is
     ) external virtual onlyKmsTxSender {
         DecryptionStorage storage $ = _getDecryptionStorage();
 
-        // Make sure the decryptionId corresponds to a generated decryption request.
-        if (decryptionId > $._decryptionRequestCounter || decryptionId == 0) {
+        // Make sure the decryptionId corresponds to a generated public decryption request.
+        // TODO: The 2nd condition is a temporary check to support on-going consensus when the
+        // contract is upgraded. We should remove it in the next version
+        // See https://github.com/zama-ai/fhevm-internal/issues/377
```
melanciani, line 326: add `DEPRECATED: remove for mainnet` comment

gateway-contracts/contracts/shared/KmsRequestCounters.sol
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// Bit position to left shift for initializing the counters
+uint256 constant REQUEST_TYPE_SHIFT = 248;
+
+// Define the counters' initial values in order to generate globally unique requestIds per request type
+// for the KMS
+
+// Public decrypt requestId format in bytes: [0000 0001 | counter_1..31]
+uint256 constant PUBLIC_DECRYPT_COUNTER_BASE = uint256(RequestType.PublicDecrypt) << REQUEST_TYPE_SHIFT;
```
melanciani, line 24: check with coprocessor and fhevm-relayer if they handle uint256 request IDs

gateway-contracts/contracts/Decryption.sol
```diff
@@ -160,12 +161,12 @@ contract Decryption is
...
+    uint256 private constant MINOR_VERSION = 4;
     uint256 private constant PATCH_VERSION = 0;
 
     /// Constant used for making sure the version number using in the `reinitializer` modifier is
     /// identical between `initializeFromEmptyProxy` and the reinitializeVX` method
-    uint64 private constant REINITIALIZER_VERSION = 5;
+    uint64 private constant REINITIALIZER_VERSION = 6;
 
     /// @notice The contract's variable storage struct (@dev see ERC-7201)
     /// @custom:storage-location erc7201:fhevm_gateway.storage.Decryption
```
melanciani, line 172: add deprecated comment for `_decryptionRequestCounter` below (like `/// @notice DEPRECATED: to remove for mainnet`)

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -361,16 +414,22 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         return $.userDecryptionThreshold;
     }
 
+    /// @dev See {IGatewayConfig-getKmsStrongMajorityThreshold}.
+    function getKmsStrongMajorityThreshold() external view virtual returns (uint256) {
```
melanciani: is it ok if this threshold is dynamic ? cc @jot2re @kc1212  (or fixed like like user/public decypytion above)

gateway-contracts/.env.example
```diff
@@ -36,21 +36,25 @@ NUM_KMS_NODES="4" # (number)
 KMS_TX_SENDER_ADDRESS_0="0xc1d91b49A1B3D1324E93F86778C44a03f1063f1b" # accounts[3] (address)
 KMS_SIGNER_ADDRESS_0="0x305F1F471e9baCFF2b3549F9601f9A4BEafc94e1" # accounts[4] (address)
 KMS_NODE_IP_ADDRESS_0="127.0.0.1" # (string)
+KMS_NODE_S3_BUCKET_URL_0="s3://kms-bucket-1" # (string)
```
melanciani, line 42: do not forget to update the wiki with new vars

gateway-contracts/contracts/KmsManagement.sol
```diff
@@ -1,126 +1,248 @@
...
+    // ----------------------------------------------------------------------------------------------
+    // Contract storage:
+    // ----------------------------------------------------------------------------------------------
+
+    /**
+     * @notice The contract's variable storage struct (@dev see ERC-7201)
+     */
     /// @custom:storage-location erc7201:fhevm_gateway.storage.KmsManagement
     struct KmsManagementStorage {
-        /// @notice The FHE params digest tied to a given FHE params name
```
melanciani, line 42: see with infra if it's fine to deploy it again, instead of upgrade

if so, remove all deprecated state var

gateway-contracts/contracts/shared/Structs.sol
```diff
@@ -95,3 +97,12 @@ struct DelegationAccounts {
     /// @notice The address of the account that receives the delegation.
     address delegatedAddress;
 }
+
+// ----------------------------------------------------------------------------------------------
+// Deprecated structs:
```
melanciani, line 103: add deprecated comment (same as other)

gateway-contracts/contracts/shared/Structs.sol
```diff
@@ -10,13 +10,15 @@ struct ProtocolMetadata {
 }
 
 /// @notice Struct that represents a KMS (Key Management Service) node
-struct KmsNode {
+struct KmsNodeV2 {
```
melanciani, line 14: add comment to say we need to rename to `KmsNode` for mainnet

gateway-contracts/contracts/GatewayConfig.sol
kc1212: what is this threshold? is it keygen?

if it's keygen then the threshold should work the same way as public decryption since the result of keygen should be the same for ever honest party

jot2re: @RomanBredehoft what do you mean with dynamic? You mean if we can change it later on?

gateway-contracts/contracts/shared/KmsRequestCounters.sol
isaacdecoded: Coprocessor and Relayer issues updated [here](https://github.com/zama-ai/planning-blockchain/issues/820#issue-3363820431) and [here](https://github.com/zama-ai/console/issues/539#issue-3363701135) after checking this with them.

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: @jot2re afaiu, dynamic here refers to that this `getKmsStrongMajorityThreshold` function is calculating the threshold based on the registered list of `kmsSignerAddresses` (i.e., `($.kmsSignerAddresses.length * 2) / 3 + 1;`) while the other ones like `getUserDecryptionThreshold` return the fixed state value (i.e., `$.userDecryptionThreshold`)

kc1212: I think if you have a configured threshold like publicDecryptionThreshold, then you should use that for keygen too. I don't see any reason for those two to be different.

But to understand better, what happens when this threshold is reached? is it similar to public decryption where you check whether there's a majority? if you want to check for majority then you need to wait 2t + 1 responses, but your computation is `2(n/3) + 1`, this may lead to something a bit suboptimal when n is not exactly 3t + 1.

For example, n=6, t=1, so n >= 3t + 1. (note that t can only be 1 here, not 2)
But you compute 2( 6/3 ) + 1 = 5. This suggests you're waiting for 5 responses before doing the majority vote. But it's enough in this case to wait for only 3 responses because 3 = 2t+1.

jot2re: I agree with @kc1212 ; I am not sure why these thresholds come different places. Especially when they are the same (when n=3t+1).

isaacdecoded: I see 🤔 it seems like we need to revisit these thresholds again (sorry because it's something we have discussed several times already 🙏). [This is the related issue](https://github.com/zama-ai/fhevm-gateway/issues/387) from which we implemented them separately (public and user decryption thresholds).

Also, these thresholds are set via environment variables as [seen here](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/.env.example#L27-L29).

Should we then unify them and use the same configured threshold state variable for public and user decryptions, and also key generations? @kc1212 @jot2re 

> But to understand better, what happens when this threshold is reached? is it similar to public decryption where you check whether there's a majority?

Yes, this PR is introducing this new "majority threshold" = `($.kmsSignerAddresses.length * 2) / 3 + 1` to check whether a consensus for key generation responses has been reached between the KMS nodes and emit the respective events `KeygenRequest` or `ActivateKey`.

kc1212: I think you can just rename `PUBLIC_DECRYPTION_THRESHOLD` into something like `MAJORITY_VOTE_THRESHOLD` and use it for both public decryption and keygen. 

`USER_DECRYPTION_THRESHOLD` can remain the same since it's not majority vote

isaacdecoded: ~~As discussed with @leventdem, we want to keep it simple for now by not refactoring the GatewayConfig contract that much about these thresholds. So we'll reuse the current `PUBLIC_DECRYPTION_THRESHOLD ` for the keygen while adding important tracking comments and an [issue](https://github.com/zama-ai/fhevm-internal/issues/414) for revisiting this later.~~

After discussing again with @jatZama, we will continue with a more future-proof and less confusing approach by keeping these thresholds separated, so we will add a new threshold `KEYGEN_THRESHOLD`.

jatZama: like Levent said during the daily, it seems more intuitive and clearer to have two separate arguments, one for KEYGEN_THRESHOLD and a separate one for PUBLIC_DECRYPTION_THRESHOLD, because this is more readable, and more future-proof as well, what if some day we decide to change specs again?

gateway-contracts/contracts/interfaces/IDecryption.sol
```diff
@@ -35,12 +35,12 @@ interface IDecryption {
...
     /**
      * @notice Emitted when an public decryption request is made.
-     * @param decryptionId The decryption request ID.
+     * @param publicDecryptionId The decryption request ID.
      * @param snsCtMaterials The handles, key IDs and SNS ciphertexts to decrypt.
      * @param extraData Generic bytes metadata for versioned payloads. First byte is for the version.
      */
     event PublicDecryptionRequest(
-        uint256 indexed decryptionId,
+        uint256 indexed publicDecryptionId,
```
jatZama: Are offchain component listening to this event (kms connector I guess only here) aware of this change?

gateway-contracts/contracts/interfaces/IDecryption.sol
```diff
@@ -61,14 +61,14 @@ interface IDecryption {
...
-     * @param decryptionId The decryption request ID.
+     * @param userDecryptionId The decryption request ID.
      * @param snsCtMaterials The handles, key IDs and SNS ciphertexts to decrypt.
      * @param userAddress The user's address.
      * @param publicKey The user's public key for used reencryption.
      * @param extraData Generic bytes metadata for versioned payloads. First byte is for the version.
      */
     event UserDecryptionRequest(
-        uint256 indexed decryptionId,
+        uint256 indexed userDecryptionId,
```
jatZama: same question for kms connector

gateway-contracts/contracts/interfaces/IDecryption.sol
isaacdecoded: @eudelins-zama is this possible to include this parameter name change in the ongoing PR for the KMS connector? 🙏

gateway-contracts/contracts/interfaces/IDecryption.sol
jatZama: same here @eudelins-zama

gateway-contracts/contracts/interfaces/IGatewayConfig.sol
```diff
@@ -301,7 +317,7 @@ interface IGatewayConfig {
      * @notice Get the metadata of the KMS node with the given transaction sender address.
      * @return The KMS node's metadata.
      */
-    function getKmsNode(address kmsTxSenderAddress) external view returns (KmsNode memory);
+    function getKmsNode(address kmsTxSenderAddress) external view returns (KmsNodeV2 memory);
```
jatZama, line 339: and i guess here as well right? @isaacdecoded @eudelins-zama

gateway-contracts/contracts/KmsManagement.sol
```diff
@@ -1,126 +1,248 @@
...
     uint256 private constant PATCH_VERSION = 0;
 
-    /// Constant used for making sure the version number using in the `reinitializer` modifier is
-    /// identical between `initializeFromEmptyProxy` and the reinitializeVX` method
-    uint64 private constant REINITIALIZER_VERSION = 2;
+    /**
+     * @dev Constant used for making sure the version number using in the `reinitializer` modifier
+     * is identical between `initializeFromEmptyProxy` and the reinitializeVX` method
+     */
+    uint64 private constant REINITIALIZER_VERSION = 3;
```
jatZama: are you sure you want to bump the reinitializer of KMSManagement? i thought this one would be redeployed from scratch, not upgraded.

gateway-contracts/contracts/KmsManagement.sol
```diff
@@ -1,126 +1,248 @@
...
+import { UUPSUpgradeableEmptyProxy } from "./shared/UUPSUpgradeableEmptyProxy.sol";
+import { GatewayConfigChecks } from "./shared/GatewayConfigChecks.sol";
+import { Pausable } from "./shared/Pausable.sol";
+import { PREP_KEYGEN_COUNTER_BASE, KEY_COUNTER_BASE, CRS_COUNTER_BASE } from "./shared/KmsRequestCounters.sol";
+
+/**
+ * @title KMS Management contract
+ * @dev See {IKmsManagement}.
+ */
 contract KmsManagement is
```
jatZama: I think this contract should be renamed to KMSManagement, rationale is to be aligned with KMSVerifier name on host-contracts.

gateway-contracts/contracts/KmsManagement.sol
```diff
@@ -1,126 +1,248 @@
...
+        /// @notice The digest of the signed struct on which consensus was reached for a request
+        mapping(uint256 requestId => bytes32 digest) consensusDigest;
     }
 
-    /// @dev Storage location has been computed using the following command:
-    /// @dev keccak256(abi.encode(uint256(keccak256("fhevm_gateway.storage.KmsManagement")) - 1))
-    /// @dev & ~bytes32(uint256(0xff))
+    /**
+     * @dev Storage location has been computed using the following command:
+     * keccak256(abi.encode(uint256(keccak256("fhevm_gateway.storage.KmsManagement")) - 1))
```
jatZama: Would be nice to also update the storage slot after renaming of the contract since we have the occasion here since we deploy from scratch.

jatZama: New value would be : 
➜ // keccak256(abi.encode(uint256(keccak256("fhevm_gateway.storage.KMSManagement")) - 1))
`0x52e3d903674697c366245bdc532b6735f22bb42391635b8e0488806aba29452b`

gateway-contracts/contracts/interfaces/IGatewayConfig.sol
eudelins-zama: This function is not used in connector so no need for this one

gateway-contracts/contracts/interfaces/IDecryption.sol
eudelins-zama: Is it really useful to rename this field? I don't feel it add any information... (we know it's the decryption ID of a public decryption as the type is PublicDecryptionRequest)

On connector side, it is not that hard to adapt, but for consistency, it would mean I have to update my DB table and make the renaming everywhere this field is used, so I prefer to confirm it is a change we **really want**

gateway-contracts/contracts/interfaces/IDecryption.sol
eudelins-zama: Same here

gateway-contracts/contracts/Decryption.sol
```diff
@@ -261,25 +286,26 @@ contract Decryption is
...
-        uint256 decryptionId = $._decryptionRequestCounter;
+        $.publicDecryptionCounter++;
+        uint256 publicDecryptionId = $.publicDecryptionCounter;
 
         /// @dev The handles are used during response calls for the EIP712 signature validation.
-        $.publicCtHandles[decryptionId] = ctHandles;
+        $.publicCtHandles[publicDecryptionId] = ctHandles;
 
-        emit PublicDecryptionRequest(decryptionId, snsCtMaterials, extraData);
+        emit PublicDecryptionRequest(publicDecryptionId, snsCtMaterials, extraData);
```
jatZama, line 309: there might be changes needed on relayer side here and in userDecryptRequest due to new name of this variable in the event (and due to the big initial offset, but we talked about this already)

isaacdecoded: @manoranjith tagging you just to track it soon later 🙏

gateway-contracts/contracts/interfaces/IDecryption.sol
isaacdecoded: @eudelins-zama I see, it seems like this change is more impactful than we thought 🤔 let me see if we can avoid these renamings then, but initially it was required because our mocking script was breaking at some point. Will check it again, thanks.

eudelins-zama: Again, it's not too hard on connector's side, it's just lot of small changes I feel we could avoid.

Let me know if you are able to fix the mocking script somehow otherwise I could do the changes :innocent:

gateway-contracts/contracts/Decryption.sol
isaacdecoded: The change on these Decryption contract events has been rolled back [here in this commit](https://github.com/zama-ai/fhevm/pull/772/commits/e55d360bd360a17f886c5a0ed262261b6a6a43fc#diff-14c62001db03729747b5ad7bc744639d1f05199712fca3b9e99187c3c47b4994). Their current definitions are unchanged.

gateway-contracts/contracts/interfaces/IDecryption.sol
isaacdecoded: The change on these Decryption contract events has been rolled back [here in this commit](https://github.com/zama-ai/fhevm/pull/772/commits/e55d360bd360a17f886c5a0ed262261b6a6a43fc#diff-14c62001db03729747b5ad7bc744639d1f05199712fca3b9e99187c3c47b4994). Their current definitions are unchanged.

gateway-contracts/contracts/interfaces/IDecryption.sol
isaacdecoded: Same here 😬

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -17,9 +18,37 @@ import "./shared/Pausable.sol";
  * @dev See https://github.com/zama-ai/fhevm-gateway/issues/98 for more details.
  */
 contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeableEmptyProxy, Pausable {
+    // ----------------------------------------------------------------------------------------------
+    // Upgrade compatibility (V2 -> V3):
+    // ----------------------------------------------------------------------------------------------
+
+    /// @notice Input data used to upgrade the contract from V2 to V3.
+    struct V3UpgradeInput {
```
jatZama, line 26: why did you use a struct here? this does not make much sense for me, especially since `reinitializeV3` method also takes a third argument (keygenThreshold).

gateway-contracts/contracts/KMSManagement.sol
```diff
@@ -0,0 +1,642 @@
...
+        // Check that the signer has not already signed for this key generation response
+        if ($.kmsHasSignedForResponse[keyId][kmsSigner]) {
+            revert KmsAlreadySignedForKeygen(keyId, kmsSigner);
+        }
+
+        $.kmsHasSignedForResponse[keyId][kmsSigner] = true;
+
+        // Store the KMS transaction sender address and the s3 bucket URL for the keygen response,
+        // event if the consensus has already been reached
+        string[] memory consensusUrls = _storeConsensusMaterials(keyId, digest);
```
jatZama, line 304: This design is wasting gas unnecessarily by restoring same urls already in storage somewhere else. Maybe we could just store indexes instead. I saw same error in `crsResponse` in this same code, as well as in the other coprocessor context one.

jatZama: But maybe it is not so urgent here, because keygen is rarely invoked.
On the other hand it was a lot more worse in the context PR, because currently  the urls are stored there for each handle added in ACL (this is very common operation we want to optimize to not get same issue as userDecryptResponse)

gateway-contracts/.env.example
isaacdecoded: Added in this [commit](https://github.com/zama-ai/fhevm/wiki/0.9.0/0e566640cde4715c5aabe681ead3f6bc6d369ab3) 👍

isaacdecoded (approved):

isaacdecoded:
@mergifyio queue

PR #778, opened 2025-08-25
feat(kms-connector): kms sharding support

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/330


kms-connector/config/kms-worker.toml
```diff
@@ -8,10 +8,12 @@
...
 
-# KMS-core gRPC endpoint (required)
+# List of KMS-core gRPC endpoints (required)
 # Format: http://host:port
-# ENV: KMS_CONNECTOR_KMS_CORE_ENDPOINT
-kms_core_endpoint = "http://[::1]:50051"
+# ENV: KMS_CONNECTOR_KMS_CORE_ENDPOINTS
+# If using environment variables, use a comma separated list of URLs
+# Ex: KMS_CONNECTOR_KMS_CORE_ENDPOINTS="http://[::1]:50051,http://[::1]:50052"
+kms_core_endpoint = ["http://[::1]:50051"]
```
piizama: @eudelins-zama Can you add a fallback to the `kms_core_endpoint` config key if `kms_core_endpoints` is not defined. We should avoid any breaking change on configuration files or env vars as much as possible. The old property can simply marked as "deprecated".

eudelins-zama: Should be fixed by 6274447b25d889cbe4106ebdaa030097167dc679

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
```diff
@@ -187,18 +200,23 @@ impl KmsClient {
...
+                async move { client.get_user_decryption_result(request).await }
             },
         )
         .await?;
 
         KmsGrpcResponse::try_from((request_id, grpc_response))
     }
+
+    fn choose_client(&self, request_id: RequestId) -> CoreServiceEndpointClient<Channel> {
+        let request_id = request_id.request_id.parse::<usize>().unwrap_or_default();
```
maksymsur: Perhaps we need to log a warning/error here if it defaults? These cases shall be monitored IMO

maksymsur (approved):
Just a minor comment - overall looks good! Let's keep the approach simple initially, validate its efficiency and measure the actual gains, then polish it based on real-world data.

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
eudelins-zama: Fixed in 68c6e2b2b550815306739b7f3cb6c3be8b771d50

PR #780, opened 2025-08-25
Make transaction_id column non null

antoniupop:


coprocessor/fhevm-engine/sns-worker/src/executor.rs
```diff
@@ -599,6 +599,15 @@ async fn update_computations_status(
             )
             .execute(db_txn.as_mut())
             .await?;
+            sqlx::query!(
```
goshawk-3, line 607: nitpick: Maybe add some comments to document the change

goshawk-3 (approved):
LGTM

PR #781, opened 2025-08-25
chore(kms-connector): remove verify_coprocessor config

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/347

@piizama regarding our earlier discussion, here the config field that was removed was an optional one that was not used by the connector, and that should not have been configured either by infra team. So I guess it is fine to remove it, right? Plus, even if this field stay configured, it will just be ignored by the connector, so really no impact I would say


leventdem (approved):
Let's clean a bit!

piizama (approved):

PR #797, opened 2025-08-30
Fix DFG cleanup

antoniupop:


coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -223,15 +200,37 @@ FOR UPDATE SKIP LOCKED            ",
         let key_cache = tenant_key_cache.read().await;
         // Clear unneeded work items
         for (_, work) in work_by_tenant.iter_mut() {
-            let mut work_to_remove = vec![];
+            let mut work_to_remove: BTreeSet<usize> = BTreeSet::new();
```
goshawk-3, line 203: nitpick: This can be a HashSet?

goshawk-3 (approved):
LGTM

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
antoniupop: As we need to traverse the set in order, this is probably more efficient than sorting after the fact.

PR #802, opened 2025-09-02
feat(gateway-contracts): add `Clone` traits in rust bindings

fd0r:
https://github.com/foundry-rs/foundry/pull/11516


fd0r:
This is a bit of a dirty fix because I patched my locally built foundry to add the `Clone` trait as `extra_derives` then built the bindings with it.

fd0r:
Actually the way I patched `foundry` was wrong.

https://github.com/alloy-rs/core/pull/1003

PR #808, opened 2025-09-02
fix(coprocessor): cargo audit, dependency upgrade/cleanup

rudy-6-4:


rudy-6-4:
It remove https://rustsec.org/advisories/RUSTSEC-2025-0055.
There are still indirect dependencies to paste and derivative but it remove the direct dependecy and some indirect one..

antoniupop (approved):

PR #812, opened 2025-09-02
feat(kms-connector): retrieve s3 url in extra_data

eudelins-zama:
Ref https://github.com/zama-ai/fhevm-internal/issues/404


kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
```diff
@@ -406,6 +216,17 @@ pub fn compute_digest(ct: &[u8]) -> Vec<u8> {
     result
 }
 
+pub fn extract_fallback_url(s3_config: &S3Config) -> String {
```
melanciani, line 245: are we sure we want a fallback url ? what's the purpose of it exactly ?

melanciani (approved):
thanks !

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
eudelins-zama: It has been useful in the past when zws had issue with S3 setup, it offered a way to use the S3 bucket we wanted as a fallback to debug the rest of the stack if needed

dartdart26 (approved):

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
```diff
@@ -118,14 +114,14 @@ where
         decryption_id: U256,
         key_id: &str,
         sns_materials: Vec<SnsCiphertextMaterial>,
+        extra_data: &[u8],
     ) -> anyhow::Result<Vec<TypedCiphertext>> {
-        // Retrieve ciphertext materials from S3
+        let s3_url_matrix = decode_s3_url_matrix(extra_data)?;
```
melanciani: I updated https://github.com/zama-ai/fhevm-internal/issues/404#issuecomment-3323575775 : the urls are no longer in extraData and instead are a new field to the event

melanciani (changes requested):
small change: storage urls are no longer in extraData and can be found as a new field in the request event

eudelins-zama:
Should be good now @RomanBredehoft (cc @dartdart26)

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
```diff
@@ -37,208 +27,42 @@ pub struct S3Service<P: Provider> {
...
     pub async fn retrieve_sns_ciphertext_materials(
         &self,
         sns_materials: Vec<SnsCiphertextMaterial>,
+        s3_urls_matrix: Vec<Vec<String>>,
     ) -> Vec<TypedCiphertext> {
         let mut sns_ciphertext_materials = Vec::new();
 
-        for sns_material in sns_materials {
-            // Get S3 URL and retrieve ciphertext
+        for (sns_material, mut s3_urls) in sns_materials.into_iter().zip(s3_urls_matrix) {
```
melanciani, line 52: shouldn't you maybe check that both `sns_materials` and `s3_urls_matrix` have the sane length with a proper error ? just in case ?

melanciani (approved):
looks good to me !

kms-connector/crates/kms-worker/src/core/event_processor/s3.rs
eudelins-zama: Good catch! I did in aaea886bfd39fc358a3106943b151f7e9f3c1cc6 

Not raising errors immediately, will try to use fallback config first FYI

melanciani: alright, but still not sure I'm a fan of the fallback url : how do you chose the coprocessor storage to use for the fallback ? 

this means we somewhat have to define/harcode a "more trusted" copro for its storage url. But this feels a bit wrong no ? What if the chosen coprocessor becomes malicious, or its storage URL changes ? It looks like the gateway is no longer the single point of truth as these cases will require additional manual update in each KMS connector

What I'm trying to say is that I feel like if there is an issue with the URLs sent by the gateway, it's most likely an issue in the protocol that should raise an error and that should be fixed in the gateway itself. I don't think the connector should have a patch for these cases 

wdyt ?

eudelins-zama: For now, this was just used by infra as a debugging solution on devnet/testnet, not meant to be used once the env is working well.

But we could open an issue to completely remove it if it is OK on infra side, I just don't think this PR is the good place for that, so I tried to make it coherent with current implementation

melanciani: yes I see, well yes I'll let you create an issue on that because I indeed we shouldn't keep this code, but yes agree maybe not in this PR !

eudelins-zama:
Closing, no longer wanted

PR #818, opened 2025-09-04
KMS connector helm chart refactor

piizama:
* Uniformize config values with kms-core
* Rework common configuration values
* Enable probes by default
* Disable tracing by default


tawadaa (approved):
LGTM, thanks!

PR #823, opened 2025-09-05
feat(coprocessor): handle graceful shutdown cases in txn-sender

dartdart26:
This commit assumes that `provider_max_retries` and `provider_retry_interval` are set such that it would take a long time until the provider stops retrying. If that happens, we consider the provider no longer usable and we shut down the txn-sender. This keeps code simpler and avoids the need to refactor and retest retry behaviour. Reason we do it is that once retries are exhausted, the provider is no longer usable and `BackendGone` is returned.

We could consider using HTTP instead of WS, but that is left for future investigation.

The following other changes are made:
 * propagete errors without converting them to strings, e.g. `bail(a)` instead of `bail("xxx {}", e)`
 * remove the `provider()` and `check_provider_connection()` functions from the `TransactionOperation` trait as they are used for health checking and we do not need the health check to go to just the first operation - instead, we use the provider from the `TransactionSender`
* since the provider can retry for a long time, if a signal to stop the sender is received, we add a graceful timeout such that the sender stops in a timely manner and doesn't wait for a long time for the provider - for that, we introduce the `graceful_shutdown_timeout` config option
* plug the cancellation token to `get_chain_id()` on startup, such that the sender can stop in a timely manner
* use `tokio::spawn()` instead of join! for starting the sender and the HTTP server

### DIFF
--- a/coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
+++ b/coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
@@ -122,6 +122,9 @@ struct Conf {
 
     #[arg(long, default_value = "120", value_parser = clap::value_parser!(u32).range(100..))]
     gas_limit_overprovision_percent: u32,
+
+    #[arg(long, default_value = "8s", value_parser = parse_duration)]
+    graceful_shutdown_timeout: Duration,
 }
 
 fn install_signal_handlers(cancel_token: CancellationToken) -> anyhow::Result<()> {
@@ -129,10 +132,11 @@ fn install_signal_handlers(cancel_token: CancellationToken) -> anyhow::Result<()
     let mut sigterm = signal(SignalKind::terminate())?;
     tokio::spawn(async move {
         tokio::select! {
-            _ = sigint.recv() => (),
-            _ = sigterm.recv() => ()
+            _ = sigint.recv() => { info!("received SIGINT"); },
+            _ = sigterm.recv() => { info!("received SIGTERM"); },
         }
         cancel_token.cancel();
+        info!("Cancellation signal sent over the token");
     });
     Ok(())
 }
@@ -149,7 +153,19 @@ async fn main() -> anyhow::Result<()> {
         .with_max_level(conf.log_level)
         .init();
 
-    let chain_id = get_chain_id(conf.gateway_url.clone(), conf.provider_retry_interval).await;
+    let cancel_token = CancellationToken::new();
+    install_signal_handlers(cancel_token.clone())?;
+
+    let chain_id = get_chain_id(
+        conf.gateway_url.clone(),
+        conf.graceful_shutdown_timeout,
+        cancel_token.clone(),
+    )
+    .await;
+    if chain_id.is_none() {
+        info!("Cancellation requested before getting chain ID during startup, exiting");
+        return Ok(());
+    }
     let abstract_signer: AbstractSigner;
     match conf.signer_type {
         SignerType::PrivateKey => {
@@ -160,7 +176,7 @@ async fn main() -> anyhow::Result<()> {
                 ));
             }
             let mut signer = PrivateKeySigner::from_str(conf.private_key.unwrap().trim())?;
-            signer.set_chain_id(Some(chain_id));
+            signer.set_chain_id(chain_id);
             abstract_signer = make_abstract_signer(signer);
         }
         SignerType::AwsKms => {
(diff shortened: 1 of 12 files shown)


coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
```diff
@@ -149,7 +153,19 @@ async fn main() -> anyhow::Result<()> {
...
+        conf.gateway_url.clone(),
+        conf.graceful_shutdown_timeout,
+        cancel_token.clone(),
+    )
+    .await;
+    if chain_id.is_none() {
+        info!("Cancellation requested before getting chain ID during startup, exiting");
+        return Ok(());
+    }
     let abstract_signer: AbstractSigner;
```
eudelins-zama: nit: I feel the more adequate pattern here would be to use `select!` macro on `get_chain_id` and `cancel_token.cancelled()`, instead of providing the cancel token to get_chain_id and returning an option

coprocessor/fhevm-engine/transaction-sender/src/lib.rs
```diff
@@ -76,11 +81,24 @@ impl HealthStatus {
...
+            info!("Cancellation requested before getting chain ID");
+            return None;
+        }
         let provider = match ProviderBuilder::new()
-            .connect_ws(WsConnect::new(ws_url.clone()))
+            .connect_ws(
+                WsConnect::new(ws_url.clone())
+                    .with_max_retries(1)
+                    .with_retry_interval(retry_interval),
+            )
```
eudelins-zama: nit: see above comment first.

But if you want to keep cancel_token in this fn, shouldn't you use `select!` tokio macro on `connect_ws` + `cancel_token.cancelled`?

coprocessor/fhevm-engine/transaction-sender/src/transaction_sender.rs
```diff
@@ -173,10 +185,39 @@ impl<P: Provider<Ethereum> + Clone + 'static> TransactionSender<P> {
...
+                            break Err(e.into());
+                        }
+                        None => {
+                            info!("All operations stopped");
+                            break Ok(())
+                        }
+                    }
+                }
+            }
         }
```
eudelins-zama, line 220: nit: don't you want to use `.join_all` instead of looping on `join_next` here?

coprocessor/fhevm-engine/transaction-sender/src/lib.rs
```diff
@@ -113,3 +131,16 @@ pub async fn get_chain_id(ws_url: Url, retry_interval: Duration) -> u64 {
...
+        if let Some(t) = cause.downcast_ref::<TransportError>() {
+            matches!(
+                t,
+                TransportError::Transport(TransportErrorKind::BackendGone)
+            )
+        } else {
+            false
+        }
+    })
+}
```
eudelins-zama, line 135: If this is the only time we need downcasting, I guess it's fine, but if such fn tends to multiply, it might worth defining proper error enum

eudelins-zama (approved):
Small improvements/comments suggested here and there, but otherwise looks good!

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
dartdart26: I think that makes sense, I will do it as you suggest.

coprocessor/fhevm-engine/transaction-sender/src/lib.rs
dartdart26: I think you previous comment makes sense, will do it as you suggest.

coprocessor/fhevm-engine/transaction-sender/src/transaction_sender.rs
dartdart26: I opted for `join_next()` due to this comment on `join_all()`:

```
If any tasks on the `JoinSet` fail with an [`JoinError`], then this call
to `join_all` will panic and all remaining tasks on the `JoinSet` are
cancelled. To handle errors in any other way, manually call [`join_next`]
in a loop.
```

coprocessor/fhevm-engine/transaction-sender/src/lib.rs
dartdart26: I did it as it is not a library, but lives in the same crate and is used only internally. I evaluated a proper enum, but felt it is an overkill for now in the way the codebase is structured.

PR #833, opened 2025-09-06
Fix scalar compression: handle scalars separately from ciphertexts

Budalebah:
This PR properly handles scalars separately from ciphertexts.

## Problem
- Scalars are not ciphertexts, but the code was trying to compress them like ciphertexts
- This caused a panic: "cannot compress a scalar"

## Solution
- Scalars are now serialized directly instead of being compressed
- Added proper handling for scalar compression/decompression in all relevant functions
- Scalars use type number 200 and are handled separately from ciphertext compression

## Changes
- Updated `compress()` function to handle scalars separately
- Updated all `decompress()` functions to handle scalar type (200)
- Removed panic and replaced with proper serialization

## Files Changed
- `coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs`


immortal-tofu:
Your PR seems broken.

PR #839, opened 2025-09-07
docs(library-solidity): specify that rand bounds must be power of 2

arajasek:
AFAIC tell, the upper bound when asking for randomness must always be a power of 2 (see [here](https://github.com/zama-ai/fhevm/blob/7e12c8bfc1782dfa465323b200cd449201208dd5/host-contracts/contracts/FHEVMExecutor.sol#L905-L923)). If so, we should specify this in documentation, and fixup examples that use non powers of two.

Apologies if I'm missing something.


aquint-zama:
@cla-bot check

docs/solidity-guides/operations/random.md
```diff
@@ -36,13 +36,13 @@ function randomBoolean() public returns (ebool) {
...
+To generate random numbers within a specific range, you can specify an **upper bound**. The specified upper bound must be a power of 2. The random number will be in the range `[0, upperBound - 1]`.
 
 ```solidity
 // Generate random numbers with upper bounds
-euint8 r8 = FHE.randEuint8(100);      // Random number between 0-99
-euint16 r16 = FHE.randEuint16(1000);  // Random number between 0-999
-euint32 r32 = FHE.randEuint32(1000000); // Random number between 0-999999
+euint8 r8 = FHE.randEuint8(32);      // Random number between 0-31
+euint16 r16 = FHE.randEuint16(512);  // Random number between 0-511
+euint32 r32 = FHE.randEuint32(65536); // Random number between 0-65535
```
immortal-tofu, line 45: Wrong range values

immortal-tofu (changes requested):

docs/solidity-guides/operations/random.md
arajasek: Can you please explain? I didn't pick the max possible upper bound for each uint type, just an example power of 2 that fits in the size.

immortal-tofu: Ho yes sorry, your comment was regarding the value entered and not the maximum value. My bad 👍

immortal-tofu (approved):

PR #841, opened 2025-09-08
feat(protocol-contracts): add protocol contracts with token

jatZama:


protocol-contracts/package.json
```diff
@@ -0,0 +1,81 @@
+{
+  "name": "@layerzerolabs/oft-adapter-example",
```
isaacdecoded: nit: should we name it like `@fhevm/protocol-contracts` instead?

protocol-contracts/README.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+You can still select either both networks or alternatively uniquely the `arbitrum-testnet` network this time. Now enter `ZamaOFT` when requested for the deploy script tag. This will deploy the `ArbitrumOFT` contract on Arbitrum Sepolia testnet.
+
+## Step 5 : Etherscan verification (Optional)
+
+In order verify the two contracts (`ZamaERC20` and `ZamaOFTAdapter`) on Etherscan for the Ethereum Sepolia network, use this command (don't worry if the following scripts return you an error or an invalid explorer URL link - see the note at the end of this step):
+
+```
+pnpm verify:etherscan:ethereum:sepolia
```
isaacdecoded, line 71: nit: maybe replace `pnpm` by `npm` also here in this readme file.

protocol-contracts/contracts/ZamaOFTAdapter.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.22;
+
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+import { OFTAdapter } from "@layerzerolabs/oft-evm/contracts/OFTAdapter.sol";
+
+/**
+ * @title OFTAdapter Contract
```
isaacdecoded: nit: `OFTAdapter` -> `ZamaOFTAdapter`

protocol-contracts/contracts/ZamaERC20.sol
```diff
@@ -0,0 +1,41 @@
...
+     */
+    function mint(address _to, uint256 _amount) public onlyRole(MINTER_ROLE) {
+        _mint(_to, _amount);
+    }
+
+    /**
+     * @notice Anyone could burn tokens he owns.
+     * @param amount Number of tokens to burn from own balance.
+     */
+    function burn(uint256 amount) public {
```
mortendahl, line 31: Is it worth using [`ERC20Burnable`](https://docs.openzeppelin.com/contracts/5.x/api/token/erc20#ERC20Burnable) instead?

protocol-contracts/contracts/ZamaOFTAdapter.sol
mortendahl: Are the comments even relevant here? They seem more appropriate for the general `OFTAdapter` than for `ZamaOFTAdapter`

protocol-contracts/README.md
jatZama: Good catch, i should redo the whole deployment process now to check npm still works. I did the change here since every other component of the monorepo uses npm, but did not test it, but layerzero recommends pnpm and this was tested only with pnpm for the moment. Give me few minutes to check everything still works with nom.

protocol-contracts/contracts/ZamaERC20.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/ERC20Permit.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/ERC1363.sol";
+import "@openzeppelin/contracts/access/AccessControl.sol";
+
+// @dev WARNING: This is for testing purposes only
+contract ZamaERC20 is ERC20, ERC20Permit, ERC1363, AccessControl {
```
mortendahl, line 10: @jatZama is there any downsides to extending `ERC20Votes` as well, in case we ever need it?

jatZama: This PR was to have a first version of the protocol-contracts directory ready to be used for OZ and us. We can rediscuss the token design in a future PR.  I thought that the goal is to merge this first asap. Am I wrong?

protocol-contracts/contracts/ZamaOFTAdapter.sol
jatZama: I have no strong opinion here, maybe for security it makes sense to keep them, but I could remove them if you want.

protocol-contracts/package.json
jatZama: Yes, will do it now.

protocol-contracts/contracts/ZamaERC20.sol
jatZama: This contract gives also access to a `burnFrom` function, do we need it? But maybe we should discuss this in a future issue/PR, see: https://github.com/zama-ai/fhevm/pull/841#discussion_r2330697778

protocol-contracts/README.md
jatZama: Unfortunately, due to non-retrocompatible new version of layer zero plugin, the npm script is stuck forever at the wiring step, so i am rolling back to using pnpm for now.

protocol-contracts/package.json
jatZama: Done.

jatZama:
I roll back the protcol-contracts to use pnpm instead of npm in last commit, since I noticed that the new version of layer-zero plugins is not retro-compatible, I tried to pin to an older version inside the package.json to no avail (maybe I did not update correct package). So making it compatible with npm instead of pnpm seems a lot less trivial than expected. I have opened this issue https://github.com/zama-ai/fhevm-internal/issues/416  to do it later. 
I also opened a second issue to track new specs for the token, following all the suggestions from Morten in the comments of this PR: https://github.com/zama-ai/fhevm-internal/issues/417
For now let's merge this with current working version of pnpm and contracts. Can one of you approve it? @mortendahl @isaacdecoded

isaacdecoded (approved):
LGTM! 👍

PR #845, opened 2025-09-09
chore(gateway-contracts): user decrypt shares in events only

eudelins-zama:
associated fhevm-relayer PR: https://github.com/zama-ai/console/pull/659

closes https://github.com/zama-ai/fhevm-internal/issues/331


jatZama (approved):

gateway-contracts/contracts/Decryption.sol
```diff
@@ -194,7 +194,7 @@ contract Decryption is
         // User decryption state variables:
         // ----------------------------------------------------------------------------------------------
         /// @notice Verified signatures for a user decryption.
-        mapping(uint256 decryptionId => bytes[] verifiedSignatures) _verifiedUserDecryptSignatures;
+        mapping(uint256 decryptionId => uint256 verifiedSignatures) _verifiedUserDecryptSignatures;
```
isaacdecoded: Not sure if this change could have conflicts during upgrade due to incompatible storage layout. Meaning that we should keep it like this and add the new definition at the end of the layout struct. wdyt? @jatZama

eudelins-zama: FYI, I used what @jatZama did in his branch, so I'm not fully aware of reason of this change.

Could we leave `bytes[]` for now? Or would it break stuff?

jatZama: fixed

gateway-contracts/contracts/interfaces/IDecryption.sol
```diff
@@ -78,17 +78,21 @@ interface IDecryption {
...
         uint256 indexed decryptionId,
-        bytes[] userDecryptedShares,
-        bytes[] signatures,
+        uint256 counterShares,
+        bytes userDecryptedShare,
+        bytes signature,
         bytes extraData
     );
 
+    event UserDecryptionResponseConsensusReached(uint256 indexed decryptionId);
```
melanciani: should we really name it that way ? there is no real consensus for user decryption, we only emit this event once we have `threshold` response, since we cannot verify the shares on-chain

we suggested `UserDecryptionResponseThreshold` in the wiki but it could be `UserDecryptionResponseThresholdReached`

jatZama: as you wish, but changing the name of this event must be done asap, and maybe it is already too late because Relayer team already assumed this signature

melanciani: just updated the name as @manoranjith confirmed it was fine !

gateway-contracts/contracts/Decryption.sol
```diff
@@ -542,27 +540,20 @@ contract Decryption is
...
         // It is important to consider the same mapping fields used for the consensus
         // A "late" valid KMS transaction sender address will still be added in the list.
         // We thus use a zero digest (default value for `bytes32`) to still be able to retrieve the
         // list later independently of the decryption response type (public or user).
-        $.consensusTxSenderAddresses[decryptionId][0].push(msg.sender);
+        address[] storage txSenderAddresses = $.consensusTxSenderAddresses[decryptionId][0];
+        txSenderAddresses.push(msg.sender);
+
+        // Store the user decrypted share for the user decryption response.
+        emit UserDecryptionResponse(decryptionId, txSenderAddresses.length, userDecryptedShare, signature, extraData);
```
jatZama: Can you replace `txSenderAddresses.length` by `txSenderAddresses.length-1` to still index events by starting with 0?

melanciani: done ! but I renamed `counterShares` to `indexShare` to better reflect its purpose

jatZama (approved):
LGTM

isaacdecoded (approved):
Great! 👏

PR #848, opened 2025-09-10
Fix perofrmance of query for work

antoniupop:


immortal-tofu (approved):
Not sure to understand what is .sqlx files, otherwise LGTM

antoniupop:
> Not sure to understand what is .sqlx files, otherwise LGTM

Those are the query cache, just automatically generated by sqlx

antoniupop:
Dependency check is fixed in separate commit in main - could be ported here as well, but is a separate issue.

PR #860, opened 2025-09-12
feat(gateway-contracts): replace pauser multisig by PauserSet contract

jatZama:
This should close issue https://github.com/zama-ai/fhevm-internal/issues/384


jatZama:
For some unknown reason it is telling me bindings are not up to date, but when i run make update-bindings locally, it tells me they are updated. cc: @eudelins-zama would you know how is this possible?

gateway-contracts/contracts/Decryption.sol
```diff
@@ -160,12 +160,12 @@ contract Decryption is
...
     string private constant CONTRACT_NAME = "Decryption";
     uint256 private constant MAJOR_VERSION = 0;
-    uint256 private constant MINOR_VERSION = 3;
+    uint256 private constant MINOR_VERSION = 4;
     uint256 private constant PATCH_VERSION = 0;
 
     /// Constant used for making sure the version number using in the `reinitializer` modifier is
     /// identical between `initializeFromEmptyProxy` and the reinitializeVX` method
-    uint64 private constant REINITIALIZER_VERSION = 5;
+    uint64 private constant REINITIALIZER_VERSION = 6;
```
isaacdecoded, line 168: Just wondering why this version bump is needed for this contract?

gateway-contracts/contracts/InputVerification.sol
```diff
@@ -67,12 +67,12 @@ contract InputVerification is
...
     string private constant CONTRACT_NAME = "InputVerification";
     uint256 private constant MAJOR_VERSION = 0;
-    uint256 private constant MINOR_VERSION = 1;
+    uint256 private constant MINOR_VERSION = 2;
     uint256 private constant PATCH_VERSION = 0;
 
     /// Constant used for making sure the version number using in the `reinitializer` modifier is
     /// identical between `initializeFromEmptyProxy` and the reinitializeVX` method
-    uint64 private constant REINITIALIZER_VERSION = 3;
+    uint64 private constant REINITIALIZER_VERSION = 4;
```
isaacdecoded, line 75: Also wondering why the version bump is needed here.

jatZama: It is needed for sure because it inherits from a modified version of Pausable.

gateway-contracts/contracts/Decryption.sol
jatZama: It is needed for sure because it inherits from a modified version of `Pausable`.

isaacdecoded: Ah! I see, thanks! 👍

gateway-contracts/contracts/shared/Pausable.sol
```diff
@@ -34,7 +34,7 @@ abstract contract Pausable is PausableUpgradeable {
     error NotOwnerOrGatewayConfig(address notOwnerOrGatewayConfig);
 
     modifier onlyPauser() {
-        if (msg.sender != _GATEWAY_CONFIG.getPauser()) {
+        if (!_GATEWAY_CONFIG.isPauser(msg.sender)) {
```
melanciani, line 37: wouldn't it make more sense to directly store and call the `PauserSet` contract here, instead of going through the gateway config one ?

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -331,10 +297,9 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         }
     }
 
-    /// @dev See {IGatewayConfig-getPauser}.
-    function getPauser() external view virtual returns (address) {
-        GatewayConfigStorage storage $ = _getGatewayConfigStorage();
-        return $.pauser;
+    /// @dev See {IGatewayConfig-isPauser}.
+    function isPauser(address account) external view virtual returns (bool) {
```
melanciani, line 301: following https://github.com/zama-ai/fhevm/pull/860/files#r2371661722, I wonder if it atually makes sense to have this function in GatewayConfig, why not only go through PauserSet directly ?

gateway-contracts/contracts/shared/Pausable.sol
jatZama: I think it is better more sense to keep it this way, because we would like to centralise the PauserSet address. If for any reason we want to update the pauserSet, since it is not an upgradeable contract, the only way would be to redeploy a new contract and upgrade the gatewayConfig. If you store it instead in the different contracts requiring pausing, you will need otherwise to upgrade each of them one by one. Wdyt?

gateway-contracts/contracts/GatewayConfig.sol
jatZama: Same reason to not do this as https://github.com/zama-ai/fhevm/pull/860/files#r2371761520

jatZama: We should see gatewayConfig as the "single source of truth" in general whenever possible, making debugging and upgrading mre straightforward.

gateway-contracts/contracts/shared/Pausable.sol
melanciani: that's a good point yes, I agree then thanks for the explanation !

then, should the `PAUSER_SET` address be a private constant in the gateway config, which requires upgrading the contract to update the address ? or should we instead make it a variable that can be updated with a `updatePauserSet` (restricted to the owner), which seems simpler to manage ?

jatZama: Both are fine, to be honest, this kind of upgrade should never happen. I would say I prefer current approach with constant, this way we simplify governance. Remember that for every action protected by onlyOwner, we would need to define specific scopes with specific thresholds and list of owners in Aragon DAO, so it is not "simpler" (also costing more gas if we use storage instead of gas, but this is negligible for rare actions).

melanciani: alright I see, then good to me I guess

melanciani (approved):
looks good, thanks !

gateway-contracts/.env.example
```diff
@@ -131,3 +127,8 @@ FHE_PARAMS_DIGEST="0xd61fa44c57b6e7526a46e1072d63332ab4eb38e050fe54dd51fe995f905
 # is dynamically set after contracts are deployed.
 # For tests, this address is read directly from the addresses/.env.gateway file, so the value in this file is ignored.
 GATEWAY_CONFIG_ADDRESS="0xC7D45661a345eC5cA0e8521CFEF7e32FDA0Daa68" # (address)
+
+# Pausers
+NUM_PAUSERS="1"
+PAUSER_ADDRESS_0="0xa44366bAA26296c1409AD1e284264212029F02f1" # accounts[2] (address)
+PAUSER_PRIVATE_KEY_0="0x7ae52cf0d3011ef7fecbe22d9537aeda1a9e42a0596e8def5d49970eb59e7a40" # accounts[2], private key (bytes32)
```
melanciani, line 134: not sure this is useful, right ?

PR #867, opened 2025-09-14
Fix typo in CHANGELOG: 'see the magic happens' → 'see the magic happen'

HusseinAdeiza:
This PR corrects a typo in test-suite/CHANGELOG.md by changing "see the magic happens" to "see the magic happen."

Correction:
"see the magic happen"
("happens" is incorrect; "magic happen" is the correct idiom.)

"between 20 and 150 seconds"
This phrasing is fine, but for clarity you could use: "between 20 and 150 seconds per test" if you want to be more explicit (optional, not a typo).

Note this doesn't affect any code functionality


Eikix:
Kindly closing as stale.

HusseinAdeiza:
@Eikix no reward ?

PR #875, opened 2025-09-18
feat(host-contracts): introduce coprocessor contexts logic

isaacdecoded:
This PR introduces two types of coprocessor contexts into the InputVerifier contract: active and suspended. The active context is always present and initialized during the contract deployment. The suspended one is always the contextId being replaced when calling the `addNewContextAndSuspendCurrentOne`, and it could be "removed" by setting it into a deactivated state.

On the other hand, the signers set is now retrieved based on a given coprocessorContextId, which must match the currently stored active or suspended contexts.

Closes https://github.com/zama-ai/fhevm-internal/issues/112


host-contracts/contracts/InputVerifier.sol
```diff
@@ -113,14 +119,25 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+        address[] activeCoprocessorContextSigners;
+        /// @notice Suspended coprocessor context ID.
+        uint256 suspendedCoprocessorContextId;
+        /// @notice List of signers for the suspended coprocessor context ID.
+        address[] suspendedCoprocessorContextSigners;
     }
 
     /// Constant used for making sure the version number used in the `reinitializer` modifier is
     /// identical between `initializeFromEmptyProxy` and the `reinitializeVX` method
-    uint64 private constant REINITIALIZER_VERSION = 4;
```
jatZama: It does not make sense to bump REINITIALIZER_VERSION again, current version deployed on Sepolia is still using version 3. This is because they did not add the ACL centralized owner commit in last upgrade.

jatZama: Please always check what is currently deployed before bumping or not the reinitializer , see https://sepolia.etherscan.io/address/0xf2f502c4b384acbd6d3fcc897747fd6d41833bd5#code

host-contracts/contracts/InputVerifier.sol
```diff
@@ -135,31 +152,53 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
     }
 
     /**
-     * @notice Re-initializes the contract from V1.
+     * @notice Re-initializes the contract from V3.
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    function reinitializeV3() public virtual reinitializer(REINITIALIZER_VERSION) {}
+    function reinitializeV4(
```
jatZama: Here as well, keep `reinitializeV3` name please.

host-contracts/contracts/InputVerifier.sol
```diff
@@ -428,25 +521,40 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
     }
 
     function _verifyEIP712(CiphertextVerification memory ctVerif, bytes[] memory signatures) internal virtual {
+        // Decode the contextId from the extraData.
+        (, uint256 contextId) = abi.decode(ctVerif.extraData, (uint256, uint256));
```
jatZama: what is the extraData content here? can you comment what are the two uint256 corresponding to?

host-contracts/contracts/InputVerifier.sol
```diff
@@ -326,34 +368,85 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+        for (uint256 i = 0; i < coprocessorSigners.length; i++) {
+            if (coprocessorSigners[i] == account) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    function updateActiveCoprocessorContext(uint256 contextId, address[] calldata signers) public virtual onlyACLOwner {
+        if (contextId == 0) {
```
jatZama: Shouldn't you add a check to check also that contextId is different from the `suspendedCoprocessorContextId` ?It seems it would make sense to avoid mistakes.

host-contracts/contracts/InputVerifier.sol
```diff
@@ -326,34 +368,85 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+        InputVerifierStorage storage $ = _getInputVerifierStorage();
+        $.activeCoprocessorContextId = contextId;
+        $.activeCoprocessorContextSigners = signers;
+    }
+
+    function updateSuspendedCoprocessorContext(
+        uint256 contextId,
+        address[] calldata signers
+    ) public virtual onlyACLOwner {
+        if (contextId == 0) {
```
jatZama: Shouldn't you add a check to check also that contextId is different from the `activeCoprocessorContextId` ?It seems it would make sense to avoid mistakes.

host-contracts/contracts/InputVerifier.sol
```diff
@@ -326,34 +368,85 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+        address[] memory coprocessorSigners = getCoprocessorSigners(contextId);
+        for (uint256 i = 0; i < coprocessorSigners.length; i++) {
+            if (coprocessorSigners[i] == account) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    function updateActiveCoprocessorContext(uint256 contextId, address[] calldata signers) public virtual onlyACLOwner {
```
jatZama: I am really confused by this implementation. I don't see the need of this function.
What I had in mind was that the admin would NOT be able to update active and suspended contexts independently, this seems extremely risky. What should be done instead I think: 
- remove both `updateActiveCoprocessorContext` and `updateSuspendedCoprocessorContext` 
- add `addNewContextAndSuspendOldOne` , the name is explicit, it would move previous context to the suspended state and move new one to the active state.
- add `removeSuspendedContext` , it would mark the suspended one as inactive.

Maybe @RomanBredehoft can confirm.

host-contracts/contracts/InputVerifier.sol
```diff
@@ -113,14 +119,25 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+        // TODO: remove this for mainnet.
         uint256 threshold; /// @notice The threshold for the number of signers required for a signature to be valid
+        /// @notice Current active coprocessor context ID.
+        uint256 activeCoprocessorContextId;
+        /// @notice List of signers for the current active coprocessor context ID.
+        address[] activeCoprocessorContextSigners;
+        /// @notice Suspended coprocessor context ID.
+        uint256 suspendedCoprocessorContextId;
+        /// @notice List of signers for the suspended coprocessor context ID.
+        address[] suspendedCoprocessorContextSigners;
```
jatZama: For security reason, I think we should also add a mapping `mapping(uint256=>bool) alreadyUsedContextId` to check that a contextId which was used at least once in the past could never be reused again, otherwise this would mean potentially that old signed inputs could be reused maybe in the future if at some point (maybe by mistake) an old contextId was reused in the future.

host-contracts/contracts/InputVerifier.sol
```diff
@@ -326,34 +368,85 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
 
     /**
-     * @notice          Returns the list of signers.
+     * @notice          Returns the list of signers of a specific context ID.
      * @dev             If there are too many signers, it could be out-of-gas.
+     * @param contextId The context ID of the signer addresses to return.
      * @return signers  List of signers.
      */
-    function getCoprocessorSigners() public view virtual returns (address[] memory) {
+    function getCoprocessorSigners(uint256 contextId) public view virtual returns (address[] memory) {
```
jatZama: can you make this function maybe also return a boolean `isSuspended` ? not just the list of signers. Could be helpful for debugging.

host-contracts/contracts/InputVerifier.sol
jatZama: I am wondering if you could not do something more gas efficient
because looping on 2 whole arrays and doing SSTORES inside the loop is so expensive
maybe you can just have 2 mappings:
- `mapping(contextId=>CoprocesorContextState) contextStates` (also to be used to avoid old contextId to be reused) where `CoproceesorContextState` would be an enum with all possible states:
`enum { NeverUsed, Active, Suspended, Removed }`.
- `mapping(contextId=>address[]) coprocessorContextArrays`

Wdyt?

jatZama: Like this you avoid half of the SSTOREs, you dont need to copy in storage already stored coprocessors from previous context

host-contracts/contracts/InputVerifier.sol
melanciani: this will be removed as we'll no longer put the contextId in the extraData, as updated here : https://github.com/zama-ai/fhevm-internal/issues/112

host-contracts/contracts/InputVerifier.sol
melanciani: are you sure about this ? I feel like we should keep this function simple. We could provide another function for getting the status for debugging if really needed

host-contracts/contracts/InputVerifier.sol
```diff
@@ -72,6 +69,26 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
     /// @notice Returned when signatures verification fails.
     error SignaturesVerificationFailed();
 
+    /// @notice Returned when the set of signers is empty.
+    error EmptyContextSignerAddresses(uint256 contextId);
```
melanciani: I find `EmptyCoprocessorSignerAddresses` more explicit and aligned with gateway names !

host-contracts/contracts/InputVerifier.sol
```diff
@@ -72,6 +69,26 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+
+    /// @notice Returned when the context ID is zero.
+    error ZeroContextId();
```
melanciani: in gateway I use `InvalidNullContextId`, as we never really name errors using `Zero`, but could be just `NullContextId`

host-contracts/contracts/InputVerifier.sol
```diff
@@ -72,6 +69,26 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+
+    /// @notice Returned when the context ID is not marked as active or suspended.
+    error InvalidContextId(uint256 contextId);
+
+    /// @notice Returned when the context ID is already used (i.e., not in the Unused state).
+    error ContextAlreadyUsed(uint256 contextId);
+
+    /// @notice The state of a coprocessor context ID.
+    enum CoprocessorContextState {
```
melanciani: I think it would be best to align with [the gateway PR](https://github.com/zama-ai/fhevm/pull/211/files#diff-a72e4b0c21d92619913077c97f9bc86e7990c469eeafd726f20330882f88123eR5) and use `NotInitialized` instead of us `Unused`

host-contracts/contracts/InputVerifier.sol
```diff
@@ -135,31 +163,59 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, EI
...
+     * @param initialContextId Initial active context ID.
+     * @param initialContextSigners Initial list of signers for the active context ID.
      */
     /// @custom:oz-upgrades-validate-as-initializer
     function initializeFromEmptyProxy(
         address verifyingContractSource,
         uint64 chainIDSource,
-        address[] calldata initialSigners
+        uint256 initialContextId,
+        address[] calldata initialContextSigners
```
melanciani: like https://github.com/zama-ai/fhevm/pull/875/files#r2372782609, I think `initialCoprocessorSigners` is more explicit

host-contracts/contracts/InputVerifier.sol
jatZama: Sorry `isSuspended` was wrong. I think we should return actually `isActiveOrSuspended` here to know if the input would revert or not.
If `isActiveOrSuspended` is `true` then the relayer-sdk signature check will pass, otherwise, it should raise an error.

host-contracts/contracts/InputVerifier.sol
melanciani, line 93: maybe take the opportunity of this PR to remove this struct (+ update associated functions) ? like in gateway, it's not really necessary

host-contracts/contracts/InputVerifier.sol
```diff
@@ -112,9 +111,20 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
 
     /// @custom:storage-location erc7201:fhevm.storage.InputVerifier
     struct InputVerifierStorage {
+        // TODO: remove this for mainnet.
```
melanciani: let's do like in gateway, for ex here : 
```
// DEPRECATED
mapping(address => bool) isSigner; 
```
or maybe
```
// DEPRECATED: remove in next state reset
mapping(address => bool) isSigner; 
```
wdyt cc @jatZama ?

because we won't remove this for mainnet

host-contracts/contracts/InputVerifier.sol
```diff
@@ -211,28 +231,42 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+            if (inputProof.length == 0) {
+                revert EmptyInputProof();
+            }
             uint256 numHandles = uint256(uint8(inputProof[0]));
             uint256 numSigners = uint256(uint8(inputProof[1]));
 
+            // Extract the coprocessor context ID from inputProof.
+            uint256 coprocessorContextId;
+            assembly {
+                coprocessorContextId := mload(add(inputProof, 0x22)) // 0x20 offset for array prefix + 2 bytes for numHandles and numSigners
```
melanciani, line 255: ok great, just updated https://github.com/zama-ai/relayer-sdk/issues/311 for making relayer-sdk aware of the expected encoding

host-contracts/contracts/InputVerifier.sol
```diff
@@ -243,7 +277,7 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
             for (uint256 j = 0; j < numSigners; j++) {
                 signatures[j] = new bytes(65);
                 for (uint256 i = 0; i < 65; i++) {
-                    signatures[j][i] = inputProof[2 + 32 * numHandles + 65 * j + i];
+                    signatures[j][i] = inputProof[34 + 32 * numHandles + 65 * j + i];
```
melanciani, line 292: worth adding a small comment on why this encoding , like in l268 !

host-contracts/contracts/InputVerifier.sol
```diff
@@ -271,7 +306,7 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
             if (numHandles <= indexHandle || indexHandle > 254) revert InvalidIndex();
             uint256 element;
             for (uint256 j = 0; j < 32; j++) {
-                element |= uint256(uint8(inputProof[2 + indexHandle * 32 + j])) << (8 * (31 - j));
+                element |= uint256(uint8(inputProof[34 + indexHandle * 32 + j])) << (8 * (31 - j));
```
melanciani: worth adding a small comment on why this encoding , like in l268 !

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+     * @dev If there are too many signers, it could be out-of-gas.
+     * @param coprocessorContextId The coprocessor context ID of the signer addresses to return.
+     * @return signers  List of signers.
      */
-    function addSigner(address signer) public virtual onlyACLOwner {
-        if (signer == address(0)) {
-            revert SignerNull();
-        }
-
+    function getCoprocessorSigners(uint256 coprocessorContextId) public view virtual returns (address[] memory) {
```
melanciani: I think @jatZama suggested that this functions also returns a boolean to indicate if the context is active or suspended (true), or not (false)

he suggested `isActiveOrSuspended` name but we could maybe use a simpler name/definition, I suggest maybe `isOperating`

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+     * @notice Get the threshold for required signatures.
+     * @param coprocessorContextId The coprocessor context ID of the signer addresses to get the threshold from.
+     * @return threshold Threshold for number of signatures verification.
      */
-    function removeSigner(address signer) public virtual onlyACLOwner {
-        InputVerifierStorage storage $ = _getInputVerifierStorage();
-        if (!$.isSigner[signer]) {
-            revert NotASigner();
-        }
+    function getThreshold(uint256 coprocessorContextId) public view virtual returns (uint256) {
```
melanciani: ~~I suggest we also add a boolean iin return, same as getCoprocessorSigners, using isOperating (or isActiveOrSuspended if we prefer) , wdyt @jatZama~~

EDIT: maybe not in the end, let's only add it to getCoprocessorSigners as this will be the most used functions, and lets keep the other functions transparent from this

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
-                _applyThreshold();
-                emit SignerRemoved(signer);
-                return;
+        // The majority threshold is the number of coprocessors that is required to validate consensus.
+        // It is currently defined as a strict majority within the coprocessor context (50% + 1).
+        return coprocessorSigners.length / 2 + 1;
+    }
+
+    /**
+     * @notice Returns whether the account address is a valid signer.
```
melanciani: let's update the comments : `Returns whether the account address is a valid signer within a context.`

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+     * @param coprocessorContextId The coprocessor context ID of the signer addresses to check the signer against.
+     * @param account Account address.
+     * @return isSigner Whether the account is a valid signer.
```
melanciani: same : `Whether the account is a valid signer within a context.`

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+    function isCoprocessorContextActiveOrSuspended(uint256 coprocessorContextId) public view virtual returns (bool) {
         InputVerifierStorage storage $ = _getInputVerifierStorage();
-        return $.signers;
+        if (
+            $.coprocessorContextStates[coprocessorContextId] == CoprocessorContextState.Active ||
+            $.coprocessorContextStates[coprocessorContextId] == CoprocessorContextState.Suspended
+        ) {
+            return true;
+        }
+        revert InvalidContextId(coprocessorContextId);
```
melanciani: why revert here ? I think we should return False instead, especially now that we trey to avoid reverting in view functions

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
     /**
-     * @notice          Returns the list of signers.
-     * @dev             If there are too many signers, it could be out-of-gas.
-     * @return signers  List of signers.
+     * @notice Returns whether the coprocessor context ID is active or suspended.
+     * @param coprocessorContextId The coprocessor context ID to check.
+     * @return isActiveOrSuspended Whether the coprocessor context ID is active or suspended.
      */
-    function getCoprocessorSigners() public view virtual returns (address[] memory) {
+    function isCoprocessorContextActiveOrSuspended(uint256 coprocessorContextId) public view virtual returns (bool) {
```
melanciani: do we need to make this function public ? imo we could remove this and instead expose a more general getter `getCoprocessorContextStatus` that returns the status of a context

host-contracts/contracts/InputVerifier.sol
```diff
@@ -112,9 +111,20 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+        // TODO: remove this for mainnet.
         address[] signers; /// @notice Array to keep track of all signers
+        // TODO: remove this for mainnet.
         uint256 threshold; /// @notice The threshold for the number of signers required for a signature to be valid
+        /// @notice Current active coprocessor context ID.
+        uint256 activeCoprocessorContextId;
+        /// @notice Suspended coprocessor context ID.
+        uint256 suspendedCoprocessorContextId;
+        /// @notice Mapping to keep track of coprocessor context states.
+        mapping(uint256 contextId => CoprocessorContextState contextState) coprocessorContextStates;
```
melanciani: in the gateway PR and specs, we talk about `status` instead of `state` !

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+        }
+        revert InvalidContextId(coprocessorContextId);
     }
 
-    /**
-     * @notice              Get the threshold for signature.
-     * @return threshold    Threshold for signature verification.
-     */
-    function getThreshold() public view virtual returns (uint256) {
+    function addNewContextAndSuspendCurrentOne(
```
melanciani: I think naming this function `addCoprocessorContext` like in gateway is enough. The fact that it also suspends the current one: 
- is more of an internal logic
- is not always true: most of the time, there won't be any suspended context anyway (as it already have been deactivated)

host-contracts/contracts/InputVerifier.sol
melanciani: we should have a reinitializeV2 ! you can make the current signers as the first context with context ID 1

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
         InputVerifierStorage storage $ = _getInputVerifierStorage();
-        return $.threshold;
+
+        // Check that the new context ID is not already used.
+        if ($.coprocessorContextStates[newContextId] != CoprocessorContextState.NotInitialized) {
+            revert ContextAlreadyInitialized(newContextId);
+        }
+
+        // Suspend the current active context.
+        $.coprocessorContextStates[$.activeCoprocessorContextId] = CoprocessorContextState.Suspended;
```
melanciani: I think it would be best to define the following functions as internal, to avoid forgetting steps : 
- `_suspendContext`
- `_activateContext`
- `_deactivateContext`

for ex `_activateContext` could be re-used in ini functions + here

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+        $.activeCoprocessorContextId = newContextId;
     }
 
-    /**
-     * @notice              Returns whether the account address is a valid signer.
-     * @param account       Account address.
-     * @return isSigner     Whether the account is a valid signer.
-     */
-    function isSigner(address account) public view virtual returns (bool) {
+    function removeSuspendedCoprocessorContext() public virtual onlyACLOwner {
```
melanciani: let's follow the same names as in the gateway : `deactivateSuspendedCoprocessorContext`

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
         InputVerifierStorage storage $ = _getInputVerifierStorage();
-        return $.isSigner[account];
+
+        // Mark the suspended context as deactivated.
+        $.coprocessorContextStates[$.suspendedCoprocessorContextId] = CoprocessorContextState.Deactivated;
```
melanciani: this function should: 
- only write if `suspendedCoprocessorContextId != 0` (no need to rewrite each time if there is no suspended context)
- if it is !=0, then :
    - reset `suspendedCoprocessorContextId` to `0` as well
    - emit an event

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+        }
+
+        // Suspend the current active context.
+        $.coprocessorContextStates[$.activeCoprocessorContextId] = CoprocessorContextState.Suspended;
+        $.suspendedCoprocessorContextId = $.activeCoprocessorContextId;
+
+        // Activate the new context.
+        $.coprocessorContextStates[newContextId] = CoprocessorContextState.Active;
+        $.coprocessorContextSigners[newContextId] = newContextSigners;
+        $.activeCoprocessorContextId = newContextId;
```
melanciani: this function should also emit 2 events : 
- for context suspension
- for context activation

host-contracts/contracts/InputVerifier.sol
melanciani: probably better to rename this function `getCoprocessorThreshold` btw

host-contracts/contracts/InputVerifier.sol
```diff
@@ -280,79 +315,95 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+        return coprocessorSigners.length / 2 + 1;
+    }
+
+    /**
+     * @notice Returns whether the account address is a valid signer.
+     * @param coprocessorContextId The coprocessor context ID of the signer addresses to check the signer against.
+     * @param account Account address.
+     * @return isSigner Whether the account is a valid signer.
+     */
+    function isSigner(uint256 coprocessorContextId, address account) public view virtual returns (bool) {
```
melanciani, line 380: does this function need to be public cc @jatZama ?

host-contracts/tasks/upgradeContracts.ts
```diff
@@ -251,7 +251,19 @@ task('task:upgradeInputVerifier')
     }
     const proxyAddress = getRequiredEnvVar('INPUT_VERIFIER_CONTRACT_ADDRESS');
 
-    await upgradeCurrentToNew(proxyAddress, currentImplementation, newImplementation, verifyContract, hre);
+    // Load the initial coprocessor context from environment variables.
+    let initialContextSigners: string[] = [];
+    const numSigners = getRequiredEnvVar('NUM_COPROCESSORS');
+    const initialContextId = getRequiredEnvVar('COPROCESSOR_CONTEXT_ID');
```
melanciani: no need to provide the context ID for the reinitializer I think, it will be 1 at first

melanciani (changes requested):
great !! made several comments, mostly to align the logic and names with what has been done in the gateway PR !

host-contracts/contracts/InputVerifier.sol
melanciani: minor version should be updated !

melanciani: reinitializer version should be updated !

isaacdecoded: Made a first approach, but removing it affects the `verifyInput` method, which easily makes the contract compilation fail with the "stack too deep" error. Created this issue here to revisit later: https://github.com/zama-ai/fhevm-internal/issues/472

host-contracts/contracts/InputVerifier.sol
melanciani: created https://github.com/zama-ai/fhevm-internal/issues/484 for that

host-contracts/contracts/InputVerifier.sol
melanciani: i think we can keep it public yes

host-contracts/contracts/InputVerifier.sol
```diff
@@ -134,22 +162,19 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
-        uint256 initialSignersLen = initialSigners.length;
-        if (initialSignersLen == 0) {
-            revert InitialSignersSetIsEmpty();
-        }
-        for (uint256 i = 0; i < initialSignersLen; i++) {
-            addSigner(initialSigners[i]);
-        }
+
+        // Activate the initial context.
+        _activateContext(1, initialCoprocessorSigners);
```
melanciani: no this is wrong : someone might deploy the contract one day in the future and the contextId might be much greater in the gateway : the `contextId` should be an input to `initializeFromEmptyProxy` 

only `reinitializeV2` should not consider it, as it will be used only to upgrade the preivous version (without context) to the new one (so with initial context)

melanciani (changes requested):
great work ! have one request changes that needs to be addressed

isaacdecoded:
Outdated implementation.

PR #877, opened 2025-09-22
chore(common): upgrade rust golden image

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/430


golden-container-images/rust-glibc/Dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM cgr.dev/zama.ai/glibc-dynamic:15.2.0-dev AS builder
+FROM cgr.dev/zama.ai/glibc-dynamic:15.2.0-dev AS prod
```
aquint-zama, line 1: would it not be easier to drop our own golden image in favor of chainguard rust image?

eudelins-zama: I'm not fully aware of why these golden images were created in the first place. From what I read, it's standard practice to make every deps installation  (`apk install`, ...) needed for all services' Dockerfile in a single place, to avoid each service does it on its own.
So if this is the reason we decided to use a golden image, then using Chainguard image won't help here. Maybe @tawadaa could answer but she's off right now.

chilcano: You still can use chainguard image and install packages using `apk install` as a golden image, however, adding external packages could introduce vulns a common image used for many services. We should be careful with golden images.

Best practices: https://www.notion.so/zamaai/Best-practices-for-Dockerfile-1355a7358d5e80e1b366f78fe5326773

eudelins-zama: But this is what we are already doing right @chilcano ? Or am I missing something here ?

fegmorte: As @eudelins-zama said we use chainguard based image to create a golden rust image with our dependencies to avoid redundancies accross build. So it's still necessary to maintain this image. @chilcano this image is scanned and it doesn't contain vulns anymore in kms and maybe there is in fhevm but btw it's not deplooyed in production. It's just like a stage image and the vulns don't end up in the final image.
But be careful @eudelins-zama you're re introduce what i removed in my pr regarding SCCACHE

eudelins-zama: @fegmorte thanks for the clarification !

I don't think I re-introduce what you removed, once my PR will be rebased on top of yours, SSCACHE will be removed as I did modify the associated lines in my PR. At worse we could have conflict, but here I don't think it would be the case :wink:

chilcano (approved):
LGTM

PR #889, opened 2025-09-24
chore(kms-connector): enable tls for sqlx

eudelins-zama:
e2e tests are passing: https://github.com/zama-ai/fhevm/actions/runs/17976964176/job/51132743478

### DIFF
--- a/kms-connector/Cargo.lock
+++ b/kms-connector/Cargo.lock
@@ -7078,6 +7078,7 @@ dependencies = [
  "memchr",
  "once_cell",
  "percent-encoding",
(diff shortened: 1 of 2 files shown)


aquint-zama:
there is no configuration file to update?

eudelins-zama:
@aquint-zama Not on kms-connector side. Everything is handled via the postgres URL, which already exists in the config

On infra side, we might add `?sslmode=require` suffix in DB urls to enforce TLS, but it might not even be needed

aquint-zama (approved):

PR #891, opened 2025-09-24
ci(test-suite): add e2e tests orchestrator

tawadaa:


.github/workflows/trigger-all-builds.yml
```diff
@@ -0,0 +1,84 @@
+name: trigger-all-builds
+
+on:
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: "Reason for triggering all builds"
+        default: "Manual trigger for e2e testing"
+        type: string
```
eudelins-zama: Is that an AI hallucination ? What's this `reason` input otherwise ?

.github/workflows/orchestrate-e2e-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            commit_sha="${{ github.event.pull_request.head.sha }}"
+            echo "all-builds-complete=true" >> $GITHUB_OUTPUT
+            echo "image-tag=${commit_sha:0:7}" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+          
+          commit_sha="${{ github.event.workflow_run.head_sha }}"
+          echo "image-tag=${commit_sha:0:7}" >> $GITHUB_OUTPUT
```
eudelins-zama: I think something's wrong with the overall logic around image tag.

From what I understood, you trigger every workflows that build docker images. But if the "check-change" job of one service's workflow does not detect any change, then the image of the service won't be built, correct?

Thus the image tag you gather here won't work for services that did not built their image. We need a way to **really** enforce the built of every docker images, and in that case, what you did could work.

But it means we would build some images for nothing... (and thus it would require more logic, or dedicated tags or whatever to use the correct image for each service)

tawadaa: good catch, the trigger-all-builds point is to simplify, but you are absolutly right on the fact that some will be skipped because of the check-changes status

tawadaa:
Outdated

PR #892, opened 2025-09-24
chore(kms-connector): deps upgrade

eudelins-zama:
Few change were required to adapt to:
- `kms-grpc`: placeholder where added for upcoming features
- `alloy`: the behavior of the mock provider was updated, the event stream is never closed, so I had to use a separated task etc. for unit tests


dartdart26 (approved):
Looks good!

PR #896, opened 2025-09-25
feat(host-contracts): add PauserSet for host

jatZama:


host-contracts/.env.example
```diff
@@ -19,6 +19,16 @@ ETHERSCAN_API_KEY="XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
...
+
+# Pauser 1
+PAUSER_ADDRESS_0="0xa44366bAA26296c1409AD1e284264212029F02f1" # accounts[2] (address)
+
+# Pauser 2
+PAUSER_ADDRESS_1="0xc1d91b49A1B3D1324E93F86778C44a03f1063f1b" # accounts[3] (address)
+
+# The first pauser's private key
+# This is required for local tests and running the pausing task. It must correspond to one of the pauser's private key
+PAUSER_PRIVATE_KEY="0x7ae52cf0d3011ef7fecbe22d9537aeda1a9e42a0596e8def5d49970eb59e7a40" # accounts[22], private key (bytes32)
```
melanciani: small detail but inline comment should say `accounts[2]` not `accounts[22]`

melanciani (approved):
thanks !! looks good, just.a small detail to clean

host-contracts/.env.example
jatZama: thanks, good catch, will fix now

PR #919, opened 2025-09-29
feat(coprocessor): add transaction-based performance metrics 

goshawk-3:
### TLDR:

- Adds a **transaction_id** to all major OpenTelemetry spans across all services, enabling the collection of per-operation and per-service timing data for each L1 transaction.
- Introduces  a new histogram metric (**coprocessor_l1_txn_latency_seconds**) to track latency of L1 transactions (mint, transfer, etc ..)
-  Introduces  a new histogram metric (**coprocessor_zkproof_txn_latency_seconds**) to track  latency of zkproof transactions

Implementation details:
- A new table `transactions`  stores both `created_at` and `completed_at` timestamp of a single (L1 or L2) transaction:
-  Tables (allowed_handles, pbs_computations, verify_proofs and ciphertext_digest)  include a new column - `transaction_id`
- Both listeners (gw-listener and host-listener) register the transaction arrival time
- Each worker and sender can read the transaction_id from any of PG tables to populate the OTEL spans.
- The transaction_sender register the time of completion of a transaction


coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
```diff
@@ -127,3 +156,287 @@ pub fn end_span_with_err(mut span: BoxedSpan, desc: String) {
...
+            chain_id,
+            block_number as i64
+        )
+        .execute(pool)
+        .await?;
+
+        // Do some GC
+        sqlx::query!(
+            r#"
+                DELETE FROM transactions
```
rudy-6-4, line 280: without index it's a linear scan over 1 day of all the transactions

rudy-6-4: you can GC irrespective of chain_id ? so it needs only 1 index on created_at

coprocessor/fhevm-engine/db-migration/migrations/20250929064611_create_transactions_table.sql
```diff
@@ -0,0 +1,27 @@
...
+ALTER TABLE verify_proofs
+ADD COLUMN IF NOT EXISTS transaction_id bytea NULL;
+CREATE INDEX IF NOT EXISTS idx_verify_proofs_transactions ON verify_proofs USING HASH (transaction_id);
+
+-- Add transaction_id column to ciphertext_digest (if not present)
+ALTER TABLE ciphertext_digest
+ADD COLUMN IF NOT EXISTS transaction_id bytea NULL;
+CREATE INDEX IF NOT EXISTS idx_ciphertext_digest_transactions ON ciphertext_digest USING HASH (transaction_id);
+
+CREATE TABLE transactions (
```
rudy-6-4, line 21: need some index to support the existing query efficiently

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -456,9 +469,18 @@ impl Database {
         &self,
         tx: &mut Transaction<'_>,
         event: &Log<AclContractEvents>,
+        transaction_hash: &Option<Handle>,
```
rudy-6-4, line 476: event + transaction_hash + block_number could be a struct (like struct LogTfhe) shared or not (but less useful than for tfhe).

rudy-6-4, line 476: event + transaction_hash + block_number could be a struct (like struct LogTfhe) shared or not (but less useful than for tfhe).

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -363,6 +365,17 @@ impl Database {
         let insert_computation_bytes = |tx, result, dependencies_handles, dependencies_bytes, scalar_byte| {
             self.insert_computation_bytes(tx, tenant_id, result, dependencies_handles, dependencies_bytes, fhe_operation, scalar_byte, log)
         };
+
+        let _t = telemetry::tracer(
+            "handle_tfhe_event",
+            &log.transaction_hash.map(|h| h.to_vec()),
+        );
+
+        self.record_transaction_begin(
```
rudy-6-4, line 378: should be done at the block level and not the event level. Ie in db_insert_block_no_retry once for each transaction_hash
(could be bulk insert later as improvement)

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -363,6 +365,17 @@ impl Database {
         let insert_computation_bytes = |tx, result, dependencies_handles, dependencies_bytes, scalar_byte| {
             self.insert_computation_bytes(tx, tenant_id, result, dependencies_handles, dependencies_bytes, fhe_operation, scalar_byte, log)
         };
+
+        let _t = telemetry::tracer(
```
rudy-6-4, line 373: could be done once at block level ?

coprocessor/fhevm-engine/transaction-sender/src/ops/add_ciphertext.rs
```diff
@@ -340,6 +350,9 @@ where
 
         let mut join_set = JoinSet::new();
         for row in rows.into_iter() {
+            let transaction_id = row.transaction_id.clone();
+            let t = telemetry::tracer("prepare_add_ciphertext", &transaction_id); // TODO: maybe move to the task
```
rudy-6-4: clean the comment ?

rudy-6-4 (commented):
Not sure about GC transactions table performance. Could be too slow if called too often.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
goshawk-3: There is a particular reason to call it inside `insert_tfhe_event`.

I wanted to make ensure that the stress-test tool that calls `insert_tfhe_event` also registers the transaction.

If there is any perf impact, we can refactor this into a batch insert.

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
goshawk-3: The chain_id is here so that each listener (either gw-listener or host-listener) cleans up its own transactions but you're right that this can gradually become slow over time.

I'll remove the chain_id for now.

coprocessor/fhevm-engine/db-migration/migrations/20250929064611_create_transactions_table.sql
goshawk-3: Which query do you mean here?

coprocessor/fhevm-engine/transaction-sender/src/ops/add_ciphertext.rs
goshawk-3: Done

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: to be done later if useful

rudy-6-4 (approved):

PR #921, opened 2025-09-29
chore(kms-connector): use alloy chain id filler

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/447

e2e tests run: https://github.com/zama-ai/fhevm/actions/runs/18091651538/job/51473480301


dartdart26 (approved):
Looks good!

Could we get the chain ID from the provider? Or we want to make sure the correct one is set in conf?

eudelins-zama:
> Could we get the chain ID from the provider? Or we want to make sure the correct one is set in conf?

@dartdart26 We want to make sure the correct one is set in conf, otherwise the txs would all fail because EIP-712 would break. So it is fine relying on the one in the conf for the filler

PR #924, opened 2025-09-29
chore(test-suite): call addPausers in host and gateway docker compose

melanciani:
docker image `bef879c` for host-contracts comes from this build : https://github.com/zama-ai/fhevm/actions/runs/18042117265 (last docker build on main for host-contracts)


jatZama (approved):
Ok but maybe we should add an addPausers task run as well, I think it might be needed by the infra team at some point, wdyt?

jatZama (approved):

jatZama:
LGTM

PR #925, opened 2025-09-29
feat(library-solidity): add feature to request decryption without sav…

immortal-tofu:
…ing handles


host-contracts/codegen/templates.ts
```diff
@@ -1130,7 +1153,7 @@ function generateSolidityDecryptionOracleMethods(fheTypes: AdjustedFheType[]): s
         bytes32[] memory handlesList,
         bytes memory cleartexts,
         bytes memory decryptionProof
-    ) private returns (bool) {
+    ) internal returns (bool) {
```
jatZama, line 1159: Ok I see the point of making this low level function internal; but if you do so at least you should change the first word in the comment above the function.

jatZama: Since this function is made internal now, I notice something very painful for dev tooling (i.e  mocked mode): the mocked relayer would no longer work, meaning onchain async decrypt will no longer work in mocked mode, because you can no longer emit this event `DecryptionFulfilled(requestID);`  which is really mandatory at the moment with how we use async decrypt in mocked mode.

immortal-tofu: Indeed, I see your point. Would it be possible to include an other event in `verifySignatures` which could be used in mock mode? cc @0xalexbel

jatZama: ah no I think in mocked mode the feature should still work, but on a real chain since we cannot detect if DecryptionFulfilled is emitted or not, it won't work and we will have to sleep instead of waiting for n event in this case, or otherwise emit it the event manually.

jatZama: I think this is fine, we can say the scope of mocked mode is only the hardhat local node, so there is no issue here.
If we really need to use some kind of utility for the end to end test, we can either sleep , or emit this event manually in case we use the `verifySignatures` instead of the `checkSignatures` function. I think it is fine for us.

host-contracts/codegen/templates.ts
```diff
@@ -1079,10 +1079,33 @@ function generateSolidityDecryptionOracleMethods(fheTypes: AdjustedFheType[]): s
...
+    /**
+     * @dev     Calls the DecryptionOracle contract to request the decryption of a list of handles, with a custom msgValue.
+     * @notice  Also does the needed call to ACL::allowForDecryption with requested handles.
+     */
+    function requestDecryption(
+        bytes32[] memory ctsHandles,
+        bytes4 callbackSelector,
+        uint256 msgValue
+    ) internal returns (uint256 requestID) {
+      requestID = requestDecryptionWithoutSavingHandles(ctsHandles, callbackSelector, 0);
```
jatZama: There is an important issue here, `msgValue` should replace the 0 as last argument.

immortal-tofu:
LGTM

0xalexbel (approved):

PR #932, opened 2025-09-30
chore(gateway-contracts): clean and rename KmsManagement to KMSGeneration

melanciani:
closes https://github.com/zama-ai/fhevm-internal/issues/377


test-suite/fhevm/env/staging/.env.gateway-sc
```diff
@@ -5,6 +5,7 @@ HARDHAT_NETWORK=staging
 RPC_URL=http://gateway-node:8546
 CHAIN_ID_GATEWAY=54321
 GATEWAY_CONFIG_ADDRESS=0x812b06e1CDCE800494b79fFE4f925A504a9A9810
+KMS_GENERATION_ADDRESS=0x35760912360E875DA50D40a74305575c23D55783
```
melanciani, line 8: this is needed because we should not use internal addresses in docker-compose (same as add host chains) !

gateway-contracts/contracts/KMSGeneration.sol
```diff
@@ -145,28 +158,18 @@ contract KMSManagement is
...
 
     /**
      * @dev Storage location has been computed using the following command:
-     * keccak256(abi.encode(uint256(keccak256("fhevm_gateway.storage.KMSManagement")) - 1))
+     * keccak256(abi.encode(uint256(keccak256("fhevm_gateway.storage.KMSGeneration")) - 1))
      * & ~bytes32(uint256(0xff))
      */
-    bytes32 private constant KMS_MANAGEMENT_STORAGE_LOCATION =
+    bytes32 private constant KMS_GENERATION_STORAGE_LOCATION =
         0x52e3d903674697c366245bdc532b6735f22bb42391635b8e0488806aba29452b;
```
isaacdecoded: Shouldn't the hash also change? I think it's: `0x0b8fdb1f0a6356dd20a6cbc6f9668fac23b85f96575d10e333e603faa794ac00`

melanciani: good catch thanks

PR #957, opened 2025-10-01
chore(gateway-contracts): replace check view funcs by boolean ones

melanciani:
docker images: 
- gateway-contracts: `b3736bf` from [this build](https://github.com/zama-ai/fhevm/actions/runs/18194417345)
- fhevm-relayer: `ce851ac` from [this build](https://github.com/zama-ai/console/actions/runs/18194502150) based on [this PR](https://github.com/zama-ai/console/pull/684)

note: this breaks the relayer in 2 manners : 
- retry mechanism for decryption requests: `check..` functions in `Decryption` contract no longer exist
- bindings: some errors have been moved to other contracts, or simply removed

closes https://github.com/zama-ai/fhevm-internal/issues/421


gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -104,13 +103,7 @@ contract CiphertextCommits is ICiphertextCommits, UUPSUpgradeableEmptyProxy, Gat
         uint256 keyId,
         bytes32 ciphertextDigest,
         bytes32 snsCiphertextDigest
-    ) external virtual onlyCoprocessorTxSender {
-        // Extract the chainId from the ciphertext handle
-        uint256 chainId = HandleOps.extractChainId(ctHandle);
```
jatZama, line 109: Why was this removed? How is it linked to replacing reverting view functions by returning boolean ones?

melanciani: see the line below, it's more about cleaning this as we already exposed the `onlyHandleFromRegisteredHostChain` modifier for this very purpose

(and I realized this because `checkHostChainIsRegistered` has now been removed from `GATEWAY_CONFIG`)

jatZama: ok thanks, i missed the modifier.

gateway-contracts/contracts/interfaces/IDecryption.sol
```diff
@@ -282,45 +276,48 @@ interface IDecryption {
...
         DelegationAccounts calldata delegationAccounts,
         CtHandleContractPair[] calldata ctHandleContractPairs,
         address[] calldata contractAddresses,
         bytes calldata extraData
-    ) external view;
+    ) external view returns (bool);
 
     /**
-     * @notice Checks if a (public, user, delegated user) decryption is done.
+     * @notice Indicate if a (public, user, delegated user) decryption is done.
```
isaacdecoded: super-nitpick: I've seen [at some places](https://github.com/zama-ai/fhevm/pull/957/files#diff-14c62001db03729747b5ad7bc744639d1f05199712fca3b9e99187c3c47b4994R301) using "Indicates" and in others "Indicate"

melanciani: indeed good catch, I sometimes forget that

test-suite/benchmarks/tasks/shared.ts
```diff
@@ -110,36 +110,60 @@ export interface KeyPair {
...
 
 /**
- * Checks if a contract function reverts when called with the given inputs.
+ * Calls a contract function and returns the result.
  * @param contractAddress The address of the contract.
  * @param functionSignature The function signature, e.g. "myFunction(uint256,address)".
  * @param inputs The array of input values to encode and pass to the function.
  * @param provider The ethers provider.
  * @param ethers The ethers object.
  * @returns true if the function call reverts, false otherwise.
```
jatZama: this comment must be updated...

test-suite/benchmarks/tasks/shared.ts
```diff
@@ -110,36 +110,60 @@ export interface KeyPair {
...
-    await provider.call({
-      to: contractAddress,
-      data: data
-    });
-    return false;
-  } catch (error) {
-    return true;
+  // Decode the result
+  const decodedResult: Array<any> = await callFunction(contractAddress, functionSignature, inputs, provider, ethers);
+  if (decodedResult.length !== 1) {
```
jatZama: does this really make sense? remember that the return value is an array of bytes, so it returns a single byte this condition pass, but the byte could be something else than 0x00 or 0x01. I think this should be fixed.

jatZama (changes requested):
Contracts and tests are good.
I found two mistakes in the utils function: 
- small comment about reverting which should be changed to say we return boolean (instead of revert).
- logic error to check if the return value is boolean, it is not sufficient to check the length of bytes array, you must also validate this single byte is a boolean (0 or 1).

test-suite/benchmarks/tasks/shared.ts
melanciani: are you sure it's bytes ? it's decoded in `callFunction` 

but I talked with Simon and I believe this directory will be removed anyway ..

jatZama: I was not with Simon during this conversation, what was the topic? I dont get it

melanciani: I just asked him if this directory was still used for benchmarking, and it's not (we now use `gateway-stress` dir). I'll remove the directory in another PR

melanciani: reverted the changes and opened a PR to remove the `benchmarks` directory: https://github.com/zama-ai/fhevm/pull/965

jatZama (approved):

PR #959, opened 2025-10-02
chore(charts): update charts

antoniupop:


charts/coprocessor/values.yaml
```diff
@@ -218,6 +218,8 @@ tfheWorker:
       value: "0xa1880e99d86F081E8D3868A8C4732C8f65dfdB11"
   args:
     - --coprocessor-private-key=/accounts/coprocessor.hex
+    - --coprocessor-fhe-threads=64
```
maxnovawind, line 221: Do we need to tune these values for the testnet environment with the HPC7a.96xlarge instance?

maxnovawind (approved):

PR #963, opened 2025-10-02
feat(charts): add new args and env vars

tawadaa:


charts/kms-connector/values.yaml
```diff
@@ -1,85 +1,175 @@
...
+    #       name: connector-database
+    #       key: password
+
+    # =========================================================================
+    # NEW ENVIRONMENT VARIABLES
+    # =========================================================================
+    
+    # The address of the KMSGeneration gateway contract
+    - name: KMS_GENERATION_ADDRESS
+      value: ""
```
eudelins-zama: To remove, it's already in `kmsGeneration` field of `commonConfig`

charts/kms-connector/values.yaml
```diff
@@ -1,85 +1,175 @@
...
+    
+    # =========================================================================
+    # DEPRECATED ENVIRONMENT VARIABLES
+    # =========================================================================
+    # The following environment variables are no longer used:
+    # - KMS_MANAGEMENT_CONTRACT_ADDRESS
```
eudelins-zama: To remove, it was not used in any `values.yaml` files of the `gitops` repo AFAIK

I don't want to add too much irrelevant information in this file

charts/kms-connector/values.yaml
```diff
@@ -1,85 +1,175 @@
...
+# -----------------------------------------------------------------------------
+# Sets up the database schema for KMS connector
 kmsConnectorDbMigration:
   enabled: true
   annotations:
+  
   image:
     name: ghcr.io/zama-ai/fhevm/kms-connector/db-migration
-    tag: v0.8.0
+    tag: v0.9.0
```
eudelins-zama, line 69: I think the `README.md` of the kms-connector helm chart also need to be updated, because it currently states that default value for this is `v0.8.0`

charts/coprocessor/values.yaml
```diff
@@ -49,35 +87,50 @@ dbMigration:
...
   nameOverride:
+  
   image:
     name: ghcr.io/zama-ai/fhevm-coprocessor
-    tag: v0.7.0-rc2
+    tag: v0.9.0
+  
   replicas: 1
+  
+  # Environment variables
```
rudy-6-4, line 133: Don't we need ETHEREUM_RPC_URL, ACL_CONTRACT_ADDRESS, FHEVM_EXECUTOR_CONTRACT_ADDRESS as env var ?

charts/kms-connector/values.yaml
melanciani: done

charts/coprocessor/values.yaml
melanciani: I believe we've only added _new_ env vars. Are you referring to new ones as well ?

tawadaa: well, it would be nice if you can add missing values, you can sync with `fhevm-dev` env

rudy-6-4: No to exising ones (that are use as actual value for command line parameter).

tawadaa: @rudy-6-4 let's keep this version, we can update the existing missing values in another PR

melanciani (approved):
thanks !

isaacdecoded (approved):
LGTM! 💯

PR #974, opened 2025-10-02
docs(gateway-contracts): clarify new env vars

melanciani:


charts/contracts/values.yaml
```diff
@@ -44,7 +44,7 @@ scDeploy:
     # =========================================================================
 
     # KMS generation threshold
-    # Should be set to `floor(n_kms/2) + 1` with `n_kms` the number of KMS nodes
+    # Should be set to `floor(2*n_kms/3) + 1` with `n_kms` the number of KMS nodes
```
melanciani: @isaacdecoded keygen threshold should be 2/3rd, not 1/2 : https://github.com/zama-ai/tech-spec/blob/main/architecture/completed_intercomponent_flows/key_chain.md#step-7

I think we got confused with the majority threshold defined for _triggering_ the keygen : https://github.com/zama-ai/tech-spec/blob/main/architecture/definitions.md#gateway-contracts

isaacdecoded (approved):
Approved but edited: just remembered about this discussion [here](https://github.com/zama-ai/fhevm/pull/772#discussion_r2324926764) in which it was commented about a "suboptimal behavior when `n` is not exactly `3t + 1`" 🤔

cc @RomanBredehoft @kc1212

melanciani:
good point @isaacdecoded , we discussed with MPC team and decided to use `2t+1` instead in the end !

PR #980, opened 2025-10-03
chore(contracts): add verifyPauserSet task

melanciani:


jatZama (approved):
LGTM! Now let's test on a real blockexplorer (Sepolia for host and blockscout for gateway contracts).

PR #989, opened 2025-10-03
feat(coprocessor): dump ciphertext digests along with handles in stre…

antoniupop:
…ss generator

### DIFF
--- a/coprocessor/fhevm-engine/stress-test-generator/src/utils.rs
+++ b/coprocessor/fhevm-engine/stress-test-generator/src/utils.rs
@@ -272,6 +272,31 @@ pub async fn query_and_save_pks(
     Ok((pks, public_params))
 }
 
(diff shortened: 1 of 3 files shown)


coprocessor/fhevm-engine/stress-test-generator/src/utils.rs
```diff
@@ -272,6 +272,31 @@ pub async fn query_and_save_pks(
     Ok((pks, public_params))
 }
 
+pub async fn get_ciphertext_digests(
+    handle: &[u8],
+    pool: &sqlx::PgPool,
+) -> Result<(Vec<u8>, Vec<u8>), Box<dyn std::error::Error>> {
+    loop {
```
goshawk-3: Better restrict this loop using retries count limit so that it times out in case of S3 failures.

goshawk-3 (approved):

PR #1022, opened 2025-10-06
fix(kms-connector): use eth_sendRawTransactionSync

eudelins-zama:
e2e tests are passing: https://github.com/zama-ai/fhevm/actions/runs/18306478320/job/52124802024#step:10:11

Btw if you wonder why I had to use concrete `FillProvider` in `NonceManagedProvider`, it is partially explained by this issue I opened on `alloy`'s repo (https://github.com/alloy-rs/alloy/issues/2997)

### DIFF
--- a/kms-connector/crates/utils/src/provider.rs
+++ b/kms-connector/crates/utils/src/provider.rs
@@ -1,7 +1,7 @@
 use alloy::{
     consensus::Account,
-    eips::{BlockId, BlockNumberOrTag},
-    network::{Network, TransactionBuilder},
+    eips::{BlockId, BlockNumberOrTag, Encodable2718},
+    network::{Ethereum, Network, TransactionBuilder},
     primitives::{
         Address, B256, BlockHash, BlockNumber, Bytes, StorageKey, StorageValue, TxHash, U64, U128,
         U256,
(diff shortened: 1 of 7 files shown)


kms-connector/crates/utils/src/provider.rs
```diff
@@ -38,23 +39,68 @@ pub type FillersWithoutNonceManagement =
...
+            .await?
+            .try_into_envelope()
+            .map_err(|e| TransportError::LocalUsageError(Box::new(e)))?
+            .encode_2718(&mut tx_bytes);
+
+        let res = self
+            .client()
+            .request("eth_sendRawTransactionSync", (Bytes::from(tx_bytes),))
+            .await;
+        if res.is_err() {
```
dartdart26, line 90: Is the error here just for networking or also for say nonce mismatch?

dartdart26 (approved):
Looks good, just one question.

kms-connector/crates/utils/src/provider.rs
eudelins-zama: I think it is also for nonce mismatch

dartdart26: If that's the case, then it is good.

PR #1023, opened 2025-10-06
fix(db): improve safety and clarity in database connection and cache initialization

hariom2509:
- Updated 'new_pool' to safely parse the database URL using 'match' instead of '.expect("bad url")'.
  This prevents an ungraceful panic and logs a clear error message if the URL is invalid.
- Corrected the 'statement_timeout' option from "10000" (10s) to "5000" (5s) to align with the comment.
- Replaced '.unwrap()' with '.expect("capacity must be > 0")' in LRU cache initialization for clearer intent.
  Although 'MINIMUM_BUCKET_CACHE_SIZE' ensures nonzero values, this change improves readability
  and provides an explicit error message if future changes break that assumption.

These updates improve code robustness, prevent silent crashes, and make configuration intent clearer.


hariom2509:
Updated 'new_pool' to safely parse the database URL using 'match' instead of '.expect("bad url")'.
This prevents an ungraceful panic and logs a clear error message if the URL is invalid.
Corrected the 'statement_timeout' option from "10000" (10s) to "5000" (5s) to align with the comment.
Replaced '.unwrap()' with '.expect("capacity must be > 0")' in LRU cache initialization for clearer intent.
Although 'MINIMUM_BUCKET_CACHE_SIZE' ensures nonzero values, this change improves readability
and provides an explicit error message if future changes break that assumption.

aquint-zama:
@cla-bot check

PR #1026, opened 2025-10-07
feat(host-contracts): delegate user decrypt

jatZama:
Closes https://github.com/zama-ai/fhevm-internal/issues/474


host-contracts/contracts/ACL.sol
```diff
@@ -139,6 +223,93 @@ contract ACL is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, PausableUpgr
...
+     * @notice Delegates the access of handles, for instance, in the context of account
+     *  abstraction for issuing user decryption requests from a smart contract account.
+     * @param delegate The address of the account that receives the delegation.
+     * @param contractAddress The contract address to delegate access to.
+     * @param expiryDate Expiry date in seconds, between 1 hour and 1 year in the future.
+     */
+    function delegateAccount(
+        address delegate,
+        address contractAddress,
+        uint64 expiryDate
```
rudy-6-4, line 236: this is not a date. expiryDelay ?

jatZama: No this code is correct, it should be a date, as per what was discussed with @immortal-tofu , and I think personally it makes sense.
The FHE.sol library could easily handle the conversion between delay and date.

jatZama: A bit more details: 
I think one of the reasons for using date instead of delay is security. Imagine you broadcast a tx with too low gas price, it will stay in mempool maybe for several hours, so this could lead to unforeseen consequence, where delegation is given for longer than expected.

jatZama:
Good work, but I think you forgot to rerun selectors computation in your last commit, since you changed one of the custom errors signatures, right?
Aslo what was the rationale behind renaming NewDelegation event to DelegateAccount? I don't remember any discussion on this topic.

isaacdecoded:
> Good work, but I think you forgot to rerun selectors computation in your last commit, since you changed one of the custom errors signatures, right? 

I think the selectors are up-to-date already. 

> Aslo what was the rationale behind renaming NewDelegation event to DelegateAccount? I don't remember any discussion on this topic.

Basically, to follow the same convention we have been using for others events, I mean, using same/similar name as the contract method that triggers it. E.g., `RevokedDelegation` for `revokeDelegation()`, `AllowedForDecryption` for `allowForDecryption()` , etc.

host-contracts/contracts/ACL.sol
```diff
@@ -139,6 +223,93 @@ contract ACL is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, PausableUpgr
...
+            revert SenderCannotBeContractAddress(contractAddress);
+        }
+        if (delegate == msg.sender) {
+            revert SenderCannotBeDelegate(delegate);
+        }
+        if (delegate == contractAddress) {
+            revert DelegateCannotBeContractAddress(contractAddress);
+        }
+
+        uint64 newExpiryDate = uint64(block.timestamp) + expiryDate;
```
obatirou: Shouldn't this be `uint64 newExpiryDate = expiryDate;` ?
The doc says `Expiry date in seconds, between 1 hour and 1 year in the future`

The guards at the beginning of the function are 
```solidity
        if (expiryDate < block.timestamp + 1 hours) {
            revert ExpiryDateBeforeOneHour();
        }
        if (expiryDate > block.timestamp + 365 days) {
            revert ExpiryDateAfterOneYear();
        }
```
which is absolute and consistent with the comment

Let me know if I am missing something

jatZama: you are right, this must be changed, and we need additional Foundry tests to catch this bug first.

jatZama: this was due to confusion between delay and date, at first we wanted to use delay, then we switched to date for security

isaacdecoded (approved):
LGTM! 👍

host-contracts/contracts/ACL.sol
```diff
@@ -139,6 +223,93 @@ contract ACL is UUPSUpgradeableEmptyProxy, Ownable2StepUpgradeable, PausableUpgr
...
     }
 
+    /**
+     * @notice Delegates the access of handles, for instance, in the context of account
+     *  abstraction for issuing user decryption requests from a smart contract account.
+     * @param delegate The address of the account that receives the delegation.
+     * @param contractAddress The contract address to delegate access to.
+     * @param expiryDate Expiry date in seconds, between 1 hour and 1 year in the future.
+     */
+    function delegateAccount(
```
obatirou, line 233: As discussed IRL, should naming reflect it is only for user decryption?
(to apply where relevant in the file)

jatZama: I am not sure we should change naming of ACL function at this stage, but maybe I am wrong. I think natspecs are already quite explicit. Perhaps ask Clément if you have a better proposal for the name?

jatZama: IMO the most important thing would be to get correct namings in the FHE.sol library, since this is the library exposed to the end user/developer.

obatirou: The natspec is not strict about being about only for user decryption. It only says "for instance". 
I think ˋdelegateUserDecryptionToAccount` would be more explicit.
All events and functions in this file are generalist: they delegate or revoke to an account but nothing points to being only valid for user decryption.
I agree the FHE solidity library is most important but having a clear flow here should also be considered. 

I might miss something of course.

jatZama: I was waiting for a proposal, ok then i will ask @immortal-tofu if he agrees to change name to `delegateUserDecryptionToAccount` in ACL.sol.

PR #1031, opened 2025-10-08
feat(kms-connector): shared otlp context

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/475

Here's what it looks like: 
Screenshot From 2025-10-08 09-50-18

Only issue is that the parent span's duration is not shown as the sum of the 3 child spans. This is because the `handle_gateway_event` span is automatically closed when the function exit (implementation of the `tracing` crate).

I spent some time digging this, and the only way I see to resolve this would be to implement a way for `tx-sender` to warn `gw-listener` that he finished its job (grpc call for instance), and to only exit the `handle_gateway_event` when the `gw-listener` receive the ping from the `tx-sender`. It might be a bit overkill for now, I asked @leventdem about this and we decided to postpone this for now, and merge this as it is.


dartdart26 (approved):
Looks good!

PR #1033, opened 2025-10-08
chore(kms-connector): improve error handling

eudelins-zama:
Improve error handling in the connector:
- Remove use of `unwrap_or` for fhe type extraction from ct handle
- If tx get reverted, retry instead of immediately putting the tx back in the DB
- Improve default values for configs


dartdart26 (approved):
Looks good!

PR #1034, opened 2025-10-08
feat(gateway-contracts): remove useless url storing and cleanup

jatZama:
closes https://github.com/zama-ai/fhevm-internal/issues/470
refs https://github.com/zama-ai/fhevm-internal/issues/471


melanciani (approved):
thanks !

PR #1037, opened 2025-10-08
feat(kms-connector): re-estimate gas on out-of-gas failed tx and resend

jatZama:


kms-connector/crates/tx-sender/src/core/tx_sender.rs
```diff
@@ -348,7 +348,7 @@ where
         let current_gas = match call.gas {
             Some(gas) => gas,
             None => self
-                .decryption_contract
+                .kms_generation_contract
```
jatZama, line 351: I think this line change has no impact btw, since it is the same provider used on both contracts, but i put it just in case (because initial reason for this PR was fixing keygen issue, which happens via kms_generation contract on gateway).

eudelins-zama: Indeed

eudelins-zama (commented):
Tbh I don't think this PR improve the robustness of the connector's tx-sender.

There is already error handling using `get_revert_reason` above in the code, so it duplicates this.
Also we don't want to retry only on out-of-gas, but also on any kind of network error and so on.
Also, IMO, we don't want to force `call.gas` to `None` at the beginning of the fn. Because we want to increase gas at each attempts, otherwise, if a tx fail and the gas estimation is not increased, the replacement tx won't be accepted.

jatZama:
> Tbh I don't think this PR improve the robustness of the connector's tx-sender.
> 
> There is already error handling using `get_revert_reason` above in the code, so it duplicates this. Also we don't want to retry only on out-of-gas, but also on any kind of network error and so on. Also, IMO, we don't want to force `call.gas` to `None` at the beginning of the fn. Because we want to increase gas at each attempts, otherwise, if a tx fail and the gas estimation is not increased, the replacement tx won't be accepted.

You are again confusing gas and gas price i am afraid, this is really needed to re-estimate gas which was never done.

eudelins-zama (approved):

PR #1040, opened 2025-10-09
chore(gateway-contracts): cleanup MultichainACL and add default contextId

melanciani:
closes https://github.com/zama-ai/fhevm-internal/issues/471


jatZama:
There is a typo in the title of this PR fyi

gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -75,6 +75,11 @@ contract CiphertextCommits is ICiphertextCommits, UUPSUpgradeableEmptyProxy, Gat
             alreadyAddedCoprocessorTxSenders;
         /// @notice The coprocessor transaction senders involved in a consensus for a ciphertext material addition.
         mapping(bytes32 addCiphertextHash => address[] coprocessorTxSenderAddresses) coprocessorTxSenderAddresses;
+        // ----------------------------------------------------------------------------------------------
+        // Coprocessor context state variables:
+        // ----------------------------------------------------------------------------------------------
+        /// @notice The coprocessor context ID associated to the add ciphertext
+        mapping(bytes32 addCiphertextHash => uint256 contextId) addCiphertextContextId;
```
jatZama, line 82: Why is this necessary now? We want to avoid adding state variables before upgrade.

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -50,41 +50,19 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+        // Common state variables:
         // ----------------------------------------------------------------------------------------------
-        /// @notice The coprocessor transaction senders involved in a consensus for allowing a public decryption.
-        mapping(bytes32 ctHandle => address[] coprocessorTxSenderAddresses) allowPublicDecryptConsensusTxSenders;
-        // prettier-ignore
-        /// @notice The coprocessor transaction senders involved in a consensus for allowing an account.
-        mapping(bytes32 ctHandle => mapping(address accountAddress =>
-            address[] coprocessorTxSenderAddresses))
-               allowAccountConsensusTxSenders;
+        /// @notice The coprocessors that have already made an allow call.
```
jatZama, line 49: I am totally lost at this point...

gateway-contracts/contracts/CiphertextCommits.sol
melanciani: to avoid adding workaround logic  for backward compat in upcoming copro context contract

gateway-contracts/contracts/MultichainACL.sol
melanciani: to follow same pattern as in ct commits : https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/CiphertextCommits.sol#L129

it removes the need of multiple state vars, and will be useful for delegation call once we integrate it

gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -129,6 +134,12 @@ contract CiphertextCommits is ICiphertextCommits, UUPSUpgradeableEmptyProxy, Gat
         bytes32 addCiphertextHash = keccak256(abi.encode(ctHandle, keyId, ciphertextDigest, snsCiphertextDigest));
         $.addCiphertextHashCounters[addCiphertextHash]++;
 
+        // Associate the handle to coprocessor context ID 1 to anticipate their introduction in V2.
+        // Only set the context ID if it hasn't been set yet to avoid multiple identical SSTOREs.
+        if ($.addCiphertextContextId[addCiphertextHash] == 0) {
```
melanciani, line 139: btw @jatZama @isaacdecoded  this if if just here to avoid storing at each call, we only need this once, but we don't want it in the consensus to handle cases where we upgrade contracts while consensus is on-going

this is expected and we will do in the copro context PR : https://github.com/zama-ai/fhevm/pull/211/files#diff-55a63ae115a7238c4ce9ede59c15c663ceeae104003624e744471592b60548ffR151

jatZama:
Thanks for your answers, I approved.

jatZama (approved):

PR #1052, opened 2025-10-10
chore(kms-connector): improve error handling

eudelins-zama:


dartdart26 (approved):
Looks good!

kms-connector/config/tx-sender.toml
```diff
@@ -70,13 +70,13 @@ private_key = "8da4ef21b864d2cc526dbdb2a120bd2874c36c9d0a1fb7f8c63d7f7a8b41de8f"
...
+# The limit number of KMS Core responses to fetch from the database (optional, defaults to 50)
 # ENV: KMS_CONNECTOR_RESPONSES_BATCH_SIZE
-# responses_batch_size = 10
+# responses_batch_size = 50
 
-# Number of retries for transactions sent to the Gateway (optional, defaults to 3)
+# Number of retries for transactions sent to the Gateway (optional, defaults to 4)
 # ENV: KMS_CONNECTOR_TX_RETRIES
-# tx_retries = 3
+# tx_retries = 4
```
jatZama, line 79: is it normal to put  4 here but keep the default_grpc_request_retries in raw.rs set to 3?

eudelins-zama: Yes these values are independent.

For now, the kms-worker almost never need to retry one of its request to the KMS (compared to the tx-sender). So I think it is fine to leave 3 (+ its configurable via env var anyway...).

jatZama: Ah yes i only now noticed there were 2 raw.rs files, and the one of txSender was updated to 4 as expected.

PR #1076, opened 2025-10-14
ci(test-suite): add e2e test orchestrator for merge queue

tawadaa:


.github/workflows/test-suite-e2e-tests-mq.yml
```diff
@@ -0,0 +1,248 @@
...
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36  # v3.0.2
+        id: filter
+        with:
+          filters: |
+            fhevm:
+              - 'test-suite/fhevm/**'
+  fhevm-e2e-test:
+    name: test-suite-e2e-tests-mq/fhevm-e2e-test (bpr)
+    needs: check-changes
+    if: ${{ needs.check-changes.outputs.changes-fhevm == 'true' || github.event_name == 'release' || github.event_name == 'workflow_dispatch' || startsWith(github.head_ref, 'mergify/merge-queue/') }}
```
eudelins-zama, line 68: Are we sure we want to run this workflow on `github.event_name == 'release'`?

Should we keep both `test-suite-e2e-tests.yml` and `test-suite-e2e-tests-mq.yml`?

.github/workflows/test-suite-e2e-tests-mq.yml
```diff
@@ -0,0 +1,248 @@
...
+          echo "::endgroup::"
+          echo "::group::Transaction Sender Logs (filtered)"
+          ./fhevm-cli logs coprocessor-transaction-sender | grep -v "Selected 0 rows to process"
+          echo "::endgroup::"
+          echo "::group::Host Listener"
+          ./fhevm-cli logs coprocessor-host-listener
+          echo "::endgroup::"
+          echo "::group::Gateway Listener"
+          ./fhevm-cli logs coprocessor-gw-listener
+          echo "::endgroup::"
```
eudelins-zama, line 242: Would be great to store all logs as discussed during our team events.

But could be done in another PR later of course

.github/workflows/coprocessor-gw-listener-docker-build.yml
```diff
@@ -64,3 +76,27 @@ jobs:
...
+    runs-on: ubuntu-latest
+    outputs:
+      image-built: ${{ steps.check-build.outputs.image-built }}
+      image-tag: ${{ steps.check-build.outputs.image-tag }}
+    steps:
+      - name: Check if image was built
+        id: check-build
+        run: |
+          # Check if docker build job ran and succeeded
+          if [[ "${{ needs.build.result }}" == "success" ]]; then
```
eudelins-zama, line 94: Are we sure all these `output-build-status` are userful?

AFAIK, the `test-suite-orchestrate` have access to the `${{ github.sha }}` and thus the image tag already. Plus, I think it can check by itself whether the `build.result` step was equal to `success` or not right?

.github/workflows/test-suite-e2e-tests-mq.yml
eudelins-zama: Also, I think we should delete `test-suite-e2e-tests-with-build.yml` right?

eudelins-zama: If we keep both `test-suite-e2e-tests.yml` and `test-suite-e2e-tests-mq.yml`, what is the purpose of `test-suite-orchestrate-e2e-tests.yml`?
Couldn't we just use the `trigger-builds` job directly as the first job of the `test-suite-e2e-tests-mq.yml`?

tawadaa: we don't need `github.event_name == 'release'`, it's an old test, let me clean it
Yes, for now, we keep both of them, in order not to disturb the current e2e tests pipeline, until validating the merge queue pipeline

tawadaa: > If we keep both test-suite-e2e-tests.yml and test-suite-e2e-tests-mq.yml, what is the purpose of test-suite-orchestrate-e2e-tests.yml? Couldn't we just use the trigger-builds job directly as the first job of the test-suite-e2e-tests-mq.yml?

There are mainly 2 approches, either separate simple workflows or 1 complex workflow, I prefer the first approcah as it's easier to maintain.
In our case, it might appear that test-suite-e2e-tests-mq.yml and test-suite-orchestrate-e2e-tests.yml can be easily merged, but I prefer to keep them separate for few reasons:

- They have different roles, one is for orchestration and the other is for execution
- We might need to launch manually test-suite-e2e-tests-mq.yml for some reason e.g hotfix

eudelins-zama: Fine for me, thx for the details :slightly_smiling_face:

eudelins-zama:
Closing as this one was merged: https://github.com/zama-ai/fhevm/pull/1435

PR #1079, opened 2025-10-15
refactor(host-contracts): rename ACL delegation methods and events

isaacdecoded:
Refs https://github.com/zama-ai/fhevm-internal/issues/474


host-contracts/contracts/ACL.sol
```diff
@@ -272,22 +272,22 @@ contract ACL is
         // Set the delegation expiry date.
         delegation.expiryDate = newExpiryDate;
 
-        emit DelegatedAccount(
+        emit DelegatedForUserDecryption(
             msg.sender,
             delegate,
             contractAddress,
-            delegation.delegationCounter++,
+            ++delegation.delegationCounter,
```
isaacdecoded: I think it's better to start this delegation counter from value `1` instead of `0`, similar to how we do in [Gateway for zkProof](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/InputVerification.sol#L184-L185).

wdyt? @jatZama

jatZama: Ok for me

host-contracts/test/acl/acl.t.sol
```diff
@@ -191,26 +191,25 @@ contract ACLTest is Test {
         uint64 expiryDate = uint64(block.timestamp) + 7 hours;
         uint64 oldExpiryDate = delegation.expiryDate;
         uint64 newExpiryDate = expiryDate;
-        uint64 delegationCounter = delegation.delegationCounter++;
 
         vm.prank(sender);
         vm.expectEmit(address(acl));
-        emit ACLEvents.DelegatedAccount(
+        emit ACLEvents.DelegatedForUserDecryption(
```
jatZama, line 208: is it possible to change the test name of the test function?

host-contracts/test/acl/acl.t.sol
```diff
@@ -191,26 +191,25 @@ contract ACLTest is Test {
...
             sender,
             delegate,
             contractAddress,
-            delegationCounter,
+            ++delegation.delegationCounter,
```
jatZama: is it possible to double check that the counter is indeed incremented inside storage, using vm.load cheatcode. Same thing in revoke test.
https://getfoundry.sh/reference/cheatcodes/load/

jatZama (changes requested):
2 small requests

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -779,10 +781,10 @@ pub fn acl_result_handles(event: &Log<AclContractEvents>) -> Vec<Handle> {
             allowed_for_decryption.handlesList.clone()
         }
         AclContractEvents::Initialized(_)
-        | AclContractEvents::DelegatedAccount(_)
+        | AclContractEvents::DelegatedForUserDecryption(_)
```
rudy-6-4, line 784: Can't we keep the same name as function ?
delegateForUserDecryption

rudy-6-4 (approved):

host-contracts/test/acl/acl.t.sol
jatZama: Forget about using vm.load cheatcode, since this is inside nested mapping, too much headache.
Instead, just extend the ACL contract as a MockACL one which will expose easily a getter view function giving access at least to the struct counter, or even better the whole struct itself to check all its fields are correctly being updated as expected. For this I advise you to reuse same pattern as the foundry test for HCULimit, using an HCULimitMock contract extending original one here https://github.com/zama-ai/fhevm/blob/main/host-contracts/test/hcuLimit/HCULimit.t.sol#L17

So in summary we need 2 things: 
- we need previous test with the `MockACL` checking that the `Delegation` struct in storage is being updated as expected after each delegate/revoke tx (within a chain of delegate/revoke).
- as discussed before lunch, also test that within a similar (or same) chain of delegate/revoke calls in some custom order, each event emits an increasing counter incrementing by `1` each time.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
jatZama: I think it is preferable to keep same convention as already existing events, such as `Initialized`, `Allowed` etc.

host-contracts/contracts/ACL.sol
```diff
@@ -224,34 +227,36 @@ contract ACL is
...
+        uint64 expirationDate
     ) public virtual whenNotPaused {
-        if (expiryDate < block.timestamp + 1 hours) {
-            revert ExpiryDateBeforeOneHour();
+        if (expirationDate < block.timestamp + 1 hours) {
+            revert ExpirationDateBeforeOneHour();
         }
-        if (expiryDate > block.timestamp + 365 days) {
-            revert ExpiryDateAfterOneYear();
+        if (expirationDate > block.timestamp + 365 days) {
```
jatZama: I thought this check needed to be deleted, after the OZ requirement...
Correct me if i'm wrong cc: @immortal-tofu @0xalexbel

jatZama:
Everything looks fine, and new foundry tests are great.
Only remaining request would be to remove the upper bound check on the delegation, since OpenZeppelin requested the ability to get unlimited delegation. In practise this would be done by passing a huge number such as type(uint64).max, but for this to be possible you need to remove the check on ` if (expirationDate > block.timestamp + 365 days)` at line 244 of ACL.sol.

host-contracts/contracts/ACL.sol
isaacdecoded: Yes, good catch. I forgot to remove the upper bound check. Done [here](https://github.com/zama-ai/fhevm/pull/1079/commits/c411f445263b96b00930816d9bff5b1073a2d9b7).

host-contracts/contracts/ACL.sol
```diff
@@ -424,6 +410,30 @@ contract ACL is
         return $.allowedForDecryption[handle];
     }
 
+    /**
+     * @notice Returns whether an account is delegated to access the handle for user decryption.
+     * @param delegate The address of the account that receives the delegation.
```
rudy-6-4: reversed order ? delegate delegator

host-contracts/contracts/ACL.sol
```diff
@@ -424,6 +410,30 @@ contract ACL is
...
+     * @param delegator The address of the account that delegates access to its handles.
+     * @param handle The handle to check for delegated user decryption.
```
rudy-6-4: reverse handle / contractAddress ?

host-contracts/contracts/ACL.sol
```diff
@@ -332,43 +339,22 @@ contract ACL is
...
-
-    /**
-     * @notice Get the delegation expiry date.
-     * @param delegate The address of the account that receives the delegation.
-     * @param delegator The address of the account that delegates access to its handles.
-     * @param contractAddress The contract address to delegate access to.
-     * @return expiryDate the expiryDate (0 means delegation is inactive).
-     */
-    function getDelegationExpiryDate(
+    function getUserDecryptionDelegationExpirationDate(
```
rudy-6-4, line 357: can we have a consistent order, e.g. delegator then delegate ?

jatZama (approved):

host-contracts/contracts/ACL.sol
```diff
@@ -224,34 +224,42 @@ contract ACL is
...
 
         ACLStorage storage $ = _getACLStorage();
-        Delegation storage delegation = $.delegations[msg.sender][delegate][contractAddress];
+        UserDecryptionDelegation storage userDecryptionDelegation = $.userDecryptionDelegations[msg.sender][delegate][
+            contractAddress
+        ];
         uint256 blockNumber = block.number;
 
-        if (delegation.lastBlockDelegateOrRevoke == blockNumber) {
+        if (userDecryptionDelegation.lastBlockDelegateOrRevoke == blockNumber) {
```
rudy-6-4, line 257: Should it be said in doc. At least for revocation ? (you cannot optimistically delegate and retract in the same block)

rudy-6-4 (approved):

host-contracts/contracts/ACL.sol
jatZama: Yes this should be explained by @0xalexbel in the FHE.sol library and docs.

isaacdecoded: @0xalexbel just tagging you here for tracking it, and to be sure we update the docs as needed, as Rudy is pointing out.

PR #1089, opened 2025-10-17
feat(gateway-contracts): implement user decryption delegation logic

isaacdecoded:
Closes https://github.com/zama-ai/fhevm-internal/issues/485


gateway-contracts/contracts/interfaces/IMultichainACL.sol
```diff
@@ -63,6 +107,42 @@ interface IMultichainACL {
...
+     * @param expiryDate The expiration date for the intended delegation.
+     * @param delegationCounter A counter specific to the (delegator, delegate, contract) triple tied to the delegation.
+     */
+    function delegateUserDecryption(
+        uint256 chainId,
+        address delegator,
+        address delegate,
+        address contractAddress,
+        uint64 expiryDate,
+        uint64 delegationCounter
```
rudy-6-4: the counter position is not consistent with ACL.sol event.

gateway-contracts/contracts/interfaces/IMultichainACL.sol
```diff
@@ -21,6 +21,24 @@ interface IMultichainACL {
...
     event AllowPublicDecrypt(bytes32 indexed ctHandle);
 
+    /**
+     * @notice Emitted when a user decryption is delegated to a delegate and contract addresses.
+     * @param chainId The chain ID of the registered host chain where the contract is deployed.
+     * @param delegator The address of the account that delegates access to its handles.
+     * @param delegate The address of the account that receives the delegation.
+     * @param contractAddress The address of the contract that is part of the user decryption context.
+     */
+    event DelegateUserDecryption(uint256 indexed chainId, address delegator, address delegate, address contractAddress);
```
jatZama: I think it would be better to add counter, oldExpiryDate and newExpiryDate like in ACL.sol https://github.com/zama-ai/fhevm/blob/main/host-contracts/contracts/ACL.sol#L275 and almost same for revoke event. Wdyt?

isaacdecoded: Yes, I agree. Added in [this commit](https://github.com/zama-ai/fhevm/pull/1089/commits/5526f4f554869c0f4b8bcc98dbd81a9943f9841f#diff-54b96465377fce818073cb022331ab574e54ef3056a8c1b125682a6ab43e99f2R34-R60).

gateway-contracts/contracts/interfaces/IMultichainACL.sol
isaacdecoded: Ok, yes, it's now aligned in [this commit](https://github.com/zama-ai/fhevm/pull/1089/commits/5526f4f554869c0f4b8bcc98dbd81a9943f9841f#diff-54b96465377fce818073cb022331ab574e54ef3056a8c1b125682a6ab43e99f2R144-R145).

gateway-contracts/contracts/interfaces/IMultichainACL.sol
```diff
@@ -76,6 +176,20 @@ interface IMultichainACL {
...
     function isAccountAllowed(bytes32 ctHandle, address accountAddress) external view returns (bool);
 
+    /**
+     * @notice Indicates if the delegator has delegated access to the delegate and contract address.
+     * @param chainId The chain ID of the registered host chain where the contract is deployed.
+     * @param delegator The address of the account that delegates access to its handles.
+     * @param delegate The address of the account that has the delegation access.
+     * @param contractAddress The address of the contract that is part of the user decryption context.
+     */
+    function isUserDecryptionDelegated(
```
melanciani, line 238: haven't followed all discussions, but I'm surprised to see the notion of "user decryption" instead of "account" here. I thought we wanted to have a more general notion of ACLs as we did with `AllowAccount` to highlight that an account gets access to a handle for not only user decryption (only op currently available), but also future things like bridging. And maybe this is not relevant anymore ?

gateway-contracts/contracts/interfaces/IMultichainACL.sol
```diff
@@ -63,6 +127,42 @@ interface IMultichainACL {
...
+    ) external;
+
+    /**
+     * @notice Revokes the access for user decryption previously delegated to the delegate and contract addresses.
+     * @param chainId The chain ID of the registered host chain where the contract is deployed.
+     * @param delegator The address of the account that revokes access to its handles.
+     * @param delegate The address of the account that stops receiving the delegation.
+     * @param contractAddress The address of the contract that was part of the user decryption context.
+     * @param delegationCounter A counter specific to the (delegator, delegate, contract) triple tied to the delegation.
+     * @param expirationDate The expiration date for the intended delegation.
```
melanciani: `expirationDate` is a uint64, but I see no comments on what it should be : in days ? in seconds ? imo worth mentioning this at least here

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -2,19 +2,27 @@
...
  * @title MultichainACL smart contract
  * @notice See {IMultichainACL}
  */
-contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwnable, GatewayConfigChecks {
+contract MultichainACL is
+    IMultichainACL,
+    UUPSUpgradeableEmptyProxy,
+    GatewayOwnable,
+    GatewayConfigChecks,
+    MulticallUpgradeable
```
melanciani, line 26: worth a comment somewhere saying why we need `MulticallUpgradeable` in `MultichainACL` specifically

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -54,13 +62,28 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
         mapping(bytes32 allowHash => address[] coprocessorTxSenderAddresses) allowConsensusTxSenders;
         /// @notice Allowed public decryptions.
         mapping(bytes32 allowHash => bool isAllowed) isAllowed;
-        /// @notice The coprocessor transaction senders involved in a consensus for delegating an account.
-        mapping(bytes32 delegateAccountHash => address[] coprocessorTxSenderAddresses) delegateAccountConsensusTxSenders;
```
melanciani, line 58: this breaks storage layout for upgrades. , you need a new variable for `delegateOrRevokeUserDecryptionTxSenders` or you keep the initial name, no ?

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+            delegate,
+            contractAddress,
+            delegationCounter,
+            expirationDate
+        );
+
+        mapping(address => bool) storage alreadyDelegatedUserDecryptionCoprocessors = $
+            .alreadyDelegatedOrRevokedUserDecryptionCoprocessors[delegateUserDecryptionHash];
+
+        /// @dev Check if the coprocessor has already delegated the user decryption.
```
melanciani: you can remove the `/ @dev` !

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+            chainId,
+            delegator,
+            delegate,
+            contractAddress,
+            delegationCounter,
+            expirationDate
+        );
+
+        mapping(address => bool) storage alreadyDelegatedUserDecryptionCoprocessors = $
+            .alreadyDelegatedOrRevokedUserDecryptionCoprocessors[delegateUserDecryptionHash];
```
melanciani: I'm not sure to get why a coprocessor cannot delegate AND revoke ? if this is intended, please explain in the comments why we need this

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+            // Mark the delegate user decryption hash as having reached consensus for delegation.
+            $.delegatedOrRevokedUserDecryptionHashes[delegateUserDecryptionHash] = true;
+
+            emit DelegateUserDecryption(
+                chainId,
+                delegator,
+                delegate,
+                contractAddress,
+                delegationCounter,
+                oldExpirationDate,
```
melanciani, line 280: why do we consider the `oldExpirationDate` in the event ? shouldn't it just be the one asked for in the function ?

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+        ) {
+            UserDecryptionDelegation storage userDecryptionDelegation = $.userDecryptionDelegations[chainId][delegator][
+                delegate
+            ][contractAddress];
+
+            // Check that the delegation counter is greater than a previous one.
+            if (delegationCounter <= userDecryptionDelegation.delegationCounter) {
+                revert UserDecryptionDelegationCounterTooLow(delegationCounter);
+            }
+
```
melanciani, line 265: we don't need to check anything on the expiration date ? the counter if enough ?

gateway-contracts/contracts/interfaces/IMultichainACL.sol
isaacdecoded: Yes, based on [this conversation](https://github.com/zama-ai/fhevm/pull/1026#discussion_r2429677634) in the ACL implementation, we discussed renaming it in a more "intention-revealing" way. Then, we'd like to also use similar name in the FHE library and MultichainACL.

gateway-contracts/contracts/MultichainACL.sol
isaacdecoded: Good catch ✊ I forgot that the storage layout turns incompatible not only from removing/altering the definitions but also from just renaming the variables. Thanks.

gateway-contracts/contracts/MultichainACL.sol
isaacdecoded: The reason is that the `delegationCounter` is included in the preimage of the `delegateUserDecryptionHash`, meaning that a coprocessor shouldn't delegate AND revoke a user decryption delegation for the same counter. I'll add a comment in the code, yes.

gateway-contracts/contracts/MultichainACL.sol
isaacdecoded: In case of processing two user decryption delegations sequentially, maybe it could make sense to show the previous expiration date replaced by the last processed delegation in the emitted event.

gateway-contracts/contracts/MultichainACL.sol
isaacdecoded: The check of the expiration date, which is just validating it's [one hour in the future](https://github.com/zama-ai/fhevm/pull/1079/files#diff-342c378b72fda8ec0e0e240388d7031d0e2753ebae955933659cf704af896818R238-R240), is already done at the ACL host contract level. So, for the MultichainACL contract it's enough to check that the user decryption delegations are processed sequentially by checking the counters.

gateway-contracts/contracts/interfaces/IMultichainACL.sol
melanciani: I see, still a bit surprised, I don't think the FHE library has anything to do with the gateway, but you all agreed than fine I guess

gateway-contracts/contracts/MultichainACL.sol
melanciani: you should enable back upgrade tests for this contract in the CI btw, by removing `if: false` from the `MultichainACL` upgrade step

gateway-contracts/contracts/MultichainACL.sol
melanciani: ah, interesting point ! I see, it makes sense indeed

gateway-contracts/contracts/MultichainACL.sol
melanciani: I'm not sure to get why you would need both expiration dates honestly, but if the guild agrees on this than fine I guess

gateway-contracts/contracts/MultichainACL.sol
melanciani: hmm well you could still put a right counter but lie with the expiration date no ? I guess it's "fine" since we ask for consensus anyway, but in that case imo it's either :
- check both the counter and expiration date
- or check nothing else (other than consensus)

unless I'm missing something ? maybe worth syncing with the guild here

in any case, I guess that's worth a comment in the code !

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+        // A "late" valid coprocessor transaction sender address will still be added in the list.
+        $.delegateOrRevokeUserDecryptionTxSenders[delegateUserDecryptionHash].push(msg.sender);
+
+        // Send the event only if the consensus is reached in the current call.
+        // This means a "late" response will not be reverted, just ignored and no event will be emitted.
+        if (
+            !$.delegatedOrRevokedUserDecryptionHashes[delegateUserDecryptionHash] &&
+            _isConsensusReached($.delegateOrRevokeUserDecryptionTxSenders[delegateUserDecryptionHash].length)
+        ) {
+            UserDecryptionDelegation storage userDecryptionDelegation = $.userDecryptionDelegations[chainId][delegator][
```
melanciani, line 253: btw shouldn't these load + following check(s) be done outside of consensus ? Imo we should try to make gas usage more "amortized" between logic with and without consensus

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+            delegate,
+            contractAddress,
+            delegationCounter,
+            expirationDate
+        );
+
+        mapping(address => bool) storage alreadyRevokedUserDecryptionCoprocessors = $
+            .alreadyDelegatedOrRevokedUserDecryptionCoprocessors[delegateUserDecryptionHash];
+
+        /// @dev Check if the coprocessor has already revoked the user decryption delegation.
```
melanciani: you can remove the `/ @dev` !

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+            chainId,
+            delegator,
+            delegate,
+            contractAddress,
+            delegationCounter,
+            expirationDate
+        );
+
+        mapping(address => bool) storage alreadyRevokedUserDecryptionCoprocessors = $
+            .alreadyDelegatedOrRevokedUserDecryptionCoprocessors[delegateUserDecryptionHash];
```
melanciani: maybe you want to simplify namings and have 2 different storage variables in order to have 2 different errors depending on if a copro already delegated or revoked , pretty much like we've done in `_checkCoprocessorAlreadyResponded` in the InputVerification ?

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+        // A "late" valid coprocessor transaction sender address will still be added in the list.
+        $.delegateOrRevokeUserDecryptionTxSenders[delegateUserDecryptionHash].push(msg.sender);
+
+        // Send the event if and only if the consensus is reached in the current call.
+        // This means a "late" response will not be reverted, just ignored and no event will be emitted.
+        if (
+            !$.delegatedOrRevokedUserDecryptionHashes[delegateUserDecryptionHash] &&
+            _isConsensusReached($.delegateOrRevokeUserDecryptionTxSenders[delegateUserDecryptionHash].length)
+        ) {
+            UserDecryptionDelegation storage userDecryptionDelegation = $.userDecryptionDelegations[chainId][delegator][
```
melanciani, line 333: same as above for the following: 
- imo it can be moved out of consensus section
- probably worth checking the expiration date
- not sure why old expiration date is included in `RevokeUserDecryption`

gateway-contracts/contracts/MultichainACL.sol
jatZama: I would say let's keep it, it makes things clear that we are overriding old expiration date, and add very little gas (not an indexed topic). This is easier for debugging , if old expiration was null we know it was previously revoked or never delegated, but if it was not null then this is an override.

gateway-contracts/contracts/MultichainACL.sol
jatZama: only the SLOAD could be put outside eventually, but not sure we want to do this, since it would increase gas cost for nothing for other coprocessors. long term the design should be reconsidered using same pattern as for the new user decryption flow. Btw this won't help too much on gas amortization, since an SLOAD is still very cheap compared to the following SSTOREs which are the real origin of the bump in gas at consensus.
Very important: for security you should absolutely not put the check on the counter outside consensus, this could not work since we only assume correct order at consensus (not possible before, or there is high chance it would revert) for honest coprocessors.
For the moment, what Isaac did seems to be fine for me here also.

gateway-contracts/contracts/MultichainACL.sol
jatZama: no the attack you are describing is not possible, since both counter and expiration date are part of the preimage of the DelegateUserDecryptionHash. Counter check is mandatory for security, since we assume honest node are always sending there counter in correct order. Otherwise there is a security attack: a malicious node could by himself frontrun a "future counter" out of order at the moment of reaching consensus and create a gap.
What Isaac did seems to be fine for me.

gateway-contracts/contracts/MultichainACL.sol
jatZama: I think it is fine to use a single storage variable like Isaac did, to save gas and to make implementation more straightforward.

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -188,6 +364,24 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
         return $.isAllowed[allowHash];
     }
 
+    /**
+     * @notice See {IMultichainACL-isUserDecryptionDelegated}.
+     */
+    function isUserDecryptionDelegated(
```
rudy-6-4, line 387: who is going to call that ? app contract will rely on host ACL.sol.

gateway-contracts/contracts/MultichainACL.sol
jatZama: - as explained above, for me it is ok to put it in consensus while waiting eventually to use more efficient design, anyways this is not creating the huge bump in gas at consensus, because an SLOAD is quite cheap compared to the following SSTOREs.
- no need to check expirationDate, as explained by Isaac previously, clocks are not necessarily in sync between chains, and we rely just on counter for security.
- for better traceability and because this adds almost no gas.

jatZama (approved):
Thanks @isaacdecoded for your good work. To me it looks ready to be merged.

gateway-contracts/contracts/MultichainACL.sol
jatZama: It will be called by the Decryption contract on gateway at least this is necessary there.

gateway-contracts/contracts/MultichainACL.sol
melanciani: the attack I explained is not precise enough indeed: I was talking about if a majority of copro becomes malicious, which I know is against the initial assumption, but I mentioned this because else I didn't get the point of checking the counter if it wasn't for this. EDIT: but I now get that the reason of the counter check is otherwise, so I guess it's ok yes

> Otherwise there is a security attack: a malicious node could by himself frontrun a "future counter" out of order at the moment of reaching consensus and create a gap.
What Isaac did seems to be fine for me.

~~since the counter if part of the preimage hash, how could the malicious node frontrun a "future counter" at consensus ? 🤔  if he gives a different counter, the hash will be different and the consensus won't be reached , so I'm not sure I get the issue here~~

EDIT: ok I think I get it: this is in case of multiple delegation calls in parallel, and we want to make sure to "approve" (ie trigger consensus) in the right order. This seems a very important concern and it should be explained in the comments

gateway-contracts/contracts/MultichainACL.sol
melanciani: > only the SLOAD could be put outside eventually, but not sure we want to do this, since it would increase gas cost for nothing for other coprocessors. long term the design should be reconsidered using same pattern as for the new user decryption flow. Btw this won't help too much on gas amortization, since an SLOAD is still very cheap compared to the following SSTOREs which are the real origin of the bump in gas at consensus.

ok sure, I agree it's not of a big deal since gas different is not huge

> Very important: for security you should absolutely not put the check on the counter outside consensus, this could not work since we only assume correct order at consensus (not possible before, or there is high chance it would revert) for honest coprocessors.

ah ok I think I get it, this is to handle in case of multiple delegation calls in parallel. This seems a very important concern and it should be explained in the comments

gateway-contracts/contracts/MultichainACL.sol
melanciani: I agree on the gas saving part, less about the straightforward part, what I suggest helps debugging since it allows considering more precise revert errors. Imo we should align the patterns by either use this one and accept this issue, or do like in Inputverification and accept a small gas increase, but not keep both in the end

melanciani (changes requested):
overall logic seems ok, but I would prefer to add explanatory comments in key parts !

gateway-contracts/contracts/MultichainACL.sol
jatZama: if you want to use two mappings, then you should do two checks instead of one, so implementation is a bit more complex, if you think it is ok despite gas increase then ok.

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+                chainId,
+                delegator,
+                delegate,
+                contractAddress,
+                delegationCounter,
+                expirationDate,
+                msg.sender
+            );
+        }
+
```
melanciani: also let's anticipate https://github.com/zama-ai/fhevm-internal/issues/446 and add an event for each calls (not only for the consensus) !

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +192,159 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+        // Mark the coprocessor has already revoked for the delegate user decryption hash.
+        alreadyRevokedUserDecryptionCoprocessors[msg.sender] = true;
+
+        // Store the coprocessor transaction sender address for the delegate user decryption hash.
+        // A "late" valid coprocessor transaction sender address will still be added in the list.
+        $.delegateOrRevokeUserDecryptionTxSenders[delegateUserDecryptionHash].push(msg.sender);
+
```
melanciani, line 326: same, let's anticipate https://github.com/zama-ai/fhevm-internal/issues/446 and add an event for each calls (not only for the consensus) !

rudy-6-4 (approved):

jatZama (approved):

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -54,13 +64,37 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
         mapping(bytes32 allowHash => address[] coprocessorTxSenderAddresses) allowConsensusTxSenders;
         /// @notice Allowed public decryptions.
         mapping(bytes32 allowHash => bool isAllowed) isAllowed;
-        /// @notice The coprocessor transaction senders involved in a consensus for delegating an account.
+        /// @notice DEPRECATED: to remove in a state reset.
```
melanciani, line 67: could you add a comment in https://github.com/zama-ai/fhevm-internal/issues/484 so that we remember about this ? thanks !

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -188,6 +419,24 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+    ) external view returns (bool) {
+        MultichainACLStorage storage $ = _getMultichainACLStorage();
+
+        UserDecryptionDelegation storage userDecryptionDelegation = $.userDecryptionDelegations[chainId][delegator][
+            delegate
+        ][contractAddress];
+
+        return userDecryptionDelegation.expirationDate >= block.timestamp;
+    }
+
```
melanciani, line 401: just realized : you should add 2 getters for fetching the tx senders (one for delegate, one for revoke) !, like we already do for other allows !

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +203,203 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+            contractAddress,
+            delegationCounter,
+            expirationDate
+        );
+
+        mapping(address => bool) storage alreadyDelegatedUserDecryptionCoprocessors = $
+            .alreadyDelegatedUserDecryptionCoprocessors[delegateUserDecryptionHash];
+
+        // Check if the coprocessor has already delegated the user decryption.
+        if (alreadyDelegatedUserDecryptionCoprocessors[msg.sender]) {
```
melanciani: I think it's more future proof if we use a common internal function here for these checks, like we do in input verification contract

if you want to avoid the additional SLOAD, you could make this internal function return the loaded values and use them in the rest of the code !

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +203,203 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+        if (
+            !$.delegatedOrRevokedUserDecryptionHashes[delegateUserDecryptionHash] &&
+            _isConsensusReached($.delegateUserDecryptionTxSenders[delegateUserDecryptionHash].length)
+        ) {
+            UserDecryptionDelegation storage userDecryptionDelegation = $.userDecryptionDelegations[chainId][delegator][
+                delegate
+            ][contractAddress];
+
+            /**
+             * @dev Check that the delegation counter is strictly greater than the previous one.
```
melanciani: bad comment format: better to use `//` to align with the rest of the comments ! and no need for `@dev` it's not natspec comment !

gateway-contracts/contracts/MultichainACL.sol
```diff
@@ -169,6 +203,203 @@ contract MultichainACL is IMultichainACL, UUPSUpgradeableEmptyProxy, GatewayOwna
...
+        if (
+            !$.delegatedOrRevokedUserDecryptionHashes[delegateUserDecryptionHash] &&
+            _isConsensusReached($.revokeUserDecryptionTxSenders[delegateUserDecryptionHash].length)
+        ) {
+            UserDecryptionDelegation storage userDecryptionDelegation = $.userDecryptionDelegations[chainId][delegator][
+                delegate
+            ][contractAddress];
+
+            /**
+             * @dev Check that the delegation counter is strictly greater than the previous one.
```
melanciani: same comment for comment format

gateway-contracts/contracts/MultichainACL.sol
isaacdecoded: Yes, it's true. I'm wondering what components will use these getters, maybe the relayer?

gateway-contracts/contracts/MultichainACL.sol
isaacdecoded: Yes, I thought about it, but in the end I didn't implement that way because of the extra SLOAD and returning two `mapping(address => bool) storage` seems a little odd. But I'll try again and let's see 👍

melanciani: I think in the gateway, gas usage is less of an issue so it's better if we make the code more future proof + follow software dev best practices even if it asks a few additional sload

melanciani (approved):
looks good thanks !!

gateway-contracts/contracts/MultichainACL.sol
isaacdecoded: Done! 👍

PR #1092, opened 2025-10-17
feat(coprocessor): host-listener + transaction-sender, delegations

rudy-6-4:
Closes https://github.com/zama-ai/fhevm-internal/issues/10


.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -182,6 +182,13 @@ jobs:
         run: |
           ./fhevm-cli test public-decrypt-http-mixed
 
+      - name: Delegate User Decryption (partial test)
+        working-directory: test-suite/fhevm
+        run: |
+          ./fhevm-cli test delegate-user-decryption && echo "Should fail" && exit 1 || echo "Failed as expected"
+          ./fhevm-cli logs coprocessor-transaction-sender | grep DelegationRow
+          ./fhevm-cli logs coprocessor-transaction-sender | grep "DelegationForUserDecryption txn succeeded"
```
isaacdecoded: Is this possible to also check if the transaction that propagates the delegation to MultichainACL (Gateway) also succeeds?

rudy-6-4: It's the one that is checked here

coprocessor/fhevm-engine/transaction-sender/Cargo.toml
```diff
@@ -33,6 +33,7 @@ humantime = { workspace = true }
 
 # local dependencies
 fhevm-engine-common = { path = "../fhevm-engine-common" }
+fhevm_gateway_bindings = { path = "../../../gateway-contracts/rust_bindings" }
```
isaacdecoded, line 37: Just for the record, remember to replace this with git deps here and at the gw-listener component. Also, to rollback changes in the make command at the Gateway for generating the bindings. cc @eudelins-zama

coprocessor/fhevm-engine/host-listener/build.rs
```diff
@@ -7,7 +7,13 @@ use semver::Version;
...
 
 fn build_contracts() {
-    println!("cargo:rerun-if-changed=../../../host-contracts/contracts");
+    println!(
+        "cargo:rerun-if-changed=../../../host-contracts/contracts/ACL.sol"
+    );
+    println!(
+        "cargo:rerun-if-changed=../../../host-contracts/contracts/ACLEvents.sol"
+    );
+    println!("cargo:rerun-if-changed=../../../host-contracts/contracts/FHEVMExecutor.sol");
```
isaacdecoded, line 16: Should we consider implementing the rust bindings for the host-contracts as well? In such case, do you know if there's an issue ticket already?

coprocessor/fhevm-engine/transaction-sender/src/ops/add_ciphertext.rs
```diff
@@ -182,8 +177,11 @@ impl<P: Provider<Ethereum> + Clone + 'static> AddCiphertextOperation<P> {
     fn already_added_error(&self, err: &RpcError<TransportErrorKind>) -> Option<Address> {
         err.as_error_resp()
             .and_then(|payload| payload.as_decoded_interface_error::<CiphertextCommitsErrors>())
-            .map(|error| match error {
-                CiphertextCommitsErrors::CoprocessorAlreadyAdded(c) => c.coprocessorTxSenderAddress,
+            .and_then(|error| match error {
+                CiphertextCommitsErrors::CoprocessorAlreadyAdded(c) => {
+                    Some(Address::from_slice(c.ctHandle.as_ref()))
```
rudy-6-4: should be tx sender

coprocessor/fhevm-engine/host-listener/build.rs
eudelins-zama: See https://github.com/zama-ai/fhevm-internal/issues/371 @isaacdecoded

coprocessor/fhevm-engine/transaction-sender/Cargo.toml
eudelins-zama: Indeed.

I suggest to do the `git` dep import in the `Cargo.toml` of the workspace instead of in each sub-crate as well (in sub-crate then only import via `fhevm_gateway_bindings.workspace = true`.

rudy-6-4: Yes that will be done after this one is merged.

coprocessor/fhevm-engine/host-listener/build.rs
rudy-6-4: will be happy to use them when available

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -103,5 +103,5 @@ jobs:
 
     - name: Run tests
       run: |
-        DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor TEST_GLOBAL_LOCALSTACK=1 cargo test --release -- --test-threads=1
+        DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor TEST_GLOBAL_LOCALSTACK=1 cargo test -p --transaction-sender --release -- --test-threads=1
```
eudelins-zama: Why this `-p --transaction-sender`?

rudy-6-4: mmm typo, I want to only test transaction sender temporarly because it's broken on CI but not locally

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -182,6 +182,14 @@ jobs:
...
           ./fhevm-cli test public-decrypt-http-mixed
 
+      - name: Delegate User Decryption (partial test)
+        working-directory: test-suite/fhevm
+        run: |
+          ./fhevm-cli test delegate-user-decryption
+          echo Checking transaction-sender logs for DelegateUserDecryption
+          ./fhevm-cli logs coprocessor-transaction-sender | grep "DelegateUserDecryption txn succeeded"
+          echo Checking transaction-sender logs for RevokeUserDecryptionDelegation
+          ./fhevm-cli logs coprocessor-transaction-sender | grep "RevokeUserDecryptionDelegation txn succeeded"
```
eudelins-zama: Shouldn't this be directly included in the `./fhevm-cli test delegate-user-decryption` command?

I mean for users of local e2e tests, it would be great for them to just have to run `./fhevm-cli test delegate-user-decryption` to be sure the test worked, without having to dig into the logs

charts/coprocessor/values.yaml
```diff
@@ -755,8 +757,15 @@ txSender:
     - --gas-limit-overprovision-percent=120
     - --graceful-shutdown-timeout=8s
     - --service-name=txn-sender
+    # New in 0.10
     - --metric-host-txn-latency=0.1:60.0:0.1
     - --metric-zkproof-txn-latency=0.1:60.0:0.1
+    - --host-chain-url=$(ETHEREUM_RPC_URL)
+    - --delegation-block-delay=10  # not applied immediately to avoid reorg
+    - --delegation-clear-after-n_blocks=648000  # ~3 months at 12s block time, only on succesful delgations
```
eudelins-zama: `succesful delgations` -> `successful delegations`

.github/workflows/test-suite-e2e-tests.yml
rudy-6-4: Yes maybe. It's a temporary test for the CI waiting for the delegation to be finished. I'll see if it's easy to move.

coprocessor/fhevm-engine/db-migration/migrations/20251013083601_delegations.sql
```diff
@@ -0,0 +1,28 @@
...
+    delegator BYTEA NOT NULL,
+    delegate BYTEA NOT NULL,
+    contract_address BYTEA NOT NULL,
+    delegation_counter BIGINT NOT NULL,
+    old_expiration_date BIGINT NOT NULL, -- 0 = first time delegation
+    new_expiration_date BIGINT NOT NULL, -- 0 = revoke
+    host_chain_id BIGINT NOT NULL,
+    block_number BIGINT NOT NULL,
+    block_hash BYTEA NOT NULL, -- to check finality
+    transaction_id BYTEA,
```
obatirou, line 12: Why transaction id can be null?

obatirou: I suppose due to delegate_user_decrypt_idempotent_errorcall 
Need to read the entire PR to understand

rudy-6-4: It's because the transaction_id of the log received from the host chain could be None.

obatirou: If this happens this means the transaction is not yet included in a bloc.
The events are processed optimistically then?
This null value will only be updated when there is a catch up launched right?

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
```diff
@@ -180,6 +221,62 @@ fn parse_args() -> Conf {
...
+                return Ok(provider);
+            }
+            Err(e) => {
+                error!(
+                    name,
+                    error = %e,
+                    retry_interval = ?conf.provider_retry_interval,
+                    "Failed to connect to chain on startup, retrying"
+                );
+                tokio::time::sleep(conf.provider_retry_interval).await;
```
obatirou, line 274: in case of cancel, this will need to wait until the sleep is finished here?

coprocessor/fhevm-engine/db-migration/migrations/20251013083601_delegations.sql
rudy-6-4: I have no idea why it would happen. But given the type it's a possibility. No it will not be updated later.

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
rudy-6-4: Yes. It's a short time (default is 4s) and is the same behavior as before PR. (the code was moved to a function to avoid duplication).

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
```diff
@@ -0,0 +1,705 @@
...
+            reorg_out_blocks = ?reorg_out_blocks,
+            "No rows updated when updating reorg out delegation"
+        );
+    }
+    // clean table past blocks except for errors
+    let cleaned = sqlx::query!(
+        r#"
+        DELETE FROM delegate_user_decrypt
+        WHERE block_number < $1
+        AND gateway_nb_attempts = 0
```
obatirou, line 678: any delegate_user_decrypt that had a retry will not be cleaned even if it succeed?

rudy-6-4: yes errors will stay in db for now, once it's running correctly for some time we can have a look and do some extra clean up

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
```diff
@@ -0,0 +1,705 @@
...
+            .await;
+        let operation = if delegation.new_expiration_date == 0 {
+            "RevokeUserDecryptionDelegation"
+        } else {
+            "DelegateUserDecryption"
+        };
+        let transaction = match transaction {
+            Ok(txn) => txn,
+            Err(e) if self.idempotency_error(&e).is_some() => {
+                warn!(
```
obatirou, line 148: Shouldn't a counter be incremented like other error cases?

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
```diff
@@ -0,0 +1,705 @@
...
+                TxResult::OtherError(e) => {
+                    nb_errors += 1;
+                    update_error_delegation(&mut tx, &delegation, &e.to_string()).await;
+                    other_error = true
+                }
+            }
+        }
+        tx.commit().await?;
+        DELEGATE_USER_DECRYPT_SUCCESS_COUNTER.inc_by(nb_success);
+        DELEGATE_USER_DECRYPT_FAIL_COUNTER.inc_by(nb_errors);
```
obatirou, line 526: Isn't it already done inside of send_transaction?

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
rudy-6-4: idempotency error mean the information is already on gateway chain or is already replaced by something more up to date. If you have a better name suggestion "idempotency_error" ?
E.g. in case of a catchup where something is replayed.
So the error reflect the fact that the final state on gateway blockchain is good wrt to the transaction.

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
rudy-6-4: you are right, fixing it by removing intermediate individual counts

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
obatirou: What I meant is that `DELEGATE_USER_DECRYPT_FAIL_COUNTER` was increased for all errors except this one.
Now that intermediate individual counts are removed this is not relevant anymore anyway

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -549,15 +556,43 @@ impl Database {
                 )
                 .await?;
             }
+            AclContractEvents::DelegatedForUserDecryption(delegation) => {
```
isaacdecoded, line 560: Should we include some `info!` here similar to `AclContractEvents::AllowedForDecryption` for some debugging logging? Or is this something that is already logged somewhere else? @rudy-6-4

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
```diff
@@ -36,26 +48,53 @@ impl<P: alloy::providers::Provider<Ethereum> + Clone + 'static> NonceManagedProv
...
-                .get_next_nonce(&self.provider, signer_address)
-                .await?;
-            tx.nonce = Some(nonce);
-        }
-        let res = self.provider.send_transaction(tx).await;
-        if res.is_err() {
-            // Reset the nonce manager if the transaction sending failed.
-            *self.nonce_manager.lock().await = Default::default();
+        let mut res = Err(RpcError::UnsupportedFeature("Not reachable"));
+        // res cannot be returned non initialized
```
isaacdecoded: nit: maybe rewrite this comment for better understanding. I mean, "non initialized" refers to "Not reachable"?

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
```diff
@@ -36,26 +48,53 @@ impl<P: alloy::providers::Provider<Ethereum> + Clone + 'static> NonceManagedProv
...
+                if is_nonce_error(err) {
+                    let msg = err.to_string();
+                    warn!(msg, "Transaction failed due to nonce, resetting nonce manager and retrying immediately");
+                    continue;
+                }
+                warn!(err = ?err, "Transaction failed, resetting nonce manager");
+            }
+            return res;
         }
+        error!("Transaction failed multiple time due to nonce, aborting");
```
isaacdecoded:
```suggestion
        error!("Transaction failed too many times due to nonce issue. Aborting retries.");
```

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -663,19 +690,67 @@ impl Database {
...
+            VALUES($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, false, false)
+            ON CONFLICT DO NOTHING",
+            &delegator.into_array(),
+            &delegate.into_array(),
+            &contract_address.into_array(),
+            delegation_counter as i64,
+            old_expiration_date as i64,
+            new_expiration_date as i64,
+            chain_id as i64,
+            block_number as i64,
```
obatirou, line 737: Not sure this is safe, casting u64 to i64?
Not a problem now but might become one later.

coprocessor/fhevm-engine/host-listener/tests/integration_test.rs
```diff
@@ -526,3 +521,105 @@ async fn test_catchup_only() -> Result<(), anyhow::Error> {
...
+    let delegate = *acl_contract.address();
+    let contract_address = *acl_contract.address();
+    for (i_wallet, wallet) in wallets.iter().enumerate() {
+        let expiration_date = 3600_u64 + i_wallet as u64;
+        let wallet = wallet.clone();
+        let acl_contract = acl_contract.clone();
+        let url = url.to_string();
+        let thread = tokio::spawn(async move {
+            let delegation_counter = UNIQUE_INT.fetch_add(1, Ordering::SeqCst);
+            for _ in 1..=NB_DELEGATION_PER_WALLET {
```
obatirou, line 548: shouldn't the delegation_counter be inside the for loop?
else it will be the same delegation_counter for all delegation for a same wallet if not mistaken

rudy-6-4: we want each event to be different to be on database, it's done using this counter. Note that the test only check the event ends to the database. all the logic is now in tx-sender

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
```diff
@@ -0,0 +1,703 @@
...
+use fhevm_gateway_bindings::multichain_acl::MultichainACL::MultichainACLErrors;
+
+#[derive(Clone, Debug)]
+pub struct DelegationRow {
+    pub key: i64,
+    pub delegator: Vec<u8>,
+    pub delegate: Vec<u8>,
+    pub contract_address: Vec<u8>,
+    pub delegation_counter: u64,
+    #[allow(dead_code)]
```
obatirou, line 49: Why #[allow(dead_code)]?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: Delegations are logged later by the transaction-sender (when applied). But yes we could log it here also with other allow.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: Depend on the use case. (e.g. cast and cast back is safe).
For which value do you think it's risky ? timestamp, chain id block, number ?

obatirou: `new_expiration_date` as it can be a user input.
One might put MAX::u64 and this will wrap

obatirou: https://github.com/zama-ai/fhevm/blob/2b7f1343b735cd16576430b2d025f9c3ea820002/library-solidity/lib/FHE.sol#L8991

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
rudy-6-4: I think it was temporary and was not removed after (the type was in host-listener before). I am checking.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: for info the reading part in transaction-sender does the reverse i64 to u64

test-suite/e2e/contracts/DelegateUserDecryptDelegate.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+pragma solidity ^0.8.24;
+
+import "@fhevm/solidity/lib/FHE.sol";
+import {E2ECoprocessorConfig} from "./E2ECoprocessorConfigLocal.sol";
+
+
+/// @notice Contract for demonstrating user decryption with delegation
+contract DelegateUserDecryptDelegate is E2ECoprocessorConfig {
```
isaacdecoded, line 10: nit: or just `UserDecryptDelegate` for the contract name, since it seems verbose having "Delegate" twice.

test-suite/e2e/contracts/DelegateUserDecryptDelegator.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+pragma solidity ^0.8.24;
+
+import "@fhevm/solidity/lib/FHE.sol";
+import {E2ECoprocessorConfig} from "./E2ECoprocessorConfigLocal.sol";
+
+
+/// @notice Contract for demonstrating user decryption with delegation
+contract DelegateUserDecryptDelegator is E2ECoprocessorConfig {
```
isaacdecoded, line 10: nit: same as above, but "UserDecryptDelegator"

test-suite/e2e/contracts/DelegateUserDecryptDelegate.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+pragma solidity ^0.8.24;
+
+import "@fhevm/solidity/lib/FHE.sol";
+import {E2ECoprocessorConfig} from "./E2ECoprocessorConfigLocal.sol";
+
+
+/// @notice Contract for demonstrating user decryption with delegation
```
isaacdecoded:
```suggestion
/// @notice Contract that simulates the "delegate" account for a user decryption with delegation demonstration.
```

test-suite/e2e/contracts/DelegateUserDecryptDelegator.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+pragma solidity ^0.8.24;
+
+import "@fhevm/solidity/lib/FHE.sol";
+import {E2ECoprocessorConfig} from "./E2ECoprocessorConfigLocal.sol";
+
+
+/// @notice Contract for demonstrating user decryption with delegation
```
isaacdecoded:
```suggestion
/// @notice Contract that simulates the "delegator" account for a user decryption with delegation demonstration.
```

test-suite/e2e/test/delegateUserDecrypt/delegateUserDecryption.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    this.contractDelegate = await contractFactoryDelegate.connect(this.signers.alice).deploy();
+    await this.contractDelegate.waitForDeployment();
+    this.contractDelegateAddress = await this.contractDelegate.getAddress();
+
+    this.instances = await createInstances(this.signers);
+  });
+
+  const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+  it('test delegate user decrypt ebool', async function () {
```
isaacdecoded: Not sure, but it seems that this test case is testing the delegate and revoke use cases, not any decryption, right? In such a case, maybe better to specify this, e.g., `test delegation and revocation of user decryption`

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
```diff
@@ -0,0 +1,703 @@
...
+        }
+        let mut other_error = false;
+        let mut transient_error = false;
+        let mut nb_success = 0;
+        let mut nb_errors = 0;
+        while let Some(res) = join_set.join_next().await {
+            let Ok((delegation, tx_result)) = res else {
+                transient_error = true;
+                error!("Join error in delegation transaction sending");
+                continue;
```
obatirou: Should do a `nb_errors += 1;`  here ?

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
```diff
@@ -0,0 +1,703 @@
...
+                    update_error_delegation(&mut tx, &delegation, "transient_error").await;
+                }
+                TxResult::OtherError(e) => {
+                    nb_errors += 1;
+                    update_error_delegation(&mut tx, &delegation, &e.to_string()).await;
+                    other_error = true
+                }
+            }
+        }
+        tx.commit().await?;
```
obatirou, line 524: Correct me if wrong as my knowledge of postgres is limited. What happens if there is a DB error in one of update_error_delegation,  update_error_delegation or update_transmitted_delegation?
The commit is done only once when everything is proceed.
If there is one error, all update will be aborted no?

isaacdecoded (dismissed):
Just a few comments but LGTM overall 👍

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
rudy-6-4: yes an error in db update cancel the full db transaction.

obatirou: Isn't it a problem?
All transactions would have been sent but the update in DB would not be done

rudy-6-4: The net effect is that they will be retried later which is ok thank to the delegation counter.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: So MAX will be converted to a -1  in db, and when read back converted to MAX. That part is fine which is what the code do for now.
But a query, based on order over new_expiration_date could be wrong (like sorting by).

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: Will do

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
rudy-6-4: Removed

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: Replaced by bigdecimal

test-suite/e2e/contracts/DelegateUserDecryptDelegator.sol
rudy-6-4: Kept

test-suite/e2e/contracts/DelegateUserDecryptDelegator.sol
rudy-6-4: Corrected

test-suite/e2e/test/delegateUserDecrypt/delegateUserDecryption.ts
rudy-6-4: Corrected

test-suite/e2e/contracts/DelegateUserDecryptDelegate.sol
rudy-6-4: Corrected

test-suite/e2e/contracts/DelegateUserDecryptDelegate.sol
rudy-6-4: Kept

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
rudy-6-4: Supressed

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
rudy-6-4: Suppressed

rudy-6-4:
Note this is already in 0.10.0-3.

antoniupop (approved):

isaacdecoded (approved):

PR #1100, opened 2025-10-20
fix(coprocessor): Harmonize the metrics endpoint across all services

goshawk-3:
- fixes: https://github.com/zama-ai/fhevm-internal/issues/494
- fixes: https://github.com/zama-ai/fhevm-internal/issues/496


coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
```diff
@@ -130,6 +137,14 @@ struct Conf {
     /// service name in OTLP traces
     #[arg(long, default_value = "txn-sender")]
     pub service_name: String,
+
+    /// Prometheus metrics: coprocessor_l1_txn_latency_seconds
+    #[arg(long, default_value = "0.1:60.0:0.1", value_parser = clap::value_parser!(MetricsConfig))]
+    pub metric_l1_txn_latency: MetricsConfig,
```
dartdart26: Why is this called "l1" latency?

dartdart26 (approved):
Looks good! I think there's just a typo in txn-sender (l1), unless I miss something.

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -574,6 +552,7 @@ async fn execute_partition(
                 }
                 continue 'tx;
             }
+            drop(s);
```
antoniupop, line 555: Does that automatically call s.end() ? I'm not sure this drop is needed in any case wouldn't it happen when running out of scope?

antoniupop (approved):

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
goshawk-3: It measures the duration of processing (in copro) a single L1 transaction.

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
goshawk-3: It calls `s.end()`, we can change it to s.end() or remove it but it's not an issue for now.

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
dartdart26: How do we measure it exactly, not sure I understood that part?

dartdart26: I guess we could call it `host` instead, as a separate note. To be consistent with how we name it elsewhere and in specs too.

goshawk-3: True. `coprocessor_host_txn_latency_seconds` sounds reasonable. 

Simply put, it measures the time between the arrival of a host txn in the host-listener and the successful executions of all corresponding operations of the txn ( like pbs, tfhe, l2 submissions )

goshawk-3: Done

PR #1102, opened 2025-10-20
chore(common): update CODEOWNERS

aquint-zama:


.github/CODEOWNERS
```diff
@@ -1,7 +1,7 @@
...
 
-# All pull request should be reviewed by at least one of the members of httpz-devs
-*    @zama-ai/httpz-devs
+# All pull request should be reviewed by at least one of the members of fhevm-devs
+*    @zama-ai/fhevm-devs
 
-# All CI pull request should be reviewed by at least one of the members of httpz-devops
-.github/workflows/common-docker.yml	@zama-ai/httpz-devops
+# All CI pull request should be reviewed by at least one of the members of fhevm-devops
+.github/workflows/*release.yml	@zama-ai/fhevm-devops
```
eudelins-zama, line 7: The comment states that all CI PR should be reviewed by fhevm-devops. (not sure if it's correct, let me know what you think)

But the regex will catch only `.github/workflows/charts-helm-release.yml` right?

aquint-zama: right now yes, maybe we should add all docker build workflow with `on release` trigger?
we have to find a tradeoff

tawadaa (approved):

PR #1103, opened 2025-10-20
feat(protocol-contracts): add automated deployment script

zmalatrax:
Streamline the deployment of the OFT on Ethereum Sepolia & Arbitrum Sepolia.
The script is fully automated, and corresponds to the steps 4, 5 (optional) and 6 of the `protocol-contracts/README.md`.

It assumes that the previous steps 1-3 were properly done.
The script has a few checks to verify the required setup.

### DIFF
--- a/protocol-contracts/README.md
+++ b/protocol-contracts/README.md
@@ -2,7 +2,7 @@
 
 ## Step 1 : Initial setup
 
-```
+```bash
 cp .env.example .env
 ```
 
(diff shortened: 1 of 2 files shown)


aquint-zama:
@cla-bot check

protocol-contracts/pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
```
jatZama: How come the lock file did change?

jatZama: Pinned versions are coming from the `create-lz-oapp` official tool, and this is why Ihad to use pnpm here. Can you revert changes to this file maybe?

zmalatrax: I have a different local pnpm version (v10.18.3) than the one provided by the layerzero CLI `npx create-lz-oapp@latest`

Will revert this change

jatZama (approved):

PR #1109, opened 2025-10-20
chore(gateway-contracts): implement payment contract

melanciani:
refs https://github.com/zama-ai/fhevm-internal/issues/502
closes https://github.com/zama-ai/fhevm-internal/issues/489
closes https://github.com/zama-ai/fhevm-internal/issues/478


gateway-contracts/.env.example
```diff
@@ -138,14 +138,24 @@ PAUSER_ADDRESS_0="0x6591319B97979Acc59b7191A8B4Ec381375bFc92" # accounts[23] (ad
...
 
+# Protocol payment prices
+# The prices are in $ZAMA using 18 decimals
+INPUT_VERIFICATION_PRICE="10000000000000000000" # (uint256, 10 $ZAMA)
+PUBLIC_DECRYPTION_PRICE="10000000000000000000"  # (uint256, 10 $ZAMA)
+USER_DECRYPTION_PRICE="1000000000000000000" # (uint256, 1 $ZAMA)
+
+# OFT contracts
+ZAMA_TOKEN_OFT_ADDRESS="0xc1D733116990ce3D9e54F9eCf48a1cdD441Af4f9"
+PROTOCOL_FEES_OFT_ADDRESS="0xa50F5243C70c80a8309e3D39d8c9d958cDa83979"
```
jatZama: This is wrong name, please use the one I told you instead

jatZama: Chose either `FEES_SENDER_ADDRESS` or `FEES_SENDER_TO_BURNER_ADDRESS`

melanciani: yes will use `FEES_SENDER_TO_BURNER_ADDRESS`

gateway-contracts/contracts/ProtocolPayment.sol
```diff
@@ -0,0 +1,209 @@
...
+        ProtocolPaymentStorage storage $ = _getProtocolPaymentStorage();
+        $.userDecryptionPrice = price;
+        emit NewUserDecryptionPrice(price);
+    }
+
+    /**
+     * @notice See {IProtocolPayment-sendBalanceToBurner}.
+     * @dev This function can be called by anyone, not just the gateway owner.
+     */
+    function sendBalanceToBurner() external payable virtual {
```
jatZama: There is a function `sendFeesToBurner` in the FeesSenderToBurner contract that should be called there by anyone, not by this contract.
This contract should just do a transferFrom to the FeesSenderToBurner contract.

melanciani: you mean `transfer`, not `transferFrom` right ? 

but yes agree, it's simpler like you suggest, just updated the PR

jatZama:
I guess the "chore" in the commit message is a bit optimistic here, i would have used "feat" instead

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -112,7 +112,7 @@ jobs:
       # 2) Copy addresses from previous version to root directory: the upgrade tasks need to use the
       # internal addresses that have been deployed (ie, the previous version's addresses)
       # 3) Deploy mocked payment bridging contracts: this can be removed once fhevm 0.10 is released
-      # as the docker-compose file will consider the `deploy-mocked-payment-bringing-contracts` service
+      # as the docker-compose file will consider the `deploy-mocked-payment-bridging-contracts` service
```
jatZama: Do we need this for testnet and mainnet as well? I think this is just for devnet right?

melanciani: by "this" what do you mean exactly ? but yes anything related to mocked are only for devnet at most

for testnet/mainnet, only the addresses for ZamOFT and FeesSenderToBurner are required ?

gateway-contracts/contracts/interfaces/IProtocolPayment.sol
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * @notice Sets the price in $ZAMA for a user decryption (including delegated user decryption).
+     * @param price The price of the user decryption in $ZAMA base units (using 18 decimals).
+     */
+    function setUserDecryptionPrice(uint256 price) external;
+
+    /**
+     * @notice Sends all the protocol fees (the contract's balance) to the burner contract.
+     */
+    function sendBalance() external;
```
jatZama: You don't need this function. The contract should directly transferFrom from msg.sender to FeesSenderToBurner. I thought this is what was we agreed on...

gateway-contracts/contracts/shared/ZamaOFTUtils.sol
```diff
@@ -0,0 +1,60 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { zamaOFTAddress } from "../../addresses/PaymentBridgingAddresses.sol";
+import { protocolPaymentAddress } from "../../addresses/GatewayAddresses.sol";
+import { IProtocolPayment } from "../interfaces/IProtocolPayment.sol";
+
+/**
+ * @title ZamaOFT utils
+ */
+abstract contract ZamaOFTUtils {
```
jatZama: This abstract contract should be removed. Payment logic should be entirely contained in the Payment contract, not duplicated in Decryption/InputVerification contract via the abstract contract. Otherwise upgrading payment logic will become way more complex, necessitating updating all contracts instead of just Payment.

gateway-contracts/contracts/ProtocolPayment.sol
```diff
@@ -0,0 +1,203 @@
...
+        uint256 initialUserDecryptionPrice
+    ) public virtual onlyFromEmptyProxy reinitializer(REINITIALIZER_VERSION) {
+        ProtocolPaymentStorage storage $ = _getProtocolPaymentStorage();
+
+        $.inputVerificationPrice = initialInputVerificationPrice;
+        $.publicDecryptionPrice = initialPublicDecryptionPrice;
+        $.userDecryptionPrice = initialUserDecryptionPrice;
+
+        // Allow the FeesSenderToBurner address with "infinite" allowance
+        ZAMA_OFT.approve(FEES_SENDER_TO_BURNER_ADDRESS, type(uint256).max);
```
jatZama: You don't need this line, I would delete it.

gateway-contracts/contracts/ProtocolPayment.sol
```diff
@@ -0,0 +1,203 @@
...
+        ProtocolPaymentStorage storage $ = _getProtocolPaymentStorage();
+        $.userDecryptionPrice = price;
+        emit NewUserDecryptionPrice(price);
+    }
+
+    /**
+     * @notice See {IProtocolPayment-sendBalance}.
+     * @dev This function can be called by anyone, not just the gateway owner.
+     */
+    function sendBalance() external virtual {
```
jatZama: Remove this function as well.

jatZama:
Here is my first detailed review: 
- I would delete the abstract contract `ZamaOFTUtils`.
- I would remove the line `ZAMA_OFT.approve(FEES_SENDER_TO_BURNER_ADDRESS, type(uint256).max);` from `ProtocolPayment`.
- I would remove the function `sendBalance()` from `ProtocolPayment`.

Since `ZamaOFTUtils` is removed, I would replace the _collectfee functions by calls to external functions on `ProtocolPayment`. 

In short, my initial idea:
1/ The relayer approves `ProtocolPayment` for the OFT *only* (it actually does NOT need to approve `InputVerification` neither `Decryption`).

2/ The relayer calls `userDecryptResponse` on `Decryption` contract, and inside that call:

- `Decryption` calls `ProtocolPayment.payUserDecryption(msg.sender)`.
- This function `payUserDecryption` is protected: it checks whether the caller is either `Decryption` (for userDecryption/publicDecryption) or `InputVerification` (for zkpokVerification), otherwise it reverts. If the check passes, it performs a `transferFrom` from the relayer to ProtocolPayment.

Pseudo-code of `ProtocolPayment.payUserDecryption` would be: 
```solidity
function payUserDecryption(address account) external {
    if(msg.sender != decryptionAddress) revert CallerIsNotDecryption();
    transferFrom(account, feesSenderToBurnerAddress, userDecryptionPrice);
}
```

gateway-contracts/tasks/deployment/paymentBridging.ts
```diff
@@ -0,0 +1,103 @@
...
+  console.log(`${name} deployed successfully at address: ${contractAddress}\n`);
+
+  return contractAddress;
+}
+
+// Deploy the mocked payment bridging contracts
+// Currently, only the ZamaOFT contract is deployed as the FeesSenderToBurner contract can be
+// simply mocked with a random address (there is no logic to test on this contract)
+// We keep the command general enough if we ever need to consider additional contracts in the future
+task("task:deployMockedPaymentBridgingContracts").setAction(async function (_, hre) {
```
isaacdecoded: Is it planned to have more than one payment bridging contract in the future? I'm wondering if having just a `task:deployMockedPaymentBridgingContract` is enough.

gateway-contracts/tasks/deployment/contracts.ts
```diff
@@ -199,5 +222,53 @@ task("task:deployAllGatewayContracts").setAction(async function (_, hre) {
...
   console.log("Contract deployment done!");
 });
+
+// Deploy all the contracts
+task("task:deployAllGatewayContracts").setAction(async function (_, hre) {
+  // Deploy all the setup contracts
+  await hre.run("task:deploySetupContracts");
+
+  // Deploy all the regular contracts
+  await hre.run("task:deployRegularContracts");
```
isaacdecoded: nit: not sure about the "regular" wording here. Another idea could be `deployBaseContracts` for all initially required contracts and `deployGatewayContracts` for these "regular" ones.

gateway-contracts/contracts/mockedPaymentBridging/ZamaOFT.sol
isaacdecoded, line 1: I think we are using kebab-case for directories (i.e. mocked-payment-bridging). Also, maybe we can put it into `/mocks` directory as `ZamaOFTMock`, but not sure if this way could have conflicts with the current generated mocks.

gateway-contracts/tasks/deployment/contracts.ts
```diff
@@ -199,5 +222,53 @@ task("task:deployAllGatewayContracts").setAction(async function (_, hre) {
...
+
+    // Compile the implementation contracts
+    // The deployEmptyUUPSProxies task has generated the contracts' addresses in `addresses/*.sol`.
+    // Contracts thus need to be compiled after deploying the EmptyUUPS proxy contracts in order to
+    // use these addresses. Otherwise, irrelevant addresses will be used and, although deployment will
+    // succeed, most transactions made to the contracts will revert as inter-contract calls will fail.
+    await hre.run("compile:specific", { contract: "contracts" });
+
+    console.log(`Deploy ${name} contract:`);
+    await hre.run(`task:deploy${name}`);
```
isaacdecoded, line 271: Shouldn't the `useInternalProxyAddress` also be propagated here instead of being fixed as `true`?

gateway-contracts/tasks/deployment/contracts.ts
```diff
@@ -199,5 +222,53 @@ task("task:deployAllGatewayContracts").setAction(async function (_, hre) {
...
+  // Deploy all the setup contracts, including the mocked payment bridging ones
+  await hre.run("task:deploySetupContracts", { deployMockedPaymentBridgingContracts: true });
+
+  // Deploy all the regular contracts
+  await hre.run("task:deployRegularContracts");
+});
+
+// Deploy a single contract, after the GatewayConfig contract has been deployed
+// The new contract address will be appended to the .env.gateway and GatewayAddresses.sol files
+task("task:deploySingleContract")
```
isaacdecoded, line 251: Is this really needed to have this task even when we already have one task for each contract? e.g. task:deployKMSGeneration, task:deployInputVerification, etc.

gateway-contracts/tasks/deployment/contracts.ts
```diff
@@ -199,5 +222,53 @@ task("task:deployAllGatewayContracts").setAction(async function (_, hre) {
...
+task("task:deployAllGatewayContracts").setAction(async function (_, hre) {
+  // Deploy all the setup contracts
+  await hre.run("task:deploySetupContracts");
+
+  // Deploy all the regular contracts
+  await hre.run("task:deployRegularContracts");
+});
+
+// Deploy all the contracts, including the mocked payment bridging ones (FOR TESTS ONLY)
+task("task:deployAllGatewayContractsForTests").setAction(async function (_, hre) {
```
isaacdecoded, line 241: Maybe another approach could be executing `task:deployMockedPaymentBridgingContracts` in the hardhat.config.ts [here](https://github.com/zama-ai/fhevm/pull/1109/files#diff-4174d32e7c30d8e24f30e119f6bf99a004d8cf98648f139900be43df1d159242R67), and this way we just have one `task:deployAllGatewayContracts`.

gateway-contracts/tasks/deployment/paymentBridging.ts
melanciani: initially I had 2 (FeesSenderBurner), that's why. I didn't want to take time to revert the general flow + it'll be useful if we never need to add more in the future

nonetheless since this is only for internal purpose, I agree I can even rename `depoyMockedZamaOFT` and it won't be a problem to rename it again later if needed

gateway-contracts/tasks/deployment/contracts.ts
melanciani: this it to differentiate tasks : 
- setup contracts : proxy, pauser + (optionally) mocked payment
- "regular" contracts
- "all" contracts : combination of both of the above


I can rename "implementationContracts" if you prefer, wdyt ?

gateway-contracts/contracts/mockedPaymentBridging/ZamaOFT.sol
melanciani: we already have `emptyProxy` and `emptyProxyGatewayConfig` dirs , so not sure I get your point

also, mock contracts (used for external purposes) are completely different from mocked ZamaOFT (use for internal tests), so I'm not sure why we would combine both in the same dir

.github/workflows/gateway-contracts-upgrade-tests.yml
jatZama: Yes.

gateway-contracts/tasks/deployment/contracts.ts
melanciani: since you create a new empty proxy in `deploySingleEmptyUUPSProxy`, you need the internal addresses anyway if we keep using the current flow

but overall I think we could remove the use of internal proxy flag, this is something to look later when we have more time : https://github.com/zama-ai/fhevm-internal/issues/499

gateway-contracts/tasks/deployment/contracts.ts
melanciani: no, I tried this initially but turns out we have to deploy all contracts in a single tasks when using for hardhat test, else we need some dedicated nonce management (let's avoid that) as the network state is reset each time we run a hardhat task using the hardhat network

in other words, running 2 hardhat tasks successively that each deploys come contracts will result in different contracts having the same address

gateway-contracts/tasks/deployment/contracts.ts
melanciani: yes, the tasks you mention don't deploy a proxy conctract. When infra is going to upgrade the net from 0.9 to 0.10, they need to be able to deploy a new empty proxy + upgrade it to the new implem, and this task is exactly this !

melanciani: you can look at the upgrade test CI for the flow we'll expect them to follow

gateway-contracts/contracts/mockedPaymentBridging/ZamaOFT.sol
isaacdecoded: Yes, indeed, I forgot about the emptyProxy directories. About combining with mocks, the idea was to make it more explicit that this is a mock. But yes, I guess having it in the mockedPaymentBridging directory is clear enough.

gateway-contracts/tasks/deployment/contracts.ts
isaacdecoded: Yes, "implementationContracts" seems more accurate to me. Again, it is an opinion.

gateway-contracts/tasks/deployment/contracts.ts
isaacdecoded: Ah, I got it 👍

gateway-contracts/tasks/deployment/contracts.ts
isaacdecoded: I see. Sounds good if we can remove the internal proxy flag later on. 👍

jatZama (approved):

PR #1112, opened 2025-10-21
feat(kms-connector): improved nonce manager

eudelins-zama:
Huge kudos to @nboisde for this one :pray: 

Closes https://github.com/zama-ai/fhevm-internal/issues/672
Closes https://github.com/zama-ai/fhevm-internal/issues/671

## Code changes

No longer rely on a `CachedNonceManager` from `alloy`, but from a custom one (also used in the `relayer`) that keeps track of which nonce were successfully used, which are free to be reused in case of failure etc.

This should reduce the number of `nonce too high/low` error caught in our logs.
With the current setup, if the tx-sender get a `nonce too high/low` error, it will reinit the `CachedNonceManager` via RPC call. The problem with that is that in general, there are multiple of these errors, so each task handling a tx will do this RPC call, and the `CachedNonceManager` willl provide the same nonce to all of these tasks (because of the RPC call), instead of an incremented one.
With this new setup, no RPC call is made on error, the failed nonce is just stored in a map to be reused by another tx later on.

## Risk

The main risk I see is that there is that the nonce manager never re-sync with the RPC node on error. So it could result in a dead lock if for some reason, the nonce manager is stuck with too high nonces...

### DIFF
--- a/kms-connector/crates/utils/src/conn.rs
+++ b/kms-connector/crates/utils/src/conn.rs
@@ -11,7 +11,7 @@ use alloy::{
             WalletFiller,
         },
     },
-    transports::http::reqwest::Url,
+    transports::http::reqwest::{self, Url},
 };
 use anyhow::anyhow;
 use sqlx::{Pool, Postgres, postgres::PgPoolOptions};
@@ -23,6 +23,9 @@ pub const CONNECTION_RETRY_NUMBER: usize = 5;
 
 /// The delay between two connection attempts.
 pub const CONNECTION_RETRY_DELAY: Duration = Duration::from_secs(2);
+
+/// Timeout for requests to the Gateway's RPC node.
+const REQUEST_TIMEOUT: Duration = Duration::from_mins(1);
 
 /// Tries to establish the connection with Postgres database.
 pub async fn connect_to_db(db_url: &str, db_pool_size: u32) -> anyhow::Result<Pool<Postgres>> {
@@ -55,7 +58,8 @@ type DefaultFillers = JoinFill<
 pub type GatewayProvider = FillProvider<JoinFill<DefaultFillers, ChainIdFiller>, RootProvider>;
 
 /// The default `alloy::Provider` used to interact with the Gateway using a wallet.
-pub type WalletGatewayProvider = NonceManagedProvider<WalletGatewayProviderFillers, RootProvider>;
+pub type WalletGatewayProvider =
+    NonceManagedProvider<FillProvider<WalletGatewayProviderFillers, RootProvider>>;
 pub type WalletGatewayProviderFillers = JoinFill<
     JoinFill<JoinFill<Identity, ChainIdFiller>, FillersWithoutNonceManagement>,
     WalletFiller<EthereumWallet>,
@@ -107,7 +111,11 @@ where
 
     let gateway_url =
         Url::from_str(gateway_url).map_err(|e| anyhow!("Invalid Gateway URL: {e}"))?;
-    let provider = provider_builder_new().connect_http(gateway_url);
+    let client = reqwest::ClientBuilder::new()
+        .timeout(REQUEST_TIMEOUT)
+        .build()?;
+    let provider = provider_builder_new().connect_reqwest(client, gateway_url);
+
     info!("Connected to Gateway's RPC node successfully");
     Ok(provider)
 }
--- a/kms-connector/crates/utils/src/provider/mod.rs
+++ b/kms-connector/crates/utils/src/provider/mod.rs
@@ -1,3 +1,6 @@
+mod nonce_manager;
+
+use crate::provider::nonce_manager::ZamaNonceManager;
 use alloy::{
     consensus::Account,
     eips::{BlockId, BlockNumberOrTag, Encodable2718},
@@ -7,29 +10,25 @@ use alloy::{
         U256,
     },
     providers::{
-        EthCall, EthCallMany, EthGetBlock, FilterPollerBuilder, GetSubscription,
-        PendingTransaction, PendingTransactionBuilder, PendingTransactionConfig,
-        PendingTransactionError, Provider, ProviderCall, RootProvider, RpcWithBlock, SendableTx,
-        fillers::{
-            BlobGasFiller, CachedNonceManager, FillProvider, GasFiller, JoinFill, NonceManager,
-            TxFiller,
-        },
+        EthCall, EthCallMany, EthGetBlock, FilterPollerBuilder, PendingTransaction,
+        PendingTransactionBuilder, PendingTransactionConfig, PendingTransactionError, Provider,
+        ProviderCall, RootProvider, RpcWithBlock, SendableTx,
+        fillers::{BlobGasFiller, FillProvider, GasFiller, JoinFill, NonceManager, TxFiller},
     },
     rpc::{
         client::NoParams,
+        json_rpc::ErrorPayload,
         types::{
             AccessListResult, Bundle, EIP1186AccountProofResponse, EthCallResponse, FeeHistory,
             Filter, FilterChanges, Index, Log, SyncStatus, TransactionReceipt, TransactionRequest,
             erc4337::TransactionConditional,
-            pubsub::{Params, SubscriptionKind},
             simulate::{SimulatePayload, SimulatedBlock},
         },
     },
-    transports::{TransportError, TransportResult},
+    transports::{RpcError, TransportError, TransportErrorKind, TransportResult},
 };
-use futures::lock::Mutex;
 use serde_json::value::RawValue;
-use std::{borrow::Cow, sync::Arc};
+use std::borrow::Cow;
 
 pub type FillersWithoutNonceManagement = JoinFill<GasFiller, BlobGasFiller>;
 
@@ -46,7 +45,7 @@ where
 {
     inner: FillProvider<F, P, N>,
     signer_address: Address,
-    nonce_manager: Arc<Mutex<CachedNonceManager>>,
+    nonce_manager: ZamaNonceManager,
 }
 
 impl<F, P> NonceManagedProvider<F, P>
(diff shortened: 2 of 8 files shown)


kms-connector/crates/utils/src/provider/mod.rs
```diff
@@ -124,19 +148,25 @@ where
...
         tx.set_nonce(nonce);
-        let res = self.inner.send_transaction(tx).await;
-        if res.is_err() {
-            // Reset the nonce manager if the transaction sending failed.
-            *self.nonce_manager.lock().await = Default::default();
+        let send_tx_result = self.inner.send_transaction(tx).await;
+
+        match &send_tx_result {
+            Err(e) if is_nonce_too_low(e) => {
+                self.nonce_manager.confirm_nonce(signer_addr, nonce).await;
```
rudy-6-4, line 133: we have similar issue in tx-sender. What does confirm_nonce improve ?

eudelins-zama: It's not just `confirm_nonce`, it is the overall `NonceManager` that is improved.

Basically, it keeps in memory which nonce were sent to the blockchain node, if the tx failed, the nonce is "released" to be reused by another tx, if it is successful, the `confirm_nonce` allows to confirm the nonce was used so it won't be reused

kms-connector/crates/utils/src/provider/mod.rs
```diff
@@ -109,6 +114,25 @@ where
...
+// See https://ethereum-json-rpc.com/errors
+const ETH_INVALID_INPUT_RPC_ERROR_CODE: i64 = -32000;
+
+/// Returns `true` if the RPC error is "nonce too low" or "already known", `false` otherwise.
+fn is_nonce_too_low(error: &RpcError<TransportErrorKind>) -> bool {
+    match error {
+        RpcError::ErrorResp(ErrorPayload { code, message, .. }) => {
+            if *code == ETH_INVALID_INPUT_RPC_ERROR_CODE {
+                let lowercase_msg = message.to_lowercase();
+                lowercase_msg.starts_with("nonce too low")
```
rudy-6-4, line 75: what about "error code -32003: Nonce too high" ? should it be release_nonce like other error.
we observe it when 2 tx are done and the first revert. the second can fail with "nonce too high".
Details on slack discussion https://zama-ai.slack.com/archives/C065F16D5U3/p1761207858275019

nboisde: In case of a nonce too high, normaly, the strategy should be to retry the failed transaction and assigning a nonce with the get_next_nonce, the underlying data structure will pick up the next available nonce, which can be a nonce that creates the gap.

eudelins-zama: Yes indeed in case of nonce too high the nonce should be released to be reused later

kms-connector/crates/utils/src/provider/nonce_manager.rs
```diff
@@ -0,0 +1,522 @@
...
+};
+use async_trait::async_trait;
+use futures::lock::{Mutex, MutexGuard};
+use std::collections::{BTreeSet, HashMap, hash_map::Entry};
+use std::sync::Arc;
+use tracing::debug;
+
+/// A robust, in-memory nonce manager for a scalable transaction engine.
+#[derive(Clone, Debug, Default)]
+pub struct ZamaNonceManager {
```
dartdart26, line 15: Not a big fan of the name - maybe it should say what it does and not name it Zama.

kms-connector/crates/utils/src/provider/mod.rs
```diff
@@ -109,34 +63,82 @@ where
     }
 }
 
+// See https://ethereum-json-rpc.com/errors
+const ETH_INVALID_INPUT_RPC_ERROR_CODE: i64 = -32000;
+
+/// Returns `true` if the RPC error is "nonce too low" or "already known", `false` otherwise.
+fn is_nonce_too_low(error: &RpcError<TransportErrorKind>) -> bool {
```
dartdart26, line 70: I am a bit uneasy about this function - do we know for a fact that it handles all cases of nonce too low, will all RPC nodes be consistent?

kms-connector/crates/utils/src/conn.rs
```diff
@@ -24,6 +24,9 @@ pub const CONNECTION_RETRY_NUMBER: usize = 5;
 /// The delay between two connection attempts.
 pub const CONNECTION_RETRY_DELAY: Duration = Duration::from_secs(2);
 
+/// Timeout for requests to the Gateway's RPC node.
+const REQUEST_TIMEOUT: Duration = Duration::from_mins(1);
```
dartdart26, line 28: Should we make this configurable?

kms-connector/crates/utils/src/conn.rs
```diff
@@ -107,7 +111,11 @@ where
 
     let gateway_url =
         Url::from_str(gateway_url).map_err(|e| anyhow!("Invalid Gateway URL: {e}"))?;
-    let provider = provider_builder_new().connect_http(gateway_url);
+    let client = reqwest::ClientBuilder::new()
+        .timeout(REQUEST_TIMEOUT)
```
dartdart26, line 115: Should we maybe use tokio::timeout? Will the client timeout handle all cases of no response?

kms-connector/crates/utils/src/conn.rs
```diff
@@ -24,6 +24,9 @@ pub const CONNECTION_RETRY_NUMBER: usize = 5;
 /// The delay between two connection attempts.
 pub const CONNECTION_RETRY_DELAY: Duration = Duration::from_secs(2);
```
dartdart26, line 25: Also configurable?

dartdart26 (approved):
Looks good!

I've put some comments that you might address now or not, not critical.

Also, maybe you can elaborate on your concern about the manager not syncing often with the network - is it a concern?

Was wondering - does getting occasional nonce errors create any serious issue? Do we observe issues on prod?

kms-connector/crates/utils/src/provider/mod.rs
eudelins-zama: No indeed. But I don't think we have another way for now...

There's only this [spec](https://ethereum-json-rpc.com/errors), but it too generic for what we want to achieve

It is mentioned in [quicknode docs](https://www.quicknode.com/docs/ethereum/error-references) but it's not a guarantee that all RPC nodes will do it this way.

Wdyt?

kms-connector/crates/utils/src/conn.rs
eudelins-zama: Here's the doc: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.timeout

I think it does what we want, and has the benefit to be located in a single place, so I prefer this over `tokio::timeout`. Also note that `alloy` use this client under the hood for many RPC calls, so we can't put `tokio::timeout` everywhere anyway

wdyt?

kms-connector/crates/utils/src/conn.rs
eudelins-zama: It is just for the initial connection to the initial connection to the db/kms-core

The config is already bloated with a lot of stuff that we are advised not to use (because hard to keep all 13 MPC node config in sync). So I don't think this would be useful tbh

kms-connector/crates/utils/src/conn.rs
eudelins-zama: Why not yes :slightly_smiling_face:

kms-connector/crates/utils/src/provider/nonce_manager.rs
eudelins-zama: Will try to come up with a better one

eudelins-zama:
> Was wondering - does getting occasional nonce errors create any serious issue? Do we observe issues on prod?

I would not say serious issue. But it will create (lots of) useless retries if the Gateway cache is full of txs, because of this (if not clear let me know):

> if the tx-sender get a nonce too high/low error, it will reinit the CachedNonceManager via RPC call. The problem with that is that in general, there are multiple of these errors, so each task handling a tx will do this RPC call, and the CachedNonceManager willl provide the same nonce to all of these tasks (because of the RPC call), instead of an incremented one.

eudelins-zama:
> Also, maybe you can elaborate on your concern about the manager not syncing often with the network - is it a concern?

Let's say we send tx1, tx2 and tx3 with nonce 1, 2 and 3. RPC node returns OK for tx1 and tx2 and an error for tx3.

Now if for some reason the OK returned by the RPC node for nonce 2 was a "false positive" and that in reality, this nonce is not consumed, then when the tx-sender will retry tx3, it will try to send it with nonce 3 instead of 2, and this forever as the NonceManager never "re-sync" with the RPC node

I tried to stress tests the NonceManager on devnet to see if I could produce this scenario, but no, I did not manage to.

In theory, this should not be possible, but well, we never know...

kms-connector/crates/utils/src/conn.rs
dartdart26: Looks good, I was thinking of websockets, but we don't use them here (as we do in the coprocessor). So maybe my comment is indeed irrelevant.

kms-connector/crates/utils/src/provider/mod.rs
dartdart26: To put it differently - what would happen if this function doesn't work as intended and misses nonce errors?

eudelins-zama: Then, the nonce would be stored in the `available_nonces` for later reuse instead of being confirmed. As the nonce is too low, this will cause many useless errors

Eikix:
Is this PR stale or still ongoing?

eudelins-zama:
Closing for now as we don't currently have issue management on mainnet, so the risk is not worth it.

I'll reopen it if that changes.

PR #1116, opened 2025-10-22
feat(protocol-contracts): add PauserSetWrapper contract

jatZama:


protocol-contracts/pauserSetWrapper/README.md
```diff
@@ -0,0 +1,13 @@
+# Sample Hardhat Project
```
zmalatrax, line 1: Update readme to briefly describe contract purpose ?

protocol-contracts/pauserSetWrapper/contracts/PauserSetWrapper.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: UNLICENSED
```
zmalatrax: Are all contracts UNLICENSED?

protocol-contracts/pauserSetWrapper/contracts/PauserSetWrapper.sol
```diff
@@ -0,0 +1,23 @@
...
+pragma solidity ^0.8.24;
+
+import {IPauserSet} from "@fhevm/host-contracts/contracts/interfaces/IPauserSet.sol";
+
+contract PauserSetWrapper {
```
zmalatrax, line 6: Add NatSpec to contract/error/function

protocol-contracts/pauserSetWrapper/contracts/PauserSetWrapper.sol
melanciani: in gateway and host, we use `// SPDX-License-Identifier: BSD-3-Clause-Clear`, I think it should be the same for protocol contracts (seems that ZamaERC20 also needs to be changed)

protocol-contracts/pauserSetWrapper/README.md
jatZama: Can you do it in a future PR?

protocol-contracts/pauserSetWrapper/contracts/PauserSetWrapper.sol
jatZama: Can you do it in a future PR?

protocol-contracts/pauserSetWrapper/contracts/PauserSetWrapper.sol
jatZama: Good catch, will fix this now

jatZama: Fixed in latest commit

zmalatrax (approved):

PR #1117, opened 2025-10-22
test(test-suite): update to fhevm 0.9.3 and kms-core to 0.12.2

melanciani:


jatZama:
Maybe it would make sense to add transferOwnership tasks on host and gateway before merging, wdyt?

dd23 (approved):
Successfully tested locally. 👍

PR #1119, opened 2025-10-22
chore(contracts): add comment on optimizer and viaIR config

melanciani:


melanciani:
closing this as we need further investigation : https://github.com/zama-ai/fhevm-internal/issues/500

PR #1124, opened 2025-10-22
feat(host-contracts): ACL deny list

obatirou:
- Extend `ACL.sol` with a deny-list mapping, owner-only `blockAccount`/`unblockAccount`, and an `isAccountDenied` helper.
  - Guard `allow`, `allowForDecryption`, and `allowTransient` so denied callers revert.
  - Expand `acl.t.sol` with helpers and scenarios covering pause constraints, ownership checks, and denial reverts across all entry points.


host-contracts/contracts/ACL.sol
```diff
@@ -454,6 +476,34 @@ contract ACL is
...
+    function isAccountDenied(address account) public view virtual returns (bool) {
+        ACLStorage storage $ = _getACLStorage();
+        return $.denyList[account];
+    }
+
+    /**
+     * @notice Adds `account` to the deny list
+     * @param account Address of the account.
+     */
+    function blockAccount(address account) public virtual whenNotPaused onlyOwner {
```
jatZama: I don't think the `whenNotPaused` modifier makes sense here, and same thing for `unblockAccount`. Owner could immediately unpause, so there is not good reason to use both in conjunction. Do you agree?

0xalexbel: Agree with you. I followed the pattern for the first implementation. We can remove it and the corresponding tests as well.

jatZama: On the other hand, in addition to remove `whenNotPaused` modifier on both block/unblock functions, I think some important stuff is missing here: 
- Please add events BlockAccount(account) and UnblockAccount(account), those could be very useful for tracking and are needed for such important actions.
- check that an already blocked account cannot be blocked again, otherwise revert with a custom error.
- same for unblocking, should not be possible to unblock an already unblocked.

0xalexbel: The revert was also part of the pending questions I had.
Adding events is a great suggestion. Let's add them

host-contracts/contracts/ACL.sol
```diff
@@ -213,6 +230,11 @@ contract ACL is
                 revert SenderNotAllowed(msg.sender);
             }
         }
+
+        if (isAccountDenied(msg.sender)) {
```
jatZama: I think this check should be inside the `if (msg.sender != fhevmExecutorAddress)` just in case, although fhevmExecutorAddress should never be denied normally.

jatZama: Another reason is to save gas, because the branch `msg.sender==fhevmExecutorAddress` happens very very frequently (for all transient allowances coming from computation).

0xalexbel: Good catch! I missed this one! 💪

host-contracts/contracts/ACL.sol
```diff
@@ -94,7 +94,13 @@ contract ACL is
     error SenderNotAllowed(address sender);
 
     /**
-     * @notice Struct that represents a user decryption delegation.
+     * @notice Returned if the sender address is in the deny list.
+     * @param sender The address of the account that is denied.
+     */
+    error SenderDenied(address sender);
```
obatirou, line 112: Should this be `AccountDenied` instead ?

jatZama: SenderDenied is fine, similar to already existing SenderNotAllowed

host-contracts/contracts/ACL.sol
```diff
@@ -24,6 +24,30 @@ contract ACL is
     ACLEvents,
     MulticallUpgradeable
 {
+    /**
+     * @notice Emitted when an account is added to the deny list.
+     * @param account The address of the account that is blocked.
+     */
+    event BlockedAccount(address account);
```
jatZama: nit: I think it could be better to make the `account` argument indexed, wdyt? This is common practice for addresses arguments, for eg in ERC20 Transfer events. Same thing for the `UnblockedAccount` event.

jatZama: More importantly, these events should not be put in ACL.sol but in ACLEvents fyi.

obatirou: Done with 50eaab1

aquint-zama:
@cla-bot check

obatirou:
Closing 
Superseded by https://github.com/zama-ai/fhevm/pull/1126 due to ci permissions issues

PR #1125, opened 2025-10-22
chore(contracts): add ownership tasks

melanciani:
closes https://github.com/zama-ai/fhevm-internal/issues/501


host-contracts/.env.example
```diff
@@ -17,8 +17,8 @@ COPROCESSOR_SIGNER_ADDRESS_1="0xa5eE8292dA52d8234248709F3E217ffEBA5E8312" # acco
 COPROCESSOR_SIGNER_ADDRESS_2="0xCFbF539CB91c92ace0343c5B0487149Ad0b82078" # account[13]
 COPROCESSOR_SIGNER_ADDRESS_3="0xA951F315d5FD35Cac111dFB5250DF231FB8eF905" # account[14]
 ETHERSCAN_API_KEY="XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
-RPC_URL="http://anvil-node:8545"
-HARDHAT_NETWORK="staging"
+RPC_URL="http://127.0.0.1:8757"
```
melanciani: this (+ changes in docker-compose and hardhat.config) are made to be able to run local config easily (like `make get-accounts`), like we do in gateway

shouldn't change anything major

gateway-contracts/tasks/ownership.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { getRequiredEnvVar } from "./utils/loadVariables";
+import { pascalCaseToSnakeCase } from "./utils/stringOps";
+
+task(
+  "task:transferGatewayOwnership",
+  "Transfers ownership of the gateway contracts to the provided address. This can only be used if the current owner is an EOA.",
+)
+  .addParam(
+    "currentOwnerPrivateKey",
```
jatZama: I don't think it makes sense to expose private key as a CLI argument, for the deployment tasks it was read from env variables, i think we can do the same here. I am concerned otherwise this won't be usable by infra team... could reuse `const currentOwnerPrivateKey = getRequiredEnvVar('OWNER_PRIVATE_KEY');`

gateway-contracts/tasks/ownership.ts
```diff
@@ -0,0 +1,66 @@
...
+      `Ownership of GatewayConfig contract ${gatewayConfigContractAddress} successfully transferred to EOA ${newOwnerAddress}`,
+    );
+  });
+
+task(
+  "task:acceptGatewayOwnership",
+  `Accepts ownership of the gateway contracts. This can only be used if the new owner is an EOA.`,
+)
+  .addParam(
+    "newOwnerPrivateKey",
```
jatZama: same remark here, i think infra does not accept to put private key in a command line

gateway-contracts/tasks/ownership.ts
```diff
@@ -0,0 +1,66 @@
...
+    const gatewayConfigContractAddress = getRequiredEnvVar(gatewayConfigAddressEnvVarName);
+    const gatewayConfigContract = await ethers.getContractAt("GatewayConfig", gatewayConfigContractAddress);
+
+    // Transfer ownership of the GatewayConfig contract to the destination address.
+    const tx = await gatewayConfigContract.connect(currentOwner).transferOwnership(newOwnerAddress);
+
+    await tx.wait();
+
+    console.log(
+      `Ownership of GatewayConfig contract ${gatewayConfigContractAddress} successfully transferred to EOA ${newOwnerAddress}`,
```
jatZama: Can you correct this log message please?
The `newOwnerAddress` is not necessarily an EOA here (unlike the acceptOwnership task), so I would remove mention of EOA. Also I would precise that the receiving account should still accept ownership to validate ownership transfer, which is not clear from current message.
Maybe something like: `Ownership of GatewayConfig contract ${gatewayConfigContractAddress} is now successfully pending for account ${newOwnerAddress} - now awaiting an acceptGatewayOwnership tx from ${newOwnerAddress}  to validate ownership transfer`

host-contracts/hardhat.config.ts
```diff
@@ -87,7 +88,6 @@ function getChainConfig(chain: keyof typeof chainIds): NetworkUserConfig {
 }
 
 const config: HardhatUserConfig = {
-  defaultNetwork: 'sepolia',
```
jatZama, line 90: are you sure about that?

jatZama: i dont know if we measure well what are all implications here...

host-contracts/tasks/utils/stringOps.ts
```diff
@@ -0,0 +1,11 @@
+export function pascalCaseToSnakeCase(str: string) {
```
jatZama: Why is this needed in host? you don't use it...

melanciani: yes forgot to remove it thanks

host-contracts/tasks/ownership.ts
```diff
@@ -0,0 +1,59 @@
...
+import { task, types } from 'hardhat/config';
+
+import { getRequiredEnvVar } from './utils/loadVariables';
+
+task(
+  'task:transferHostOwnership',
+  'Transfers ownership of the host contracts to the provided address. This can only be used if the current owner is an EOA.',
+)
+  .addParam(
+    'currentOwnerPrivateKey',
```
jatZama: same remarks for both tasks as for gateway.

host-contracts/.env.example
jatZama: wdym by "run local config" exactly? not sure to understand here

melanciani: run tasks locally using hardhat network by default, eg be able to run `npx hardhat get-accounts` without having to specify the hardhat network

i think it's convenient for devX, and it makes sense as the env.example if only here to unit tests and local dev. Also, this is what we already do in the gateway !

jatZama: Then can you update the package.json please? currently we have special task "test:mock" replace it with "test".

melanciani: done, should I do the same with library solidity to align ?

host-contracts/hardhat.config.ts
melanciani: we checked with infra: the define HARDHAT_NETWORK each time, so we're good, thanks for the catch !

gateway-contracts/tasks/ownership.ts
melanciani: yes, but we only keep a transferOwnership using the deployer, as the ownership will only be transfered to the multisig governance

host-contracts/.env.example
jatZama: Yes sure.

host-contracts/package.json
```diff
@@ -15,12 +15,11 @@
     "lint:ts": "eslint --ignore-path ./.eslintignore --ext .js,.ts .",
     "prettier:check": "prettier --check \"**/*.{js,json,md,sol,ts,yml}\"",
     "prettier": "prettier --write \"**/*.{js,json,md,sol,ts,yml}\"",
-    "test:inband": "hardhat test",
-    "test": "HARDHAT_PARALLEL=1 hardhat test --parallel",
-    "test:mock": "hardhat test --network hardhat",
+    "test:parallel": "HARDHAT_PARALLEL=1 hardhat test --parallel",
```
jatZama: I think the test:parallel is deprecated, you should remove it.

library-solidity/.env.example
```diff
@@ -18,5 +18,5 @@ COPROCESSOR_SIGNER_ADDRESS_3="0xA951F315d5FD35Cac111dFB5250DF231FB8eF905" # acco
 NUM_PAUSERS="1"
 PAUSER_ADDRESS_0="0xa44366bAA26296c1409AD1e284264212029F02f1" # accounts[2] (address)
 ETHERSCAN_API_KEY="XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
-RPC_URL="http://anvil-node:8545"
-HARDHAT_NETWORK="staging"
\ No newline at end of file
+RPC_URL="http://127.0.0.1:8756"
```
jatZama, line 21: Are you sure we don't need this anymore?
I remember it was @tawadaa  who made this change long time ago.

jatZama: btw where is port defined? 8756? and where is it used?

melanciani: > btw where is port defined? 8756? and where is it used?

8756 is the default one used in hardhat.config, I just picked the same

> Are you sure we don't need this anymore?

the .env.example is only used for local dev and unit tests. Relevant values are always set in another env file for e2e testing, staging, devnet, ect 

you can check the gateway's env.example to compare if you want

jatZama (approved):

PR #1127, opened 2025-10-22
feat(protocol-contracts): update PauserSetWrapper implementation

jatZama:
Instead of letting the PauserSetWrapper able to forward any call to any function of any contract, like in previous implementation, in this new implementation we decided to only make it able to call the `pauseMinting()` function from the token. It makes it a lot less flexible, but this was actually necessary for security, since otherwise any single (potentially malicious) pauser could call the `renounceRole` method on the token inherited from `AccessControl.sol` contract, forbidding other honest pausers to pause until the admin would regrant the pauser role. (see https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.4.0/contracts/access/AccessControl.sol#L140-L155)


zmalatrax (approved):

melanciani (approved):
thanks !

PR #1129, opened 2025-10-23
Alex b/new codegen

0xalexbel:
### About Codegen
- Generates FHE.sol + Impl.sol
- Generates all the FHEVM operations test suite for FHE.sol

### The problem
1. Codegen is currently duplicated in several places:
    - `host-contracts/codegen`
    - `library-solidity/codegen`
    - `test-suite/e2e/codegen`

2. Each time we modify something in FHE.sol, we must modify:
    - `host-contracts/codegen` (currently host-contract uses a local copy of FHE.sol)
    - `library-solidity/codegen`

3. Each time a test is modified or improved we must propagate the change in each folder
    - `host-contracts/codegen`
    - `library-solidity/codegen`
    - `test-suite/e2e/codegen`

4. The tests generated by codegen for the 3 folders are:
    - identical for `host-contracts` & `library-solidity`
    - there is a minor difference with `test-suite/e2e`

5. Randomization issue: each time codegen is executed, a new `overloads.json` file is generated thus modifying all the generated tests even if the operations API has not changed. For example, if a new FHE API entry is added like `isAccountDenied`, the entire test suite is re-generated.

This issue has already been identified, but never solved.

### The solution
- Codegen should be a standalone CLI tool
- Codegen should run using a `codegen.config.json` argument
- Codegen should stay in library-solidity only.
- Codegen should be configurable to allow generation of tests adapted to each needs.

### Codegen Config:
- Number of splits: e2e requires 96 test splits, and host-contracts+library-solidity only 12 test splits.
- Shuffle: e2e shuffles the tests, host-contracts+library-solidity do not  shuffle the tests.
- Imports: e2e, host-contracts, library-solidity all uses custom imports. The new codegen allows you to customize the required imports.
- Randomization: The new codegen does not regenerate the `overloads.json` file if not needed. For e2e tests that require shuffling, an optional pseudo random flag has been added to make sure the random sequence always stays the same even if we need to regenerate shuffled tests.
- And many more...

### Related issue:
https://github.com/zama-ai/fhevm-internal/issues/423


immortal-tofu:
Why this codegen is in parent-folder? We should have this in library-solidity if it generates FHE.sol

codegen/src/main.ts
```diff
@@ -0,0 +1,305 @@
...
+  } else {
+    console.log(`Overloads is unchanged.`);
+  }
+}
+
+// function readOperatorsPrices(): PriceData {
+//   const p = path.resolve('./src/operatorsPrices.json');
+//   const jsonContent = readFileSync(p, 'utf8');
+//   return JSON.parse(jsonContent) as PriceData;
+// }
```
obatirou: why commented code? 
Same below

immortal-tofu:
Can you summarize in your desc what this PR is doing? The endgoal is to have codegen for FHE.sol in `library-solidity`, codegen for tests in `e2e-tests` codegen for host-contracts in... etc
Also, each package should use FHE.sol from `library-solidity` instead of having its local copy.

It's not clear which point is covered by your PR.

jatZama:
Can you explain how this is working exactly?
For instance I don't understand many fields in the config files, such as : 
```
{
  "numberOfTestSplits": 96,
  "noLib": true,
  "publicDecrypt": true,
  "shuffle": false,
  "shuffleWithPseuseRand": true,
  "directories": {
    "baseDir": "../test-suite/e2e"
  },
  "solidity": {
    "outDir": "./contracts/operations",
    "imports": ["@fhevm/solidity/lib/FHE.sol", ["E2ECoprocessorConfig", "../E2ECoprocessorConfigLocal.sol"]],
    "parentContract": {
      "name": "E2ECoprocessorConfig",
      "solidityFile": "../E2ECoprocessorConfigLocal.sol"
    }
  },
  "typescript": {
    "outDir": "./test/fhevmOperations",
    "imports": [
      ["instance", "../instance"],
      ["signers", "../signers"],
      ["typechain", "../../types"]
    ]
  }
}
```
Also it is hard to follow because every config file has different number of fields.

library-solidity/codegen/verify-e2e.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
jatZama: Why and where are all the verify .sh scripts needed and used?
What is the --debug flag for?

0xalexbel: 👍 The verify-xxx.sh scripts should be removed. The purpose is to check that the generated tests using new codegen are the same as the existing codegen.
👍 The --debug flag is not well named, it should be renamed as --verbose.

jatZama:
Can you elaborate on `Randomization: The new codegen also supports customizable values randomizations to avoid generating new tests even if the API has not changed.` ?

library-solidity/codegen/verify-host-contracts.sh
```diff
@@ -0,0 +1,41 @@
...
+OUT_BASE_DIR=$(jq -r '.directories.baseDir' ./codegen.host-contracts.config.json)
+SOL_REL_DIR=$(jq -r '.solidity.outDir' ./codegen.host-contracts.config.json)
+TS_REL_DIR=$(jq -r '.typescript.outDir' ./codegen.host-contracts.config.json)
+
+npm run build && ./codegen.mjs --overloads ./overloads/host-contracts.json --config ./codegen.host-contracts.config.json --debug 
+
+SOL_DIR="${FHEVM_DIR}/host-contracts/examples/tests-local.orig"
+TS_DIR="${FHEVM_DIR}/host-contracts/test/fhevmOperations-local.orig"
+
+OUT_SOL_DIR="${OUT_BASE_DIR}/${SOL_REL_DIR}"
```
jatZama: Shouldn't you also verify equality of FHE.sol and Impl.sol?

codegen/README.md
```diff
@@ -0,0 +1,38 @@
...
+# Generate e2e tests
+
+```sh
+cd codegen
+./codegen.mjs --overloads ./overloads/e2e.json --config ./codegen.e2e.config.json --debug
+```
+
+# Dry Run
+
+Use the `--dry-run --debug` options to check code generation in dry run mode.
```
obatirou: This needs to be updated with --verbose

immortal-tofu:
Remove codegen from library-solidity if your PR replaces it. It's always possible to retrive old codegen with git history if needed :)

And overall, I'd like this PR remove all codegen as expected. For now, this PR just add yet another new codegen :D

library-solidity/test/utils.ts
```diff
@@ -1,17 +1,18 @@
...
+import type { ContractMethodArgs, Signer } from 'ethers';
 import { ethers, network } from 'hardhat';
+import hre from 'hardhat';
 
 import type { Counter } from '../types';
 import { TypedContractMethod } from '../types/common';
 import { getSigners } from './signers';
 
-const hre = require('hardhat');
+//const hre = require('hardhat');
```
jatZama: Can you remove this comment?

library-solidity/package.json
```diff
@@ -19,12 +19,12 @@
     "test:mock": "hardhat test --network hardhat",
     "coverage:mock": "SOLIDITY_COVERAGE=true hardhat coverage",
     "typechain": "cross-env TS_NODE_TRANSPILE_ONLY=true hardhat typechain",
-    "codegen": "npx ts-node codegen/main.ts && npm run prettier",
-    "codegen:overloads": "npx ts-node codegen/generateOverloads.ts",
```
jatZama, line 21: why did this disappear?

library-solidity/test/utils.ts
0xalexbel: done!

library-solidity/lib-js/common.ts
```diff
@@ -0,0 +1,9 @@
+//
+// Warning
+// Hardhat v2 cannot resolve ts modules with .js extension
+//
+import { ArgumentType, OperatorArguments, ReturnType } from '../codegen/src/common';
```
jatZama, line 5: why are all the files in lib-js needed?

library-solidity/package.json
```diff
@@ -19,12 +19,12 @@
...
+    "codegen": "cd codegen && npm run codegen:library-solidity",
     "task:getEthereumAddress": "hardhat task:getEthereumAddress",
     "task:deployGateway": "hardhat task:deployGateway",
     "task:addRelayer": "hardhat task:addRelayer",
-    "task:removeRelayer": "hardhat task:removeRelayer"
+    "task:removeRelayer": "hardhat task:removeRelayer",
```
jatZama: Can you please take the chance to remove those 3 lines which are obsolete since a long time: 
```
"task:deployGateway": "hardhat task:deployGateway",
    "task:addRelayer": "hardhat task:addRelayer",
    "task:removeRelayer": "hardhat task:removeRelayer",
    ```

File: library-solidity/lib-js/common.ts (hunk 7 above)
Comment (0xalexbel) [reply library-solidity/lib-js/common.ts:5]: These are very important exports for host-contracts and possibly e2e. The idea is to get rid of codegen in host-contracts. But host-contracts still needs to import stuff from library-solidity. At this point I identified everything that is needed and regroup everything in lib-js. Maybe extra export will be needed in the future.

File: library-solidity/codegen/src/operatorsPrices.ts [hunk 9]
```diff
@@ -0,0 +1,473 @@
+import type { PriceData } from './common';
```
jatZama, line 1: Why is this a ts file instead of a json?
The Python script in cli-tools generates a json...

0xalexbel: Converting a json to a ts is trivial. Json import is problematic. Better use a ts instead. Additionally you control the Typings in a much safer way.

jatZama (commented):
You should remove codegen directories in host and e2e before merging.
Also I am wondering, where is the overloads being computed?
And why did you change the operatorsPrices form a json to a ts file? Fyi the cli-tools Python script is generating this as a .json, so it would be simpler to keep .json format.

library-solidity/lib-js/common.ts
jatZama: I don't understand, where are those used in your branch in host and e2e? As I can see it now, you kept codegen in both e2e and host.

library-solidity/codegen/src/operatorsPrices.ts
jatZama: I am not sure it is trivial? Where is the script converting the json to ts? Can you add it to streamline the cli-tools output if we need to use new values for example in a future upgrade?

library-solidity/package.json
0xalexbel: Fixed

library-solidity/codegen/src/pseudoRand.ts
```diff
@@ -0,0 +1,26 @@
+export const RND = [
```
jatZama: This function is risky I disagree to reimplement this in a custom manner sorry. I think this is better to generate pseudo random number by using a standard library. What happens now if the counter i exceeds RND.length: then you get cyclic behavior, not random one, so this is bad.

immortal-tofu:
I agree with JAT regarding deleting codegen in e2e and host contract. Maybe on host-contract a part is needed (I don't remember if we still generate a host contract?)

library-solidity/codegen/src/operatorsPrices.ts
jatZama: its ok we decided to copy paste the json from the python script and add import manually next time.

jatZama:
> I agree with JAT regarding deleting codegen in e2e and host contract. Maybe on host-contract a part is needed (I don't remember if we still generate a host contract?)

Yes HCULimit is codegenerated. But since you decided to move codegen in library, no need to keep duplicated codegen code there.

library-solidity/codegen/README.md
```diff
@@ -0,0 +1,84 @@
...
+```sh
+cd codegen
+npm run codegen:e2e
+```
+
+# Dry Run
+
+Use the `--dry-run --verbose` options to check code generation in dry run mode.
+
+# Verification scripts
```
jatZama: Delete them already please. And also delete codegen code from host and e2e.

library-solidity/codegen/verify-e2e.sh
```diff
@@ -0,0 +1,52 @@
...
```
jatZama: this should be deleted

library-solidity/codegen/verify-host-contracts.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
```
jatZama: this should be deleted

library-solidity/codegen/verify-library-solidity.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
```
jatZama: this should be deleted

library-solidity/codegen/README.md
0xalexbel: One step at a time. I suggest we do it after this PR

library-solidity/codegen/src/generateOverloads.ts
```diff
@@ -372,28 +429,15 @@ export const generateOverloads = (fheTypes: FheType[]) => {
                   encryptedTests.push(
                     safeEval(
                       test.evalTest,
-                      smallest - 4n,
                       smallest,
+                      smallest - 4n,
```
jatZama, line 433: why order changed here?

library-solidity/lib-js/common.ts
0xalexbel: That's right. lib-js is already used by library-solidity/tests and is also meant to be used by host-contracts and e2e in the next PRs

library-solidity/codegen/src/generateOverloads.ts
0xalexbel: The order has not changed (I checked the code side by side). Code is exactly the same as the original one. I guess this is mainly due to diff display.

library-solidity/codegen/src/pseudoRand.ts
0xalexbel: This is not a security-sensitive context. The goal is merely to roughly shuffle about one hundred tests. Using pseudo-random numbers is perfectly acceptable. I had initially generated a sufficiently large set of random numbers to ensure a thorough shuffle. However, if you know of a quality pseudo-random package, let's go for it ?

library-solidity/codegen/verify-library-solidity.sh
0xalexbel: Done

library-solidity/codegen/verify-host-contracts.sh
0xalexbel: Done

library-solidity/codegen/verify-e2e.sh
0xalexbel: Done

library-solidity/codegen/README.md
0xalexbel: Done. Now new codegen has been added to e2e and host

immortal-tofu (approved):

test-suite/e2e/test/utils.ts
```diff
@@ -55,13 +56,13 @@ export const waitForBlock = (blockNumber: bigint | number) => {
...
 };
 
-export const waitNBlocks = async (Nblocks: number) => {
-  const currentBlock = await ethers.provider.getBlockNumber();
-  if (network.name === 'hardhat') {
-    await produceDummyTransactions(Nblocks);
-  }
-  await waitForBlock(currentBlock + Nblocks);
-};
+// export const waitNBlocks = async (Nblocks: number) => {
```
obatirou: Commented code, that can be deleted?

test-suite/e2e/test/utils.ts
```diff
@@ -1,21 +1,22 @@
...
 import { HardhatEthersSigner } from '@nomicfoundation/hardhat-ethers/signers';
 import { toBufferBE } from 'bigint-buffer';
-import { ContractMethodArgs, Typed } from 'ethers';
+import { ContractMethodArgs, Log, TransactionReceipt, Typed } from 'ethers';
 import { Signer } from 'ethers';
 import { ethers, network } from 'hardhat';
+import hre from 'hardhat';
 
-import type { Counter } from '../types';
+//import type { Counter } from '../types';
```
obatirou: Commented code, that can be deleted?

test-suite/e2e/test/utils.ts
```diff
@@ -1,21 +1,22 @@
...
 import { TypedContractMethod } from '../types/common';
-import operatorsPrices from './operatorsPrices.json';
-import { getSigners } from './signers';
-import { ALL_FHE_TYPES } from './types';
 
-const hre = require('hardhat');
+//import { getSigners } from './signers';
```
obatirou: same

test-suite/e2e/test/utils.ts
```diff
@@ -90,25 +91,25 @@ export const createTransaction = async <A extends [...{ [I in keyof A]-?: A[I] |
...
-export const produceDummyTransactions = async (blockCount: number) => {
-  const contract = await deployCounterContract();
-  let counter = blockCount;
-  while (counter > 0) {
-    counter--;
-    const tx = await contract.increment();
-    const _ = await tx.wait();
-  }
-};
+// export const produceDummyTransactions = async (blockCount: number) => {
```
obatirou: Commented code

test-suite/e2e/test/utils.ts
```diff
@@ -90,25 +91,25 @@ export const createTransaction = async <A extends [...{ [I in keyof A]-?: A[I] |
...
+//   while (counter > 0) {
+//     counter--;
+//     const tx = await contract.increment();
+//     const _ = await tx.wait();
+//   }
+// };
 
-async function deployCounterContract(): Promise<Counter> {
-  const signers = await getSigners();
+// async function deployCounterContract(): Promise<Counter> {
```
obatirou: Commented code
see below also

jatZama (approved):

PR #1133, opened 2025-10-23
test(host-contracts): Add host stack deployment harness and assert ACL-denied accounts cannot persist FHE handles

obatirou:
Adds tests to ensure denied accounts cannot use the FHE lib function doing compute.
  - add `HostContractsDeployerTestUtils` to deploy the full host stack to its canonical proxy addresses inside Foundry tests
  - cover each deploy helper with unit tests that assert proxy wiring, initialization data, and version metadata
  - exercise an end-to-end ACL integration flow to ensure account blocking prevents FHE add/sub operations and unblock restores access

### DIFF
--- a/host-contracts/test/acl/TestIntegrationACL.t.sol
+++ b/host-contracts/test/acl/TestIntegrationACL.t.sol
@@ -6,6 +6,7 @@ import {HostContractsDeployerTestUtils} from "../utils/HostContractsDeployerTest
 import {ACL} from "../../contracts/ACL.sol";
 import {PauserSet} from "../../contracts/immutable/PauserSet.sol";
 import {FHE} from "../../lib/FHE.sol";
+import {FHEEvents} from "../../contracts/FHEEvents.sol";
 import {CoprocessorConfig} from "../../lib/Impl.sol";
 import {aclAdd, fhevmExecutorAdd, kmsVerifierAdd, pauserSetAdd} from "../../addresses/FHEVMHostAddresses.sol";
 
@@ -67,7 +68,12 @@ contract TestIntegrationACL is HostContractsDeployerTestUtils {
         caller.sub(lhs, rhs);
     }
 
-    function test_BlockingAccountDoesNotPreventAddWithoutAllowing() public {
+    // A denied account can still call the FHEVM Executor and emit Computation events.
+    // However, these events are most likely ignored on the coprocessor side.
+    // This happens because every single handle resulting from these computations will not be ACL-persistent.
+    // On the coprocessor side, any computation branch leading to a non-ACL-persistent handle (using allow/allowForDecryption)
+    // will ultimately be automatically discarded.
+    function test_BlockingAccountDoesNotPreventAddEventEmission() public {
         bytes32 lhs = caller.trivialEncrypt(7);
         bytes32 rhs = caller.trivialEncrypt(11);
         caller.addWithoutAllowing(lhs, rhs);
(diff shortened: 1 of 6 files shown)


host-contracts/test/acl/TestIntegrationACL.t.sol
```diff
@@ -0,0 +1,99 @@
...
+        CoprocessorConfig memory config = CoprocessorConfig({
+            ACLAddress: aclAddress,
+            CoprocessorAddress: fhevmExecutorAdd,
+            DecryptionOracleAddress: address(0),
+            KMSVerifierAddress: kmsVerifierAdd
+        });
+        FHE.setCoprocessor(config);
+    }
+
+    function encrypt(uint64 value) external returns (bytes32 handle) {
```
0xalexbel: Better to rename as `trivialEncrypt` since it not an actual encryption but rather a "Trivial" encryption, meaning there's no encryption at all.

host-contracts/test/acl/TestIntegrationACL.t.sol
```diff
@@ -0,0 +1,99 @@
...
+        euint64 encrypted = FHE.asEuint64(value);
+        FHE.allowThis(encrypted);
+        handle = euint64.unwrap(encrypted);
+    }
+
+    function add(bytes32 lhs, bytes32 rhs) external returns (bytes32 handle) {
```
0xalexbel, line 108: Should also add an additional similar test that does not call `FHE.allowThis(result)`. This alternative test should not revert as FHE.add call is not reverting if account is denied

jatZama:
As said yesterday evening to @0xalexbel the commit message is a bit confusing, because technically the denied contract can still produce computations but only given transient allowances to those new handles by fhevmExecutor privileged contract, so this should be unusable for later computation because there isn't any possible path leading to persistant allow or allowFroDecrypt (since we block all allow/transientAllow/allowForDecrypt initiated by the dapp on those new handles) and could be pruned by coprocessor node to ignore them, which was the case initially, but due to ReRand feature it is not pruned anymore, but @antoniupop is aware of this and will fix it in a future release.

host-contracts/test/acl/TestIntegrationACL.t.sol
```diff
@@ -0,0 +1,115 @@
...
+
+    function test_BlockingAccountDoesNotPreventAddWithoutAllowing() public {
+        bytes32 lhs = caller.trivialEncrypt(7);
+        bytes32 rhs = caller.trivialEncrypt(11);
+        caller.addWithoutAllowing(lhs, rhs);
+
+        vm.prank(OWNER);
+        acl.blockAccount(address(caller));
+
+        caller.addWithoutAllowing(lhs, rhs);
```
0xalexbel, line 87: May be you could add a comment following @jatZama remark: 
- A denied account can still call the FHEVM Executor and emit Computation events.
- However, these events are most likely ignored on the coprocessor side. This happens because every single handle resulting from these computations will not be ACL-persistent. On the coprocessor side, any computation branch leading to a non-ACL-persistent handle (using allow/allowForDecryption) will ultimately be automatically pruned (discarded).

obatirou: Sure, will add that
Might need to rename the PR and tests also

obatirou: Renamed the PR, the test and added a check for the event emission
Let me know if good for you

0xalexbel (approved):
LGTM!

PR #1140, opened 2025-10-24
feat(coprocessor): add metrics for gw-listener

dartdart26:
Additionally:
 * use the metrics server from fhevm-engine-common
 * remove the `health_check_port` alias and use it as the name of the
   config option (as we now only have health check on that port, without
   metrics) in both txn-sender and gw-listener
 * use 2 spawns instead of tokio::join! in gw-listener such that the 2
   tasks can run in parallel

### DIFF
--- a/coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
+++ b/coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 use alloy::providers::{ProviderBuilder, WsConnect};
 use alloy::{primitives::Address, transports::http::reqwest::Url};
 use clap::Parser;
-use fhevm_engine_common::telemetry;
+use fhevm_engine_common::{metrics_server, telemetry};
 use gw_listener::aws_s3::AwsS3Client;
 use gw_listener::chain_id_from_env;
 use gw_listener::gw_listener::GatewayListener;
@@ -41,9 +41,12 @@ struct Conf {
     #[arg(long, default_value = "10")]
     error_sleep_max_secs: u16,
 
-    /// HTTP server port for health checks
     #[arg(long, default_value_t = 8080)]
     health_check_port: u16,
+
+    /// Prometheus metrics server address
+    #[arg(long, default_value = "0.0.0.0:9100")]
+    metrics_addr: Option<String>,
 
     #[arg(long, default_value = "4s", value_parser = parse_duration)]
     health_check_timeout: Duration,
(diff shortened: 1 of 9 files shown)


dartdart26:
To be merged **after** https://github.com/zama-ai/fhevm/pull/1100

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
```diff
@@ -41,9 +41,9 @@ struct Conf {
     #[arg(long, default_value = "10")]
     error_sleep_max_secs: u16,
 
-    /// HTTP server port for health checks
-    #[arg(long, default_value_t = 8080)]
-    health_check_port: u16,
+    /// HTTP server port
+    #[arg(long, alias = "health-check-port", default_value_t = 8080)]
+    http_server_port: u16,
```
rudy-6-4: need to change the charts directory and maybe the doc

dartdart26: Yep.

@goshawk-3 is this related to PR 1100 in any way?

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
```diff
@@ -216,29 +226,52 @@ impl<P: Provider<Ethereum> + Clone + 'static, A: AwsS3Interface + Clone + 'stati
...
+                                            activate_key_success += 1;
+                                            info!("ActivateKey event successful");
+                                        }
                                         Err(e) if e.is::<DigestMismatchError>() => {
+                                            key_digest_mismatch += 1;
                                             error!(error = %e, "Key digest mismatch, ignoring event");
                                         }
-                                        Err(e) => return Err(e),
+                                        Err(e) => {
+                                            ACTIVATE_KEY_FAIL_COUNTER.inc();
```
rudy-6-4, line 272: why not a counter here like other metrics ?
Or since they are rare all could be direct increase of the metric.

rudy-6-4 (approved):

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
dartdart26: It's' just because we return from the whole function on an error and try again.

Alternatively, we could try this individual activate event until it succeeds (or fails with bad digest).

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
dartdart26: I will do that in a separate PR @rudy-6-4, also for other services that are missing.

coprocessor/fhevm-engine/gw-listener/src/http_server.rs
```diff
@@ -73,6 +73,7 @@ impl<
         let app = Router::new()
             .route("/healthz", get(health_handler))
             .route("/liveness", get(liveness_handler))
+            .route("/metrics", get(metrics_handler))
```
goshawk-3: Should we delete this endpoint?

goshawk-3 (approved):
LGTM

coprocessor/fhevm-engine/gw-listener/src/http_server.rs
dartdart26: Ah, yes, good catch, let me do it!

PR #1143, opened 2025-10-24
chore(kms-connector): sync with main

eudelins-zama:
Should include roughly:
* feat(kms-connector): shared otlp context
* fix(kms-connector): default otlp_context value
* chore(kms-connector): remove s3 fallback from config

e2e tests: https://github.com/zama-ai/fhevm/actions/runs/18837608423


dartdart26 (approved):
I haven't checked the code in detail, but if it is just a cherry pick - looks ok.

PR #1144, opened 2025-10-24
feat(gateway-contracts): implement exceptional keygen corrective flow

isaacdecoded:
Closes https://github.com/zama-ai/fhevm-internal/issues/505


gateway-contracts/contracts/KMSGeneration.sol
```diff
@@ -220,9 +240,11 @@ contract KMSGeneration is
...
         $.keygenIdPairs[keyId] = prepKeygenId;
 
-        // TODO: Get the epochId once resharing is implemented.
-        // See https://github.com/zama-ai/fhevm-internal/issues/151
-        uint256 epochId = 0;
+        // Generate a globally unique epochId for the resharing epoch.
+        // The counter is initialized at deployment such that epochId's first byte uniquely
+        // represents a resharing epoch, with format: [0000 0110 | counter_1..31]
+        $.epochCounter++;
+        uint256 epochId = $.epochCounter;
```
isaacdecoded: Now that we're introducing the [EPOCH_BASE_COUNTER](https://github.com/zama-ai/fhevm/pull/1144/files#diff-bacd4fa5e85fd1fceca50a8270e67f85715ed66130273b427cfa3d02ebb30f9aR41), do you think it's ok to use the epochCounter at this point already in the key generation flow? cc @RomanBredehoft @dd23

gateway-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -142,6 +158,19 @@ interface IKMSGeneration {
...
+     * @dev This is a temporary method to initialize PRSS until implementation of a proper key resharing.
+     */
+    function prssInit() external;
+
+    /**
+     * @notice Trigger the reshare of the currently active key.
+     * @dev This is a temporary method to reshare the specified key ID until implementation of a proper key resharing.
+     * @param keyId The ID of the key to reshare.
+     */
+    function retryKeygenReshare(uint256 keyId) external;
```
isaacdecoded: Do you think this other suggested name (`retryKeygenReshare`) is ok or better to keep the original `fixedKeygen` one? cc @leventdem @dd23

gateway-contracts/contracts/KMSGeneration.sol
dd23: I think we should not increase it in keygen.
We can in principle have multiple keys with the same set of operators (=same epoch - this is just not implemented yet).
cc @jot2re

gateway-contracts/contracts/interfaces/IKMSGeneration.sol
dd23: maybe `refreshKeygenReshare`?

gateway-contracts/contracts/KMSGeneration.sol
jot2re: Yes, epoch is _completely_ independent of key gen. It only relates to resharing (and hence indirectly to context)

gateway-contracts/contracts/mocks/KMSGenerationMock.sol
```diff
@@ -27,14 +27,20 @@ contract KMSGenerationMock {
...
     event ActivateCrs(uint256 crsId, string[] kmsNodeStorageUrls, bytes crsDigest);
 
+    event PRSSInit();
+
+    event RefreshKeygenReshare(uint256 prepKeygenId, uint256 keyId, uint256 epochId, ParamsType paramsType);
+
     uint256 prepKeygenCounter = 3 << 248;
     uint256 keyCounter = 4 << 248;
     uint256 crsCounter = 5 << 248;
+    uint256 epochCounter = 0;
```
eudelins-zama: Why this value is 0 instead of `6 << 248`?

Looking at `EPOCH_COUNTER_BASE` value, I guess it should be the same here right?

isaacdecoded: Yes, you're right. Initially, I didn't fully implement it, thinking that it should be tackled when properly implementing the key resharing (along with the KMS contexts). But I checked again, and I think it's ok to include already at least for the mocks. Commit [here](https://github.com/zama-ai/fhevm/pull/1144/commits/36bc48f5df99ad90545da262a7a93e2cae9dc505).

gateway-contracts/contracts/mocks/KMSGenerationMock.sol
```diff
@@ -27,14 +27,20 @@ contract KMSGenerationMock {
...
     uint256 keyCounter = 4 << 248;
     uint256 crsCounter = 5 << 248;
+    uint256 epochCounter = 6 << 248;
 
     function keygen(ParamsType paramsType) external {
         prepKeygenCounter++;
         uint256 prepKeygenId = prepKeygenCounter;
-        uint256 epochId;
+        epochCounter++;
+        uint256 epochId = epochCounter;
```
dd23: I think we said _not_ to increase the epoch Counter on keygen?

isaacdecoded: This is a mocked contract generated automatically by executing a script for this, and it is used only for external testing purposes.

eudelins-zama: Isn't that weird that the script increment the `epochId` whereas the "real" `keygen` function does not?

Anyway, not important for this feature IMO

isaacdecoded: This is because [the script](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/scripts/mock_contracts_cli.js#L432) detects all counters declared in the contract based on [this mapping](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/scripts/mock_contracts_cli.js#L23-L29). So it's associating the `epochId` with the `epochCounter` and executing the increment and then the assignment. I can try to find a way to avoid it for the `keygen` method, but I think it's not relevant now, since this PR is just a hotfix that should be properly reworked. Also, this is a mock contract, nothing to do with real/production logic.

eudelins-zama: Definitely not relevant, let's not wait time on this :muscle:

gateway-contracts/contracts/shared/KMSRequestCounters.sol
```diff
@@ -12,7 +12,8 @@ enum RequestType {
     UserDecrypt, // 2
     PrepKeygen, // 3
     Keygen, // 4
-    Crsgen // 5
+    Crsgen, // 5
+    Epoch // 6
```
melanciani: I think internally we should name it `KeyReshare` instead of the very general term `epoch` (it's not very explicit when you compare to the other IDs) !

imo it looks like `epoch` will be mostly used internally to KMS, same like they use `requestId` instead of our different `decryptionId`, `crsId` ect

gateway-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -73,6 +73,22 @@ interface IKMSGeneration {
...
+
+    /**
+     * @notice Emitted to trigger the reshare of the specified key ID.
+     * @dev This is a temporary event to reshare the specified key ID until implementation of a proper key resharing.
+     * @param prepKeygenId The ID of the preprocessing keygen request.
+     * @param keyId The ID of the key to reshare.
+     * @param epochId The ID of the resharing epoch.
+     * @param paramsType The type of FHE parameters to use.
+     */
+    event RefreshKeygenReshare(uint256 prepKeygenId, uint256 keyId, uint256 epochId, ParamsType paramsType);
```
melanciani: I'm not sure I get the `refresh` part of the name (same for the event), it sounds weird to have both refresh + reshare, I thought we suggested something like `keyReshareSameSet` !

isaacdecoded: I initially suggested `retryKeygenReshare` but [received another proposal](https://github.com/zama-ai/fhevm/pull/1144#discussion_r2460900017). I'm going to change it to `keyReshareSameSet` since it seems everybody likes this original suggestion then.

gateway-contracts/contracts/shared/KMSRequestCounters.sol
isaacdecoded: Ok, agree. I'm going to replace the epoch mention with KeyReshare/keyReshareId.

gateway-contracts/contracts/interfaces/IKMSGeneration.sol
eudelins-zama: Ping me once done :pray: @isaacdecoded

melanciani (approved):
great thanks a lot !

PR #1146, opened 2025-10-24
test(host-contracts): refactor ACL tests to use deployer utils

obatirou:
Closes https://github.com/zama-ai/fhevm-internal/issues/514

  - Reuse HostContractsDeployerTestUtils inside ACLTest so the suite exercises the real proxy stack and drops the bespoke mock upgrade flow.
  - Replace storage peeking via MockACL with a documented stdstore helper that rebuilds `UserDecryptionDelegation` from packed slots.
  - Tighten read-only assertions to view and fix setup

### DIFF
--- a/host-contracts/test/acl/acl.t.sol
+++ b/host-contracts/test/acl/acl.t.sol
@@ -1,36 +1,26 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.24;
 
-import {Test} from "forge-std/Test.sol";
 import {UnsafeUpgrades} from "@openzeppelin/foundry-upgrades/src/Upgrades.sol";
+import {StdStorage, stdStorage} from "forge-std/StdStorage.sol";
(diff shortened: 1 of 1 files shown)


host-contracts/test/acl/acl.t.sol
```diff
@@ -47,45 +37,53 @@ contract ACLTest is Test {
...
-        fhevmExecutor = acl.getFHEVMExecutorAddress();
+    /**
+     * @dev Reconstructs `UserDecryptionDelegation` directly from storage.
+     *
+     * `ACL.userDecryptionDelegations` stores three packed `uint64` fields in one slot. We point `stdstore`
+     * at `getUserDecryptionDelegationExpirationDate`, which walks the same nested mapping and records the
+     * slot that is read. With `enable_packed_slots()` the helper returns offsets for that packed word, so we
+     * can `vm.load` the slot and shift/mask the raw value back into the struct fields—no mock accessor needed.
+     */
+    function _getUserDecryptionDelegation(
```
obatirou, line 48: Would a getter on ACL be relevant for app/user usage?

0xalexbel (approved):
LGTM!

PR #1154, opened 2025-10-27
chore(library-solidity): integrate foundry

obatirou:


library-solidity/hardhat.config.ts
```diff
@@ -1,3 +1,4 @@
+require("@nomicfoundation/hardhat-foundry");
```
0xalexbel: You should use import instead of require.

0xalexbel: not sure hardhat foundry is required here. Since we should not modify any host-contract nor lib

0xalexbel: it might also produce side effects since hardhat will try to interpret mappings.txt. Better to handle remapping manually in hardhat using built-in task override. But I can handle that later. We should try to add foundry without possible interference with hardhat

obatirou: you're right, I removed it

0xalexbel (approved):

PR #1160, opened 2025-10-28
feat(protocol-contracts): add admin safe module

jatZama:


protocol-contracts/safe/hardhat.config.ts
```diff
@@ -0,0 +1,29 @@
...
+const config: HardhatUserConfig = {
+  solidity: {
+    version: "0.8.24",
+    settings: {
+      metadata: {
+        // Not including the metadata hash
+        // https://github.com/paulrberg/hardhat-template/issues/31
+        bytecodeHash: "none",
+      },
+      // Disable the optimizer when debugging
```
melanciani: this comment is wrong no ?

protocol-contracts/safe/contracts/AdminModule.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: LGPL-3.0
+pragma solidity ^0.8.0;
+import "@safe-global/safe-contracts/contracts/common/Enum.sol";
+import "@safe-global/safe-contracts/contracts/SafeL2.sol";
+
+contract AdminModule {
```
melanciani, line 6: would be great to add a bit of doc comments to add context on this contract

protocol-contracts/safe/contracts/AdminModule.sol
```diff
@@ -0,0 +1,34 @@
...
+    address immutable public ADMIN_ACCOUNT;
+    address payable immutable public SAFE_PROXY;
```
melanciani, line 8: these couldn't be made `constant` ? like we do with contract addresses in the gateway for ex

melanciani (approved):
minor comments and a question, but overall looks good !

protocol-contracts/safe/contracts/AdminModule.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: LGPL-3.0
```
zmalatrax: License required by Safe?

melanciani: ah good catch: this should definitively be `BSD-3-Clause-Clear`

protocol-contracts/safe/hardhat.config.ts
zmalatrax: Provided link by hardhat is dead((in next line) but it seems to be a guideline to deactivate the optimizer when debugging ; though it can (should) be enabled otherwise

protocol-contracts/safe/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "safe-admin-module",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
zmalatrax, line 11: Is it under ISC license?

protocol-contracts/safe/contracts/AdminModule.sol
jatZama: I will let other people do it, this is also needed for PauserSetWrapper, FeeBurner etc Simon is already aware of this need.

protocol-contracts/safe/contracts/AdminModule.sol
jatZama: This is not easily doable here, we need to pass it in the constructor here for simplicity, so it should be immutable. Gas price at usage is the same because immutable is, like constant, putting it in bytecode.

protocol-contracts/safe/contracts/AdminModule.sol
jatZama: fixed

protocol-contracts/safe/package.json
jatZama: We use ISC everywhere for package.json see for instance in gateway-contracts as well https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/package.json#L16 cc @RomanBredehoft 
I think I will keep it here.

protocol-contracts/safe/hardhat.config.ts
jatZama: fixed

protocol-contracts/safe/package.json
zmalatrax: Okay, good to know, ty

zmalatrax (approved):

PR #1163, opened 2025-10-28
chore(gateway-contracts): implement hardhat tasks for key resharing methods

isaacdecoded:
Refs https://github.com/zama-ai/fhevm-internal/issues/505


melanciani (approved):
looks good, would be great to also have tests for them in https://github.com/zama-ai/fhevm/tree/main/gateway-contracts/test/tasks !

PR #1168, opened 2025-10-28
feat(library-solidity): remove requestDecryption completely

immortal-tofu:


jatZama:
Why are we doing this now? It will break OZ library and so many examples? Better to keep it even if not usable at first. 
Furthermore, why don't we consider this function like the delegateUserDecryption feature: we expose it even if it is not fully ready, so this should be the same for async public decryption.

host-contracts/test/kmsVerifier/kmsVerifier.ts
```diff
@@ -1,111 +1,108 @@
...
-    let setSigners = await kmsVerifier.getKmsSigners();
-    const randomAccount = this.signers.carol;
-    setSigners = [...setSigners, randomAccount];
-    await expect(kmsVerifier.connect(randomAccount).defineNewContext(setSigners, 2)).to.be.revertedWithCustomError(
-      kmsVerifier,
-      'NotHostOwner',
-    );
-  });
-});
+// import { expect } from 'chai';
```
jatZama, line 1: Why are we commenting this now? It can be risky to remove so many tests

jatZama:
Fyi it was agreed offline at the office to NOT remove the Async Decrypt tests and intpuverifier, kmsverifier tests etc but keep all tests, just by rewriting them by using verifySignatures function inside the callback. cc: @0xalexbel

immortal-tofu:
Replaced by https://github.com/zama-ai/fhevm/pull/1229

PR #1171, opened 2025-10-28
feat(gateway-contracts): add monitoring events

melanciani:
we now use the following pattern: 
- new "monitoring" events: same name as function
- consensus events: '{functionName}Consensus'

exceptions: 
- public decryption response
- input proof responses

which use pattern: 
- new "monitoring" events: '{functionName}Call'
- consensus events: same name as function (like currently)

they will need to be updated to follow the other pattern once we tackle : 
- https://github.com/zama-ai/fhevm-internal/issues/295
-  https://github.com/zama-ai/fhevm-internal/issues/533

closes https://github.com/zama-ai/fhevm-internal/issues/527
closes https://github.com/zama-ai/fhevm-internal/issues/446


jatZama (approved):

isaacdecoded (approved):
LGTM! 💯

PR #1172, opened 2025-10-28
fix(protocol-contracts): fix hardhat lz dependencies to ethers v6

zmalatrax:
Fixes the interface error to query the number of decimals of the OFT in `FeesSenderToBurner.sol`
Fixes the package conflicts in feesBurner:
- reduces hardhat plugins to the hardhat-toolbox, which encapsulates all the plugins that were previously added as a direct dependency
- overrides hardhat-deploy to 1.0.4, lz uses 0.12.4 which still rely on the deprecated Etherscan v1 API.
- overrides ethers version for a specific package used by layerzerolabs that created conflicts.
Through trial and error it boiled down to this single package needed to be overriden: "@eth-optimism/contracts@0.6.0"

We cannot simply override all packages to use ethers v6, as layerzerolabs packages expect ethers v5, with breakings changes between both versions, which induce runtime errors when trying to load modules...

### DIFF
--- a/protocol-contracts/feesBurner/contracts/FeesSenderToBurner.sol
+++ b/protocol-contracts/feesBurner/contracts/FeesSenderToBurner.sol
@@ -2,7 +2,6 @@
 pragma solidity ^0.8.22;
 
 import { IOFT, SendParam, MessagingFee } from "@layerzerolabs/oft-evm/contracts/interfaces/IOFT.sol";
-import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import { OptionsBuilder } from "@layerzerolabs/oapp-evm/contracts/oapp/libs/OptionsBuilder.sol";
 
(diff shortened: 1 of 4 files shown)


protocol-contracts/feesBurner/contracts/FeesSenderToBurner.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity ^0.8.22;
 
 import { IOFT, SendParam, MessagingFee } from "@layerzerolabs/oft-evm/contracts/interfaces/IOFT.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
jatZama: Can you remove the IERC20 import please? and replace it in code to ERC20.

protocol-contracts/feesBurner/contracts/FeesSenderToBurner.sol
```diff
@@ -37,7 +38,7 @@ contract FeesSenderToBurner {
             revert UnsupportedChainID();
         }
         uint8 sharedDecimals = IOFT(_oft).sharedDecimals();
-        uint8 decimals = IOFT(_oft).decimals();
+        uint8 decimals = ERC20(_oft).decimals();
```
jatZama, line 40: I just realized it would make more sense to put decimalConversionRate in capital letters, even if it is private, since it is immutable.

jatZama (approved):

protocol-contracts/feesBurner/contracts/FeesSenderToBurner.sol
zmalatrax: renamed to `_DECIMAL_CONVERSION_RATE`

PR #1173, opened 2025-10-28
Update benchmark_parser.py

0xLayer-3:
Changelog

Fix: correct multiplier extraction for ...chunk benchmarks

What changed

Replaced strip("chunk") with removesuffix("chunk") when parsing the throughput multiplier from test names that end with "chunk".

Before

multiplier = int(lowercase_test_name.strip("chunk").split("::")[-1])

After

name_wo_suffix = lowercase_test_name.removesuffix("chunk")  # py3.9+ multiplier = int(name_wo_suffix.split("::")[-1])

Why

str.strip("chunk") removes any of the characters c/h/u/n/k from both ends, not the literal suffix "chunk". This could corrupt names (e.g., trimming valid digits/letters) and produce wrong multipliers.

removesuffix("chunk") removes the exact suffix only, yielding correct parsing.

Impact

Accurate throughput multiplier parsing for pbs_throughput...chunk tests.

Prevents silent mis-scaling of ops-per-dollar and related throughput metrics. Notes

The only functional change is the suffix parsing fix (uses removesuffix, requires Python 3.9+). If Python 3.8 support is needed, replace with:

name_wo_suffix = (
    lowercase_test_name[:-5]
    if lowercase_test_name.endswith("chunk")
    else lowercase_test_name
)

### DIFF
--- a/ci/benchmark_parser.py
+++ b/ci/benchmark_parser.py
@@ -18,9 +18,10 @@ ONE_SECOND_IN_NANOSECONDS = 1e9
 parser = argparse.ArgumentParser()
 parser.add_argument(
     "results",
-    help="Location of criterion benchmark results directory."
-    "If the --key-size option is used, then the value would have to point to"
-    "a CSV file.",
+    help=(
+        "Location of the criterion results directory. "
+        "If --object-sizes or --key-gen is used, this should point to a CSV file."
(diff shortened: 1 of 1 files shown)


aquint-zama:
@cla-bot check

PR #1182, opened 2025-10-29
feat(protocol-contracts): first version of governance OApp

jatZama:


protocol-contracts/governance/contracts/shared/Structs.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.22;
```
melanciani: btw we use `^0.8.24` in host and gateway contracts , we should probably do the same in protocol contracts

protocol-contracts/governance/contracts/GovernanceOAppReceiver.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.22;
```
melanciani: same

protocol-contracts/governance/contracts/GovernanceOAppReceiver.sol
```diff
@@ -0,0 +1,87 @@
...
+
+        uint256 targetLen = targets.length;
+        for (uint256 idx = 0; idx < targetLen; idx++) {
+            /// @dev if function signature is an empty string, datas is the full calldata already starting with the selector
+            datas[idx] = bytes(functionSignatures[idx]).length == 0
+                ? datas[idx]
+                : abi.encodePacked(bytes4(keccak256(bytes(functionSignatures[idx]))), datas[idx]);
+        }
+
+        adminSafeModule.executeSafeTransaction(targets, values, datas, operations);
```
melanciani: what is the benefit between calling adminSafeModule once outside of the loop vs several times within the above loop ? 

also, not sure I get why there would be several txs at once ? what's the use case ?

protocol-contracts/safe/contracts/AdminModule.sol
```diff
@@ -1,34 +1,61 @@
...
+        ADMIN_ACCOUNT = adminAccount;
         SAFE_PROXY = safeProxy;
     }
 
-    function execTransactionFromModuleReturnData(
-        address target,
-        uint256 value,
-        bytes memory data,
-        Enum.Operation operation
+    function executeSafeTransaction(
```
melanciani: why the name change ? if the only difference is being able to run multiple txs, then maybe the name should highlight that

protocol-contracts/governance/contracts/shared/Structs.sol
jatZama: Fixed (fyi this was the LZ default from their CLI, I forgot to change it, and @zmalatrax already noticed that)

protocol-contracts/safe/contracts/AdminModule.sol
jatZama: It is more clear than previous name. Otherwise before it was repeating the function from the Safe Proxy, so there was confusion possible.

protocol-contracts/governance/contracts/GovernanceOAppReceiver.sol
jatZama: The primary use case is to batch "upgradeToAndCall" calls, when we want to upgrade different contracts at once, instead of waiting a very long time (like 30min) on each cross chain tx.
Also, even if it was not strictly necessary, because we could already batch calls via the SafeProxy, by using delegatecall + multisend (like we do already when transferring ownership in a single step to many owners), this design is a lot better for tracability inside the Aragon DAO, because it makes it a lot more auditable, thanks to the array of strings `string[] functionSignatures`, because if we used the multisend+delegatecall trick, the calldata would be raw calldata bytes array with no possibility to add function signatures strings inside the front-end DAO.

protocol-contracts/safe/contracts/AdminModule.sol
melanciani: could it be `executeSafeTransactions` (with an `s` at least) ? to highlight you consider multiple ones here

jatZama: Done.

protocol-contracts/governance/.nvmrc
```diff
@@ -0,0 +1 @@
+v18.18.0
```
zmalatrax: Bump to v22 as in token?

jatZama: done.

melanciani (approved):

protocol-contracts/governance/contracts/shared/Structs.sol
zmalatrax: The `Structs.sol` contract might have been forgotten while bumping pragma version to `^0.8.24` (even it will be compiled with solc 0.8.24 due to the `^`)

jatZama: fixed

PR #1183, opened 2025-10-30
feat(protocol-contracts): add mainnet network

zmalatrax:
Add Ethereum and Gateway mainnet networks to FeesBurner and Token hardhat projects:
- Updates the .env.example
- Add a task to verify `ProtocolFeesBurner` using `hardhat-verify` as `etherscan-verify` is deprecated.
- Gateway Mainnet is commented out as the EndpointId is not available yet.
- Add NPM scripts for token verification

~~FeesBurner known issue:~~
The Etherscan API v1 is deprecated since 05/31/2025, ~~there are issues to verify a contract with `hardhat-verify` using Etherscan API v2 for Ethereum and Blockscout for Gateway using the same hardhat config (`hardhat.config.ts`).~~

To use Etherscan API V2 with `hardhat-verify`, the `etherscan.apiKey` field must be a **string**: otherwise even if overwriting the apiUrl (which defaults to deprecated etherscan API V1 url due to chain descriptors), it misses the mandatory url parameter `chainid`.
However, for the Gateway if `etherscan.apiKey` is a string, `hardhat-verify` recognizes it as using Etherscan API V2, and appends the `chainid` parameter, which is unwanted in this case (using BlockScout), resulting in an error.

To fix this, the hardhat tasks for verifying both contracts of the feesBurner override the apiKey if not in the right configuration for the targeted chain.

### DIFF
--- a/protocol-contracts/feesBurner/.env.example
+++ b/protocol-contracts/feesBurner/.env.example
@@ -1,9 +1,16 @@
+# Deployer account
 PRIVATE_KEY=
+# mainnet
+MAINNET_RPC_URL=
+RPC_URL_ZAMA_GATEWAY_MAINNET=
+# testnet
 SEPOLIA_RPC_URL=
 RPC_URL_ZAMA_GATEWAY_TESTNET=
+# API for contract verification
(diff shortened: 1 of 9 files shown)


jatZama (approved):
LGTM

PR #1185, opened 2025-10-30
docs(gateway-contracts): add example usage for safe deployment command

enitrat:
Adds an in-code documentation on how to properly invoke the deployment command, notably on the JSON format expected when feeding the owners addresses, which is not straightforward.


melanciani (approved):

jatZama (approved):
Very good idea.
I think @RomanBredehoft is working on putting similar task in protocol-contracts/safe at the moment.

PR #1186, opened 2025-10-30
chore(protocol-contracts): enforce `pnpm` as package manager

enitrat:
Packages under `protocol-contracts` alternatively use `npm` and `pnpm`. This PR enforces usage of `pnpm` everywhere, which is faster and better to work with.


enitrat:
i'd even go as far as using bun but i'm not sure it'd be compatible everywhere. maybe in the future

protocol-contracts/feesBurner/package.json
```diff
@@ -2,6 +2,10 @@
   "name": "feesBurner",
   "version": "1.0.0",
   "license": "BSD-3-Clause-Clear",
+    "packageManager": "pnpm@10.20.0",
```
zmalatrax: indent typo
```suggestion
  "packageManager": "pnpm@10.20.0",
```

protocol-contracts/governance/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.6.3",
   "private": true,
   "license": "MIT",
+    "packageManager": "pnpm@10.20.0",
```
enitrat:
```suggestion
  "packageManager": "pnpm@10.20.0",
```

jatZama:
I am not sure it makes sense to push this now.
As said on slack, we use npm everywhere in the monorepo already, except for 2 oapps (governance and token) because those were initialized with the lz cli-tool which strongly recommends to use pnpm...

PR #1188, opened 2025-10-30
feat(protocol-contracts): configurable token deployment as a hardhat task

enitrat:
## Rationale

The original shell scripts (`deploy_zama_oft_arbitrum_testnet.sh`, `deploy_zama_oft_gateway_testnet.sh`) used inline JS code inside a bashscript that was unreliable (as outlined by existing problems in scripts) and hot-modifications of the`hardhat.config.ts` config file on disk; on top of being duplicated for each deployment 

## Proposal

Replaced the shell scripts with a unified Hardhat task (`tasks/deployOFTBridge.ts`) that invokes each task successively. The runtime configuration `hre` is shared among tasks and addresses updated in-memory using `hre.config.networks[network].oftAdapter = { tokenAddress }`, and execute all LayerZero tasks programmatically via `hre.run()`, with uses a config-driven approach with `CONFIGS` for each network.
In the future, adding new deployment targets (e.g., mainnet networks) will only require adding a single entry to the `CONFIGS` object and its corresponding verification scripts to `VERIFICATION_SCRIPTS`.

## Other

- Removed config for arbitrum and all mentions from the readme, replacing with instructions for the Gateway Testnet instead.
- Added config for mainnet - when the L0 endpoint will be ready, we can just remove the `@ts-ignore` directive.


protocol-contracts/token/tasks/deployOFTBridge.ts
```diff
@@ -0,0 +1,202 @@
...
+ * 1. Deploy ZamaERC20 on source network
+ * 2. Update hardhat config with deployed token address
+ * 3. Deploy ZamaOFTAdapter on source network
+ * 4. Deploy ZamaOFT on destination network
+ * 5. Wire the contracts together
+ *
+ * Example usage:
+ * npx hardhat deploy:oft-bridge --target "arbitrum_testnet" --verify true
+ */
+task('deploy:oft-bridge', 'Deploy complete OFT bridge setup')
```
jatZama: Can you rename the script to `deploy:token` please?  Also would change the command:
`task('deploy:token', 'Complete setup of ZAMA token and oftAdapter and oft')`

protocol-contracts/token/tasks/deployOFTBridge.ts
```diff
@@ -0,0 +1,202 @@
...
+    .addParam('target', 'Target network', undefined, types.string)
```
jatZama: Not only the target, but also the source chain can change.
For example for mainnet, source chain will be Ethereum-Mainnet, instead of Sepolia, and destination will be Gateway-Mainnet.
So ideally I would like to see a `source` argument as well.

enitrat: this is mis-described/named; the way I see it is that it's more like a "config" param that uses the right pre-defined configuration setup. 
```ts
const CONFIGS: Record<string, DeploymentConfig> = {
    arbitrum_testnet: {
        sourceNetwork: 'ethereum-testnet',
        destNetwork: 'arbitrum-testnet',
        layerzeroConfigPath: 'layerzero.config.arbitrumtestnet.ts',
    },
    gateway_testnet: {
        sourceNetwork: 'ethereum-testnet',
        destNetwork: 'gateway-testnet',
        layerzeroConfigPath: 'layerzero.config.gatewaytestnet.ts',
    },
    // in the future
    gateway_testnet: {
        sourceNetwork: 'ethereum-mainnet',
        destNetwork: 'gateway-mainnet',
        layerzeroConfigPath: 'layerzero.config.gatewaymainnet.ts',
    },
}
```

and you'd invoke it with 

```
 npx hardhat deploy:oft-bridge --config "gateway_mainnet" --verify true
 ```

File: protocol-contracts/token/tasks/deployAndSetupToken.ts [hunk 3]
```diff
@@ -0,0 +1,186 @@
...
+/**
+ * Orchestrates the full OFT bridge deployment:
+ * 1. Deploy ZamaERC20 on source network
+ * 2. Update hardhat config with deployed token address
+ * 3. Deploy ZamaOFTAdapter on source network
+ * 4. Deploy ZamaOFT on destination network
+ * 5. Wire the contracts together
+ *
+ * Example usage:
+ * npx hardhat deploy:token --preset "arbitrum_testnet" --verify true
```
jatZama: can you remove "arbitrum_testnet" mention from this comment as well?

protocol-contracts/token/tasks/deployAndSetupToken.ts
```diff
@@ -0,0 +1,186 @@
...
+    layerzeroConfigPath: string // Path to LayerZero config file
+    verificationScripts?: {
+        // Verification scripts to run for the source and destination networks
+        source: string
+        destination: string
+    }
+}
+
+const CONFIGS: Record<string, DeploymentConfig> = {
+    gateway_testnet: {
```
jatZama: Can you already add a CONFIG for (gateway-)mainnet? even if just as a placeholder for the moment since LZ endpoint is still not deployed on mainnet, just to be future proof.
Btw I believe to reduce confusion it could be best to remove all mentions of gateway in config name as well.
ie we would have only two configs: "mainnet" (meaning gateway-mainnent and ethereum-mainnet) and "testnet" (meaning gateway-testnet+ethereum-testnet). wdyt? This would also imply renaming the layerzero config file btw.

jatZama (changes requested):

jatZama (approved):
LGTM

PR #1189, opened 2025-10-30
Rudy/fix/gw listener sync contract

rudy-6-4:


coprocessor/fhevm-engine/gw-listener/Cargo.toml
```diff
@@ -32,6 +32,7 @@ tower-http = { workspace = true }
 
 url = "2.5.7"
 fhevm-engine-common = { path = "../fhevm-engine-common" }
+fhevm_gateway_bindings = { path = "../../../gateway-contracts/rust_bindings" }
```
eudelins-zama, line 35: I recommend using git dependency instead of path dependency.

It avoids copro CI being unexpectedly broken by gateway-contracts change. See kms-connector `Cargo.toml` for an example

antoniupop (approved):
LGTM - but best get one more review, I'm not familiar with this.

gateway-contracts/Makefile
```diff
@@ -47,11 +47,16 @@ docker-compose-up: docker-compose-down
...
+check-bindings: ensure-addresses update-bindings
+	git diff --exit-code || { \
+		echo "Rust bindings are not up-to-date."; \
+		echo "Please run 'make update-bindings' to update them."; \
+		exit 1; \
+	}
 
 update-bindings: ensure-addresses
 	python3 scripts/bindings_update.py update
+	cd rust_bindings; cargo fmt --all
```
eudelins-zama, line 59: Are we sure this will not be a problem for gateway CI? cc @isaacdecoded 

Seems risky to me and would try to avoid this if possible, but maybe this works...

gateway-contracts/rust_bindings/Cargo.toml
```diff
@@ -6,6 +6,9 @@ edition = "2024"
 license = "BSD-3-Clause-Clear"
 version = "0.1.0-rc14"
 
+[lints.rust]
+unused = "allow"
+
```
eudelins-zama, line 11: Why is that required?

This is not in kms-connector. If you use git deps instead of path, I think this won't be required

gateway-contracts/Makefile
eudelins-zama: I think you are doing this because of `cargo fmt` breaking in your CI right @rudy-6-4 ?

If you use git deps instead of path, note that these changes won't be required.

eudelins-zama (changes requested):
Looks good overall! :+1: 

Just a few comments, mainly around the use of git dependency instead of path dep.

gateway-contracts/Makefile
isaacdecoded: Yes, I think we should avoid this formatting here. Also, why are we changing the `check-bindings` command above, @rudy-6-4? Seems like you're "deprecating" the `python3 scripts/bindings_update.py check` in favor of that git diff command.

gateway-contracts/rust_bindings/Cargo.toml
rudy-6-4: because the bindings have unused elements

rudy-6-4: and we don't want the warnings during copro compilation

rudy-6-4: (not sure why we see them but same as cargo fmt issue I guess)

eudelins-zama: > If you use git deps instead of path, I think this won't be required

:wink:

eudelins-zama (approved):

PR #1192, opened 2025-10-30
feat(protocol-contracts): add unified FeesBurner deployment script

zmalatrax:
Add a single hardhat task that allows to deploy FeesBurner contracts on Ethereum and Gateway:
```bash
npx hardhat deploy:feesBurner --preset <mainnet|testnet> --verify <true|false> 
```
- Choose to either deploy on Ethereum Mainnet & Gateway Mainnet, or Ethereum Testnet & Gateway Testnet with the `preset` param.
- Deploy ProtocolFeesBurner on Ethereum
- Deploy FeesSenderToBurner on Gateway
- Optionally verify both contracts on Ethereum and Gateway

In order to deploy on different networks within a single hardhat task, we need to switch the HardhatRuntimeEnvironment (`hre`), which is done thanks to the `GetHreByNetworkName(networkName: string)` function, copied and flattened from LayerZero's devtools-evm-hardhat package.


protocol-contracts/feesBurner/tasks/deployFeesBurner.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Choose whether to deploy on testnet on mainnet with the `--mainnet true` flag.
+ * Verify both contracts with the `--verify true` flag.
+ */
+task("deploy:feesBurner")
+  .addOptionalParam(
+    "mainnet",
+    "Deploy on Ethereum Mainnet & Gateway Mainnet if true. Otherwise deploy on Ethereum Sepolia & Gateway Testnet (default).",
+    false,
+    types.boolean,
+  )
```
enitrat: for consistency with the token deployment, could we do something like:

```
   .addParam('preset', 'Deployment preset to use', undefined, types.string)
 ```
 (which accepts mainnet | testnet)

File: protocol-contracts/feesBurner/tasks/deployFeesBurner.ts [hunk 2]
```diff
@@ -0,0 +1,87 @@
...
+
+const MAINNET_CONFIG: DeploymentConfig = {
+  ethereumNetwork: "ethereum-mainnet",
+  gatewayNetwork: "gateway-mainnet",
+};
+
+const TESTNET_CONFIG: DeploymentConfig = {
+  ethereumNetwork: "ethereum-testnet",
+  gatewayNetwork: "gateway-testnet",
+};
```
enitrat: single CONFIG constant with 'mainnet' and 'testnet' keys, that you index with `CONFIG[preset]`

protocol-contracts/feesBurner/deploy/FeesSenderToBurner.ts
```diff
@@ -17,7 +18,23 @@ const deploy: DeployFunction = async (hre) => {
...
+
+    const protocolFeesBurnerDeployment: Deployment = JSON.parse(
+      fs.readFileSync(`deployments/${protocolFeesBurnerNetworkname}/ProtocolFeesBurner.json`, "utf8"),
+    );
+    protocolFeesBurner = protocolFeesBurnerDeployment.address;
+  }
+
+  if (!protocolFeesBurner) {
+    throw new Error(`The ProtocolFeesBurner address cannot be empty: ${protocolFeesBurner}`);
+  }
```
enitrat: not a fan of this logic. could the hardhat task set the environment value insted? This is too dependent on target names

protocol-contracts/feesBurner/tasks/deployFeesBurner.ts
```diff
@@ -0,0 +1,87 @@
...
+    });
+
+    const protocolFeesBurnerAddress = (await hre.deployments.get(PROTOCOL_FEES_BURNER)).address;
+
+    // Deploy FeesSenderToBurner on Gateway
+    // Set HardhatRuntimeEnvironment to Gateway <Mainnet|Testnet>
+    hre = await getHreByNetworkName(config.gatewayNetwork);
+    await hre.run("deploy", {
+      tags: FEES_SENDER_TO_BURNER,
+    });
```
enitrat, line 63: right before that you could

```
process.env["PROTOCOL_FEES_BURNER_ADDRESS"] = protocolFeesBurnerAddress
```

jatZama (approved):
LGTM

protocol-contracts/feesBurner/tasks/deployFeesBurner.ts
zmalatrax: indeed, thx!

protocol-contracts/feesBurner/deploy/FeesSenderToBurner.ts
zmalatrax: indeeed, refactored to only work with env variable

enitrat (approved):

PR #1194, opened 2025-10-30
chore(protocol-contracts): add hardhat tasks for AdminModule

melanciani:
edit: what I initially did was not aligned with our goal anymore : we now use 3 tasks for ownership transfer : 
1) add the `n` new owners
2) check registered owners are correct
3) remove deployer from owners and set threshold to `floor(2n/3)+1`

closes https://github.com/zama-ai/fhevm-internal/issues/520


protocol-contracts/safe/tasks/deploy.ts
```diff
@@ -0,0 +1,167 @@
...
+    contract: proxyFactoryName,
+    args: [],
+    log: true,
+    skipIfAlreadyDeployed: false,
+  });
+
+  // Get the environment variables for the Safe setup, using default values if not provided for
+  // the optional parameters
+  const owners = [deployer];
+  const threshold = getRequiredEnvVar("SAFE_THRESHOLD");
```
jatZama: this can be only 1 by construction, since initial owner is the deployer only, so no need to make it flexible with an env variable.

protocol-contracts/safe/tasks/deploy.ts
```diff
@@ -0,0 +1,167 @@
...
+  const to = getRequiredEnvVar("SAFE_TO", ethers.ZeroAddress);
```
jatZama: tbh i think all those env variables are not needed, the script could be simplified by using defaults values everywhere here.

protocol-contracts/safe/tasks/deploy.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  // The staticCall allows to predict the address of the upcoming Safe proxy before it is deployed
+  const safeAddress =
+    await proxyFactoryContract.createProxyWithNonce.staticCall(
+      safeSingleton.address,
+      safeData,
+      0n,
+    );
+
+  if (safeAddress === ethers.ZeroAddress) {
```
jatZama: This if branch is never reachable, considering that safeData is a valid payload by construction.

protocol-contracts/safe/tasks/verify.ts
```diff
@@ -0,0 +1,18 @@
+import { task } from "hardhat/config";
+
+import { getRequiredEnvVar } from "./utils/loadVariables";
+
+task("task:verifyAdminModule")
```
jatZama, line 27: Could you add verify tasks for the SafeProxy, SafeFactory, MultiSend and SafeSingleton as well?

protocol-contracts/safe/tasks/safeOwnershipTransfer.ts
```diff
@@ -0,0 +1,156 @@
+import Safe from "@safe-global/protocol-kit";
+import MultiSendJson from "@safe-global/safe-contracts/build/artifacts/contracts/libraries/MultiSend.sol/MultiSend.json";
+import { task } from "hardhat/config";
+
+import { getRequiredEnvVar } from "./utils/loadVariables";
+
+// Deploy the SafeSmartAccount contract
```
jatZama: this comment seems false, maybe update it?

melanciani: removed

protocol-contracts/safe/tasks/safeOwnershipTransfer.ts
```diff
@@ -0,0 +1,156 @@
...
+task("task:transferSafeOwnershipFromDeployer")
+  .addParam(
+    "newOwners",
+    "Addresses of the new owners of the Safe, comma-separated",
```
jatZama, line 53: maybe you can add a concrete example of the command like Matthieu did in gateway.

protocol-contracts/safe/tasks/safeOwnershipTransfer.ts
```diff
@@ -1,156 +1,192 @@
...
       safeAddress,
-      contractNetworks,
+      network,
+      ethers,
+    );
+
+    // Define the new threshold: it should be ``floor(2n/3) + 1``, where `n` is the number of the
+    // final owners (i.e., without the deployer)
+    const currentOwners = await safeProxy.getOwners();
+    const newThreshold = Math.floor((2 * (currentOwners.length - 1)) / 3) + 1;
```
jatZama: Not sure it is a good idea to hard code this, I would prefer to pass it as an env variable here please. In the notion page, there is still an open question if we want to use 2/3 or not. Btw even this formula is false, correct one would have been Math.ceil((2 * (currentOwners.length)) / 3) ; this is different in case number of owners is exactly a multiple of 3.

melanciani: I hesitated indeed, thought it was clear we wanted 2/3rd + wanted to avoid any misconfigurations issues Will revert this

although i'm surprised by your comment on the formula: we've always considerd floor() + 1 instead of ceil for thresholds in the protocol 🤔

protocol-contracts/safe/test/tasks/safeOwnershipTransfer.ts
```diff
@@ -41,48 +40,56 @@ describe("Ownership transfers", function () {
...
-    expect(new Set(owners)).to.deep.equal(new Set(newOwners));
+    const expectedOwnersAsString = newOwners.join(",");
+    await hre.run("task:checkSafeOwners", {
+      expectedOwners: expectedOwnersAsString,
+    });
 
-    // Check that the threshold has been updated
+    // Check that the threshold is updated to the new threshold: ``floor(2n/3) + 1``, where `n` is the number of the
+    // new owners
+    const newThreshold = Math.floor((2 * newOwners.length) / 3) + 1;
```
jatZama: as said in my previous comment, this formula is not correct, and anyways should be replaced by an env variable

protocol-contracts/safe/tasks/safeOwnershipTransfer.ts
jatZama: I don't think so, if we have 18 owners (13 mpc + 5 copro) then , we should take a threshold of 12, not 13.

jatZama (approved):

PR #1197, opened 2025-10-31
fix(kms-connector): retry on Unavaible error code

eudelins-zama:
A network error is translated into a `Unavailble` code at GRPC level. Thus, we must retry in such case, to avoid removing the request from the DB as it is currently the case


melanciani (approved):

dartdart26 (approved):
Looks good.

PR #1214, opened 2025-11-01
Improve transaction handling 

lucia-w:
When we call await contact.method(), the promise resolves as soon as the transaction is broadcast to the network, not when it's mined and confirmed. improve transaction handling for more reliability.


aquint-zama:
@cla-bot check

PR #1222, opened 2025-11-03
feat(protocol-contracts): pass list of initial receivers in token constructor

jatZama:
Contracts, tests and tasks were updated.
@zmalatrax will add more tests in a future PR, but I will merge this asap for the auditors.

### DIFF
--- a/protocol-contracts/token/test/ZamaERC20.test.ts
+++ b/protocol-contracts/token/test/ZamaERC20.test.ts
@@ -30,7 +30,7 @@ describe('ZamaERC20 - Unit Test', () => {
 
     // beforeEach hook for setup that runs before each test in the block
     beforeEach(async () => {
-        // The INITIAL_SUPPLY_RECEIVER and INITIAL_ADMIN can be different from the deployer.
+        // The INITIAL_RECEIVER_0 and INITIAL_ADMIN can be different from the deployer.
         zamaERC20 = await zamaERC20Factory
             .connect(deployer)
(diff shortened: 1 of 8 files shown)


melanciani (approved):

protocol-contracts/token/test/ZamaERC20.test.ts
```diff
@@ -30,8 +30,10 @@ describe('ZamaERC20 - Unit Test', () => {
 
     // beforeEach hook for setup that runs before each test in the block
     beforeEach(async () => {
-        // The INITIAL_SUPPLY_RECEIVER and INITIAL_ADMIN can be different from the deployer.
-        zamaERC20 = await zamaERC20Factory.connect(deployer).deploy('ZAMAERC20', 'ZAMA', owner.address, admin.address)
+        // The INITIAL_RECEIVER_0 and INITIAL_ADMIN can be different from the deployer.
+        zamaERC20 = await zamaERC20Factory
+            .connect(deployer)
+            .deploy('ZAMAERC20', 'ZAMA', [owner.address], [11_000_000_000n], admin.address)
```
zmalatrax, line 36: we should add a test with multiple receivers and different amounts received ; current setup only has a single receiver.
I can do it if a subsequent PR if you want to merge this first

melanciani: makes sense indeed good catch

jatZama: Yes please don't hesitate to add such tests, thanks!

enitrat (approved):

PR #1224, opened 2025-11-03
chore(gateway-contracts): add check to prevent empty keyDigests in keygenResponse

melanciani:
refs https://github.com/zama-ai/fhevm-internal/issues/538


gateway-contracts/contracts/KMSGeneration.sol
```diff
@@ -107,14 +107,14 @@ contract KMSGeneration is
      */
     string private constant CONTRACT_NAME = "KMSGeneration";
     uint256 private constant MAJOR_VERSION = 0;
-    uint256 private constant MINOR_VERSION = 2;
+    uint256 private constant MINOR_VERSION = 3;
     uint256 private constant PATCH_VERSION = 0;
```
eudelins-zama, line 111: Just one question:

In the KMS connector, we set the default value for contracts' versions to `1`.
AFAIK, this default value is not overriden by infra config.
And version should be part of the EIP-712 signature, right?

If yes, how is it possible that everything works fine in devnet/testnet? There should be an error with the current setup right, or what am I missing?

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -126,6 +126,7 @@ jobs:
       #  See https://github.com/zama-ai/fhevm-internal/issues/379
       - name: Upgrade GatewayConfig contract
         working-directory: current-fhevm/gateway-contracts
+        if: false
```
eudelins-zama, line 129: Are these `if: false` related to https://github.com/zama-ai/fhevm-internal/issues/379?

If yes, fine for me

eudelins-zama (approved):
LGTM.

Please just check the comments I made before merging :slightly_smiling_face:

gateway-contracts/contracts/KMSGeneration.sol
melanciani: these constant values are mostly here for internal checks, we don't have real rules or use for them (and we sometimes forget to update them) !

> And version should be part of the EIP-712 signature, right?

here you are referring to the eip712 struct version, which is currently always set to 1 indeed : https://github.com/zama-ai/fhevm/pull/1224/files#diff-327512bb29edf1e6b063153d518a4b9f7b3a47baca40205a354ac44d1c7ba8abR194

hence why there is no issue

.github/workflows/gateway-contracts-upgrade-tests.yml
melanciani: yes exactly, the reason why I needed to add false here is because I made the CI target 0.9.5 , and the contracts (except KMSGeneration) haven't changed since then

eudelins-zama (changes requested):
Just removing my approval as the PR subject changed totally

gateway-contracts/contracts/KMSGeneration.sol
eudelins-zama: Ok thanks for the clarification!

eudelins-zama (approved):
LGTM

PR #1227, opened 2025-11-03
refactor(protocol-contracts): remove scaling factor from constructor

zmalatrax:
Removes the 1e18 scaling factor from the ZamaERC20 constructor: the arguments are now expected to have all the decimals ; similar behavior to the ERC20 standard

The environment variable stay in token units, the scaling factor from the token units in .env and the constructor arguments is done with ethers in the deployment script.


enitrat (approved):

jatZama (approved):
LGTM

PR #1230, opened 2025-11-03
chore(gateway-contracts): add domain separators in internal hash functions

melanciani:
this is not a high issue anymore as it was thought out initially in MultichainACL thanks to the chain ID modifier. This is still relevant for future-proof changes 

this will break backward compatibility for allow getters but it is fine as testnet is not ready yet 

refs https://github.com/zama-ai/fhevm-internal/issues/538


gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -25,6 +25,11 @@ contract CiphertextCommits is ICiphertextCommits, UUPSUpgradeableEmptyProxy, Gat
      */
     IKMSGeneration private constant KMS_GENERATION = IKMSGeneration(kmsGenerationAddress);
 
+    /**
+     * @notice The domain separator for the add ciphertext hash.
+     */
+    string private constant ADD_CIPHERTEXT_DOMAIN_SEPARATOR = "CiphertextCommits.addCiphertextMaterial";
```
jatZama: Can you replace the strings by their hash as bytes32? this is helpful to save gas.

jatZama (commented):
I would use bytes32 for domain separators, instead of strings, this is common practice and save gas (for eg EIP712).
Also don't forgot to add separators for delegate/revoke on main ,ie it is not just a cherry pick in this case.

gateway-contracts/contracts/CiphertextCommits.sol
melanciani: done !

jatZama (approved):
LGTM now

PR #1231, opened 2025-11-03
chore(gateway-contracts): add chainId check in userDecryptionRequest

melanciani:
this is more for best practice to prevent issues in the future 

closes https://github.com/zama-ai/fhevm-internal/issues/538


jatZama (approved):
Looks good and quite needed for increased security. 👍

PR #1234, opened 2025-11-03
chore(gateway-contracts): add domain separators in internal hash functions

melanciani:
I will align contract version in a following PR


jatZama:
Good but I noticed an optimization, can you swap those two lines to do the check before the second SLOAD: 
```
        mapping(address => bool) storage alreadyRevokedUserDecryptionCoprocessors = $
            .alreadyRevokedUserDecryptionCoprocessors[delegateUserDecryptionHash];

        // Check if the coprocessor has already delegated the user decryption.
        if (alreadyDelegatedUserDecryptionCoprocessors[msg.sender]) {
            revert CoprocessorAlreadyDelegatedUserDecryption(
                chainId,
                delegator,
                delegate,
                contractAddress,
                delegationCounter,
                expirationDate,
                msg.sender
            );
        }
```
This will save gas many times, by avoiding double SLOADs most of the times.

jatZama (approved):

PR #1236, opened 2025-11-03
fix(gateway-contracts): fix InputVerification/CiphertextCommits versions

melanciani:
in https://github.com/zama-ai/fhevm/pull/1230 I incorrectly updated the version of `InputVerification` instead of `CiphertextCommit` : this PR fixes this


jatZama (approved):
Good catch. Approved, but CI is red...?

PR #1241, opened 2025-11-04
chore(kms-connector): don't re-send GRPC requests to kms

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/534
Closes https://github.com/zama-ai/fhevm-internal/issues/526

e2e tests: https://github.com/zama-ai/fhevm/actions/runs/19065216389


dartdart26 (approved):
I didn't follow all the details of the refactor, but was looking specifically at the use of the `already_sent` flag and it looks correct to me.

So now `under_process` means we are waiting for the flow to complete and `already_sent` means just sent to the KMS?

eudelins-zama:
> So now under_process means we are waiting for the flow to complete and already_sent means just sent to the KMS?

@dartdart26 `under_process` is still used as it was, i.e. to "lock" the row in the DB and state that a request is currently processed by a KMS Worker. So the worker (or another worker instance) don't pick the same request twice.

And yes `already_sent` just means the request was already sent to the KMS and we can immediately poll for the result and skip the sending of the request.

PR #1246, opened 2025-11-04
fix(kms-connector): stop gw-listener if filter is dropped

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/540


dartdart26 (approved):
Looks good!

PR #1254, opened 2025-11-04
test(protocol-contracts): add post-deployment e2e tests and gas profiling for L0 messages

enitrat:
- tests for post-deployment contracts (testing aragon dao L1 gov process and l2 governance through message passing), ensures contract are properly owned and wired
- CLI tool to estimate gas of proposals using local forks.

Example usage:

```bash
  bun run scripts/gas-search-proposal.ts \
    --min-gas 50000 \
    --max-gas 500000 \
    --targets '["0xbd9b335a7d927338623b80f102e9b8734895a029"]' \
    --values '[0]' \
    --function-signatures '[""]' \
    --datas '["0x82dc1ec40000000000000000000000000000000000000000000000000000000000000001"]' \
    --operations '[0]'
 ```

### CODE REVIEW

File: protocol-contracts/deployment-cli/deployment-state/deployment-config.yaml [hunk 1]
```diff
@@ -53,7 +53,7 @@ protocol:
     dao_multisig_threshold: "2/3"
   token:
     recipients:
-      - address: "0xd2097BFD29BE52D57Ed56eabfa960d1Bb42D6bF0"
+      - address: "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
```
melanciani, line 56: could we have a comment saying where this address comes from ?

PR #1260, opened 2025-11-05
ci(common): set --locked for cargo command

fegmorte:


rudy-6-4:
I think this is cover (for copro) by the dependency check but 2 precaution are better than 1 if the check is broken (red) or if the image comes from a PR.
https://github.com/zama-ai/fhevm/blob/main/.github/workflows/coprocessor-dependency-analysis.yml#L64

PR #1263, opened 2025-11-05
chore(test-suite): bump kms core to v0.12.4

dd23:
Bumps the KMS core version to v0.12.4 in the test-suite.


eudelins-zama (approved):
Thanks!

PR #1264, opened 2025-11-05
chore(coprocessor): map verify_and_expand panics (if any) to an error

goshawk-3:


goshawk-3:
Rebased in https://github.com/zama-ai/fhevm/pull/1271

PR #1267, opened 2025-11-05
fix(coprocessor): improve handling of corner case errors

### LINKED ISSUE #1216: Worker panic on invalid FHE operation causes DoS vulnerability
## Description
The `tfhe-worker` uses `.expect()` when converting an FHE operation type from the database. If invalid or corrupted data exists in the `computations` table, the worker will **panic and crash** instead of handling the error gracefully.

## Location
- **File**: `coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs`
- **Function**: `query_for_work()`

## Code Reference
```rust
let fhe_op: SupportedFheOperations = w
    .fhe_operation
    .try_into()
    .expect("only valid fhe ops must have been put in db");
```

## Impact
- **Critical**: A single corrupted database entry can crash the entire coprocessor worker, halting all FHE computations
- **DoS Vulnerability**: An attacker who can inject invalid data into the database could repeatedly crash workers
- No error recovery or graceful degradation; the process terminates immediately

## Why It Matters for Security
The coprocessor worker is the backbone of the FHEVM protocol. A crash renders the entire system unavailable, breaking fundamental availability guarantees. In production, this could halt a confidential blockchain or dApp.

## Suggested Fix
Replace `.expect()` with robust error handling:

```rust
let fhe_op: SupportedFheOperations = match w.fhe_operation.try_into() {
    Ok(op) => op,
    Err(e) => {
        error!(target: "tfhe_worker", { error = %e, transaction_id = ?w.transaction_id }, "Invalid FHE operation in database, skipping item");
        // Mark the work item as failed and continue
        continue;
    }
};
```

## Reproduction Steps
1. Insert a row into the `computations` table with an invalid `fhe_operation` value (e.g., 999)
2. The `tfhe-worker` will poll for work and attempt to process this invalid item
3. The worker will panic and terminate

### Context

_No response_

_No response_

### LINKED ISSUE #1217: Worker panic on missing tenant key causes race condition crash
## Description
The worker uses `.expect()` when retrieving tenant encryption keys from the LRU cache. If a key is evicted between the initial check and its use (possible race condition under high load), the worker will **panic and crash**.

## Location
- **File**: `coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs`
- **Function**: `build_transaction_graph_and_execute()`

## Code Reference
```rust
let keys = rk.get(tenant_id).expect("Can't get tenant key from cache");
```

## Impact
- **Critical**: Worker crashes on cache miss, even though this is recoverable
- Race condition more likely under high load with many tenants
- No retry mechanism or cache-miss handling

## Why It Matters for Security
Encryption keys are fundamental to FHE. A simple cache miss should not be fatal. The system should reload keys from persistent storage when not in cache.

## Suggested Fix
Handle cache miss gracefully:

```rust
let keys = match rk.get(tenant_id) {
    Some(k) => k,
    None => {
        warn!(target: "tfhe_worker", { tenant_id = tenant_id }, "Tenant key not in cache, reloading...");
        // Reload key from persistent storage
        return Err("Failed to reload tenant key from database".into());
    }
};
```

## Reproduction Steps
1. Configure worker with small `tenant_key_cache_size`
2. Process computations for more tenants than cache size
3. Under load, a key may be evicted after checked but before used
4. Worker will panic with "Can't get tenant key from cache"

### Context

_No response_

_No response_

### LINKED ISSUE #1218: Unsafe error downcasting causes worker panic on unexpected error types
## Description
The error handling logic uses `.unwrap()` when downcasting a generic error to `FhevmError`. If an error of a different type occurs, the `.unwrap()` will **panic**, crashing the worker.

## Location
- **File**: `coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs`
- **Function**: `upload_transaction_graph_results()`

## Code Reference
```rust
std::mem::swap(
    &mut *err.downcast_mut::<FhevmError>().unwrap(),
    &mut swap_val,
);
```

## Impact
- Worker is fragile and crashes on unexpected error types
- Makes system difficult to debug and maintain
- Future changes introducing new error types could trigger this panic
- Violates principle that error handling should be resilient

## Suggested Fix
Use safe pattern matching:

```rust
if let Some(fhevm_err) = err.downcast_mut::<FhevmError>() {
    std::mem::swap(fhevm_err, &mut swap_val);
} else {
    error!(target: "tfhe_worker", { error = ?err }, "Unexpected error type during result processing");
    // Handle generic error case gracefully
}
```

## Reproduction Steps
1. Modify scheduler/computation logic to return a custom error type (not `FhevmError`)
2. Trigger a computation that results in this new error
3. When `upload_transaction_graph_results()` processes the result, it will try to downcast
4. The `.unwrap()` will fail, causing worker panic

### Context

_No response_

_No response_

antoniupop:


coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -436,7 +449,19 @@ async fn build_transaction_graph_and_execute<'a>(
...
     {
         let mut rk = tenant_key_cache.write().await;
-        let keys = rk.get(tenant_id).expect("Can't get tenant key from cache");
+        let keys = match rk.get(tenant_id) {
+            Some(k) => k,
+            None => {
+                // If the keys can't be located in the cache, skip
+                // executing for this tenant, retry next iteration
+                error!(target: "tfhe_worker", { tenant_id = tenant_id }, "no keys found for tenant");
+                return Err(CoprocessorError::MissingKeys {
```
goshawk-3: nitpick: Can you also call `s_compute.set_status(..)` to add the error in otel

```
s_compute.set_status(Status::Error {
        description: desc.into(),
    });

WORKER_ERRORS_COUNTER.inc();
```

goshawk-3 (approved):
LGTM

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
goshawk-3: The `set_status, counter inc and log entry` probably can be a reusable func instead

PR #1273, opened 2025-11-06
feat(kms-connector): track last block polled

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/544
Closes https://github.com/zama-ai/fhevm-internal/issues/547


kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
```diff
@@ -0,0 +1,31 @@
...
+        'UserDecryptionRequest',
+        'PrepKeygenRequest',
+        'KeygenRequest',
+        'CrsgenRequest',
+        'PrssInit',
+        'KeyReshareSameSet'
+    );
+EXCEPTION
+    WHEN duplicate_object THEN null;
+END $$;
```
dartdart26, line 13: Wondering if it is worth it to put a dependency between code and the DB. That would mean we have to execute migrations on code changes. IMO, it would be better to just use the DB as storage and keep logic/types in code only. Wdyt?

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -0,0 +1,225 @@
...
+    )
+    .execute(db_pool)
+    .await?;
+
+    if query_result.rows_affected() == 1 {
+        info!("Last block polled for {event_type} was successfully updated!");
+    } else {
+        warn!(
+            "Unexpected query result while updating last block for {event_type}: {query_result:?}"
+        );
```
dartdart26: Maybe what can be done instead is to do INSERT and ON CONFLICT DO UPDATE or similar. Ties to my previous comment about putting the types into the DB, which I think is something we can avoid.

dartdart26 (commented):
Looks good!

The only thing I am not convinced about is the types in the DB - see my previous comments.

kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
eudelins-zama: There's already a huge dependency between code and DB as each event kind has its own table. And here we are using the DB as storage of the last block polled for each event kind, so it can be reused at next restart.

So here I am not sure this would be a big concern, but I am open to other solution ofc :slightly_smiling_face:. 
Where would you store the last block polled? In the event's table itself (but then it would add an unnecessary column for 99% of the row as we are only interested in the most recent block)?

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Not sure to understand this one. No matter the type of query (`INSERT` or `UPDATE`), you want to make sure only one row was affected, and display a warning otherwise, no?

kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
dartdart26: What I meant is the `event_type` enum - does it have to be part of the DB definition?

Totally fine with storing the last block ID, just the type is concerning me :)

kms-connector/crates/gw-listener/src/core/publish.rs
dartdart26: I mean that you INSERT all the types in the migration. And you expect that UPDATE would always return 1 affected. But I was saying that if we don't do INSERT into the migration (and don't specify types there), you won't have the issue here with the warn and, also, you can use INSERT or UPDATE in 1 query basically.

kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
eudelins-zama: No it does not have to, we could use an integer as primary key, and use the enum discriminant on the Rust side to insert/select from this table. Wdyt? :slightly_smiling_face: 

Not sure it is much better tbh now that I'm thinking of it, but let me know

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Still not sure to understand sorry :sweat_smile: .
If you don't insert in the migration and we use INSERT or UPDATE in 1 query as you suggest, we would still want to display a warning if no row were affected by the query right?

kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
dartdart26: It could still be a string, but just not having to restrict to a set of strings in the DB as we do with `CREATE TYPE`, that's what I was thinking.

Because the moment you have a new type in code, you need to run a migration to update the enum for it to accept it, right?

kms-connector/crates/gw-listener/src/core/publish.rs
dartdart26: If it is the first time you INSERT it in code (if you don't INSERT in the migration), there won't be a need for the else statement.

kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
eudelins-zama: > Because the moment you have a new type in code, you need to run a migration to update the enum for it to accept it, right?

Yes indeed, but if we want to add a new type in code (new event to handle), we have to run a db-migration anyway to create a new table for this new event type. So we should altering the existing `event_type` enum at the same time.

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: But it even the INSERT statement could fail somehow, so we should display a warning in such case.

kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
dartdart26: OK, let's leave it as is.

kms-connector/crates/gw-listener/src/core/publish.rs
dartdart26: Let's leave as is.

dartdart26 (approved):
Looks good!

PR #1275, opened 2025-11-06
chore(gateway-contracts): remove safe tasks

melanciani:
the multisig contracts are now found in protocol-contracts instead, we no longer need these safe tasks in the gateway-contracts dir  

refs https://github.com/zama-ai/fhevm-internal/issues/520


enitrat (approved):

isaacdecoded (approved):
Great! ✅

PR #1279, opened 2025-11-06
chore(protocol-contracts): add selectors list

enitrat:


isaacdecoded (approved):
Great! 👍 You could also include the `update-selectors` command in a Makefile, for better dev experience, I mean.

enitrat:
ill do a global makefile for all the protocol-contracts

melanciani (approved):
thanks !

PR #1283, opened 2025-11-07
fix(protocol-contracts): ensure `ProtocolStaking` cooldown does not overflow

arr00:


protocol-contracts/staking/contracts/ProtocolStaking.sol
```diff
@@ -315,8 +315,8 @@ contract ProtocolStaking is AccessControlDefaultAdminRulesUpgradeable, ERC20Vote
         return success;
     }
 
-    function _setUnstakeCooldownPeriod(uint256 unstakeCooldownPeriod_) internal {
-        require(unstakeCooldownPeriod_ != 0, InvalidUnstakeCooldownPeriod());
+    function _setUnstakeCooldownPeriod(uint48 unstakeCooldownPeriod_) internal {
+        require(unstakeCooldownPeriod_ != 0 && unstakeCooldownPeriod_ <= 365 days, InvalidUnstakeCooldownPeriod());
```
mortendahl, line 319: What's the rational behind putting this limit of one year?

arr00: Any suggestion on a limit is good. We just wanted to limit to something so that the admin can't increase arbitrarily and lock all tokens.

mortendahl: We think one year is reasonable, were just curious on why this was set.

melanciani (approved):
thanks !

PR #1286, opened 2025-11-07
feat(kms-connector): track last block polled (#1273)

eudelins-zama:
* feat(kms-connector): track last block polled

* fix(kms-connector): update block only if greater than already stored


leventdem (approved):
+1

PR #1292, opened 2025-11-07

eudelins-zama:
As we are doing a comparison on `block_number` in the update query:
```
    "UPDATE last_block_polled SET block_number = $2 \
     WHERE event_type = $1 AND (block_number IS NULL OR block_number < $2)"
```
I realized `BYTEA` was not suitable and now use `NUMERIC` instead.

### DIFF
--- a/kms-connector/crates/gw-listener/src/core/gw_listener.rs
+++ b/kms-connector/crates/gw-listener/src/core/gw_listener.rs
@@ -251,18 +251,13 @@ where
                 .bind(event_type)
                 .fetch_one(&self.db_pool)
                 .await?
-                .try_get::<Option<Vec<u8>>, _>("block_number")?;
+                .try_get::<Option<i64>, _>("block_number")?;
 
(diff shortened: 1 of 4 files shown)


kms-connector/crates/gw-listener/src/core/gw_listener.rs
```diff
@@ -23,7 +24,7 @@ use fhevm_gateway_bindings::{
     decryption::Decryption::{self, DecryptionInstance},
     kms_generation::KMSGeneration::{self, KMSGenerationInstance},
 };
-use sqlx::{Pool, Postgres, Row};
+use sqlx::{Pool, Postgres, Row, types::BigDecimal};
```
dartdart26: Wondering whether using BIGINT (which is i64) wouldn't be simpler?

eudelins-zama: Indeed :pray:

dartdart26 (approved):
Looks good!

PR #1293, opened 2025-11-07

eudelins-zama:
e2e tests: https://github.com/zama-ai/fhevm/actions/runs/19173431171

### DIFF
--- a/kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
+++ b/kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
@@ -15,7 +15,7 @@ END $$;
 
 CREATE TABLE IF NOT EXISTS last_block_polled (
     event_type event_type NOT NULL,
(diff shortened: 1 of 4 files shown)


kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
```diff
@@ -15,7 +15,7 @@ END $$;
 
 CREATE TABLE IF NOT EXISTS last_block_polled (
     event_type event_type NOT NULL,
-    block_number BYTEA,
+    block_number BIGINT,
```
dartdart26, line 18: Putting a note here to remember to check if this table has been deployed somewhere. If yes, above fix won't work as we'd need to ATLER the table instead of changing the existing migration.

dartdart26 (approved):
Looks good! I put a comment about the migration, in case this table has been deployed somewhere. Let's keep that in mind.

PR #1294, opened 2025-11-07

eudelins-zama:


kms-connector/connector-db/migrations/20251104142103_add_last_block_polled.sql
```diff
@@ -15,7 +15,7 @@ END $$;
 
 CREATE TABLE IF NOT EXISTS last_block_polled (
     event_type event_type NOT NULL,
-    block_number BYTEA,
+    block_number BIGINT,
```
dartdart26, line 18: Same as in the other PRs for different branches - this only works if the table doesn't exist in the DB as we'd need to do ALTER to properly upgrade, in a separate migration.

dartdart26 (approved):
Looks good! Put a comment about the migrations, if table is existing (which it isn't anywhere, as far as we know).

PR #1295, opened 2025-11-07
chore(gateway-contracts): add verifyProtocolPayment and update charts

melanciani:
- add missing `verifyProtocolPayment` task
- update `charts/` for new commands for host and gateway contracts

refs https://github.com/zama-ai/fhevm-internal/issues/466


charts/contracts/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: contracts
 description: A helm chart to manage fhevm Smart Contracts Deployment
-version: 0.7.3
+version: 0.7.4
```
isaacdecoded, line 3: Also, not sure we need to bump the Chart version here, but @tawadaa or someone from infra should confirm.

isaacdecoded (approved):

charts/contracts/Chart.yaml
melanciani: we need to : else the CI fails

melanciani: we need to bump anytime we update the values.yaml, even if it's just a comment

jatZama (approved):

charts/contracts/Chart.yaml
tawadaa: yes, once you change the chart, we need to bump the chart version (otherwise the ci won't pass), a minor one is good, since we are only changing values

charts/contracts/values-add-pausers.yaml
```diff
@@ -0,0 +1,130 @@
...
+# =============================================================================
+
+# NOTE: These values are for reference only
+# The chart is for gateway and host contracts
+
+# -----------------------------------------------------------------------------
+# Add pausers to gateway contracts
+# -----------------------------------------------------------------------------
+# Adds pausers to the gateway contracts
+addGatewayPausers:
```
tawadaa: needs to be scDeploy or to be included in the condition we use in the template

charts/contracts/values-add-pausers.yaml
```diff
@@ -0,0 +1,130 @@
...
+  # Commands to run in sequence
+  commands:
+    # Add pausers to the gateway contracts
+    - npx hardhat task:addGatewayPausers
+
+# -----------------------------------------------------------------------------
+# Add pausers to the host contracts
+# -----------------------------------------------------------------------------
+# Adds pausers to the host contracts
+addHostPausers:
```
tawadaa: similar to above

tawadaa (changes requested):
LGTM, only small comment on the values title used

tawadaa (approved):
LGTM as a reference
But we will use scDeploy for all jobs, as stated in my previous comment

PR #1313, opened 2025-11-12
fix(coprocessor): gw-listener, reject suspicious urls for keys

rudy-6-4:


coprocessor/fhevm-engine/gw-listener/src/aws_s3.rs
```diff
@@ -142,6 +142,38 @@ fn bucket_from_domain(url: &Url) -> anyhow::Result<String> {
     Ok(domain_parts[0].to_owned())
 }
 
+fn is_url_suspicious(parsed_url_and_bucket: &Url) -> bool {
```
obatirou, line 145: no unit test?

Else lgtm
My understanding, url is suspicious if there are any of: 
* query
* fragment
* username
* password
* not http or https
* no host

PR #1315, opened 2025-11-12
chore(test-suite): fhevm 0.9.12, kms v0.12.4 version bump

eudelins-zama:


melanciani (approved):
thanks, maybe rename the commit to add `and kms-core to v0.12.4` !

PR #1316, opened 2025-11-12
chore(common): checkSignatures, ZamaConfig.sol, mock publicDecrypt an…

0xalexbel:
- FHE.sol : remove verifySignatures
- FHE.sol : add checkSignatures
- New ZamaConfig.sol 
- Add local coprocessor config
- Add mock publicDecrypt function
- Add publicDecrypt tests
- Add verifySignatures tests
- Refactor the HH mock typescript code: fhevmjsMock.ts, instance.ts etc.
- host-contracts & library-solidity must have the same mock code.


host-contracts/.env.example
```diff
@@ -1,3 +1,4 @@
+ENV_HOST_ADDRESSES_PATH="addresses/.env.host"
```
isaacdecoded, line 1: Since this is an environment variable used only for tests, maybe you could define it in the `hardhat.config.js file, similar to what we do [at the Gateway](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/hardhat.config.ts#L31-L34).

Same [here](https://github.com/zama-ai/fhevm/pull/1316/files#diff-4697d3d7d154579fb47657ead15d6d5cd61ea5f4338c24438762108e8542c95dR1).

jatZama: I agree with Isaac, don't put this in the .env, this was not needed until now, I don't see the point.

host-contracts/eslint.config.mjs
```diff
@@ -0,0 +1,133 @@
+import { FlatCompat } from "@eslint/eslintrc";
```
jatZama, line 1: why are all those eslint files needed now? linter was working before without those.

host-contracts/package.json
```diff
@@ -8,16 +8,21 @@
   },
   "main": "contracts/ACL.sol",
   "scripts": {
-    "compile": "HARDHAT_NETWORK=hardhat npm run deploy:emptyProxies && cross-env TS_NODE_TRANSPILE_ONLY=true hardhat compile",
-    "deploy:emptyProxies": "hardhat task:deployEmptyUUPSProxies &&  hardhat compile:specific --contract 'contracts/immutable' && hardhat task:deployPauserSet",
+    "addresses": "npm run deploy:emptyProxies",
```
isaacdecoded, line 11: nit: having this "addresses" script seems a bit redundant already having "deploy:emptyProxies"

host-contracts/lib/FHE.sol
```diff
@@ -9072,24 +9075,57 @@ library FHE {
...
-     * @notice Warning: this function never reverts, its boolean return value must be checked.
-     * @dev clearTexts is the abi-encoding of the list of all decrypted values assiociated to handlesList, in same order.
-     * @dev Only static native solidity types for clear values are supported, so clearTexts is the concatenation of all clear values appended to 32 bytes.
-     * @dev decryptionProof contains KMS signatures corresponding to clearTexts and associated handlesList, and needed metadata for KMS context.
-     **/
-    function verifySignatures(
+    /// @notice Reverts if the KMS signatures verification against the provided handles and public decryption data
+    ///         fails.
+    /// @dev The function MUST be called inside a public decryption callback function of a dApp contract
+    ///      to verify the signatures and prevent fake decryption results for being submitted.
```
zmalatrax, line 9081:
```suggestion
    ///      to verify the signatures and prevent fake decryption results from being submitted.
```

host-contracts/package.json
```diff
@@ -8,16 +8,21 @@
...
+    "tsc": "tsc --project tsconfig.json --noEmit",
+    "clean": "rm -rf addresses dist build artifacts cache typechain-types",
+    "typechain": "cross-env TS_NODE_TRANSPILE_ONLY=true hardhat typechain",
```
isaacdecoded, line 14: nit: note that the typechain is also generated when running "hardhat compile", so maybe not needed to have a separate script for generating the typechain.

host-contracts/test/encryptedERC20/EncryptedERC20.ts
```diff
@@ -114,7 +114,7 @@ describe('EncryptedERC20', function () {
       );
       expect.fail('Expected an error to be thrown - Bob should not be able to reencrypt Alice balance');
     } catch (error) {
-      expect(error.message).to.equal('User is not authorized to reencrypt this handle!');
+      expect((error as any).message).to.equal('User is not authorized to reencrypt this handle!');
```
isaacdecoded, line 117: Maybe better to use `error as Error` to avoid the usage of `any`.

host-contracts/test/signers.ts
```diff
@@ -8,12 +9,12 @@ import { waitForBalance } from './utils';
 const exec = promisify(oldExec);
 
 export interface Signers {
-  alice: HardhatEthersSigner;
-  bob: HardhatEthersSigner;
-  carol: HardhatEthersSigner;
-  dave: HardhatEthersSigner;
-  eve: HardhatEthersSigner;
-  fred: HardhatEthersSigner;
+  alice: HardhatEthersSigner | HDNodeWallet;
```
jatZama, line 12: Why do you need `HDNodeWallet` here?

host-contracts/package.json
jatZama: I am not sure what's the point indeed of changing those. Also this might break the e2e deployment script made by @enitrat but I am not sure.

host-contracts/test/instance.ts
```diff
@@ -1,295 +1,173 @@
...
- */
-export const decryptEbytes256 = async (handle: string): Promise<bigint> => {
-  if (network.name === 'hardhat') {
-    await awaitCoprocessor();
-    return BigInt(await getClearText(handle));
-  } else {
-    return getDecryptor().decryptEbytes256(await getCiphertext(handle, ethers));
-  }
-};
+export { getTxHCUFromTxReceipt, awaitCoprocessor };
```
isaacdecoded, line 173: I'm wondering what the intention of re-exporting these two functions here is, if they're already exported by `fhevmjsMocked`?

host-contracts/test/tasks/pausing.ts
```diff
@@ -3,18 +3,20 @@ import dotenv from 'dotenv';
 import fs from 'fs';
 import { ethers, run } from 'hardhat';
 
+import { ACL } from '../../typechain-types';
+
 describe('Pausing and Unpausing Tasks', function () {
   let pauserSet;
-  let acl;
+  let acl: ACL;
```
isaacdecoded, line 10: Maybe doesn't make much sense to define its type if you then need to do the type assertion `as ACL` in line 19. In such a case, should the pauserSet also have its type defined to be consistent?

host-contracts/test/fhevmjsMocked.ts
```diff
@@ -1,151 +1,164 @@
...
-  const coprocessorSigners = [];
-  const numKMSSigners = getRequiredEnvVar('NUM_COPROCESSORS');
-  for (let idx = 0; idx < +numKMSSigners; idx++) {
-    const coprocessorSigner = await hre.ethers.getSigner(getRequiredEnvVar(`COPROCESSOR_SIGNER_ADDRESS_${idx}`));
-    await checkIsHardhatSigner(coprocessorSigner);
-    coprocessorSigners.push(coprocessorSigner);
-  }
-  return coprocessorSigners;
+function loadEnvFhevmMockConfig(): EnvFhevmMockConfig {
+  const hostEnvPath = getRequiredEnvVar('ENV_HOST_ADDRESSES_PATH');
```
jatZama, line 42: Do you really need this to be inside a .env?

jatZama: Maybe for getting identical code for fhevmjsMocked between host and lib?

library-solidity/.env.example
```diff
@@ -1,10 +1,12 @@
+ENV_HOST_ADDRESSES_PATH="./fhevmTemp/addresses/.env.host"
 MNEMONIC="adapt mosquito move limb mobile illegal tree voyage juice mosquito burger raise father hope layer"
 CHAIN_ID_GATEWAY="654321"
 DEPLOYER_PRIVATE_KEY="7697c90f7863e6057fbe25674464e14b57f2c670b1a8ee0f60fb87eb9b615c4d" # account[5]
 DECRYPTION_ADDRESS="0x5ffdaAB0373E62E2ea2944776209aEf29E631A64"
 INPUT_VERIFICATION_ADDRESS="0x812b06e1CDCE800494b79fFE4f925A504a9A9810"
 NUM_KMS_NODES="1"
-KMS_THRESHOLD="1"
+# In practice, the `PUBLIC_DECRYPTION_THRESHOLD` is currently set to `floor(n/2) + 1` with `n`` the number of KMS nodes
```
jatZama, line 8: I am not sure about this comment, can you remove it? This changed several times very recently.

jatZama: Ah but this was already inside the host-contracts' .env.
Let's keep it then, even if i am not sure.

host-contracts/test/fhevmjsMocked.ts
melanciani: don't have a lot of context here, but env var should be kept only for infra / production tools as much as possible, if we can avoid them for test-only purposes, then it's much better

library-solidity/config/ZamaConfig.sol
```diff
@@ -25,41 +67,35 @@ library ZamaConfig {
             });
     }
 
-    function getEthereumProtocolId() internal pure returns (uint256) {
-        /// @note Zama Ethereum protocol id is '1'
-        return 1;
+    /// @dev chainid == 11155111
```
jatZama: wrong chain id here

library-solidity/config/ZamaConfig.sol
```diff
@@ -25,41 +67,35 @@ library ZamaConfig {
...
+    function _getLocalProtocolId() private pure returns (uint256) {
+        /// @note Development protocol id is '100000 + Zama Ethereum protocol id'
+        return 100001;
```
jatZama: maybe for local net, send back type(uint256).max , just in case. 
10001 seems too small

library-solidity/.env.example
melanciani: better to remove and replace by 
```
Should be the same as `PUBLIC_DECRYPTION_THRESHOLD` set in the Gateway
```

same for host

PR #1317, opened 2025-11-12
style(gateway-contracts): add cspell and trunk to manage linters

ClementWalter:
# Trunk

This PR adds cspell spell checking, and introduce trunk to manage the linters and formatters.

Since I wanted to avoid big file changes but still wanted to have the main `trunk check` command pass, I have ignored all directories apart from the gateway contracts one. We should enable it progressively on other folders.

[trunk](trunk.io) is a linter and formatter manager, to keep one single command for everything related to these topics. It also manage git hooks and CI (if enabled), among other options and tools.

The tl;dr here

- `trunk check` will check all enable linter (or `trunk check --fix` to auto fix)
- `trunk fmt` will run all formatter
- `git commit` will be hooked to check for commit message correctness
- you can install the Trunk VSCode/Cursor extension and just use `trunk` as default formatter for everything

Next steps:

- fix error raises with `trunk check --all` (all pre existing errors)
- enable more linters
- enable on other directories
- add trunk action in CI

See also

- https://docs.trunk.io/code-quality/code-quality

```bash
curl https://get.trunk.io -fsSL | bash
```

image

image


cspell.yaml
```diff
@@ -0,0 +1,162 @@
+version: "0.2"
+suggestionsTimeout: 5000
+ignorePaths: null
+words:
```
melanciani, line 4: so this means we have a current fixed list of words to manage ourselves ?

enitrat: all non-conventional words that are in the dictionary should be handled through exceptions, yes

.trunk/configs/.rustfmt.toml
```diff
@@ -0,0 +1 @@
+edition = "2021"
```
enitrat: 2024 is out now 🤔

cspell.yaml
ClementWalter: see also integrated quick fix feature in VSCode/Cursor
image

.trunk/configs/.rustfmt.toml
ClementWalter: in any case rustfmt was not used so far, so I've removed the config for now. This PR is just to introduce cspell and trunk to manage all existing linters and formatters, not to add new ones

gateway-contracts/rust_bindings/src/zama_oft.rs
```diff
@@ -379,6 +379,7 @@ pub mod ZamaOFT {
     #[rustfmt::skip]
     #[allow(clippy::all)]
     pub static BYTECODE: alloy_sol_types::private::Bytes = alloy_sol_types::private::Bytes::from_static(
+        // cspell:disable-next-line
```
melanciani: just to be sure: this line is added automaticly right ?

enitrat (dismissed):
lgtm

gateway-contracts/rust_bindings/src/zama_oft.rs
ClementWalter: no this one needs to be added manually. whenever you add words that are not part of the common dict, either you need to add it to the local dict, or to disable the linter for the line, like for any linter I guess

ClementWalter: like when you add `// @ts-ignore` you need to do it manually

melanciani (changes requested):
bindings are auto generated, we should not manually update them

commitlint.config.ts
```diff
@@ -0,0 +1,31 @@
+import type { UserConfig } from "@commitlint/types";
+
+const Configuration: UserConfig = {
```
melanciani, line 3: could you clarify a bit where this is considered ? is it for local commits or in CI ?

ClementWalter: there was already a similar file in the `library-solidity/commitlint.config.ts` though not as accurate as this one, based on @aquint-zama answer on the actual regex enforced for commit messages. This is just used locally in a git hook, mimicking the one actually defined elsewhere

cspell.yaml
melanciani, line 4: I still think adding `cspell` should be a fhevm discussion across teams because there are different ways of approaching this subject. For example tfhe-rs consider a tool (`typos`) that seem to enable low false positives in practice : https://github.com/zama-ai/tfhe-rs/blob/main/_typos.toml

not sure which one is best, I'm just not a huge fan of adding dictionaries that we'll need to update manually more or less often, hence why I think we should discuss this at the fhevm team level

in order not to block this PR, let's keep the fixes but remove cspell itself from the PR for now (+ create an issue not to forget about it)

.trunk/trunk.yaml
```diff
@@ -0,0 +1,124 @@
...
+      ref: v1.7.3
+      uri: https://github.com/trunk-io/plugins
+# Many linters and tools depend on runtimes - configure them here. (https://docs.trunk.io/runtimes)
+runtimes:
+  enabled:
+    - go@1.21.0
+    - node@22.16.0
+    - python@3.10.8
+# This is the section where you manage your linters. (https://docs.trunk.io/check/configuration)
+tools:
```
melanciani, line 19: a bit like cspell, maybe let's discuss this at the fhevm level, although here I'm less sure of what trunk does in practice

commitlint.config.ts
melanciani: I've just tried to create a commit locally with random message and had no errors, is there something else to configure ?

enitrat: you need to install the hook I think

ClementWalter: this is a working commitlint config, there was already a `.commitlintrc.json` file that I've replaced. depending on your current setting and how you use commitlint, you will have it or not. I personally use trunk to manage git hooks as well, but you can just do as you were doing before 🤷

cspell.yaml
ClementWalter: - typos VS cspell : ok for me I don't mind, I can update the PR. But I don't think it's good practice to keep running without anything
- fhevm VS gateway only discussion : as far as I can tell, every team currently manages their own tool and I'm not adding any restriction to any other folder. but if it's not the right time it's ok we can just pause this

.trunk/trunk.yaml
ClementWalter: see below, it's just enabled for the gateway-contract folder, not for the whole repo

ClementWalter: I have actually enabled all the linter that I found already in use

    - hadolint@2.14.0
    - clippy@1.91.0
    - markdownlint@0.45.0
    - prettier@3.6.2
    - cspell@9.3.0
    - solhint@5.1.0

only applied to `./gateway-contracts/**`

commitlint.config.ts
aquint-zama: fyi the commitlint action config is in `.github/config` I'm updating it to get rid of the ruleset in this PR #1366

PR #1329, opened 2025-11-13
chore(kms-connector): limit decryption attempts

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/511

e2e tests: https://github.com/zama-ai/fhevm/actions/runs/19370157895/job/55423698401


kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
```diff
@@ -419,74 +498,85 @@ const RETRYABLE_GRPC_CODE: [Code; 4] = [
...
             Err(e) if RETRYABLE_GRPC_CODE.contains(&e.code()) => {
                 error_counter.inc();
-                warn!("#{i}/{retries} GRPC request attempt failed: {e}");
+                warn!("#{i}/{retries} GRPC request error_count failed: {e}");
             }
             Err(e) => {
                 error_counter.inc();
-                return Err(ProcessingError::Irrecoverable(e.into()));
+                return (i, Err(ProcessingError::Irrecoverable(e.into())));
             }
```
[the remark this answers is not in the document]
eudelins-zama: Hallucination

kms-connector/crates/kms-worker/src/core/kms_worker.rs
```diff
@@ -97,12 +100,13 @@ where
 
         let response = KmsResponse::new(response_kind, otlp_context);
         if let Err(e) = response_publisher.publish_response(response).await {
```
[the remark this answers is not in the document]
eudelins-zama: No we should not, as this is an internal error and not one returned by the KMS

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
```diff
@@ -419,74 +498,85 @@ const RETRYABLE_GRPC_CODE: [Code; 4] = [
...
                 }
             }
         }
     }
     error_counter.inc();
-    Err(Status::unavailable("all result polling attempts failed"))
+    (
+        retries as u16,
+        Err(Status::unavailable("all result polling attempts failed")),
+    )
```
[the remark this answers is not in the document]
eudelins-zama: Hallucination

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
```diff
@@ -419,74 +498,85 @@ const RETRYABLE_GRPC_CODE: [Code; 4] = [
...
     Fut: Future<Output = Result<Response<R>, Status>>,
 {
-    for i in 1..=retries {
+    for i in 1..=retries as i16 {
         match request_fn().await {
             Ok(_) => {
                 success_counter.inc();
                 info!("GRPC request successfully sent to the KMS!");
-                return Ok(());
+                return (i - 1, Ok(())); // Don't count last successful attempt
```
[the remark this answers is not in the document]
eudelins-zama: Already specified in docstring of the fn

kms-connector/connector-db/migrations/20251113084850_add_decryption_error_counter.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE public_decryption_requests ADD COLUMN error_counter SMALLINT DEFAULT 0;
+ALTER TABLE public_decryption_requests ALTER COLUMN error_counter SET NOT NULL;
```
dartdart26: Do we need to have two separate ALTER statements and not just one? Or maybe there's some PSQL limitation here?

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
```diff
@@ -88,8 +90,14 @@ impl KmsClient {
...
     /// Sends the GRPC request to the KMS Core.
+    ///
+    /// Returns the number of errors encountered while sending the request, and the result of the
+    /// request.
     #[tracing::instrument(skip_all)]
-    pub async fn send_request(&self, request: &KmsGrpcRequest) -> Result<(), ProcessingError> {
+    pub async fn send_request(
+        &self,
+        request: &KmsGrpcRequest,
+    ) -> (i16, Result<(), ProcessingError>) {
```
dartdart26, line 100: This return type makes it a bit confusing, IMO. How about returning the error count in both cases, something like `Result<i16, (i16, ProcessingError)>` or maybe wrapping the return `count + ProcessingError` in another struct? Or even adding error count in `ProcessingError`?

dartdart26 (approved):
Looks good! Added some comments. I will let you to decide whether to handle them or not.

kms-connector/connector-db/migrations/20251113084850_add_decryption_error_counter.sql
eudelins-zama: We can do 1 ALTER statements indeed, will update this

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
eudelins-zama: I considered `Result<i16, (i16, ProcessingError)>`, but liked this one better. Only personal preference, and could switch if you think it benefits clarity.

> maybe wrapping the return count + ProcessingError

Could do this yes, it's a bit of boilerplate to be used only once, but it could be prettier

> Or even adding error count in ProcessingError?

`ProcessingError` is used at multiple places, and IMO in some of them we don't care about incrementing returning the error count. We only want to increase it when the KMS returns an error, not for potential internal errors

Let me know what you prefer!

eudelins-zama: Now that I think of it, we also have `(i16, Result<Response<T>, Status>)` for `poll_for_result`, and the `Result<i16, (i16, ProcessingError)>` format could not be used in this case AFAIU

dartdart26: It's not a big deal, you can leave as is if you want. To me it is a bit confusing, but if used in isolation and properly documented, it's fine.

eudelins-zama: So I guess if we don't want to go with ` (i16, Result<T, ProcessingError>)` signature, we should use a struct like

```rust
struct KmsClientResult<T, E> {
    error_counter: i16,
    operation_result: Result<T, E>
}
```

eudelins-zama: Tried the last option (custom struct), it is currently not that easy to make it work for `poll_for_result`.

I'll leave it as it is, as it is documented and used only there.

PR #1332, opened 2025-11-14
Georgi/fix/569

goshawk-3:
fixes https://github.com/zama-ai/fhevm-internal/issues/569

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/utils.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/utils.rs
@@ -64,20 +64,28 @@ pub fn safe_deserialize_sns_key<T: DeserializeOwned + Named + Unversionize>(
         .map_err(|e| FhevmError::DeserializationError(e.into()))
 }
 
-// Print first 4 and last 4 bytes of a blob as hex
-pub fn compact_hex(blob: &[u8]) -> String {
-    const OFFSET: usize = 8;
-    match blob.len() {
-        0 => String::from("0x"),
-        len if len <= 2 * OFFSET => format!("0x{}", hex::encode(blob)),
-        _ => {
-            let hex_str = hex::encode(blob);
-            format!(
+pub fn to_hex(blob: &[u8]) -> String {
+    let hex_str = hex::encode(blob);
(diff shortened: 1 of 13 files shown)


isaacdecoded (approved):
Great! 👏

coprocessor/fhevm-engine/fhevm-engine-common/src/utils.rs
```diff
@@ -64,21 +64,29 @@ pub fn safe_deserialize_sns_key<T: DeserializeOwned + Named + Unversionize>(
...
+    #[cfg(feature = "compact-hex")]
+    {
+        const OFFSET: usize = 8;
+        match blob.len() {
+            0 => String::from("0x"),
+            len if len <= 2 * OFFSET => format!("0x{}", hex_str),
+            _ => format!(
                 "0x{}...{}",
                 &hex_str[..OFFSET],
                 &hex_str[hex_str.len() - OFFSET..]
```
antoniupop, line 80: Do we need to use balanced trimming (4bytes..4bytes) here or could we have (4bytes..2bytes) instead? Saves a bit more and only the byte before last of metadata is useful among the last 4. We could even just take the first 4 bytes and drop the last part as the only info it contains is the CT type but will be the same most of the time.

antoniupop (approved):
LGTM - we can add this in tfhe-worker and scheduler in another PR.

coprocessor/fhevm-engine/fhevm-engine-common/src/utils.rs
goshawk-3: If we want to use it as a reference only - `first 4 bytes` could be enough.
For now , the the most meaningful bytes from handle suffix are `ct_idx` and `serialized_type`


```
    handle[21] = ct_idx as u8; 
    handle[22..30].copy_from_slice(&(aux_data.chain_id as u64).to_be_bytes());
    handle[30] = serialized_type as u8;
    handle[31] = current_ciphertext_version() as u8;
 ```

antoniupop: Yes, but CT index will never be visible in the compact version and I wouldn't want to move bytes around

PR #1334, opened 2025-11-14
chore(kms-connector): check decryption done

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/546

e2e tests: https://github.com/zama-ai/fhevm/actions/runs/19359282301/job/55387137249

### DIFF
--- a/kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
@@ -112,9 +112,8 @@ impl<P: Provider> DbEventProcessor<P> {
                     .await
             }
             GatewayEventKind::UserDecryption(req) => {
-                self.decryption_processor
-                    .check_decryption_not_already_done(req.decryptionId)
-                    .await?;
+                // No need to check decryption is done for user decrypt, as MPC parties don't
+                // communicate between each other for user decrypt
                 self.decryption_processor
                     .prepare_decryption_request(
                         req.decryptionId,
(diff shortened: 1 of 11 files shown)


dartdart26 (approved):
Looks good!

Just a question - is there a way to check decryption is not already done by sending the txn eagerly and checking the revert reason? This way, in the happy case we won't do eth_call?

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
```diff
@@ -109,6 +112,9 @@ impl<P: Provider> DbEventProcessor<P> {
                     .await
             }
             GatewayEventKind::UserDecryption(req) => {
+                self.decryption_processor
+                    .check_decryption_not_already_done(req.decryptionId)
+                    .await?;
```
dd23: we could in principle live without this check here, as _user decryptions_ always get processed on a single core (without communication with other cores - as opposed to public decrypts, where all MPC parties communicate).
So retrying those should be good and we either get a result, or we get ALREADY_EXISTS if we answered them before.
In both cases retries should stop from the connector, right?

eudelins-zama: Replaced the check by a comment, wdyt?

dd23: Looks good. Thx!

dd23 (approved):
Great! Looks good for the logic around `check_decryption_not_already_done`

PR #1335, opened 2025-11-14
chore(kms-connector): check decryption done

eudelins-zama:
Ref https://github.com/zama-ai/fhevm/pull/1334


dartdart26 (approved):

melanciani (approved):
has been approved in main (https://github.com/zama-ai/fhevm/pull/1334) and it seems to be identical , so looks good to me

PR #1342, opened 2025-11-14
chore(test-suite): bump e2e test versions for coprocessor

antoniupop:


eudelins-zama (commented):
Why not updating every component to `v0.10.0-3`?

Btw, I don't see the e2e tests action being run in the CI, is that expected?

antoniupop:
> Why not updating every component to `v0.10.0-3`?

Shouldn't that be updated by the PRs before being merged? The others were still -2 so assume nothing changed?

> Btw, I don't see the e2e tests action being run in the CI, is that expected?

You mean this one from the last PR?
https://github.com/zama-ai/fhevm/actions/runs/19368183773

eudelins-zama:
> Shouldn't that be updated by the PRs before being merged? The others were still -2 so assume nothing changed?

Previous PRs are not correlated to release, so I guess every time we do a release, we should update all services of the e2e tests at once, even if no update of the service was made between the 2 releases. I think this is what we've done until now, right @melanciani ?

> You mean this one from the last PR?

No, I mean whenever you update a file in `test-suite/fhevm` (`fhevm-cli` in your PR), it should trigger the e2e tests workflow in the CI, see the `check-changes` [job](https://github.com/zama-ai/fhevm/blob/main/.github/workflows/test-suite-e2e-tests.yml#L63-L81)

melanciani:
>Previous PRs are not correlated to release, so I guess every time we do a release, we should update all services of the e2e tests at once, even if no update of the service was made between the 2 releases. I think this is what we've done until now, right @melanciani ?

yes we should put all components to `0.10.0-3` for consistency 

> Btw, I don't see the e2e tests action being run in the CI, is that expected?

not expected and very weird, let's see once the new versions are pushed

antoniupop:
Yea it got triggered now

eudelins-zama (approved):

PR #1345, opened 2025-11-17
chore(gateway-contracts): improve mocked payment tasks and tests

melanciani:
- added `mint` to mocked erc20 for better test purposes
- updated mocked funding tasks to use `mint` instead of `transfer`
- increased default initial supply
- increased the default funding amount for the tx sender
- exposed 2 intermediate tasks (funding + approval) to facilitate further tests
- clarified tests


eudelins-zama (approved):
Thanks!

isaacdecoded (approved):
LGTM! 👍

PR #1348, opened 2025-11-17
chore(gateway-contracts): improve mocked payment tasks and tests (#1345)

melanciani:
refs https://github.com/zama-ai/fhevm/pull/1345


eudelins-zama (approved):
Thanks!

PR #1353, opened 2025-11-18
feat(test-suite): implement test for paused gateway contracts

isaacdecoded:
E2E tests passed [here](https://github.com/zama-ai/fhevm/actions/runs/19757703299/job/56613695862).

Refs https://github.com/zama-ai/fhevm-internal/issues/218


.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -182,6 +182,63 @@ jobs:
...
+        working-directory: test-suite/fhevm
+        run: |
+          ./fhevm-cli test paused-gateway-public-decrypt-http
+
+      - name: Unpause Gateway Contracts
+        working-directory: test-suite/fhevm
+        run: |
+          ./fhevm-cli unpause gateway
+
+      # E2E tests after unpausing the Gateway contracts
```
melanciani: I don't think this is great as it will make e2e tests much longer, maybe we could instead start with the pausing tests and then run the usual tests after unpausing the contracts ?

test-suite/e2e/test/pausedProtocol/pausedGateway.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  // InputVerification tests.
+  it('test paused gateway user input uint64 (non-trivial)', async function () {
+    const inputAlice = this.instances.alice.createEncryptedInput(
+      this.testInputContractAddress,
+      this.signers.alice.address,
+    );
+    inputAlice.add64(18446744073709550042n);
+
+    await expect(inputAlice.encrypt()).to.be.rejectedWith(new RegExp('Input request failed'));
```
melanciani, line 40: there is not way to make sure the revert is coming from `EnforcedPause` revert from the gateway ?

test-suite/e2e/test/pausedProtocol/pausedGateway.ts
```diff
@@ -0,0 +1,68 @@
...
+    await expect(
+      userDecryptSingleHandle(
+        handle,
+        this.userDecryptContractAddress,
+        this.instances.alice,
+        this.signers.alice,
+        privateKey,
+        publicKey,
+      ),
+    ).to.be.rejectedWith(new RegExp('User decrypt failed'));
```
melanciani, line 56: same

test-suite/e2e/test/pausedProtocol/pausedGateway.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  // PublicDecryption tests.
+  it('test paused gateway HTTPPublicDecrypt', async function () {
+    const handleBool = await this.httpPublicDecryptContract.xBool();
+    const handleAddress = await this.httpPublicDecryptContract.xAddress();
+    const handle32 = await this.httpPublicDecryptContract.xUint32();
+    await expect(this.instances.alice.publicDecrypt([handleAddress, handle32, handleBool])).to.be.rejectedWith(
+      new RegExp('Public decrypt failed'),
```
melanciani, line 65: same

melanciani:
a bit harder to write the tests, but would it also make sense to have a test that :
- ask for decryption/input proof
- pause directly
- checks test is ok 
to verify responses still come in after pausing ?

.github/workflows/test-suite-e2e-tests.yml
isaacdecoded: Makes sense, yes. Initially, I didn't want to change the current order to avoid affecting somehow the behavior of the tests with the pausing step, but now you said I think it's ok 👍

test-suite/e2e/test/pausedProtocol/pausedGateway.ts
isaacdecoded: I tried to reach the `EnforcedPause` error somehow, but the Relayer returns: 
```
Error: Relayer didn't response correctly. Bad status Bad Request. Content: {"message":"Transaction rejected: \"Input request failed: Transaction failed: RPC error: Could not estimate gas\""}
```

So I didn't find a way to get the error selector at this level of the test.

test-suite/e2e/test/pausedProtocol/pausedGateway.ts
isaacdecoded: Same as above, in this case the Relayer throws with:

`Error: User decrypt failed: relayer respond with HTTP code 500`

test-suite/e2e/test/pausedProtocol/pausedGateway.ts
isaacdecoded: And in this case, Relayer throws:

`Error: Public decrypt failed: relayer respond with HTTP code 500`

isaacdecoded:
> a bit harder to write the tests, but would it also make sense to have a test that :
> 
> * ask for decryption/input proof
> * pause directly
> * checks test is ok
>   to verify responses still come in after pausing ?

I'll take a second look at how to import and interact with the gateway contracts from inside the tests during execution.

test-suite/e2e/test/pausedProtocol/pausedGateway.ts
melanciani: I don't know exactly what RPC provider or tool the relayer uses, but I would say it's worth asking them if it would be possible to track down the underlying error when getting `Could not estimate gas` as it's not the first time we've seen this and it makes debugging much harder

test-suite/e2e/test/pausedProtocol/pausedGateway.ts
melanciani: would be worth asking them for here as well

melanciani:
> I'll take a second look at how to import and interact with the gateway contracts from inside the tests during execution.

great but let's avoid it if it's not relevant or it takes too much time to do !

test-suite/fhevm/fhevm-cli
```diff
@@ -177,10 +213,18 @@ case $COMMAND in
           log_message="${LIGHT_BLUE}${BOLD}[TEST] INPUT PROOF (uint64)${RESET}"
           docker_args+=("-g" "test add 42 to uint64 input and decrypt")
           ;;
+        paused-gateway-input-proof)
```
melanciani: looking at this again, maybe we could avoid making this file more verbose and just consider a single test command for gateway pausing, like : 
```
paused-gateway-requests)
          log_message="${LIGHT_BLUE}${BOLD}[TEST] PAUSED INPUT PROOF (uint64)${RESET}"
          docker_args+=("-g" "test paused gateway user input uint64|paused gateway user decrypt|paused gateway HTTPPublicDecrypt")
          ;;
```

melanciani: I mean, one of these tests should never fail because of pausing issues if the others succeed, and worse case the tests themselves have logs so tracking issues is still fine

isaacdecoded: Yes, good idea. I'm going to group these tests then, thanks! ✊

melanciani (dismissed):
thanks !

test-suite/fhevm/fhevm-cli
```diff
@@ -220,6 +256,10 @@ case $COMMAND in
           log_message="${LIGHT_BLUE}${BOLD}[TEST] RANDOM OPERATORS (SUBSET)${RESET}"
           docker_args+=("-g" "64 bits generate and decrypt|generating rand in reverting sub-call|64 bits generate with upper bound and decrypt")
           ;;
+        paused-gateway-requests)
+          log_message="${LIGHT_BLUE}${BOLD}[TEST] PAUSED INPUT PROOF (uint64)${RESET}"
```
melanciani: ah sorry, log message not up to date here

isaacdecoded: Done [here](https://github.com/zama-ai/fhevm/pull/1353/commits/941c97fddca7a63530355d2ed56e205555f6a98f), thanks for the catch!

melanciani (approved):
awesome thanks !!

isaacdecoded:
@mergifyio queue

PR #1357, opened 2025-11-18
fix(coprocessor): update tfhe-worker query for work

antoniupop:


coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -298,6 +298,18 @@ async fn query_for_work<'a>(
...
+      c.transaction_id
+    FROM (
+      SELECT transaction_id
+      FROM computations 
+      WHERE is_completed = FALSE
+        AND is_error = FALSE
+        AND is_allowed = TRUE
+      ORDER BY created_at
+      LIMIT $1
+    ) as c
```
rudy-6-4: both have same name "c" which is confusing

antoniupop: True - fixed.

goshawk-3 (approved):
LGTM

rudy-6-4 (approved):

PR #1362, opened 2025-11-19
chore(test-suite): remove inline comments in .env

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/599


melanciani (approved):
weird error from docker but ok let's do like this then 😅

PR #1369, opened 2025-11-19
chore(coprocessor): add more metrics for sns-worker

dartdart26:
Also, move code around to put metrics in one place.

### DIFF
--- a/charts/coprocessor/values.yaml
+++ b/charts/coprocessor/values.yaml
@@ -601,10 +601,12 @@ snsWorker:
     - --cleanup-interval=120s
     - --liveness-threshold=70s
     - --lifo=false
-    - --enable-compssion=true
+    - --enable-compression=true
     - --schedule-policy=rayon_parallel
     - --service-name=sns-worker
     - --log-level=INFO
+    # Only enable `gauge-update-interval-secs` for some of the workers to reduce DB load and not have duplicate metrics data for no reason.
+    # --gauge-update-interval-secs=10
     # - --pg-auto-explain-with-min-duration="30s"
     # - --keys-file-path
 
--- a/coprocessor/fhevm-engine/sns-worker/src/lib.rs
+++ b/coprocessor/fhevm-engine/sns-worker/src/lib.rs
@@ -3,6 +3,8 @@ mod executor;
 mod keyset;
 mod squash_noise;
 
+pub mod metrics;
+
 #[cfg(test)]
 mod tests;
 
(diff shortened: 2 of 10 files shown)


coprocessor/fhevm-engine/sns-worker/src/metrics.rs
```diff
@@ -0,0 +1,104 @@
...
+        "Number of uncomplete AWS uploads in sns-worker"
+    )
+    .unwrap()
+});
+
+pub fn spawn_gauge_update_routine(period: std::time::Duration, db_pool: PgPool) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        loop {
+            match sqlx::query_scalar(
+                "SELECT COUNT(*) FROM pbs_computations WHERE is_completed = FALSE",
```
goshawk-3, line 72: We'll probably deploy 16 replicas of sns-worker in prod. Shall we consider moving the `spawn_gauge_update_routine` out of the worker to avoid query duplication? 

Maybe we can have a separate service dedicated only to converting DB stats into prometheus metrics?

For now, we can make the config **pub gauge_update_interval_secs** optional and enable it only 1 or 2 replicas.

goshawk-3 (changes requested):

coprocessor/fhevm-engine/sns-worker/src/metrics.rs
dartdart26: Good point! Let me make it optional!

charts/coprocessor/values.yaml
obatirou: this should be `--enable-compression=true`, no?
How did this work before?

dartdart26: Good catch! This file is not being used in practice - instead, it is used as a reference. Let me fix it.

coprocessor/fhevm-engine/sns-worker/src/lib.rs
```diff
@@ -532,6 +543,15 @@ pub async fn run_all(
 
     let pg_mngr = pool_mngr.clone();
 
+    // Start metrics server
+    metrics_server::spawn(conf.metrics.addr.clone(), token.child_token());
+
+    // Start gauge update routine.
+    spawn_gauge_update_routine(
```
obatirou: No need for cancellation token?

dartdart26: I don't think it is important, yeah.

goshawk-3 (approved):
LGTM

PR #1374, opened 2025-11-20
Implement ZamaEthereumConfigInitializable contract

peersky:
This PR introduces ZamaEthereumConfigInitializable, an abstract contract designed to support EIP-1167 Minimal Proxy Clones (and other proxy patterns) when using the fhevm library.

It allows developers to initialize the FHEVM coprocessor configuration within an initialize() function, solving the issue where configuration is lost because constructors do not execute in the context of a proxy.

Todo: 

Constructor ideally should have bool `isTesting` variable to ensure both local tests & prod follows best practices: 

```solidity
constructor(bool isTesting)
{
   if(!isTesting) disableInitializers();
}
```

This is in a way breaking API change as current constructor takes no arguments. I just leave it open for maintainers to edit PR and figure out approach.


aquint-zama:
@cla-bot check

peersky:
Closes #1487

jatZama:
Hello, thanks for your PR. However, I would suggest something simpler and more flexible.
Instead of creating a new abstract contract, I would just add a single function:

```
function __ZamaEthereumConfig_init() internal {
        FHE.setCoprocessor(ZamaConfig.getEthereumCoprocessorConfig());
}
```

 to the current `ZamaEthereumConfig` abstract contract. With natspecs for this function explaining this should be used only inside an initializer of a proxy contract (eg clones, upgradeable contracts, etc).
 All the other logic around state management for initializer could be handled by OZ's library.
 Also adding a test for this function and an example in documentation would be greatly appreciated.
 Also don't forget to rebase on main otherwise merging on main won't work.

peersky:
> Hello, thanks for your PR. However, I would suggest something simpler and more flexible. Instead of creating a new abstract contract, I would just add a single function:
> 
> ```
> function __ZamaEthereumConfig_init() internal {
>         FHE.setCoprocessor(ZamaConfig.getEthereumCoprocessorConfig());
> }
> ```
> 
> to the current `ZamaEthereumConfig` abstract contract. With natspecs for this function explaining this should be used only inside an initializer of a proxy contract (eg clones, upgradeable contracts, etc). All the other logic around state management for initializer could be handled by OZ's library. Also adding a test for this function and an example in documentation would be greatly appreciated. Also don't forget to rebase on main otherwise merging on main won't work.

@jatZama Thanks for the review

I agree that your approach is technically cleaner and reduces boilerplate. However, I have two points I'd like to raise before we commit to that direction:

1/ **Developer Experience & Safety** The challenge with a hidden `__init` function is that it breaks the "industry standard" pattern (e.g., `ERC20` vs `ERC20Upgradeable`). If a developer imports `ZamaEthereumConfig` into a proxy setup, it isn't self-evident that initialization is required. The contract will deploy without error, but runtime execution will silently fail (or revert confusingly) because the storage slots are empty.
Explicitly naming it `ZamaEthereumConfigInitializable` acts as a "warning sign" that is easy to spot in the documentation _outline_ (sadly, not everyone really read manuals beyond outlines) and IDE autocompletion, alerting developers that they must handle state initialization differently.

2/ **Alternative: Stateless Configuration (Zero Initialization)** To address the root cause, could we remove the need for storage/initialization entirely?

Since `Impl.sol` and `ZamaConfig.sol` are internal libraries, their bytecode is inlined. Instead of reading from a storage slot (which requires `SSTORE` to set up and `SLOAD` to read), we could modify `Impl.sol` to have immutable(constant) variables baked in to the bytecode. 

It would allow zero initialization code - No need for `initialize()` calls or storage namespaces, gas savings (_1 `SLOAD` is ~100-2100 gas on every single FHE operation_) .

The trade-off is that this would effectively make the coprocessor addresses immutable. If Zama rotates the Coprocessor address, we would need to upgrade the Implementation contract rather than calling `setCoprocessor`. Given that the FHE lib is likely used behind upgradeable proxies anyway, this seems like a reasonable trade-off for the massive DX and gas wins.

jatZama:
> > Hello, thanks for your PR. However, I would suggest something simpler and more flexible. Instead of creating a new abstract contract, I would just add a single function:
> > ```
> > function __ZamaEthereumConfig_init() internal {
> >         FHE.setCoprocessor(ZamaConfig.getEthereumCoprocessorConfig());
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to the current `ZamaEthereumConfig` abstract contract. With natspecs for this function explaining this should be used only inside an initializer of a proxy contract (eg clones, upgradeable contracts, etc). All the other logic around state management for initializer could be handled by OZ's library. Also adding a test for this function and an example in documentation would be greatly appreciated. Also don't forget to rebase on main otherwise merging on main won't work.
> 
> @jatZama Thanks for the review
> 
> I agree that your approach is technically cleaner and reduces boilerplate. However, I have two points I'd like to raise before we commit to that direction:
> 
> 1/ **Developer Experience & Safety** The challenge with a hidden `__init` function is that it breaks the "industry standard" pattern (e.g., `ERC20` vs `ERC20Upgradeable`). If a developer imports `ZamaEthereumConfig` into a proxy setup, it isn't self-evident that initialization is required. The contract will deploy without error, but runtime execution will silently fail (or revert confusingly) because the storage slots are empty. Explicitly naming it `ZamaEthereumConfigInitializable` acts as a "warning sign" that is easy to spot in the documentation _outline_ (sadly, not everyone really read manuals beyond outlines) and IDE autocompletion, alerting developers that they must handle state initialization differently.
> 
> 2/ **Alternative: Stateless Configuration (Zero Initialization)** To address the root cause, could we remove the need for storage/initialization entirely?
> 
> Since `Impl.sol` and `ZamaConfig.sol` are internal libraries, their bytecode is inlined. Instead of reading from a storage slot (which requires `SSTORE` to set up and `SLOAD` to read), we could modify `Impl.sol` to have immutable(constant) variables baked in to the bytecode.
> 
> It would allow zero initialization code - No need for `initialize()` calls or storage namespaces, gas savings (_1 `SLOAD` is ~100-2100 gas on every single FHE operation_) .
> 
> The trade-off is that this would effectively make the coprocessor addresses immutable. If Zama rotates the Coprocessor address, we would need to upgrade the Implementation contract rather than calling `setCoprocessor`. Given that the FHE lib is likely used behind upgradeable proxies anyway, this seems like a reasonable trade-off for the massive DX and gas wins.

For 1/ : I agree that you could have a second abstract contract called `ZamaEthereumConfigInitializable` if it is clearer, but in this case I prefer please not to rewrite the initializer logic, which would still be inherited from OZ. Cleanest way to do it would then simply to make  another lightweight `ZamaEthereumConfigInitializable` abstract contract such as : 

```
///// TODO: some natspecs here to say this should be used with proxies
abstract contract ZamaEthereumConfigInitializable {

    /**
     * @dev Should be called inside your proxy's initialize function, 
     * typically should be protected by an `onlyInitializing` modifer.
     * This sets up the FHE coprocessor configuration for the specific network.
     */
    function __ZamaEthereumConfig_init() internal {
        FHE.setCoprocessor(ZamaConfig.getEthereumCoprocessorConfig());
    }

    function confidentialProtocolId() public view returns (uint256) {
        return ZamaConfig.getConfidentialProtocolId();
    }
}
```

Note 2 differences compared to your code: 
a/ that I removed the `onlyInitializing` to not have to import OZ library here, but I added a comment about it in natspecs. We should add a page in documentation to give a concrete exemple on how to use it with OZ library, as said in my previous message.
b/ also that I did not put a constructor for ZamaEthereumConfigInitializable, since implementations contract typically just call a `_disableInitializers();` in their constructor.

For you second point, actually we tried very hard to find a way to put the addresses as constants in bytecode, but due to a Solidity language limitation with libraries, this is not possible, so we had to write them in storage. The only soludtion could have been to make `FHE.sol` as a big abstract contract instead of a library, but we decided against doing this, because it would make devX worse, as for eg we could not use `FHE.` neither `using FHE for *` (very useful for chaining fhe ops to make code more concise) syntax anymore.

peersky:
> to not have to import OZ library here

You did not have external dependency for OZ libraries in this PR. I would still advise blocking re-initialization natively to prevent accidental overwrites, even without OZ. A simple `bool` check is lightweight:

> we tried very hard to find a way to put the addresses as constants in bytecode

We can achieve this without writing to storage. We can define the addresses as private constants and use a stateless `view` function to return the correct configuration based on the current chain ID.

```solidity
// SPDX-License-Identifier: BSD-3-Clause-Clear
pragma solidity ^0.8.24;

import {CoprocessorConfig} from "../lib/Impl.sol";

library ZamaConfig {
    error ZamaProtocolUnsupported();

    // =============================================================
    //               HARDCODED ADDRESS CONSTANTS
    // =============================================================
    
    // Sepolia (ChainID: 11155111)
    address private constant SEPOLIA_ACL = 0xf0Ffdc93b7E186bC2f8CB3dAA75D86d1930A433D;
    address private constant SEPOLIA_COPROCESSOR = 0x92C920834Ec8941d2C77D188936E1f7A6f49c127;
    address private constant SEPOLIA_KMS = 0xbE0E383937d564D7FF0BC3b46c51f0bF8d5C311A;

    // Local/Anvil (ChainID: 31337)
    address private constant LOCAL_ACL = 0x50157CFfD6bBFA2DECe204a89ec419c23ef5755D;
    address private constant LOCAL_COPROCESSOR = 0xe3a9105a3a932253A70F126eb1E3b589C643dD24;
    address private constant LOCAL_KMS = 0x901F8942346f7AB3a01F6D7613119Bca447Bb030;

    // =============================================================
    //                  THE "PURE GETTER" LOGIC
    // =============================================================

    /**
     * @notice Returns config purely from bytecode logic. 
     * @dev No SLOAD. No Initialization required. Works with Proxies instantly.
     */
    function getEthereumCoprocessorConfig() internal view returns (CoprocessorConfig memory) {
        uint256 chainId = block.chainid;

        if (chainId == 11155111) {
            return CoprocessorConfig({
                ACLAddress: SEPOLIA_ACL,
                CoprocessorAddress: SEPOLIA_COPROCESSOR,
                KMSVerifierAddress: SEPOLIA_KMS
            });
        } 
        else if (chainId == 31337) {
            return CoprocessorConfig({
                ACLAddress: LOCAL_ACL,
                CoprocessorAddress: LOCAL_COPROCESSOR,
                KMSVerifierAddress: LOCAL_KMS
            });
        }
        
        // Add Mainnet (1) logic here when available...
        
        revert ZamaProtocolUnsupported();
    }
}
```

Since Impl.sol is also an internal library, the Solidity compiler will inline the code above directly into your contract.

```solidity
// Inside Impl.sol

// 1. Remove the storage slot definition entirely
// bytes32 private constant CoprocessorConfigLocation = ...; // DELETE THIS

// 2. Update the getter to return memory, not storage
function getCoprocessorConfig() internal view returns (CoprocessorConfig memory) {
    // This looks like a function call, but in the compiled bytecode 
    // it is just a JUMP to the logic defined above.
    return ZamaConfig.getEthereumCoprocessorConfig();
}

// 3. Update consuming functions to read from memory
function add(bytes32 lhs, bytes32 rhs, bool scalar) internal returns (bytes32 result) {
    // ... setup ...
    
    // Fetch from bytecode logic (Costs ~20 gas)
    CoprocessorConfig memory config = getCoprocessorConfig(); 
    
    // Use the address
    result = IFHEVMExecutor(config.CoprocessorAddress).fheAdd(lhs, rhs, scalarByte);
}
```

This saves ~2100 gas (Cold `SLOAD`) on every FHE operation and removes the need for `initialize()`. (Note: for massive multi-chain scaling beyond ~50 networks, a deterministic `CREATE2` Registry would be the ideal stateless solution to avoid bytecode bloat).

If you agree, we can proceed with storage-based `ZamaEthereumConfigInitializable` approach for this PR to get it merged, as it's a solid middle ground that supports upgradability. 
The stateless pattern above as a potential optimization for next iterations.

jatZama:
> You do not have external dependency for OZ libraries in this repo. Are you sure you want to introduce a heavy dependency just for this logic?

Of course no, my point was exactly to **not** make our package import OZ as a real dep, but maybe I was not clear enough. But on the other hand, I don't want neither to copy/paste code coming from OZ (not maintainable to do so), the reinitializer logic should be imported from OZ in the dapp, **not in our fhevm library**. But we will show in docs examples that it is the responsibility of the dapp developer to inherit it from OZ.

peersky:
> Of course no, my point was exactly to **not** make our package import OZ as a real dep

I miss red that, I see, that's good! Edited my comment

peersky:
> Your example won't work with FHE.sol library unfortunately. My point was that this would break all tests, because you need to call these contracts if you look at the code of FHE.sol. I will let you run tests with your suggestion if you are not convinced.



I took a look at your tests. It makes sense now. Refactoring the integration tests to support stateless config would be a massive undertaking.

For a future V2, stateless with `CREATE2` at the same address on all chains would be nice to have, but you'd likely need test utils to use something like `vm.etch` to inject precompile mocks.

But for now, `ZamaEthereumConfigInitializable` is the pragmatic choice. I'll update the PR with the changes requested shortly.

jatZama:
> I took a look at your tests. It makes sense now. Refactoring the integration tests to support stateless config would be a massive undertaking.

We don't want to hardcode any constant address inside FHE.sol library, because it should be able to support any chain and potentially other protocols than Zama protocol, as long as they follow the https://eip.tools/eip/7995
This is why the pattern you suggested is not acceptable, because you import the ZamaConfig inside Impl.sol, this should not be the case.
As said before, Solidity language has limitations: you cannot import constants coming from an external file inside a Solidity library, so there is no other way to put them in storage unfortunately. At least this has the advantageous side effects of making addresses easily accessible via storage (it would be a lot harder to track if hidden inside bytecode).

jatZama:
> For a future V2, stateless with CREATE2 at the same address on all chains would be nice to have, but you'd likely need test utils to use something like vm.etch to inject precompile mocks.

We are already using vm.etch , or rather the equivalent setCode in mocked tests btw.
And we decided against using stuff like keyless deployment method (Nick's method) or using CREATE2 factory, because those are not available universally in all EVMs, make deployment more complex, and the keyless one cost a lot more gas to delploy. And anyways, we expect the FHE.sol to be usable even by multiple confidential protocols on the same chain if available, so it is not required to have unique addresses.

PR #1377, opened 2025-11-20
feat(test-suite): implement tests on paused host contracts

isaacdecoded:
E2E tests passed [here](https://github.com/zama-ai/fhevm/actions/runs/19762861881/job/56628621823).

Refs https://github.com/zama-ai/fhevm-internal/issues/218


test-suite/e2e/test/pausedProtocol/pausedHost.ts
```diff
@@ -0,0 +1,45 @@
...
+
+describe('Paused host', function () {
+  before(async function () {
+    await initSigners(2);
+    this.signers = await getSigners();
+    this.instances = await createInstances(this.signers);
+  });
+
+  // ACL.allow test.
+  it('test paused host user input uint64 (non-trivial)', async function () {
```
melanciani: I think there are some tests missing : pausing host should revert any acl/fhe ops as well, ie tests like `erc20` should revert as well

isaacdecoded: @melanciani I took a look at the EncryptedERC20 transfer flow, and it seems it requires `ACL.allow` calls as we can [see here](https://github.com/zama-ai/fhevm/blob/main/test-suite/e2e/contracts/EncryptedERC20.sol#L207-L219), and also the [mint step](https://github.com/zama-ai/fhevm/blob/main/test-suite/e2e/contracts/EncryptedERC20.sol#L60-L66) requires `ACL.allow` calls. Meaning that I'm not sure we should add the ERC20 transfer test case since the paused  behavior of the `ACL.allow` is already tested by this input proof test case.

Wdyt?

melanciani: yes it won't validate that it pauses operations indeed

then in that cas could you create a new test for pausing a single operators ? one of the `operators` one should be enough !

isaacdecoded: Operator test case added [here](https://github.com/zama-ai/fhevm/pull/1377/files#diff-cf64f92d3332679c648c4f9f68fd472e057474205c84eb1f0ee7f393d2a5846bR34-R54).

test-suite/fhevm/fhevm-cli
```diff
@@ -217,6 +217,10 @@ case $COMMAND in
           log_message="${LIGHT_BLUE}${BOLD}[TEST] PAUSED INPUT PROOF (uint64)${RESET}"
           docker_args+=("-g" "test paused gateway user input uint64")
           ;;
+        paused-host-input-proof)
```
melanciani: same as for gateway (https://github.com/zama-ai/fhevm/pull/1353/files#r2576327853): maybe better to consider a single command

test-suite/fhevm/fhevm-cli
```diff
@@ -260,6 +260,10 @@ case $COMMAND in
           log_message="${LIGHT_BLUE}${BOLD}[TEST] PAUSED INPUT PROOF (uint64)${RESET}"
           docker_args+=("-g" "test paused gateway.*")
           ;;
+        paused-host-requests)
+          log_message="${LIGHT_BLUE}${BOLD}[TEST] PAUSED HOST INPUT PROOF (uint64)${RESET}"
```
melanciani: log message not up to date here !

test-suite/fhevm/fhevm-cli
```diff
@@ -260,6 +260,10 @@ case $COMMAND in
           log_message="${LIGHT_BLUE}${BOLD}[TEST] PAUSED INPUT PROOF (uint64)${RESET}"
           docker_args+=("-g" "test paused gateway.*")
           ;;
+        paused-host-requests)
```
melanciani: `requests` makes a bit less sense for host, maybe `paused-host-conctracts` and then align the gateway command on that as well ?

test-suite/fhevm/fhevm-cli
isaacdecoded: Done [here](https://github.com/zama-ai/fhevm/pull/1377/commits/97b754632a5cdfb689daaa29d45f3bf6b0c98cac) also, thanks!

PR #1378, opened 2025-11-20
feat(coprocessor): host-listener, report any update during a catchup

rudy-6-4:


coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -922,7 +922,15 @@ async fn db_insert_block_no_retry(
             is_allowed,
             ..tfhe_log
         };
-        db.insert_tfhe_event(&mut tx, &tfhe_log).await?;
+        let inserted = db.insert_tfhe_event(&mut tx, &tfhe_log).await?;
+        if block_logs.catchup && inserted {
```
obatirou, line 949: Could be done only once outside the loop tfhe_log in tfhe_event_log?

obatirou: Depends if you want a log for each event (but maybe change the field and message to be event specific)

obatirou (dismissed):
lgtm, minor nit

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
rudy-6-4: I have merged with the existing log. And added a global summary.

obatirou (approved):

PR #1382, opened 2025-11-21
docs(common): add wallet integration guide for partners

ankurdotb:


jatZama (dismissed):

docs/examples/wallet-guide.md
```diff
@@ -0,0 +1,102 @@
...
+
+## **Core concepts in this guide**
+
+While building support for ERC-7984 confidential tokens in your wallet/app, you might come across the following terminology related to [various parts of the Zama Protocol](../protocol/architecture/overview.md). A brief explanation of common terms you might encounter are:
+
+- **FHEVM**: Zama’s [FHEVM library](../protocol/architecture/library.md) that supports computations on encrypted values. Encrypted values are represented on‑chain as **ciphertext handles** (bytes32).
+- **Host chain**: The EVM network your users connect to in a wallet with confidential smart contracts. Example: Ethereum / Ethereum Sepolia.
+- **Gateway chain**: Zama’s Arbitrum L3 [Gateway chain](../protocol/architecture/gateway.md) that coordinates FHE encryptions/decryptions.
+- **Relayer**: Off‑chain [Relayer](../protocol/architecture/relayer_oracle.md) that registers encrypted inputs, coordinate decryptions, and return results to users or contracts. Wallets and dApps talk to the Relayer via the JavaScript SDK.
+- **ACL:** Access control for ciphertext handles. Contracts grant per‑address permissions so a user can re‑encrypt and read their own data.
```
jatZama: Not sure you want to keep re-encrypt. This terminology has been deprecated and replaced by "user decrypt" almost everywhere, even if technically it made sense.

jatZama:
Don't forget to amend commit messages to include scope, otherwise CI will keep failing.
For eg : `docs(common): blaba` or `docs(library-solidity): blabla`

docs/examples/wallet-guide.md
ankurdotb: I deliberately tried to avoid language that is Zama-specific and wouldn't be understandable unless someone was already familiar with our specific terminology like "user decrypt", "public decrypt" etc. I expect the audience to be product teams at first pass, then developers will use the linked deep-dives.

jatZama: But re-encrypt is even less understandable than user decrypt. I am not sure it makes sense to reintroduce reencrypt here. But I will let leadership decide. Maybe @mortendahl or @immortal-tofu could settle on this.

ankurdotb: Let me think of a way to rephrase it

ankurdotb: @jatZama Rewrote that line to drop any reference to encryptions (not strictly necessary at that point I reckon). Hope this update version works!

mortendahl: I would not use "reencrypt", since the terminology elsewhere is "user decrypt". But sounds like it's been rewritten anyways.

jatZama (approved):

PR #1386, opened 2025-11-24
feat: host listener poller

obatirou:
Closes https://github.com/zama-ai/fhevm-internal/issues/633


coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
```diff
@@ -0,0 +1,129 @@
...
+struct Args {
+    #[arg(
+        long = "url",
+        alias = "rpc-url",
+        help = "L1 node HTTP JSON-RPC endpoint (HTTP only; ws not supported)"
+    )]
+    url: String,
+
+    #[arg(long, help = "ACL contract address to monitor")]
+    acl_contract_address: Option<Address>,
```
rudy-6-4: Make it mandatory

coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    #[arg(long, help = "TFHE contract address to monitor")]
+    tfhe_contract_address: Option<Address>,
```
rudy-6-4: Make it mandatory

coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
```diff
@@ -0,0 +1,129 @@
...
+        help = "Backoff between retry attempts for RPC/DB failures in milliseconds"
+    )]
+    retry_interval_ms: u64,
+
+    #[arg(
+        long,
+        default_value_t = 10,
+        help = "Maximum number of HTTP/RPC retry attempts (in addition to the initial attempt) before failing an operation"
+    )]
+    max_http_retries: u64,
```
rudy-6-4: That should be bigger ? especially if retry_interval_ms is made small by error.
E.g. 
If the chain node is not reachable for 11s. It stops.
Hoping it will restarted, and doing its auto catchup.

Could be a lot bigger or set as a time like 60 seconds

coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
```diff
@@ -0,0 +1,129 @@
...
+        help = "Depth behind the head considered final (in blocks)"
+    )]
+    finality_lag: u64,
+
+    #[arg(
+        long,
+        default_value_t = 100,
+        help = "Maximum number of blocks to process per iteration"
+    )]
+    batch_size: u64,
```
rudy-6-4, line 48: no sure about the name, it's like paging in catchup (to avoid error ofchain node).

coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
obatirou: You mean the default value?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -461,7 +461,47 @@ impl Database {
         Ok(())
     }
 
-    pub async fn read_last_valid_block(&mut self) -> Option<i64> {
+    pub async fn get_last_caught_up_block(
```
rudy-6-4: prefix/suffix with poller

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -461,7 +461,47 @@ impl Database {
...
+        )
+        .bind(chain_id)
+        .bind(block)
+        .execute(&pool)
+        .await?;
+
+        Ok(())
+    }
+
+    pub async fn read_last_valid_block(&self) -> Option<i64> {
```
rudy-6-4, line 517: prefix/suffix with ???

coprocessor/fhevm-engine/host-listener/src/poller/http_client.rs
```diff
@@ -0,0 +1,272 @@
...
+                self.provider.get_chain_id().await.map_err(|e| anyhow!(e))
+            },
+        )
+        .await
+    }
+
+    pub async fn latest_block_number(
+        &self,
+    ) -> Result<(u64, u64), RetryError<anyhow::Error>> {
+        retry_with_delay(
```
rudy-6-4: could have a self.retry_with_delay to avoid always passing self.retry_interval and self.max_retries.
This make retry with lamnda param also simpler to read

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
```diff
@@ -304,7 +304,7 @@ async fn test_listener_no_event_loss(
     // Kill the listener
```
rudy-6-4, line 299: file could be renamed to be clear there is now 2 listeners with dedicated tests

test-suite/fhevm/docker-compose/coprocessor-docker-compose.yml
```diff
@@ -43,6 +43,26 @@ services:
...
+        - type=gha,mode=max
+    env_file:
+      - ../env/staging/.env.coprocessor.local
+    command:
+      - host_listener_poller
+      - --database-url=${DATABASE_URL}
+      - --coprocessor-api-key=${TENANT_API_KEY}
+      - --acl-contract-address=${ACL_CONTRACT_ADDRESS}
+      - --tfhe-contract-address=${FHEVM_EXECUTOR_CONTRACT_ADDRESS}
+      - --url=${RPC_URL}
```
rudy-6-4: Probably in ws format ?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
obatirou: Why prefix?
This one is used by both

coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
obatirou: as you prefer 
No strong opinion on this naming, both ok for me

test-suite/fhevm/docker-compose/coprocessor-docker-compose.yml
```diff
@@ -37,12 +37,35 @@ services:
...
       - --tfhe-contract-address=${FHEVM_EXECUTOR_CONTRACT_ADDRESS}
-      - --url=${RPC_URL}
+      - --url=${RPC_WS_URL}
       - --initial-block-time=1
     depends_on:
       coprocessor-db-migration:
         condition: service_completed_successfully
 
+  coprocessor-poller:
+    container_name: coprocessor-poller
```
antoniupop: Name is too vague -  `coprocessor-host-listener-poller` ?

charts/coprocessor/values.yaml
```diff
@@ -214,6 +214,91 @@ hostListener:
       maxSurge: 1
       maxUnavailable: 0
 
+# -----------------------------------------------------------------------------
+# Poller
+# -----------------------------------------------------------------------------
+# Processes blockchain events from the host chain
+poller:
```
antoniupop: Same for naming - preferably we use names that are precise, we could have other polling services, so I'd rather have the name explicitly link this to host listener. So `hostPoller` or even better `hostListenerPoller`.

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -187,6 +187,12 @@ jobs:
         run: |
           ./fhevm-cli test delegate-user-decryption
 
+      - name: Host listener poller test
+        working-directory: test-suite/fhevm
+        run: |
+          docker stop coprocessor-host-listener
+          ./fhevm-cli test erc20
+
```
eudelins-zama, line 201: Shouldn't the coprocessor host-listener be restarted after the erc20 test?

My point is that this currently work because this test is the last one in the workflow. But if we add more tests in the future, or we change the position of this test, the test-suite should still pass IMO, and I don't think it is currently the case

obatirou: Good point, will make the change 👍 
If the poller is correctly implemented (no bugs hopefully), even if coprocessor-host-listener is stopped the test-suite should pass with a delay. But I agree, the rest of the suite should not rely on the poller which should be considered as a backup of the host listener.

obatirou: done with [bc2033a](https://github.com/zama-ai/fhevm/pull/1386/commits/bc2033a4458d36bc4e04fc08891a7a2d21ae9353)

coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
rudy-6-4: yes maybe something covering 1 minute

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -0,0 +1,103 @@
...
+                }
+                db.handle_acl_event(
+                    &mut tx,
+                    &event,
+                    &log.transaction_hash,
+                    chain_id,
+                    block_hash.as_ref(),
+                    block_number,
+                )
+                .await?;
```
nboisde: Just a question, what happens in this case: if the db.handle_acl_event fails ? 
What is the behaviour ? Is the insertion for all thoses logs are retried ?

coprocessor/fhevm-engine/host-listener/src/poller/http_client.rs
```diff
@@ -0,0 +1,254 @@
...
+        mut op: F,
+    ) -> Result<(T, u64), RetryError<E>>
+    where
+        F: FnMut() -> Fut,
+        Fut: Future<Output = Result<T, E>>,
+        E: Display + Debug,
+    {
+        let mut retries = 0;
+        loop {
+            match op().await {
```
nboisde: Really nice feature

nboisde (commented):
I don't know if you ran this single component pointing on actual live environments, but could worth the try.
And how do you run multiples pollers if needed, you create new binaries instances ? Is it possible with your current db setup ?

Otherwise, lgtm, thanks @obatirou

obatirou:
> I don't know if you ran this single component pointing on actual live environments, but could worth the try. And how do you run multiples pollers if needed, you create new binaries instances ? Is it possible with your current db setup ?
> 
> Otherwise, lgtm, thanks @obatirou

I did not tried on actual live environments, only in local e2e with fhevmcli.
For now, the poller is designed to have only _one poller per chain_id_.
If we want multiple pollers on the same chain_id, coordination would be needed on `last_caught_up_block`, and possibly more cases to handle I suppose. We could do that in a second time if necessary? One is not enough for now?

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
obatirou: `ingest_block_logs` is used by `ingest_with_retry` which have a retry_interval up to 10 times in `poller/mod.rs`.
The block ingestion runs in a single database transaction. So if fails here the db tx is not committed and retried after the `retry_interval`.

nboisde:
Lets try with one sure @obatirou ! We could set up for Ethereum with 2 blocks latency for handling reorgs with certitude. 
Interesting numbers for Ethereum:
10-25 reorgs / day
max depth / day: 1 
max known depth: 1
avg block time: 12.08

Theoretical max depth is 15 I think.
With 2 blocks latency it should handle reorgs properly and catch up missed events approximately in 24 seconds delay.

charts/coprocessor/values.yaml
```diff
@@ -214,6 +214,91 @@ hostListener:
       maxSurge: 1
       maxUnavailable: 0
 
+# -----------------------------------------------------------------------------
+# Host Listener Poller
+# -----------------------------------------------------------------------------
+# Processes blockchain events from the host chain
+hostListenerPoller:
```
piizama, line 221: We are missing the template part associated with these values. It should consist in a Deployment similar to the regular hostListener.

piizama (changes requested):

charts/coprocessor/values.yaml
obatirou: Done with [acfe4b1](https://github.com/zama-ai/fhevm/pull/1386/commits/acfe4b14dd6e4570181eb230f12dc0cfc6f1e1c5)

piizama (dismissed):

coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
```diff
@@ -0,0 +1,129 @@
...
+    acl_contract_address: Address,
+
+    #[arg(long, help = "TFHE contract address to monitor")]
+    tfhe_contract_address: Address,
+
+    #[arg(long, help = "PostgreSQL connection URL")]
+    database_url: DatabaseURL,
+
+    #[arg(long, help = "Coprocessor API key")]
+    coprocessor_api_key: Option<Uuid>,
```
rudy-6-4, line 34: This one could also be mandatory, it's optional in host-listener for historical test reason. (could be done later)

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -0,0 +1,112 @@
...
+        if acl_contract_address.is_none() || is_acl_address {
+            if let Ok(event) =
+                AclContract::AclContractEvents::decode_log(&log.inner)
+            {
+                info!(acl_event = ?event, "ACL event");
+                let handles = acl_result_handles(&event);
+                for handle in handles {
+                    is_allowed.insert(handle.to_vec());
+                }
+                db.handle_acl_event(
```
rudy-6-4: some rebase miss, here the insertion should be checked to display missed event

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -863,116 +853,6 @@ async fn db_insert_block(
...
-        let transaction_hash = log.transaction_hash;
-        if acl_contract_address.is_none() || is_acl_address {
-            if let Ok(event) =
-                AclContract::AclContractEvents::decode_log(&log.inner)
-            {
-                let handles = acl_result_handles(&event);
-                for handle in handles {
-                    is_allowed.insert(handle.to_vec());
-                }
-                let inserted = db
```
rudy-6-4, line 903: This part has been lost

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -863,116 +853,6 @@ async fn db_insert_block(
...
-                    .handle_acl_event(
-                        &mut tx,
-                        &event,
-                        &log.transaction_hash,
-                        chain_id,
-                        block_hash.as_ref(),
-                        block_number,
-                    )
-                    .await?;
-                if block_logs.catchup && inserted {
```
rudy-6-4, line 913: This part has been lost

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -863,116 +853,6 @@ async fn db_insert_block(
...
-            if let Some(result_handle) = tfhe_result_handle(&tfhe_log.event) {
-                is_allowed.contains(&result_handle.to_vec())
-            } else {
-                false
-            };
-        let tfhe_log = LogTfhe {
-            is_allowed,
-            ..tfhe_log
-        };
-        let inserted = db.insert_tfhe_event(&mut tx, &tfhe_log).await?;
```
rudy-6-4, line 959: This part has been lost

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -863,116 +853,6 @@ async fn db_insert_block(
...
-        if block_logs.catchup && inserted {
```
rudy-6-4, line 960: This part has been lost

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -863,116 +853,6 @@ async fn db_insert_block(
...
-            info!(tfhe_log = ?tfhe_log, "TFHE event missed before");
-            catchup_insertion += 1;
-        } else {
-            info!(tfhe_log = ?tfhe_log, "TFHE event");
-        }
-    }
-    if catchup_insertion == block_logs.logs.len() {
```
rudy-6-4, line 967: This part has been lost

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -0,0 +1,112 @@
...
+            if let Some(result_handle) = tfhe_result_handle(&tfhe_log.event) {
+                is_allowed.contains(&result_handle.to_vec())
+            } else {
+                false
+            };
+        let tfhe_log = LogTfhe {
+            is_allowed,
+            ..tfhe_log
+        };
+        db.insert_tfhe_event(&mut tx, &tfhe_log).await?;
```
rudy-6-4: some rebase miss, here the insertion should be checked to display missed event

coprocessor/fhevm-engine/host-listener/src/poller/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+        last_caught_up_block = last_caught_up_block,
+        finality_lag = config.finality_lag,
+        batch_size = config.batch_size,
+        poll_interval_ms = config.poll_interval.as_millis(),
+        retry_interval_ms = config.retry_interval.as_millis(),
+        "Starting host-listener poller"
+    );
+
+    loop {
+        let (latest, latest_retries) = match client.latest_block_number().await
```
rudy-6-4: what happens if the client gets invalid ? Can the provider be in a bad state at some point ?

coprocessor/fhevm-engine/host-listener/src/poller/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+                Err(err) => {
+                    http_retries += err.retries;
+                    error!(
+                        block = block,
+                        retries = err.retries,
+                        error = %err.error,
+                        "Failed to fetch logs for block after retries"
+                    );
+                    rpc_errors += 1;
+                    break;
```
rudy-6-4, line 278: maybe a comment saying the other block will be retried later, next loop turn ?

coprocessor/fhevm-engine/host-listener/src/poller/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+                    Err(err) => {
+                        http_retries += err.retries;
+                        error!(
+                            block = block,
+                            retries = err.retries,
+                            error = %err.error,
+                            "Failed to fetch header for block after retries"
+                        );
+                        rpc_errors += 1;
+                        break;
```
rudy-6-4: maybe a comment saying the other blocks will be retried later, next loop turn ?

charts/coprocessor/values.yaml
```diff
@@ -214,6 +214,91 @@ hostListener:
...
+      valueFrom:
+        secretKeyRef:
+          name: coprocessor-api-key
+          key: coprocessor-api-key
+
+  # Command line arguments for the host listener poller
+  args:
+    - --database-url=$(DATABASE_URL)
+    - --coprocessor-api-key=$(TENANT_API_KEY)
+    - --url=$(ETHEREUM_RPC_URL)
```
rudy-6-4: Could have HTTP explicitely, like ETHEREUM_RPC_HTTP_URL

coprocessor/fhevm-engine/host-listener/src/poller/mod.rs
obatirou: This is a very good point.
There is no reconnection mechanism

obatirou: Done [4330d7b](https://github.com/zama-ai/fhevm/pull/1386/commits/4330d7b5cfb5d57f76bee366a63411c96ce04c9c)

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -0,0 +1,148 @@
...
+    acl_contract_address: &Option<Address>,
+    tfhe_contract_address: &Option<Address>,
+) -> Result<(), sqlx::Error> {
+    let mut tx = db.new_transaction().await?;
+    let mut is_allowed = HashSet::<Handle>::new();
+    let mut tfhe_event_log = vec![];
+    let block_hash = block_logs.summary.hash;
+    let block_number = block_logs.summary.number;
+    let mut catchup_insertion = 0;
+    let summary_offset_dt = OffsetDateTime::from_unix_timestamp(
```
rudy-6-4: the name is adding confusion, it took me e few seconds to see offset and dt are the type of the value.
eg dt  looks like delta-time
offset could be an offset :)
maybe it should be in a function, like fn block_timestampz to see it relate to database type conversion.
block_date_time_utc ?

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
obatirou: let's do that later

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
obatirou: [e533246](https://github.com/zama-ai/fhevm/pull/1386/commits/e533246b7d0f87c103593f2fbdac594feece1550)

rudy-6-4 (approved):

PR #1387, opened 2025-11-25
feat(protocol-contracts): add mainnet lz config for governance

jatZama:


protocol-contracts/governance/layerzero.config.mainnet.ts
```diff
@@ -0,0 +1,49 @@
...
+// Ethereum -> ZamaGatewayMainnet
+
+// We don't use enforced executor gas options here, so we should ensure `options` is set to the correct value through profiling the gas usage of calling GovernanceOAppReceiver._lzReceive(...) on the destination chain.
+// To learn more, read https://docs.layerzero.network/v2/concepts/applications/oapp-standard#execution-options-and-enforced-settings
+
+const pathways: TwoWayConfig[] = [
+    [
+        ethereumContract, // Chain A contract
+        zamaMainnetContract, // Chain B contract
+        // TODO: Add custom ZAMA DVN in next line?
```
melanciani, line 26: is this todo relevant to keep ?

jatZama: Yes, it could be done next year, and I think it would be better for security.

melanciani (approved):
looks good to me !

PR #1388, opened 2025-11-25
feat(coprocessor): send sync and nonce mngmt fixes in txn-sender

dartdart26:
* Use `eth_sendRawTransactionSync` for sending txns to the GW
* Do not consume a nonce if gas estimation fails due to a revert
* Default `error-sleep-max-secs` from 300 to 4 seconds
* Deprecate `required-txn-confirmations` config param
* Use `txn-receipt-timeout-secs` conf param as send sync timeout and set
default from 10 to 4 seconds
* Update alloy version to 1.1.2 for `Provider::send_transaction_sync`
* Deprecate `retry-immediately-on-nonce-error`

Above should fix issues that were leading to constant retries due to a
txn that gets reverted, leading to a consumed nonce. Furthermore, using
send sync should decrease latency and load.

For now, we keep the strategy of redoing the whole batch on failure. We
might change it in a future commit in order to keep this one simpler.


coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
```diff
@@ -59,6 +67,97 @@ impl<P: alloy::providers::Provider<Ethereum> + Clone + 'static> NonceManagedProv
...
+        .await
+        .map_err(|_| TransportErrorKind::custom_str("eth_sendRawTransactionSync timeout"))
+        .flatten();
+
+        if res.is_err() {
+            // Reset the nonce manager if the transaction sending failed.
+            *self.nonce_manager.lock().await = Default::default();
+        }
+        res
+    }
```
eudelins-zama, line 90: As I tried to explain during our sync, [send_transaction_sync](https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.send_transaction_sync) is now supported in the `Provider` trait (was not the case when I wrote this piece of code).

Up to you to use it or not, but it would avoid adding the `F` generic for the fillers everywhere and so on. I'm currently working on it on the connector side and could ping you once ready.

dartdart26: Oh, I haven't realized that. I guess I'd better rewrite it this way. Thanks!

charts/coprocessor/values.yaml
```diff
@@ -739,14 +739,13 @@ txSender:
     - --verify-proof-resp-batch-limit=128
     - --verify-proof-resp-max-retries=6
     - --verify-proof-remove-after-max-retries
-    - --add-ciphertexts-batch-limit=10
-    - --allow-handle-batch-limit=10
+    - --add-ciphertexts-batch-limit=100
+    - --allow-handle-batch-limit=100
```
antoniupop: Has this been verified not to pose problems now? Just wondering if we should test it before increasing.

dartdart26: It hasn't been verified. Let's change it in another commit then.

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
dartdart26: Done.

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
```diff
@@ -59,6 +63,84 @@ impl<P: alloy::providers::Provider<Ethereum> + Clone + 'static> NonceManagedProv
...
+            tx.nonce = Some(nonce);
+        }
+
+        let res = tokio::time::timeout(timeout, self.provider.send_transaction_sync(tx))
+            .await
+            .map_err(|_| TransportErrorKind::custom_str("eth_sendRawTransactionSync timeout"))
+            .flatten();
+
+        if res.is_err() {
+            // Reset the nonce manager if the transaction sending failed.
```
nboisde, line 86: This is where i see the biggest flaw in the  implementation. 
Because in case of error, you are reseting the nonce automatically. This is reducing the capacity of parallelism and introduce a lot of nonce too high errors. 
Could be resolved in a next PR, by using a different nonce management strategy.

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
nboisde: Thanks for the tip @eudelins-zama !

nboisde (dismissed):
This is a great improvement and could improve the overall tps with the introduction of eth_sendRawTransactionSync, see a flaw nonetheless that could be fixed in a next improvement (see the comment).

coprocessor/fhevm-engine/transaction-sender/src/config.rs
```diff
@@ -51,13 +49,12 @@ impl Default for ConfigSettings {
             verify_proof_remove_after_max_retries: true,
             db_polling_interval_secs: 5,
             error_sleep_initial_secs: 1,
-            error_sleep_max_secs: 300,
+            error_sleep_max_secs: 4,
```
rudy-6-4, line 59: do we still need exponential backo off for sleep if this conf is working well (to be seen later and for a next PR)

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
```diff
@@ -59,6 +63,84 @@ impl<P: alloy::providers::Provider<Ethereum> + Clone + 'static> NonceManagedProv
...
+
+    // If `txn_request.gas` is set, overprovision it by the given percent.
+    // If `txn_request.gas` is not set, estimate the gas limit and then overprovision it by the given percent.
+    // If the percent is less than 100, an error is returned.
+    pub async fn overprovision_gas_limit(
+        &self,
+        txn_request: impl Into<TransactionRequest>,
+        percent: u32,
+    ) -> TransportResult<TransactionRequest> {
+        if percent < 100 {
```
rudy-6-4: Should be done on top level ? it's too late here for an error ? hre could be capped to 100

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
```diff
@@ -297,8 +298,7 @@ async fn main() -> anyhow::Result<()> {
         add_ciphertexts_max_retries: conf.add_ciphertexts_max_retries,
         allow_handle_batch_limit: conf.allow_handle_batch_limit,
         allow_handle_max_retries: conf.allow_handle_max_retries,
-        txn_receipt_timeout_secs: conf.txn_receipt_timeout_secs,
-        required_txn_confirmations: conf.required_txn_confirmations,
+        send_txn_sync_timeout_secs: conf.send_txn_sync_timeout_secs,
```
rudy-6-4, line 384: check gas_limit_overprovision_percent  line 308

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
eudelins-zama: FYI, I use it in this [kms-connector's PR](https://github.com/zama-ai/fhevm/pull/1112), if it can help :innocent:

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
dartdart26: I agree. In my mind, there should not be errors in the happy case. We can change it in a future commit, but for the system to be fast, maybe we should also focus on the errors we get from the GW and try to understand how to make them less frequent, e.g. HTTP errors. That would also improve a future implementation further.

But basically, if there are no errors, we should get good speed out of this approach for the time being.

I also want to limit changes in this PR.

coprocessor/fhevm-engine/transaction-sender/src/config.rs
dartdart26: I guess maybe we don't need it for now. But if we send to a GW that has very rare errors, maybe it has some benefit.

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
dartdart26: Good point, let me think how to avoid it.

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
dartdart26: Didn't get what you mean here.

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
dartdart26: I've already done it when you first mentioned it. I am closing this here.

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
dartdart26: I've made it default to 120 if set wrongly.

charts/coprocessor/values.yaml
dartdart26: Reverted to 10.

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
rudy-6-4: You could check gas_limit_overprovision_percent at line 308 once. So the watning that will be seen on all call is seen here at start as an error maybe.

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
```diff
@@ -1,103 +1,145 @@
...
+    /// If `txn_request.gas` is set, overprovision it by the given percent.
+    /// If `txn_request.gas` is not set, estimate the gas limit and then overprovision it by the given percent.
+    /// If the percent is less than 100, DEFAULT_GAS_LIMIT_OVERPROVISION_PERCENT is used.
+    pub async fn overprovision_gas_limit(
+        &self,
+        txn_request: impl Into<TransactionRequest>,
+        percent: u32,
+    ) -> TransportResult<TransactionRequest> {
+        let percent = if percent < 100 {
+            warn!(
```
rudy-6-4, line 101: ok thx it's better

rudy-6-4 (approved):

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
rudy-6-4: I think removing allmost all logical contract error (could still have logical one due to concurrency) during gaz provision will help a lot.

PR #1389, opened 2025-11-25
ci(test-suite): run random tests in all pull requests

melanciani:


.github/workflows/test-suite-e2e-operators-tests.yml
```diff
@@ -120,11 +123,14 @@ jobs:
...
+      # Operator tests can take several hours to complete, so we only run them when triggered manually
       - name: All operators tests
+        if: github.event_name == 'workflow_dispatch'
         working-directory: test-suite/fhevm
         run: |
           ./fhevm-cli test operators
 
+      # Random tests are quick and can be run at each PR
       - name: Random operators tests
         working-directory: test-suite/fhevm
```
eudelins-zama: Don't you want to move this to the `test-suite-e2e-tests.yml` workflow instead?

eudelins-zama: Feels weird to filter which test to run in the `test-suite-e2e-operators-tests.yml` based on `pull_request` vs `workflow_dispatch` IMO

melanciani: @jatZama mentioned that they should stay along operator tests (probably for clarity)

also keeping them here makes it a bit parallelized to avoid making e2e tests longer, if we add them in regular e2e tests, it will make them 8 min longer

but for that I'll need to add `check-changes`

alternative would be to pick one or 2 and run them in regular e2e tests instead, let's discuss this once they are fixed

melanciani: in the end I've put a subset of random tests in regular CI instead as adding `check-changes` was annoying to do

melanciani:
let's first do in main and then cherry picks the 2 commits from https://github.com/zama-ai/fhevm/pull/1391 and https://github.com/zama-ai/fhevm/pull/1390

PR #1390, opened 2025-11-26
fix(test-suite): relayer-sdk v0.3.0-6 and Rand

0xalexbel:
- Tests are now running relayer-sdk 0.3.0-6
- Fix Rand tests
- Fix fhevmOperationsXX tests
- updated codegen to support new publicDecrypt return values

refs https://github.com/zama-ai/fhevm-internal/issues/674


jatZama (commented):
Since you had to change the testgen, I don't understand why you did. not also update code generated tests in solidity library and host?

jatZama (approved):

PR #1397, opened 2025-11-26
chore(protocol-contracts): add staking tasks

melanciani:
closes https://github.com/zama-ai/fhevm-internal/issues/621


protocol-contracts/staking/.env.example
```diff
@@ -0,0 +1,77 @@
...
+ZAMA_TOKEN_ADDRESS=0x0000000000000000000000000000000000000001
+
+# Address of the governance DAO
+DAO_ADDRESS=0x70997970C51812dc3A010C7d01b50e0d17dc79C8 # accounts[1] (address)
+
+# ----------------------------------------------------------------------------
+# Coprocessor
+# ----------------------------------------------------------------------------
+PROTOCOL_STAKING_COPRO_TOKEN_NAME=ProtocolStakingCoprocessor
+PROTOCOL_STAKING_COPRO_TOKEN_SYMBOL=stZAMAfhe
```
mortendahl: There were some comments about naming, see this thread: https://zama-ai.slack.com/archives/C0320EL7Y4U/p1764001564308419

Long story short, I think we are going with stZamaCoprocessors and stZamaKMS

protocol-contracts/staking/tasks/registerEligibleOperators.ts
```diff
@@ -0,0 +1,101 @@
...
+import { task, types } from 'hardhat/config';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+
+// Register an operator as eligible in the protocol staking contract
+// Example usage:
+// npx hardhat task:registerOperatorInProtocolStaking \
+//   --operatorStakingAddress 0x1234567890123456789012345678901234567890 \
+//   --protocolStakingProxyAddress 0x1234567890123456789012345678901234567890 \
+//   --network ethereum-testnet
+task('task:registerOperatorInProtocolStaking')
```
mortendahl: Maybe it's just naming, but this seems more like making operator eligible than "registering" them (whatever that means)

mortendahl (commented):
Nice work!

protocol-contracts/staking/.env.example
jatZama: a this is the symbol, not the name, so it makes more sense to use capital letters at least for the ZAMA part: `stZAMAcoprocessors` and `stZAMAkms` wdyt?

protocol-contracts/staking/.env.example
```diff
@@ -0,0 +1,88 @@
...
+# Address of the underlying staking ERC20
+ZAMA_TOKEN_ADDRESS=0x0000000000000000000000000000000000000001
+
+# Address of the governance DAO
+DAO_ADDRESS=0x70997970C51812dc3A010C7d01b50e0d17dc79C8 # accounts[1] (address)
+
+# ----------------------------------------------------------------------------
+# Coprocessor
+# ----------------------------------------------------------------------------
+PROTOCOL_STAKING_COPRO_TOKEN_NAME=ProtocolStakingCoprocessor
```
jatZama: `ProtocolStakingCoprocessors` with an s

protocol-contracts/staking/.env.example
```diff
@@ -0,0 +1,88 @@
...
+# Cooldown period in seconds before unstaked tokens can be released
+PROTOCOL_STAKING_KMS_COOLDOWN_PERIOD=604800 # 7 days
+
+# ----------------------------------------------------------------------------
+# OperatorStaking deployment
+# ----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
+# Coprocessor
+# ----------------------------------------------------------------------------
+# The number of OperatorStaking for coprocessors must be lower or equal to the number of addresses defined below
```
jatZama: why lower? maybe best to impose equality to avoid mistakes. wdyt?

protocol-contracts/staking/contracts/ProtocolStaking.sol
```diff
@@ -19,7 +19,7 @@ interface IERC20Mintable is IERC20 {
 
 /**
  * @dev Staking contract that distributes newly minted tokens to eligible accounts at a configurable flow rate.
- *
```
jatZama: can you cancel this change for the moment? even if it does nothing, for the moment better not touch any contract, to not disturb OZ team?

protocol-contracts/staking/tasks/accounts.ts
```diff
@@ -0,0 +1,39 @@
+import { task } from 'hardhat/config';
+import { HardhatNetworkHDAccountsConfig } from 'hardhat/types';
+
+const NUM_ACCOUNTS = 10;
+
+// Use this task to get the list of accounts (addresses, private keys, public keys)
+// Example usage:
+// npx hardhat get-accounts
+task('get-accounts', 'Prints the list of accounts').setAction(async (_, { ethers, network }) => {
```
jatZama, line 9: Do you really need this task?

protocol-contracts/staking/.env.example
melanciani: just to be clear: these are default testing values, we'll provide real ones when we deploy them

but yes I can better align them to avoid confusion

protocol-contracts/staking/tasks/registerEligibleOperators.ts
melanciani: yes good point I will do that

protocol-contracts/staking/.env.example
melanciani: we've always written like this copy-pasted from gateway-contracts), but yes we could say that indeed

protocol-contracts/staking/contracts/ProtocolStaking.sol
melanciani: this is prettier, I missed it indeed, will revert

protocol-contracts/staking/tasks/accounts.ts
melanciani: it's useful to easily get test addresses yes, I've already put it in other protocol-contracts repo. It's not a big deal to leave it no ?

jatZama: ah ok now i understand.

protocol-contracts/staking/contracts/ProtocolStaking.sol
jatZama: i dont have strong opinion here, maybe it is fine, i am not sure.

protocol-contracts/staking/tasks/verify/verifyOperatorStaking.ts
```diff
@@ -0,0 +1,36 @@
...
+      address: operatorStakingAddress,
+      constructorArguments: [],
+    });
+  }
+});
+
+// Verify all operator rewarder contracts
+// Example usage:
+// npx hardhat task:verifyAllOperatorRewarderContracts --network ethereum-testnet
+task('task:verifyAllOperatorRewarderContracts').setAction(async function (_, hre) {
```
jatZama: Actually since they will all follow same implementation, they will have same bytecode, since there is no implementation specific constants, so this is not needed. Just verifying one contract would work.

jatZama: Verifying a single instance of each contract (ProtocolStaking, OperatorStaking and OperatorRewarder) should be enough.

protocol-contracts/staking/hardhat.config.ts
```diff
@@ -1,15 +1,48 @@
+import './tasks/accounts';
+import './tasks/deployment/';
+import './tasks/ownership/';
+import './tasks/registerEligibleOperators';
+import './tasks/setRewardRate';
+import './tasks/verify';
```
jatZama, line 8: Why there is no ending `/` for this one but you use slash at the end of 
import './tasks/deployment/';
import './tasks/ownership/'; ?

protocol-contracts/staking/tasks/ownership/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './protocolStakingManagerRole';
+export * from './protocolStakingGovernorRole';
```
jatZama, line 2: Please add as well a task for transferring ownership for OperatorStaking. This might be needed if for instance we want to set the initial owner fees on behalf on an operator (so deployer will be initial owner of OperatorStaking) then we transfer ownership to the operator.

protocol-contracts/staking/tasks/ownership/protocolStakingGovernorRole.ts
```diff
@@ -0,0 +1,56 @@
+import { PROTOCOL_STAKING_CONTRACT_NAME } from '../deployment';
+import { getProtocolStakingCoproProxyAddress, getProtocolStakingKMSProxyAddress } from '../utils/getAddresses';
+import { getRequiredEnvVar } from '../utils/loadVariables';
+import { task } from 'hardhat/config';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+
+// Being a transfer of the governor role to the DAO for a protocol staking contract using the deployer account
```
jatZama: typo : "Being" should be replaced by "Begin"

protocol-contracts/staking/tasks/ownership/protocolStakingGovernorRole.ts
```diff
@@ -0,0 +1,56 @@
...
+  const DAO_ADDRESS = getRequiredEnvVar('DAO_ADDRESS');
+
+  // Begin the transfer of the governor role to the DAO
+  await protocolStaking.beginDefaultAdminTransfer(DAO_ADDRESS);
+
+  log(`Governor role of protocol staking contract at address ${protocolStakingProxyAddress} 
+      being transferred to DAO address ${DAO_ADDRESS} on network ${network.name}`);
+}
+
+// Being a transfer of the governor role to the DAO for the protocol staking contracts
```
jatZama: same typo here as well

protocol-contracts/staking/tasks/ownership/protocolStakingGovernorRole.ts
```diff
@@ -0,0 +1,56 @@
...
+// This is te first step of a 2-step process to transfer the governor role to the DAO
+// The DAO then needs to accept the transfer by calling `acceptDefaultAdminTransfer()` function on
+// both protocol staking contracts
+// Example usage:
+// npx hardhat task:beingTransferProtocolStakingGovernorRolesToDAO --network ethereum-testnet
```
jatZama: same typo here as well bis

protocol-contracts/staking/tasks/ownership/protocolStakingGovernorRole.ts
```diff
@@ -0,0 +1,56 @@
...
+task('task:beingTransferProtocolStakingGovernorRolesToDAO').setAction(async function (
```
jatZama: same typo here as well - tris

protocol-contracts/staking/tasks/ownership/protocolStakingGovernorRole.ts
```diff
@@ -0,0 +1,56 @@
...
+  _,
+  hre: HardhatRuntimeEnvironment,
+) {
+  const { log } = hre.deployments;
+
+  log('Being a transfer of the governor role to the DAO for the coprocessor protocol staking contract...');
```
jatZama: here also

protocol-contracts/staking/tasks/ownership/protocolStakingGovernorRole.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const protocolStakingCoproProxyAddress = await getProtocolStakingCoproProxyAddress(hre);
+
+  transferGovernorRole(protocolStakingCoproProxyAddress, hre);
+
+  log('Being a transfer of the governor role to the DAO for the KMS protocol staking contract...');
```
jatZama: and here

jatZama (commented):
Nice work, I finished my review, but I put several comments.

protocol-contracts/staking/tasks/verify/verifyOperatorStaking.ts
melanciani: ok great good catch thanks

protocol-contracts/staking/.env.example
mortendahl: > `stZAMAcoprocessors` and `stZAMAkms`

Yes I agree

protocol-contracts/staking/hardhat.config.ts
melanciani: it's because these are directories, but / are not needed I'll remove them

protocol-contracts/staking/tasks/ownership/index.ts
melanciani: well it won't be needed anyway of we introduce the ownerFee in the constructor, but yes better to have it just in case I'll add it

jatZama: you could remove it later, once you add this feature. I was saying this just in case, and to do initial deployment for corner3 asap, because OZ seems to not plan to do this anytime soon.

jatZama (dismissed):

protocol-contracts/token/package.json
```diff
@@ -66,7 +66,9 @@
     "solhint": "^4.1.1",
     "solidity-bytes-utils": "^0.8.2",
     "ts-node": "^10.9.2",
-    "typescript": "^5.4.4"
+    "typescript": "^5.4.4",
+    "@safe-global/safe-core-sdk-types": "npm:@safe-global/types-kit@^1.0.0",
```
melanciani: I couldn't run the deploy task without this

also, i needed to npm install force as `npm i` fails for incompatible deps

jatZama: Please don't modify package stuff. Did you use pnpm (not npm) and follow Readme instructions (point 4) to deploy? Should work this way.

melanciani: ah true we are using `pnpm` in this specific directory .. will revert yes

protocol-contracts/staking/tasks/addEligibleAccount.ts
```diff
@@ -75,7 +75,7 @@ task('task:addAllCoproOperatorsAsEligible').setAction(async function (_, hre: Ha
 
     if (i < operatorStakingAddresses.length - 1) {
       // Wait for 5 seconds before adding the next operator as eligible in order to avoid underpriced transaction issues
-      await wait(5);
+      await wait(5, hre.network.name);
```
jatZama: this does not seem very robust, couldn't you just add a `await receipt.wait(1)` in the task definition instead?

protocol-contracts/staking/tasks/addEligibleAccount.ts
```diff
@@ -97,7 +97,7 @@ task('task:addAllKMSOperatorsAsEligible').setAction(async function (_, hre: Hard
...
```
jatZama: same here

protocol-contracts/staking/tasks/addEligibleAccount.ts
```diff
@@ -110,7 +110,9 @@ task('task:addAllOperatorsAsEligible').setAction(async function (_, hre: Hardhat
 
   await hre.run('task:addAllCoproOperatorsAsEligible');
 
-  await wait(5);
+  await wait(5, hre.network.name);
```
jatZama: and here

protocol-contracts/staking/hardhat.config.ts
```diff
@@ -10,14 +10,45 @@ import '@nomicfoundation/hardhat-ethers';
...
 import { HardhatUserConfig, HttpNetworkAccountsUserConfig } from 'hardhat/types';
+import { resolve } from 'path';
 import 'solidity-coverage';
 
+// Run the test with environment variables from `.env.example`
+task('test', 'Runs the test suite with environment variables from .env.example')
+  .addOptionalParam('skipSetup', 'Set to true to skip setup tasks', false, types.boolean)
+  .setAction(async (taskArgs, hre, runSuper) => {
+    // Load `.env.example`
+    const envExamplePath = resolve(__dirname, '.env.example');
```
jatZama: Sorry to be nit-picky here, but I am not a huge fan of overriding the default `test` hardhat task, even if it made sense in the specific context of gateway-contracts project, because here remember that this project will also be used by OZ team, and this could have unforeseen side effects.
IMO it would be best to not override the default behavior of test task here, and instead put all the needed deployment steps as a fixture directly inside the hardhat tests related to tasks. This seems a lot cleaner in this case, wdyt?

protocol-contracts/staking/hardhat.config.ts
```diff
@@ -1,7 +1,6 @@
 import './tasks/accounts';
 import './tasks/addEligibleAccount';
 import './tasks/deployment';
-import './tasks/deposit';
```
jatZama: Not sure why you decided to remove this task?

protocol-contracts/staking/tasks/addEligibleAccount.ts
melanciani: yes it's definitively much better, not even sure why i didn't do this in the first place

protocol-contracts/staking/hardhat.config.ts
melanciani: I thought it was not needed for testnet, putting it back

protocol-contracts/staking/hardhat.config.ts
melanciani: yes I see what you mean and I agree. The simplest solution I found was to take these tests out of the test directory and rename the task (because I want to only run these deployment tasks once, else it'll be too verbose)

melanciani: lmk wyt

jatZama (approved):

PR #1405, opened 2025-11-27
test(test-suite): update fhevm versions to 0.10.0

melanciani:


eudelins-zama (commented):
CI is not passing on `main` but work on `release/0.10.x`...

I ran the `git diff main..release/0.10.x  test-suite` command, and there are some differences, notably in the `test-suite/e2e/test/delegateUserDecrypt/delegateUserDecryption.ts` file, maybe cc @rudy-6-4

melanciani:
~~yes @eudelins-zama seems like it has been modified in main recently : https://github.com/zama-ai/fhevm/pull/1092/files~~

~~but it is somewhat expected, we should consider new images if anything has been updated in main but not for 0.10, so best thing to do is keep the test-suite images introduced recently by Alex : https://github.com/zama-ai/fhevm/pull/1390/files#diff-6392bf845e5d28c9e71606db47d7889600ded5ba9775f1dfe7264d305e2a9668R41~~

edit: sorry, looking at it again, I think there's a mistake indeed, as the PRs I showed above are supposed to be the same as the ones included in the 0.10 branch. Let's clarify this

eudelins-zama (approved):

PR #1406, opened 2025-11-27
fix(coprocessor): fix handling of cyclic dependence errors

antoniupop:


rudy-6-4 (approved):

coprocessor/fhevm-engine/scheduler/src/dfg.rs
```diff
@@ -145,14 +151,72 @@ impl DFTxGraph {
...
+                    let prod = self
+                        .graph
+                        .node_weight(*producer)
+                        .ok_or(SchedulerError::DataflowGraphError)?;
+                    let cons = self
+                        .graph
+                        .node_weight(*consumer)
+                        .ok_or(SchedulerError::DataflowGraphError)?;
+                    error!(target: "scheduler", { producer_id = ?hex::encode(prod.transaction_id.clone()), consumer_id = ?hex::encode(cons.transaction_id.clone()) },
+		       "Dependence cycle when adding dependence");
```
rudy-6-4: MAybe it could say the detection of cycle is not complete since it's unexpected to end here ?

antoniupop: Yes, this is guarded against above, but if for some reason the above detection fails, this will return an error and nothing gets processed to be conservative. I'll update the message.

PR #1408, opened 2025-11-28
feat(coprocessor): host-listener --only-catchup-loop mode

obatirou:
Close https://github.com/zama-ai/fhevm-internal/issues/687

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
@@ -154,6 +154,7 @@ struct InfiniteLogIter {
     last_valid_block: Option<u64>,
     start_at_block: Option<i64>,
     end_at_block: Option<i64>,
+    absolute_end_at_block: Option<u64>,
     catchup_margin: u64,
     catchup_paging: u64,
     pub tick_timeout: HeartBeat,
@@ -213,6 +214,7 @@ impl InfiniteLogIter {
             last_valid_block: None,
             start_at_block: args.start_at_block,
             end_at_block: args.end_at_block,
+            absolute_end_at_block: None,
             catchup_paging: args.catchup_paging.max(1),
             catchup_margin: args.catchup_margin,
             tick_timeout: HeartBeat::default(),
@@ -674,9 +676,12 @@ impl InfiniteLogIter {
                 Ok(provider) => {
                     let catch_up_from =
                         self.catchup_block_from(&provider).await;
+                    let end_at_block =
+                        self.resolve_end_at_block(&provider).await;
+                    self.absolute_end_at_block = end_at_block;
                     self.catchup_blocks = Some((
                         catch_up_from.as_number().unwrap_or(0),
-                        self.resolve_end_at_block(&provider).await,
+                        end_at_block,
                     ));
                     // note subscribing to real-time before reading catchup
                     // events to have the minimal gap between the two
@@ -769,7 +774,7 @@ impl InfiniteLogIter {
     }
 
     async fn end_at_block_reached(&self) -> bool {
-        let Some(end_at_block) = self.end_at_block else {
+        let Some(end_at_block) = self.absolute_end_at_block else {
             return false;
         };
         let current_block_number =
@@ -780,7 +785,7 @@ impl InfiniteLogIter {
             } else {
                 return false;
             };
-        current_block_number > end_at_block as u64
+        current_block_number > end_at_block
     }
 
     async fn next(&mut self) -> Option<BlockLogs<Log>> {
@@ -851,6 +856,7 @@ impl InfiniteLogIter {
         self.catchup_blocks = None;
         self.next_blocklogs.clear();
         self.last_valid_block = None;
+        self.absolute_end_at_block = None;
         self.block_history =
             BlockHistory::new(self.reorg_maximum_duration_in_blocks as usize);
     }
@@ -904,47 +910,6 @@ async fn db_insert_block(
         retries -= 1;
         db.reconnect().await;
         tokio::time::sleep(Duration::from_millis(500)).await;
-    }
-}
-
-/// Runs the catchup-only loop mode **forever**.
-///
-/// This function processes blocks from `start_at_block` to `stop_at_block`,
-/// sleeps for `catchup_sleep_secs`, then repeats indefinitely.
-/// It only returns `Err` on connection failures; otherwise it loops forever.
-async fn run_catchup_only_loop(
-    args: &Args,
-    log_iter: &mut InfiniteLogIter,
-    db: &mut Database,
-    chain_id: ChainId,
-    acl_contract_address: &Option<Address>,
-    tfhe_contract_address: &Option<Address>,
-) -> anyhow::Result<()> {
-    info!("Starting catchup-only loop mode");
-
-    loop {
-        log_iter.new_log_stream(true).await;
-
-        // Process blocks until end_at_block is reached
-        while let Some(block_logs) = log_iter.next().await {
-            let _ = db_insert_block(
-                chain_id,
-                db,
-                &block_logs,
-                acl_contract_address,
-                tfhe_contract_address,
-            )
-            .await;
-        }
-
-        info!(
-            sleep_secs = args.catchup_sleep_secs,
-            "Catchup loop iteration complete, sleeping"
-        );
-        tokio::time::sleep(Duration::from_secs(args.catchup_sleep_secs)).await;
-
-        // Reset state for next iteration
-        log_iter.reset_for_catchup_loop();
     }
 }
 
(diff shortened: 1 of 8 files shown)


coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -123,6 +128,21 @@ pub struct Args {
...
+        help = "Run only catchup loop without real-time subscription"
+    )]
+    pub only_catchup_loop: bool,
+
+    #[arg(
+        long,
+        default_value_t = 960u64,
+        help = "Sleep duration in seconds between catchup iterations (only with --only-catchup-loop)"
+    )]
+    pub catchup_sleep_secs: u64,
```
rudy-6-4: default could be none, in this case we sleep the time cover by the blocks ? like you go 1000 blocks in the past, you sleep 1000 blocks less a margin ?
ok maybe too complicated.
960 is too big depending on the start option (80 blocks),
E.g. with --start -40, there are holes
Maybe a straight error like, "catchup_sleep_secs" is too big compared to --start -40.
Also ositive index can maybe rejected with only_catchup_loop true.

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -219,6 +239,21 @@ impl InfiniteLogIter {
         Ok(provider.get_chain_id().await?)
     }
 
+    /// Resolves `end_at_block` to an absolute block number.
+    /// If `end_at_block` is negative, it is interpreted as relative to the current block.
+    async fn resolve_end_at_block(
+        &self,
+        provider: &BlockchainProvider,
+    ) -> Option<u64> {
+        match self.end_at_block? {
```
rudy-6-4: I like patterns but not sure it helps readability here.
```
let n = self.end_at_block?;
if n>= 0
``

File: coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs [hunk 3]
```diff
@@ -812,6 +851,90 @@ impl InfiniteLogIter {
         self.next_blocklogs.push_back(block_logs);
         self.next_blocklogs.pop_front()
     }
+
+    /// Setup catchup-only mode without real-time subscription.
+    /// Connects to the provider and initializes catchup range.
+    async fn setup_catchup_only(&mut self) -> anyhow::Result<()> {
```
rudy-6-4: Not sure why you do that. The initial code probably work out of the box since host-listener is already used to only catchup in the catchup instance.

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -739,6 +774,10 @@ impl InfiniteLogIter {
         let Some(end_at_block) = self.end_at_block else {
             return false;
         };
+        if end_at_block < 0 {
```
rudy-6-4: why ? this is super usefull if you want to do an isolated catchup.
--start -120 --stop -1

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
obatirou: It is to avoid subscribing to real-time blocks as it is not necessary for catchup. But we could use existing function. Subscribing to blocks would be wasteful but would avoid new functions

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
obatirou: Added validation of argument and reduced default value

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -207,6 +227,20 @@ impl InfiniteLogIter {
...
+    async fn resolve_end_at_block(
+        &self,
+        provider: &BlockchainProvider,
+    ) -> Option<u64> {
+        let n = self.end_at_block?;
+        if n >= 0 {
+            return Some(n as u64);
+        }
+        let last_block = provider.get_block_number().await.ok()?;
+        Some(last_block.saturating_sub(n.unsigned_abs()))
```
rudy-6-4, line 250: end_of_block should be updated here to the absolute value ? so that it can be compared later to current block

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -735,7 +769,7 @@ impl InfiniteLogIter {
             } else {
                 return false;
             };
-        current_block_number > end_at_block
+        current_block_number > end_at_block as u64
```
rudy-6-4: This won't work with end_of_block = -1 = 2*64 - 1

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
```diff
@@ -522,6 +561,139 @@ async fn test_catchup_only() -> Result<(), anyhow::Error> {
...
+    Ok(())
+}
+
+#[tokio::test]
+#[serial(db)]
+async fn test_catchup_only_relative_end() -> Result<(), anyhow::Error> {
+    let nb_event_per_wallet = 5;
+    let outcome = run_catchup_only_scenario(nb_event_per_wallet, 15, |args| {
+        args.start_at_block = Some(-50); // 50 blocks from current
+        args.end_at_block = Some(-5); // 5 blocks from current (more recent)
```
rudy-6-4, line 663: how does it succeed with negative end_at_block ? is it looping or end was not detected ?

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
rudy-6-4: could be in absolute_end_of_block to avoid having to put back the initial negative value

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
rudy-6-4: use self.absolute_end_of_block

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -853,10 +896,74 @@ async fn db_insert_block(
...
+            let _ = db_insert_block(
+                chain_id,
+                db,
+                &block_logs,
+                acl_contract_address,
+                tfhe_contract_address,
+            )
+            .await;
+        }
+
```
rudy-6-4: this duplicates the main body. add a if !catchup loop break here and it can be fused with the main function.

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -949,6 +1056,20 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
...
+            &acl_contract_address,
+            &tfhe_contract_address,
+        )
+        .await;
+    }
+
+    // Normal mode: real-time listener with catchup
     log_iter.new_log_stream(true).await;
 
     while let Some(block_logs) = log_iter.next().await {
```
rudy-6-4: the loop from run_catchup_only_loop could be done here with just a "if !catchup break"

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
rudy-6-4: maybe it should check that it's doing a new catchup aftert the first one

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
obatirou: good catch

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
obatirou: it is looping as end is not detected due to `end_at_block as u64` will be a huge number
Also note that the end_at_block is per iteration

obatirou: added a second iteration catchup

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -764,10 +803,16 @@ impl InfiniteLogIter {
                 return self.next_blocklogs.pop_front();
             };
             if self.end_at_block_reached().await {
-                eprintln!(
-                    "End at block reached: {}",
-                    self.end_at_block.unwrap()
-                );
+                let absolute = self.absolute_end_at_block.unwrap();
```
rudy-6-4: The unwrap are perfectly safe here but it could be made explicit.
```
match self.end_at_block {
     Some(index) if index < 0 =>
           eprintln!()
```
But maybe you can just compare them and print them as-is.
```
if self.end_at_block() == self.absolute_end_at_block() {
    ...
} else {
     eprintln!("End at block reached: {}", self.absolute_end_at_block());
}
```
Either way or do nothing, is ok.

rudy-6-4 (dismissed):

rudy-6-4 (dismissed):

piizama (approved):

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -764,10 +803,16 @@ impl InfiniteLogIter {
                 return self.next_blocklogs.pop_front();
             };
             if self.end_at_block_reached().await {
-                eprintln!(
-                    "End at block reached: {}",
-                    self.end_at_block.unwrap()
-                );
+                match self.end_at_block {
+                    Some(n) if n < 0 => eprintln!(
```
rudy-6-4, line 807: Maybe having the messages could be in logs so we can inspect them in case the catchup mode does not stop properly.

rudy-6-4 (approved):

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
obatirou: Let's do that in a new issue.
We should release asap to test

PR #1433, opened 2025-12-02
chore(common): extension recommandation

ClementWalter:


coprocessor/.gitignore
```diff
@@ -14,7 +14,6 @@ dist/
 
 # Editor directories and files
 .idea/
-.vscode/
```
melanciani, line 17: should we keep `.vscode/` and only allow `.vscode/extensions.json` to avoid having other vscode files commited by mistake in the future ?

ClementWalter: afaik every file in this folder should be committed
image

melanciani: well I don't think I agree with this, there could be some security issues + most of these files are for user settings that don't make much sense to share between all team members. Also, not all people use vscode, so let's not start adding all settings from all IDE 

I asked claude to confirm this : 
Screenshot 2025-12-03 at 10 42 50

ClementWalter: perplexity is wrong

- it's not because there is a .vscode that anyone need to use vscode
- personal preferences don't go into the workspace setting but user settings
- doesn't include package-lock specific info, any update in settings in this folder should be shared
- "false sense" is quite a weak argument

melanciani: overall I disagree as we should probably not enforce settings even at the workspace level (other repos like tfhe-rs decided to avoid this for example), but more generally I don't see the point of allowing new .vscode config files in the repo if there is no proper reason for doing it for now. The extension.json makes a lot of sense, the rest I doubt it for now, so let's only allow new config files in the future if we ever need to, but forbid doing it by mistake until then !

PR #1437, opened 2025-12-02
feat(coprocessor): add ZK proof gauges to txn-sender

dartdart26:
These cover both txn sending by `transaction-sender` and `zkproof-worker` operation.

Also make `--host-chain-url` a long-only argument for compatibility with `-h` for help.

### DIFF
--- a/coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
+++ b/coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
@@ -50,7 +50,7 @@ struct Conf {
     #[arg(short, long)]
     gateway_url: Url,
 
-    #[arg(short, long)]
+    #[arg(long)]
     host_chain_url: Url,
 
     #[arg(short, long, value_enum, default_value = "private-key")]
--- a/coprocessor/fhevm-engine/transaction-sender/src/metrics.rs
+++ b/coprocessor/fhevm-engine/transaction-sender/src/metrics.rs
@@ -94,6 +94,22 @@ pub(crate) static ADD_CIPHERTEXT_MATERIAL_UNSENT: LazyLock<IntGauge> = LazyLock:
     .unwrap()
 });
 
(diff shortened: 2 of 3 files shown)


coprocessor/fhevm-engine/transaction-sender/src/metrics.rs
```diff
@@ -127,6 +143,32 @@ pub fn spawn_gauge_update_routine(period: std::time::Duration, db_pool: PgPool)
...
+                Ok(count) => {
+                    info!(verify_proof_pending = %count, "Fetched pending verify proofs count");
+                    VERIFY_PROOF_PENDING.set(count);
+                }
+                Err(e) => {
+                    error!(error = %e, "Failed to fetch pending verify proofs count");
+                }
+            }
+
+            match sqlx::query_scalar("SELECT COUNT(*) FROM verify_proofs WHERE verified = TRUE")
```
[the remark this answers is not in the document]
dartdart26: Wow, good catch!

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
```diff
@@ -50,7 +50,7 @@ struct Conf {
     #[arg(short, long)]
     gateway_url: Url,
 
-    #[arg(short, long)]
+    #[arg(long)]
```
rudy-6-4, line 53: are you sure it's not used ?

rudy-6-4 (approved):

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
dartdart26: I think the full one is used everywhere, also in the values file. Not 100% sure if someone else is not using the short one though.

If I keep the short one, it doesn't even start on my end, complaining about `-h`.

@tawadaa Could you please check if only `--host-chain-url` is used and not `-h`?

dartdart26:
@Mergifyio queue

PR #1438, opened 2025-12-02
fix(coprocessor): use block hash to identify transaction

rudy-6-4:


coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -336,7 +351,8 @@ SELECT
   c.is_scalar,
   c.is_allowed, 
   c.dependence_chain_id,
-  c.transaction_id
+  c.transaction_id,
+  c.block_hash
 FROM computations c
 JOIN selected_computations sc
   ON  c.transaction_id = sc.transaction_id
```
rudy-6-4, line 372: missing block_hash

rudy-6-4:
add a test with duplicate transaction id that forms a transaction cycle without the block hash

coprocessor/fhevm-engine/db-migration/migrations/20251202093917_computations_block_hash.sql
obatirou, line 1: If not mistaken the primary key is (tenant_id, output_handle, transaction_id) for computation.
Is it ok to not included the block hash? Can output_handle change during a reorg?

obatirou: My understanding is yes: handle represent a computation, in case of reorg, either it is different and be inserted or handles are equals and be drop due to duplication

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -292,9 +294,11 @@ impl Database {
...
                 is_allowed,
                 created_at,
-                schedule_order
+                schedule_order,
+                block_hash,
+                block_number
             )
-            VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $9)
+            VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $9, $10, $11)
             ON CONFLICT (tenant_id, output_handle, transaction_id) DO NOTHING
```
rudy-6-4: add block_hash here

coprocessor/fhevm-engine/tfhe-worker/src/server.rs
```diff
@@ -696,9 +696,11 @@ impl CoprocessorService {
                         is_scalar,
                         dependence_chain_id,
                         transaction_id,
-                        is_allowed
+                        is_allowed,
+                        block_number,
+                        block_hash
```
obatirou, line 701: seems it is inverted compared to the insert in coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Also no created_at or schedule_order are added?

obatirou: from my understanding, I suppose this is ok as it is and endpoint that receives computation requests from external clients instead of host listener but not sure to understand how the tfhe worker will order those if there is no schedule_order. The default value will be inserting time and if I recall correctly this is problematic.

Is this function actually used?

coprocessor/fhevm-engine/db-migration/migrations/20251202093917_computations_block_hash.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE computations
+  ADD COLUMN IF NOT EXISTS block_hash BYTEA NOT NULL DEFAULT '\x00'::BYTEA,
+  ADD COLUMN IF NOT EXISTS block_number BIGINT NOT NULL DEFAULT 0;
```
obatirou, line 3: why have the block_number?
I do not see any usage in the current PR code

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -303,9 +318,9 @@ async fn query_for_work<'a>(
 WITH selected_computations AS (
   (
     SELECT DISTINCT
-      c_creation_order.transaction_id
+      c_creation_order.transaction_id, c_creation_order.block_hash
```
obatirou, line 335: should there be a condition that check if this block was reorg?
shouldn't the host listener add this information to the computation table?

It will allow to ensure those computation are not picked up

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -189,10 +192,19 @@ impl<'a> Scheduler<'a> {
...
                 set.spawn_blocking(move || execute_partition(args, index, 0, sks, cpk, loop_ctx));
+            } else {
+                info!("task not ready");
             }
         }
+        if set.is_empty() {
+            warn!(target: "scheduler", "No ready tasks to schedule - possible cyclic dependence in graph");
+            return Ok(());
+        }
+        info!("wait result {}", set.len());
```
rudy-6-4, line 205: ici j'ai un set de taille 4

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -189,10 +192,19 @@ impl<'a> Scheduler<'a> {
...
         while let Some(result) = set.join_next().await {
+            info!("has result {}", result.is_ok());
```
rudy-6-4, line 207: au bout d'un tres long moment j'ai un "has result true"

coprocessor/fhevm-engine/tfhe-worker/src/tests/errors.rs
```diff
@@ -674,3 +696,249 @@ async fn test_coprocessor_computation_errors() -> Result<(), Box<dyn std::error:
 
     Ok(())
 }
+
+
+#[tokio::test]
+async fn test_coprocessor_reorb_tx_id_collision_errors() -> Result<(), Box<dyn std::error::Error>> {
```
rudy-6-4, line 702: Test qui ne marche pas

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
rudy-6-4: the info is not available in db

obatirou: should this be added?
update: not sure this is useful for the small number of case we will have.

coprocessor/fhevm-engine/db-migration/migrations/20251202093917_computations_block_hash.sql
rudy-6-4: no it's for easier debug, it's informational only

PR #1450, opened 2025-12-03
fix(coprocessor): add missing indices for ciphertext_digest table

dartdart26:
* `idx_ciphertext_digest_handle` - we want the txn-sender to select by
   handle only
 * `idx_ciphertext_digest_unsent` - we want the txn-sender to
   select unsent txns with ordered limited retries efficiently

Furthermore, add fair ordering of rows in the ciphertext_digest table by
tracking when entries are created and order by that in the txn-sender.

Also, update the `svm-rs-builds` dependency to fix an issue with it
failing to fetch 0.8.31 prerelease version during build time from the
Soliditylang JSON file.

### DIFF
--- a/coprocessor/fhevm-engine/db-migration/migrations/20251203140023_ciphertext_digest_idx_sent_and_handle.sql
+++ b/coprocessor/fhevm-engine/db-migration/migrations/20251203140023_ciphertext_digest_idx_sent_and_handle.sql
@@ -1,0 +1,11 @@
+-- Track when entries are created for fair queuing of unsent transactions.
+ALTER TABLE ciphertext_digest
+ADD COLUMN IF NOT EXISTS created_at TIMESTAMP NOT NULL DEFAULT NOW();
+
+-- Handle SELECTs on handle only by the txn-sender.
+CREATE INDEX IF NOT EXISTS idx_ciphertext_digest_handle
+ON ciphertext_digest (handle);
+
+-- Handle SELECTs on unsent txns with limited retries by the txn-sender.
+CREATE INDEX IF NOT EXISTS idx_ciphertext_digest_unsent
+ON ciphertext_digest (txn_is_sent, created_at);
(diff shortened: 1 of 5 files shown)


antoniupop (dismissed):

coprocessor/fhevm-engine/db-migration/migrations/20251203140023_ciphertext_digest_idx_sent_and_handle.sql
```diff
@@ -0,0 +1,11 @@
+-- Track when entries are created for fair queuing of unsent transactions.
+ALTER TABLE ciphertext_digest
+ADD COLUMN IF NOT EXISTS created_at TIMESTAMP NOT NULL DEFAULT NOW();
+
+-- Handle SELECTs on handle only by the txn-sender.
+CREATE INDEX IF NOT EXISTS idx_ciphertext_digest_handle
+ON ciphertext_digest (handle);
```
antoniupop, line 7: Do you think we'd want to restrict index to `txn_is_sent=false` and filter in teh query?

coprocessor/fhevm-engine/db-migration/migrations/20251203140023_ciphertext_digest_idx_sent_and_handle.sql
```diff
@@ -0,0 +1,11 @@
...
+
+-- Handle SELECTs on unsent txns with limited retries by the txn-sender.
+CREATE INDEX IF NOT EXISTS idx_ciphertext_digest_unsent
+ON ciphertext_digest (txn_is_sent, created_at);
```
antoniupop, line 11: Same here, we could have index on creted_at, but only `where txn_is_sent=false`

dartdart26: The thing is that there is a query in fhevm-exporter that uses = true still.

coprocessor/fhevm-engine/db-migration/migrations/20251203140023_ciphertext_digest_idx_sent_and_handle.sql
dartdart26: The index is for when inserting WHERE handle = X.

antoniupop (approved):

coprocessor/fhevm-engine/db-migration/migrations/20251203140023_ciphertext_digest_idx_sent_and_handle.sql
dartdart26: My comment isn't good actually, yeah, sorry. I will merge as is for now to unblock CI.

PR #1459, opened 2025-12-04
ci(library-solidity): merge hardhat and foundry tests

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/720


.github/workflows/library-solidity-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: library-solidity-forge-tests
+name: library-solidity-hardhat-tests
```
melanciani: maybe `library-solidity-tests` no ?

eudelins-zama: Nice catch !
Fixed

melanciani (approved):
thanks !

PR #1465, opened 2025-12-04
feat(coprocessor): host listener poller port to 0.10

obatirou:


charts/coprocessor/values.yaml
```diff
@@ -214,6 +214,91 @@ hostListener:
...
+
+  # Command line arguments for the host listener poller
+  args:
+    - --database-url=$(DATABASE_URL)
+    - --coprocessor-api-key=$(TENANT_API_KEY)
+    - --url=$(ETHEREUM_RPC_HTTP_URL)
+    - --acl-contract-address=$(ACL_CONTRACT_ADDRESS)
+    - --tfhe-contract-address=$(FHEVM_EXECUTOR_CONTRACT_ADDRESS)
+    - --log-level=INFO
+    - --health-port=8080
```
rudy-6-4, line 263: Not done in previous review but we could have all param here, even the optional.
Particulartly the one about reliability or observability, so infra can change them easily.
E.g. metrics-addr, service-name, retry, delay ...
```
#[arg(
        long,
        default_value_t = 100,
        help = "Maximum number of blocks to process per iteration"
    )]
    batch_size: u64,

    #[arg(
        long,
        default_value_t = 1000,
        help = "Sleep duration between iterations in milliseconds"
    )]
    poll_interval_ms: u64,

    #[arg(
        long,
        default_value_t = 1000,
        help = "Backoff between retry attempts for RPC/DB failures in milliseconds"
    )]
    retry_interval_ms: u64,

    #[arg(
        long,
        default_value_t = 45,
        help = "Maximum number of HTTP/RPC retry attempts (in addition to the initial attempt) before failing an operation"
    )]
    max_http_retries: u32,

    #[arg(
        long,
        default_value_t = 1000,
        help = "Rate limiting budget for RPC calls during block catchup (compute units per second). Higher values = less throttling"
    )]
    rpc_compute_units_per_second: u64,

    #[arg(
        long,
        help = "Address for Prometheus metrics HTTP server (e.g. 0.0.0.0:9100); if unset, metrics server is disabled"
    )]
    metrics_addr: Option<String>,

    #[arg(long, default_value_t = 8080, help = "Health check port")]
    health_port: u16,

    #[arg(
        long,
        value_parser = clap::value_parser!(Level),
        default_value_t = Level::INFO
    )]
    log_level: Level,

    #[arg(long, default_value = "host-listener-poller")]
    service_name: String,
```

coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[arg(
+        long,
+        default_value_t = 100,
+        help = "Maximum number of blocks to process per iteration"
+    )]
+    batch_size: u64,
+
+    #[arg(
+        long,
+        default_value_t = 1000,
```
rudy-6-4: just thinking, could be half block time for the principle. since there is a lag of several block there is no need to poll every second ?

rudy-6-4 (dismissed):

rudy-6-4 (approved):

PR #1469, opened 2025-12-04
chore(protocol-contracts): clean hardhat config

melanciani:
things added/removed in https://github.com/zama-ai/fhevm/pull/1397/ by mistake

### DIFF
--- a/protocol-contracts/staking/hardhat.config.ts
+++ b/protocol-contracts/staking/hardhat.config.ts
@@ -27,6 +27,7 @@ import 'solidity-coverage';
 // To make use of automatic environment setup:
 // - Duplicate .env.example file and name it .env
 // - Fill in the environment variables
+dotenv.config();
 
 // Set your preferred authentication method
 //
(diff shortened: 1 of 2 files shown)


protocol-contracts/staking/hardhat.config.ts
```diff
@@ -109,15 +110,6 @@ const config: HardhatUserConfig = {
     alice: {
       default: 1, // wallet address of index[1], of the mnemonic in .env
     },
-    bob: {
```
jatZama, line 112: Aryeh from openzeppelin might need those, especially since batching will be introduced with the confidential staking aggregator, so not confident if we should remove them.

protocol-contracts/staking/hardhat.config.ts
```diff
@@ -27,6 +27,7 @@ import 'solidity-coverage';
 // To make use of automatic environment setup:
 // - Duplicate .env.example file and name it .env
 // - Fill in the environment variables
+dotenv.config();
```
jatZama, line 30: Why is this needed? for tasks tests I guess?

protocol-contracts/staking/hardhat.config.ts
melanciani: for context, I added them in for some tests in https://github.com/zama-ai/fhevm/pull/1397 but in the end I don't use them, so I think it's better, they can still add them if they need to

protocol-contracts/staking/hardhat.config.ts
melanciani: for context, I removed this in https://github.com/zama-ai/fhevm/pull/1397 by mistake, it something done by OZ

but yes in practice this loads `.env` and is useful for tasks

jatZama (approved):
LGTM

PR #1480, opened 2025-12-05
chore(coprocessor): improve poller charts and update pool interval

obatirou:


rudy-6-4 (approved):

charts/coprocessor/values.yaml
```diff
@@ -243,13 +243,28 @@ hostListenerPoller:
...
   args:
+    ### Required parameters
     - --database-url=$(DATABASE_URL)
     - --coprocessor-api-key=$(TENANT_API_KEY)
     - --url=$(ETHEREUM_RPC_HTTP_URL)
     - --acl-contract-address=$(ACL_CONTRACT_ADDRESS)
     - --tfhe-contract-address=$(FHEVM_EXECUTOR_CONTRACT_ADDRESS)
+
+    ### Polling and block processing parameters
+    - --finality-lag=3
```
rudy-6-4, line 254: the value should be realistic, sobetter use 16 or higher ?

obatirou: Yes this is realistic on ethereum
on etherscan reorg are of 1 block depth

PR #1482, opened 2025-12-05
feat(protocol-contracts): improve fees management in staking contracts

melanciani:

### DIFF
--- a/protocol-contracts/staking/contracts/OperatorRewarder.sol
+++ b/protocol-contracts/staking/contracts/OperatorRewarder.sol
@@ -146,6 +146,8 @@ contract OperatorRewarder is Ownable {
      * @param basisPoints Maximum fee in basis points (max 10000).
      */
     function setMaxfee(uint16 basisPoints) public virtual onlyOwner {
+        require(basisPoints != _maxfeeBasisPoints, MaxFeeAlreadySet(basisPoints, _maxfeeBasisPoints));
+
         _setMaxfee(basisPoints);
     }
 
@@ -155,6 +157,8 @@ contract OperatorRewarder is Ownable {
      * @param basisPoints Fee in basis points (cannot be greater than the maximum fee).
      */
     function setFee(uint16 basisPoints) public virtual onlyBeneficiary {
+        require(basisPoints != feeBasisPoints(), FeeAlreadySet(basisPoints, feeBasisPoints()));
+
         _setFee(basisPoints);
     }
 
@@ -320,7 +324,6 @@ contract OperatorRewarder is Ownable {
      */
     function _setMaxfee(uint16 basisPoints) internal virtual {
         require(basisPoints <= 10000, InvalidBasisPoints(basisPoints));
-        require(basisPoints != _maxfeeBasisPoints, MaxFeeAlreadySet(basisPoints, _maxfeeBasisPoints));
 
         if (basisPoints < _feeBasisPoints) {
             _setFee(basisPoints);
@@ -337,7 +340,6 @@ contract OperatorRewarder is Ownable {
      */
     function _setFee(uint16 basisPoints) internal virtual {
         require(basisPoints <= 10000, InvalidBasisPoints(basisPoints));
-        require(basisPoints != feeBasisPoints(), FeeAlreadySet(basisPoints, feeBasisPoints()));
         require(basisPoints <= _maxfeeBasisPoints, MaxBasisPointsExceeded(basisPoints, _maxfeeBasisPoints));
 
         _claimFee();
--- a/protocol-contracts/staking/contracts/OperatorRewarder.sol
+++ b/protocol-contracts/staking/contracts/OperatorRewarder.sol
@@ -292,6 +292,10 @@ contract OperatorRewarder is Ownable {
     /**
      * @notice Transfers the beneficiary address.
      * @param newBeneficiary The new beneficiary address.
+     * @dev Transferring the beneficiary address does not trigger a claim of unclaimed fees for the
+     * old beneficiary on purpose. This is to avoid losing unclaimed fees in case a beneficiary loses
+     * access to their private key. It is acceptable as the owner (who can set the beneficiary) is
+     * expected to be a governance DAO.
      */
     function _transferBeneficiary(address newBeneficiary) internal virtual {
         require(newBeneficiary != address(0), InvalidBeneficiary(address(0)));
--- a/protocol-contracts/staking/contracts/OperatorRewarder.sol
+++ b/protocol-contracts/staking/contracts/OperatorRewarder.sol
@@ -27,7 +27,7 @@ contract OperatorRewarder is Ownable {
     ProtocolStaking private immutable _protocolStaking;
     OperatorStaking private immutable _operatorStaking;
     address private _beneficiary;
-    uint16 private _maxfeeBasisPoints;
+    uint16 private _maxFeeBasisPoints;
     uint16 private _feeBasisPoints;
     bool private _shutdown;
     uint256 private _lastClaimTotalAssetsPlusPaidRewards;
@@ -42,7 +42,7 @@ contract OperatorRewarder is Ownable {
     event Shutdown();
 
     /// @notice Emitted when the maximum fee is updated.
-    event MaxfeeUpdated(uint16 oldFee, uint16 newFee);
+    event MaxFeeUpdated(uint16 oldFee, uint16 newFee);
 
     /// @notice Emitted when the fee is updated.
     event FeeUpdated(uint16 oldFee, uint16 newFee);
@@ -87,10 +87,10 @@ contract OperatorRewarder is Ownable {
     /**
      * @notice Initializes the OperatorRewarder contract.
      * @param owner The owner address.
-     * @param beneficiary_ The beneficiary address.
+     * @param beneficiary_ The address that can set and claim fees.
      * @param protocolStaking_ The ProtocolStaking contract address.
      * @param operatorStaking_ The OperatorStaking contract address.
-     * @param initialMaxfeeBasisPoints_ The initial max fee basis points.
+     * @param initialMaxFeeBasisPoints_ The initial max fee basis points.
      * @param initialFeeBasisPoints_ The initial fee basis points.
      */
     constructor(
@@ -98,14 +98,14 @@ contract OperatorRewarder is Ownable {
         address beneficiary_,
         ProtocolStaking protocolStaking_,
         OperatorStaking operatorStaking_,
-        uint16 initialMaxfeeBasisPoints_,
+        uint16 initialMaxFeeBasisPoints_,
         uint16 initialFeeBasisPoints_
     ) Ownable(owner) {
         _transferBeneficiary(beneficiary_);
         _token = IERC20(protocolStaking_.stakingToken());
         _protocolStaking = protocolStaking_;
         _operatorStaking = operatorStaking_;
-        _setMaxfee(initialMaxfeeBasisPoints_);
+        _setMaxFee(initialMaxFeeBasisPoints_);
         _setFee(initialFeeBasisPoints_);
     }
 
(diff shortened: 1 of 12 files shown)


protocol-contracts/staking/contracts/OperatorStaking.sol
```diff
@@ -69,19 +69,34 @@ contract OperatorStaking is ERC20, Ownable, ReentrancyGuardTransient {
...
+     * @param beneficiary The beneficiary address.
+     * @param initialMaxfeeBasisPoints_ The initial maximum fee basis points for the OperatorRewarder contract.
+     * @param initialFeeBasisPoints_ The initial fee basis points for the OperatorRewarder contract.
      */
     constructor(
         string memory name,
         string memory symbol,
         ProtocolStaking protocolStaking_,
-        address owner
+        address owner,
```
jatZama, line 80: As per the discussion with Morten, it seems we agreed that OperatorStaking and OperatorRewarder contracts should no longer be inheriting from Ownable. Instead, they should read the owner directly from the one stored inside ProtocolStaking instance.

protocol-contracts/staking/tasks/deposit.ts
```diff
@@ -14,7 +14,7 @@ import { HardhatRuntimeEnvironment } from 'hardhat/types';
 // --network testnet
 task('task:depositOperatorStakingFromDeployer')
   .addParam('assets', 'The amount of assets to deposit into the operator staking contract', 0n, types.bigint)
-  .addParam('receiver', 'The address of the operator rewarder contract to set the owner fee for', '', types.string)
+  .addParam('receiver', 'The address of the operator rewarder contract to set the fee for', '', types.string)
```
jatZama: the comment here is incorrect: it is not setting the fee here.

jatZama: Also I noticed something a bit confusing in the code of this task , you define zamaTokenMock variable, but it should work even with non mocked, real token. Maybe you should just importe the interface of ERC20 instead of relying on mock contract to reduce confusion, wdyt?

protocol-contracts/staking/tasks/setFee.ts
```diff
@@ -4,28 +4,23 @@ import { getRequiredEnvVar } from './utils/loadVariables';
...
-    0n,
-    types.bigint,
-  )
+// npx hardhat task:setFee --fee 2000 --operatorRewarderAddress 0x1234567890123456789012345678901234567890 --network testnet
+task('task:setFee')
+  .addParam('fee', 'The fee to set for the operator rewarder contract (in 1/100th of a percent)', 0n, types.bigint)
   .addParam(
     'operatorRewarderAddress',
-    'The address of the operator rewarder contract to set the owner fee for',
+    'The address of the operator rewarder contract to set the fee for',
```
jatZama, line 16: Maybe it would simplify operations to pass the operatorStaking address here instead of the operatorRewarder address? the OperatorRewarder address could be pulled from OperatorStaking:rewarder() view function. Wdyt?

protocol-contracts/staking/tasks/setFee.ts
```diff
@@ -76,18 +71,18 @@ task('task:setAllCoprocessorOwnerFees').setAction(async function (_, hre: Hardha
   }
 
   for (let i = 0; i < operatorRewarderAddresses.length; i++) {
-    // Get the owner fee for the operator rewarder contracts
-    const ownerFee = BigInt(parseInt(getRequiredEnvVar(`OPERATOR_REWARDER_COPRO_OWNER_FEE_${i}`)));
-    await hre.run('task:setOwnerFee', { ownerFee, operatorRewarderAddress: operatorRewarderAddresses[i] });
+    // Get the fee for the operator rewarder contracts
+    const fee = BigInt(parseInt(getRequiredEnvVar(`OPERATOR_REWARDER_COPRO_FEE_${i}`)));
```
jatZama, line 75: same remark here, maybe only using OperatorStaking address would simplify operations.

protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -207,16 +276,74 @@ contract OperatorRewarder is Ownable {
         token_.safeTransfer(to, amount);
     }
 
+    /**
+     * @notice Transfers the beneficiary address.
+     * @param newBeneficiary The new beneficiary address.
+     */
+    function _transferBeneficiary(address newBeneficiary) internal virtual {
+        if (newBeneficiary == address(0)) {
```
jatZama: I think you should also check that newBeneficiary!=oldBeneficiary; otherwise revert. This would allow to catch early some mistakes.

protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -207,16 +276,74 @@ contract OperatorRewarder is Ownable {
...
+    }
+
+    /**
+     * @notice Sets the fee in basis points (1/100th of a percent).
+     * Unpaid fees are claimed and transferred to the beneficiary.
+     * @param basisPoints Fee in basis points (cannot be greater than the maximum fee).
+     */
+    function _setFee(uint16 basisPoints) internal virtual {
+        require(basisPoints <= 10000, InvalidBasisPoints(basisPoints));
+        require(basisPoints <= _maxfeeBasisPoints, MaxBasisPointsExceeded(basisPoints, _maxfeeBasisPoints));
```
jatZama: optional: here as well you could add a check to revert if basisPoints is equal to old value, but this is less critical than the beneficiary check imo.

protocol-contracts/staking/contracts/OperatorStaking.sol
melanciani: yes indeed will do in a follow up PR !

protocol-contracts/staking/tasks/deposit.ts
melanciani: yes makes a lot of sense

protocol-contracts/staking/tasks/setFee.ts
melanciani: I'm not convinced why it would simply things, the tools already stores and accesses the rewarder addresses easily. Also it would feel weird to call a task for a setter that is not on the contract address you provide, no ?

protocol-contracts/staking/tasks/setFee.ts
melanciani: same answer

protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -207,16 +289,74 @@ contract OperatorRewarder is Ownable {
...
+     * @param newBeneficiary The new beneficiary address.
+     */
+    function _transferBeneficiary(address newBeneficiary) internal virtual {
+        require(newBeneficiary != address(0), InvalidBeneficiary(address(0)));
+        require(newBeneficiary != _beneficiary, BeneficiaryAlreadySet(newBeneficiary));
+
+        address oldBeneficiary = _beneficiary;
+        _beneficiary = newBeneficiary;
+        emit BeneficiaryTransferred(oldBeneficiary, newBeneficiary);
+    }
```
arr00, line 307: Should claim the fee to the old beneficiary prior to switching to a new beneficiary.

protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -207,16 +289,74 @@ contract OperatorRewarder is Ownable {
...
+
+    /**
+     * @notice Sets the maximum fee in basis points (1/100th of a percent) that the beneficiary
+     * can set.
+     * If the new max fee is lower than the current fee:
+     * - the fee is set to the new max fee
+     * - the unpaid fees are claimed and transferred to the beneficiary
+     * @param basisPoints Maximum fee in basis points (max 10000).
+     */
+    function _setMaxfee(uint16 basisPoints) internal virtual {
```
arr00:
```suggestion
    function _setMaxFee(uint16 basisPoints) internal virtual {
```

arr00: Shouldn't it be maxFee everywhere (as opposed to maxfee)

protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -133,6 +199,14 @@ contract OperatorRewarder is Ownable {
         }
     }
 
+    /**
+     * @notice Returns the beneficiary address.
+     * @return The beneficiary address.
```
arr00: Saying who the `beneficiary` is would be more informative. Such as, the address that receives and sets fees.

arr00:
Who will be the `owner`? Is this going to be a trusted party or governance system? In that case, should we consider upgradeability here?

Upgradeability was initially ruled out as allowing the operator themselves to control that would be far too permissive.

protocol-contracts/staking/contracts/OperatorRewarder.sol
melanciani: we've indeed discussed about this internally and we went towards not doing it because it allows not losing unclaimed fees in case the beneficiary key is lost !

we were ok with it because our goal is to make the owner the governance DAO so we  don't consider that there is a risk of the DAO "stealing" the unclaimed fees from the old beneficiary 

wdyt ?

protocol-contracts/staking/contracts/OperatorRewarder.sol
melanciani: ah yes good catch, of course !

protocol-contracts/staking/contracts/OperatorRewarder.sol
melanciani: totally agree, will do

melanciani:
> Who will be the owner? Is this going to be a trusted party or governance system? In that case, should we consider upgradeability here?

> Upgradeability was initially ruled out as allowing the operator themselves to control that would be far too permissive.

the owner will be a governance DAO system, so yes it does make sense to consider upgradability in the end ! I will add it in a follow-up PR

btw note that in https://github.com/zama-ai/fhevm/pull/1540 we make the owner of ProtocolStaking the owner of OperatorStaking and OperatorRewarder ! thanks for the suggestion

protocol-contracts/staking/contracts/OperatorRewarder.sol
arr00: I guess that's fine. Given that claiming the fee is gated to the beneficiary, we do have a guarantee that if the beneficiary key is lost, no fees will be lost after the key is lost.

protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -207,16 +289,74 @@ contract OperatorRewarder is Ownable {
...
+        _maxFeeBasisPoints = basisPoints;
+    }
+
+    /**
+     * @notice Sets the fee in basis points (1/100th of a percent).
+     * Unpaid fees are claimed and transferred to the beneficiary.
+     * @param basisPoints Fee in basis points (cannot be greater than the maximum fee).
+     */
+    function _setFee(uint16 basisPoints) internal virtual {
+        require(basisPoints <= 10000, InvalidBasisPoints(basisPoints));
```
arr00: Note that this is already enforced by the `require` statement below.

protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -207,16 +289,74 @@ contract OperatorRewarder is Ownable {
...
+     * can set.
+     * If the new max fee is lower than the current fee:
+     * - the fee is set to the new max fee
+     * - the unpaid fees are claimed and transferred to the beneficiary
+     * @param basisPoints Maximum fee in basis points (max 10000).
+     */
+    function _setMaxFee(uint16 basisPoints) internal virtual {
+        require(basisPoints <= 10000, InvalidBasisPoints(basisPoints));
+
+        if (basisPoints < _feeBasisPoints) {
```
arr00: Prefer to use the getters when those exist `feeBasisPoints()`

protocol-contracts/staking/tasks/setFee.ts
jatZama: ok

protocol-contracts/staking/tasks/setFee.ts
jatZama: ok

jatZama (approved):

arr00 (approved):

PR #1498, opened 2025-12-06
docs(common): update sdk docs for user and public decryption

manoranjith:
- In both cases, add an info banner that the total bit length of ciphertexts being decrypted should not exceed 2048 bits.

- In user decrypt, emphasize that API supports batch decryption.

### DIFF
--- a/docs/sdk-guides/public-decryption.md
+++ b/docs/sdk-guides/public-decryption.md
@@ -25,6 +25,10 @@ const handles = [
 const values = instance.publicDecrypt(handles);
 ```
 
+{% hint style="info" %}
+The total bit length of all ciphertexts being decrypted in a single request must not exceed 2048 bits. Each encrypted type has a specific bit length, for instance `euint8` uses 8 bits and `euint16` uses 16 bits. For the full list of encrypted types and their corresponding bit lengths, refer to the [encrypted types documentation](../solidity-guides/types.md#list-of-encrypted-types).
+{% endhint %}
+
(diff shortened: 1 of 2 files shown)


docs/sdk-guides/public-decryption.md
```diff
@@ -25,6 +25,10 @@ const handles = [
 const values = instance.publicDecrypt(handles);
 ```
 
+{% hint style="info" %}
+The total bit length of all ciphertexts being decrypted in a single request must not exceed 2048 bits. Each encrypted type has a specific bit length, for instance `euint8` uses 8 bits and `euint16` uses 16 bits. For the full list of encrypted types and their corresponding bit lengths, refer to the [encrypted types documentation](../solidity-guides/types.md#list-of-encrypted-types).
```
jatZama, line 29: You should add a warning: there is an exception for ebool taking 2 bits instead of the expected 1 bit, due to a tfhe-rs quirk. I don't know exactly this is the case, we should ask tfhe-rs team.

docs/sdk-guides/user-decryption.md
```diff
@@ -58,6 +59,8 @@ The user needs to have created an instance object prior to that (for more contex
 // contractAddress: [`string`]
 
 const keypair = instance.generateKeypair();
+// userDecrypt can take a batch of handles (with their corresponding contract addresses).
+// In this example we only pass one handle.
```
jatZama, line 63: You should add a specific example with at least 2 handles at once.

PR #1505, opened 2025-12-08
chore(kms-connector): more granular metrics

eudelins-zama:
* chore(kms-connector): more granular metrics

* docs(kms-connector): update metrics documentation


dartdart26 (approved):
Looks good!

PR #1506, opened 2025-12-08
feat(coprocessor): add a non-blocking, distributed locking mechanism across multiple tfhe-workers

goshawk-3:
- fixes https://github.com/zama-ai/fhevm-internal/issues/742

This adds a non-blocking, distributed locking mechanism that coordinates dependence-chain processing across  tfhe-workers replicas.

A worker can acquire a lock of the next available dependence-chain entry for processing ordered by last_updated_at (FIFO queue-like approach).

Ownership expires after a timeout, enabling work-stealing by other workers for resilience.

New CLI params
- --worker_id
- --dcid_ttl_sec


coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
```diff
@@ -0,0 +1,208 @@
...
+        info!(target: "deps_chain", ?row, "Acquired lock");
+
+        Ok(Some(row.dependence_chain_id))
+    }
+
+    /// Release all locks held by this worker
+    ///
+    /// If host-listener has marked the dependence chain as 'updated' in the meantime,
+    /// we don't overwrite its status
+    pub async fn release_all_owned_locks(&self) -> Result<u64, sqlx::Error> {
```
obatirou: it seems both release* function do not update self.lock but only the DB?

coprocessor/fhevm-engine/tfhe-worker/benches/utils.rs
```diff
@@ -108,6 +108,7 @@ async fn start_coprocessor(rx: Receiver<bool>, app_port: u16, db_url: &str) {
         health_check_port: 8080,
         metric_rerand_batch_latency: MetricsConfig::default(),
         metric_fhe_batch_latency: MetricsConfig::default(),
+        worker_id: None,
```
rudy-6-4, line 111: Couldn't they invent a unique id on start or from a table ? If we have dozen workers, that could be error prone.

coprocessor/fhevm-engine/db-migration/migrations/20251205154454_create_dependence_chain_table.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE dependence_chain (
+    dependence_chain_id  bytea PRIMARY KEY,
+
+    -- Scheduling / Coordination
+    status              TEXT NOT NULL CHECK (status IN (
+                                'updated', 'processing', 'processed'
+                            )),
+    error_message              TEXT,  -- optional error message if processing failed
```
rudy-6-4: There is no default, what HL should insert ?

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
goshawk-3: Correct, that was fixed in recent commits

coprocessor/fhevm-engine/tfhe-worker/benches/utils.rs
goshawk-3: That's optional. If `worker_id` is not provided, the worker generates its own uuid.

coprocessor/fhevm-engine/db-migration/migrations/20251205154454_create_dependence_chain_table.sql
goshawk-3: For error_message - nothing

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -137,6 +156,8 @@ async fn tfhe_worker_cycle(
 
         // Execute transactions segregated by tenant
         for (tenant_id, ref mut tenant_txs) in transactions.iter_mut() {
+            deps_chain_mngr.extend_current_lock().await?;
```
obatirou: results is not checked
What append if it returns none?

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
```diff
@@ -0,0 +1,286 @@
...
+    ) -> Result<Option<(Vec<u8>, LockingReason)>, sqlx::Error> {
+        let dependence_chain_id = match &self.lock {
+            Some(lock) => lock.dependence_chain_id.clone(),
+            None => {
+                debug!("No lock to extend");
+                return Ok(None);
+            }
+        };
+
+        sqlx::query!(
```
obatirou: result is not checked. Could 0 rows be affected?

goshawk-3: Definitely!

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
goshawk-3: This is just a best-effort attempt to extend the lock and prevent other workers from trying to lock the same DCID. Worst-case scenario, it returns None if the lock has expired. However, the worker has already secured  exclusive access to the transaction computations in the `Computations` table.

I can add a log::warn here too.

obatirou: Thank you for the clarification 
a warn could be useful indeed

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -322,8 +395,9 @@ WITH selected_computations AS (
       WHERE is_completed = FALSE
         AND is_error = FALSE
         AND is_allowed = TRUE
+        AND ($1::bytea IS NULL OR dependence_chain_id = $1)
       ORDER BY schedule_order
-      LIMIT $1
+      LIMIT $2
     ) as c_schedule_order
```
antoniupop, line 401: I would say at some point we won't need the union here (probably not even need the schedule order anymore as that would be the role of the last_updated_at in a sense), but can keep for now. We'll have to refactor/simplify this handling, but we need to think of a way to deal with missing events wrt. dependence chain selection as well at some point. If there's "lots" of workers it's less of a problem as only some would be stuck

goshawk-3:
Closed as duplicate https://github.com/zama-ai/fhevm/pull/1550

PR #1519, opened 2025-12-09
fix(coprocessor): host-listener, dependency chain

rudy-6-4:
Compute chain dependency on the transaction level instead of ops level.


coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -0,0 +1,181 @@
...
+fn scan_transactions(logs: &[LogTfhe]) -> (HandleToTransaction, Transactions) {
+    let nb_logs = logs.len();
+    let mut prev_tx = None;
+    let mut seen_txs = HashSet::with_capacity(nb_logs);
+    let mut txs = VecDeque::with_capacity(nb_logs);
+    let mut handle_to_tx = HashMap::with_capacity(nb_logs);
+    for (age, log) in logs.iter().enumerate() {
+        let tx = log.transaction_hash.unwrap_or_default();
+        if let Some(handle) = tfhe_result_handle(&log.event) {
+            handle_to_tx.insert(handle, (age, tx));
```
antoniupop: Not sure if it's possible, but ideally this should be filtered to only include allowed handles.

rudy-6-4: It is filtered for the past cache, but still needed here to detect internal dependency. Otherwise it could promote an old handle (several block before) as dependency chain vs the fresher one. But this policy could be changed if you think it's better the other way around for transaction internal dependency.r

rudy-6-4:
To be updated after https://github.com/zama-ai/fhevm/pull/1553 is merged.

rudy-6-4:
merged in feature branch,

PR #1521, opened 2025-12-09
feat(gateway-contracts): store the hash instead of the full handles

ClementWalter:
Remove the `publicCtHandles` mapping from the `Decryption` contract and
use the `publicCtHandlesHashes` mapping instead.

The hash could most probably also be used as the `decryptionId` for the public decryption request.
This has not been done in this PR as it may require some other changes (if the id is used elsewhere and expecting a given encoding).

Closes: https://github.com/zama-ai/fhevm-internal/issues/747


jatZama:
~This will break EIP712 based verification in a lot of components, so you will have to also modify EIP712_PUBLIC_DECRYPT_TYPE and the logic inside _validateDecryptionResponseEIP712Signature. Please note that whenver a change is made to EIP712 structs, it is a significant change requiring change of many components before being merged: host contracts, solidity library, mock mode, relayer-sdk, and KMS core, etc. Normally this kind of change need coordination between different teams and require 2 weeks of work before being merged.~
EDIT: Sorry, I read too fast, it does nto break actually the EIP712, but it is a bit weird to have different struct defined than what is actually used in the  EIP712_PUBLIC_DECRYPT_TYPE string. Main problem I think is storage layout breaking change for upgradeability.

dd23:
Just note that this will require an update on the KMS  side (at least, maybe elsewhere) as we change `PublicDecryptVerification`.

ClementWalter:
I don't understand: I see nowhere where `publicCtHandles` is used. I'm just changing internal storage layout, why would this require change from other parts of the protocol?

dd23:
The KMS is signing the `PublicDecryptVerification` after it has publicly decrypted a value.
So this struct needs to match across components: https://github.com/zama-ai/kms/blob/10f73a3418f17ef263c894a25b283d34349dfcaf/core/grpc/src/solidity_types.rs#L31-L35

ClementWalter:
the struct is not shared, there are several definition of `PublicDecryptVerification`. nothing enforces to have them to be the same. not saying that it's good to have several struct with the same name (!) nor several with the same name and not the same defition. But it's what it is. I don't think that the proposed change requires any other update on the codebase to work

ClementWalter:
also @jatZama if it breaks a lot of things and the CI is all green, it's quite concerning tbf

gateway-contracts/contracts/Decryption.sol
```diff
@@ -257,7 +257,7 @@ contract Decryption is
             mapping(bytes32 digest => bytes[] verifiedSignatures))
                 verifiedPublicDecryptSignatures;
         /// @notice Handles of the ciphertexts requested for a public decryption
-        mapping(uint256 decryptionId => bytes32[] ctHandles) publicCtHandles;
+        mapping(uint256 decryptionId => bytes32 ctHandlesHash) publicCtHandlesHashes;
```
jatZama, line 260: This kind of storage layout change will break upgradeability I am afraid.
Btw don't forget to bump minor version and reinitializer value.
I agree CI should be improved to catch this kind of bug, I think it will be handled better once we have mergify support enabled, running correct e2e tests. Currently CI is lacking, focusing more on unit tests than e2e , but I am not an expert in CI.

dd23:
It seems E2E tests are not running for some reason. These will break with the changes in this PR.

ClementWalter:
ok closing, but tbh the fact that changing an internal storage layout that doesn't update any public api, for a storage var that is stored once, read once, is quite concerning to me

PR #1525, opened 2025-12-09
fix(coprocessor): host-listener, restart on last validated block on disconnect

rudy-6-4:


rudy-6-4:
In 0.10 https://github.com/zama-ai/fhevm/pull/1500

antoniupop (approved):

PR #1532, opened 2025-12-10
refactor(gateway-contracts): remove redundant decryptionDone storage for public decryption

enitrat:
## Summary

This PR optimizes public decryption by removing a redundant storage write in the `Decryption` contract, saving approximately 20,000 gas per public decryption request.

### Storage Optimization
- **Renamed `decryptionDone` to `userDecryptionDone`**: The mapping is now scoped to only user/delegated user decryption, where it's actually needed
- **Removed redundant SSTORE in `publicDecryptionResponse()`**: Public decryption consensus is now tracked solely through `decryptionConsensusDigest`, eliminating an unnecessary storage write

## Gas Impact

- **Savings**: ~20,000 gas per public decryption request (1 cold SSTORE removed)
- **No additional cost**: The check now uses `decryptionConsensusDigest` which is already read/written in the same function

## Compatibility

- **External interfaces**: No changes - `isDecryptionDone()` works identically for both types
- **kms-connector compatibility**: Verified that the off-chain component calls `isDecryptionDone()` which maintains the same behavior, not the internal mapping

Close https://github.com/zama-ai/fhevm-internal/issues/751


ClementWalter (commented):
lgtm

gateway-contracts/contracts/Decryption.sol
```diff
@@ -231,8 +231,8 @@ contract Decryption is
         // ----------------------------------------------------------------------------------------------
         // Common decryption state variables:
         // ----------------------------------------------------------------------------------------------
-        /// @notice Whether a (public, user, delegated user) decryption is done
-        mapping(uint256 decryptionId => bool decryptionDone) decryptionDone;
+        /// @notice Whether a (user, delegated user) decryption is done
+        mapping(uint256 decryptionId => bool userDecryptionDone) userDecryptionDone;
```
isaacdecoded, line 235: Note that, unfortunately, we can't rename state variables, but deprecate them and add the new ones at the end of the storage struct. This is because of upgrade conflicts between storage layouts. Also, we must bump the contract and the reinitializer versions in order to properly upgrade the contract once the change is introduced.

enitrat: ah ok cool thanks for raising that!

enitrat: >  Also, we must bump the contract and the reinitializer versions in order to properly upgrade the contract once the change is introduced.

what does this mean?

isaacdecoded: This means:
1. Bump internal contract version [here](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L215)
2. And also bump the reinitializer version [here](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L224)

The second one is relevant for upgrading the contract, so the [reinitializerVX method](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L308) can be called with the correct versioning. For example, on testnet we have deployed now [v0.10.x with reinitializer version 4](https://github.com/zama-ai/fhevm/blob/v0.10.0/gateway-contracts/contracts/Decryption.sol#L224), so when we need to upgrade to v0.11.x we should have this reinitializer version incremented to 5 in order to properly call the `reinitializerV4()` method.

PR #1535, opened 2025-12-10
chore(protocol-contracts): remove ProtocolOperatorRegistry

melanciani:
following work done in https://github.com/zama-ai/fhevm/pull/1482 with respect to the owner changes, we believe the registry does not make much sens anymore 


arr00:
Was about to comment about this in #1482. Is there a plan on how to associate operators with their staking contracts? Is having that hardcoded somewhere acceptable long-term?

melanciani:
> Is there a plan on how to associate operators with their staking contracts? Is having that hardcoded somewhere acceptable long-term?

@arr00 internally we've said we're not going to do this at first, but we can always add something later if needed !

jatZama (approved):
LGTM

PR #1536, opened 2025-12-10
test(test-suite): e2e for input, compute and decrypt

dartdart26:
Also, add ability to run the tests on the TestInput contract without deploying it each time via the `TEST_INPUT_CONTRACT_ADDRESS` env var.

Intention is to use this a single entry point for a smoke test.


antoniupop (dismissed):

dartdart26:
E2E did pass here: https://github.com/zama-ai/fhevm/actions/runs/20102597168/job/57677374223?pr=1536

I will be bypassing the rules to merge with the version of `v0.10.3` that I haven't tagged yet.

antoniupop (approved):

PR #1541, opened 2025-12-10
feat(protocol-contracts): make OperatorStaking inherit ERC1363

melanciani:
suggested by @arr00 to help migrating to a new operator staking contract (if ever needed)


jatZama (approved):
LGTM

PR #1547, opened 2025-12-10
chore(protocol-contracts): reword staking to delegating in OperatorStaking/Rewarder

melanciani:
we now want to have users "delegate" rather than "stake" in OperatorStaking (being "delegators" instead of "stakers")

### DIFF
--- a/protocol-contracts/staking/contracts/OperatorStaking.sol
+++ b/protocol-contracts/staking/contracts/OperatorStaking.sol
@@ -20,12 +20,12 @@ import {ProtocolStaking} from "./ProtocolStaking.sol";
 /**
  * @title OperatorStaking
  * @custom:security-contact security@zama.ai
- * @notice Allows users to stake assets and receive shares, with support for reward distribution.
+ * @notice Allows users to delegate assets to an operator staker and receive shares, with support for reward distribution.
  * @dev Integrates with ProtocolStaking and OperatorRewarder contracts. Inspired by ERC7540 but not fully compliant.
  * Also inherits ERC1363 to ease of users with potential OperatorStaking contract migrations.
  *
  * NOTE: This contract supports slashing on the `ProtocolStaking` level, meaning that the overall stake of this contract
- * may decrease due to slashing. These losses are symmetrically passed to restakers on the `OperatorStaking` level.
+ * may decrease due to slashing. These losses are symmetrically passed to delegators on the `OperatorStaking` level.
  * Slashing must first decrease the `ProtocolStaking` balance of this contract before affecting pending withdrawals.
  */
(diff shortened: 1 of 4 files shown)


protocol-contracts/staking/contracts/OperatorStaking.sol
```diff
@@ -174,9 +174,10 @@ contract OperatorStaking is ERC20, Ownable, ReentrancyGuardTransient {
      * accounting for all in-flight redemptions are restaked into the `ProtocolStaking` contract.
      *
      * NOTE: Excess tokens will be in the `OperatorStaking` contract the operator is slashed
-     * during a redemption flow. Anyone can call this function to restake those tokens.
+     * during a redemption flow or if donations are made to it. Anyone can call this function to
+     * restake those tokens.
      */
-    function restake() public virtual {
+    function restakeOperatorExcess() public virtual {
```
mortendahl: Having `Operator` in `restakeOperatorExcess` makes it sound like this is something belonging to the operator. Maybe simply `stakeExcess`?

mortendahl (commented):
Minor comment, otherwise lgtm

protocol-contracts/staking/contracts/OperatorStaking.sol
melanciani: I see what you mean, the operator was to highlight the fact that we restake at the operatorStaking contract level

melanciani: I'll update, `stakeExcess` was @jatZama 's initial suggestion

jatZama (dismissed):
LGTM

protocol-contracts/staking/contracts/OperatorStaking.sol
```diff
@@ -191,9 +191,10 @@ contract OperatorStaking is ERC1363, Ownable, ReentrancyGuardTransient {
      * accounting for all in-flight redemptions are restaked into the `ProtocolStaking` contract.
      *
      * NOTE: Excess tokens will be in the `OperatorStaking` contract the operator is slashed
-     * during a redemption flow. Anyone can call this function to restake those tokens.
+     * during a redemption flow or if donations are made to it. Anyone can call this function to
+     * restake those tokens.
      */
-    function restake() public virtual {
+    function restakeExcess() public virtual {
```
jatZama: Morten was suggesting `stakeExcess` actually (I know I suggested restakeExcess at first). Not sure we settle on this name then.

melanciani: ah didn't even notices in his message good catch

jatZama (approved):

PR #1550, opened 2025-12-11
feat(coprocessor): add a non-blocking, distributed locking mechanism in tfhe-worker

goshawk-3:
- fixes https://github.com/zama-ai/fhevm-internal/issues/742

This adds a non-blocking, distributed locking mechanism that coordinates dependence-chain processing across multiple tfhe-workers replicas.

A worker can acquire a lock of the next available dependence-chain entry for processing ordered by last_updated_at (FIFO queue-like approach). 

A permission to acquire a DCID depends on either 
- dependency_count is 0 and DCID is not locked
or
- dependency_count is 0 and DCID is locked but the lock has expired

- Ownership expires after a timeout, enabling work-stealing by other workers for resilience.
- GC procedure is regularly executed to clean up processed DCIDs


rudy-6-4:
values.yaml new parameters are missing

goshawk-3:
See also: https://github.com/zama-ai/fhevm/pull/1506#discussion_r2610175689

antoniupop:
> New CLI params
> 
> * --worker-id
> * --dcid-ttl-sec
> * --dcid-timeslice-sec
> * --disable-dcid-locking

Please could you update the charts with these (or any new params added) - I think we've mostly converged on the arch, so would be good to start planning for deployment.

rudy-6-4 (dismissed):

antoniupop (dismissed):

antoniupop:
e2e CI green https://github.com/zama-ai/fhevm/actions/runs/20568622480/job/59071264500

PR #1561, opened 2025-12-12
feat(kms-connector): garbage collection

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/749

## DB changes

This PR adds the `status` (custom enum) and `updated_at` (timestamp) fields while removing the `under_process` field of all the requests/responses tables.
The `status` field can be:
- `pending`: waiting to be processed
- `under_process`: currently processed, so the request/response is "locked"
- `completed` (will be deleted later by the garbage collector)
- `failed` (will be deleted later by the garbage collector)

### Request/Response deletion

Before this PR, the kms-connector tried to immediatly delete request/response once it has been processed.
Now it will instead mark it as `completed` (or `failed`), and a garbage collector will handle them later on.
The garbage collector will:
- delete decryption requests/responses
- keep other kind of requests/responses
  - Why? Because they are related to keygen or initialization of the KMS, so they are pretty rare, and it could be useful to keep a trace of them

As we want to run a single instance of this garbage collector, it is running alongside the `TransactionSender`, as it is the only micro-service of the kms-connector that should also be a single instance.

### Decryption Request/Response unlocking

Before this PR, if a `KmsWorker` picks and locks a decryption request from the DB to process it, if it crashes or is restarted for some reason, then this request will be locked forever and never processed.
Now the garbage collector will unlock the request if it is in a `under_process` status for too long.

Same behavior for the `TransactionSender` and decryption responses.

Note that we only unlock decryption requests/responses and not the other, because as explained above, they are related to keygen/init of the KMS, so are pretty rare, and under the protocol's control. Thus it is easier to manually investigate what is happening in such case, and manually unlock or re-trigger requests if needed. That way, it avoid running multiple SQL queries that will be useless 99.99% of the time.

## Code changes

I removed the `KmsResponseRemover` abstraction. All of these abstractions were mostly created in the 1st place because a migration from Postgres DB to Kafka was something to consider.
But IMO, it seems less probable these days, and it was easier/cleaner to achieve the update of the `KmsResponse` `status` using method than a dedicated struct to pass everywhere.


dartdart26:
I think what could be nice is to have a high-level description of the change as part of the PR (and commit message) as that would make it easier for the reviewer and also to track what changes have been made if going through commits later. Copilot has taken a stab at it, but I am sure you as an author would have much better context :)

kms-connector/config/tx-sender.toml
```diff
@@ -101,3 +101,17 @@ private_key = "8da4ef21b864d2cc526dbdb2a120bd2874c36c9d0a1fb7f8c63d7f7a8b41de8f"
...
 # task_limit = 1000
+
+# The interval between garbage collection runs, in minutes (optional, defaults to 5).
+# ENV: KMS_CONNECTOR_GC_RUN_INTERVAL_MINS
+# gc_run_interval_mins = 5
+
+# The expiration time in minutes for completed/failed decryptions, after which they will be
+# deleted (optional, defaults to 1440 = 24 hours).
+# ENV: KMS_CONNECTOR_GC_DECRYPTION_EXPIRY_MINS
+# gc_decryption_expiry_mins = 1440
```
dartdart26, line 112: Wondering if the default should be less, maybe a few hours at most? Wdyt?

kms-connector/config/tx-sender.toml
```diff
@@ -101,3 +101,17 @@ private_key = "8da4ef21b864d2cc526dbdb2a120bd2874c36c9d0a1fb7f8c63d7f7a8b41de8f"
...
+
+# The time limit in minutes for decryption to be under process, after which they will be
+# considered as pending again (optional, defaults to 10).
+# ENV: KMS_CONNECTOR_GC_DECRYPTION_UNDER_PROCESS_LIMIT_MINS
+# gc_decryption_under_process_limit_mins = 10
```
dartdart26: Should that be say 1 min? Wdyt?

kms-connector/connector-db/migrations/20251208161022_implem_garbage_collection.sql
```diff
@@ -0,0 +1,322 @@
...
+-- Any remaining locked crsgen/keygen operations in DB are failures (stuck operations), but there
+-- should be none AFAIK.
+UPDATE prep_keygen_requests SET status = 'failed' WHERE under_process = TRUE;
+UPDATE keygen_requests SET status = 'failed' WHERE under_process = TRUE;
+UPDATE crsgen_requests SET status = 'failed' WHERE under_process = TRUE;
+UPDATE prss_init SET status = 'failed' WHERE under_process = TRUE;
+UPDATE key_reshare_same_set SET status = 'failed' WHERE under_process = TRUE;
+UPDATE prep_keygen_responses SET status = 'failed' WHERE under_process = TRUE;
+UPDATE keygen_responses SET status = 'failed' WHERE under_process = TRUE;
+UPDATE crsgen_responses SET status = 'failed' WHERE under_process = TRUE;
```
dartdart26, line 53: Should the status be set to `pending` here?

kms-connector/connector-db/migrations/20251208161022_implem_garbage_collection.sql
```diff
@@ -0,0 +1,322 @@
...
+--                                Create indexes for decryption                                 --
+--------------------------------------------------------------------------------------------------
+CREATE INDEX idx_public_decryption_requests_status ON public_decryption_requests (status);
+CREATE INDEX idx_public_decryption_requests_status_updated_at ON public_decryption_requests (status, updated_at);
+CREATE INDEX idx_user_decryption_requests_status ON user_decryption_requests (status);
+CREATE INDEX idx_user_decryption_requests_status_updated_at ON user_decryption_requests (status, updated_at);
+CREATE INDEX idx_public_decryption_responses_status ON public_decryption_responses (status);
+CREATE INDEX idx_public_decryption_responses_status_updated_at ON public_decryption_responses (status, updated_at);
+CREATE INDEX idx_user_decryption_responses_status ON user_decryption_responses (status);
+CREATE INDEX idx_user_decryption_responses_status_updated_at ON user_decryption_responses (status, updated_at);
```
dartdart26: Do we need two indexes here? Since `status` is leftmost, I think we can remove the ones on `status` only and keep the composite ones. Let me know if I miss something that would prevent us from doing that.

kms-connector/crates/kms-worker/src/core/event_picker/picker.rs
```diff
@@ -110,18 +110,19 @@ impl DbEventPicker {
...
-                    WHERE under_process = FALSE
+                    WHERE status = 'pending'
                     LIMIT $1 FOR UPDATE SKIP LOCKED
                 ) AS req
                 WHERE public_decryption_requests.decryption_id = req.decryption_id
                 RETURNING req.decryption_id, sns_ct_materials, extra_data, otlp_context, already_sent, error_counter
             ",
         )
         .bind(self.events_batch_size as i16)
+        .bind(chrono::Utc::now().naive_utc())
```
dartdart26: I think it would be better to use the DB time as we use it as NOW() in some places already and also makes it uniform across services.

kms-connector/crates/tx-sender/src/monitoring/garbage_collection.rs
dartdart26, line 1: Nit: Maybe we can name this file `garbage_collection.rs` for clarity.

dartdart26 (commented):
Nice work! I am almost ready to approve, just wanted to discuss some of the comments I put first.

eudelins-zama:
> I think what could be nice is to have a high-level description of the change as part of the PR (and commit message) as that would make it easier for the reviewer and also to track what changes have been made if going through commits later. Copilot has taken a stab at it, but I am sure you as an author would have much better context :)

You're right @dartdart26, done! :slightly_smiling_face:

kms-connector/crates/kms-worker/src/core/event_picker/picker.rs
eudelins-zama: Yes indeed + I added some Postgres triggers to auto-update the `update_at` fields on update, so it's pointless to use this `bind`.

kms-connector/connector-db/migrations/20251208161022_implem_garbage_collection.sql
eudelins-zama: Not super familiar with DB indexes tbh.

As we are querying requests/responses two ways:
- only by `status=pending` to select requests/responses to process
- with `status=xxx && (NOW() - updated_at > time_limit)` to delete/unlock requests/responses

I though the two indexes were useful. But I might be wrong, wdyt?

kms-connector/connector-db/migrations/20251208161022_implem_garbage_collection.sql
eudelins-zama: Imagine there is a keygen leftover from our initial keygen that was stuck in `under_process` state for some reason.
Then I don't think we want to send it to the KMS now, but instedad consider it as `failed`, it is safer right?

kms-connector/config/tx-sender.toml
eudelins-zama: If we poll for a decryption result that is not ready yet, the KMS maintain the GRPC connection for 60s before responding with an `Unavailable` error.
And we are doing at least 3 attempts by default.

So I think we should not put this limit to less than 3 minutes, I put 10 to be safer. Let me know wdyt

kms-connector/config/tx-sender.toml
eudelins-zama: I am not familiar with DB indexes tbh.

We are querying decryption in two ways:
- using `status=pending` fetch requests/responses to process
- using `status=xxx && (NOW() - updated_at > $time_limit)` to delete/unlock requests/responses

So I thought the two indexes would be useful, but I might be wrong. Wdyt?

kms-connector/crates/kms-worker/src/core/event_picker/picker.rs
eudelins-zama: Fixed in f5b2b68f761f4c446d32a9b6ee500f387594ba2e

kms-connector/crates/tx-sender/src/monitoring/garbage_collection.rs
eudelins-zama: Fixed in 553dbfcb62e2d0365d0c2a6325cfc99650769c51

kms-connector/connector-db/migrations/20251208161022_implem_garbage_collection.sql
dartdart26: Since `status` is part of the composite index on the left side, AFAICT, you don' need a separate one just for it. You could use an EXPLAIN query to verify no table scans (though that might vary with table size).

> A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but the index is most efficient when there are constraints on the leading (leftmost) columns. The exact rule is that equality constraints on leading columns, plus any inequality constraints on the first column that does not have an equality constraint, will always be used to limit the portion of the index that is scanned.

https://www.postgresql.org/docs/current/indexes-multicolumn.html

kms-connector/config/tx-sender.toml
dartdart26: I responded to that in a separate comment, I think that's a copy/paste issue.

kms-connector/config/tx-sender.toml
dartdart26: Got it. Maybe we reduce it to say 6 minutes?

kms-connector/connector-db/migrations/20251208161022_implem_garbage_collection.sql
dartdart26: I might have been confused by the comment as it says `pending`. Your thinking makes sense to me, we can set to `failed`.

eudelins-zama: `pending` was just for decryption, will try to make the comment a bit clearer :slightly_smiling_face:

kms-connector/config/tx-sender.toml
eudelins-zama: Yes indeed, sorry :sweat_smile:.

I have no strong opinion on this, I just think we have two possible approaches:
- either we keep a very short decryption expiry (like few minutes) just to avoid the race condition that make us process a decryption twice, and that's it. If we have an error during the processing of a decryption, we would rely 100% on logs as we are doing right now
- either we keep it longer (like few days) to let us time to investigate with logs + the DB if a decryption goes wrong

Wdyt?

dartdart26: I guess we can keep them for 24h, let's keep as is.

kms-connector/connector-db/migrations/20251208161022_implem_garbage_collection.sql
eudelins-zama: Fixed in 4c76f875eff42d48d5de22f73d1d8b0677565f1c

kms-connector/config/tx-sender.toml
eudelins-zama: Fixed in 38bd3499875a5823fbb25a7580a3b32f8b590af5

dartdart26 (approved):
Great work, thanks!

eudelins-zama:
@mergifyio queue

PR #1570, opened 2025-12-13
fix(coprocessor): fix query for work duplication

antoniupop:
This PR fixes duplication of records fetched which occurs after the introduction of the union which can return duplicate transaction IDs.


goshawk-3 (approved):
LGTM

antoniupop:
@mergifyio queue

PR #1595, opened 2025-12-17
chore(coprocessor): lock version of bincode lib

antoniupop:
Subsequent versions could be supply chain attacks - make sure we don't upgrade this by accident until we remove it from the codebase.


rudy-6-4 (approved):

antoniupop:
@mergifyio queue

PR #1605, opened 2025-12-18
feat(protocol-contracts): return release time on calls to `requestRedeem`

arr00:
This PR adds a return type of `uint48` to the `requestRedeem` function. The return value is the timestamp at which the redemption request is ready for redemption--after the cooldown is complete. This make building on top of the `OperatorStaking` contract much simpler.

We also opted to revert when calls to `redeem` or `requestRedeem` specify 0 shares.

### DIFF
--- a/protocol-contracts/staking/contracts/OperatorStaking.sol
+++ b/protocol-contracts/staking/contracts/OperatorStaking.sol
@@ -170,9 +170,10 @@ contract OperatorStaking is ERC1363Upgradeable, ReentrancyGuardTransient, UUPSUp
      * @param shares Amount of shares to redeem.
      * @param controller The controller address for the request.
      * @param ownerRedeem The owner of the shares.
+     * @return releaseTime The timestamp when the assets will be available for withdrawal.
      */
-    function requestRedeem(uint208 shares, address controller, address ownerRedeem) public virtual {
-        if (shares == 0) return;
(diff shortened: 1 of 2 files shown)


protocol-contracts/staking/contracts/OperatorStaking.sol
```diff
@@ -196,6 +200,8 @@ contract OperatorStaking is ERC1363Upgradeable, ReentrancyGuardTransient, UUPSUp
         $._redeemRequests[controller].push(releaseTime, controllerSharesRedeemed + shares);
 
         emit RedeemRequest(controller, ownerRedeem, 0, msg.sender, shares, releaseTime);
```
jatZama, line 202: Just realized this now, but we no longer need the 3rd argument here, which is always 0, in the RedeemRequest event, initially it was only here because we tried to adhere to the ERC7540 (async vaults), but since we dont follow this ERC anyways, especially now that we added the releaseTime in the event, I think we could remove it... wdyt?

jatZama: But if we do this, we will probably need to update the front-end app as well, so it might be too late actually.

melanciani: we will update this in https://github.com/zama-ai/fhevm/pull/1681 + consider the change in the frontend once it is merged yes !

melanciani (approved):
thanks !

PR #1606, opened 2025-12-18
chore(protocol-contracts): add compiler settings and clean licenses

melanciani:
in this PR: 
- add compiler settings to be able to compile contracts
    - update optimizer runs to 800
- update SPDX licenses in all contracts for clear licensing 
- add makefile for easy make commands

refs https://github.com/zama-ai/fhevm-internal/issues/796

### DIFF
--- a/protocol-contracts/confidential-wrapper/hardhat.config.ts
+++ b/protocol-contracts/confidential-wrapper/hardhat.config.ts
@@ -79,7 +79,7 @@ const config: HardhatUserConfig = {
           // https://hardhat.org/hardhat-network/#solidity-optimizer-support
           optimizer: {
             enabled: true,
-            runs: 200,
+            runs: 800,
           },
           evmVersion: "cancun",
         },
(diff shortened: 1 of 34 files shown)


protocol-contracts/confidential-wrapper/hardhat.config.ts
```diff
@@ -66,21 +66,39 @@ const config: HardhatUserConfig = {
...
+          metadata: {
+            // Not including the metadata hash
+            // https://github.com/paulrberg/hardhat-template/issues/31
+            bytecodeHash: "none",
+          },
+          // Disable the optimizer when debugging
+          // https://hardhat.org/hardhat-network/#solidity-optimizer-support
+          optimizer: {
+            enabled: true,
+            runs: 200,
```
jatZama: I think it could be better to use `runs: 800` here. Like we did for host/gateway, etc.

melanciani: done, and i confirm it still comiles

created https://github.com/zama-ai/fhevm-internal/issues/800 for doing the same on staking

jatZama (approved):

PR #1610, opened 2025-12-18
feat(protocol-contracts): add ConfidentialTokensRegistry contract

melanciani:
closes https://github.com/zama-ai/fhevm-internal/issues/792


protocol-contracts/confidential-token-wrappers-registry/test/confidentialTokenWrappersRegistry.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      confidentialToken3,
+      registry: fixture.registry,
+    });
+  });
+
+  describe('Access Control', function () {
+    it('should not upgrade if not authorized', async function () {
+      await expect(
+        this.registry.connect(this.anyone).upgradeToAndCall(this.registry.target, '0x'),
+      ).to.be.revertedWithCustomError(this.registry, 'OwnableUnauthorizedAccount');
```
jatZama, line 48: nitpick, but it might be more sensible to call `upgradeToAndCall` with another address than current implementation ie `this.registry`, for clarity.

protocol-contracts/confidential-tokens-registry/contracts/ConfidentialTokensRegistry.sol
```diff
@@ -0,0 +1,193 @@
...
+        $._revokedConfidentialTokens[confidentialTokenAddress] = false;
+        emit ConfidentialTokenReinstated(confidentialTokenAddress);
+    }
+
+    /**
+     * @notice Returns the address of the confidential token associated with a token.
+     * @param tokenAddress The address of the token.
+     * @return The address of the confidential token.
+     */
+    function getConfidentialTokenAddress(address tokenAddress) public view returns (address) {
```
jatZama: For the 2 methods `getConfidentialTokenAddress` and `getTokenAddress`, I think it would make sense to: 
1/ revert the getters if there is nothing registered yet (i.e mappings to null address).
2/ make them return a tuple, (address token, bool isRevoked), in case there was already something registered in the past, but with a `isRevoked` to `true` if it has been revoked. This is what Morten had in mind I think, and returning a couple is better from a devX perspective, to not forget to handle the revoked edge case inside dapps relying on the registry.

melanciani: I see, did not think of that, I will do the changes yes

jatZama: Finally, for point 2/, I realize your idea is probably better of just returning an address, and have the boolean returned in another function.
I am still unsure about how to handle point 1/, should we revert or not...

jatZama: Maybe at least putting a comment in natspec, to explain if returned address is null, it means the token was never whitelisted.

melanciani: > it means the token was never whitelisted.

or that its confidential token has been revoked

protocol-contracts/confidential-token-wrappers-registry/test/confidentialTokenWrappersRegistry.test.ts
melanciani: done !

protocol-contracts/confidential-tokens-registry/contracts/ConfidentialTokensRegistry.sol
melanciani: I've added a comment !

protocol-contracts/confidential-tokens-registry/tasks/verify.ts
```diff
@@ -0,0 +1,49 @@
...
+    console.log(`Verifying confidential tokens registry implementation contract at ${implementationAddress}...\n`);
+    await run('verify:verify', {
+      address: implementationAddress,
+      constructorArguments: [],
+    });
+  });
+
+// Verify the confidential tokens registry contract using environment variables
+// Example usage:
+// npx hardhat task:verifyConfidentialTokensRegistryFromEnv --network testnet
```
jatZama: why not re-using same previous task but with an optional flag instead?

jatZama: you can have one unique task with 2 optional falgs : proxyAddress and fromEnv, and check that only one is present, and branch. But that's fine.

melanciani: done

protocol-contracts/confidential-tokens-registry/tasks/verify.ts
```diff
@@ -0,0 +1,55 @@
...
+task('task:verifyConfidentialTokensRegistryFromEnv').setAction(async function (_, hre) {
```
jatZama: Maybe you can remove this task now then? Since it can be replaced by previous one that you just updated it seems.

protocol-contracts/confidential-tokens-registry/contracts/ConfidentialTokensRegistry.sol
```diff
@@ -0,0 +1,197 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.27;
+
+import {Ownable2StepUpgradeable} from "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
+import {UUPSUpgradeable} from "@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol";
+
+/**
+ * @title ConfidentialTokensRegistry
+ * @notice A registry contract to map ERC20 token addresses to their ERC7984 confidential token addresses.
```
jatZama: Maybe you could add in natspecs above contract definition, that this should be used mainly to link a regular ERC20 token to its corresponding whitelisted confidential fhevm **wrapper** contract. There is no mention at all about confidential **wrappers** anywhere in current code, which could be confusing.

protocol-contracts/confidential-tokens-registry/contracts/ConfidentialTokensRegistry.sol
```diff
@@ -0,0 +1,197 @@
...
+
+        emit ConfidentialTokenRevoked(tokenAddress, confidentialTokenAddress);
+    }
+
+    /**
+     * @notice Reinstate an ERC7984 confidential token in case it was revoked. It does not restore
+     * the association with a token, it only allows it to be associated with one again.
+     * @param confidentialTokenAddress The address of the ERC7984 confidential token to reinstate.
+     */
+    function reinstateConfidentialToken(address confidentialTokenAddress) external onlyOwner {
```
jatZama: There is a weird state transition allowed by current implementation, when a confidential token is revoked and then reinstated, then it could be registered to a different ERC20 token, from which it was never registered before. Do we want to allow this? Are we ok with current design? It could be abused, but maybe we are fine with it, since we trust the owner to act honestly and never do this, but maybe at least then it should be documented if we don't add a safe guard in code.

protocol-contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol
```diff
@@ -0,0 +1,197 @@
...
+        $._confidentialTokensToTokens[confidentialTokenAddress] = tokenAddress;
+
+        emit ConfidentialTokenRegistered(tokenAddress, confidentialTokenAddress);
+    }
+
+    /**
+     * @notice Revoke an ERC7984 confidential token.
+     * @param confidentialTokenAddress The address of the ERC7984 confidential token to revoke.
+     */
+    function revokeConfidentialToken(address confidentialTokenAddress) external onlyOwner {
```
jatZama, line 142: Maybe we can also add a `revokeToken` function? Not sure about this tho.

jatZama:
I put few comments, but globally the PR looks very good. One of the suggestions I put was to add some mention of `wrapper` somewhere in the natspecs, for the sake of clarity.
The other main suggestion, was to add a check on `supportsInterface` on the confidential token, since EIP165 is a requirement to [EIP7984](https://eips.ethereum.org/EIPS/eip-7984), I think we should leverage it here.

protocol-contracts/confidential-tokens-registry/contracts/ConfidentialTokensRegistry.sol
```diff
@@ -0,0 +1,197 @@
...
+    function registerConfidentialToken(address tokenAddress, address confidentialTokenAddress) external onlyOwner {
+        if (tokenAddress == address(0)) {
+            revert TokenZeroAddress();
+        }
+        if (confidentialTokenAddress == address(0)) {
+            revert ConfidentialTokenZeroAddress();
+        }
+
+        // The confidential token must not be revoked.
+        if (isConfidentialTokenRevoked(confidentialTokenAddress)) {
```
jatZama: Can you please add a check on `supportsInterface(0x4958f2a4)` which should be return true on the confidential token, since EIP165 is a requirement to [EIP7984](https://eips.ethereum.org/EIPS/eip-7984), I think we should leverage it here. This should be handled with a try/catch, to handle reverts differently since in this case it means passed address does not implement the `supportsInterface` method probably.

jatZama: You should add this check `supportsInterface(0x4958f2a4)==true` in the register and revoke functions (and handling it with try/catch in case underlying contract does not implement it, to revert gracefully with another specific error).
Since there is a pending PR in the OZ library for the wrapper [PR](https://github.com/OpenZeppelin/openzeppelin-confidential-contracts/pull/267/files) adding an override to the supportsInterface function of the wrapper, maybe we could check both interfaces (ie supportsInterface(0x4958f2a4) &&  supportsInterface(wrapperInterfaceId)) instead of just the ERC7984 one, but I am not sure we want to be so restrictive in the registry, because it should be compatible with any kind of wrapper, not just the OZ one. (That being said, in the wrapper PR I would add this overrided method tho as commented there, but this is another topic).

protocol-contracts/confidential-tokens-registry/contracts/ConfidentialTokensRegistry.sol
melanciani: in the end we'll remove `reinstate`for now and we'll think about it in the future once we really need it, let's make things simple for now

protocol-contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol
melanciani: I don't think it makes much sense as it would mean you could imagine having a confidential wrapper defined in the registry but not associated to any token. We should really think them as pairs. It's for the same reason we only have one `register` function basically

protocol-contracts/confidential-tokens-registry/contracts/ConfidentialTokensRegistry.sol
melanciani: yes just added

I've only included this in the register function since the revoke function first makes sure the confidential token has been registered

protocol-contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol
```diff
@@ -3,18 +3,37 @@ pragma solidity ^0.8.27;
 
 import {Ownable2StepUpgradeable} from "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 import {UUPSUpgradeable} from "@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol";
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
 
 /**
  * @title ConfidentialTokensRegistry
- * @notice A registry contract to map ERC20 token addresses to their ERC7984 confidential token addresses.
+ * @notice A registry contract to map ERC20 token addresses to their corresponding ERC7984
+ * confidential fhevm wrapper addresses, also called confidential tokens.
```
jatZama: I find this natspec a bit confusing... I would have said:
 `/**
 * @title ConfidentialTokenWrappersRegistry
 * @notice A registry contract mapping ERC-20 tokens to their corresponding confidential FHEVM wrapper extensions of ERC-7984.`

protocol-contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol
```diff
@@ -27,6 +46,13 @@ contract ConfidentialTokensRegistry is Ownable2StepUpgradeable, UUPSUpgradeable
     /// @notice Error thrown when the confidential token address is zero.
     error ConfidentialTokenZeroAddress();
 
+    /// @notice Error thrown when a confidential token is not a valid ERC7984 confidential token
+    /// because it does not support the ERC7984 interface (0x4958f2a4).
+    error NotERC7984(address confidentialTokenAddress);
+
+    /// @notice Error thrown when a confidential token does not support ERC165's `supportsInterface`.
```
jatZama, line 53: This error name does not make sense, in the confidential token standard, it is a requirement to follow EIP165.

protocol-contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol
```diff
@@ -117,74 +150,95 @@ contract ConfidentialTokensRegistry is Ownable2StepUpgradeable, UUPSUpgradeable
...
     function _authorizeUpgrade(address newImplementation) internal override onlyOwner {}
 
+    /**
+     * @notice Validates that a confidential token supports the ERC7984 interface (0x4958f2a4) via
+     * `supportsInterface` from ERC165.
+     * See https://eips.ethereum.org/EIPS/eip-7984.
+     * @param confidentialTokenAddress The address of the confidential token to validate.
+     */
+    function _validateERC7984(address confidentialTokenAddress) internal view {
+        (bool success, bytes memory returnData) = confidentialTokenAddress.staticcall(
```
jatZama, line 226: As suggested in my previous comment, it would have been easier to handle this via a try-catch, not sure why you choosed this low level approach?

protocol-contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol
melanciani: yes but here the "confidential token" is the address given as an input, the error is about pointing out that the given confidential token address does not support erc 165 

maybe you can suggest another wording if you prefer ?

protocol-contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol
melanciani: with a try/catch , it was not catching when an address does not implement `supportsInterface` (it was instead throwing `function returned an unexpected amount of data`), at least here it is more explicit, no ?

protocol-contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol
jatZama: I would have suggested `AddressIsNotAnERC7984` for eg

protocol-contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol
jatZama: You are right, let's keep the low level call here.

jatZama (approved):

PR #1635, opened 2025-12-22
feat(protocol-contracts): replaced revoked by isValid, new view methods

jatZama:


protocol-contracts/confidential-token-wrappers-registry/.env.example
```diff
@@ -1,5 +1,5 @@
 # Authentication (use one of these)
-MNEMONIC=
+MNEMONIC="adapt mosquito move limb mobile illegal tree voyage juice mosquito burger raise father hope layer" # (string)
```
melanciani: you don't have to put it normally for tests

melanciani (changes requested):
yes please add new tests for your new features, but everything looks great !

melanciani (approved):
thanks a lot !!

protocol-contracts/confidential-token-wrappers-registry/.env.example
jatZama: done

PR #1636, opened 2025-12-22
chore(protocol-contracts): add verifyAllOperatorStakingContracts task

melanciani:
following changes from https://github.com/zama-ai/fhevm/pull/1629, I forgot that it's still useful to verify all contracts so that etherscan links the proxies with the implementations !

### DIFF
--- a/protocol-contracts/staking/tasks/verify.ts
+++ b/protocol-contracts/staking/tasks/verify.ts
@@ -59,31 +59,43 @@ task('task:verifyAllProtocolStakingContracts').setAction(async function (_, hre)
   }
 });
 
-// Verify a operator staking contract
+// Verify a single operator staking contract by proxy address
 // Example usage:
-// npx hardhat task:verifyOperatorStaking --network testnet
-task('task:verifyOperatorStaking').setAction(async function (_, hre) {
-  const { run, upgrades } = hre;
+// npx hardhat task:verifyOperatorStaking --proxyAddress 0x1234567890123456789012345678901234567890 --network testnet
+task('task:verifyOperatorStaking')
+  .addParam('proxyAddress', 'The address of the operator staking proxy contract to verify', '', types.string)
+  .setAction(async function ({ proxyAddress }, hre) {
+    const { run, upgrades } = hre;
 
-  // Get the first operator staking proxy address
-  // Since all operator staking contracts share the same implementation, we only have to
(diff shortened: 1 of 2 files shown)


protocol-contracts/staking/tasks/verify.ts
```diff
@@ -59,31 +59,43 @@ task('task:verifyAllProtocolStakingContracts').setAction(async function (_, hre)
...
-    address: implementationAddress,
-    constructorArguments: [],
-  });
+// Verify all operator staking contracts (proxies and implementations)
+// Since all operator staking contracts share the same implementation, we normally only have to
+// verify one of them. However, since they are proxied, verifying all of them has the benefit of linking
+// their proxies with their implementations on Etherscan.
+// Example usage:
+// npx hardhat task:verifyAllOperatorStakingContracts --network testnet
+task('task:verifyAllOperatorStakingContracts').setAction(async function (_, hre) {
```
jatZama, line 92: I am not sure you need this looping task here, since all instances will use same implementation.

melanciani: so I checked again with confidential wrapper contracts and I confirm that etherscan does not automatically link the proxies with the implementation contract although they already have been verified , ie I still needed to run a verify task to do the linking.

Examples of logs I just had : 
```
Verifying implementation: 0x1000...
The contract 0x1000... has already been verified on the block explorer. [...]

Verifying proxy: 0xA000...
Failed to verify ERC1967Proxy contract at 0xA000...: Already Verified
Linking proxy 0xA000... with implementation
Successfully linked proxy to implementation.
```

I guess it is related to what is said in this page https://info.etherscan.com/what-is-proxy-contract/ : 
> What if the implementation address was upgraded?
If the proxy contract’s implementation address had been upgraded, just re-run the verification process and the new implementation address will be noted and stored in our database.

jatZama: Ok so behavior is slightly different than on blockscout. But iirc, I think this could also be done easily manually in this case, in 1 or 2 clicks.

jatZama (approved):

protocol-contracts/staking/tasks/verify.ts
melanciani: yes indeed it's exactly what Etherscan says in the doc

PR #1647, opened 2025-12-24
docs(protocol-contracts): write first integration doc for wrappers

melanciani:
closes https://github.com/zama-ai/fhevm-internal/issues/818


protocol-contracts/confidential-wrapper-upgradeable/docs/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Important Considerations
+
+### Zero balance requirement
+
+Accounts with a zero balance that have never held tokens cannot be the `from` address in transfers. This will revert with `ERC7984ZeroBalance`.
+
+### Ciphertext uniqueness assumption
+
+The unwrap mechanism stores requests in a mapping keyed by ciphertext. The current implementation assumes ciphertexts are unique. This holds in practice but be aware of this architectural decision as it is not true in the general case.
```
jatZama: "This holds *in this very specific case* in practice.. etc"

You could also say it is because there is a sequential relationship between the computed burn amount handles due to their dependency on totalSupply, and due to how handles stemming from computations are computed under the hood, but maybe this is too much details?

protocol-contracts/confidential-wrapper-upgradeable/docs/README.md
```diff
@@ -0,0 +1,292 @@
...
+On top of the above unwrap request considerations:
+- `msg.sender` must be approved by ACL for the given `encryptedAmount` ⚠️ (see [ACL documentation](https://docs.zama.org/protocol/protocol/overview/library#access-control)).
+
+
+#### 2) Finalize unwrap
+
+> ℹ️ **Public decryption:** The encrypted burned amount emitted by the `UnwrapRequested` event must be publicly decrypted to get the `cleartextAmount` along its `decryptionProof`. More information in the [`relayer-sdk` documentation](https://docs.zama.org/protocol/relayer-sdk-guides/fhevm-relayer/decryption/public-decryption).
+
+```solidity
+wrapper.finalizeUnwrap(burntAmount, cleartextAmount, decryptionProof);
```
jatZama, line 91: I think it is important to say here that this burntAmount handle should be retrieved from previously emitted UnwrapRequest event here, to help integrators. Not obvious otherwise.

protocol-contracts/confidential-wrapper-upgradeable/docs/README.md
```diff
@@ -0,0 +1,292 @@
...
+|---------------------|------------------|------|--------|
+| 18 | 6 | 10^12 | 1 wrapped = 10^12 underlying |
+| 6 | 6 | 1 | 1:1 mapping |
+| 2 | 2 | 1 | 1:1 mapping |
+
+### Check supplies
+
+#### Non-confidential total supply
+
+The wrapper exposes a non-confidential view of the total supply, computed from the underlying ERC20 balance held by the wrapper contract. This value may be higher than `confidentialTotalSupply()` if tokens are sent directly to the wrapper outside of the wrapping process.
```
jatZama, line 162: Can you add that this view function could potentially be used to get a TVS total value shielded approximation useful for front ends?

protocol-contracts/confidential-wrapper-upgradeable/docs/README.md
```diff
@@ -0,0 +1,292 @@
...
+| `InvalidUnwrapRequest(amount)` | Finalizing non-existent unwrap request |
+| `ERC7984TotalSupplyOverflow()` | Minting would exceed uint64 max |
+
+---
+
+## Important Considerations
+
+### Zero balance requirement
+
+Accounts with a zero balance that have never held tokens cannot be the `from` address in transfers. This will revert with `ERC7984ZeroBalance`.
```
jatZama: Not sure why this error is separate and not in previous table with all other errors?

protocol-contracts/confidential-wrapper-upgradeable/docs/README.md
melanciani: I mean I already mention it just above : 
> The encrypted burned amount emitted by the `UnwrapRequested` event must be publicly

i'll put "burntAmount" in the text to be more clear, but do you want to emphasize even more on it ?

protocol-contracts/confidential-wrapper-upgradeable/docs/README.md
melanciani: it is part of the above errors ! but yes I'll move it to relevant sections

protocol-contracts/confidential-wrapper-upgradeable/docs/README.md
melanciani: in the end we've said we're not going to explain this in details because it's too specific, we'll maybe add a doc section based on this later: https://github.com/zama-ai/fhevm-internal/issues/819

jatZama (approved):

PR #1655, opened 2025-12-26
chore(coprocessor): disable default features for pinned bincode

Ocheretovich:
Explicitly disables default features for the pinned bincode dependency to avoid potential feature drift.
No version or behavior change intended.


Ocheretovich:
Re-trigger CLA check, please.

PR #1680, opened 2025-12-30
fix(kms-connector): reduce port collision in tests

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/825

I reduce the usage of the `pick_free_port` fn as much as possible, because I suspect it to be responsible for port collisions in integration tests due to race conditions (see our slack discussion for more details).

I still use this fn for healthcheck tests, because it is more convenient this way, and these tests are run "alone", so I think the race condition is less likely to happen.


dartdart26 (approved):

eudelins-zama:
@mergifyio queue

PR #1687, opened 2025-12-30
fix(protocol-contracts): add indexed parameters to staking events (N-05)

melanciani:
Note: I've only added indexes to `address` parameters

### DIFF
--- a/protocol-contracts/staking/contracts/OperatorRewarder.sol
+++ b/protocol-contracts/staking/contracts/OperatorRewarder.sol
@@ -40,7 +40,7 @@ contract OperatorRewarder {
      * @param oldBeneficiary The previous beneficiary address.
      * @param newBeneficiary The new beneficiary address.
      */
-    event BeneficiaryTransferred(address oldBeneficiary, address newBeneficiary);
+    event BeneficiaryTransferred(address indexed oldBeneficiary, address indexed newBeneficiary);
 
     /// @notice Emitted when the contract is shut down.
     event Shutdown();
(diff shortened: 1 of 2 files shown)


protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -36,19 +36,19 @@ contract OperatorRewarder {
...
     /// @notice Emitted when the beneficiary is transferred.
-    event BeneficiaryTransferred(address oldBeneficiary, address newBeneficiary);
+    event BeneficiaryTransferred(address indexed oldBeneficiary, address indexed newBeneficiary);
 
     /// @notice Emitted when the contract is shut down.
     event Shutdown();
 
     /// @notice Emitted when the maximum fee is updated.
-    event MaxFeeUpdated(uint16 oldFee, uint16 newFee);
+    event MaxFeeUpdated(uint16 indexed oldFee, uint16 indexed newFee);
```
jatZama, line 54: Do we really need to index fees arguments? I think it would make more sense to only index addresses in those events, since indexed arguments increase gas, and I am not entirely sure it makes sense to filter on fees, but it might make sense because these events (MaxFeeUpdated and FeeUpdated) should be rarely emitted, so you might keep them indexed, no strong opinion here.

melanciani: I also wasn't sure indeed but did it as this is what was recommended. If we feel there is no big benefit maybe it's better not to index them indeed

I behind the rationale behind this change is that these events are related to important configuration function that won't be often used indeed

melanciani: I removed the non-addresses ones

jatZama (approved):

PR #1688, opened 2025-12-30
fix(protocol-contracts): add named parameters to staking mappings (N-01)

melanciani:
Add named parameters to mapping declarations in OperatorRewarder and OperatorStaking contracts for improved code readability (Solidity 0.8.18+ feature).

### DIFF
--- a/protocol-contracts/staking/contracts/OperatorRewarder.sol
+++ b/protocol-contracts/staking/contracts/OperatorRewarder.sol
@@ -32,7 +32,7 @@ contract OperatorRewarder {
     uint256 private _lastClaimTotalAssetsPlusPaidRewards;
     uint256 private _totalRewardsPaid;
     int256 private _totalVirtualRewardsPaid;
-    mapping(address account => int256 rewardsPaid) private _rewardsPaid;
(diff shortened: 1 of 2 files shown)


jatZama (dismissed):

protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -32,8 +32,8 @@ contract OperatorRewarder {
     uint256 private _lastClaimTotalAssetsPlusPaidRewards;
     uint256 private _totalRewardsPaid;
     int256 private _totalVirtualRewardsPaid;
-    mapping(address => int256) private _rewardsPaid;
-    mapping(address => address) private _authorizedClaimers;
+    mapping(address account => int256 rewardsPaid) private _rewardsPaid;
```
jatZama: Wdyt of replacing the `account` key name here with `receiver` for clarity?
And for the same reason, you would replace the `account` argument of the `claimRewards` function below with `receiver`? It seems more explicit.

melanciani: yes good idea, done !

jatZama (approved):

PR #1689, opened 2025-12-30
fix(protocol-contracts): remove unused function parameters in staking (N-04)

melanciani:
Remove unused first parameter from `pendingRedeemRequest` and `claimableRedeemRequest` functions in OperatorStaking contract.


jatZama (approved):
I approve, since we no longer plan to follow the async vault standard, same as for the event update, but do not forget to update the front-end here as well.

PR #1691, opened 2025-12-30
fix(protocol-contracts): skip zero-amount stake/unstake operations (N-08)

melanciani:
NOTE: as opposed to the audit report, I introduced a revert in `unstake` instead of an early return like in `stake` because it is unclear what releaseTime we should consider then. We'll first need to confirm this with auditors

- Add early return for zero-amount stake in ProtocolStaking
- Add revert with ZeroUnstakeAmount error for zero-amount unstake in ProtocolStaking  
- Add revert with NoAssetsToWithdraw error in OperatorStaking.requestRedeem when no assets to withdraw
- Add tests for new behaviors

### DIFF
--- a/protocol-contracts/staking/contracts/OperatorStaking.sol
+++ b/protocol-contracts/staking/contracts/OperatorStaking.sol
@@ -75,6 +75,9 @@ contract OperatorStaking is ERC1363Upgradeable, ReentrancyGuardTransient, UUPSUp
     /// @dev Thrown when the controller address is not valid (e.g., zero address).
     error InvalidController();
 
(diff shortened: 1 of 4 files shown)


arr00:
This introduces a bug, and I recommend against taking this approach.

https://github.com/zama-ai/fhevm/blob/b8d5b24386bf859e2dd65920e9a5c5f1636f372e/protocol-contracts/staking/contracts/OperatorStaking.sol#L197

The line linked above makes it so that if there are excess funds in the `OperatorStaking` contract over the requested redemption, they must be restaked via `stakeExcess`.

melanciani:
we believe that addressing the issue does more harm than good, let's close the PR

PR #1692, opened 2025-12-30
fix(coprocessor): db migration, improve indexing for sns worker fetching work

rudy-6-4:


rudy-6-4:
This is to support
```sql
SELECT a.*, c.ciphertext
        FROM pbs_computations a
        JOIN ciphertexts c 
        ON a.handle = c.handle
        WHERE c.ciphertext IS NOT NULL
        AND a.is_completed = FALSE
        ORDER BY a.created_at {}
        FOR UPDATE SKIP LOCKED
        LIMIT $1;
```

antoniupop (approved):

PR #1693, opened 2025-12-30
fix(coprocessor): host-listener, always picking 1 past chain for connexity across block

rudy-6-4:
…


rudy-6-4:
no longer relevant

PR #1696, opened 2025-12-31
feat(coprocessor): add mechanism to release dependence chains when no progress

antoniupop:


coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -21,6 +21,7 @@ struct Transaction {
     output_tx: HashSet<TransactionHash>,
     linear_chain: TransactionHash,
     size: usize,
+    tx_topo_oder_in_block: usize,
```
rudy-6-4: Typo oder

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -469,6 +474,9 @@ pub async fn dependence_chains(
         let tx_hash = log.transaction_hash.unwrap_or_default();
         if let Some(tx) = txs.get(&tx_hash) {
             log.dependence_chain = tx.linear_chain;
+            log.block_timestamp = log.block_timestamp.saturating_add(
+                time::Duration::microseconds(tx.tx_topo_oder_in_block as i64),
+            );
```
rudy-6-4: Isn't it the same as before_size the same in last_updated_at ? We could keep just one of the 2 ?

antoniupop: I might have misunderstood - but this doesn't seem to work, I get the same timestamps so probably that evaluates to 0 all the time. If it's the same as last_updated_at, then we should reuse indeed - Can you please do the switch?

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
antoniupop: Looks like I can't spell anymore - I'll fix these tomorrow :+1:

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
rudy-6-4: offset 0 should be on new chain with no dependency inside the same block. (where you could have 0 with the top order)

antoniupop: Yes - I was trying to have each Tx in block ordered with an offset so we take them in a schedulable order.

If we have e.g. 200 tx in a block and they're all dependent, then if we take less than 200 in a batch, we can deadlock if we don't get the head of the chain by chance. That happened on testnet (which is why batch size is now set to 400, which is not ideal).

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -235,10 +237,13 @@ fn topological_order(ordered_txs: &mut Vec<Transaction>) {
     }
     ordered_txs.clear();
     debug!("Reordered txs: {:?}", reordered);
+    let mut topo_order_position = 0;
     for tx_hash in reordered.iter() {
-        let Some(tx) = txs.remove(tx_hash) else {
+        let Some(mut tx) = txs.remove(tx_hash) else {
             continue;
         };
+        tx.tx_topo_oder_in_block = topo_order_position;
```
rudy-6-4: on force un ordre entre transaction des chaines independantes, si un worker en prend plusieurs ça pourrait limiter le parallelisme entre les 2 chaines ?

rudy-6-4: tx.before_size est directement utilisable ici pour faire quelque chose d'équivalent (compte la chaine de dependence minimal pour atteindre la tx)

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
```diff
@@ -268,13 +270,18 @@ impl LockMngr {
...
+            // released from others in the same block.
+            let update_at = update_at.saturating_add(time::Duration::microseconds(1));
+            sqlx::query!(
             r#"
             UPDATE dependence_chain
             SET
                 worker_id = NULL,
                 lock_acquired_at = NULL,
                 lock_expires_at = NULL,
+                last_updated_at = $4::timestamp,
```
rudy-6-4, line 281: that field is supposed to be write-only for the host listener

rudy-6-4: is it safe (for worker side, HL doesn't care) that the worker overwrites a freshly updated value by HL ?

antoniupop: This is only written to if the worker found uncompleted computations in the chain, but it couldn't execute them. If that's the case, the HL will not have written a new value. Ideally the value written should be consistent between the two (created_at of the earliest un-executed computation in the chain and last_updated_at in dep_chain)

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
antoniupop: We only try to force the order in which computations are fetched from DB, not executed. If we fetch in a different order we're guaranteed to be at best as good as this order, but often we'll fetch uncomputable computations because we don't fetch in an order that's computable.

antoniupop: If tx.before_size is usable, then sure - we just need to propagate this to the DB in computations

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -487,17 +488,14 @@ WHERE c.transaction_id IN (
                     inputs,
                     is_allowed: w.is_allowed,
                 });
-                if w.created_at < earliest_created_at {
-                    earliest_created_at = w.created_at;
-                }
             }
```
antoniupop, line 488: I don't think we can drop the update of the earliest as the query is not guaranteed to return elements in order. The order by clause is only on the inner select that selects the transaction set.

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -132,11 +133,31 @@ async fn fill_tx_dependence_maps(
...
+                        let mut h = HashSet::new();
+                        h.insert(tx.tx_hash);
+                        h
+                    });
             } else if let Some(dep_tx_hash) =
                 past_chains.write().await.get(input_handle)
             {
                 // extra block, this is directly a chain hash
                 tx.input_tx.insert(*dep_tx_hash);
+                used_txs_chains
```
rudy-6-4, line 151: how is it different from txs.get(dep_hash).output_tx set ?
txA.output_tx is all other tx using txA ?

rudy-6-4: ok it needs to be mutable after

rudy-6-4 (approved):

PR #1698, opened 2025-12-31
chore(kms-connector): simplify config parsing

eudelins-zama:
Refacto of the config parsing.

Removed the `RawConfig` deserialization step which was then "parsed" into the final `Config` struct.
Now, the `Config` is directly deserialized and validated using `serde_with` annotation.
It avoids having to update multiple files when we want to add a new field in the config.


dartdart26 (approved):
Nice!

eudelins-zama:
@mergifyio queue

PR #1701, opened 2026-01-02
fix(protocol-contracts): round up unpaid fee to prevent extra reward distribution

melanciani:

### DIFF
--- a/protocol-contracts/staking/contracts/OperatorRewarder.sol
+++ b/protocol-contracts/staking/contracts/OperatorRewarder.sol
@@ -419,7 +419,13 @@ contract OperatorRewarder {
 
     function _unpaidFee(uint256 totalAssetsPlusPaidRewards) internal view returns (uint256) {
         uint256 totalAssetsPlusPaidRewardsDelta = totalAssetsPlusPaidRewards - _lastClaimTotalAssetsPlusPaidRewards;
-        return (totalAssetsPlusPaidRewardsDelta * feeBasisPoints()) / 10_000;
+
+        // Use ceiling instead of floor to avoid extra rewarding delegators when computing allocations.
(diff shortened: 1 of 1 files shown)


protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -419,7 +419,13 @@ contract OperatorRewarder {
...
     function _unpaidFee(uint256 totalAssetsPlusPaidRewards) internal view returns (uint256) {
         uint256 totalAssetsPlusPaidRewardsDelta = totalAssetsPlusPaidRewards - _lastClaimTotalAssetsPlusPaidRewards;
-        return (totalAssetsPlusPaidRewardsDelta * feeBasisPoints()) / 10_000;
+
+        // Use ceiling instead of floor to avoid extra rewarding delegators when computing allocations.
+        // Although floor is the correct behavior for calculating the fee to pay to the beneficiary,
+        // delegator allocations would be slightly higher than expected as it is based on total
+        // accumulated rewards subtracted by this unpaid fee amount, reversing the rounding direction.
+        // This could prevent the last depositor (or the beneficiary) from claiming their rewards.
+        return totalAssetsPlusPaidRewardsDelta.mulDiv(feeBasisPoints(), 10_000, Math.Rounding.Ceil);
```
jatZama, line 428: This is not what OZ suggested for the fix. You should keep original rounding down behaviour, because this is the correct behaviour for the _claimFee function.
On the other hand, you should create a new function variant called `_unpaidFeeRoundUp` or `_unpaidFeeCeil` for eg, which contain this formula with the rounding up, and which would be used to compute the `historicalReward`.

melanciani:
issue has been closed as false positive, closing this PR

PR #1715, opened 2026-01-05
fix(coprocessor): add missing indexes on verify_proofs and dependence_chain tables

antoniupop:


melanciani:
typo in the title !

coprocessor/fhevm-engine/db-migration/migrations/20260105120000_dependence_chain_proofs_indexing.sql
```diff
@@ -0,0 +1,23 @@
+--- Improve indexing for verify_proofs table
+
+CREATE INDEX IF NOT EXISTS idx_verify_proofs_verified_not_null
+ON verify_proofs (retry_count)
+WHERE verified IS NOT NULL;
+
+CREATE INDEX IF NOT EXISTS idx_verify_proofs_verified_null
+ON verify_proofs (zk_proof_id)
+WHERE verified IS NULL;
```
dartdart26: Which query is this targeting? `zk_proof_id` is a primary key, no? I mean do we need this additional index?

dartdart26 (commented):
Looks good! I just have a question re `idx_verify_proofs_verified_null` and whether we actually need it.

coprocessor/fhevm-engine/db-migration/migrations/20260105120000_dependence_chain_proofs_indexing.sql
antoniupop: It's for the 2 queries that filter by `verified is null` which should be more efficient (but I may very well be wrong - maybe an index on `verify_proos(verified) where verified is null` is what we want...

dartdart26: But I guess we need an index on `verified`, no?

dartdart26: I guess we want to drop this index then:

```CREATE INDEX IF NOT EXISTS idx_verify_proofs_verified_retry ON verify_proofs(verified, retry_count, zk_proof_id);```

Do you know which query has a sequential scan?

Comment (dartdart26) [reply coprocessor/fhevm-engine/db-migration/migrations/20260105120000_dependence_chain_proofs_indexing.sql]: I am looking at it deeper, let's chat about it tomorrow.

File: coprocessor/fhevm-engine/db-migration/migrations/20260105120000_dependence_chain_proofs_indexing.sql [hunk 2]
```diff
@@ -0,0 +1,23 @@
+--- Improve indexing for verify_proofs table
+
+CREATE INDEX IF NOT EXISTS idx_verify_proofs_verified_not_null
+ON verify_proofs (retry_count)
+WHERE verified IS NOT NULL;
```
dartdart26: We do have this query too:

```
 "DELETE FROM verify_proofs WHERE retry_count >= $1",
```

So maybe we can omit `WHERE verified IS NOT NULL`?

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/db-migration/migrations/20260105120000_dependence_chain_proofs_indexing.sql
antoniupop: Oh yes that may be better then

coprocessor/fhevm-engine/db-migration/migrations/20260105120000_dependence_chain_proofs_indexing.sql
antoniupop: I removed it for now.

coprocessor/fhevm-engine/db-migration/migrations/20260105120000_dependence_chain_proofs_indexing.sql
antoniupop: Done

dartdart26 (approved):
Looks good!

PR #1717, opened 2026-01-05
fix(protocol-contracts): add start() in OperatorRewarder (H-01)

melanciani:

### DIFF
--- a/protocol-contracts/staking/contracts/OperatorRewarder.sol
+++ b/protocol-contracts/staking/contracts/OperatorRewarder.sol
@@ -29,7 +29,7 @@ contract OperatorRewarder {
     uint16 private _maxFeeBasisPoints;
     uint16 private _feeBasisPoints;
     bool private _shutdown;
-    bool private _initialized;
+    bool private _started;
     uint256 private _lastClaimTotalAssetsPlusPaidRewards;
     uint256 private _totalRewardsPaid;
     int256 private _totalVirtualRewardsPaid;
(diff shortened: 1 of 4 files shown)


jatZama:
I would replace `initialize` with `startup` (opposite of already existing `shutdown`) or `start` name, to avoid confusion with the proxy initializer.

melanciani:
yes indeed this completely went over my head, I renamed to `start`

protocol-contracts/staking/contracts/OperatorRewarder.sol
```diff
@@ -241,7 +275,7 @@ contract OperatorRewarder {
      * @param to Recipient address.
      * @param shares Number of shares transferred.
      */
-    function transferHook(address from, address to, uint256 shares) public virtual onlyOperatorStaking {
+    function transferHook(address from, address to, uint256 shares) public virtual whenStarted onlyOperatorStaking {
```
jatZama: You don't need this `whenStarted` modifier for the `transferHook` function, since it is called only via the `OperatorStaking` contract.

protocol-contracts/staking/test/OperatorStaking.test.ts
```diff
@@ -642,14 +640,34 @@ describe('OperatorStaking', function () {
         await this.mock.connect(this.delegator2).deposit(ethers.parseEther('3'), this.delegator2);
         await timeIncreaseNoMine(10);
 
+        // Deploy the new rewarder contract after earnings have been accumulated in the old rewarder
+        // to better reflect the "stake state" test
```
jatZama: I think there is a typo, you probably meant "stale state" instead of "stake state".

protocol-contracts/staking/contracts/OperatorRewarder.sol
melanciani: indeed !

jatZama (approved):

PR #1723, opened 2026-01-06
fix(coprocessor): host-listener, add timeout to get logs calls

rudy-6-4:
From a case seen on testnet where the host-listener was stuck by alloy looping on bad node message. The timeout will cut that loop and force host-listener to retry and reconnect.

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
@@ -40,6 +40,8 @@ pub const DEFAULT_DEPENDENCE_CACHE_SIZE: u16 = 10_000;
 pub const DEFAULT_DEPENDENCE_BY_CONNEXITY: bool = false;
 pub const DEFAULT_DEPENDENCE_CROSS_BLOCK: bool = true;
 
+const TIMEOUT_REQUEST_ON_WEBSOCKET: u64 = 15;
+
 #[derive(Parser, Debug, Clone)]
 #[command(version, about, long_about = None)]
 pub struct Args {
@@ -156,6 +158,13 @@ pub struct Args {
         help = "Sleep duration in seconds between catchup loop iterations"
     )]
     pub catchup_loop_sleep_secs: u64,
+
+    #[arg(
+        long,
+        default_value_t = TIMEOUT_REQUEST_ON_WEBSOCKET,
+        help = "Timeout in seconds for RPC calls over websocket"
+    )]
+    pub timeout_request_websocket: u64,
 }
 
 // TODO: to merge with Levent works
@@ -180,6 +189,7 @@ struct InfiniteLogIter {
     reorg_maximum_duration_in_blocks: u64, // in blocks
     block_history: BlockHistory,           // to detect reorgs
     catchup_finalization_in_blocks: u64,
+    timeout_request_websocket: u64,
 }
 
 enum BlockOrTimeoutOrNone {
@@ -243,6 +253,7 @@ impl InfiniteLogIter {
                 args.reorg_maximum_duration_in_blocks as usize,
             ),
             catchup_finalization_in_blocks: args.catchup_finalization_in_blocks,
+            timeout_request_websocket: args.timeout_request_websocket,
         }
     }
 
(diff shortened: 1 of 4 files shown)


coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -37,6 +37,8 @@ const RETRY_GET_BLOCK_DELAY_IN_MS: u64 = 100;
 
 const DEFAULT_BLOCK_TIME: u64 = 12;
 
+const TIMEOUT_GET_LOGS_WEBSOCKET: u64 = 120;
```
obatirou: How this was chosen? 
Should this be configurable?

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -296,13 +298,27 @@ impl InfiniteLogIter {
...
         };
-        provider.get_logs(&filter).await.map_err(|err| {
-            if eth_rpc_err::too_much_blocks_or_events(&err) {
-                anyhow::anyhow!("Too much blocks or events: {err}")
-            } else {
-                anyhow::anyhow!("Cannot get logs for {filter:?} due to {err}")
+        // Timeout to prevent hanging indefinitely on buggy node
+        match tokio::time::timeout(
+            Duration::from_secs(TIMEOUT_GET_LOGS_WEBSOCKET),
+            provider.get_logs(&filter),
```
obatirou, line 330: Do other functions called on provider could be in the same case?

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
rudy-6-4: Made it configurable and shorter (default to 30s).

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
rudy-6-4: Yes there are some other functions in the most frequent path that would benefit a timeout as well (e.g. get_current_block, get_block_by_hash). I have added them and added a provider+stream reset in case it happens in the block timeout case (eg when new block is more than 2 second late + alloy is stuck)

rudy-6-4:
@claude /pr-review

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
obatirou, line 264: Is this one ok to not have timeouts?
Same for
* resolve_end_at_block
* catchup_block_from

obatirou: this one has no timeout and panics on failure, is this expected?

obatirou (dismissed):

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
rudy-6-4: I didn't make this ones because they were only at startup/reconnect and would imply introducing new error propagation in new_log_stream. But that's clearly a good idea, I'll have a second look to see how to extend to new_log_stream. I think it could be in a next PR to merge that part.

rudy-6-4: No that a long standing initial bug. Could go in a PR introducing new_log_stream error propagation.

rudy-6-4: There is also a panic after many retry.

rudy-6-4: Ok convinced I'm changing the retry loop to handle all kind of error. Note that panic is only at startup and is on  purpose.

obatirou: Should this be done in an other PR?

antoniupop (dismissed):

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
rudy-6-4: Yes. https://github.com/zama-ai/fhevm/pull/1857

obatirou (approved):

rudy-6-4:
@mergifyio queue

rudy-6-4:
@mergifyio queue

rudy-6-4:
@mergifyio queue

eudelins-zama:
@mergifyio requeue

PR #1748, opened 2026-01-08
fix(gateway-contracts): improve forge version checking with range support

enitrat:
## Summary
- Replace hardcoded list of allowed forge versions with flexible version range check
- Use simple tuple comparison for semver validation (no external dependencies)
- Accept all forge 1.x versions >= 1.3.1

## Context
The previous implementation used a hardcoded list `ALLOWED_FORGE_VERSIONS` containing specific version strings like "1.3.1-v1.3.1", "1.3.1-stable", "1.3.2-stable". This approach:
- Required updating the script for every new forge patch release
- Made it difficult to support multiple version formats
- Was inflexible for accepting newer compatible versions

## Changes
This PR improves the forge version validation by:
1. Defining `MIN_FORGE_VERSION = (1, 3, 1)` and `MAX_FORGE_VERSION = (2, 0, 0)` as tuple constants
2. Adding a simple `parse_semver()` helper to convert version strings to tuples
3. Using tuple comparison to check `>= 1.3.1` and `< 2.0.0`

## Benefits
- **Zero dependencies**: Uses Python built-in tuple comparison instead of external `packaging` library
- **Low maintenance**: Accepts any forge 1.x version >= 1.3.1 without needing updates
- **Semver-aware**: Protects against major version breaking changes (2.x)
- **Simple**: Easy to understand and maintain

## Test Plan
- [x] Tested with forge version "1.3.1-stable" - passes
- [x] Tested with forge version "1.4.0" - passes
- [x] Tested with forge version "1.9.0" - passes
- [x] Tested with forge version "2.0.0" - fails with appropriate error message
- [x] Tested with forge version "1.3.0" - fails with appropriate error message

### DIFF
--- a/gateway-contracts/scripts/bindings_update.py
+++ b/gateway-contracts/scripts/bindings_update.py
@@ -10,7 +10,6 @@ import tempfile
 from argparse import ArgumentParser
 from enum import Enum
 from pathlib import Path
-from packaging import version
 
 GW_ROOT_DIR = Path(os.path.dirname(__file__)).parent
 GW_CRATE_DIR = GW_ROOT_DIR.joinpath("rust_bindings")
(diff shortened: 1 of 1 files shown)


eudelins-zama (approved):
LGTM, thx!

enitrat:
@mergify queue

enitrat:
@mergify queue

PR #1757, opened 2026-01-09
fix(coprocessor): persist computed ct128 as local backup

goshawk-3:
Persist computed ct128 values in a dedicated `ciphertexts128` table as a failover mechanism when S3 uploads fail and require retry.

Entries are deleted after a successful L2 `add_ciphertext` call. As an additional safeguard, a periodic GC procedure removes stale entries in cases where the txn-sender fails to clean up as expected.

fixes: https://github.com/zama-ai/fhevm-internal/issues/761
fixes: https://github.com/zama-ai/fhevm-internal/issues/865


antoniupop (dismissed):

dartdart26:
@goshawk-3 Is this still relevant?

goshawk-3:
> @goshawk-3 Is this still relevant?

It is no longer relevant, as it has been merged into main ref: https://github.com/zama-ai/fhevm/pull/1790.

PR #1763, opened 2026-01-09
fix(kms-connector): add prometheus content-type header

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/856

### DIFF
--- a/kms-connector/crates/utils/src/monitoring/otlp.rs
+++ b/kms-connector/crates/utils/src/monitoring/otlp.rs
@@ -1,4 +1,4 @@
-use actix_web::{HttpResponse, http::StatusCode};
+use actix_web::HttpResponse;
 use anyhow::anyhow;
 use opentelemetry::{
     global,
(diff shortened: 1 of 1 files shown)


kms-connector/crates/utils/src/monitoring/otlp.rs
```diff
@@ -51,13 +51,20 @@ pub fn init_otlp_setup(service_name: String) -> anyhow::Result<()> {
     Ok(())
 }
 
+/// The content type header for Prometheus text-based metrics format.
+///
+/// For more details, check https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format
+const PROMETHEUS_TEXT_FORMAT_HEADER: (&str, &str) = ("Content-Type", "text/plain; version=0.0.4");
```
dartdart26: Is version mandatory?

dartdart26 (dismissed):

kms-connector/crates/utils/src/monitoring/otlp.rs
eudelins-zama: According to the docs: 'A missing version value will lead to a fall-back to the most recent text format version".

So I guess it's safer to keep this version for now

dartdart26: I don't know what that means, but if they say we need to use it - let's keep it.

dartdart26 (approved):

kms-connector/crates/utils/src/monitoring/otlp.rs
eudelins-zama: Ok so in the end, the `prometheus` crate is exposing this version, so let's use it instead of hard-coding it ourselves

eudelins-zama:
@mergifyio queue

PR #1787, opened 2026-01-15
fix(coprocessor): gw-listener, recycle provider preventively inactivity

rudy-6-4:


rudy-6-4:
https://github.com/zama-ai/fhevm/pull/1788

PR #1788, opened 2026-01-15
fix(coprocessor): verify input  via eth_getLogs

dartdart26:
Instead of relying on subscriptions that are not reliable in all cases, use eth_getLogs to fetch verify input requests.

Also, change:
 * poll interval from 1 sec to 500 ms to reduce latency
 
Since gw-listener catches up automatically via the last processed block
number in the DB, call the manual process "replay" instead of "catchup".
Replay is requested via the cmd line. Also, keep the old name as an
alias for backward compatibility. Furthermore, replay is not
KMS-specific anymore, so remove KMS from the name.

Add more logging in health checks and replay.

### DIFF
--- a/coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
+++ b/coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
@@ -88,11 +88,11 @@ impl<P: Provider<Ethereum> + Clone + 'static, A: AwsS3Interface + Clone + 'stati
             let s = self.clone();
             let d = db_pool.clone();
             tokio::spawn(async move {
-                let mut catchup_kms_generation_from = s.conf.catchup_kms_generation_from_block;
+                let mut replay_from_block = s.conf.replay_from_block;
                 let mut sleep_duration = s.conf.error_sleep_initial_secs as u64;
                 loop {
                     match s
-                        .run_get_logs(&d, &mut sleep_duration, &mut catchup_kms_generation_from)
+                        .run_get_logs(&d, &mut sleep_duration, &mut replay_from_block)
                         .await
                     {
                         Ok(_) => {
@@ -117,12 +117,13 @@ impl<P: Provider<Ethereum> + Clone + 'static, A: AwsS3Interface + Clone + 'stati
         &self,
         db_pool: &Pool<Postgres>,
         sleep_duration: &mut u64,
-        catchup_kms_generation_from: &mut Option<i64>,
+        replay_from_block: &mut Option<i64>,
     ) -> anyhow::Result<()> {
         let mut ticker = tokio::time::interval(self.conf.get_logs_poll_interval);
         let mut last_processed_block_num = self.get_last_processed_block_num(db_pool).await?;
-        if let Some(from_block) = *catchup_kms_generation_from {
-            info!(from_block, "Catchup on KMSGeneration, start");
+        let mut number_of_last_processed_updates: u64 = 0;
+        if let Some(from_block) = *replay_from_block {
+            info!(from_block, "Replay starts");
             let from_block = if from_block >= 0 {
                 // start from specified block
                 from_block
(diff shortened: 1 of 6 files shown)


coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
```diff
@@ -232,7 +166,7 @@ impl<P: Provider<Ethereum> + Clone + 'static, A: AwsS3Interface + Clone + 'stati
                     };
 
                     let filter = Filter::new()
-                        .address(self.kms_generation_address)
+                        .address(vec![self.kms_generation_address, self.input_verification_address])
```
rudy-6-4, line 174: Nice reuse

dartdart26: You do the same in the host listener basically, right?

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
```diff
@@ -250,47 +184,66 @@ impl<P: Provider<Ethereum> + Clone + 'static, A: AwsS3Interface + Clone + 'stati
...
-                                        }
-                                        Err(e) => {
-                                            ACTIVATE_CRS_FAIL_COUNTER.inc();
-                                            return Err(e);
-                                        }
+                        if log.address() == self.input_verification_address {
+                            if let Ok(event) = InputVerification::InputVerificationEvents::decode_log(&log.inner) {
+                                match event.data {
+                                    InputVerification::InputVerificationEvents::VerifyProofRequest(request) => {
+                                        self.verify_proof_request(db_pool, request, log.clone()).await?;
```
rudy-6-4: Need some retry here but can be in a next PR.

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
dartdart26: Since we have `?` at the end, it will return and it won't update the last processed block and we will do it again, no?

Eikix (dismissed):
Great! This is way better and reuses code from KMS logic which is lighter and nicer.

We can check with Q&A if this change feels good user-side (wrt 500ms polling intervals)

EDIT: Consider whether individual event processing failures should be isolated vs failing the entire batch

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
dartdart26: I mean reasoning here is that we just insert into the DB. If we can't do that, we must try again as nothing else will work.

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
```diff
@@ -250,47 +184,66 @@ impl<P: Provider<Ethereum> + Clone + 'static, A: AwsS3Interface + Clone + 'stati
...
-                                            key_digest_mismatch += 1;
-                                            error!(error = %e, "Key digest mismatch, ignoring event");
-                                        }
-                                        Err(e) => {
-                                            ACTIVATE_KEY_FAIL_COUNTER.inc();
-                                            return Err(e);
+                                }
+                            } else {
+                                error!(log = ?log, "Failed to decode InputVerification event log");
+                            }
```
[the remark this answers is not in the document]
dartdart26: InputVerification just inserts into the DB and reasoning is that if we can't write to the DB, nothing else will work anyway.

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
rudy-6-4: If there is a retry at higher level it's ok, but we should t loose an event without retries

dartdart26: If something fails, we never update the last processed block (unless it is an error we can't recover from like bad digest for KMS events). For input, if it fails, we will now try again as the only way for it to fails for the DB to be down.

antoniupop (approved):

maxnovawind (approved):

rudy-6-4 (approved):

dartdart26:
@Mergifyio queue

PR #1790, opened 2026-01-16
fix(coprocessor): persist computed ct128 as local backup

goshawk-3:
Persist computed ct128 values in a dedicated `ciphertexts128` table as a failover mechanism when S3 uploads fail and require retry.

Entries are deleted after a successful L2 `add_ciphertext` call. As an additional safeguard, a periodic GC procedure removes stale entries in cases where the txn-sender fails to clean up as expected.

fixes: https://github.com/zama-ai/fhevm-internal/issues/865


coprocessor/fhevm-engine/db-migration/migrations/20260106150619_create_ciphertexts128_table.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE ciphertexts128 (
+    tenant_id   INTEGER   NOT NULL,
+    handle      BYTEA     NOT NULL,
+    ciphertext  BYTEA     NULL,
+    created_at  TIMESTAMP NOT NULL DEFAULT NOW(),
+
+    PRIMARY KEY (tenant_id, handle)
+);
```
antoniupop, line 8: Nice - do we add a separate migration for deleting not the column but at least the very old CT128 from DB or should we plan to do it by hand?

antoniupop (approved):

coprocessor/fhevm-engine/db-migration/migrations/20260106150619_create_ciphertexts128_table.sql
goshawk-3: AFAIK, without dropping the column, we cannot physically reclaim storage from very old ct128 values. We can only update them to NULL, and the on-disk storage will not be reclaimed unless a VACUUM FULL is done.

I suggest we release this fix in 0.11.0 and later we drop the column in 0.11.1 (when it's safe to do so)

I'll file an issue for `drop column` migration.

antoniupop: Yes please add the issue, we need to make sure we don't forget.

goshawk-3:
@mergifyio queue

PR #1795, opened 2026-01-16
fix(coprocessor): verify input via eth_getLogs

dartdart26:
Instead of relying on subscriptions that are not reliable in all cases, use eth_getLogs to fetch verify input requests.

Also, change:
 * poll interval from 1 sec to 500 ms to reduce latency
 
Since gw-listener catches up automatically via the last processed block number in the DB, call the manual process "replay" instead of "catchup". Replay is requested via the cmd line. Also, keep the old name as an alias for backward compatibility. Furthermore, replay is not KMS-specific anymore, so remove KMS from the name.

Add more logging in health checks and replay.

### DIFF
--- a/.github/workflows/coprocessor-dependency-analysis.yml
+++ b/.github/workflows/coprocessor-dependency-analysis.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Install cargo tools
         run: |
           cargo binstall --no-confirm --force \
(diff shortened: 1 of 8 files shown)


antoniupop (dismissed):

Eikix:
Already on main, we can close this

PR #1796, opened 2026-01-16
feat(protocol-contracts): add bnb configs

jatZama:


protocol-contracts/token/scripts/copyAbiToDeployments.ts
```diff
@@ -0,0 +1,103 @@
...
+function loadJson<T>(filePath: string): T {
+  const content = fs.readFileSync(filePath, 'utf-8');
+  return JSON.parse(content) as T;
+}
+
+function saveJson(filePath: string, data: unknown): void {
+  fs.writeFileSync(filePath, JSON.stringify(data, null, 2) + '\n');
+}
+
+function isDeploymentFile(fileName: string): boolean {
```
melanciani, line 31: these conditions seem a bit too broad, but I get the point of it and its usage looks good so I guess it's fine

protocol-contracts/token/tasks/sendOFT.ts
```diff
@@ -34,7 +34,7 @@ task('lz:oft:send', 'Sends OFT tokens cross‐chain from EVM chains')
     .addParam('dstEid', 'Destination endpoint ID', undefined, types.int)
     .addParam('amount', 'Amount to send (human readable units, e.g. "1.5")', undefined, types.string)
     .addParam('to', 'Recipient address (20-byte hex for EVM)', undefined, types.string)
-    .addOptionalParam('oappConfig', 'Path to LayerZero config file', 'layerzero.config.testnet.ts', types.string)
+    .addOptionalParam('oappConfig', 'Path to LayerZero config file', 'layerzero.config.mainnet.ts', types.string)
```
melanciani: I think would be best to keep default values as testnet no ? to avoid any unwanted issues

protocol-contracts/token/ADD_BNB_OFT.md
```diff
@@ -0,0 +1,146 @@
...
+
+```
+npx ts-node scripts/convertToAragonProposal.ts ethereum-wiring.json aragonProposal.json
+```
+
+This will convert the `ethereum-wiring.json` file to a new `aragonProposal.json` which could be directly uploaded inside the Aragon App UI, when creating an Aragon proposal, to streamling the process.
+
+More precisely, in Aragon App, when you reach "Step 2 of 3" of proposal creation, click on the `Upload` button there, in select the newly created `aragonProposal.json` file to upload it and create the wiring proposal on Ethereum.
+
+![Aragon Proposal Upload](AragonUpload.png)
```
melanciani: could you create a `img` dir or something similar to avoid having images at the root of the directory ?

protocol-contracts/token/layerzero.config.mainnet.bnb.ts
```diff
@@ -0,0 +1,67 @@
...
+        value: 0,
+    },
+]
+
+// With the config generator, pathways declared are automatically bidirectional
+// i.e. if you declare A,B there's no need to declare B,A
+const pathways: TwoWayConfig[] = [
+    [
+        ethereumContract, // Chain A contract
+        zamaMainnetContract, // Chain B contract
```
melanciani, line 44: if this is for BNB mainnet, why do we still need to write the gateway link ? is it because this config will essentially "override" the old one ?

protocol-contracts/token/layerzero.config.testnet.bnb.ts
```diff
@@ -0,0 +1,66 @@
...
+        value: 0,
+    },
+]
+
+// With the config generator, pathways declared are automatically bidirectional
+// i.e. if you declare A,B there's no need to declare B,A
+const pathways: TwoWayConfig[] = [
+    [
+        sepoliaContract, // Chain A contract
+        zamaTestnetContract, // Chain B contract
```
melanciani, line 44: same question

protocol-contracts/token/ADD_BNB_OFT.md
jatZama: I added it in an images/ folder.

protocol-contracts/token/layerzero.config.mainnet.bnb.ts
jatZama: yes, it can be useful if we want to redeploy also from scratch the whole current setup. I feel it is cleaner to have some place summarizing the whole config, even if it is not strictly needed.

protocol-contracts/token/layerzero.config.testnet.bnb.ts
jatZama: see my answer to previous comment

protocol-contracts/token/tasks/sendOFT.ts
jatZama: Ok I reverted to testnet default in my last commit.

protocol-contracts/token/scripts/copyAbiToDeployments.ts
jatZama: There is no big risk here, those artifacts are automatically generated or seeded via the ADD_BNB_OFT guide.

melanciani (approved):
thanks a lot !!

PR #1794, opened 2026-01-16
feat(kms-connector): add ACL verification for decryption requests

enitrat:
## Summary

- Add host chain ACL verification to KMS Connector before processing decryption requests
- KMS now independently verifies authorization by calling view functions on the host chain ACL contract

### New Features
- **Host chain configuration**: Added `HostChainConfig` struct with RPC URL, chain ID, and ACL contract address
- **Public decryption ACL check**: Calls `isAllowedForDecryption(handle)` for each handle
- **User decryption ACL check**: Calls `isAllowed(handle, userAddress)` for the requester

### Error Handling
- ACL denials → `Irrecoverable` error (immediate failure, no retry)
- Network errors → `Recoverable` error (retry with counter, max 200 attempts)

### Files Modified
| File | Changes |
|------|---------|
| `config.rs` | Added `HostChainConfig` struct |
| `decryption.rs` | Added `check_acl_for_public_decryption()` and `check_acl_for_user_decryption()` |
| `processor.rs` | Integrated ACL checks, fixed error counter increment |
| `kms_worker.rs` | Added host chain provider initialization |
| `kms-worker.toml` | Added `[host_chain]` configuration section |

## Test Plan

- [x] Unit tests for ACL authorization (6 parametrized cases)
- [x] Unit test for decryption-not-already-done check
- [x] Integration tests for ACL denied scenarios
- [x] Integration tests for ACL transport error scenarios
- [x] Config loading tests updated

```
cargo test -p kms-worker --lib
test result: ok. 14 passed; 0 failed
```

## Related Issues

Closes #849

PR #1797, opened 2026-01-19
feat(kms-connector): acl check for decryption

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/849
Closes https://github.com/zama-ai/fhevm-internal/issues/890


kms-connector/Cargo.toml
```diff
@@ -18,7 +18,8 @@ gw-listener.path = "crates/gw-listener"
 kms-worker.path = "crates/kms-worker"
 tx-sender.path = "crates/tx-sender"
 connector-utils.path = "crates/utils"
-fhevm_gateway_bindings = { git = "https://github.com/zama-ai/fhevm.git", tag = "v0.10.3", default-features = false }
+fhevm_gateway_bindings = { git = "https://github.com/zama-ai/fhevm.git", rev = "96f31548861bedbf9b3d99f66e42364cebb69ba9", default-features = false }
+fhevm_host_bindings = { git = "https://github.com/zama-ai/fhevm.git", rev = "96f31548861bedbf9b3d99f66e42364cebb69ba9", default-features = false }
```
enitrat, line 22: why are we not using local paths to monorepo here?

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
```diff
@@ -138,6 +146,12 @@ impl<P: Provider> DbEventProcessor<P> {
             GatewayEventKind::UserDecryption(req) => {
                 // No need to check decryption is done for user decrypt, as MPC parties don't
                 // communicate between each other for user decrypt
+                if !acl_checked {
+                    self.decryption_processor
+                        .check_ciphertexts_allowed_for_decryption(&req.snsCtMaterials)
```
enitrat: [AUTOMATED]

**Issue: Wrong ACL function for User Decryption**

The same `check_ciphertexts_allowed_for_decryption()` function (which calls `isAllowedForDecryption(handle)`) is used for both PublicDecryption and UserDecryption.

According to the specification in issue #849, different ACL functions should be used:

| Decryption Type | Required ACL Check |
|-----------------|-------------------|
| Public | `isAllowedForDecryption(handle)` ✅ |
| User | `isAllowed(handle, userAddress)` ❌ Not implemented |
| Delegated User | `isHandleDelegatedForUserDecryption(...)` ❌ Not implemented |

`isAllowedForDecryption` checks if a handle is marked for **public** decryption, while `isAllowed(handle, user)` checks if a **specific user** has access to the handle.

**Suggested fix:**
Create a separate method for user decryption that calls `isAllowed(handle, req.userAddress)` instead.

Confidence: 95/100

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
```diff
@@ -119,13 +119,21 @@ impl<P: Provider> DbEventProcessor<P> {
...
     ) -> Result<KmsGrpcRequest, ProcessingError> {
+        let acl_checked = event.acl_checked;
         match &event.kind {
             GatewayEventKind::PublicDecryption(req) => {
                 self.decryption_processor
                     .check_decryption_not_already_done(req.decryptionId)
                     .await?;
+                if !acl_checked {
+                    self.decryption_processor
+                        .check_ciphertexts_allowed_for_decryption(&req.snsCtMaterials)
```
enitrat: [AUTOMATED]

**Consideration: TOCTOU (Time-of-Check to Time-of-Use) window**

Once `acl_checked = true` is set and persisted, retries will skip the ACL check. If ACL is revoked between the initial check and retry (due to network failure, etc.), the revocation may not be enforced.

**Questions:**
1. Is this an acceptable trade-off for performance optimization?
2. Should there be a TTL on `acl_checked` to force periodic re-verification?
3. What is the expected SLA for ACL revocation enforcement?

If this is an accepted limitation, please document it. Otherwise, consider adding a timestamp `acl_checked_at` and re-checking after a configurable TTL.

Confidence: 80/100

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
```diff
@@ -119,13 +119,21 @@ impl<P: Provider> DbEventProcessor<P> {
     #[tracing::instrument(skip_all)]
     async fn prepare_request(
         &self,
-        event: &GatewayEvent,
+        event: &mut GatewayEvent,
     ) -> Result<KmsGrpcRequest, ProcessingError> {
+        let acl_checked = event.acl_checked;
         match &event.kind {
             GatewayEventKind::PublicDecryption(req) => {
                 self.decryption_processor
```
enitrat, line 126: [AUTOMATED]

**Issue: Error counter not incremented for recoverable ACL errors**

When `check_ciphertexts_allowed_for_decryption()` fails with a `ProcessingError::Recoverable` error (e.g., host chain RPC timeout), the `event.error_counter` is not incremented.

This means:
- Events will retry indefinitely if the host chain is unavailable
- `max_decryption_attempts` limit will never trigger
- Database may fill with stuck events

**Note:** PR #1794 addresses a similar issue. Please ensure consistency.

Confidence: 90/100

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
```diff
@@ -71,6 +84,35 @@ where
         Ok(())
     }
 
+    pub async fn check_ciphertexts_allowed_for_decryption(
```
enitrat: [AUTOMATED]

**Question: Delegated User Decryption support**

The specification in issue https://github.com/zama-ai/fhevm-internal/issues/849 mentions three ACL check functions:
1. `isAllowedForDecryption(handle)` - for public decryption
2. `isAllowed(handle, userAddress)` - for user decryption
3. `isHandleDelegatedForUserDecryption(delegator, delegate, contract, handle)` - for delegated decryption

Is delegated user decryption currently supported by the Gateway? If so, which ACL check should be used?

If not yet supported, please document this as out-of-scope for this PR.

Confidence: 85/100

enitrat (changes requested):
[AUTOMATED]

## Code Review Summary

This PR adds ACL verification to the KMS Connector for defense-in-depth security. The architecture is sound, but there are implementation issues to address.

### Blocking Issues

1. **Wrong ACL function for User Decryption** - Uses `isAllowedForDecryption()` for both public and user decryption. User decryption should use `isAllowed(handle, userAddress)`.

2. **Error counter not incremented** - Recoverable ACL errors don't increment the error counter, potentially causing infinite retries.

### Questions for Author

1. Is delegated user decryption (`isHandleDelegatedForUserDecryption`) in scope?
2. Is the TOCTOU window on `acl_checked` persistence acceptable?

### Positive Observations

- ✅ Defense-in-depth architecture
- ✅ Chain ID extraction from handles correctly implemented
- ✅ Clean provider generics (`GP`, `HP`)
- ✅ Database persistence optimization for retries
- ✅ Unit tests for basic ACL scenarios

enitrat:
[AUTOMATED]

## 💡 Testing Tip: Using `rstest` for Parameterized ACL Tests

PR #1794 implements excellent test coverage for ACL checks using `rstest` parameterized tests. Adopting this pattern here would have caught the ACL function issue before review.

### Example from PR #1794:

```rust
#[derive(Debug, Clone, Copy)]
enum AclCheckType {
    Public,
    User,
}

#[derive(Debug, Clone, Copy)]
enum MockResponse {
    Allowed,
    NotAllowed,
    TransportError,
}

#[rstest]
#[case::public_allowed(AclCheckType::Public, MockResponse::Allowed, ExpectedResult::Success)]
#[case::public_not_allowed(AclCheckType::Public, MockResponse::NotAllowed, ExpectedResult::Irrecoverable)]
#[case::public_transport_error(AclCheckType::Public, MockResponse::TransportError, ExpectedResult::Recoverable)]
#[case::user_allowed(AclCheckType::User, MockResponse::Allowed, ExpectedResult::Success)]
#[case::user_not_allowed(AclCheckType::User, MockResponse::NotAllowed, ExpectedResult::Irrecoverable)]
#[case::user_transport_error(AclCheckType::User, MockResponse::TransportError, ExpectedResult::Recoverable)]
#[tokio::test]
async fn check_acl_authorization(
    #[case] check_type: AclCheckType,
    #[case] mock_response: MockResponse,
    #[case] expected: ExpectedResult,
) {
    // Test implementation...
    let result = match check_type {
        AclCheckType::Public => {
            processor.check_acl_for_public_decryption(&materials).await
        }
        AclCheckType::User => {
            processor.check_acl_for_user_decryption(&materials, user_address).await
            //                                                  ^^^^^^^^^^^^
            // This parameter forces you to implement a separate function!
        }
    };
}
```

### Why this matters:

Writing the test with `AclCheckType::User` requiring a `user_address` parameter **forces the implementation** to have a separate function that accepts and uses that address. The current single-function approach would fail to compile with this test structure.

### Suggested integration test (also from PR #1794):

```rust
#[rstest]
#[case::public_denied(EventType::PublicDecryptionRequest, AclBehavior::Denied)]
#[case::user_denied(EventType::UserDecryptionRequest, AclBehavior::Denied)]
#[case::public_transport_error(EventType::PublicDecryptionRequest, AclBehavior::TransportError)]
#[case::user_transport_error(EventType::UserDecryptionRequest, AclBehavior::TransportError)]
#[tokio::test]
async fn test_decryption_acl_behavior(...) {
    // Verifies that denied requests end up with status "failed"
    // Verifies that transport errors keep status "pending" (for retry)
}
```

---

## 📍 Follow-up: Error Counter Increment Location

For the error counter fix, PR #1794 implements it in `inner_process()` (processor.rs):

```rust
async fn inner_process(&mut self, event: &mut GatewayEvent) -> Result<...> {
    let request = match self.prepare_request(event).await {
        Ok(req) => req,
        Err(ProcessingError::Recoverable(e)) => {
            event.error_counter += 1;  // <-- Add this
            return Err(ProcessingError::Recoverable(e));
        }
        Err(e) => return Err(e),
    };
    // ...
}
```

This ensures **all** recoverable errors during `prepare_request()` (including ACL check failures) count toward `max_decryption_attempts`.

kms-connector/Cargo.toml
eudelins-zama: Agreed that we should move to using local paths at some point.

We used to do this when we were working without feature branch, thus we wanted to stick with a specific version of the bindings to avoid breaking kms-connector CI without knowing why.

But this would require a bit of work (modification of the integration test setup + probably the CI if we want to do this correctly) so we should do in a dedicated PR IMO :slightly_smiling_face:

enitrat: https://github.com/zama-ai/fhevm-internal/issues/891

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
eudelins-zama: Indeed! Fixed

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
eudelins-zama: Fixed. I'm not sure to fully know how the revocation mechanism is planned for the future, but anyway, the performance gain to persist the ACL in DB would not have been that significant...

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
eudelins-zama: The error counter is mainly to count error in the GRPC communication with the KMS Core, so this behavior is what we want

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
enitrat: I think this will need to coordinate with what @isaacdecoded is working on?

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
```diff
@@ -71,6 +84,65 @@ where
...
+                    "{} is not allowed for decrypt!",
+                    hex::encode(ct.ctHandle)
+                )));
+            }
+        }
+
+        Ok(())
+    }
+
+    pub async fn check_ciphertexts_allowed_for_user_decryption(
```
enitrat, line 116: needs a test now that it was added

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
```diff
@@ -213,4 +292,48 @@ mod tests {
             ProcessingError::Irrecoverable(_)
         ));
     }
+
+    #[tokio::test]
+    async fn check_ciphertexts_allowed_for_decryption() {
```
enitrat: prefer splitting into two tests: 

- `check_ciphertexts_allowed_for_decryption`
- `check_ciphertexts_not_allowed_for_decryption`

or, even better, with parametrization to avoid duplicating boilerplate

kms-connector/crates/utils/src/types/fhe.rs
```diff
@@ -2,20 +2,38 @@ use alloy::{hex, primitives::U256};
...
+            handle.len()
+        ))
+    }
+}
+
+/// Extracts the chain id from a ciphertext's handle.
+pub fn extract_chain_id_from_handle(handle: &[u8]) -> anyhow::Result<u64> {
+    let err_prefix = "Failed to extract chain_id from handle.";
+    if handle.len() >= 32 {
+        let chain_id_bytes = handle[22..30]
```
enitrat, line 28: So there's a funny thing here about encoding `chain_id`: technically, there's nothing in the ethereum protocol that prevents you from having a chain_id coded on 256 bits. If that was the case, then this logic of considering that the chain id fits in 8 bytes _could_ cause issues.

Actually, it did cause issues for us at Kakarot, because our own chainID was more than 8 bytes long. In theory, nothing disallows it, but in practice, wallets like metamask define a `MAX_SAFE_CHAIN_ID = 4,503,599,627,370,476` (~2^53) value and don't support chains above that. So for a while, we couldnt get our chain to work with wallets, until we changed our chainId.

No issue here just thought about it while reviewing this...

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
eudelins-zama: Yeah indeed, I reached out to him this morning, and for delegated decryption, we might need a dedicated event for the connector to check the ACL properly

kms-connector/Cargo.toml
eudelins-zama: Thx!

kms-connector/crates/utils/src/types/fhe.rs
eudelins-zama: Good to know indeed. I know we already discussed this topic, but I thought this was not supposed to be an issue in any case.
Seems it *could* be in niche case though...

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
eudelins-zama: Already the case unless I missed something so resolving

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
eudelins-zama: Fixed in 249e5efbde85b511756a1d7c6030508b15d28877.

I fully agree with the philosophy of single responsibility for unit tests.
But on the other hand there is now more boilerplate  with parametrized tests (enum, macros, match on the cases) than before, and tests are a bit harder to read IMO.
I'll let @dartdart26 tell us what he thinks of this to see if we keep the commit or not.

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
enitrat: I can see the `check_ciphertexts_allowed_for_public_decryption` test, but not  `check_ciphertexts_allowed_for_user_decryption`

eudelins-zama: Yep indeed, got confused sorry

eudelins-zama: Fixed

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
enitrat: > But on the other hand there is now more boilerplate with parametrized tests (enum, macros, match on the cases) than before, and tests are a bit harder to read IMO.


I agree it's not easy to read here, but I think that there's a good way to go around it:

1. If you have individual test cases testing a combination of (input, expected_result), then you're repeating 3 times the exact test setup. which makes un-necessary boilerplate
2. In the solution you propose you don't really leverage the parametrization. The thing we want to test is: _"Given an input I, an expected outcome O, is my function working properly?_"; but in your commit you structured the test to run a specific, hardcoded _case_, not a tuple `(I, O)`, which defeats the purpose of the parametrization a bit. The idea is that by checking a test + it's parametrization, you have the entire set of `(I, O)` that are tested by this test. This also prevents duplicating logic inside the test code itself (verbose match arms). The right way would be (I think) something like this:

```rust
  #[derive(Debug, Clone, Copy, PartialEq)]
  enum ExpectedOutcome {
      Ok,
      Recoverable,
      Irrecoverable,
  }

  enum MockResponse {
      Failure(&'static str),
      Success(bool),
  }

  #[rstest]
  #[case::transport_error(MockResponse::Failure("Transport Error"), ExpectedOutcome::Recoverable)]
  #[case::allowed(MockResponse::Success(true), ExpectedOutcome::Ok)]
  #[case::not_allowed(MockResponse::Success(false), ExpectedOutcome::Irrecoverable)]
  #[tokio::test]
  async fn check_ciphertexts_allowed_for_public_decryption(
      #[case] mock_response: MockResponse,
      #[case] expected: ExpectedOutcome,
  ) {
      let asserter = Asserter::new();
      let mock_provider = ProviderBuilder::new().disable_recommended_fillers().on_mocked_client(asserter.clone());
      let config = ConfigBuilder::default().build().unwrap();
      let acl_contracts_mock = AclContracts::new_mocked(asserter.clone());
      let s3_service = S3Service::new_mocked();
      let sns_ciphertexts = vec![rand_sns_ct()];
      let decryption_processor = DecryptionProcessor::new(&config, mock_provider, acl_contracts_mock, s3_service);

      match mock_response {
          MockResponse::Failure(msg) => asserter.push_failure_msg(msg),
          MockResponse::Success(val) => asserter.push_success(&val.abi_encode()),
      }

      let result = decryption_processor
          .check_ciphertexts_allowed_for_public_decryption(&sns_ciphertexts)
          .await;

      match expected {
          ExpectedOutcome::Ok => { result.unwrap(); }
          ExpectedOutcome::Recoverable => {
              assert!(matches!(result.unwrap_err(), ProcessingError::Recoverable(_)));
          }
          ExpectedOutcome::Irrecoverable => {
              assert!(matches!(result.unwrap_err(), ProcessingError::Irrecoverable(_)));
          }
      }
  }
  ```

Comment (eudelins-zama) [reply kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs]: Agreed your approach better leverages the parametrized test. But still compared to the initial multi-steps test, it's harder to grasp IMO

Comment (enitrat) [reply kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs]: I'm also fine having non-parametrized tests, with a fixture for `decryption_processor` and then just a list of tests like this if you prefer:

- decryption_check_succeeds_when_not_done
- decryption_check_irrecoverable_when_already_done
- decryption_check_recovers_transport_error

etc

this way you don't have the boilerplate and you have the "easy-to-grasp" property of just reading the test names

File: kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs (hunk 5 above)
Comment (enitrat) [reply kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs]: I am extracting this in another issue for tracking @eudelins-zama   https://github.com/zama-ai/fhevm-internal/issues/894

Comment (eudelins-zama) [reply kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs]: Good idea, thanks!

File: kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs [hunk 9]
```diff
@@ -71,6 +84,65 @@ where
...
+                return Err(ProcessingError::Irrecoverable(anyhow!(
+                    "No ACL contract config found for chain id {ct_chain_id}"
+                )));
+            };
+
+            if !acl_contract
+                .isAllowed(ct.ctHandle, user_address)
+                .call()
+                .await
+                .map_err(|e| ProcessingError::Recoverable(anyhow::Error::from(e)))?
```
enitrat, line 134: this is only checking (handle, user_address) but in `_extractCtHandlesCheckConformanceUser` (on GW) we used to do:

```
            // Check that the allowed and contract accounts have access to the handles.
            _checkIsAccountAllowed(ctHandle, allowedAddress);
            _checkIsAccountAllowed(ctHandle, contractAddress);
            
```

eudelins-zama: Indeed!
But we would need to pass the contractAddresses to the kms-connector via the UserDecryptionRequest event then

PR #1802, opened 2026-01-20
feat(protocol-contracts): pausers fetching script for both chains

jatZama:


protocol-contracts/config-checker/.env.example
```diff
@@ -0,0 +1,4 @@
+PAUSER_SET_ETHEREUM="0xbBfE1680b4a63ED05f7F80CE330BED7C992A586C"
+PAUSER_SET_GATEWAY="0x571ecb596fCc5c840DA35CbeCA175580db50ac1b"
+RPC_ETHEREUM="https://ethereum-rpc.publicnode.com"
+RPC_GATEWAY=
```
mortendahl: Missing value?

protocol-contracts/chains-config-checker/utils/get-deployment-block.js
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env node
+
+const { ethers } = require('ethers');
+
+const DEFAULT_RPC_URL = 'https://ethereum-rpc.publicnode.com';
+
+async function findDeploymentBlock(address, options = {}) {
```
melanciani, line 7: do we really need this func ? I mean, we only have 2 pauser contracts, we can hardcode their deployment block right now and simplify the script no ?

jatZama: Yes we absolutely need it, because it is common for future scripts, such as the one which will scan events for Aragon multisigs, etc.

protocol-contracts/config-checker/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "config-checker",
```
melanciani: I think the project (including the directory, packagejson and readme) should be more specific than just `config-checker` : 
- if it's only about pausers: rename it `pausers-checker` for ex
- if it's meant to be a bit more general in the future: `governance-checker` or `onchain-config-checker`

melanciani (dismissed):
2 minor comments, but otherwise looks good

protocol-contracts/config-checker/package.json
jatZama: This is not just governance, this is also to check owner of host and gateway contracts, etc.

jatZama: I can rename to `onchain-config-checker` if you really want.

jatZama: and for staking, etc

jatZama: Done renaming to `chains-config-checker`

melanciani (approved):
thanks !

PR #1815, opened 2026-01-21
chore(protocol-contracts): remove non-payment protocol-contracts from repo

melanciani:


melanciani:
doing this again on another PR

PR #1840, opened 2026-01-26
fix(test-suite): add resilient smoke runner for inputFlow

Eikix:
## Summary
- add a dedicated smoke runner (Hardhat runtime) for inputFlow with minimal tx resilience (nonce preflight, timeout + fee bump, optional auto-cancel)
- support deploy or attach to TestInput; chain-aware coprocessor config for sepolia/mainnet deploys
- add per-network RPC vars + Hardhat vars support, devnet env template, README updates

## Why
- smoke tests can stall on pending nonce/underpriced txs
- deploys on sepolia/mainnet need correct coprocessor config and relayer auth

## Testing
- `SMOKE_FORCE_DEPLOY=1 npx hardhat run --network sepolia scripts/smoke-inputflow.ts`
- `TEST_INPUT_CONTRACT_ADDRESS=<deployed> npx hardhat run --network sepolia scripts/smoke-inputflow.ts`
- `SMOKE_FORCE_DEPLOY=1 npx hardhat run --network mainnet scripts/smoke-inputflow.ts`

## Handoff / Infra
- Cron: `npx hardhat run --network mainnet test-suite/e2e/scripts/smoke-inputflow.ts`
- Env: `MAINNET_ETH_RPC_URL`, `MNEMONIC`, `RELAYER_URL`, `ZAMA_FHEVM_API_KEY`, `CHAIN_ID_GATEWAY`, `CHAIN_ID_HOST`, `DECRYPTION_ADDRESS`, `INPUT_VERIFICATION_ADDRESS`, `ACL_CONTRACT_ADDRESS`, `KMS_VERIFIER_CONTRACT_ADDRESS`, `INPUT_VERIFIER_CONTRACT_ADDRESS`, `TEST_INPUT_CONTRACT_ADDRESS`
- Refs: zama-ai/fhevm-internal#928, zama-ai/infra-internal#32

closes https://github.com/zama-ai/infra-internal/issues/32
closes https://github.com/zama-ai/fhevm-internal/issues/928


Eikix:
Optional follow-up: add `SMOKE_OUTPUT_FILE` (or similar) so a successful deploy can persist `TEST_INPUT_CONTRACT_ADDRESS` for the next run. Not required for this PR, but would make ops smoother.

Eikix:
@cla-bot check

Eikix:
## Changes from review

### Simplifications
- Simplified RPC URL warning conditions (removed redundant checks)
- Refactored `cancelBacklog` to reuse `sendWithRetries` for retry/fee-bump resilience

### Post-success cleanup
- Moved backlog clearing off the critical path - test runs first, cleanup happens after success
- Added skip warnings with reasons (`low_balance`, `backlog_too_large`) for observability

### BetterStack heartbeat integration
- Pings `BETTERSTACK_HEARTBEAT_URL` on success (if set)
- Reports failure status with error message via `?status=1&msg=...`
- Documented in README

### Note for follow-up
Consider renaming `zwsDev` → `devnet` and removing `staging` (they're currently identical) in a separate PR.

---

## Test Results

### Sepolia
| Mode | Contract | Result |
|------|----------|--------|
| `SMOKE_FORCE_DEPLOY=1` | `0x18b27e96e73b8c62aCb3046eE091015f0Ac730ad` | ✅ |
| `TEST_INPUT_CONTRACT_ADDRESS=...` | attach | ✅ |

### Mainnet
| Mode | Contract | Tx Hash | Result |
|------|----------|---------|--------|
| `SMOKE_FORCE_DEPLOY=1` | `0x12982F50CA9596E354370F2B272D542285223c7B` | [0x5667ef42...](https://etherscan.io/tx/0x5667ef42171a6fbfebfb768a6d12973628d15c89f48aedca0ce844286581551c) | ✅ |
| `TEST_INPUT_CONTRACT_ADDRESS=...` | attach | [0xd21444eb...](https://etherscan.io/tx/0xd21444ebe2c583a18f12468a3af0377425d05401e83da02dbf5adc01b922d0e4) | ✅ |

test-suite/e2e/contracts/E2ECoprocessorConfigLocal.sol
```diff
@@ -1,10 +1,15 @@
...
 pragma solidity ^0.8.24;
 
+import {ZamaConfig} from "@fhevm/solidity/config/ZamaConfig.sol";
 import {CoprocessorConfig, FHE} from "@fhevm/solidity/lib/FHE.sol";
 
 library DefaultCoprocessorConfig {
-    function getConfig() internal pure returns (CoprocessorConfig memory) {
+    function getConfig() internal view returns (CoprocessorConfig memory) {
+        if (block.chainid == 1 || block.chainid == 11155111) {
+            return ZamaConfig.getEthereumCoprocessorConfig();
```
dartdart26: Does that work for both chain IDs?

test-suite/e2e/README.md
```diff
@@ -11,3 +11,45 @@ REPORT_GAS=true npx hardhat test
...
+
+The smoke runner uses the same environment variables as the e2e tests:
+
+- `RPC_URL` (or `SEPOLIA_ETH_RPC_URL` / `MAINNET_ETH_RPC_URL`), `MNEMONIC`
+- `RELAYER_URL`
+- `KMS_VERIFIER_CONTRACT_ADDRESS`, `ACL_CONTRACT_ADDRESS`, `INPUT_VERIFIER_CONTRACT_ADDRESS`
+- `CHAIN_ID_GATEWAY`, `CHAIN_ID_HOST`
+- `DECRYPTION_ADDRESS`, `INPUT_VERIFICATION_ADDRESS`
+
+Mainnet additionally requires `ZAMA_FHEVM_API_KEY` and `TEST_INPUT_CONTRACT_ADDRESS` unless `SMOKE_FORCE_DEPLOY=1`.
```
dartdart26: In general, if we are to only run smoke in a separate runner, maybe we can remove `TEST_INPUT_CONTRACT_ADDRESS` as it was added mainly to avoid it on mainnet. I mean, AFAIU, if `SMOKE_FORCE_DEPLOY` is not 1, then we don't need `TEST_INPUT_CONTRACT_ADDESS`?

test-suite/e2e/hardhat.config.ts
```diff
@@ -90,37 +89,51 @@ const chainIds = {
...
       break;
     case 'sepolia':
-      jsonRpcUrl = process.env.RPC_URL || defaultRpcUrl;
-      if (jsonRpcUrl === defaultRpcUrl && !process.env.RPC_URL) {
+      jsonRpcUrl =
+        process.env.SEPOLIA_ETH_RPC_URL ??
+        (vars.has('SEPOLIA_ETH_RPC_URL') ? vars.get('SEPOLIA_ETH_RPC_URL') : undefined) ??
+        process.env.RPC_URL ??
+        (vars.has('RPC_URL') ? vars.get('RPC_URL') : undefined) ??
+        defaultRpcUrl;
```
dartdart26: Are these changes relevant for this PR?

test-suite/e2e/test/instance.ts
```diff
@@ -1,20 +1,69 @@
...
+
+const inputAdd =
+  process.env.INPUT_VERIFIER_CONTRACT_ADDRESS ??
+  vars.get('INPUT_VERIFIER_CONTRACT_ADDRESS', defaults?.inputVerifierContractAddress ?? '');
+if (!inputAdd) throw new Error('Missing required env var INPUT_VERIFIER_CONTRACT_ADDRESS');
+
+const gatewayChainRaw =
+  process.env.CHAIN_ID_GATEWAY ?? vars.get('CHAIN_ID_GATEWAY', defaults?.gatewayChainId?.toString() ?? '');
+const gatewayChainID = gatewayChainRaw ? Number(gatewayChainRaw) : undefined;
+if (!Number.isFinite(gatewayChainID)) throw new Error('Missing required env var CHAIN_ID_GATEWAY');
```
dartdart26: That's a bit hard to parse, IMO. Do we need this type of handling?

test-suite/e2e/scripts/smoke-inputflow.ts
```diff
@@ -0,0 +1,420 @@
...
+  }
+}
+
+runSmoke().catch(async (error) => {
+  console.error(`SMOKE_FAILED ${String(error)}`);
+  process.exitCode = 1;
+
+  const heartbeatUrl = process.env.BETTERSTACK_HEARTBEAT_URL;
+  if (heartbeatUrl) {
+    const msg = encodeURIComponent(String(error).slice(0, 100));
```
dartdart26: Do we need such msg handling?

test-suite/e2e/scripts/smoke-inputflow.ts
```diff
@@ -0,0 +1,420 @@
...
+  const provider = ethers.provider;
+  const signerIndices = parseIndices(process.env.SMOKE_SIGNER_INDICES ?? DEFAULT_SIGNER_INDICES);
+  const timeoutMs = Math.trunc(Number(process.env.SMOKE_TX_TIMEOUT_SECS) || DEFAULT_TX_TIMEOUT_SECS) * 1000;
+  const maxRetries = Math.trunc(Number(process.env.SMOKE_TX_MAX_RETRIES) || DEFAULT_TX_MAX_RETRIES);
+  const feeBump = Number(process.env.SMOKE_FEE_BUMP) || DEFAULT_FEE_BUMP;
+  const maxBacklog = Math.trunc(Number(process.env.SMOKE_MAX_BACKLOG) || DEFAULT_MAX_BACKLOG);
+  const decryptTimeoutMs =
+    Math.trunc(Number(process.env.SMOKE_DECRYPT_TIMEOUT_SECS) || DEFAULT_DECRYPT_TIMEOUT_SECS) * 1000;
+  const cancelRaw = process.env.SMOKE_CANCEL_BACKLOG;
+  const allowCancel = cancelRaw === undefined || cancelRaw === '' ? true : cancelRaw.trim().toLowerCase() !== 'false';
```
dartdart26: Looks a bit hard to parse, at least to me.

dartdart26 (dismissed):
Looks good overall, just some minor comments.

It is a bit more verbose in some places and interaction with existing env vars make it a bit complex, IMO. I can't verify if it is correct in all cases just by looking at it, but should be good for the smoke test.

test-suite/e2e/contracts/E2ECoprocessorConfigLocal.sol
Eikix: Yes, `ZamaConfig.getEthereumCoprocessorConfig()` handles both chain IDs internally - it checks `block.chainid` and returns the appropriate config:
- chainId 1 → mainnet addresses
- chainId 11155111 → Sepolia addresses
- chainId 31337 → local addresses

The check here was redundant. Simplified in 7de7569d to just delegate to ZamaConfig entirely.

test-suite/e2e/README.md
Eikix: I think there might be a misunderstanding - the logic is actually:

- `SMOKE_FORCE_DEPLOY=1` → deploys a new contract (no need for `TEST_INPUT_CONTRACT_ADDRESS`)
- `SMOKE_FORCE_DEPLOY` not set → **requires** `TEST_INPUT_CONTRACT_ADDRESS` to attach to an existing contract

The goal is to **not** redeploy on every mainnet smoke run (to save gas). So on mainnet we:
1. Deploy once with `SMOKE_FORCE_DEPLOY=1`
2. Save the contract address
3. Run subsequent smoke tests with `TEST_INPUT_CONTRACT_ADDRESS=<addr>` (no force deploy)

So `TEST_INPUT_CONTRACT_ADDRESS` is needed for the normal case (reusing existing contract). Were you suggesting a different approach?

test-suite/e2e/hardhat.config.ts
Eikix: You're right, the verbose chaining was hard to parse. Simplified in 518cc3f9:

- staging/zwsDev: `process.env.RPC_URL ?? vars.get('RPC_URL', defaultRpcUrl)`
- sepolia: requires `SEPOLIA_ETH_RPC_URL` explicitly (throws if not set)
- mainnet: requires `MAINNET_ETH_RPC_URL` explicitly (throws if not set)

Removed the `RPC_URL` fallback for sepolia/mainnet - if you're targeting a specific network, you should set its specific env var. Clearer and less error-prone.

test-suite/e2e/README.md
dartdart26: Isn't it the case that if `SMOKE_FORCE_DEPLOY` is not set and `TEST_INPUT_CONTRACT_ADDRESS` is not set too, we will deploy? If that's the case, not sure we need `SMOKE_FORCE_DEPLOY`?

test-suite/e2e/test/instance.ts
Eikix: Agreed - simplified significantly in 1379516c.

Chain IDs, contract addresses, and URLs don't need hardhat vars (not secrets). Now just:
```typescript
const gatewayChainID = Number(process.env.CHAIN_ID_GATEWAY) || defaults?.gatewayChainId;
if (!gatewayChainID) throw new Error('CHAIN_ID_GATEWAY is required');
```

Only `ZAMA_FHEVM_API_KEY` keeps vars support since it's actually a secret.

Went from 50 lines to 33 lines in the config section.

test-suite/e2e/scripts/smoke-inputflow.ts
Eikix: You're right - simplified in 6ccc4ab7.

BetterStack has a dedicated `/fail` endpoint, so now it's just:
```typescript
await fetch(`${heartbeatUrl}/fail`).catch(() => {});
```

No need for URL encoding or message truncation. Error details are already in the console logs.

test-suite/e2e/scripts/smoke-inputflow.ts
Eikix: Simplified in 53961702:

```typescript
const allowCancel = process.env.SMOKE_CANCEL_BACKLOG !== 'false';
const forceDeploy = Boolean(process.env.SMOKE_FORCE_DEPLOY);
```

Also removed unnecessary `Math.trunc()` calls throughout.

test-suite/e2e/README.md
Eikix: The current logic actually throws an error if neither is set:

```typescript
if (forceDeploy) {
  // deploy
} else if (existingContractAddress) {
  // attach  
} else {
  throw new Error('TEST_INPUT_CONTRACT_ADDRESS is required...');
}
```

We just tested this - running without `TEST_INPUT_CONTRACT_ADDRESS` gives:
```
SMOKE_FAILED Error: TEST_INPUT_CONTRACT_ADDRESS is required for smoke runs. Set SMOKE_FORCE_DEPLOY=1 to deploy explicitly.
```

The explicit `SMOKE_FORCE_DEPLOY` was intentional to prevent accidental deploys on mainnet.

That said, we're open to your suggestion - using `TEST_INPUT_CONTRACT_ADDRESS` as the sole determinant:
- Set → attach
- Not set → deploy

This would simplify to one env var instead of two. Let us know if you'd prefer that approach.

piizama: I also agree with @dartdart26, for the sake of simplicity we can keep the old behavior.

@Eikix, In this case we can set `SMOKE_FORCE_DEPLOY=1` by default (maybe rename `SMOKE_FORCE_DEPLOY` -> `SMOKE_DEPLOY_CONTRACT`). The operator can then set it to 0 to prevent any contract deployment which is a good feature to have.

Additionally, it would be nice to be able to deploy the contract only without running the smoke tests, maybe by introducing a: `SMOKE_RUN_TESTS` variable that default to `true` but that can be set to false to not run the tests.

test-suite/e2e/scripts/smoke-inputflow.ts
```diff
@@ -0,0 +1,416 @@
...
+  }
+}
+
+runSmoke().catch(async (error) => {
+  console.error(`SMOKE_FAILED ${String(error)}`);
+  process.exitCode = 1;
+
+  const heartbeatUrl = process.env.BETTERSTACK_HEARTBEAT_URL;
+  if (heartbeatUrl) {
+    await fetch(`${heartbeatUrl}/fail`).catch(() => {});
```
piizama: Can you also pass any error logs and information about failure to the body of the `/fail` endpoint, see: https://betterstack.com/docs/uptime/cron-and-heartbeat-monitor/#reporting-failures

piizama (changes requested):

Eikix:
## Addressed in 06e0498f

1. SMOKE_FORCE_DEPLOY → SMOKE_DEPLOY_CONTRACT (defaults to 1)
2. Added SMOKE_RUN_TESTS (set to 0 for deploy-only)
3. BetterStack /1 endpoint with error body
4. Strict 0/1 boolean parsing

Tested on Sepolia ✅

test-suite/e2e/README.md
Eikix: Done in 06e0498f:
- Renamed to `SMOKE_DEPLOY_CONTRACT` (defaults to `1`)
- Added `SMOKE_RUN_TESTS` (defaults to `1`, set to `0` for deploy-only)
- Both use strict `0`/`1` parsing with fail-fast on invalid values

test-suite/e2e/scripts/smoke-inputflow.ts
Eikix: Done in 06e0498f - now using `/1` endpoint with error message in POST body (truncated to 10KB)

test-suite/e2e/scripts/smoke-inputflow.ts
```diff
@@ -0,0 +1,444 @@
...
+  const errorMessage = String(error);
+  console.error(`SMOKE_FAILED ${errorMessage}`);
+  process.exitCode = 1;
+
+  const heartbeatUrl = process.env.BETTERSTACK_HEARTBEAT_URL;
+  if (heartbeatUrl) {
+    await fetch(`${heartbeatUrl}/1`, {
+      method: 'POST',
+      body: errorMessage.slice(0, 10000),
+    }).catch(() => {});
```
obatirou: Should a warn be done if the heartbeat fails ?

dartdart26 (dismissed):

piizama (dismissed):

test-suite/e2e/scripts/smoke-inputflow.ts
Eikix: Done!

piizama (dismissed):

Eikix:
## Summary of fixes in this PR

### 1. Heartbeat failure logging (`40be8fc7`)
Log warning when BetterStack heartbeat fails on smoke test failure, matching success behavior.

### 2. Constructor-based config injection (`76208a8e`, `649e47fe`)
**Problem**: Smoke tests used `E2ECoprocessorConfigLocal.sol` which gets sed-patched by Pierre's E2E runner. This doesn't work for standalone smoke tests on devnet/sepolia/mainnet.

**Solution**: Created `SmokeTestInput.sol` with constructor injection:
```solidity
constructor(address aclAddress, address coprocessorAddress, address kmsVerifierAddress)
```
Addresses come from `test/instance.ts` which:
- Uses SDK defaults for sepolia/mainnet
- Falls back to env vars for devnet

### 3. Preserved E2E compatibility
`TestInput.sol` kept unchanged (inherits `E2ECoprocessorConfig`) so Pierre's sed-based E2E workflow continues to work.

### 4. CI fix (cherry-picked from #1861)
Bumped actionlint version to fix `@actions/tool-cache` conflict.

---

**Tested**: Sepolia deploy + attach ✅ | Mainnet deploy + attach ✅

test-suite/e2e/scripts/smoke-inputflow.ts
```diff
@@ -0,0 +1,463 @@
...
+    try {
+      tx = await send({ nonce, ...fees });
+    } catch (error) {
+      lastError = error instanceof Error ? error : new Error(String(error));
+      console.warn(`SMOKE_TX_SEND_FAILED label=${label} nonce=${nonce} attempt=${attempt}`);
+      continue;
+    }
+
+    console.log(`SMOKE_TX_SENT label=${label} nonce=${nonce} hash=${tx.hash} attempt=${attempt}`);
+    const receipt = await waitForReceipt(provider, tx.hash, timeoutMs);
```
maxnovawind: Why not using the function waitForTransaction with ethers pkg,
https://docs.ethers.org/v6/api/providers/#Provider-waitForTransaction ?

Eikix: Update: Just tested — `provider.waitForTransaction` throws:

```
NotImplementedError: Method 'HardhatEthersProvider.waitForTransaction' is not implemented
```

So the custom polling is actually required when running through Hardhat.

test-suite/e2e/scripts/smoke-inputflow.ts
```diff
@@ -0,0 +1,463 @@
...
+};
+
+const waitForReceipt = async (
+  provider: Provider,
+  txHash: string,
+  timeoutMs: number,
+): Promise<TransactionReceipt | null> => {
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
+    const receipt = await provider.getTransactionReceipt(txHash);
```
maxnovawind: I think using ethers.js v6 might not be possible due to the dependency on Hardhat. But with v6 we can rely on the library itself and use tx.wait() with TRANSACTION_REPLACED, as described here: https://docs.ethers.org/v6/api/providers/#TransactionResponse-wait. This is the native ethers way to handle same-nonce, new-hash replacements.

Eikix: tx.wait() is available, let me see if i can refactor

Eikix: [refactored](https://github.com/zama-ai/fhevm/pull/1840/commits/1cd5ea1979b58454122203d062c427fdaf13553c), can you TAL?

maxnovawind: Seem good for me. Minor change: more make sense to check if the receipt is null first then check if receipt.status === 1 (https://github.com/zama-ai/fhevm/blob/1cd5ea1979b58454122203d062c427fdaf13553c/test-suite/e2e/scripts/smoke-inputflow.ts#L169).

maxnovawind: I hope you validate it by testing locally. The provider Web3 in every language is a mess, especially when it comes to interacting with the blockchain.

Eikix: > Seem good for me. Minor change: more make sense to check if the receipt is null first then check if receipt.status === 1 (
> 
> https://github.com/zama-ai/fhevm/blob/1cd5ea1979b58454122203d062c427fdaf13553c/test-suite/e2e/scripts/smoke-inputflow.ts#L169
> 
> ).

If that's okay with you we can skip this minor change as it's not crucial to including in v0.11 and we're short on time (wrt docker time for re-build etc.). Right now the head of the PR is tested on testnet by Pierre

piizama (dismissed):

Eikix:
@Mergifyio queue

Eikix:
@Mergifyio unqueue

piizama (dismissed):

dartdart26 (dismissed):

Eikix:
@Mergifyio queue

Eikix:
https://github.com/Mergifyio unqueue

Eikix:
@Mergifyio queue

test-suite/e2e/test/instance.ts
```diff
@@ -1,20 +1,68 @@
...
+if (!verifyingContractAddressInputVerification) throw new Error('INPUT_VERIFICATION_ADDRESS is required');
+
+const relayerUrl = process.env.RELAYER_URL || defaults?.relayerUrl;
+if (!relayerUrl) throw new Error('RELAYER_URL is required');
+
+// API key is a secret - support hardhat vars for secure storage
+const apiKey = process.env.ZAMA_FHEVM_API_KEY ?? vars.get('ZAMA_FHEVM_API_KEY', '');
+const isMainnet = network.name === 'mainnet' || network.config.chainId === 1;
+if (isMainnet && !apiKey) {
+  throw new Error('ZAMA_FHEVM_API_KEY is required for mainnet');
```
maxnovawind: Can we remove this condition ? Since the smoke is done internally there is no need to have this condition.

Eikix: what do you mean? it's good to fail fast no?

test-suite/e2e/scripts/smoke-inputflow.ts
```diff
@@ -0,0 +1,524 @@
...
+};
+
+const formatSignerState = (state: SignerState): string => {
+  const balanceStr = ethers.formatEther(state.balance);
+  const lowBalanceWarning = state.balance < LOW_BALANCE_THRESHOLD ? ' ⚠️ LOW_BALANCE' : '';
+  return `index=${state.index} address=${state.address} latest=${state.latest} pending=${state.pending} balance=${balanceStr} ETH${lowBalanceWarning}`;
+};
+
+const bumpFees = (base: FeeData, multiplier: number): FeeData => {
+  const scaled = BigInt(Math.round(multiplier * 100));
```
nboisde, line 146: Just taking a glance here, you should not use BigInt in js/ts for the gas prices or transaction values, you could take a glance at ethers bigNumbers library.
Should be ok here, but could be harmful on some other edge cases.

Eikix: i thought ethers v6 uses native bigint and everyone is starting to use it again?

or is it for loss of precision?

Eikix:
@claude /pr-review

piizama (approved):

tawadaa (approved):

Eikix:
@mergify queue

PR #1853, opened 2026-01-28
feat(coprocessor): sns-worker, push metadata on S3

rudy-6-4:


rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

coprocessor/fhevm-engine/sns-worker/src/json_sidecar.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub key_id: String,
+    pub tx_hash: String,
+    pub handle: String,
+    pub digest: String,
+    pub signed_tuple: String,
+    pub signed_digest: String,
+    pub signer: String,
+    pub created_at: String,
+    pub format: String,
+}
```
eudelins-zama, line 33: Just an overall comment here.

I think this kind of stuff should live in a shared crate (like `fhevm_ct_commits` or whatever) to be used by both copro, kms-connector and maybe even Relayer.
And using proper types instead of `String` everywhere if possible (don't know the constraint of S3 metadata for now).

I don't fully get why we have the `tx_hash` but let's discuss it in our sync.

rudy-6-4 (commented):
Adapt sidecar to follow https://github.com/zama-ai/tech-spec/blob/1bd8135fece711ee5432978f33871f945ed17757/rfcs/023-ciphertext-commits-removal.md#canonical-signed-payload

coprocessor/fhevm-engine/sns-worker/src/json_sidecar.rs
rudy-6-4: All fields are compatible with text content (ie json compatible), eg bytes content are hex formatted to be properly included in the json file. Do you want another formatting compatible with text content ? It could be arrays of integers  or base64 ?

eudelins-zama: I was thinking we could just do something like `.metadata("ct-attestation", serde_json::serialize(sidecar))`.

Then, the sidecar can have proper type in its fields, we just need to implement the `Serlialize/Deserialize` trait.
Best case is just a `derive`.
Worst case would be that we need to manually implement it if some of our types do not implement them.

rudy-6-4:
@claude /pr-review

coprocessor/fhevm-engine/sns-worker/src/lib.rs
```diff
@@ -250,13 +261,17 @@ impl HandleItem {
         &self,
         db_txn: &mut Transaction<'_, Postgres>,
     ) -> Result<(), ExecutionError> {
+        let ct128_format: i16 = self.ct128.format().into();
+
         sqlx::query!(
-            "INSERT INTO ciphertext_digest (host_chain_id, key_id_gw, handle, transaction_id)
-            VALUES ($1, $2, $3, $4) ON CONFLICT DO NOTHING",
+            "INSERT INTO ciphertext_digest
```
rudy-6-4: add a field to track old vs new format, will help during this migration and futur ones

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -310,41 +494,52 @@ async fn upload_ciphertexts(
...
-                    .send()
-                    .instrument(ct128_upload_span.clone()),
-                UploadResult::CtType128((ct128_digest.clone(), ct128_upload_span)),
-            ));
+            let result = UploadResult::CtType128(ct128_digest.clone());
+            let upload_ct = upload_ct(
+                ct128_upload_span.clone(),
+                client.clone(),
+                bucket.clone(),
+                key,
```
rudy-6-4: add context id to the key

rudy-6-4:
@claude /pr-review

antoniupop:
Agent review points (the ones that look relevant):

  - Critical: ct64 and ct128 can overwrite each other in the Helm deployment. The chart passes the same S3_BUCKET_NAME to both flags in charts/coprocessor/templates/coprocessor-sns-
    worker-deployment.yaml:66, while the new uploader writes ct128 and ct64 to the same primary key in coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs:537 and coprocessor/fhevm-
    engine/sns-worker/src/aws_upload.rs:630. With one bucket, both objects race on <handle>/<context_id>. Tests use separate buckets in coprocessor/fhevm-engine/sns-worker/src/tests/
    mod.rs:986, so they miss this. Add a type prefix or enforce separate buckets.

  - High: the uploaded path/signature uses context ID 0, but CiphertextCommits stores context ID 1 for added ciphertexts. See coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs:43
    and gateway-contracts/contracts/CiphertextCommits.sol:137. If verifiers reconstruct the attestation with the on-chain context, every signature will fail, and /1 lookups will miss
    objects.

  - Medium: transient head_object failures are converted to “missing object” in coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs:545 and coprocessor/fhevm-engine/sns-worker/src/
    aws_upload.rs:637. For retry rows where bytes are not loaded, this becomes a non-transient “bytes unavailable” failure instead of marking S3 unhealthy.

  - Medium: s3_format_version is row-level and set by each variant update independently in coprocessor/fhevm-engine/sns-worker/src/lib.rs:313 and coprocessor/fhevm-engine/sns-worker/
    src/lib.rs:342. Fully uploaded old rows are marked 0 by the migration but never backfilled, because retry only selects rows with a missing digest in coprocessor/fhevm-engine/sns-
    worker/src/aws_upload.rs:765. Consumers need explicit v0 fallback, or this needs a real migration/backfill plan.

eudelins-zama:
> High: the uploaded path/signature uses context ID 0, but CiphertextCommits stores context ID 1 for added ciphertexts. See coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs:43
> and gateway-contracts/contracts/CiphertextCommits.sol:137

Good catch! Maybe we should use `1` indeed, and update the RFC as well

rudy-6-4:
> Agent review points (the ones that look relevant):
> 
>     * Critical: ct64 and ct128 can overwrite each other in the Helm deployment. The chart passes the same S3_BUCKET_NAME to both flags in charts/coprocessor/templates/coprocessor-sns-
>       worker-deployment.yaml:66, while the new uploader writes ct128 and ct64 to the same primary key in coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs:537 and coprocessor/fhevm-
>       engine/sns-worker/src/aws_upload.rs:630. With one bucket, both objects race on /<context_id>. Tests use separate buckets in coprocessor/fhevm-engine/sns-worker/src/tests/
>       mod.rs:986, so they miss this. Add a type prefix or enforce separate buckets.
> 
It's not in part of the PR. But I can fix it in the PR.

rudy-6-4:
> Agent review points (the ones that look relevant):
> 
>     * Medium: s3_format_version is row-level and set by each variant update independently in coprocessor/fhevm-engine/sns-worker/src/lib.rs:313 and coprocessor/fhevm-engine/sns-worker/
>       src/lib.rs:342. Fully uploaded old rows are marked 0 by the migration but never backfilled, because retry only selects rows with a missing digest in coprocessor/fhevm-engine/sns-
>       worker/src/aws_upload.rs:765. Consumers need explicit v0 fallback, or this needs a real migration/backfill plan.

Migration is done in a next PR. This PR only do the new behavior while still posting on old key.

rudy-6-4:
> > High: the uploaded path/signature uses context ID 0, but CiphertextCommits stores context ID 1 for added ciphertexts. See coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs:43
> > and gateway-contracts/contracts/CiphertextCommits.sol:137
> 
> Good catch! Maybe we should use `1` indeed, and update the RFC as well

rfc is fixed and code aligned.

coprocessor/fhevm-engine/sns-worker/src/tests/mod.rs
```diff
@@ -26,10 +33,14 @@ use test_harness::{
...
 };
-use tokio::{sync::mpsc, time::timeout};
+use tokio::{
+    sync::mpsc,
+    time::{sleep, timeout},
+};
 use tracing::{info, Level};
 
 const LISTEN_CHANNEL: &str = "sns_worker_chan";
+const COPROCESSOR_CONTEXT_ID: U256 = U256::ZERO;
```
eudelins-zama: Just don't forget to use `U256::ONE` here 🙂

rudy-6-4:
> * Medium: transient head_object failures are converted to “missing object” in coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs:545 and coprocessor/fhevm-engine/sns-worker/src/
>       aws_upload.rs:637. For retry rows where bytes are not loaded, this becomes a non-transient “bytes unavailable” failure instead of marking S3 unhealthy.

Fixed

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -242,8 +268,219 @@ async fn run_uploader_loop(
...
+        Ciphertext128Format::Unknown => {
+            anyhow::bail!("Cannot build ciphertext attestation with unknown ct128 format")
+        }
+    }
+}
+
+fn upload_material(task: &HandleItem) -> anyhow::Result<UploadMaterial> {
+    if task.ct64_compressed.is_empty() && task.ct128.is_empty() {
+        let err = anyhow::anyhow!("Invalid upload task without ciphertext bytes: {:?}", task);
+        error!(?task, error = %err, "Upload task has no ciphertext bytes");
```
antoniupop, line 563: would this log the entire CT if it fires? maybe handle is sufficient? chain id as well

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -242,8 +268,219 @@ async fn run_uploader_loop(
...
+        return Err(err);
+    }
+
+    if task.ct64_compressed.is_empty() && task.ct64_digest.is_none() {
+        let err = anyhow::anyhow!(
+            "Invalid upload task for handle {}: missing ct64 digest and ciphertext bytes",
+            to_hex(&task.handle),
+        );
+        error!(?task, error = %err, "Upload task has no ct64 material");
```
antoniupop, line 572: same as above

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -242,8 +268,219 @@ async fn run_uploader_loop(
...
+    } else {
+        compute_digest(task.ct64_compressed.as_ref())
+    };
+
+    let ct128_digest = if let Some(digest) = &task.ct128_digest {
+        digest.clone()
+    } else if !task.ct128.is_empty() {
+        compute_digest(task.ct128.bytes())
+    } else {
+        NO_SNS_CIPHERTEXT_DIGEST.to_vec()
```
antoniupop, line 587: This makes snsCiphertextDigest = 0x0000... and may need to be clarified with KMS connector and RFC 23, I don't think it's specified so could have strange behaviour cc @eudelins-zama 
Corner case, but could cause problems

antoniupop (commented):
Overall looks good, I think it's nearly ready. I spotted a couple issues (inline comments) and an agent found some more, the first one looks serious, second looks like it's imposible in practice, but could happen if we change behaviour later:

- s3_format_version can be set to 1 while the ct64 object was never migrated. The flag is row-level but updated per-variant. Concrete case: a pre-migration row with ct64 already uploaded (old digest key) and ct128 pending. Recovery re-uploads both at the new path; if ct128 succeeds but ct64's PUT fails transiently, update_ct128_uploaded (lib.rs:322) commits s3_format_version = 1 and the row now has both digests non-NULL — so fetch_pending_uploads (which only selects rows with a NULL digest) never picks it up again. Result: a row claiming v1 whose ct64 exists only at the legacy digest key with no attestation. Harmless in Step 1, but the RFC's pre-Step-2 migration ("all previously committed handles remain resolvable") would silently skip it if the backfill keys on s3_format_version = 1. Setting the version only after both variants are confirmed at the new path (or tracking it per-variant) would close this.

1. s3_format_version is a row-level claim. The migration comment defines it as a property of the row's S3 state: 0 = initial format, 1 = handle and signed digest metadata. One value
  speaks for two objects (the ct64 object in bucket_ct64 and the ct128 object in bucket_ct128).
  2. But it's stamped per-variant. Both update_ct64_uploaded (lib.rs:345) and update_ct128_uploaded (lib.rs:317) set s3_format_version = 1 as a side effect of marking their own variant
  uploaded. Whichever variant completes first stamps the whole row as "migrated."
  3. Retry is keyed on digest NULLness, not on the version. fetch_pending_uploads selects WHERE ciphertext IS NULL OR ciphertext128 IS NULL (aws_upload.rs:750). For legacy rows
  (version 0), the rewrite of an already-uploaded variant is opportunistic: it happens only because the row got selected for the other variant's sake. Nothing in the selection
  predicate tracks "this v0 variant still needs its rewrite."

  The failure walkthrough. Take the population this recovery logic was built for — a partially-uploaded legacy row. Say ct64 was uploaded pre-PR (object at digest key hex(d64), no
  attestation), ct128 never made it: row has ciphertext = d64, ciphertext128 = NULL, backfilled s3_format_version = 0.

  1. The row matches the pending query (ct128 digest is NULL). Because version ≠ 1, recovery fetches bytes for both variants so it can rewrite the old ct64 object too.
  2. upload_ciphertexts finds neither object at the new <hex(handle)>/1 path and PUTs both.
  3. The ct128 upload succeeds; the ct64 PUT fails on an S3 hiccup.
  4. In the result loop, the ct128 success calls update_ct128_uploaded (aws_upload.rs:702) — which sets ciphertext128 = d128 and s3_format_version = 1. The ct64 failure only sets
  transient_error (aws_upload.rs:710). Then trx.commit() at aws_upload.rs:727 persists the ct128 update regardless — intentionally, so a retry doesn't re-upload what succeeded.
  5. But now the row has both digests non-NULL, so it never matches the pending query again. The "transient" error implies a retry that can never be scheduled for this row.

  End state: ct128 is fine (new path + attestation + legacy copy), but ct64 exists only at the legacy digest key with no attestation — and the row says s3_format_version = 1, i.e.,
  fully migrated.

  The mirror case is nastier. Legacy row with ct128 uploaded (ciphertext128 = d128, old digest key) and ct64 pending: recovery uploads ct64 at the new path → stamps version 1; the
  ct128 rewrite fails transiently → row is complete → never revisited. Now it's the ct128 new-path object that's missing — and that's the path the RFC's Step-1 shadow verifier HEADs.
  Shadow mode would at least surface this as an on-chain/off-chain discrepancy, but per the RFC those discrepancies are supposed to indicate verifier bugs or attestation gaps, and "do
  not proceed to Step 2 until the disagreement rate is zero" — so these rows become permanent noise blocking the validation gate, with no mechanism to fix themselves.



- Rows with ciphertext128_format = 0 now stall recovery. 0 (Unknown) is DB-legal per the CHECK constraint, but Ciphertext128Format::from_i16(0) returns None. The new code parses the format unconditionally: in fetch_pending_uploads it continues (row silently never recovered), and in the DatabaseLock arm the ? kills the uploader loop each iteration. The old code only parsed when ct128 bytes were present. Current writers never write 0, so this is latent — but gating the parse on "a real ct128 digest exists" would keep ct64-only recovery alive for such rows.

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
eudelins-zama: Is that case really possible? What would you submit to `addCiphertextMaterial` in such case ([here](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/CiphertextCommits.sol#L116))?
As long as we have the `CiphertextCommits` contract, I don't think this should be possible...

But indeed, it's not specified, and worth clarifying on when this case is possible, and how should the kms-connector handle this

rudy-6-4: KMS will not see it. In that case the ct128 is not uploaded, but it make the metadata of the ct64 with a dummy ct128 digest. (both digests are present for both ciphertexts)

antoniupop: If you're sure, then it's fine!

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
rudy-6-4: I have updated the debug printer

rudy-6-4:
@claude  /pr-review and also check that `Overall looks good, I think it's nearly ready. I spotted a couple issues (inline comments) and an agent found some more, the first one looks serious, second looks like it's imposible in practice, but could happen if we change behaviour later:

    s3_format_version can be set to 1 while the ct64 object was never migrated. The flag is row-level but updated per-variant. Concrete case: a pre-migration row with ct64 already uploaded (old digest key) and ct128 pending. Recovery re-uploads both at the new path; if ct128 succeeds but ct64's PUT fails transiently, update_ct128_uploaded (lib.rs:322) commits s3_format_version = 1 and the row now has both digests non-NULL — so fetch_pending_uploads (which only selects rows with a NULL digest) never picks it up again. Result: a row claiming v1 whose ct64 exists only at the legacy digest key with no attestation. Harmless in Step 1, but the RFC's pre-Step-2 migration ("all previously committed handles remain resolvable") would silently skip it if the backfill keys on s3_format_version = 1. Setting the version only after both variants are confirmed at the new path (or tracking it per-variant) would close this.

    s3_format_version is a row-level claim. The migration comment defines it as a property of the row's S3 state: 0 = initial format, 1 = handle and signed digest metadata. One value
    speaks for two objects (the ct64 object in bucket_ct64 and the ct128 object in bucket_ct128).
    But it's stamped per-variant. Both update_ct64_uploaded (lib.rs:345) and update_ct128_uploaded (lib.rs:317) set s3_format_version = 1 as a side effect of marking their own variant
    uploaded. Whichever variant completes first stamps the whole row as "migrated."
    Retry is keyed on digest NULLness, not on the version. fetch_pending_uploads selects WHERE ciphertext IS NULL OR ciphertext128 IS NULL (aws_upload.rs:750). For legacy rows
    (version 0), the rewrite of an already-uploaded variant is opportunistic: it happens only because the row got selected for the other variant's sake. Nothing in the selection
    predicate tracks "this v0 variant still needs its rewrite."

The failure walkthrough. Take the population this recovery logic was built for — a partially-uploaded legacy row. Say ct64 was uploaded pre-PR (object at digest key hex(d64), no
attestation), ct128 never made it: row has ciphertext = d64, ciphertext128 = NULL, backfilled s3_format_version = 0.

    The row matches the pending query (ct128 digest is NULL). Because version ≠ 1, recovery fetches bytes for both variants so it can rewrite the old ct64 object too.
    upload_ciphertexts finds neither object at the new <hex(handle)>/1 path and PUTs both.
    The ct128 upload succeeds; the ct64 PUT fails on an S3 hiccup.
    In the result loop, the ct128 success calls update_ct128_uploaded (aws_upload.rs:702) — which sets ciphertext128 = d128 and s3_format_version = 1. The ct64 failure only sets
    transient_error (aws_upload.rs:710). Then trx.commit() at aws_upload.rs:727 persists the ct128 update regardless — intentionally, so a retry doesn't re-upload what succeeded.
    But now the row has both digests non-NULL, so it never matches the pending query again. The "transient" error implies a retry that can never be scheduled for this row.

End state: ct128 is fine (new path + attestation + legacy copy), but ct64 exists only at the legacy digest key with no attestation — and the row says s3_format_version = 1, i.e.,
fully migrated.

The mirror case is nastier. Legacy row with ct128 uploaded (ciphertext128 = d128, old digest key) and ct64 pending: recovery uploads ct64 at the new path → stamps version 1; the
ct128 rewrite fails transiently → row is complete → never revisited. Now it's the ct128 new-path object that's missing — and that's the path the RFC's Step-1 shadow verifier HEADs.
Shadow mode would at least surface this as an on-chain/off-chain discrepancy, but per the RFC those discrepancies are supposed to indicate verifier bugs or attestation gaps, and "do
not proceed to Step 2 until the disagreement rate is zero" — so these rows become permanent noise blocking the validation gate, with no mechanism to fix themselves.

    Rows with ciphertext128_format = 0 now stall recovery. 0 (Unknown) is DB-legal per the CHECK constraint, but Ciphertext128Format::from_i16(0) returns None. The new code parses the format unconditionally: in fetch_pending_uploads it continues (row silently never recovered), and in the DatabaseLock arm the ? kills the uploader loop each iteration. The old code only parsed when ct128 bytes were present. Current writers never write 0, so this is latent — but gating the parse on "a real ct128 digest exists" would keep ct64-only recovery alive for such rows.` is taken into account by the lasts 3 commits.

rudy-6-4:
@claude /pr-review and also check that antoniu's last remarks are answered by the 3 last fixing commits.

antoniupop (dismissed):
LGTM!

eudelins-zama:
@mergifyio queue

eudelins-zama:
@Mergifyio unqueue

eudelins-zama:
@Mergifyio dequeue

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -242,239 +274,524 @@ async fn run_uploader_loop(
...
-                } else {
-                    drop(span);
-                    task.update_ct64_uploaded(&mut trx, digest).await?;
-                }
+            Err(err) => {
+                transient_error = Err(ExecutionError::InternalError(err.to_string()).into());
+                error!(error = %err, handle = handle_as_hex, "Upload task fail");
             }
         }
     }
```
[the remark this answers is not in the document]
rudy-6-4: The trx must be commited to ensure the task ins inserted in db. I am adding a comment.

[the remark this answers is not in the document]
rudy-6-4:
```
let mut trx = pool.begin().await?;

                let item = match job {
                    UploadJob::Normal(item) => {
                        item.enqueue_upload_task(&mut trx).await?;  // THIS NEED TO BE COMMITED
                        item
                    }
```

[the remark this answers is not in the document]
rudy-6-4: Done a more uniform commiting for new upload task .

coprocessor/fhevm-engine/sns-worker/src/lib.rs
```diff
@@ -292,53 +337,38 @@ impl HandleItem {
...
 
         info!(
-            "Mark ct64 as uploaded, handle: {}, digest: {}",
+            "Mark ciphertexts as uploaded, handle: {}, ct64_digest: {}, ct128_digest: {}, format: {:?}, s3_format_version: {}",
             to_hex(&self.handle),
-            to_hex(&digest)
+            to_hex(&ct64_digest),
+            to_hex(&ct128_digest),
+            self.ct128.format(),
+            s3_format_version,
```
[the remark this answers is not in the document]
rudy-6-4: Done

rudy-6-4:
E2E tests: https://github.com/zama-ai/fhevm/actions/runs/27698088386/job/81926717162

antoniupop (dismissed):
Looks all good!

antoniupop:
@mergifyio queue

eudelins-zama:
@mergifyio refresh

rudy-6-4:
Failed build during mergify, rerun: https://github.com/zama-ai/fhevm/actions/runs/27760137591

antoniupop (approved):

rudy-6-4:
@mergifyio queue

eudelins-zama:
@mergifyio refresh

PR #1856, opened 2026-01-29
refactor(coprocessor): remove tenants notion from DB and related code

dartdart26:
Replace with:
 * keys table - containing FHE keys
 * crs table - containing the CRS
 * host_chains table - containing host chain info

Rationale is:
 * the fact that we now have only one FHE keychain (that can have multiple FHE keys)
 * CRS and keys can be changed independently via events from the GW
 * we want to support multiple host chains
 * we want to remove remains of previous design with tenants, IDs, tenant API keys, etc.

We also simplify code where it no longer needs tenants, chain IDs, etc.

Co-Authored-By: Oba <obatirou@gmail.com>


obatirou:
It seems the sns code was not updated yet and still reference tenant_id

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
```diff
@@ -0,0 +1,69 @@
...
+ALTER TABLE keys DROP COLUMN acl_contract_address;
+
+
+-- allowed_handles.tenant_id no longer needed.
+ALTER TABLE allowed_handles DROP COLUMN tenant_id;
+ALTER TABLE allowed_handles ADD PRIMARY KEY (handle, account_address);
+
+-- ciphertext_digest.tenant_id no longer needed. Instead, put the host_chain_id there directly.
+ALTER TABLE ciphertext_digest ADD COLUMN host_chain_id BIGINT NOT NULL CHECK (host_chain_id > 0);
+UPDATE ciphertext_digest SET host_chain_id = (SELECT chain_id FROM keys WHERE tenant_id = ciphertext_digest.tenant_id);
```
obatirou: It seems host_chain_id column is added but with a NOT NULL clause without a default value. Does this work?
The data is filled with the next update statement.

dartdart26: Good catch! Let me fix it by making it NULL and then NOT NULL after the UPDATE.

Btw, on a somewhat related note, I am assuming there is one entry in the keys (former tenants) table, just FYI. I will document that in the migration.

dartdart26:
> It seems the sns code was not updated yet and still reference tenant_id

Yes, still not updated.

coprocessor/fhevm-engine/gw-listener/src/database.rs
```diff
@@ -1,114 +1,63 @@
...
+    let oid = write_large_object_in_chunks_tx(tx, &key_record.sns_pk, CHUNK_SIZE).await?;
+    let query = sqlx::query!(
+        "INSERT INTO keys (key_id, pks_key, sks_key, sns_pk)
+        VALUES ($1, $2, $3, $4)
+        ON CONFLICT (key_id) DO NOTHING",
+        key_record.key_id.as_ref(),
+        key_record.pks_key.as_ref(),
+        key_record.sks_key.as_ref(),
+        oid,
+    );
```
obatirou, line 48: might be rare to update the keys but writing the large object and then check on conflict do nothing would mean you could have a large object in DB that is referenced nowhere?

dartdart26: Good catch! Let me think how to handle that properly.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
```diff
@@ -0,0 +1,76 @@
...
+ALTER TABLE allowed_handles ADD PRIMARY KEY (handle, account_address);
+
+-- ciphertext_digest.tenant_id no longer needed. Instead, put the host_chain_id there directly.
+ALTER TABLE ciphertext_digest ADD COLUMN host_chain_id BIGINT DEFAULT NULL;
+UPDATE ciphertext_digest SET host_chain_id = (SELECT chain_id FROM keys WHERE tenant_id = ciphertext_digest.tenant_id);
+ALTER TABLE ciphertext_digest ALTER COLUMN host_chain_id SET NOT NULL;
+ALTER TABLE ciphertext_digest ADD CONSTRAINT ciphertext_digest_host_chain_id_positive CHECK (host_chain_id > 0);
+ALTER TABLE ciphertext_digest DROP COLUMN tenant_id;
+ALTER TABLE ciphertext_digest ADD PRIMARY KEY (handle);
+ALTER TABLE ciphertext_digest ADD COLUMN key_id BYTEA NOT NULL; -- TODO for the future is to make this an index
```
obatirou: Same here with NOT NULL without default value

dartdart26: Let me remember if we actually need that and, if yes, fix it.

coprocessor/fhevm-engine/gw-listener/src/database.rs
dartdart26: Actually, here we use a transaction. So either both fail or both succeed. Should be a non-issue, right?

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: Done, I made it empty and then I added NOT NULL.

coprocessor/fhevm-engine/gw-listener/src/database.rs
obatirou: In case of conflict it will not fail with the DO NOTHING ?
So if tx succeed but there was a conflict you would have the large object referenced nowhere 
Unless I miss something?

dartdart26: Oh, I misunderstood your comment. Let me look.

dartdart26: But I now do:

```
        ON CONFLICT (key_id_gw) DO UPDATE SET
            key_id = '',
            pks_key = EXCLUDED.pks_key,
            sks_key = EXCLUDED.sks_key,
            sns_pk = EXCLUDED.sns_pk",
        &key_record.key_id_gw,
        key_record.pks_key.as_ref(),
        key_record.sks_key.as_ref(),
        oid,
 ```
 
 So basically it will still update. Does that solve the issue you are thinking of?

Comment (obatirou) [reply coprocessor/fhevm-engine/gw-listener/src/database.rs:48]: Oh sorry, did not saw the new version of this part of code!
In that case yes. Do we want to keep the old oid in DB?

Comment (dartdart26) [reply coprocessor/fhevm-engine/gw-listener/src/database.rs:48]: Maybe yes, due to possible race conditions between events (for key change) and computation queues for SnS. But I haven't figured out everything at that stage, it could be that it is not strictly necessary.

Comment (obatirou) [reply coprocessor/fhevm-engine/gw-listener/src/database.rs:48]: Ok thank you 
Closing this thread now that the main issue is addressed.

**Comment by dartdart26** [comment]
E2E pass here: https://github.com/zama-ai/fhevm/actions/runs/21904504850/job/63241088115

**Comment by dartdart26** [comment]
@claude /pr-review

**Review by Eikix — DISMISSED** [review]
codex review:

Thanks for the huge refactor, direction looks right and the tenant cleanup is real progress.

I went through this theme-by-theme and aligned to the scope in fhevm-internal#925 (single keychain + multichain host support + no tenant API key). I’m requesting changes for a few blockers.

Blockers

1) key_id uniqueness is incompatible with current GW insert/update flow
- Migration adds `UNIQUE (key_id)` in `coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql`.
- GW writer always sets `key_id = ''` in `coprocessor/fhevm-engine/gw-listener/src/database.rs`.
- This prevents inserting multiple key rows (new `key_id_gw`) and directly conflicts with the multi-key model introduced by this refactor.

2) zkproof-worker key/CRS never refresh after startup
- `latest_key` and `latest_crs` are loaded once before the worker loop in `coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs`.
- After key/CRS rotation events, verifier keeps stale material until process restart.
- That breaks the “keys and CRS can evolve independently” part of the new model.

3) SNS key_id_gw provenance during rotation is unsafe
- In `coprocessor/fhevm-engine/sns-worker/src/executor.rs`, tasks are stamped with the current “latest” `key_id_gw` (there is already a TODO around this).
- During rotation, this can mismatch the actual key provenance for ciphertexts in the batch and push wrong `key_id_gw` into `ciphertext_digest`.

4) Helm init script still references removed API-key vars
- In `charts/coprocessor/templates/coprocessor-init-config.yaml`, script still uses `${API_KEY_SECRET_NAME}`/`${API_KEY_SECRET_KEY}` while API key values were removed from chart config.
- This can break init/config setup path and undercuts the config simplification goal.

Notes on dedupe
- I’m intentionally not reopening already-discussed threads about LO lifecycle / conflict handling since that discussion is already there.
- If useful, I can add focused follow-up comments per file once the above blockers are addressed.

File: coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql [hunk 4]
```diff
@@ -0,0 +1,122 @@
...
+);
+
+-- Move CRS from keys to crs. Use an empty ID for the existing CRS.
+INSERT INTO crs (crs_id, crs)
+    SELECT ''::BYTEA, public_params FROM keys;
+ALTER TABLE keys DROP COLUMN public_params;
+
+-- Host chains.
+CREATE TABLE host_chains (
+    chain_id BIGINT PRIMARY KEY NOT NULL CHECK (chain_id > 0),
```
Eikix: not sure it is important for this PR but solana doesn't have chainID, we'll have to find a formula for it
solana clusters (devnet testnet mainnet) identify by genesis blockhash

but we can decide to keep chain_id as bigint and just find a way to turn the genesis blockhash into a large number

coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs
```diff
@@ -0,0 +1,44 @@
+use anyhow::Result;
+use sqlx::{PgPool, Row};
+use std::collections::HashMap;
+
+#[derive(Clone)]
+pub struct HostChain {
+    pub chain_id: i64,
+    pub name: String,
+    pub acl_contract_address: String,
```
Eikix, line 10: why is this a string?

coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs
```diff
@@ -0,0 +1,44 @@
+use anyhow::Result;
+use sqlx::{PgPool, Row};
+use std::collections::HashMap;
+
+#[derive(Clone)]
+pub struct HostChain {
+    pub chain_id: i64,
+    pub name: String,
```
Eikix, line 9: do we need a name? shouldn't we do something like name_from_id(chain_id: u64) -> String util?

update: actually having a name is probably simpler

coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub acl_contract_address: String,
+}
+
+#[derive(Clone)]
+pub struct HostChainsCache {
+    map: HashMap<i64, HostChain>,
```
Eikix: do i understand correctly that we use i64 instead of u64 because of postgre format?

coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs
```diff
@@ -0,0 +1,44 @@
+use anyhow::Result;
+use sqlx::{PgPool, Row};
+use std::collections::HashMap;
+
+#[derive(Clone)]
+pub struct HostChain {
+    pub chain_id: i64,
```
Eikix: do we want chain_id to be signed? afaik its not signed (if we take EVM)

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: Good point. Maybe let's tackle that later and yeah, as you say, find a way to convert to BIGINT that is unique across all chains.

coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs
dartdart26: Because we store it as hex string (with potential checksum) across components. I agree we might change to bytes array, but I didn't want to change migrations, env vars, etc.

coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs
dartdart26: I put it just for convenience for whoever manages that table, not used by the software itself.

coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs
dartdart26: Yeah, unfortunately it is i64 everywhere across the code. I did check and there are no chains with ID bigger than 2 ^ 53, but no ideal. I prefer we don't change that everywhere in this PR.

coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs
dartdart26: We do and we also want it to be u256 (AFAIR), but see comment above.

Eikix:
Non-blocking, but worth tightening: we still mix `i64` and `u64` for host chain IDs and rely on `as u64` casts in runtime paths.

- `HostChain` stores `chain_id` as `i64`: [host_chains.rs#L7](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs#L7)
- ZK worker reads DB `host_chain_id` as `i64`: [verifier.rs#L281](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs#L281)
- Then casts to `u64` when encoding handle metadata: [verifier.rs#L539](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs#L539)
- Tx-sender also casts DB values with `as u64`: [delegate_user_decrypt.rs#L546](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs#L546)
- Meanwhile GW side defines chain id as `u64`: [gw-listener/src/lib.rs#L16](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/gw-listener/src/lib.rs#L16)

Could we introduce a shared `HostChainId` newtype (DB-facing `i64`, domain-facing checked `u64`) and keep conversions only at DB boundaries? It would make semantics explicit and remove silent cast risk.

dartdart26:
Re @claude review:

> 1. Large Object Leak on Key Update
> 2. Unique Constraint Violation When Inserting Multiple Keys

I am aware of those, but I want to handle them in a separate commit.

> 3. Migration Edge Case: Empty Keys Table

This shouldn't be possible as we can't have 0 keys and entries in ciphertext_digest, but I will put a guard in the migration.

> 4. Potential Infinite Loop in DbKeyCache::fetch()

I wanted to keep existing behavior, but I think it makes sense to return an error instead of assert.

> 5. Unwrap in SNS Executor

I agree, but I will not solve that issue in this PR.

Re minor stuff:

> 1. Line length violations: Several SQL strings exceed the 80-char limit from rustfmt.toml. This is common for SQL and may be acceptable.

I won't fix that.

> 2. Missing transaction commit: read_large_object_in_chunks() begins a transaction but doesn't explicitly commit (relies on implicit behavior on drop).

I think that is fine as is.

> 3. CrsCache/HostChainsCache never refresh: These caches are loaded once at startup. New CRS or host chains added while workers are running won't be visible until restart. This may be intentional but should be documented.

That will be done in the future as we add keys/chains on-the-fly.

dartdart26:
@claude I did the changes, check again.

dartdart26:
> Non-blocking, but worth tightening: we still mix `i64` and `u64` for host chain IDs and rely on `as u64` casts in runtime paths.
> 
> * `HostChain` stores `chain_id` as `i64`: [host_chains.rs#L7](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs#L7)
> * ZK worker reads DB `host_chain_id` as `i64`: [verifier.rs#L281](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs#L281)
> * Then casts to `u64` when encoding handle metadata: [verifier.rs#L539](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs#L539)
> * Tx-sender also casts DB values with `as u64`: [delegate_user_decrypt.rs#L546](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs#L546)
> * Meanwhile GW side defines chain id as `u64`: [gw-listener/src/lib.rs#L16](https://github.com/zama-ai/fhevm/blob/petar/db-keys-refactor/coprocessor/fhevm-engine/gw-listener/src/lib.rs#L16)
> 
> Could we introduce a shared `HostChainId` newtype (DB-facing `i64`, domain-facing checked `u64`) and keep conversions only at DB boundaries? It would make semantics explicit and remove silent cast risk.

Can we make that a separate issue and do it after this PR gets in?

dartdart26:
> 4. Helm init script still references removed API-key vars

Let me remove them.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
```diff
@@ -0,0 +1,128 @@
+BEGIN;
+
+-- Create the keys table.
+ALTER TABLE tenants RENAME TO keys;
```
rudy-6-4: Could be easier for readability (PR and later) to start with new table.
Copy and drop tenants.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
```diff
@@ -0,0 +1,128 @@
...
+-- key_id contains the key ID from the server key metadata (that is used in ciphertext metadata). key_id for the existing key is empty.
+-- key_id_gw contains the key ID from the GW event (that could be different from key_id).
+ALTER TABLE keys RENAME COLUMN key_id TO key_id_gw;
+ALTER TABLE keys ALTER COLUMN key_id_gw SET NOT NULL;
+ALTER TABLE keys ADD CONSTRAINT unique_key_id_gw UNIQUE (key_id_gw);
+ALTER TABLE keys ADD COLUMN key_id BYTEA NOT NULL DEFAULT ''::BYTEA;
+ALTER TABLE keys ALTER COLUMN key_id SET NOT NULL;
+ALTER TABLE keys ADD CONSTRAINT unique_key_id UNIQUE (key_id);
+
+ALTER TABLE keys ADD COLUMN created_at TIMESTAMPTZ NOT NULL DEFAULT NOW();
```
rudy-6-4: Could we add the key_id_gw meta info (from the KeyGen event) like the block and  the key digest.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
```diff
@@ -0,0 +1,128 @@
...
+);
+
+-- Move CRS from keys to crs. Use an empty ID for the existing CRS.
+INSERT INTO crs (crs_id, crs)
+    SELECT ''::BYTEA, public_params FROM keys;
+ALTER TABLE keys DROP COLUMN public_params;
+
+-- Host chains.
+CREATE TABLE host_chains (
+    chain_id BIGINT PRIMARY KEY NOT NULL CHECK (chain_id > 0),
```
rudy-6-4: Can't it be a string or a number. Bigint is not good to store u64.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
```diff
@@ -0,0 +1,128 @@
...
+ALTER TABLE allowed_handles DROP COLUMN tenant_id;
+ALTER TABLE allowed_handles ADD PRIMARY KEY (handle, account_address);
+
+-- input_blobs.tenant_id no longer needed.
+ALTER TABLE input_blobs DROP CONSTRAINT input_blobs_pkey;
+ALTER TABLE input_blobs DROP COLUMN tenant_id;
+ALTER TABLE input_blobs ADD PRIMARY KEY (blob_hash);
+
+-- ciphertext_digest.tenant_id no longer needed. Instead, put the host_chain_id there directly.
+ALTER TABLE ciphertext_digest ADD COLUMN host_chain_id BIGINT DEFAULT NULL;
```
rudy-6-4, line 80: same remarks as other chain_id column.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: I am not sure I got what you mean, could you please elaborate?

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: I think I agree. Let me see if I can do that easily.

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -44,6 +44,7 @@ enum DeviceSelection {
 pub struct Scheduler<'a> {
     graph: &'a mut DFComponentGraph,
     edges: Dag<(), ComponentEdge>,
+    #[cfg(not(feature = "gpu"))]
```
rudy-6-4, line 47: is it really for this PR ?

coprocessor/fhevm-engine/scheduler/src/dfg.rs
```diff
@@ -25,8 +25,6 @@ use fhevm_engine_common::types::{Handle, SupportedFheOperations};
 pub struct ExecNode {
     df_nodes: Vec<NodeIndex>,
     dependence_counter: AtomicUsize,
-    #[cfg(feature = "gpu")]
```
rudy-6-4, line 28: ?

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: Maybe we can make it a string and use decimal or other values for other types of IDs, cc @Eikix.

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
dartdart26: I made a warning disappear by doing so.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -181,64 +181,51 @@ async fn tfhe_worker_cycle(
 
             continue;
         }
-        query_tenants_and_keys(
-            &transactions,
-            &tenant_key_cache,
+
```
rudy-6-4, line 184: looks unrelated

coprocessor/fhevm-engine/scheduler/src/dfg.rs
dartdart26: I remove the `locality: i32,` variable below as it is not used.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
dartdart26: Not sure what you mean here?

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -709,32 +635,29 @@ async fn upload_transaction_graph_results<'a>(
...
         let mut s = tracer.start_with_context("update_computation", loop_ctx);
-        s.set_attribute(KeyValue::new("tenant_id", *tenant_id as i64));
         s.set_attributes(
             handles_to_update
                 .iter()
                 .map(|(h, _)| KeyValue::new("handle", format!("0x{}", hex::encode(h)))),
         );
         let (handles_vec, txn_ids_vec): (Vec<_>, Vec<_>) = handles_to_update.into_iter().unzip();
         let comp_updated = query!(
-                "
```
rudy-6-4, line 723: you should avoid reindent for this kind of PR (no asking to go back). (sorry if it's cargo fmt madness)

coprocessor/fhevm-engine/tfhe-worker/docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
     ports:
       - '5432:5432'
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      test: [ "CMD-SHELL", "pg_isready -U postgres" ]
```
rudy-6-4, line 14: ?

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
dartdart26: I haven't done any reindent, this is the output of `cargo fmt` as it passes now.

coprocessor/fhevm-engine/db-migration/initialize_db.sh
```diff
@@ -25,95 +25,18 @@ sqlx database create || { echo "Failed to create database."; exit 1; }
...
-  local size=$(psql "$db_url" -t -c "SELECT pg_size_pretty(SUM(octet_length(data))) FROM pg_largeobject WHERE loid = $oid" | tr -d ' ')
-  echo "Imported file. Size: $size" >&2
-  
-  echo "$oid"
-}
-
-echo "Fake OID"
-FAKE_KEY_FILE=$(mktemp)
-touch "$FAKE_KEY_FILE"
-SNS_PK_OID=$(import_large_file "$FAKE_KEY_FILE" "$DATABASE_URL")
```
rudy-6-4, line 108: where is it done now ?

dartdart26: This is not needed in this file. It is done in gw-listener.

coprocessor/fhevm-engine/tfhe-worker/docker-compose.yml
dartdart26: It looks like this is the correct yaml format.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
rudy-6-4: I was not sure it was in relation to the PR (could be rebase error).

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
rudy-6-4: There were some complains that information about keys were missing for debug or verification. Like the digests of the key so we can check if it's the one expected or the block the key was inserted for manual/visual verification.
But this PR is big, so it can wait.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: Or maybe juste BYTEA and that's it. And we convert to u256 when needed.

dartdart26: Actually, I don't know if this EIP is valid, but using 63 bits is more than enough: https://eips.ethereum.org/EIPS/eip-2294

dartdart26: Let's chat about whether we want to make this something bigger than i64 (which is 2 ^ 63 for unsigned and is quite a big number).

Re casting to u64 - I can go over the code and minimize that, but I think we need to keep it for metadata as it is expecting u64.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: Now that I think about it, we expect chain ID in the GW, so we'd have to have an ID anyway, whether it perfectly matches the host chain ID or not.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: Maybe we should do further changes in a separate PR. I don't think we have a requirement to have an u64 chain ID. @rudy-6-4 @Eikix

obatirou:
Note: a search all on this branch show there are still mention of tenants
ex:
* `pub type TenantId = i32;` in `tfhe_event_propagate.rs`
* `tenant_txs` parameter in `tfhe_worker.rs`
* `tenant-key-cache-size` in doc + `deamon_cli.rs`
etc

coprocessor/fhevm-engine/fhevm-engine-common/src/db_keys.rs
```diff
@@ -0,0 +1,439 @@
...
+    #[cfg(feature = "gpu")]
+    pub gpu_sks: Vec<tfhe::CudaServerKey>,
+
+    pub pks: tfhe::CompactPublicKey,
+
+    pub cks: Option<tfhe::ClientKey>,
+}
+
+const CHUNK_SIZE: i32 = 64 * 1024; // 64KiB
+pub async fn read_keys_from_large_object_by_key_id(
```
obatirou: Seems not used anywhere?
read_keys_from_large_object_by_key_id_gw is used in keyset.rs

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
```diff
@@ -0,0 +1,128 @@
...
+BEGIN
+    IF (SELECT COUNT(*) FROM keys) > 1 THEN
+        RAISE EXCEPTION 'Expected zero or one row in keys table, but found %', (SELECT COUNT(*) FROM keys);
+    END IF;
+END $$;
+
+ALTER TABLE keys DROP COLUMN tenant_api_key;
+ALTER TABLE keys DROP COLUMN is_admin;
+ALTER TABLE keys DROP COLUMN sns_sk;
+ALTER TABLE keys DROP COLUMN verifying_contract_address;;
```
antoniupop: ;;

dartdart26:
> Note: a search all on this branch show there are still mention of tenants ex:
> 
> * `pub type TenantId = i32;` in `tfhe_event_propagate.rs`
> * `tenant_txs` parameter in `tfhe_worker.rs`
> * `tenant-key-cache-size` in doc + `deamon_cli.rs`
>   etc

Thanks, I will have a look. I think I left some of the CLI parameters as aliases for backwards compatibility, temporarily.

coprocessor/fhevm-engine/fhevm-engine-common/src/db_keys.rs
dartdart26: Let me check!

dartdart26: Yeah, it is no longer used, let me remove it.

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: I introduced the ChainId type as suggested by @Eikix. It can't be constructed with an invalid value. It supports values from 0 to 2 ^ 63.

dartdart26:
> Note: a search all on this branch show there are still mention of tenants ex:
> 
> * `pub type TenantId = i32;` in `tfhe_event_propagate.rs`
> * `tenant_txs` parameter in `tfhe_worker.rs`
> * `tenant-key-cache-size` in doc + `deamon_cli.rs`
>   etc

Removing them.

coprocessor/fhevm-engine/fhevm-engine-common/src/host_chains.rs
dartdart26: As I mentioned in another comment, I added the `ChainId` type to be more explicit about it. Thanks for suggesting it!

coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql
dartdart26: I did as you suggested!

dartdart26:
@claude I've addressed all comments. Do you see something else that I need to look at?

obatirou (approved):

dartdart26:
E2E passed here: https://github.com/zama-ai/fhevm/actions/runs/21951874363

dartdart26:
@mergify queue

rudy-6-4 (approved):

PR #1857, opened 2026-01-29
fix(coprocessor): host-listener, improve new_log_stream with timeout

rudy-6-4:


rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

obatirou:
I do not see get_chain_id that was mentioned in the first PR?
It is not needed to add a timeout to it?

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -1116,8 +1091,11 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
             .map(|n| n - args.catchup_margin as i64);
     }
 
+    // Check connection works
+    log_iter.new_log_stream_no_retry().await?;
+
     loop {
-        log_iter.new_log_stream(true).await;
+        log_iter.stream = None; // force new connection each iteration
```
obatirou, line 1094: Why it is needed on each iteration?

rudy-6-4: It's to force using a new provider + connection to remove any state like the explicit call new_log_stream did before. Could be done in reset_for_catchup_loop.

rudy-6-4:
> I do not see get_chain_id that was mentioned in the first PR? It is not needed to add a timeout to it?

chain_id is only used at startup but yes in an automatic retry it could stuck.
adding it since it's simple

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -1116,8 +1091,11 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
             .map(|n| n - args.catchup_margin as i64);
     }
 
+    // Check connection works
+    log_iter.new_log_stream_no_retry().await?;
```
obatirou, line 1091: Why this check without retry?
This will force a restart?

rudy-6-4: Because we want it fail quick on start up, like if the url is wrong or the node is down, it should be the visible for operator. E.g. during upgrade or on argocd directly. Prevous behavior was a limited number of retry.

rudy-6-4: Yes it shut down the host listener, so it will restart and retry on production environment.

rudy-6-4:
@claude /pr/review

obatirou (approved):

rudy-6-4:
@mergifyio queue

rudy-6-4:
@mergifyio queue

PR #1870, opened 2026-02-01
fix(coprocessor): use npm ci for deterministic builds

enitrat:
## Summary

Use `npm ci` with proper lockfile access for deterministic Docker builds.

**Problem:** The Dockerfiles were using `npm install` which:
- Can modify `package-lock.json` if versions don't match
- Resolves dependencies at build time (non-deterministic)
- Produces different results depending on npm registry state

**Solution:** Copy the root `package-lock.json` into the build context and use `npm ci`:

**Files changed:**
- `coprocessor/fhevm-engine/host-listener/Dockerfile`
- `coprocessor/fhevm-engine/Dockerfile.workspace`

### DIFF
--- a/coprocessor/fhevm-engine/Dockerfile.workspace
+++ b/coprocessor/fhevm-engine/Dockerfile.workspace
@@ -41,9 +41,9 @@ COPY package.json package-lock.json ./
 COPY host-contracts ./host-contracts
 
 # Compile host-contracts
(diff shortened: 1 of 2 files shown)


antoniupop (approved):

rudy-6-4 (approved):

enitrat:
@mergify queue

PR #1879, opened 2026-02-02
fix(coprocessor): cut cyclic deps in host-listener ordering

Eikix:
## Summary
- prune cyclic back-edges during topological ordering so dependency_count can reach zero
- keep depth_size consistent after pruning
- add a 6-node cycle regression test mirroring max observed cycle size

Closes https://github.com/zama-ai/fhevm-internal/issues/946

## Testing
- cargo fmt --check
- SQLX_OFFLINE=true cargo check -p host-listener --all-targets
- SQLX_OFFLINE=true cargo clippy -p host-listener --all-targets -- -W clippy::perf -W clippy::suspicious -W clippy::style -D warnings
- cargo test -p host-listener test_topological_order_cycle_cut_max_observed

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
@@ -278,18 +278,31 @@ fn topological_order(
                     let mut pruned_inputs = input_txs;
                     pruned_inputs.retain(|dep| !cycle_deps.contains(dep));
                     let mut depth_size = 0;
+                    let mut remaining_unseen = Vec::new();
                     for input_tx in &pruned_inputs {
-                        if let Some(dep_tx) = txs.get(input_tx) {
-                            depth_size =
-                                depth_size.max(dep_tx.depth_size + dep_tx.size);
+                        let Some(dep_tx) = txs.get(input_tx) else {
+                            continue;
+                        };
+                        depth_size =
+                            depth_size.max(dep_tx.depth_size + dep_tx.size);
+                        if !done_tx.contains(input_tx) {
+                            remaining_unseen.push(*input_tx);
                         }
                     }
                     if let Some(tx) = txs.get_mut(&tx_hash) {
                         tx.input_tx.retain(|dep| !cycle_deps.contains(dep));
-                        tx.depth_size = depth_size;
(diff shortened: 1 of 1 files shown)


Eikix:
@claude /pr-review

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -260,14 +265,29 @@ fn topological_order(
...
                 for unseen_tx_hash in unseen.iter() {
                     error!("Reordering transaction: tx {:?} depends on unseen tx {:?}", tx, txs.get(unseen_tx_hash));
                     if stacked_tx.contains(unseen_tx_hash) {
                         error!("Fake cyclic dependency detected for transaction {:?}, cutting", tx_hash);
                         cut_cycle = true;
+                        cycle_deps.insert(*unseen_tx_hash);
                     }
                 }
                 if cut_cycle {
+                    let mut pruned_inputs = input_txs;
```
rudy-6-4, line 287: I think this should be merged with the general case line 261.
depth size handling is duplicated with   reordered and  done_tx update

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -240,23 +264,8 @@ fn topological_order(
...
-                        depth_size =
-                            depth_size.max(dep_tx.depth_size + dep_tx.size);
-                    }
-                }
-                if !done_tx.contains(input_tx) {
-                    unseen.push(*input_tx);
-                }
-            }
+            let (unseen, depth_size) =
+                compute_unseen_and_depth(&input_txs, &txs, &done_tx);
```
rudy-6-4, line 268: ok it's still done twice (when cycle) but at least the code is shared

Eikix: > ok it's still done twice but at least the code is shared

should I change something?

rudy-6-4: No. Let's go with that

rudy-6-4 (approved):

Eikix:
@mergify queue

PR #1881, opened 2026-02-02
fix(coprocessor): gate test bypass behind feature flag

enitrat:
The `extract_server_key_without_ns` function had a hardcoded bypass for `"key_bytes"` that was part of production code

Wrap the bypass in `#[cfg(feature = "test_bypass_key_extraction")]` and enable it only for dev builds via self-referential dev-dependency (already done in other parts of the codebase)


rudy-6-4 (approved):

antoniupop (approved):

enitrat:
@mergify queue

enitrat:
@mergifyio requeue

PR #1884, opened 2026-02-02
fix(coprocessor): honor ordered txs for handle deps

Eikix:
## Summary
- honor block order when mapping handle -> producing tx and when building tx dependency maps
- drop duplicate allowed handle warning (expected for trivial encrypt cases)

Closes https://github.com/zama-ai/fhevm-internal/issues/947

## Testing
- cargo check --manifest-path coprocessor/fhevm-engine/Cargo.toml
- cargo clippy --manifest-path coprocessor/fhevm-engine/Cargo.toml

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
@@ -91,7 +91,7 @@ fn tx_of_handle(
             continue;
         };
         for handle in &tx.allowed_handle {
-            handle_creator.entry(*handle).or_insert(tx.tx_hash);
+            handle_creator.insert(*handle, tx.tx_hash);
         }
     }
     for tx_hash in ordered_txs_hash {
--- a/coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
@@ -77,6 +77,7 @@ fn scan_transactions(
 }
 
 fn tx_of_handle(
+    ordered_txs_hash: &[TransactionHash],
     ordered_txs: &HashMap<TransactionHash, Transaction>,
 ) -> (
     HashMap<Handle, TransactionHash>,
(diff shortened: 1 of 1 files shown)


Eikix:
Follow-up: tx_of_handle now iterates ordered_txs_hash so time order wins for duplicate handle producers; HashMap is only for lookup. Commit c80f1a11.

Eikix:
Rudy note addressed: fill_tx_dependence_maps now iterates ordered_txs_hash as well (commit fc790584), so both creator/consumer maps and dependency population follow initial order.

Eikix:
Squashed the two ordering fixes into one commit (78f236d4) and force-pushed. Ordered txs are now used for handle-creator selection and dependency map population.

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -85,12 +86,30 @@ fn tx_of_handle(
...
+            continue;
+        };
         for handle in &tx.allowed_handle {
-            handle_creator.insert(*handle, tx.tx_hash);
+            match handle_creator.entry(*handle) {
+                Entry::Vacant(e) => {
+                    e.insert(tx.tx_hash);
+                }
+                Entry::Occupied(e) => {
+                    warn!(
```
rudy-6-4: You don't want to log, that occurs frequently if one can allow a trivial encrypt.

rudy-6-4 (dismissed):

Eikix:
Addressed Rudy’s note: removed the duplicate allowed-handle warning (trivial encrypt duplicates are expected). Commit 23f23011.

Eikix:
History cleaned: squashed to a single commit (508f9d42) and force-pushed.

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -85,12 +86,18 @@ fn tx_of_handle(
     // handle to tx maps
     let mut handle_creator = HashMap::new(); // no intermediate value
     let mut handle_consumer = HashMap::new();
-    for tx in ordered_txs.values() {
+    for tx_hash in ordered_txs_hash {
+        let Some(tx) = ordered_txs.get(tx_hash) else {
+            continue;
+        };
         for handle in &tx.allowed_handle {
-            handle_creator.insert(*handle, tx.tx_hash);
```
rudy-6-4, line 90: so this is choosing the first one instead of the last one on purpose ?

rudy-6-4 (dismissed):

Eikix:
Re: duplicate producer choice — I switched it to match 041b9915: we iterate in block order and overwrite, so the *latest* producer in the block wins deterministically (commit 5c7dbebe).

Eikix:
Squashed again; single commit now 12b0041f. Behavior matches 041b9915: iterate in block order, last producer wins deterministically.

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
Eikix: Yes — intentionally. To match 041b9915 behavior, we iterate in block order and overwrite, so the *latest* producer in the block wins deterministically. (Now in 12b0041f.)

rudy-6-4 (approved):

Eikix:
@mergify queue

PR #1898, opened 2026-02-03
fix(coprocessor): backport hotfix for HL dependence counters updates when false cycles present

antoniupop:


Eikix (dismissed):

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -85,12 +86,18 @@ fn tx_of_handle(
     // handle to tx maps
     let mut handle_creator = HashMap::new(); // no intermediate value
     let mut handle_consumer = HashMap::new();
-    for tx in ordered_txs.values() {
+    for tx_hash in ordered_txs_hash {
+        let Some(tx) = ordered_txs.get(tx_hash) else {
+            continue;
+        };
```
rudy-6-4, line 92: As suggested in a previous code by Elias, let's keep the first tx that create the handle. So a tx cannot référence a futur handle when it exists twice in the block and is not in the cache.

coprocessor/fhevm-engine/scheduler/src/dfg.rs
```diff
@@ -331,12 +331,11 @@ impl DFComponentGraph {
                     } else {
                         dependence_pairs.push((producer[0].0, consumer));
                     }
-                } else {
-                    self.needed_map
-                        .entry(i.clone())
-                        .and_modify(|uses| uses.push(consumer))
-                        .or_insert(vec![consumer]);
                 }
+                self.needed_map
```
rudy-6-4, line 335: So even if it's produced it could use the existing one if ready ? It won't rely on waiting the existing one if not ready ?

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
rudy-6-4: Superseeded by fully relying on tx order

PR #1900, opened 2026-02-04
fix(coprocessor): dependence chain update during reorg in case of collisions (#1768)

antoniupop:
Backport from release/v0.10.x

fix(coprocessor): improve dependence chain update in HL during reorg in case of collisions

fix(coprocessor): allow TFHE worker to lock earliest dependence chain on no progress

chore(coprocessor): add index for pending dependence chains where dependence count is not zero

fix(coprocessor): add threshold for ignoring dependency_count and make dependent list unique


Eikix (approved):
Backport check only: patch-id matches e2918f4a on release/0.10.x (PR #1768). Single commit, no extra changes. LGTM.

PR #1901, opened 2026-02-04
fix(coprocessor): host-listener, trust tx & logs order from get_logs

rudy-6-4:


Eikix:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

Eikix:
@claude /pr-review can you explain how this PR solves each potential edge case (multiple txs in the block that generate the same handle, etc.), any type of cycle. TLDR -> think of what could go wrong and look whether this PR covers every edge case.

Eikix:
@claude /pr-review can you explain how this PR solves each potential edge case (multiple txs in the block that generate the same handle, etc.), any type of cycle. TLDR -> think of what could go wrong and look whether this PR covers every edge case.

Eikix:
Perhaps we can add a guard that does a local sort by (blockNumber, transactionIndex, logIndex) on txs, to harden the assumption (ethereum blocks are ordered but perhaps our RPC node not)
comparatively not too expensive?

rudy-6-4:
> Perhaps we can add a guard that does a local sort by (blockNumber, transactionIndex, logIndex) on txs, to harden the assumption (ethereum blocks are ordered but perhaps our RPC node not) comparatively not too expensive?

This is for one block only. yes if the index are available in logs, we could check the index.

PR #1904, opened 2026-02-04
chore(gateway-contracts): remove MultichainACL contract

zmalatrax:
Remove MultichainACL from Gateway contracts and coprocessor. ACL verification is now handled by the KMS instead of on-chain checks, eliminating state duplication, reducing gas costs, and simplifying the Gateway architecture.

  **Changes:**
  - Delete MultichainACL.sol, IMultichainACL.sol, MultichainACLChecks.sol and related examples/mocks
  - Remove ACL checks from Decryption.sol (public, user, and delegated decryption flows)
  - Remove ACL propagation operations from coprocessor transaction-sender
  - Update deployment scripts, upgrade tasks, and tests

  Closes zama-ai/fhevm-internal#848

`isUserDecryptionReady` and `isDelegatedUserDecryptionReady` have an unused parameter, here they are simply silenced; their fate will be determined in a subsequent PR, tracked by https://github.com/zama-ai/fhevm-internal/issues/1001


gateway-contracts/contracts/Decryption.sol
isaacdecoded, line 1: Note that we also need to bump the [contract](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L214-L216) and [reinitializer](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L224) versions, and the [reinitializerVX method](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L308) as well.

.github/workflows/gateway-contracts-upgrade-tests.yml
isaacdecoded, line 1: As the decryption contract is being updated, we should enable the upgrade test for it [here](https://github.com/zama-ai/fhevm/pull/1904/files#diff-439c64678a6191c111c4a14343dd38cf212877d607101d59609fb8746eefa58eR134-R142)

host-contracts/contracts/ACL.sol
```diff
@@ -271,6 +271,7 @@ contract ACL is
         /**
          * @dev Arbitrum block timestamps may be up to one hour ahead of L1.
          *
+         * TODO: still needed once multichainACL on gateway has been removed?
          * Since the expiration is propagated to the Gateway MultichainACL contract deployed on Arbitrum,
```
isaacdecoded: I think this comment will also deprecate, right? 🤔 I mean, there shouldn't be any MultichainACL propagation anymore after this PR. Indeed, more than the comment, we should double-check if this propagation logic is still relevant or if it was removed already from the Coprocessor side.

cc @rudy-6-4 @dartdart26

zmalatrax: Yes, I've added the comment to flag this design choice and decide whether it should be deleted or not 🙏

.github/workflows/gateway-contracts-upgrade-tests.yml
zmalatrax: indeed, will set the flag to true (as it is still not automatic afaiu)

zmalatrax: done in https://github.com/zama-ai/fhevm/pull/1904/changes/f2b28db9e2e01197b5231248836df9d9942d9094

gateway-contracts/contracts/Decryption.sol
zmalatrax: ty!
done in https://github.com/zama-ai/fhevm/pull/1904/changes/f2b28db9e2e01197b5231248836df9d9942d9094

enitrat:
@claude /pr-review

.github/workflows/gateway-contracts-upgrade-tests.yml
```diff
@@ -133,7 +133,7 @@ jobs:
       #  See https://github.com/zama-ai/fhevm-internal/issues/379
       - name: Upgrade Decryption contract
         working-directory: current-fhevm/gateway-contracts
-        if: false
+        # if: true
```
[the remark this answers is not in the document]
zmalatrax: if: true triggers linter (`if: false` does not though), it's commented out

gateway-contracts/contracts/Decryption.sol
```diff
@@ -714,22 +683,18 @@ contract Decryption is
         CtHandleContractPair[] calldata ctHandleContractPairs,
         bytes calldata /* extraData */
     ) external view virtual returns (bool) {
+        // Silence unused variable warning
+        userAddress;
```
isaacdecoded: Shouldn't we remove the userAddress argument instead? Or are we keeping this to avoid breaking other components (e.g., the Relayer)?

gateway-contracts/contracts/Decryption.sol
```diff
@@ -744,34 +709,17 @@ contract Decryption is
         CtHandleContractPair[] calldata ctHandleContractPairs,
         bytes calldata /* extraData */
     ) external view virtual returns (bool) {
+        // Silence unused variable warning
+        delegationAccounts;
```
isaacdecoded: Idem https://github.com/zama-ai/fhevm/pull/1904/changes#r2811337539.

.github/workflows/gateway-contracts-upgrade-tests.yml
isaacdecoded: I think this should be uncommented so the CI will trigger this step that tests the Decryption contract upgrade.

gateway-contracts/contracts/Decryption.sol
```diff
@@ -693,13 +665,10 @@ contract Decryption is
             return false;
         }
 
-        // For each handle, check that it is allowed for public decryption and that the ciphertext
-        // material represented by it has been added.
+        // For each handle, check that the ciphertext material represented by it has been added.
```
isaacdecoded, line 668:
```suggestion
        // Check that ciphertext material has been added for each cthandle.
```

gateway-contracts/contracts/Decryption.sol
```diff
@@ -693,13 +665,10 @@ contract Decryption is
...
+        // ACL checks are performed by the KMS.
```
isaacdecoded, line 669: Not sure if we want to mention this KMS responsibility here. Or maybe we can tag it as a `@dev ACL checks are performed by the KMS.` NatSpec.

gateway-contracts/contracts/Decryption.sol
zmalatrax: iirc only the relayer is using this function ; I think removing the unused parameters is the best choice (between keeping it as is, removing unused params and entirely removing the function)
It's tracked in this issue: https://github.com/zama-ai/fhevm-internal/issues/1001

isaacdecoded: I also agree with removing them 👍 maybe cc @manoranjith

.github/workflows/gateway-contracts-upgrade-tests.yml
zmalatrax: `if: true` triggers the action-lint ; advising to remove the line entirely.

I'm fine with removing the line, but commenting it is clearer to avoid forgetting to put the flag back for future releases where decryption contract does not need an upgrade. (This manual if-flag thing is quite error-prone in any case)

It was set to `if: false` before ; I can reset it and comment it, having `# if: false` instead.

isaacdecoded: Ah you're absolutely right, so sorry for looping-back to it 🙏

gateway-contracts/contracts/Decryption.sol
zmalatrax: Indeed we can simply remove it, not part of the function requirements anymore

gateway-contracts/contracts/Decryption.sol
zmalatrax: removed in [57b2114](https://github.com/zama-ai/fhevm/pull/1904/commits/57b21147868ef204197ffb6fd54d51ff060ed89a)

isaacdecoded (approved):
LGTM! 👍

PR #1905, opened 2026-02-04
feat(coprocessor): throttle dependent ops

Eikix:
## What
- DB-backed monotonic clamp for dependent ops scheduling (prevents schedule_order inversions across main/catchup/poller).
- New dependence_chain_schedule table (1 row per chain) used to clamp via upsert+returning.
- Limiter stays optional; dependent-ops-rate-per-min=0 still disables.

## Why
- Throttling writes schedule_order into the future; later inserts with smaller schedule_order can invert execution order.
- Per-replica state isn’t enough with multiple HL types. DB clamp gives a single monotonic timeline across restarts/instances.

## How
- For dependent ops only: compute schedule_order, apply limiter delay, then clamp_schedule_order_db:
  - INSERT … ON CONFLICT DO UPDATE with GREATEST(last + 1µs, computed)
  - returns clamped schedule_order
- New migration: 20260204120000_dependence_chain_schedule.sql

## Impact / load
- +1 upsert per dependent op (PK lookup, no scans).
- Contention only on hot chains (already serialized work).
- Non-dependent ops unaffected.

## Risks / mitigations
- Hot-row contention → limited to single chain; acceptable for testnet throttling.
- Missing migration → clear startup failure; documented.
- Precision edge (µs clamp) → relies on DB timestamp precision; uses 1µs to preserve order.

## Why this is minimal
- No new worker, queue, or scheduler changes.
- No protocol changes; uses existing schedule_order.

## Tracking
- Closes https://github.com/zama-ai/fhevm-internal/issues/1003
- Related PR: https://github.com/zama-ai/fhevm/pull/1907

## Testing
- cargo fmt --manifest-path coprocessor/fhevm-engine/host-listener/Cargo.toml
- SQLX_OFFLINE=true cargo check -p host-listener
- SQLX_OFFLINE=true cargo clippy -p host-listener -- -D warnings


Eikix:
Quick note on Antoniu's inversion concern: the PoC already clamps schedule_order to be monotonic per dependence chain when limiter is enabled (schedule_order = max(base+defer, last_scheduled+1µs)). This prevents TXb from being scheduled earlier than TXa even if delay shrinks later.\n\nCaveat: the clamp state is in-memory per replica; it resets on restart and doesn’t coordinate across replicas. If we need strict monotonicity across restarts/replicas we’d need to persist last_scheduled per chain (bigger change).

Eikix:
Next step: implement a minimal PoC that uses DB as source of truth for dependency limit! ETA tmr

Eikix:
@claude /pr-review

Eikix:
Closing, in favour of https://github.com/zama-ai/fhevm/pull/1907 which replaces it

PR #1907, opened 2026-02-04
feat(coprocessor): slow lane for dependent ops

Eikix:
## What
- Add binary scheduling priority on `dependence_chain`: `0` (fast) / `1` (slow).
- Classify slow chains in host-listener ingest using **unweighted** dependent-op count (`+1` per newly inserted, allowed TFHE op with dependencies) per chain per ingest pass.
- Add `--dependent-ops-max-per-chain` (`0` disables slow-lane classification).
- Persist monotonic priority with `GREATEST(existing, incoming)` so concurrent HL types cannot downgrade.
- Keep worker fast-first ordering in normal path (`schedule_priority`, then `last_updated_at`).
- Keep oldest-first fallback (`acquire_early_lock`) as liveness escape hatch when no progress is possible.

## Inheritance model
- Add ingest-only parent metadata (`inheritance_parents`) to improve slow-lane inheritance without changing scheduling behavior.
- Scheduling still uses `dependencies` (no-fork parallelism behavior unchanged).
- Slow-lane inheritance now uses `inheritance_parents` so parallel splits are less likely to drop lineage for throttling decisions.

## Off mode (`--dependent-ops-max-per-chain=0`)
- No new slow-lane decisions are made.
- Startup promotes chains back to fast (`schedule_priority=0`) using advisory-lock serialized, **batched** updates.

## Why
- Isolate heavy dependent chains from normal traffic without dropping data.
- Avoid backoff-style priority churn/inversions.
- Improve protection against parent-slow/child-fast splits when chain lineage is partially split for parallelism.

## How
- DB migration adds `schedule_priority` and aligns pending-chain index with worker acquisition order.
- Ingest computes per-chain dependent-op totals and marks over-cap chains as slow.
- Ingest also:
  - inherits slow from known slow parents in DB,
  - propagates transitively to dependents in the current batch graph.

## Impact
- Hot path remains in-memory for classification/propagation.
- Priority persistence is monotonic and idempotent.
- Off-mode promotion is bounded per batch to reduce DB pressure spikes.

## Tracking
- Closes https://github.com/zama-ai/fhevm-internal/issues/1003
- Related PR: https://github.com/zama-ai/fhevm/pull/1905

## Validation
- `SQLX_OFFLINE=true cargo +1.91.1 clippy -p host-listener --all-targets -- -D warnings`
- `SQLX_OFFLINE=true cargo +1.91.1 clippy -p tfhe-worker --all-targets -- -D warnings`
- `SQLX_OFFLINE=true cargo +1.91.1 test -p host-listener --test host_listener_integration_tests test_slow_lane_threshold_matrix_locally -- --nocapture`
- `SQLX_OFFLINE=true cargo +1.91.1 test -p host-listener --test host_listener_integration_tests test_slow_lane_cross_block_sustained_below_cap_stays_fast_locally -- --nocapture`
- `SQLX_OFFLINE=true cargo +1.91.1 test -p host-listener --test host_listener_integration_tests test_slow_lane_cross_block_parent_lookup_finds_known_slow_parent_locally -- --nocapture`
- `SQLX_OFFLINE=true cargo +1.91.1 test -p host-listener --test host_listener_integration_tests test_slow_lane_priority_is_monotonic_across_blocks_locally -- --nocapture`
- `SQLX_OFFLINE=true cargo +1.91.1 test -p host-listener --test host_listener_integration_tests test_slow_lane_off_mode_promotes_all_chains_on_startup_locally -- --nocapture`
- `SQLX_OFFLINE=true cargo +1.91.1 test -p tfhe-worker test_acquire_early_lock_ignores_priority -- --nocapture`


Eikix:
@claude /pr-review

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -783,7 +902,7 @@ impl Database {
                 .iter()
                 .map(|h| h.to_vec())
                 .collect::<Vec<_>>();
-            sqlx::query!(
+            sqlx::query(
```
rudy-6-4: can't we keep compile time verified query ?  schedule_lane is just an integer, should work.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -280,6 +376,29 @@ impl Database {
...
         let is_scalar = !scalar_byte.is_zero();
         let output_handle = result.to_vec();
+        let dependence_chain_id = log.dependence_chain.to_vec();
+        let schedule_order =
+            log.block_timestamp
+                .saturating_add(TimeDuration::microseconds(
+                    log.tx_depth_size as i64,
+                ));
+        let has_dependencies = log.is_allowed && !dependencies.is_empty();
+        if has_dependencies {
```
rudy-6-4: can't it be done on the caller side ?
the db object  should no handle that.

rudy-6-4 (commented):
If easy could be done in ingest function instead of database struct. (I understand the existing state database is handy for that)

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: ingest would be  perfect but insert_tfhe_event is ok

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -280,6 +376,29 @@ impl Database {
...
+            if let Some(limiter) = &self.dependent_ops_limiter {
+                let defer_for = limiter.lock().await.defer_duration();
+                if defer_for.is_zero() {
```
rudy-6-4: you probably want to do it per chain once for all instead  of each ops ?
so could move to  update_dependence_chain

Eikix:
@claude /pr-review

Eikix:
I am still thinking about the mental model of the slow lane, will take some time to think of something fair and good.

Eikix:
@claude /pr-review

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
```diff
@@ -175,7 +177,7 @@ impl LockMngr {
                             AND
                             dependency_count = 0     -- No pending dependencies
                         )
-                ORDER BY last_updated_at ASC        -- FIFO
+                ORDER BY schedule_priority ASC, last_updated_at ASC -- highest priority first
```
antoniupop, line 180: Index?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -945,6 +999,18 @@ pub fn acl_result_handles(event: &Log<AclContractEvents>) -> Vec<Handle> {
     }
 }
 
+pub fn tfhe_dependent_op_weight(op: &TfheContractEvents) -> u32 {
+    use TfheContract as C;
+    use TfheContractEvents as E;
+    match op {
+        E::FheMul(C::FheMul { .. }) => 7,
+        E::FheDiv(C::FheDiv { .. }) => 5,
+        E::FheRem(C::FheRem { .. }) => 8,
```
antoniupop: Where are these coming from?

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -125,6 +125,13 @@ pub struct Args {
     )]
     pub dependence_cross_block: bool,
 
+    #[arg(
+        long,
+        default_value_t = 0,
+        help = "Max weighted dependent ops per chain before slow-lane (0 disables)"
+    )]
+    pub dependent_ops_max_per_chain: u32,
```
antoniupop, line 130: How would this feature be turned off completely?

coprocessor/fhevm-engine/db-migration/migrations/20260205130000_dependence_chain_schedule_priority.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE dependence_chain
+  RENAME COLUMN schedule_lane TO schedule_priority;
```
antoniupop: Only one migration needed

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
antoniupop, line 180: It feels like this adds a perf impact on DB load even if we try to disable the feature... I think.

antoniupop (changes requested):

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Added the provenance inline above `tfhe_dependent_op_weight`: these are HCULimit relative ratios used as weights (mul=7x, div=5x, rem=8x, baseline=1x).

coprocessor/fhevm-engine/db-migration/migrations/20260205130000_dependence_chain_schedule_priority.sql
Eikix: Done in d531e99b. I collapsed this to one migration: `20260204130000_dependence_chain_schedule_priority.sql`, and removed the follow-up rename migration.

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
Eikix: Added index alignment in the same migration. We now recreate `idx_pending_dependence_chain` as `(schedule_priority, last_updated_at, dependence_chain_id)` with the same partial predicate, so it matches the acquisition ORDER BY.

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
Eikix: Now fully off with `--dependent-ops-max-per-chain=0`: no slow-lane decisions, no dependent-op throttling accounting, and host-listener/poller reset existing `schedule_priority` to 0 on startup.

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
Eikix: Addressed two parts here: (1) ingest now skips slow-lane accounting work entirely when cap=0, and (2) startup reset forces all priorities back to fast. So disable mode goes back to FIFO behavior, and the pending-chain index now matches the new ORDER BY.

Eikix:
@claude /pr-review

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
Eikix: Follow-up in 2181c22a: agreed on DB load concern. We dropped the unbounded startup UPDATE and replaced it with bounded batch reset under off-mode, scoped to pending schedulable rows (`status=updated`, `worker_id IS NULL`, `dependency_count=0`) with `FOR UPDATE SKIP LOCKED`.

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
Eikix: Follow-up in 2181c22a: removed startup full-table reset (too heavy). In off-mode (`--dependent-ops-max-per-chain=0`), we now do bounded resets in ingest (`5000` pending slow chains max per block tx), plus no new slow-lane decisions.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: I understand that, but I think Claude (or codex) is oversimplifying by a huge margin - and that's where the human needs to step in and check. If you look at HCU, it has a lot more entries because that's based off of bench numbers from TFHE-rs. 
Here the AI has decided to omit info about whether operands are both encrypted or not and instead pick numbers from different groups in different ops (not sure there's a logic to choices here), which matters: 64bit MUL scalar is ~200ms on CPU while same REM is ~650ms and DIV ~430.

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
antoniupop: order by sched prio may have an impact as well - not sure

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
Eikix: Follow-up validation: this is now lazy promotion only (bounded reset batches in ingest, no eager per-chain promote-on-touch path). Clippy all-targets is clean on head.

Eikix: Updated in a8f2c20b: I simplified this further. Off-mode now promotes seen chains to fast in the same ingest transaction (no bounded reset helper path anymore).

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Agree, good catch. Current weights are only a throttle signal, not a runtime-accurate HCU model. They’re intentionally coarse, but I agree this can look like false precision. If you want, I can simplify this PR to unweighted dependent-op counting (1 per dependent op) and keep the same per-chain cap behavior.

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
Eikix: Agreed this can impact DB behavior. Current mitigation is: (1) index now matches worker ORDER BY (`schedule_priority, last_updated_at, dependence_chain_id`) for pending unlocked rows, and (2) off-mode (`--dependent-ops-max-per-chain=0`) stops creating new slow chains and promotes seen slow chains back to fast during ingest. So off-mode converges back toward FIFO without a full-table reset.

Eikix:
Quick strategy recap (scope: https://github.com/zama-ai/fhevm-internal/issues/1003):

- Keep scheduling model binary per dependence chain: fast (`0`) / slow (`1`).
- Mark a chain slow only at ingest when its dependent-op load for that block exceeds `--dependent-ops-max-per-chain`.
- Persist monotonic priority with `GREATEST(existing, incoming)` so concurrent HL types cannot downgrade a chain.
- Worker acquisition stays `ORDER BY schedule_priority, last_updated_at`: fast lane first, slow lane when fast is empty.
- Off mode (`--dependent-ops-max-per-chain=0`) creates no new slow-lane decisions and lazily promotes seen slow chains back to fast during ingest (no full-table reset).

This keeps the patch focused while isolating heavy dependent chains from normal traffic.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: If we can't refine, then let's keep these as approximate worst case cost per op.

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
antoniupop: As you add a cli param, you also need to update the charts (fhevm/charts/coprocessor/values.yaml)

antoniupop (changes requested):
Please add charts - otherwise ok. I'll approve after update

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
Eikix: Done in f009646d. I added `--dependent-ops-max-per-chain` in `charts/coprocessor/values.yaml` for all three host-listener flavors (`hostListener`, `hostListenerPoller`, `hostListenerCatchupOnly`) so the new CLI knob is surfaced in charts.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Applied in b1f7956d. I simplified to a conservative coarse ordering for throttling only: `rem=6`, `div=4`, `mul=2`, baseline `1` (and `0` for non-op events). This keeps it worst-case oriented without pretending to model full HCU dimensions.

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
Eikix: Also fixed charts CI in b1f7956d by bumping `charts/coprocessor/Chart.yaml` version to `0.7.14` (values changed).

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Updated in c9507b9e: we simplified to unweighted counting (`+1` per dependent op) and removed op-specific weighting code. This makes the knob semantics explicit (`max dependent ops per chain per block`) and avoids pseudo-precision.

Eikix:
Removed weights, it'll make reasoning about the setting easier

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
```diff
@@ -251,6 +257,188 @@ async fn setup(node_chain_id: Option<u64>) -> Result<Setup, anyhow::Error> {
...
+    let mut payload = Vec::with_capacity(
+        "trivialEncrypt".len() + std::mem::size_of::<[u8; 32]>() + 1,
+    );
+    payload.extend_from_slice("trivialEncrypt".as_bytes());
+    payload.extend_from_slice(&val.to_be_bytes::<32>());
+    payload.push(to_type);
+    keccak256(payload)
+}
+
+fn fhe_add_handle(
```
Eikix, line 276: don't we have a better way to create fixtures / scenarios?

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
```diff
@@ -251,6 +257,188 @@ async fn setup(node_chain_id: Option<u64>) -> Result<Setup, anyhow::Error> {
...
+    )
+    .await?;
+    assert!(
+        receipts.iter().all(|receipt| receipt.status()),
+        "every burst tx must succeed"
+    );
+    Ok((receipts, current))
+}
+
+async fn chain_id_for_output_handle(
```
Eikix: never use chain_id imo but dep_chain_id to never ambiguate host chain id and dependence chain id

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
Eikix: Yep, agreed. I refactored this by adding `ingest_dependent_burst_seeded(...)` so tests compose scenarios with one helper instead of repeating `emit + ingest + options` blocks. That keeps the slow-lane test intent clearer and reduces fixture noise.

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
Eikix: Done. Renamed helper + variables to `dep_chain_id` in this path (`dep_chain_id_for_output_handle`, `heavy_dep_chain_id`, `fast_dep_chain_id`) so we don’t confuse host chain id with dependence-chain id.

Eikix:
Pushed a follow-up commit with the local validation hardening we discussed.

What changed:
- Added `test_slow_lane_threshold_matrix_locally` (cap=64):
  - depth=63 -> fast (`schedule_priority=0`)
  - depth=64 -> fast (`schedule_priority=0`)
  - depth=65 -> slow (`schedule_priority=1`)
- Added `test_slow_lane_cross_block_sustained_below_cap_stays_fast_locally` to lock current behavior across multiple blocks (below-cap bursts stay fast).
- Updated the host-listener README runbook with the key-bootstrap gate and recovery checks before slow-lane validation (ActivateKey + `sns_pk` OID change + sns-worker keyset fetch).

Validation run locally:
- package fmt/check/clippy (via pre-commit on commit)
- `cargo +1.91.1 test -p host-listener --test host_listener_integration_tests test_slow_lane_threshold_matrix_locally -- --nocapture`
- `cargo +1.91.1 test -p host-listener --test host_listener_integration_tests test_slow_lane_cross_block_sustained_below_cap_stays_fast_locally -- --nocapture`

Please take another look when you have a minute.

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -173,8 +239,78 @@ pub async fn ingest_block_logs(
             chains,
             block_timestamp,
             &block_logs.summary,
+            &schedule_priority_by_chain,
         )
         .await?;
     }
+    if !slow_lane_enabled {
+        let promoted = db
+            .promote_seen_schedule_priorities_fast(&mut tx, &seen_chain_ids)
```
Eikix: this is a convoluted name, not so clear what the util does first glance

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -173,8 +239,78 @@ pub async fn ingest_block_logs(
...
+                count = promoted,
+                "Slow-lane disabled: promoted seen chains to fast"
+            );
+        }
+    }
     tx.commit().await
 }
+
+#[cfg(test)]
+mod tests {
```
Eikix, line 327: does this respect the repo's style? to add tests in the same file as a module?

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -173,8 +239,78 @@ pub async fn ingest_block_logs(
...
+    use fhevm_engine_common::types::SchedulePriority;
+
+    use super::compute_schedule_priorities;
+    use crate::database::tfhe_event_propagate::{Chain, ChainHash};
+
+    fn mk_chain(last_byte: u8) -> Chain {
```
Eikix: what is this?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -773,8 +839,13 @@ impl Database {
         chains: OrderedChains,
         block_timestamp: PrimitiveDateTime,
         block_summary: &BlockSummary,
+        schedule_priority_by_chain: &HashMap<ChainHash, SchedulePriority>,
     ) -> Result<(), SqlxError> {
         for chain in chains {
+            let chain_id = chain.hash.to_vec();
```
Eikix: chain_id should be reserved for host chain id, dep_chain_id is less ambiguous

Eikix:
Small cleanup follow-up pushed in `205bf45d`.

What changed:
- Renamed ambiguous dep-chain identifiers (`chain_id` -> `dep_chain_id`) in dependence-chain update path.
- Renamed off-mode promotion helper to be explicit (`promote_seen_dep_chains_to_fast_priority`).
- Renamed slow-lane classification helper/tests in ingest for clearer intent (`classify_slow_lane_priorities`, `fixture_dep_chain`).

No behavior change intended.

Validation:
- `cargo +1.91.1 clippy -p host-listener --all-targets -- -D warnings`
- `cargo +1.91.1 test -p host-listener --lib`

Could you take another pass on readability/maintainability?

Eikix:
Pushed one follow-up commit to make the local validation path reliable:

- `99e89620` hardens `test-suite/fhevm/scripts/validate-slow-lane.sh`
  - avoids `lo_get(sns_pk)` full LOB reads (was failing with `large object read request is too large`)
  - infers/exports missing `COPROCESSOR_*_VERSION` vars from running containers before compose overrides

What I re-ran:
- `./scripts/validate-slow-lane.sh --integration-only`
  - passes all 3 deterministic assertions:
    - threshold matrix
    - cross-block sustained below cap stays fast
    - off-mode promotion-on-seen

Important observation from stack runs:
- `operators` load with `cap=1` is not a strong threshold validator because it often emits ~1 counted dependent op per tx-chain, so slow-lane may not trigger even when load is high.
- This means integration tests are currently the auditable signal for classifier correctness; stack mode is still useful as a smoke path.

Would appreciate a fresh review on current head when you have a moment.

Eikix:
Pushed one follow-up commit to make the local validation path reliable:

- `99e89620` hardens `test-suite/fhevm/scripts/validate-slow-lane.sh`
  - avoids `lo_get(sns_pk)` full LOB reads (was failing with `large object read request is too large`)
  - infers/exports missing `COPROCESSOR_*_VERSION` vars from running containers before compose overrides

What I re-ran:
- `./scripts/validate-slow-lane.sh --integration-only`
  - passes all 3 deterministic assertions:
    - threshold matrix
    - cross-block sustained below cap stays fast
    - off-mode promotion-on-seen

Important observation from stack runs:
- `operators` load with `cap=1` is not a strong threshold validator because it often emits ~1 counted dependent op per tx-chain, so slow-lane may not trigger even when load is high.
- This means integration tests are currently the auditable signal for classifier correctness; stack mode is still useful as a smoke path.

Would appreciate a fresh review on current head when you have a moment.

Eikix:
Test comment via gh api.

Eikix:
Pushed `79be2ebb` with the latest validation hardening.

What changed:
- Added deterministic contention scenario for slow-lane validation:
  - `test-suite/e2e/contracts/operations/SlowLaneContention.sol`
  - `test-suite/e2e/test/slowlane/slowLaneContention.ts`
- Heavy chain now marks intermediate results with `FHE.allowThis(...)` so ingest-side counting (`is_allowed && has_deps`) is actually exercised.
- Updated `validate-slow-lane.sh` to run the deterministic e2e grep by default and hardened bootstrap gate to avoid false negatives when key material is already present.
- Updated host-listener README runbook accordingly.

Acceptance evidence from local stack:
- `./scripts/validate-slow-lane.sh --stack-only --cap 2` now passes.
- Observed post-baseline: `total=2, fast=1, slow=1`.
- DB audit for same run showed heavy chain with `allowed_dep_ops=8` and light chain with `allowed_dep_ops=1`.

Integration checks also pass:
- `test_slow_lane_threshold_matrix_locally`
- `test_slow_lane_cross_block_sustained_below_cap_stays_fast_locally`
- `test_slow_lane_off_mode_promotes_seen_chain_locally`

Could someone take another review pass on this revision?

Eikix:
Pushed `92517566` with the fallback-ordering hardening.

What changed:
- `acquire_early_lock` now orders by `schedule_priority ASC, last_updated_at ASC` (same fast-first ordering model as `acquire_next_lock`).
- Added regression test `test_acquire_early_lock_prefers_fast_lane`.

Why:
- This closes the gap where the no-progress fallback path could bypass slow-lane priority and pick slow work ahead of fast work.

Validation:
- `SQLX_OFFLINE=true cargo +1.91.1 test -p tfhe-worker prefers_fast_lane -- --nocapture`
- `rustfmt +1.91.1 --edition 2021 --check coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs coprocessor/fhevm-engine/tfhe-worker/src/tests/dependence_chain.rs`

Can you take another pass on this update?

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -20,6 +21,43 @@ pub struct BlockLogs<T> {
...
 }
 
+#[derive(Copy, Clone, Debug)]
+pub struct IngestOptions {
+    pub dependence_by_connexity: bool,
+    pub dependence_cross_block: bool,
+    pub dependent_ops_max_per_chain: u32,
+}
+
+fn classify_slow_lane_priorities(
```
Eikix: it's still not clear what this function does. 
perhaps a docomment might help 

why do we need to return number of allowed and number of throttled? and why did we call these variables like this?

Eikix: i'm surprised this isn't doable with a one line filter in rust, and inlined (such that this doesn't need to be a function)

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -139,14 +176,31 @@ pub async fn ingest_block_logs(
...
+    let seen_dep_chain_ids = if slow_lane_enabled {
+        Vec::new()
+    } else {
+        chains.iter().map(|chain| chain.hash.to_vec()).collect()
+    };
+    let mut dependent_ops_by_chain: HashMap<ChainHash, u64> = HashMap::new();
     for tfhe_log in tfhe_event_log {
         let inserted = db.insert_tfhe_event(&mut tx, &tfhe_log).await?;
         at_least_one_insertion |= inserted;
+        if slow_lane_enabled
```
Eikix: can you justify how these 4 conditions "AND"ed are a good proxy for dep chain size ++?

coprocessor/fhevm-engine/db-migration/migrations/20260204130000_dependence_chain_schedule_priority.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE dependence_chain
+  ADD COLUMN IF NOT EXISTS schedule_priority SMALLINT NOT NULL DEFAULT 0;
+
+-- Keep lock acquisition ordering index aligned with:
+-- ORDER BY schedule_priority ASC, last_updated_at ASC
+DROP INDEX IF EXISTS idx_pending_dependence_chain;
```
Eikix, line 6: how can we build confidence that is this the right index?

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
Eikix: Good point. I removed that helper entirely in `bb2b9c76` and inlined classification in `ingest_block_logs`.

I also renamed metrics locals to `fast_lane_ops` / `slow_lane_ops` so the meaning is explicit.

coprocessor/fhevm-engine/db-migration/migrations/20260204130000_dependence_chain_schedule_priority.sql
Eikix: How I’d justify confidence on this index:

1) query shape match: worker hot-path uses `WHERE status=updated AND worker_id IS NULL AND dependency_count=0` + `ORDER BY schedule_priority, last_updated_at`.
2) index shape match: partial predicate is identical, key prefix is identical.
3) behavior check: contention tests validate fast-first acquisition ordering.

Also note: this index is for `acquire_next_lock` (the schedulable path). We now keep fast-first ordering in `acquire_early_lock` too, but that query intentionally ignores `dependency_count` and uses a different path.

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
Eikix: Yep, fair ask.

Those conditions are now explicit in code as named booleans + comment:
- `inserted`: count only new rows in this ingest pass (avoid over-counting duplicates/replays)
- `is_allowed`: only work that can actually become schedulable now
- `has_dependencies`: only dependent ops (not roots/trivial events)

So this metric tracks “new dependent schedulable work per chain per ingest”, which is the intended proxy for lane assignment.

coprocessor/fhevm-engine/db-migration/migrations/20260204130000_dependence_chain_schedule_priority.sql
Eikix: Good point. I added an explicit migration-contract test in `host_listener_integration_tests` (commit `6075b748`) so this is auditable in CI, not just reasoned manually.

It asserts:
- `schedule_priority` exists with type/default/nullability contract
- `idx_pending_dependence_chain` exists
- key order is `schedule_priority -> last_updated_at -> dependence_chain_id`
- partial predicate still matches schedulable rows (`status=updated`, `worker_id IS NULL`, `dependency_count=0`)

Validation run:
`SQLX_OFFLINE=true cargo +1.91.1 test -p host-listener --test host_listener_integration_tests test_schedule_priority_migration_contract -- --nocapture`

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -280,6 +304,12 @@ impl Database {
     ) -> Result<bool, SqlxError> {
         let is_scalar = !scalar_byte.is_zero();
         let output_handle = result.to_vec();
+        let dependence_chain_id = log.dependence_chain.to_vec();
```
Eikix: we don't need this 'beauty' change. it adds noise to the PR for no reason

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -39,6 +43,17 @@ pub type ScalarByte = FixedBytes<1>;
...
 pub type ChainHash = TransactionHash;
 
+static SLOW_LANE_MARKED_CHAINS_TOTAL: LazyLock<IntCounterVec> = LazyLock::new(
+    || {
+        register_int_counter_vec!(
+            "host_listener_slow_lane_marked_chains_total",
+            "Number of dependence chains marked slow by host-listener classification",
+            &["chain_id"]
+        )
+        .unwrap()
```
Eikix: do we use unwraps for metric declaration in the codebase?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -773,8 +823,16 @@ impl Database {
         chains: OrderedChains,
         block_timestamp: PrimitiveDateTime,
         block_summary: &BlockSummary,
+        slow_dep_chain_ids: &HashSet<ChainHash>,
     ) -> Result<(), SqlxError> {
         for chain in chains {
+            let dep_chain_id = chain.hash.to_vec();
```
Eikix: is this variable used anywhere?

test-suite/fhevm/scripts/validate-slow-lane.sh
```diff
@@ -0,0 +1,351 @@
+#!/usr/bin/env bash
```
Eikix: do we need this? a lot of opaque LoC, not sure it is worth it to commit.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Kept compile-time SQL verification on the persistence path () for dependence-chain updates.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Classification stays in ingest. DB helpers only persist chain state and are no longer deciding lane assignment themselves.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Done: slow-lane classification is per chain in ingest, then persisted once during .

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: We removed operation weighting from this PR. It is now unweighted  per inserted allowed dependent op, so the cap semantics stay explicit.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Applied. I removed the extra temp vars and inlined those one-use values to keep this diff smaller.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Switched this metric registration to  so failure context is explicit.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Good catch. That temp var was only used once; it is now inlined/removed.

test-suite/fhevm/scripts/validate-slow-lane.sh
Eikix: Removed from this PR. I kept only the concise local-stack notes in the host-listener README.

Eikix:
Pushed follow-up `5ced7000` to address the inheritance concern across block cuts.

What changed:
- If a current-batch chain depends on a parent chain that is already slow in DB, we now inherit slow for that child chain.
- Then we propagate transitively within the current ingest graph (parent -> child -> grandchild) before persisting priorities.
- This keeps monotonic behavior (`GREATEST` in upsert stays unchanged) and avoids reintroducing per-chain backoff inversions.

Scope:
- only `host-listener` ingest + one DB helper query
- no migration
- no full-table reset scan
- off mode (`--dependent-ops-max-per-chain=0`) still bypasses slow classification

Validation run:
- `SQLX_OFFLINE=true cargo +1.91.1 clippy -p host-listener --all-targets -- -D warnings`
- `SQLX_OFFLINE=true cargo +1.91.1 test -p host-listener --lib`

Eikix:
Risk map (what can still go wrong, and how this PR responds)

1) If a chain exceeds `--dependent-ops-max-per-chain` in one ingest block,
then that chain is marked slow (`schedule_priority=1`).
Response: worker picks fast first; heavy chain is deprioritized.

2) If a child chain is in the same ingest graph and depends on a slow parent,
then child inherits slow (transitively).
Response: avoids parent-slow/child-fast split for known edges in current batch.

3) If parent was already slow from previous blocks,
then current child can still inherit slow.
Response: ingest now reads slow parent ids from DB and propagates before upsert.

4) If parent is marked slow concurrently *after* this ingest read,
then child may stay fast for this pass.
Response: monotonic `GREATEST` prevents downgrade; next ingest/fallback can recover.

5) If dependency edges are missing/cut (restart cache loss, block-local visibility, cycle-cut ambiguity),
then inheritance cannot infer the relation.
Response: worker no-progress path eventually uses `acquire_oldest_lock` to recover progress.

6) If host-listener restarts and cuts perceived chains,
then some future chains can be classified fast until edges are seen again.
Response: persisted slow chains stay slow in DB; known parent links still inherit.
Limitation: truly unseen links are not protected by inheritance.

7) If feature is off (`--dependent-ops-max-per-chain=0`),
then no new slow decisions are made.
Response: expected behavior; system runs without slow-lane policy.

8) If attacker spreads load across many independent chains all below cap,
then this PR will not classify them as slow.
Response: out of scope for this patch (needs separate policy knobs).

9) If fast lane keeps selecting non-executable work due to imperfect dependency counters,
then throughput degrades before fallback triggers.
Response: `acquire_oldest_lock` is the escape hatch (recovery, not prevention).

Direct answer to the cache-clean concern:
- Yes, one cache clean / graph cut can temporarily make new chains look fast if dependency links are unseen.
- So this feature is strong on known edges + persisted slow parents, but not a formal guarantee under incomplete dependency visibility.

Eikix:
CI note: reran GPU setup again.

- Failing check remains `coprocessor-gpu-tests/setup-instance`
- Re-ran job multiple times; non-GPU checks remain green
- Current retry in progress: https://github.com/zama-ai/fhevm/actions/runs/21861466136/job/63117711751

This still looks infra-transient (`Start remote instance`) rather than code-regression.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -981,6 +981,35 @@ pub fn is_ebytes_type(inp: i16) -> bool {
...
+pub enum SchedulePriority {
+    #[default]
+    Fast = 0,
+    Slow = 1,
+}
+
+impl SchedulePriority {
+    pub const FAST: Self = Self::Fast;
+    pub const SLOW: Self = Self::Slow;
+}
```
obatirou: nit: not sure this is useful?

.github/workflows/charts-helm-checks.yml
```diff
@@ -44,6 +44,24 @@ jobs:
...
+              --set hostListener.enabled=true \
+              --set hostListenerPoller.enabled=true \
+              --set hostListenerCatchupOnly.enabled=true \
+              --set dependentOps.maxPerChain=${cap_sentinel})"
+            all_count="$(printf '%s\n' "$rendered" | grep -c -- '--dependent-ops-max-per-chain=' || true)"
+            shared_count="$(printf '%s\n' "$rendered" | grep -c -- "--dependent-ops-max-per-chain=${cap_sentinel}" || true)"
+            if [ "$all_count" -ne 3 ] || [ "$shared_count" -ne 3 ]; then
+              echo "expected 3 shared --dependent-ops-max-per-chain=${cap_sentinel} args, got all=${all_count} shared=${shared_count}"
+              exit 1
+            fi
```
obatirou: Seems unnecessary: `--dependent-ops-max-per-chain={{ int .Values.dependentOps.maxPerChain }}`
after `toYaml <hostListenerTYPE>` is sufficient to ensure the 3 host listener use the same `maxPerChain` no?

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -981,6 +981,35 @@ pub fn is_ebytes_type(inp: i16) -> bool {
...
+
+impl From<i16> for SchedulePriority {
+    fn from(value: i16) -> Self {
+        if value <= 0 {
+            Self::Fast
+        } else {
+            Self::Slow
+        }
+    }
+}
```
obatirou: Shouldn't this be 0 or 1 else raise errror?
Might be better to have explicit value.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -129,12 +144,79 @@ impl Database {
...
+
+        let rows = sqlx::query(
+            r#"
+            SELECT dependence_chain_id
+            FROM dependence_chain
+            WHERE schedule_priority = $1
+              AND dependence_chain_id = ANY($2::bytea[])
+            "#,
+        )
+        .bind(i16::from(SchedulePriority::SLOW))
```
obatirou: nit: it is bind here but not in promote_seen_dep_chains_to_fast_priority

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -36,14 +44,43 @@ fn block_date_time_utc(timestamp: u64) -> PrimitiveDateTime {
     PrimitiveDateTime::new(offset.date(), offset.time())
 }
 
+fn propagate_slow_lane_to_dependents(
```
obatirou, line 47: As slow is sticky, would it make sense to be more conservative on which dependent chain to be marked slow?
My understanding is that here if we have a chain A marked as slow, all dependent chains will be marked as slow and subsequent children also.
According to the [comment](https://github.com/zama-ai/fhevm/pull/1907#issuecomment-3877235375), it is to avoid parent-slow/child-fast split for known edges in current batch.
But in case there is too much propagation, the fast lane could be emptied?

Eikix: Good point. Yes, fast lane can shrink in worst case if a big connected subgraph is rooted in a slow parent.

That tradeoff is intentional here: we prefer preserving executable order (no parent-slow/child-fast split) over aggressive parallelism, because the split caused stalls/inversion-like behavior.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
Eikix: Fair call. I removed those alias constants (`FAST`/`SLOW`) and kept plain enum variants only.

So now this stays just:
- `SchedulePriority::Fast`
- `SchedulePriority::Slow`

.github/workflows/charts-helm-checks.yml
Eikix: Yep, agreed. This sentinel check was redundant and I dropped that workflow step.

We now rely on the chart values/rendering directly instead of a hardcoded CI-only assertion.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
Eikix: Agree on explicitness.

In the latest version I removed the unused `i16 -> SchedulePriority` conversion entirely (we weren’t reading typed priority from DB in this path).

If we add inbound parsing later, I’ll make it strict 0/1 + error.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Yep, good catch.

I aligned it: `promote_seen_dep_chains_to_fast_priority` now binds the fast value as well (`SET schedule_priority = $1`), same style as the other query.

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
obatirou: Ok so it is accepted that in case of a big spike due to a new contract deployed and a lot of new contracts depending on this one which are highly used, all dependance chains depending on those will be marked as slow.
So slow lane can be become huge (even if not necessary slow as it was only a spike).
It seems this is a edge case to note and that could be mitigated by disabling slow lane temporarily right?
Should this be documented? 
If other edges cases are known (with potential solution), a preemptive runbook regarding this feature could make sense.

Also thinking about dormant chain: if the fast lane is filed with activity, the escape hatch never trigger.
Hence, we could have slow lane never being picked up.
A solution would be again to disable temporarily the slow lane or increase number of workers.

Let me know if I miss something.

Eikix: Yes true, slow lane can grow very fast on a spike, that's to help conserve throughput for non 'problematic' traffic. 

Also true, if fast lane is always busy, then slow lane can stall. But we envision slow lane mostly for testnet only. 

Runbook is def a good idea, so handler knows that they can disable slowlane. Note that disabling slow lane is lazy: slow dep chains will be promoted to fast when they are seen again (to avoid big DB updates ~all slow => fast)

obatirou: Also thinking that if only for testnet, deployments should fail if deployed on other networks with this option.

obatirou: Could be a check in main before the run

antoniupop: I wouldn't entirely reject using this on main some day. It would be a fair last-line of defence against dos.

antoniupop (dismissed):
Please do add the info in the runbook about disabling slow lane.

Eikix:
Adding a runbook draft here since Notion access is restricted on my side.

Suggested insertion: `COP-RB01` -> Resolution -> new **Step 2.5**.

**Step 2.5 — Slow-lane side-effect check (testnet only)**  
*Note: slow lane is normally enabled only on testnet.*

Use metrics + logs in the same incident window.

1. **Metric signal (Prometheus/Grafana):** check `rate(host_listener_slow_lane_marked_chains_total[5m])`. If this remains `> 0` while uncompleted computations rise and completion rate is flat/low, slow-lane is a likely contributor.
2. **Log correlation (Grafana Loki):**
   - host-listener: confirm ongoing slow-lane / schedule-priority activity
   - tfhe-worker: confirm repeated no-progress / fallback behavior in the same period
3. **Temporary mitigation (testnet):** disable slow lane by setting `dependentOps.maxPerChain=0` through standard deployment flow, applied to all host-listener types together (`main`, `poller`, `catchup-only`).
4. **Recovery confirmation:** if completion resumes and backlog slope decreases shortly after disabling, classify incident as likely slow-lane side effect and escalate for threshold/policy tuning before re-enabling.
5. If no recovery, continue standard COP-RB01 flow (worker/HL health, catchup, capacity).

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -129,12 +145,88 @@ impl Database {
...
+    pub async fn promote_all_dep_chains_to_fast_priority(
+        &self,
+    ) -> Result<u64, SqlxError> {
+        let mut connection = self.pool().await.acquire().await?;
+        sqlx::query_scalar::<_, bool>("SELECT pg_advisory_lock($1)")
+            .bind(SLOW_LANE_RESET_ADVISORY_LOCK_KEY)
+            .fetch_one(connection.deref_mut())
+            .await?;
+
+        let rows = sqlx::query(
```
Eikix: use query!

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
Eikix, line 47: is this unit tested?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -129,12 +145,88 @@ impl Database {
...
+    pub async fn find_slow_dep_chain_ids(
+        &self,
+        tx: &mut Transaction<'_>,
+        dep_chain_ids: &[Vec<u8>],
+    ) -> Result<HashSet<ChainHash>, SqlxError> {
+        if dep_chain_ids.is_empty() {
+            return Ok(HashSet::new());
+        }
+
+        let rows = sqlx::query(
```
Eikix: use query!

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
Eikix: Yes. We cover this with host-listener integration tests: test_slow_lane_threshold_matrix_locally, test_slow_lane_cross_block_sustained_below_cap_stays_fast_locally, and test_slow_lane_off_mode_promotes_all_chains_on_startup_locally. Given these + unit coverage around lane ordering, I do not think we need extra tests in this PR.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Done in 72109d93. Switched the slow-lane helper queries in tfhe_event_propagate to `query!`, and regenerated workspace SQLx metadata.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: Done in the same commit (72109d93): converted this helper to `query!` as well and updated `.sqlx` accordingly.

Eikix:
@claude /pr-review 
- does this PR respect the conventions from this repo?
- is it well tested?
- what could go wrong? list the edge cases and look at how this PR addresses every possible edge case introduced by the slow lane feature
- what kind of worst case pressure does the slow lane (or inheritence of slow lane for dep chains) apply on the DB
- is the code lean and high quality?

charts/coprocessor/values.yaml
```diff
@@ -1,6 +1,14 @@
...
 # FHEVM Coprocessor Configuration
 # =============================================================================
+
+# -----------------------------------------------------------------------------
+# Shared host-listener dependence settings (applies to all HL types)
+# -----------------------------------------------------------------------------
+dependentOps:
+  # Max dependent ops per chain per ingested block before slow-lane.
+  # 0 disables slow-lane decisions.
+  maxPerChain: 0
```
rudy-6-4, line 11: do we only enable it when testnet is under load ?

charts/coprocessor/templates/coprocessor-host-listener-catchup-only-deployment.yaml
```diff
@@ -47,7 +47,12 @@ spec:
           image: {{ .Values.hostListenerCatchupOnly.image.name }}:{{ .Values.hostListenerCatchupOnly.image.tag }}
           command: ["host_listener"]
           args:
-{{ toYaml .Values.hostListenerCatchupOnly.args | nindent 12 }}
+            {{- range $arg := .Values.hostListenerCatchupOnly.args }}
```
rudy-6-4, line 50: not sure what's the benefit of this ?

Eikix: afaik we iterate args to drop any manually set --dependent-ops-max-per-chain=*
then append one canonical value from dependentOps.maxPerChain so all HL types use exactly the same cap without duplicates.

charts/coprocessor/values.yaml
Eikix: i think we enable it by default on testnet with decent value (like ~64)

Eikix:
@claude /pr-review Can you zoom into the integration test scenario, as well as the most critical functions of the PR, and go through an intentional protocol to build confidence that this PR does what it says it does.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -235,6 +354,11 @@ impl Database {
     ) -> Result<bool, SqlxError> {
         let is_scalar = !scalar_byte.is_zero();
         let output_handle = result.to_vec();
+        let schedule_order =
```
Eikix: no beautifying, this is out of scope for the PR

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
```diff
@@ -241,7 +243,7 @@ impl LockMngr {
                     status = 'updated'      -- Marked as updated by host-listener
                     AND
                     worker_id IS NULL       -- Ensure no other workers own it
-                ORDER BY last_updated_at ASC        -- FIFO
+                ORDER BY last_updated_at ASC, schedule_priority ASC
```
antoniupop, line 246: The schedule priority ordering shouldn't matter here as we only select one. I guess it's possible that two have the same earliest time order and both be executable at the same time.

coprocessor/fhevm-engine/host-listener/README.md
```diff
@@ -33,6 +33,76 @@ By default the listener propagate TFHE operation events to the database.
 You can change the database url using --database-url, it defaults to a local test database url.
 If you want to disable TFHE operation events propagation, you can provide an empty database-url.
 
+### Dependent ops throttling (optional)
+
+One flag enables slow‑lane assignment for heavy dependent chains:
+
+- `--dependent-ops-max-per-chain` (0 disables)
```
antoniupop: There's no explanation of what this flag actually means:
- what dependence patterns are concerned (is it dependence depth)
- what is the scope (is it looking specifically at just block level or taking into account past blocks)
- what are sensible values (with a justification - e.g. on CPU, FHE op cost mutiplier is X compared to GPU Y so if we're on a chain with block duration Z then we can at most cope with so many dep ops and we need to set this at such and such values based on arch)

antoniupop (dismissed):
Looks good - the readme or some comment in the charts (values) should maybe give some details on what this means and how to pick good values as it will be very host-dependent. Moving to a more uniform time rate rather than a block rate could be good later so it can just be one number for all hosts based on copro HW

coprocessor/fhevm-engine/host-listener/README.md
```diff
@@ -39,16 +39,40 @@ One flag enables slow‑lane assignment for heavy dependent chains:
...
+  - newly inserted,
+  - `is_allowed = true`,
+  - and has at least one input handle (depends on previous outputs).
+- It is **not** cumulative across past blocks.
+- It is **not** dependency depth.
+
+Why only allowed ops are counted:
+
+- Slow lane is meant to throttle near-term executable pressure.
+- Non-allowed ops are ACL-gated and may never become executable in the same horizon; counting them would over-throttle.
```
antoniupop: This is not quite accurate, non-allowed just means that their outputs are not allowed, but the vast majority are required to be executed. Only unneeded computations can be pruned (i.e. errors in the contract as dead code).

coprocessor/fhevm-engine/fhevm-engine-common/src/chain_id.rs
```diff
@@ -50,7 +50,7 @@ impl TryFrom<u64> for ChainId {
     type Error = InvalidChainId;
 
     fn try_from(value: u64) -> Result<Self, Self::Error> {
-        if value <= i64::MAX as u64 {
+        if i64::try_from(value).is_ok() {
```
Eikix, line 53: this is a boyscout commit otherwise clippy is not happy

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -50,7 +68,8 @@ pub type ChainCache = RwLock<lru::LruCache<Handle, ChainHash>>;
 pub type OrderedChains = Vec<Chain>;
 
 const MINIMUM_BUCKET_CACHE_SIZE: u16 = 16;
-
+const SLOW_LANE_RESET_ADVISORY_LOCK_KEY: i64 = 1_907_001;
```
Eikix: What is this? Why such a magic number?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -68,7 +68,7 @@ pub type ChainCache = RwLock<lru::LruCache<Handle, ChainHash>>;
 pub type OrderedChains = Vec<Chain>;
 
 const MINIMUM_BUCKET_CACHE_SIZE: u16 = 16;
-const SLOW_LANE_RESET_ADVISORY_LOCK_KEY: i64 = 1_907_001;
+const SLOW_LANE_RESET_ADVISORY_LOCK_KEY_BASE: i64 = 1_907_000_000;
```
antoniupop, line 71: Is this meaningful? Should this use a hash instead? Not sure what best practice would be here.

antoniupop (approved):

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
Eikix: I agree the current form can look like a magic number, but 
We namespace this lock by `chain_id` today (base + chain_id), so it’s deterministic per network and avoids cross-chain collisions for this code path. I'm okay to move forward with this for now

Eikix:
@mergify queue

PR #1926, opened 2026-02-06
chore(common): fhevm-devops codeowner of dockerfiles

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/1019

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,7 +3,12 @@
 # All pull request should be reviewed by at least one of the members of fhevm-devs
 *    @zama-ai/fhevm-devs
 
-/.github/workflows/	@zama-ai/fhevm-devops @zama-ai/fhevm-devs
+# Production Dockerfiles should be reviewed by members of fhevm-devops
+**/Dockerfile @zama-ai/fhevm-devops
+
+# Test tooling Dockerfiles can be reviewed by members of fhevm-devs
+coprocessor/fhevm-engine/stress-test-generator/Dockerfile @zama-ai/fhevm-devs
+test-suite/gateway-stress/Dockerfile @zama-ai/fhevm-devs
 
 # Gateway Team ownership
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -3,7 +3,12 @@
 # All pull request should be reviewed by at least one of the members of fhevm-devs
 *    @zama-ai/fhevm-devs
 
-/.github/workflows/	@zama-ai/fhevm-devops @zama-ai/fhevm-devs
+# Production Dockerfiles should be reviewed by members of fhevm-devops
+**/Dockerfile @zama-ai/fhevm-devops
+
+# Test tooling Dockerfiles can be reviewed by members of fhevm-devs
+coprocessor/fhevm-engine/stress-test-generator/Dockerfile
```
aquint-zama: I guess you need to put them first as iirc it's a first match strategy
btw groups `@zama-ai/fhevm-devs` is missing no?

aquint-zama (changes requested):

.github/CODEOWNERS
eudelins-zama: I don't think it's a first match strategy, because the first line of the file is:
`*    @zama-ai/fhevm-devs`

> btw groups @zama-ai/fhevm-devs is missing no?

Good catch, it's fixed :slightly_smiling_face:

eudelins-zama: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

```
# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
*.js    @js-owner #This is an inline comment.
```

aquint-zama (approved):
indeed it's last matching pattern that win

eudelins-zama:
@mergifyio queue

PR #1925, opened 2026-02-06
chore(kms-connector): time dep upgrade

eudelins-zama:
Fix https://rustsec.org/advisories/RUSTSEC-2026-0009.html

PR #1928, opened 2026-02-06
fix(coprocessor): make ciphertext compression fallible

Eikix:
## Summary
This is the standalone fix for `zama-ai/fhevm-internal#1020`.

- Make `SupportedFheCiphertexts::compress()` fallible and remove panic/expect usage.
- Convert TFHE compression failures (including panic payloads from internal asserts) into typed `FhevmError` values.
- Bubble compression errors at caller sites instead of killing execution:
  - scheduler compression paths
  - tfhe-worker server compression paths
  - zkproof-worker handle creation path
- Add a focused unit test for scalar compression error behavior.

## Scope
Compression error propagation only.
No gRPC-refactor changes, no unrelated cleanup.

## Validation
- `cd coprocessor/fhevm-engine && SQLX_OFFLINE=true cargo check -p fhevm-engine-common -p scheduler -p zkproof-worker -p tfhe-worker`
- `cd coprocessor/fhevm-engine && SQLX_OFFLINE=true cargo test -p fhevm-engine-common -p scheduler -p zkproof-worker -p tfhe-worker --no-run`
- `cd coprocessor/fhevm-engine && SQLX_OFFLINE=true cargo test -p fhevm-engine-common compress_scalar_returns_error -- --nocapture`

closes https://github.com/zama-ai/fhevm-internal/issues/1020

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
@@ -355,25 +355,6 @@ fn panic_payload_to_string(payload: Box<dyn std::any::Any + Send>) -> String {
     }
 }
 
-fn build_compressed_ciphertext_list(
-    builder: CompressedCiphertextListBuilder,
-) -> std::result::Result<tfhe::CompressedCiphertextList, FhevmError> {
-    // `builder.build()` may panic because TFHE uses internal `assert!` in this path.
-    // Catch it to avoid crashing worker tasks and convert to typed domain errors.
-    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| builder.build())) {
-        Ok(Ok(list)) => Ok(list),
-        Ok(Err(error)) => Err(FhevmError::CiphertextCompressionError(error)),
-        Err(panic_payload) => {
-            let message = panic_payload_to_string(panic_payload);
-            if message == "Ciphertexts must have empty carries to be compressed" {
-                return Err(FhevmError::CiphertextCompressionRequiresEmptyCarries);
-            }
-
-            Err(FhevmError::CiphertextCompressionPanic { message })
-        }
-    }
-}
-
 #[derive(Clone)]
 pub enum SupportedFheCiphertexts {
     FheBool(tfhe::FheBool),
@@ -597,7 +578,19 @@ impl SupportedFheCiphertexts {
         if !self.clone().to_ciphertext64().block_carries_are_empty() {
             return Err(FhevmError::CiphertextCompressionRequiresEmptyCarries);
         }
-        let list = build_compressed_ciphertext_list(builder)?;
+        let list = match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| builder.build()))
+        {
+            Ok(Ok(list)) => list,
+            Ok(Err(error)) => return Err(FhevmError::CiphertextCompressionError(error)),
+            Err(panic_payload) => {
+                let message = panic_payload_to_string(panic_payload);
+                if message == "Ciphertexts must have empty carries to be compressed" {
+                    return Err(FhevmError::CiphertextCompressionRequiresEmptyCarries);
+                }
+
+                return Err(FhevmError::CiphertextCompressionPanic { message });
+            }
+        };
         Ok((type_num, safe_serialize(&list)))
     }
 
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
@@ -575,9 +575,6 @@ impl SupportedFheCiphertexts {
             SupportedFheCiphertexts::FheBytes128(c) => builder.push(c.clone()),
             SupportedFheCiphertexts::FheBytes256(c) => builder.push(c.clone()),
         };
-        if !self.clone().to_ciphertext64().block_carries_are_empty() {
-            return Err(FhevmError::CiphertextCompressionRequiresEmptyCarries);
-        }
         let list = match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| builder.build()))
         {
             Ok(Ok(list)) => list,
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
@@ -555,8 +555,7 @@ impl SupportedFheCiphertexts {
         }
     }
 
-    pub fn compress(&self) -> std::result::Result<(i16, Vec<u8>), FhevmError> {
-        let type_num = self.type_num();
+    pub fn compress(&self) -> std::result::Result<Vec<u8>, FhevmError> {
         let mut builder = CompressedCiphertextListBuilder::new();
         match self {
             SupportedFheCiphertexts::Scalar(_) => {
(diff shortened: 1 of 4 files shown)


coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -539,12 +563,20 @@ impl SupportedFheCiphertexts {
...
             SupportedFheCiphertexts::FheBytes256(c) => builder.push(c.clone()),
             SupportedFheCiphertexts::Scalar(_) => {
-                // TODO: Need to fix that, scalars are not ciphertexts.
-                panic!("cannot compress a scalar");
+                return Err(FhevmError::CannotCompressScalar);
             }
         };
-        let list = builder.build().expect("ciphertext compression");
-        (type_num, safe_serialize(&list))
+        let list = match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| builder.build()))
```
Eikix, line 577: is this the best we can do for prod code?
it looks a bit "rough"

Eikix: Fair point. I pushed a cleanup in 8bd7e04c: `compress()` now does an explicit carry precheck (`block_carries_are_empty`) and returns a typed error early. We keep `catch_unwind` only as a final guard against upstream TFHE `assert!` panics so this path never kills worker execution if a new assert appears in the future.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -322,6 +343,33 @@ impl std::fmt::Display for FhevmError {
     }
 }
 
+fn panic_payload_to_string(payload: Box<dyn std::any::Any + Send>) -> String {
```
Eikix, line 348: can you explai nthis? 
perhaps a short docomment

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -538,13 +594,10 @@ impl SupportedFheCiphertexts {
             SupportedFheCiphertexts::FheBytes64(c) => builder.push(c.clone()),
             SupportedFheCiphertexts::FheBytes128(c) => builder.push(c.clone()),
             SupportedFheCiphertexts::FheBytes256(c) => builder.push(c.clone()),
-            SupportedFheCiphertexts::Scalar(_) => {
-                // TODO: Need to fix that, scalars are not ciphertexts.
-                panic!("cannot compress a scalar");
-            }
+            SupportedFheCiphertexts::Scalar(_) => unreachable!("checked above"),
```
Eikix: why not return an error here then instead of above for scalar?

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
Eikix: Done in `b8cc514f`.

I added short comments above:
- `panic_payload_to_string` (why we normalize panic payloads)
- `build_compressed_ciphertext_list` (why we use `catch_unwind` here)

The goal is to make this guard path readable for reviewers/operators without changing behavior.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
Eikix: Yep, agreed. Updated in `b8cc514f`.

I moved scalar handling into the `match` itself:
- `Scalar(_) => return Err(FhevmError::CannotCompressScalar)`

So we don’t rely on `unreachable!("checked above")` anymore.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -322,6 +343,37 @@ impl std::fmt::Display for FhevmError {
...
+    if let Some(message) = payload.downcast_ref::<&str>() {
+        (*message).to_string()
+    } else if let Some(message) = payload.downcast_ref::<String>() {
+        message.clone()
+    } else {
+        "unknown panic payload".to_string()
+    }
+}
+
+fn build_compressed_ciphertext_list(
```
Eikix: why do we need this wrapper?

Eikix: Good call. I removed the wrapper in `a874cee6` and inlined the `catch_unwind` -> `FhevmError` mapping directly in `compress()`, so there is one less indirection.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -538,13 +574,24 @@ impl SupportedFheCiphertexts {
             SupportedFheCiphertexts::FheBytes64(c) => builder.push(c.clone()),
             SupportedFheCiphertexts::FheBytes128(c) => builder.push(c.clone()),
             SupportedFheCiphertexts::FheBytes256(c) => builder.push(c.clone()),
-            SupportedFheCiphertexts::Scalar(_) => {
-                // TODO: Need to fix that, scalars are not ciphertexts.
-                panic!("cannot compress a scalar");
+        };
+        if !self.clone().to_ciphertext64().block_carries_are_empty() {
```
rudy-6-4: could have detrimental effect on gpu (force a move to cpu) ? why not catching the panic ? since it won't happens often it's better to not check in advance.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -522,10 +555,13 @@ impl SupportedFheCiphertexts {
         }
     }
 
-    pub fn compress(&self) -> (i16, Vec<u8>) {
+    pub fn compress(&self) -> std::result::Result<(i16, Vec<u8>), FhevmError> {
```
rudy-6-4: if not use in different places, could return only the Vec and not the i16 which is part of self.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
Eikix: good catch

coprocessor/fhevm-engine/Cargo.lock
```diff
@@ -2378,9 +2378,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
```
rudy-6-4, line 2379: Doesn't look necessary for the task. Make another in another PR if you want to bump it? An if required on main better make it separate PR every one has it faster.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
Eikix: Good catch, applied in `2cd9101e`.

`compress()` now returns only `Vec<u8>`, and callsites read the ciphertext type from `self.type_num()` where needed.

coprocessor/fhevm-engine/Cargo.lock
Eikix: i thought id be able to rebase on Link: https://github.com/zama-ai/fhevm/pull/1925 but its only for kms
you want me to open something like this for us? or its ok?

Eikix: https://github.com/zama-ai/fhevm/pull/1938

Eikix:
normally ready for re review

rudy-6-4 (approved):

Eikix:
@mergify queue

PR #1936, opened 2026-02-06
feat(copro): PoC implementation 

goshawk-3:
fixes: https://github.com/zama-ai/fhevm-internal/issues/932


coprocessor/fhevm-engine/tfhe-compute-node/src/tfhe_compute.rs
```diff
@@ -0,0 +1,246 @@
...
+            None => {
+                // No job received, continue to next cycle
+                continue;
+            }
+        };
+
+        info!(delivery = ?delivery, is_local, "Received FHE task msg");
+
+        // Spawn blocking task to execute the partition
+        if let Err(e) = spawn_partition_execution(ctx.clone(), &mut set, delivery).await {
```
antoniupop: For now we're of course targeting the current tfhe-rs API, but we will get a new DFG/DAG interface in Q2 so we will be moving the equivalent of the current scheduler down into tfhe-rs.

All worker-level parallelism (CPU + GPU/HPU) will be delegated.

antoniupop:
I haven't seen the dispatcher (not sure if missing commit or I just didn't see it?) - this is the most important/urgent component as we need to remove the HL part so we can integrate the new generic listener instead.

PR #1942, opened 2026-02-09
feat(coprocessor): standardize OTLP init + shutdown guard

Eikix:
## Summary
- unify coprocessor OTEL init and shutdown lifecycle in `fhevm-engine-common::telemetry`
- centralize `sns-worker` subscriber + optional OTEL layer wiring in common telemetry
- keep behavior minimal: global tracer provider only, no extra propagator config

## Validation
- pre-commit passed (`cargo fmt`, `cargo check`, `clippy`)
- targeted check passed:
  `SQLX_OFFLINE=true cargo check -p fhevm-engine-common -p sns-worker -p gw-listener -p host-listener -p tfhe-worker -p transaction-sender -p zkproof-worker`
- telemetry unit tests pass in `fhevm-engine-common`

## Scope
- no gRPC migration
- no operation-level span migration

Closes https://github.com/zama-ai/fhevm-internal/issues/1006

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
@@ -135,8 +135,6 @@ pub fn init_otel_tracing(
 
     let (tracer, trace_provider) = build_tracer_provider(service_name, tracer_name)?;
     install_global_tracer_provider(trace_provider.clone());
-    // TODO(fhevm-internal#935): remove explicit tracer return once all coprocessor
-    // instrumentation is fully migrated to tracing spans and layer-only wiring.
     Ok(Some((tracer, OtelGuard::new(trace_provider))))
 }
 
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
@@ -66,27 +66,19 @@ pub(crate) static ZKPROOF_TXN_LATENCY_HISTOGRAM: LazyLock<Histogram> = LazyLock:
     )
 });
 
+const DEFAULT_OTEL_TRACER_NAME: &str = "otlp-layer";
+
 pub fn init_otel(
     service_name: &str,
 ) -> Result<Option<TracerProviderGuard>, Box<dyn std::error::Error + Send + Sync + 'static>> {
-    let runtime = setup_otel(service_name, "otlp-layer")?;
+    let runtime = setup_otel(service_name, DEFAULT_OTEL_TRACER_NAME)?;
     Ok(runtime.map(|(_tracer, guard)| guard))
 }
 
 pub fn init_json_subscriber(log_level: tracing::Level) {
-    tracing_subscriber::registry()
-        .with(tracing_subscriber::filter::LevelFilter::from_level(
-            log_level,
-        ))
-        .with(
-            tracing_subscriber::fmt::layer()
-                .json()
-                .with_target(false)
-                .with_current_span(true)
-                .with_span_list(false)
-                .with_level(true),
-        )
-        .init();
+    // Reuse the OTEL-aware setup path with OTEL disabled.
+    init_json_subscriber_with_otel(log_level, "", DEFAULT_OTEL_TRACER_NAME)
+        .expect("subscriber init without OTEL should not fail");
 }
 
 pub fn init_json_subscriber_with_otel(
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
@@ -108,9 +108,8 @@ fn setup_otel_guard(
     service_name: &str,
     tracer_name: &'static str,
 ) -> Result<TracerProviderGuard, Box<dyn std::error::Error + Send + Sync + 'static>> {
-    let (_tracer, trace_provider) = build_tracer_provider(service_name, tracer_name)?;
-    install_global_tracer_provider(trace_provider.clone());
-    Ok(TracerProviderGuard::new(trace_provider))
+    let (_tracer, guard) = setup_otel_with_tracer(service_name, tracer_name)?;
+    Ok(guard)
 }
 
 fn setup_otel_with_tracer(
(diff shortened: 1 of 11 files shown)


coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
```diff
@@ -20,6 +25,37 @@ use tracing::{debug, info, warn};
...
+
+    fn shutdown_if_needed(&mut self) {
+        if let Some(provider) = self.provider.take() {
+            if let Err(err) = provider.shutdown() {
+                warn!(error = %err, "Failed to shutdown OTLP tracer provider");
+            }
+        }
+    }
+
+    pub fn shutdown(mut self) {
```
Eikix: why do we need shutdown if the Drop trait impl uses shutdown_if_needed

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
```diff
@@ -20,6 +25,37 @@ use tracing::{debug, info, warn};
...
+}
+
+impl OtlpShutdownGuard {
+    fn new(provider: SdkTracerProvider) -> Self {
+        Self {
+            provider: Some(provider),
+        }
+    }
+
+    fn shutdown_if_needed(&mut self) {
```
Eikix: why is it named like that?

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
```diff
@@ -41,14 +77,47 @@ pub(crate) static ZKPROOF_TXN_LATENCY_HISTOGRAM: LazyLock<Histogram> = LazyLock:
 pub fn setup_otlp(
     service_name: &str,
 ) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
-    let _ = setup_otlp_tracer(service_name, "otlp-layer")?;
+    let (_tracer, trace_provider) = setup_otlp_tracer_impl(service_name, "otlp-layer")?;
```
Eikix: why are we naming a function with _impl suffix?

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
```diff
@@ -41,14 +77,47 @@ pub(crate) static ZKPROOF_TXN_LATENCY_HISTOGRAM: LazyLock<Histogram> = LazyLock:
...
+    set_provider_and_propagator(trace_provider);
```
Eikix: why is this an extra function

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
```diff
@@ -66,9 +135,16 @@ pub fn setup_otlp_tracer(
         .build();
 
     let tracer = trace_provider.tracer(tracer_name);
-    opentelemetry::global::set_tracer_provider(trace_provider);
 
-    Ok(tracer)
+    Ok((tracer, trace_provider))
+}
+
+fn set_provider_and_propagator(trace_provider: SdkTracerProvider) {
```
Eikix: can you explain why we need this?

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
Eikix: agree. renamed it to `shutdown_once` in 44d9824c; previous name was clunky.

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
Eikix: yep good catch. we do not need explicit `shutdown(self)` if `Drop` already handles it. removed that public method and kept a single drop path in 44d9824c.

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
Eikix: kept it but renamed to `install_global_otel`. reason: provider + propagator should always be installed together across all setup variants, so this avoids drift.

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
Eikix: agree as well. renamed `setup_otlp_tracer_impl` to `build_otlp_tracer_and_provider` in 44d9824c.

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
Eikix: this installs both global tracer provider and composite propagator (`tracecontext` + `baggage`). without this step, spans can export but parent context propagation across services becomes inconsistent.

Eikix:
@claude /pr-review
is a drop guard good practice or outdated?

Eikix:
@claude /pr-review

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
```diff
@@ -66,9 +146,11 @@ pub fn setup_otlp_tracer(
         .build();
 
     let tracer = trace_provider.tracer(tracer_name);
-    opentelemetry::global::set_tracer_provider(trace_provider);
+    Ok((tracer, trace_provider))
+}
 
-    Ok(tracer)
+fn install_global_tracer_provider(trace_provider: SdkTracerProvider) {
```
obatirou: You could inline

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
```diff
@@ -38,17 +66,69 @@ pub(crate) static ZKPROOF_TXN_LATENCY_HISTOGRAM: LazyLock<Histogram> = LazyLock:
...
+    };
+
+    let telemetry_layer = tracing_opentelemetry::layer().with_tracer(tracer);
+    base.with(telemetry_layer).try_init()?;
+    install_global_tracer_provider(trace_provider.clone());
+    Ok(Some(TracerProviderGuard::new(trace_provider)))
 }
 
-pub fn setup_otlp_tracer(
+fn setup_otel_guard(
```
obatirou: could be inlined also

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
Eikix: good call, inlined in 60f4a1f9 to reduce indirection.

coprocessor/fhevm-engine/fhevm-engine-common/src/telemetry.rs
Eikix: yep agreed, removed this wrapper and inlined the logic in 60f4a1f9.

obatirou (approved):

Eikix:
@mergify queue

PR #1945, opened 2026-02-09
fix(coprocessor): host-listener, trust tx & logs order from get_logs

rudy-6-4:


antoniupop (approved):
LGTM - we may want to add a warning that this is only guaranteed to work with host chains that offer total order in getLogs.

PR #1950, opened 2026-02-09
fix(coprocessor): maximise dependence fetch from DB 

antoniupop:


antoniupop:
@claude /pr-review

antoniupop:
@claude /pr-review

Eikix:
local review shows:
`The new deferred-dependency path has no direct tests in this PR. Adding coverage for “handle fetched => edge dropped” and “handle missing => edge restored + cycle handling” would reduce regression risk.`

obatirou (approved):
Lgtm 
My understanding: it introduces a 2 step edge insertion for cross-tx dependencies. 
* when building the graph: defer the cross-tx dependence after db fetch in consumer (tfhe-worker).
* when resolving dependence in consumer: 
    * keep only handles that are not part of fetched ones from DB and add them to a copy of the graph with cycles allowed (digraph)
    * run TarjanScc to see if there are cycles
    * only commit to self.graph once it is confirmed that there were no cycle. 

If resolve_dependences throw an error, nothing is executed as graph cannot be executed

antoniupop:
@mergifyio queue

PR #1953, opened 2026-02-10
feat(coprocessor): do not retry on invalid signer/sender reversal from the GW

obatirou:
Closes zama-ai/fhevm-internal#835

Depending on https://github.com/zama-ai/fhevm/pull/1856

### DIFF
--- a/coprocessor/fhevm-engine/transaction-sender/contracts/CiphertextCommits.sol
+++ b/coprocessor/fhevm-engine/transaction-sender/contracts/CiphertextCommits.sol
@@ -5,9 +5,7 @@ pragma solidity ^0.8.24;
 /// source: github.com/zama-ai/fhevm-gateway/blob/main/contracts/CiphertextCommits.sol
 contract CiphertextCommits {
     error CoprocessorAlreadyAdded(bytes32 ctHandle, address coprocessorTxSenderAddress);
-    error NotCoprocessorSigner(address signerAddress);
     error NotCoprocessorTxSender(address txSenderAddress);
-    error CoprocessorSignerDoesNotMatchTxSender(address signerAddress, address txSenderAddress);
 
     event AddCiphertextMaterial(
         bytes32 indexed ctHandle,
@@ -21,9 +19,7 @@ contract CiphertextCommits {
 
     enum ConfigErrorMode {
         None,
-        NotCoprocessorSigner,
-        NotCoprocessorTxSender,
-        CoprocessorSignerDoesNotMatchTxSender
+        NotCoprocessorTxSender
     }
 
     constructor(bool _alreadyAddedRevert) {
@@ -31,19 +27,13 @@ contract CiphertextCommits {
     }
 
     function setConfigErrorMode(uint8 mode) external {
-        require(mode <= uint8(ConfigErrorMode.CoprocessorSignerDoesNotMatchTxSender), "invalid mode");
+        require(mode <= uint8(ConfigErrorMode.NotCoprocessorTxSender), "invalid mode");
         configErrorMode = ConfigErrorMode(mode);
     }
 
     function maybeRevertConfigError() internal view {
-        if (configErrorMode == ConfigErrorMode.NotCoprocessorSigner) {
-            revert NotCoprocessorSigner(msg.sender);
-        }
         if (configErrorMode == ConfigErrorMode.NotCoprocessorTxSender) {
             revert NotCoprocessorTxSender(msg.sender);
-        }
-        if (configErrorMode == ConfigErrorMode.CoprocessorSignerDoesNotMatchTxSender) {
-            revert CoprocessorSignerDoesNotMatchTxSender(address(0x1234), msg.sender);
         }
     }
 
--- a/coprocessor/fhevm-engine/transaction-sender/contracts/CiphertextCommits.sol
+++ b/coprocessor/fhevm-engine/transaction-sender/contracts/CiphertextCommits.sol
@@ -5,6 +5,9 @@ pragma solidity ^0.8.24;
 /// source: github.com/zama-ai/fhevm-gateway/blob/main/contracts/CiphertextCommits.sol
 contract CiphertextCommits {
     error CoprocessorAlreadyAdded(bytes32 ctHandle, address coprocessorTxSenderAddress);
+    error NotCoprocessorSigner(address signerAddress);
+    error NotCoprocessorTxSender(address txSenderAddress);
+    error CoprocessorSignerDoesNotMatchTxSender(address signerAddress, address txSenderAddress);
 
     event AddCiphertextMaterial(
         bytes32 indexed ctHandle,
@@ -14,9 +17,34 @@ contract CiphertextCommits {
     );
 
     bool alreadyAddedRevert;
+    ConfigErrorMode configErrorMode;
+
+    enum ConfigErrorMode {
+        None,
+        NotCoprocessorSigner,
+        NotCoprocessorTxSender,
+        CoprocessorSignerDoesNotMatchTxSender
+    }
 
     constructor(bool _alreadyAddedRevert) {
         alreadyAddedRevert = _alreadyAddedRevert;
+    }
+
+    function setConfigErrorMode(uint8 mode) external {
+        require(mode <= uint8(ConfigErrorMode.CoprocessorSignerDoesNotMatchTxSender), "invalid mode");
+        configErrorMode = ConfigErrorMode(mode);
+    }
+
+    function maybeRevertConfigError() internal view {
+        if (configErrorMode == ConfigErrorMode.NotCoprocessorSigner) {
+            revert NotCoprocessorSigner(msg.sender);
+        }
+        if (configErrorMode == ConfigErrorMode.NotCoprocessorTxSender) {
+            revert NotCoprocessorTxSender(msg.sender);
+        }
+        if (configErrorMode == ConfigErrorMode.CoprocessorSignerDoesNotMatchTxSender) {
+            revert CoprocessorSignerDoesNotMatchTxSender(address(0x1234), msg.sender);
+        }
     }
 
     function addCiphertextMaterial(
@@ -25,6 +53,7 @@ contract CiphertextCommits {
         bytes32 ciphertextDigest,
         bytes32 snsCiphertextDigest
     ) public {
+        maybeRevertConfigError();
         if (alreadyAddedRevert) {
             revert CoprocessorAlreadyAdded(ctHandle, msg.sender);
         }
--- a/coprocessor/fhevm-engine/transaction-sender/contracts/CiphertextCommits.sol
+++ b/coprocessor/fhevm-engine/transaction-sender/contracts/CiphertextCommits.sol
@@ -31,19 +31,15 @@ contract CiphertextCommits {
         configErrorMode = ConfigErrorMode(mode);
     }
 
-    function maybeRevertConfigError() internal view {
-        if (configErrorMode == ConfigErrorMode.NotCoprocessorTxSender) {
-            revert NotCoprocessorTxSender(msg.sender);
-        }
-    }
-
     function addCiphertextMaterial(
         bytes32 ctHandle,
         uint256 /* keyId */,
         bytes32 ciphertextDigest,
         bytes32 snsCiphertextDigest
     ) public {
-        maybeRevertConfigError();
+        if (configErrorMode == ConfigErrorMode.NotCoprocessorTxSender) {
+            revert NotCoprocessorTxSender(msg.sender);
+        }
         if (alreadyAddedRevert) {
             revert CoprocessorAlreadyAdded(ctHandle, msg.sender);
         }
--- a/coprocessor/fhevm-engine/transaction-sender/src/ops/allow_handle.rs
+++ b/coprocessor/fhevm-engine/transaction-sender/src/ops/allow_handle.rs
@@ -126,7 +126,7 @@ where
                         key = %key,
                         "Detected non-retryable gateway coprocessor config error while allowing handle"
                     );
-                    self.mark_allow_handle_terminal_config_error(
+                    self.stop_retrying_allow_handle_on_config_error(
                         key,
                         &terminal_config_error.to_string(),
                     )
@@ -338,7 +338,7 @@ where
         Ok(())
     }
 
-    async fn mark_allow_handle_terminal_config_error(
+    async fn stop_retrying_allow_handle_on_config_error(
         &self,
         key: &Key,
         error: &str,
--- a/coprocessor/fhevm-engine/transaction-sender/src/ops/allow_handle.rs
+++ b/coprocessor/fhevm-engine/transaction-sender/src/ops/allow_handle.rs
@@ -121,7 +121,7 @@ where
                 }
                 if let Some(terminal_config_error) = try_extract_terminal_config_error(&e) {
                     ALLOW_HANDLE_FAIL_COUNTER.inc();
-                    error!(
+                    warn!(
                         error = %terminal_config_error,
                         key = %key,
                         "Detected non-retryable gateway coprocessor config error while allowing handle"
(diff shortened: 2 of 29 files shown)


obatirou:
@claude /pr-review all findings are addressed
regarding, config-error detection from trace is root-frame only, it is expected as the tx revert and we only check root callFrame due standard proxy in front of Gateway contracts

Eikix:
All 4 ops repeat “if config error => mark terminal” + transient handling branches.
can you centralize error classification into one helper and keep per-op DB writes local?

coprocessor/fhevm-engine/transaction-sender/contracts/CiphertextCommits.sol
```diff
@@ -5,6 +5,9 @@ pragma solidity ^0.8.24;
 /// source: github.com/zama-ai/fhevm-gateway/blob/main/contracts/CiphertextCommits.sol
 contract CiphertextCommits {
     error CoprocessorAlreadyAdded(bytes32 ctHandle, address coprocessorTxSenderAddress);
+    error NotCoprocessorSigner(address signerAddress);
```
Eikix: why do we need contract changes? (applicable to all .sol modifs)

obatirou:
> All 4 ops repeat “if config error => mark terminal” + transient handling branches. can you centralize error classification into one helper and keep per-op DB writes local?

I would prefer do this change in a refactoring PR to avoid mixing refactor that would impact non feature related code + new feature

coprocessor/fhevm-engine/transaction-sender/contracts/CiphertextCommits.sol
obatirou: Mocks contracts are changed to introduce the error that would be returned by the prod gateway contracts. It allows to validate the behavior in integration tests (like `add_ciphertext_terminal_on_gw_config_error`)

coprocessor/fhevm-engine/transaction-sender/src/ops/allow_handle.rs
```diff
@@ -324,6 +361,34 @@ where
         .await?;
         Ok(())
     }
+
+    async fn mark_allow_handle_terminal_config_error(
+        &self,
+        key: &Key,
+        error: &str,
+    ) -> anyhow::Result<()> {
+        let res = sqlx::query(
```
Eikix: afai understand, the codebase convention is to use `query!`

Eikix:
There is a liveness/timing issue we should make explicit:

If a coprocessor tx sender is not yet registered on Gateway, and we terminalize/stop retries on config errors, we can end up in a bad state where the sender never resumes automatically once it *does* get registered. So even with this PR, “stop retry” can be unsafe unless we define a clean re-entry path.

A smaller alternative (likely much smaller diff) could be:

1. Add `send_enabled` in tx-sender.
2. While `send_enabled=false`, do only dry-run checks (`eth_call`), no on-chain sends.
3. Listen to `GatewayConfig.UpdateCoprocessors(...)`.
4. If this node’s `txSenderAddress` is present in the updated coprocessor set, enable sending (ideally after a few confirmations).
5. Before first real send, run one final `eth_call` guard, then proceed.

Failsafe if the event is missed:
- Keep a periodic reconciliation loop (and run it on startup) that checks on-chain config directly (`isCoprocessorTxSender(...)` / equivalent getter).
- If on-chain state says we are registered, enable sending even if no event was observed.
- Event listener remains the fast path; polling/reconciliation is the safety net.

This gives us:
- safer bootstrap timing,
- deterministic activation when Gateway config is ready,
- no permanent “stuck terminal” rows after config is fixed,
- and likely a much smaller initial PR (then we can add mined-revert/trace logic later if still needed).

obatirou:
> There is a liveness/timing issue we should make explicit:
> 
> If a coprocessor tx sender is not yet registered on Gateway, and we terminalize/stop retries on config errors, we can end up in a bad state where the sender never resumes automatically once it _does_ get registered. So even with this PR, “stop retry” can be unsafe unless we define a clean re-entry path.
> 
> A smaller alternative (likely much smaller diff) could be:
> 
> 1. Add `send_enabled` in tx-sender.
> 2. While `send_enabled=false`, do only dry-run checks (`eth_call`), no on-chain sends.
> 3. Listen to `GatewayConfig.UpdateCoprocessors(...)`.
> 4. If this node’s `txSenderAddress` is present in the updated coprocessor set, enable sending (ideally after a few confirmations).
> 5. Before first real send, run one final `eth_call` guard, then proceed.
> 
> Failsafe if the event is missed:
> 
> * Keep a periodic reconciliation loop (and run it on startup) that checks on-chain config directly (`isCoprocessorTxSender(...)` / equivalent getter).
> * If on-chain state says we are registered, enable sending even if no event was observed.
> * Event listener remains the fast path; polling/reconciliation is the safety net.
> 
> This gives us:
> 
> * safer bootstrap timing,
> * deterministic activation when Gateway config is ready,
> * no permanent “stuck terminal” rows after config is fixed,
> * and likely a much smaller initial PR (then we can add mined-revert/trace logic later if still needed).

We keep the implementation as simulation is on and sufficient
* No new code/behavior
* No periodic check with RPC calls to handle
* No test refactoring: if loop introduced at startup all tests need to be updated to introduce a gateway config and ensure the coproc is registered 

Also note that row marked as non retryable can be ignored as it means the gateway can reach consensus without them.

coprocessor/fhevm-engine/transaction-sender/src/ops/common.rs
```diff
@@ -13,3 +19,35 @@ pub(crate) fn try_into_array<const SIZE: usize>(vec: Vec<u8>) -> Result<[u8; SIZ
...
+}
+
+pub(crate) fn try_extract_terminal_config_error(
+    err: &RpcError<TransportErrorKind>,
+) -> Option<CoprocessorConfigError> {
+    err.as_error_resp()
+        .and_then(|payload| payload.as_decoded_interface_error::<GatewayConfigChecksErrors>())
+        .and_then(|decoded| match decoded {
+            GatewayConfigChecksErrors::NotCoprocessorSigner(inner) => Some(
+                CoprocessorConfigError::NotCoprocessorSigner(inner.signerAddress),
```
Eikix, line 64: why are there three different gateway errors? probably there's three different scenarii, but it's not super clear what they are from reading the errors

coprocessor/fhevm-engine/transaction-sender/src/ops/common.rs
```diff
@@ -13,3 +19,35 @@ pub(crate) fn try_into_array<const SIZE: usize>(vec: Vec<u8>) -> Result<[u8; SIZ
...
+pub(crate) enum CoprocessorConfigError {
+    #[error("NotCoprocessorSigner({0})")]
+    NotCoprocessorSigner(Address),
+    #[error("NotCoprocessorTxSender({0})")]
+    NotCoprocessorTxSender(Address),
+    #[error("CoprocessorSignerDoesNotMatchTxSender({signer},{tx_sender})")]
+    CoprocessorSignerDoesNotMatchTxSender { signer: Address, tx_sender: Address },
+}
+
+pub(crate) fn try_extract_terminal_config_error(
```
Eikix: could be nice to have doccomment on that function specifically to explain the inner workings

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
```diff
@@ -147,15 +149,23 @@ impl<P: Provider<Ethereum> + Clone + 'static> DelegateUserDecryptOperation<P> {
...
+                if is_transient_error(&error) {
+                    warn!(
+                        %error,
+                        ?delegation,
+                        "{operation} sending with transient error. Will retry indefinitely"
+                    );
+                    return TxResult::TransientError;
+                }
+                if let Some(terminal_config_error) = try_extract_terminal_config_error(&error) {
+                    error!(
```
Eikix: does it count as an error semantically or a warn?

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
```diff
@@ -587,6 +612,47 @@ pub async fn update_error_delegation(
...
+    tx: &mut DbTransaction<'_>,
+    delegation: &DelegationRow,
+    error: &str,
+    terminal_attempts: u64,
+) {
+    error!(
+        %error,
+        ?delegation,
+        terminal_attempts,
+        "Updating delegation with terminal gateway config error"
```
Eikix: maybe it's just me but I don't understand this sentence "Updating delegation with terminal gateway config error" 

what is `terminal_attempts` semantically?

coprocessor/fhevm-engine/transaction-sender/src/ops/common.rs
obatirou: Those are the checks done on the different gateway contracts: 
For example: 
https://github.com/zama-ai/fhevm/blob/ad06d432e30bd5e4265029286fe3c0549edb1b7b/gateway-contracts/contracts/InputVerification.sol#L213 with only `onlyCoprocessorTxSender`

And one case for `_checkCoprocessorSignerMatchesTxSender`
* https://github.com/zama-ai/fhevm/blob/ad06d432e30bd5e4265029286fe3c0549edb1b7b/gateway-contracts/contracts/shared/GatewayConfigChecks.sol#L154
* https://github.com/zama-ai/fhevm/blob/ad06d432e30bd5e4265029286fe3c0549edb1b7b/gateway-contracts/contracts/InputVerification.sol#L244-L246

Will fix this by adding documentation to mocks and the extract method

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
obatirou: an error (non retryable)

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
obatirou: it is to set to the maximum of retry `gateway_nb_attempts` to avoid retries on this row.

on call site
```
 mark_delegation_terminal_config_error(
                        &mut tx,
                        &delegation,
                        &e.to_string(),
                        self.conf.delegation_max_retry + 1,
                    )
```

coprocessor/fhevm-engine/transaction-sender/src/ops/verify_proof.rs
```diff
@@ -214,6 +228,28 @@ where
         }
         Ok(())
     }
+
+    async fn mark_verify_proof_terminal_config_error(
```
Eikix: I feel as though the code is hard to understand at first glance
If this sets retry_count to max, why is it called `mark_verify_proof_terminal_config_error` and not `set_verify_proof_retry_count_to_max`? 

Same for other similar utils?

coprocessor/fhevm-engine/transaction-sender/tests/add_ciphertext_tests.rs
```diff
@@ -610,3 +610,136 @@ async fn retry_on_aws_kms_error(#[case] signer_type: SignerType) -> anyhow::Resu
     run_handle.await??;
     Ok(())
 }
+
+#[rstest]
+#[case::private_key(SignerType::PrivateKey)]
+#[tokio::test]
+#[serial(db)]
+async fn add_ciphertext_terminal_on_gw_config_error(
```
Eikix: Maybe it is me but I have an issue with the 'terminal' semantic. It's hard to understand the test at first glance. 
I know Naming debates are a nit but since the PR is complex in terms of scenario, it really would help to have easy names.

add_ciphertext_skipped_on_coprocessor_unauthorized or if we keep gw config generic
add_ciphertext_no_retry_on_gw_config_error

or if we want to express semantically that there's not even 1 try, we can find another

coprocessor/fhevm-engine/transaction-sender/src/ops/verify_proof.rs
obatirou: `set_verify_proof_retry_count_to_max` is an implementation detail imo?
Could you detail why it is hard to understand to improve it?
The goal here is to say what is done. But maybe the problem is the term `terminal` that is confusing?
maybe something more explicit would be better like `stop_retrying_verify_proof_on_config_error`?

obatirou: or maybe `fail_verify_proof_on_config_error`

coprocessor/fhevm-engine/transaction-sender/tests/add_ciphertext_tests.rs
obatirou: seems to match my comment above about the terminal term.
https://github.com/zama-ai/fhevm/pull/1953#discussion_r2828108416

coprocessor/fhevm-engine/transaction-sender/src/ops/verify_proof.rs
```diff
@@ -214,6 +228,28 @@ where
...
+        &self,
+        zk_proof_id: i64,
+        error: &str,
+    ) -> anyhow::Result<()> {
```
Eikix, line 238: is it common/intentional to pair anyhow and thiserror? i see that we import result from anyhow and error from thiserror?
i dont have knowledge on this

coprocessor/fhevm-engine/transaction-sender/src/ops/verify_proof.rs
Eikix: `stop_retrying_verify_proof_on_config_error` thats perfect!

obatirou: Great!
Will make the change

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
```diff
@@ -587,6 +612,47 @@ pub async fn update_error_delegation(
     }
 }
 
+pub async fn mark_delegation_terminal_config_error(
+    tx: &mut DbTransaction<'_>,
+    delegation: &DelegationRow,
+    error: &str,
+    terminal_attempts: u64,
+) {
+    error!(
```
Eikix: i wonder if we want to qualify this as an error for the logs, wdyt

Eikix:
This could be tested locally for confidence by using the local fhevm-cli with multiple coprocs (and altering slightly the fhevm-cli to make one coproc not registered) and try sending the relevant test paths 
```
./fhevm-cli test input-proof
./fhevm-cli test delegated-user-decryption
./fhevm-cli test erc20
```

coprocessor/fhevm-engine/transaction-sender/src/ops/verify_proof.rs
obatirou: It was to stay consistent with the codebase + execute.await? will send back a `sqlx::Error` and with anyhow we  do not need to define a custom error from my understranding

coprocessor/fhevm-engine/transaction-sender/src/ops/delegate_user_decrypt.rs
obatirou: I agree
after look into it more it would be more a warn as it is a known handled behavior.
it does not make sense to keep only delegate as error

obatirou:
@claude /pr-review

Eikix:
Is there a way to test this in test-suite or in the fhevm-cli? 
This would greatly increase confidence in our ability to merge it as is

Eikix (approved):
lgtm

obatirou:
> Is there a way to test this in test-suite or in the fhevm-cli? This would greatly increase confidence in our ability to merge it as is

Modified `./fhevm-cli` locally to run 2 coprocessors with one unregistered. 
We can see the warning logs being output + no retry for the 3 different tests 

```
./fhevm-cli test input-proof
./fhevm-cli test delegated-user-decryption
./fhevm-cli test erc20
```

Capture d’écran 2026-02-23 à 16 22 19
Capture d’écran 2026-02-23 à 16 21 45
Capture d’écran 2026-02-23 à 16 21 07

obatirou:
@mergify queue

PR #1971, opened 2026-02-12
feat(coprocessor): zkproof-worker optional host-chain scope for per-host scaling

obatirou:
## Summary
- Add `--host-chain-id` CLI flag so each `zkproof-worker` instance can be scoped to process proofs for a single host chain, enabling horizontal per-chain scaling.
- When unset, existing all-chain behavior is preserved (no behavioral change for current deployments).
- Unknown chain IDs are validated at startup against `HostChainsCache`, failing fast before any work begins.
- No schema or migration changes required.

Closes https://github.com/zama-ai/fhevm-internal/issues/921


obatirou:
No needed for now

PR #2002, opened 2026-02-17
ci(coprocessor): retry solc install on ETXTBSY

Eikix:
## Summary
- remove the CI-level solc preinstall hack
- add a small retry wrapper around `Solc::find_or_install` in coprocessor build scripts
- retry only when the install fails with `ETXTBSY` / "Text file busy" and keep other failures immediate

## Why
The flake is a concurrent install/execute race on the same solc binary path (`~/.svm/.../solc-*`).
This fix handles that race at the call site used by the failing build scripts, without relying on workflow wiring or hardcoded CI versions.

## Scope
- `coprocessor/fhevm-engine/host-listener/build.rs`
- `coprocessor/fhevm-engine/gw-listener/build.rs`
- `coprocessor/fhevm-engine/transaction-sender/build.rs`

No runtime behavior changes.

## Validation
Pre-commit checks passed locally on this branch:
- cargo fmt check
- cargo check
- clippy (pedantic)

### DIFF
--- a/.github/workflows/coprocessor-cargo-tests.yml
+++ b/.github/workflows/coprocessor-cargo-tests.yml
@@ -74,6 +74,11 @@ jobs:
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
         restore-keys: ${{ runner.os }}-cargo-
 
+    - name: Preinstall solc via host-listener build script
+      run: |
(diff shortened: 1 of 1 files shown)


Eikix:
Superseded by #2003 with a code-level ETXTBSY retry fix (no workflow hack).

PR #2004, opened 2026-02-17
feat(host-contracts): add minimal per-block HCU cap with whitelist bypass

Eikix:
## Context
Implements the minimal production feature discussed in [zama-ai/fhevm-internal#1053](https://github.com/zama-ai/fhevm-internal/issues/1053): block-level HCU throttling with whitelist bypass.

No benchmark-only artifacts in this PR.

## What is included
- `HCULimit` block-level cap for non-whitelisted callers:
  - `publicHCUPerBlock`
  - packed per-block counter
  - `HCUBlockLimitExceeded`
- Owner controls:
  - `setHCUPerBlock(uint192)`
  - `setBlockHCUWhitelist(address,bool)`
  - `setBlockHCUWhitelistBatch(address[],bool)`
- Runtime caller-context wiring from executor to HCULimit:
  - `HCULimit.setHCUCallerContext(address)`
  - `FHEVMExecutor` now sets context before HCU checks.
- Read helper:
  - `HCULimit.getBlockMeter()`
- Tests:
  - block cap exceed -> revert
  - whitelist bypass
  - whitelist removal
  - block rollover reset
  - non-executor context setter rejection

## What was removed
- `contracts/bench/HCUBlockMeterGasBench.sol`
- `tasks/hcuGasBench.ts`
- `docs/hcu_block_meter_gas.md`
- `test/encryptedERC20/EncryptedERC20.blockMeterCheckpoints.gas.ts`
- hardhat task import for benchmark

## Validation
Ran locally:
```bash
cd host-contracts
set -a; source .env.example; set +a
npx hardhat compile
npx hardhat test test/upgrades/upgrades.ts --network hardhat
npx hardhat test test/hcuLimit/HCULimit.ts --network hardhat
npx hardhat test test/encryptedERC20/EncryptedERC20.ts --network hardhat
npx hardhat test test/encryptedERC20/EncryptedERC20.gas.ts --network hardhat
```

Note: I could not run foundry tests in this environment because `forge-std` dependency paths are missing locally.


host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,6 +1404,57 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
             );
     }
 
+    /**
+     * @notice Sets the block-level HCU limit for non-whitelisted callers.
+     * @param hcuPerBlock New block-level cap.
+     */
+    function setHCUPerBlock(uint192 hcuPerBlock) external onlyACLOwner {
```
Eikix: Goal: owner-settable global per-block HCU cap for all non-whitelisted callers.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,6 +1404,57 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        publicHCUPerBlock = hcuPerBlock;
+        emit HCUPerBlockSet(hcuPerBlock);
+    }
+
+    /**
+     * @notice Sets whitelist status for one caller.
+     * @param account Caller to update.
+     * @param isWhitelisted Whether caller bypasses the block cap.
+     */
+    function setBlockHCUWhitelist(address account, bool isWhitelisted) external onlyACLOwner {
```
Eikix: Goal: single-address whitelist toggle so priority callers can bypass the block cap.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,6 +1404,57 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        blockHCUWhitelist[account] = isWhitelisted;
+        emit BlockHCUWhitelistSet(account, isWhitelisted);
+    }
+
+    /**
+     * @notice Sets whitelist status for a batch of callers.
+     * @param accounts Callers to update.
+     * @param isWhitelisted Whether callers bypass the block cap.
+     */
+    function setBlockHCUWhitelistBatch(address[] calldata accounts, bool isWhitelisted) external onlyACLOwner {
```
Eikix: Goal: batch whitelist updates to keep operational/governance overhead low.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,6 +1404,57 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+            emit BlockHCUWhitelistSet(accounts[i], isWhitelisted);
+        }
+    }
+
+    /**
+     * @notice Sets the original caller context for block-level HCU checks.
+     * @dev Must be called by FHEVMExecutor in the same transaction before checkHCUFor*.
+     * @param callerAddress Original FHEVMExecutor caller (EOA or contract).
+     */
+    function setHCUCallerContext(address callerAddress) external {
```
Eikix: Goal: capture original executor caller in transient storage for per-caller whitelist enforcement.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,6 +1404,57 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        bytes32 slot = HCU_CALLER_CONTEXT_SLOT;
+        assembly {
+            tstore(slot, callerAddress)
+        }
+    }
+
+    /**
+     * @notice Returns current block counter state.
+     */
+    function getBlockMeter() external view returns (uint64 blockNumber, uint192 usedHCU) {
```
Eikix: Goal: expose block-meter state for observability and operational debugging.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1441,13 +1514,41 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
             revert HCUTransactionLimitExceeded();
         }
         _setHCUForTransaction(transactionHCU);
     }
 
+    /**
+     * @notice Updates and verifies the HCU block limit.
+     * @param opHCU The HCU for the operation.
+     */
+    function _updateAndVerifyHCUBlockLimit(uint256 opHCU) internal virtual {
```
Eikix: Goal: enforce block-level HCU throttling before tx-level accounting unless cap is disabled or caller is whitelisted.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1494,6 +1595,24 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
         }
     }
 
+    function _packBlockMeter(uint64 blockNumber, uint192 usedHCU) internal pure virtual returns (uint256) {
```
Eikix: Goal: pack block number + used HCU into one slot to minimize persistent storage cost.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1494,6 +1595,24 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        return (uint256(usedHCU) << 64) | uint256(blockNumber);
+    }
+
+    function _unpackBlockMeter(uint256 packed) internal pure virtual returns (uint64 blockNumber, uint192 usedHCU) {
```
Eikix: Goal: decode packed block-meter state for checks and external readbacks.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1494,6 +1595,24 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        blockNumber = uint64(packed);
+        usedHCU = uint192(packed >> 64);
+    }
+
+    function _getHCUCallerContext() internal view virtual returns (address caller) {
```
Eikix: Goal: read the transient caller context written by executor in the same transaction.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -906,10 +912,15 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
         _checkBelowMaxBound(upperBound, randType);
         result = keccak256(abi.encodePacked(Operators.fheRandBounded, upperBound, randType, seed));
         result = _appendMetadataToPrehandle(result, randType);
+        _setHCUCallerContext();
         hcuLimit.checkHCUForFheRandBounded(randType, result);
         acl.allowTransient(result, msg.sender);
     }
 
+    function _setHCUCallerContext() internal virtual {
```
Eikix: Goal: propagate original msg.sender into HCULimit so whitelist checks apply to the true caller, not the executor.

Eikix:
Per-block accounting happens in `HCULimit._updateAndVerifyHCUBlockLimit(uint256 opHCU)`:

- `host-contracts/contracts/HCULimit.sol:1530`

Flow in that function:
1. Skip if block cap is disabled (`publicHCUPerBlock == type(uint192).max`) or caller is whitelisted.
2. Read packed meter (`publicBlockMeterPacked`) via `_unpackBlockMeter`.
3. If block changed, reset used HCU to `0`.
4. Add `opHCU` for current op.
5. Revert with `HCUBlockLimitExceeded` if above/equal cap.
6. Persist updated packed `(blockNumber, usedHCU)` via `_packBlockMeter`.

This is invoked on every HCU-charged operation from `_updateAndVerifyHCUTransactionLimit`:

- `host-contracts/contracts/HCULimit.sol:1516`

So every existing `checkHCUFor*` path now performs block-level accounting before tx-level accounting.

host-contracts/contracts/FHEVMExecutor.sol
Eikix: What if within a single tx, there are multiple calls to FHEVMExecutor from different contracts via sub-calls?

host-contracts/contracts/HCULimit.sol
```diff
@@ -70,7 +89,10 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
      * @notice  Initializes the contract.
      */
     /// @custom:oz-upgrades-validate-as-initializer
-    function initializeFromEmptyProxy() public virtual onlyFromEmptyProxy reinitializer(REINITIALIZER_VERSION) {}
+    function initializeFromEmptyProxy() public virtual onlyFromEmptyProxy reinitializer(REINITIALIZER_VERSION) {
```
Eikix: how come we set publicHCUPerBlock here but we don't set other constructor values?
is it best practice to set variables in this entrypoint?

host-contracts/contracts/HCULimit.sol
```diff
@@ -1441,13 +1514,41 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        }
+
+        (uint64 storedBlock, uint192 storedHCU) = _unpackBlockMeter(publicBlockMeterPacked);
+        uint64 currentBlock = uint64(block.number);
+        if (storedBlock != currentBlock) {
+            storedHCU = 0;
+        }
+
+        uint256 nextHCU = uint256(storedHCU) + opHCU;
+        if (nextHCU >= publicHCUPerBlock) {
```
Eikix: Can you compare a u256 to a u192?

host-contracts/contracts/HCULimit.sol
Eikix: Yep, for upgradeable proxies this is the right place for mutable state defaults. `publicHCUPerBlock` lives in proxy storage, so it must be initialized via `initializeFromEmptyProxy`/reinitializer. The other values here are constants/immutables in bytecode, so there is nothing to initialize in storage for them.

host-contracts/contracts/HCULimit.sol
Eikix: Yes, Solidity promotes the smaller int type to the larger one for the comparison. I still pushed an explicit cast in `15736b40` (`nextHCU >= uint256(publicHCUPerBlock)`) to make it unambiguous/readable.

host-contracts/contracts/FHEVMExecutor.sol
Eikix: Good call. We set caller context before every HCU check path, so if a tx does multiple sub-calls through different contracts, each call overwrites the transient slot right before its own accounting. The whitelist check always sees the current sub-call sender (`msg.sender` at executor entry), not a stale value.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1441,13 +1525,44 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+    /**
+     * @notice Updates and verifies the HCU block limit.
+     * @param opHCU The HCU for the operation.
+     */
+    function _updateAndVerifyHCUBlockLimit(uint256 opHCU) internal virtual {
+        if (publicHCUPerBlock == type(uint192).max) {
+            return;
+        }
+
+        address caller = _getHCUCallerContext();
```
Eikix: Reviewer note: we consume and clear caller context right after read to prevent stale/leaked context reuse across later sub-calls in the same tx.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,6 +1408,64 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        assembly {
+            tstore(callerSlot, callerAddress)
+            tstore(setSlot, 1)
+        }
+    }
+
+    /**
+     * @notice Returns current block counter state.
+     */
+    function getBlockMeter() external view returns (uint64 blockNumber, uint192 usedHCU) {
```
Eikix: Reviewer note: getBlockMeter now returns effective state for the current block; if storage is stale (new block, no op yet), we return (currentBlock, 0).

host-contracts/contracts/HCULimit.sol
```diff
@@ -53,9 +62,23 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
     /// @dev This is the maximum number of homomorphic complexity units that can be used in a single transaction.
     uint256 private constant MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX = 20_000_000;
 
+    /// @dev Transient storage slot used to pass the original caller context from FHEVMExecutor.
+    bytes32 private constant HCU_CALLER_CONTEXT_SLOT = keccak256("fhevm.hcu.caller.context");
+    bytes32 private constant HCU_CALLER_CONTEXT_SET_SLOT = keccak256("fhevm.hcu.caller.context.set");
```
Eikix: Reviewer note: we intentionally split caller value + explicit set-flag in transient storage so missing context cannot silently fall back to address(0).

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,6 +1408,64 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+            emit BlockHCUWhitelistSet(accounts[i], isWhitelisted);
+        }
+    }
+
+    /**
+     * @notice Sets the original caller context for block-level HCU checks.
+     * @dev Must be called by FHEVMExecutor in the same transaction before checkHCUFor*.
+     * @param callerAddress Original FHEVMExecutor caller (EOA or contract).
+     */
+    function setHCUCallerContext(address callerAddress) external {
```
Eikix: Intent: executor sets caller context right before HCU accounting so nested sub-calls always use the current contract caller, not stale tx-level data.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1494,6 +1609,39 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+    function _packBlockMeter(uint64 blockNumber, uint192 usedHCU) internal pure virtual returns (uint256) {
+        return (uint256(usedHCU) << 64) | uint256(blockNumber);
+    }
+
+    function _unpackBlockMeter(uint256 packed) internal pure virtual returns (uint64 blockNumber, uint192 usedHCU) {
+        blockNumber = uint64(packed);
+        usedHCU = uint192(packed >> 64);
+    }
+
+    function _getHCUCallerContext() internal view virtual returns (address caller) {
```
Eikix: Fail-fast on missing context is intentional: if wiring is broken we revert early, instead of silently accounting against address(0) and hiding integration bugs.

host-contracts/contracts/HCULimit.sol
```diff
@@ -70,7 +93,10 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
      * @notice  Initializes the contract.
      */
     /// @custom:oz-upgrades-validate-as-initializer
-    function initializeFromEmptyProxy() public virtual onlyFromEmptyProxy reinitializer(REINITIALIZER_VERSION) {}
+    function initializeFromEmptyProxy() public virtual onlyFromEmptyProxy reinitializer(REINITIALIZER_VERSION) {
```
Eikix: Only mutable proxy state is initialized here (publicHCUPerBlock). ACL/FHEVMExecutor addresses are constants/immutables in bytecode, so there is no storage init for them.

Eikix:
@claude /pr-review

Eikix:
Added the whitelist-admin transfer flow we discussed, scoped only to block-HCU whitelist management.

What changed:
- `HCULimit`: added `blockHCUWhitelistAdmin` mapping.
- `HCULimit`: added `setBlockHCUWhitelistAdmin(account, admin)` (ACL owner only, account must already be whitelisted).
- `HCULimit`: added `moveBlockHCUWhitelist(from, to)` (callable by `from`'s whitelist admin) to atomically move whitelist + admin to a replacement address.
- `HCULimit`: unwhitelisting now clears admin state (`setBlockHCUWhitelist` + batch variant).
- Added focused tests for:
  - admin assignment precondition
  - non-admin transfer revert
  - successful move (old loses whitelist/admin, new gets both)
  - destination already-whitelisted revert
  - zero-address destination revert
  - admin cleanup on unwhitelist

Validation run locally:
- `forge test --match-path test/hcuLimit/HCULimit.t.sol`
- `DOTENV_CONFIG_PATH=.env.example HARDHAT_NETWORK=hardhat npx hardhat test test/upgrades/upgrades.ts`
- `make update-bindings` (host)

Eikix:
@claude /pr-review

Eikix:
Applied the stricter safety policy for `setHCUPerBlock`:

- `setHCUPerBlock` now accepts only:
  - `type(uint192).max` (disabled), or
  - values `>= MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX`.
- `0`, `1`, or any value below tx max now reverts with `InvalidHCUPerBlock`.

Tests updated accordingly:
- Added setter tests for:
  - revert when below tx max
  - accept tx max
  - existing disable sentinel behavior
- Block-cap behavior unit tests now use a mock-only helper (`setHCUPerBlockUnsafeForTest`) to keep low-cap scenario coverage without weakening production setter constraints.
- Invariant handler `setCap` now bounds fuzzed caps to valid values, preserving `fail_on_revert = true` stability.
- Encrypted ERC20 HCU tests were adapted to new semantics:
  - under valid cap, assert block meter accounting for non-whitelisted caller
  - assert whitelist bypass keeps meter at 0

Validation run:
- `forge test --match-path test/hcuLimit/HCULimit.t.sol`
- `forge test --match-path test/hcuLimit/HCULimit.invariants.t.sol`
- `DOTENV_CONFIG_PATH=.env.example HARDHAT_NETWORK=hardhat npx hardhat test test/encryptedERC20/EncryptedERC20.HCU.ts`
- `DOTENV_CONFIG_PATH=.env.example HARDHAT_NETWORK=hardhat npx hardhat test test/upgrades/upgrades.ts`

Eikix:
@claude /pr-review

host-contracts/contracts/HCULimit.sol
```diff
@@ -88,13 +137,10 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
-        FheType resultType,
-        bytes1 scalarByte,
-        bytes32 lhs,
-        bytes32 rhs,
-        bytes32 result
-    ) external virtual {
+    function checkHCUForFheAdd(FheType resultType, bytes1 scalarByte, bytes32 lhs, bytes32 rhs, bytes32 result)
+        external
+        virtual
+    {
```
obatirou: formating problem 
This should not change

host-contracts/contracts/HCULimit.sol
```diff
@@ -140,13 +186,10 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
-        FheType resultType,
-        bytes1 scalarByte,
-        bytes32 lhs,
-        bytes32 rhs,
-        bytes32 result
-    ) external virtual {
+    function checkHCUForFheSub(FheType resultType, bytes1 scalarByte, bytes32 lhs, bytes32 rhs, bytes32 result)
+        external
+        virtual
+    {
```
obatirou: same

host-contracts/contracts/HCULimit.sol
```diff
@@ -1368,18 +1354,133 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+                emit BlockHCUWhitelistAdminSet(accounts[i], address(0));
+            }
+            emit BlockHCUWhitelistSet(accounts[i], isWhitelisted);
+        }
+    }
+
+    /**
+     * @notice Sets the admin that can transfer whitelist status for a whitelisted account.
+     * @param account Whitelisted account whose admin is updated.
+     * @param admin Admin allowed to move whitelist from `account` to a replacement address.
```
obatirou: I do not understand what this is and why it is needed

host-contracts/contracts/HCULimit.sol
```diff
@@ -1368,18 +1354,133 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        emit BlockHCUWhitelistAdminSet(account, admin);
+    }
+
+    /**
+     * @notice Moves whitelist status from one address to another.
+     * @dev Intended for testnet contract redeployments where old contract address is replaced.
+     * @param from Current whitelisted address.
+     * @param to Replacement address that becomes whitelisted.
+     */
+    function moveBlockHCUWhitelist(address from, address to) external {
```
obatirou: Same, what is the point of this function?

host-contracts/contracts/HCULimit.sol
Eikix: this is to make the transfer of a whitelist easier on testnet -> you are approved once to have your app whitelisted
now if you redeploy the app from scratch (which happens many times), you can transfer the whitelist to the new deployment

maybe it is overkill if you think so

obatirou: imo overkill and pre mature optimization before a real need.
I would not assume redeploy the app from scratch often
Proxy should be used by dev for this?

obatirou: I would also assume only security focused app would be approved: hence some audits done or security minded/ experienced dev which should prevent the need for redeploy app from scratch.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1368,18 +1354,133 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        publicHCUPerBlock = hcuPerBlock;
+        emit HCUPerBlockSet(hcuPerBlock);
+    }
+
+    /**
+     * @notice Sets whitelist status for one caller.
+     * @param account Caller to update.
+     * @param isWhitelisted Whether caller bypasses the block cap.
+     */
+    function setBlockHCUWhitelist(address account, bool isWhitelisted) external onlyACLOwner {
```
obatirou: I would use addToWhitelist / removeFromWhitelist (+ batch variants) to avoid mistakes

host-contracts/contracts/HCULimit.sol
```diff
@@ -53,9 +78,26 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
     /// @dev This is the maximum number of homomorphic complexity units that can be used in a single transaction.
     uint256 private constant MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX = 20_000_000;
 
+    /// @dev Transient storage slot used to pass the original caller context from FHEVMExecutor.
+    bytes32 private constant HCU_CALLER_CONTEXT_SLOT = keccak256("fhevm.hcu.caller.context");
+    bytes32 private constant HCU_CALLER_CONTEXT_SET_SLOT = keccak256("fhevm.hcu.caller.context.set");
+
+    /// @notice Maximum homomorphic complexity units per block for non-whitelisted callers.
+    /// @dev Set to type(uint192).max to disable block-level throttling.
+    uint192 public publicHCUPerBlock;
```
obatirou: Also would use `publicHCUCapPerBlock`

host-contracts/test/hcuLimit/HCULimit.invariants.t.sol
```diff
@@ -0,0 +1,210 @@
...
+
+    function invariant_blockMeterMatchesCurrentBlock() public view {
+        (uint64 blockNumber,) = hcuLimit.getBlockMeter();
+        assertEq(blockNumber, uint64(block.number));
+    }
+
+    function invariant_blockMeterUsesCastGranularity() public view {
+        (, uint192 usedHCU) = hcuLimit.getBlockMeter();
+        assertEq(uint256(usedHCU) % 32, 0);
+    }
```
obatirou: What's this?
According to the fuzzing workshop, Montyly suggested to define invariants in english first and commit them.
Then match them to a corresponding test.

Eikix: can you explain how that would look like (the english first then commit then test)?
though i agree this invariant is useless, i'm removing it

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,6 +1415,82 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        bytes32 callerSlot = HCU_CALLER_CONTEXT_SLOT;
+        bytes32 setSlot = HCU_CALLER_CONTEXT_SET_SLOT;
+        assembly {
+            tstore(callerSlot, callerAddress)
+            tstore(setSlot, 1)
+        }
+    }
+
+    /**
+     * @notice Returns current block counter state.
```
Eikix: This doc comment is too vague. Let's make it more precise.

host-contracts/contracts/FHEVMExecutor.sol
Eikix: Let's add documentation

host-contracts/contracts/HCULimit.sol
```diff
@@ -1441,13 +1550,44 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+
         uint256 transactionHCU = opHCU + _getHCUForTransaction();
         if (transactionHCU >= MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX) {
             revert HCUTransactionLimitExceeded();
         }
         _setHCUForTransaction(transactionHCU);
     }
 
+    /**
+     * @notice Updates and verifies the HCU block limit.
```
Eikix: This isn't precise enough either.

host-contracts/test/encryptedERC20/EncryptedERC20.HCU.ts
```diff
@@ -86,4 +97,75 @@ describe('EncryptedERC20:HCU', function () {
     // Le euint64 (149000) + And ebool (25000) + Select euint64 (55000) + Sub euint64 (162000)
     expect(HCUMaxDepthTransferFrom).to.eq(391_000, 'HCU Depth incorrect');
   });
+
+  it('should account transferFrom in block meter when cap is enabled', async function () {
```
Eikix: we shouldn't call it block meter but block HCU meter

host-contracts/contracts/HCULimit.sol
```diff
@@ -53,9 +65,23 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+
+    /// @notice Maximum homomorphic complexity units per block for non-whitelisted callers.
+    /// @dev Set to type(uint192).max to disable block-level throttling.
+    uint192 public publicHCUCapPerBlock;
+
+    /// @dev Packed as: [usedHcu:192 | blockNumber:64]
+    uint256 private publicBlockMeterPacked;
+
+    /// @notice Whitelisted callers bypass block-level cap.
+    mapping(address => bool) public blockHCUWhitelist;
```
obatirou: please use ERC-7201 pattern with HCULimitStorageLocation defined below.
Surprisingly the storage location is already defined even though never used in this contract

host-contracts/contracts/HCULimit.sol
```diff
@@ -15,9 +15,21 @@ import {FheType} from "./shared/FheType.sol";
  * @dev The contract is designed to be used with the FHEVMExecutor contract.
  */
 contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
+    event HCUPerBlockSet(uint192 hcuPerBlock);
+    event BlockHCUWhitelistSet(address indexed account, bool isWhitelisted);
```
obatirou: add natspec + define after errors to stay consistent in style with other contracts

host-contracts/contracts/HCULimit.sol
obatirou, line 78: should be updated?

host-contracts/contracts/HCULimit.sol
```diff
@@ -70,15 +96,22 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+     * @dev This must be called when upgrading already-initialized proxies to this implementation,
+     *      so block HCU limiting remains disabled by default (`type(uint192).max`).
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    // function reinitializeV2() public virtual reinitializer(REINITIALIZER_VERSION) {}
+    function reinitializeV3() public virtual reinitializer(REINITIALIZER_VERSION) {
+        publicHCUCapPerBlock = type(uint192).max;
+        emit HCUPerBlockSet(type(uint192).max);
+    }
```
obatirou: should take in input the publicHCUCapPerBlock so it can be set at upgrade

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -697,6 +698,7 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
         if ((1 << uint8(toType)) & supportedTypes == 0) revert UnsupportedType();
         result = keccak256(abi.encodePacked(Operators.trivialEncrypt, pt, toType, acl, block.chainid));
         result = _appendMetadataToPrehandle(result, toType);
+        _setHCUCallerContext();
         hcuLimit.checkHCUForTrivialEncrypt(toType, result);
```
obatirou: maybe we could pass msg.sender to the checkHCUFor* instead.
Need to think about it

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -906,10 +912,19 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
         hcuLimit.checkHCUForFheRandBounded(randType, result);
         acl.allowTransient(result, msg.sender);
     }
 
+    /**
+     * @notice Propagates the original caller to HCULimit for block HCU whitelist checks.
+     * @dev Must be called immediately before each `hcuLimit.checkHCUFor*` path.
+     */
+    function _setHCUCallerContext() internal virtual {
+        hcuLimit.setHCUCallerContext(msg.sender);
```
obatirou: It will add a lot of gas if not mistaken as it is a cross contract call used for all operations

host-contracts/contracts/HCULimit.sol
```diff
@@ -53,9 +65,23 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
     /// @dev This is the maximum number of homomorphic complexity units that can be used in a single transaction.
     uint256 private constant MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX = 20_000_000;
 
+    /// @dev Transient storage slot used to pass the original caller context from FHEVMExecutor.
+    bytes32 private constant HCU_CALLER_CONTEXT_SLOT = keccak256("fhevm.hcu.caller.context");
+    bytes32 private constant HCU_CALLER_CONTEXT_SET_SLOT = keccak256("fhevm.hcu.caller.context.set");
+
+    /// @notice Maximum homomorphic complexity units per block for non-whitelisted callers.
+    /// @dev Set to type(uint192).max to disable block-level throttling.
+    uint192 public publicHCUCapPerBlock;
```
obatirou: why uint192?

host-contracts/foundry.toml
```diff
@@ -17,5 +17,10 @@ remappings = [
 # The number of fuzz runs for fuzz tests
 runs = 2560
 
+[profile.default.invariant]
+runs = 512
+depth = 48
```
obatirou, line 22: how did you choose those numbers?

library-solidity/tasks/taskDeploy.ts
obatirou, line 221: It is a deployment of new HCULimit 
In this case it will be an upgrade of already deployed proxies with task:upgradeHCULimit and I suppose upgradeFHEVMExecutor if not mistaken

host-contracts/contracts/HCULimit.sol
Eikix: so we can pack it with the block number (u64) to save a few thousand gas per tx

host-contracts/foundry.toml
Eikix: took something somewhat similar to morpho https://github.com/morpho-org/morpho-blue/blob/0427acea2f2c9066195a6824fb75967f9a9e7645/foundry.toml#L11

host-contracts/contracts/FHEVMExecutor.sol
Eikix: not sure! call can have very low gas cost, i could measure it

host-contracts/contracts/FHEVMExecutor.sol
Eikix: we'd have to think about the gas implications with data, not sure it's worth the gas save

host-contracts/contracts/FHEVMExecutor.sol
obatirou: It is a call + TLOAD + TSORE for all operations (+300) and will be linear with the number of operations used in a tx. We should mesure and see what's best 
Also passing msg.sender ensure that _setHCUCallerContext is not forgotten in subsequents upgrades + remove transient storage logic. 
Still need to wrap my head around this

host-contracts/contracts/HCULimit.sol
Eikix: do you think it's easier to have the upgrade without a cap, so the feature is just turned off (e.g. for mainnet or devnet we won't have to think about it) and then when we want to set it, the script can simply use the setter method?

host-contracts/contracts/FHEVMExecutor.sol
Eikix: yeah looking at the data it might come up expensive, i agree
for a erc20 test, the `_setHCUCallerContext` path increases gas cost by 3%

Eikix: though changing every `HCULimit.checkHCUFor*` is heavy ->
If HCULimit and FHEVMExecutor are not upgraded in lockstep, calls will fail due selector/signature mismatch and FHE ops can revert globally.

All checkHCUFor* selectors change. Any mocks/tests/tools/bindings expecting old ABI break until updated.

If someone passes the wrong address (e.g. not msg.sender), whitelist and block-meter accounting become wrong (priority bypass or over-throttling).

Eikix: Switched caller context from 2 transient slots to 1 slot

jatZama:
You forgot to update the HCULimit codegen which is located at https://github.com/zama-ai/fhevm/blob/main/library-solidity/codegen/src/hcuLimitGenerator.ts

host-contracts/contracts/HCULimit.sol
obatirou: in case we want to deploy it directly it should be passed as parameter.
If it is deactivated, you can pass the` type(uint192).max`

library-solidity/tasks/taskDeploy.ts
obatirou: Still need to be updated

jatZama:
You should also bump the reinitializer version of FHEVMExecutor since this will need an upgrade it seems.

host-contracts/contracts/FHEVMExecutor.sol
obatirou: @jatZama what are your thoughts on this?
maybe we could pass msg.sender to the checkHCUFor* instead

It avoids a call + transient storage
And ensure  _setHCUCallerContext is not forgotten in subsequents upgrades if any

host-contracts/contracts/FHEVMExecutor.sol
jatZama: Why not pass directly msg.sender via the hcuLimit.checkHCUForTrivialEncrypt function?

library-solidity/tasks/taskDeploy.ts
jatZama: i dont think you need to update the deployment script of the library, this should not be needed, you can deploy it from an empty proxy still

obatirou: Indeed, was wrong on that it can be deployed and the address passed to consumers

Eikix:
ready for re-review

host-contracts/contracts/HCULimit.sol
```diff
@@ -70,15 +102,24 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
      * @notice  Initializes the contract.
      */
     /// @custom:oz-upgrades-validate-as-initializer
-    function initializeFromEmptyProxy() public virtual onlyFromEmptyProxy reinitializer(REINITIALIZER_VERSION) {}
+    function initializeFromEmptyProxy() public virtual onlyFromEmptyProxy reinitializer(REINITIALIZER_VERSION) {
+        HCULimitStorage storage $ = _getHCULimitStorage();
+        $.publicHCUCapPerBlock = type(uint192).max;
```
jatZama: Can you please add an argument in the 2 reinitializer methods please for the `publicHCUCapPerBlock`? This will greatly simplify the upgrade process (now handled by the DAO, so please let's make this as simple as possible).

host-contracts/contracts/HCULimit.sol
```diff
@@ -40,7 +55,7 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
     uint256 private constant MINOR_VERSION = 1;
 
     /// @notice Patch version of the contract.
-    uint256 private constant PATCH_VERSION = 0;
+    uint256 private constant PATCH_VERSION = 1;
```
jatZama: Please keep PATCH_VERSION at 0, and bump MINOR_VERSION instead here.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -161,7 +161,7 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    // function reinitializeV2() public virtual reinitializer(REINITIALIZER_VERSION) {}
+    function reinitializeV3() public virtual reinitializer(REINITIALIZER_VERSION) {}
```
jatZama: No, this should be named reinitializeV2, not V3.

host-contracts/contracts/HCULimit.sol
```diff
@@ -18,6 +18,12 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
     /// @notice Returned if the sender is not the FHEVMExecutor.
     error CallerMustBeFHEVMExecutorContract();
 
+    /// @notice Returned if the block exceeds the maximum allowed homomorphic complexity units.
+    error HCUBlockLimitExceeded();
```
jatZama, line 22: Can you please run `make update-conformance` to update all selectors from host-contracts?

jatZama:
Don't forget to update the Foundry tests of host-contracts and to run the make command for conformance to get new selectors.

host-contracts/contracts/FHEVMExecutor.sol
jatZama: This should be called `reinitializeV2`, not `reinitializeV3`.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -138,7 +138,7 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
 
     /// Constant used for making sure the version number used in the `reinitializer` modifier is
     /// identical between `initializeFromEmptyProxy` and the `reinitializeVX` method
-    uint64 private constant REINITIALIZER_VERSION = 2;
+    uint64 private constant REINITIALIZER_VERSION = 3;
```
jatZama, line 141: Please also bump the MINOR_VERSION here.

host-contracts/contracts/HCULimit.sol
```diff
@@ -70,15 +102,24 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
-     * @notice Re-initializes the contract from V1.
-     * @dev Define a `reinitializeVX` function once the contract needs to be upgraded.
+     * @notice Re-initializes the contract from V2.
+     * @dev This must be called when upgrading already-initialized proxies to this implementation,
+     *      so block HCU limiting remains disabled by default (`type(uint192).max`).
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    // function reinitializeV2() public virtual reinitializer(REINITIALIZER_VERSION) {}
+    function reinitializeV3() public virtual reinitializer(REINITIALIZER_VERSION) {
```
jatZama: Same here, this function :
1/ should be called reinitializeV2 instead of reinitializeV3.
2/ should also take as argument the initial publicHCUCapPerBlock value.

host-contracts/contracts/HCULimit.sol
jatZama: Fully agree with @obatirou here, now upgrades take a very long time via the DAO, so better add it as argument here to make this atomic.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,11 +1471,97 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
             );
     }
 
+    /**
+     * @notice Sets the block-level HCU limit for non-whitelisted callers.
+     * @dev Set to type(uint192).max to disable block-level throttling.
+     * @param hcuPerBlock New block-level cap.
+     */
+    function setHCUPerBlock(uint192 hcuPerBlock) external onlyACLOwner {
+        if (hcuPerBlock != type(uint192).max && hcuPerBlock < MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX) {
```
jatZama: You can simplify this condition just using: 
```
if (hcuPerBlock < MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX)
```

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,11 +1471,97 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
             );
     }
 
+    /**
+     * @notice Sets the block-level HCU limit for non-whitelisted callers.
+     * @dev Set to type(uint192).max to disable block-level throttling.
+     * @param hcuPerBlock New block-level cap.
+     */
+    function setHCUPerBlock(uint192 hcuPerBlock) external onlyACLOwner {
```
jatZama: Can you refactor this via calling an internal function, such as:
```
function setHCUPerBlock(uint192 hcuPerBlock) external onlyACLOwner {
_setHCUPerBlock(hcuPerBlock);
}
```
Then reuse this _setHCUPerBlock internal function (containing exactly same logic except without the onlyACLOwner modifier) inside the 2 reinitializers methods please?

host-contracts/contracts/HCULimit.sol
jatZama: This should be setup with the event emitted via an internal setter (see next comment).

host-contracts/contracts/HCULimit.sol
```diff
@@ -70,15 +102,24 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+     * @notice Re-initializes the contract from V2.
+     * @dev This must be called when upgrading already-initialized proxies to this implementation,
+     *      so block HCU limiting remains disabled by default (`type(uint192).max`).
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    // function reinitializeV2() public virtual reinitializer(REINITIALIZER_VERSION) {}
+    function reinitializeV3() public virtual reinitializer(REINITIALIZER_VERSION) {
+        HCULimitStorage storage $ = _getHCULimitStorage();
+        $.publicHCUCapPerBlock = type(uint192).max;
```
jatZama: This should be setup with the event emitted via an internal setter (see next comment).

host-contracts/contracts/HCULimit.sol
```diff
@@ -87,13 +128,15 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
      * @param lhs The left-hand side operand.
      * @param rhs The right-hand side operand.
      * @param result Result.
+     * @param caller Original caller address from FHEVMExecutor.
```
jatZama: Maybe this comment could be slightly improved by mentionning this corresponds to the dapp contract, for eg wdyt "Original dapp caller address from FHEVMExecutor."

host-contracts/contracts/HCULimit.sol
jatZama: I am even wondering now if this check is even needed.
I would argue for the the sake of flexibility, why not removing it entirely? Maybe on some very fast chains we would realize that even the harcoded MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX constant is too high, so setting a lower HCU block limit in this case would be a good solution to mitigate DDOS. Wdyt?

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,11 +1471,97 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        }
+        HCULimitStorage storage $ = _getHCULimitStorage();
+        $.publicHCUCapPerBlock = hcuPerBlock;
+        emit HCUPerBlockSet(hcuPerBlock);
+    }
+
+    /**
+     * @notice Returns the public block HCU cap.
+     */
+    function publicHCUCapPerBlock() public view virtual returns (uint192) {
```
jatZama: Nit : I prefer to put all those public view getters towards the end of the contract, next to the `getVersion` function, and also can you update their name to make them start with the "get" verb for improved readability?

jatZama: Could also use the `is` verb, i.e `getPublicHCUCapPerBlock`, `isBlockHCUWhitelist`

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,11 +1471,97 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+     */
+    function removeFromBlockHCUWhitelist(address account) external onlyACLOwner {
+        _setBlockHCUWhitelist(account, false);
+    }
+
+    /**
+     * @notice Adds a batch of callers to the block-cap whitelist.
+     * @param accounts Callers to whitelist.
+     */
+    function addToBlockHCUWhitelistBatch(address[] calldata accounts) external onlyACLOwner {
```
jatZama: Those batch variants (addToBlockHCUWhitelistBatch and removeFromBlockHCUWhitelistBatch) are not strictly needed btw, personally I would remove them to reduce deployment gas. Since the DAO owning the contract can already batch calls in a single proposal.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,11 +1471,97 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        for (uint256 i = 0; i < accounts.length; i++) {
+            _setBlockHCUWhitelist(accounts[i], false);
+        }
+    }
+
+    /**
+     * @notice Returns the effective public block HCU meter for the current block.
+     * @dev If storage still contains a previous block meter, returns `(block.number, 0)`.
+     */
+    function getBlockMeter() external view returns (uint64 blockNumber, uint192 usedHCU) {
```
jatZama: nit: This public getter has a good name, but as for previous ones, I would prefer to place it around the getVersion method, towards the end of the code.

host-contracts/contracts/HCULimit.sol
```diff
@@ -53,14 +68,31 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
     /// @dev This is the maximum number of homomorphic complexity units that can be used in a single transaction.
     uint256 private constant MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX = 20_000_000;
 
+    /// @custom:storage-location erc7201:fhevm.storage.HCULimit
+    struct HCULimitStorage {
+        /// @notice Maximum homomorphic complexity units per block for non-whitelisted callers.
+        /// @dev Set to type(uint192).max to disable block-level throttling.
+        uint192 publicHCUCapPerBlock;
+        /// @dev Packed as: [usedHcu:192 | blockNumber:64]
```
jatZama: I would prefer not to do manual packing here for readability please. The compiler is smart enough to do the packing automatically, just replace `uint256 publicBlockMeterPacked` by two variables in the struct: `uint192 usedHcu; uint64 blockNumber;`.
It is a lot more readable, and you don't waste gas.

jatZama: I think you can even do better, and pack the 3 variables into a single storage slot, just by using 3 uint64 for the 3 varaibles: 
publicHCUCapPerBlock, usedHcu and blockNumber.

jatZama: This will save even more gas and be more readable.
Also maybe I would rename : `publicHCUCapPerBlock` -> `globalHCUCapPerBlock` and `blockNumber`-> `lastSeenBlockNumber` and `usedHcu` -> `usedHCU` or `usedBlockHCU`.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1382,11 +1471,97 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        }
+    }
+
+    /**
+     * @notice Returns the effective public block HCU meter for the current block.
+     * @dev If storage still contains a previous block meter, returns `(block.number, 0)`.
+     */
+    function getBlockMeter() external view returns (uint64 blockNumber, uint192 usedHCU) {
+        HCULimitStorage storage $ = _getHCULimitStorage();
+        (uint64 storedBlock, uint192 storedHCU) = _unpackBlockMeter($.publicBlockMeterPacked);
```
jatZama: I would prefer to also make `storedHCU` as a uint64 (as well as the global block limit) and pull these directly from storage, instead of manually packing/unpacking. The Solidity compiler does automatic storage packing.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1439,15 +1616,47 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
 
+    /**
+     * @notice Updates and enforces the public block HCU cap for one operation.
+     * @dev No-op if the cap is disabled or caller is whitelisted.
+     * @param opHCU HCU cost of the current operation.
+     * @param caller Original caller address from FHEVMExecutor.
+     */
+    function _updateAndVerifyHCUBlockLimit(uint256 opHCU, address caller) internal virtual {
+        HCULimitStorage storage $ = _getHCULimitStorage();
+        if ($.publicHCUCapPerBlock == type(uint192).max) {
```
jatZama: I don't think this check is needed tbh, even if switching to uint64, the limit is so high it would never be reached in practise, and even if it does, the tx would revert due to an overflow here.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1494,6 +1703,26 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        usedHCU = uint192(packed >> 64);
+    }
+
+    /**
+     * @notice Sets whitelist status for one caller.
+     * @param account Caller to update.
+     * @param isWhitelisted Whether caller bypasses the block cap.
+     */
+    function _setBlockHCUWhitelist(address account, bool isWhitelisted) internal {
+        HCULimitStorage storage $ = _getHCULimitStorage();
```
jatZama, line 1702: Can you add some error handling here, reverting if you try to whitelist an already whitelisted address, and also reverting (with another error) if you try to remove from whitelist an already non-whitelisted account?

host-contracts/contracts/HCULimit.sol
obatirou: I agree but in that case, should MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX be also configurable for fast chains as the HCUPerBlock is only for non whitelisted contracts ?

jatZama: Fair point. I don't have a very strong opinion on this, it would increase gas cost if adding another storage variable, but maybe not so much if we pack them in 4 uint64s.
Maybe for fast chains we would probably never whitelist any account tho? Not sure.

jatZama: the issue is that we have 2 constants atm: MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX and MAX_HOMOMORPHIC_COMPUTE_UNITS_DEPTH_PER_TX, so adding 2 sotrage variables is not great, because 5 uint64 do not fit in a single uint256. So not a big fan tbh.

jatZama: I think in an "emergency DDOS situation" what would happen is governance would set a very low global block cap, and remove from whitelist all privileged addresses. This could be done in a single proposal I think (unless whitelisted accounts is a huge set)...

jatZama: There is another way though, maybe best solution would be to use 3 uint64 (lastSeenBlock, usedBlockHCU and globalHCUPerBlock) + 2 uint32 for replacing the 2 current constants : MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX and MAX_HOMOMORPHIC_COMPUTE_UNITS_DEPTH_PER_TX.
This seems to be the best tradeoff for flexibility and gas savings.
WDYT?

jatZama: Actually we could probably use 5 uint32s (ie uint32 for every storage variable), this seems enough.

jatZama: Or 5 uint48 to have more margin (but this means we could not add cheaply more storage vars in the future for any reason).

host-contracts/contracts/HCULimit.sol
Eikix: > I think you can even do better, and pack the 3 variables into a single storage slot, just by using 3 uint64 for the 3 varaibles: publicHCUCapPerBlock, usedHcu and blockNumber.

ok!

host-contracts/contracts/HCULimit.sol
obatirou: > Actually we could probably use 5 uint32s (ie uint32 for every storage variable), this seems enough.

Indeed could be a solution

> Maybe for fast chains we would probably never whitelist any account tho? Not sure

> I think in an "emergency DDOS situation" what would happen is governance would set a very low global block cap, and remove from whitelist all privileged addresses. This could be done in a single proposal I think (unless whitelisted accounts is a huge set)...

Should this be a new issue / discussion?
It seems out of the scope of per block HCU and not trivial / need inputs from execs regarding fast chains?
Those points should be discussed for example and think about consequences.

jatZama: Maybe yes you could, but I am not fully convinced it makes so much sense ask leadership implementation details. We could also think about the best way to be more flexible and future-proof.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1351,42 +1442,50 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
-                    Strings.toString(MINOR_VERSION),
-                    ".",
-                    Strings.toString(PATCH_VERSION)
-                )
-            );
+    function addToBlockHCUWhitelist(address account) external onlyACLOwner {
+        HCULimitStorage storage $ = _getHCULimitStorage();
+        if ($.blockHCUWhitelist[account]) revert AlreadyBlockHCUWhitelisted(account);
+        $.blockHCUWhitelist[account] = true;
+        emit BlockHCUWhitelistSet(account, true);
```
jatZama: Nit: maybe you could replace this event with the something more readable without a boolean: for eg: `emit BlockHCUWhitelistSet(account, true);` ->  `emit BlockHCUWhitelistAdded(account);` and the next event `emit BlockHCUWhitelistSet(account, false);` -> `emit BlockHCUWhitelistSetRemoved(account);`.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1439,15 +1540,45 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
      * @notice Updates and verifies the HCU transaction limit.
      * @param opHCU The HCU for the operation.
+     * @param caller Original caller address for block-level checks.
      */
-    function _updateAndVerifyHCUTransactionLimit(uint256 opHCU) internal virtual {
+    function _updateAndVerifyHCUTransactionLimit(uint256 opHCU, address caller) internal virtual {
+        _updateAndVerifyHCUBlockLimit(opHCU, caller);
+
         uint256 transactionHCU = opHCU + _getHCUForTransaction();
         if (transactionHCU >= MAX_HOMOMORPHIC_COMPUTE_UNITS_PER_TX) {
```
jatZama: nit : I just realized it would have been more sound/readable to put strict inequalities here on all thresholds. But this issue was already present before your PR...

host-contracts/contracts/HCULimit.sol
```diff
@@ -1439,15 +1540,45 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        }
+
+        uint64 currentBlock = uint64(block.number);
+        uint64 storedHCU = $.usedBlockHCU;
+        if ($.lastSeenBlockNumber != currentBlock) {
+            storedHCU = 0;
+        }
+
+        uint256 nextHCU = uint256(storedHCU) + opHCU;
+        if (nextHCU >= uint256($.globalHCUCapPerBlock)) {
```
jatZama: nit : So if you decide to put strict inequalities to the previously existing tx and depth constants, it would also make sense to use a strict inequality here.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -161,7 +161,7 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    // function reinitializeV2() public virtual reinitializer(REINITIALIZER_VERSION) {}
+    function reinitializeV2() public virtual reinitializer(REINITIALIZER_VERSION) {}
```
jatZama, line 164: Please don't forget to bump the REINITIALIZER_VERSION constant.

host-contracts/contracts/HCULimit.sol
```diff
@@ -53,6 +71,18 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        uint64 usedBlockHCU;
+        /// @notice Last seen block number for the block meter.
+        uint64 lastSeenBlockNumber;
+        /// @notice Whitelisted callers bypass block-level cap.
+        mapping(address => bool) blockHCUWhitelist;
+    }
+
     /// Constant used for making sure the version number used in the `reinitializer` modifier is
     /// identical between `initializeFromEmptyProxy` and the `reinitializeVX` method
     uint64 private constant REINITIALIZER_VERSION = 2;
```
jatZama: Please don't forget to bump the REINITIALIZER_VERSION to 3 here.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -138,7 +138,7 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
 
     /// Constant used for making sure the version number used in the `reinitializer` modifier is
     /// identical between `initializeFromEmptyProxy` and the `reinitializeVX` method
-    uint64 private constant REINITIALIZER_VERSION = 3;
+    uint64 private constant REINITIALIZER_VERSION = 2;
```
jatZama: `REINITIALIZER_VERSION` should be 3 here.

host-contracts/contracts/HCULimit.sol
```diff
@@ -71,17 +74,18 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        /// @notice Last seen block number for the block meter.
+        uint64 lastSeenBlockNumber;
         /// @notice Whitelisted callers bypass block-level cap.
         mapping(address => bool) blockHCUWhitelist;
     }
 
     /// Constant used for making sure the version number used in the `reinitializer` modifier is
     /// identical between `initializeFromEmptyProxy` and the `reinitializeVX` method
-    uint64 private constant REINITIALIZER_VERSION = 3;
+    uint64 private constant REINITIALIZER_VERSION = 2;
```
jatZama: `REINITIALIZER_VERSION` should be 3 here.

host-contracts/contracts/HCULimit.sol
```diff
@@ -1632,11 +1667,29 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
      * @notice Sets the global HCU cap per block.
      * @param hcuPerBlock New cap value.
      */
-    function _setHCUPerBlock(uint64 hcuPerBlock) internal {
+    function _setHCUPerBlock(uint48 hcuPerBlock) internal {
```
jatZama, line 1677: I think it would be better to enforce expected invariants in those 3 new setters here now that we have some flexibility:
hcuPerBlock >= maxHCUPerTx -> to be enforced in _setHCUPerBlock
and:
maxHCUPerTx >= maxHCUDepthPerTx -> to be enforced in _setMaxHCUDepthPerTx AND _setMaxHCUPerTx

host-contracts/tasks/upgradeContracts.ts
```diff
@@ -287,8 +287,34 @@ task('task:upgradeHCULimit')
     true,
     types.boolean,
   )
+  .addOptionalParam(
+    'hcuCapPerBlock',
+    'Global HCU cap per block passed to reinitializeV2 (default: uint48 max)',
+    '281474976710655',
```
jatZama: nit : put a comment such as `// type(uint48).max`

host-contracts/test/encryptedERC20/EncryptedERC20.HCU.ts
```diff
@@ -86,4 +99,75 @@ describe('EncryptedERC20:HCU', function () {
...
+    const encryptedAllowanceAmount = await inputAlice.encrypt();
+    const approveTx = await this.erc20['approve(address,bytes32,bytes)'](
+      this.signers.bob.address,
+      encryptedAllowanceAmount.handles[0],
+      encryptedAllowanceAmount.inputProof,
+    );
+    await approveTx.wait();
+
+    const ownerHcuLimit = this.hcuLimit.connect(this.signers.fred);
+    await ownerHcuLimit.setHCUPerBlock(MIN_HCU_PER_BLOCK);
```
jatZama: I don't think you need to call this setter anymore.

host-contracts/test/encryptedERC20/EncryptedERC20.HCU.ts
```diff
@@ -86,4 +99,75 @@ describe('EncryptedERC20:HCU', function () {
...
```
jatZama: this setter as well seems unnecessary.

host-contracts/test/encryptedERC20/EncryptedERC20.HCU.ts
```diff
@@ -86,4 +99,75 @@ describe('EncryptedERC20:HCU', function () {
...
+      this.signers.alice.address,
+      this.signers.bob.address,
+      encryptedTransferAmount.handles[0],
+      encryptedTransferAmount.inputProof,
+    );
+    const receipt = await tx.wait();
+    expect(receipt?.status).to.eq(1);
+
+    const [, usedHCU] = await this.hcuLimit.getBlockMeter();
+    expect(usedHCU).to.be.greaterThan(0n);
```
jatZama, line 115: you don't need to do two tx (approve then transferFrom), you could directly check `expect(usedHCU).to.be.greaterThan(0n);` right after the approve tx for instance.

jatZama (approved):
LGTM, thanks!

Eikix:
@mergify queue

PR #2013, opened 2026-02-19
Antoniu/test multi copro

antoniupop:


Eikix:
Did this need to be in v0.12?

PR #2014, opened 2026-02-20
fix(host-contracts): add domain separator and prev block hash to handle hashing

antoniupop:
The addition of domain separators is standard practice -  needed in all hashes to avoid risk of collision when hashing the same thing in different contexts.

The previous block hash is needed to prevent collisions between handles. This is an issue for security and for consensus under re-randomisation. Full context in #guild-rerand


Eikix:
question, what is our strategy for when to switch the version of the host contracts? like for this change, should we change the patch version?
reinitialize version? etc.

obatirou (commented):
Lgtm to me 
Want to sync with past maintainers for the version and how to handle upgrades / audits

enitrat:
hey, I think we need to ping devx on that, as it might be a breaking change for how the hardhat tooling works (not 100% sure but I know they have a MockExecutor with some logic wrt handle logic compute)

antoniupop:
> hey, I think we need to ping devx on that, as it might be a breaking change for how the hardhat tooling works (not 100% sure but I know they have a MockExecutor with some logic wrt handle logic compute)

cc @0xalexbel @poppyseedDev 
TLDR: we need domain separators for hashing in handle generation - I added these in input handles (8 months ago), but missed the computed handles. If this breaks the mock, let's sync. This is not going to be merged right away.

**Confirmed with @0xalexbel this should not impact mock**

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -859,15 +862,15 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
+        result = keccak256(abi.encodePacked(COMPUTATION_DOMAIN_SEPARATOR, op, lhs, middle, rhs, acl, block.chainid, blockhash(block.number - 1)));
         result = _appendMetadataToPrehandle(result, middleType);
         acl.allowTransient(result, msg.sender);
     }
 
     function _generateSeed() internal virtual returns (bytes16 seed) {
         FHEVMExecutorStorage storage $ = _getFHEVMExecutorStorage();
         seed = bytes16(
-            keccak256(abi.encodePacked($.counterRand, acl, block.chainid, blockhash(block.number - 1), block.timestamp))
+            keccak256(abi.encodePacked(COMPUTATION_DOMAIN_SEPARATOR, $.counterRand, acl, block.chainid, blockhash(block.number - 1), block.timestamp))
```
jat9292: I don't think you need the `COMPUTATION_DOMAIN_SEPARATOR`, this line is computing the seed, it is not here that the handle is computed, but in `_generateRand` and `_generateRandBounded` methods.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -883,7 +886,7 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
 
         /// @dev Unsupported erandom type.
         if ((1 << uint8(randType)) & supportedTypes == 0) revert UnsupportedType();
-        result = keccak256(abi.encodePacked(Operators.fheRand, randType, seed));
+        result = keccak256(abi.encodePacked(COMPUTATION_DOMAIN_SEPARATOR, Operators.fheRand, randType, seed));
```
jat9292, line 890: I think you forgot to put `blockhash(block.number - 1)` here.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -904,7 +907,7 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
         if ((1 << uint8(randType)) & supportedTypes == 0) revert UnsupportedType();
         if (!_isPowerOfTwo(upperBound)) revert NotPowerOfTwo();
         _checkBelowMaxBound(upperBound, randType);
-        result = keccak256(abi.encodePacked(Operators.fheRandBounded, upperBound, randType, seed));
+        result = keccak256(abi.encodePacked(COMPUTATION_DOMAIN_SEPARATOR, Operators.fheRandBounded, upperBound, randType, seed));
```
jat9292, line 911: I think you forgot to put `blockhash(block.number - 1)` here.

host-contracts/contracts/FHEVMExecutor.sol
jat9292: Maybe instead you should use a specific separator here, such as `SEED_DOMAIN_SEPARATOR`

jat9292:
I put few comments already in the code, noticed few small issues.
Also I had a more general comment here, but since I am missing some context I am asking directly here: if you use blockhash in the handles preimage to avoid collisions due to reorg, why not also use block.timestamp (like I already did for the oprf seed)? This would mitigate a lot of reorgs collisions cases : 99.4% of reorgs are due to missed slots, those would be caught by this method, to reduce collision risk even further.

Mocked mode in the hardhat plugin has to be updated obviously. Could you also share what separator do you use for user encrypted inputs (aka "input handles") to make the mocked mode more realistic, and share it with Alex B and me?

Also please note that this PR introduces a big change: previously, all trivially encrypted handles for a given plaintext gave the same handle as result, but this is no longer the case, you will get different handles for the same trivially encrypted value except if in the same block. Hopefully no already deployed contract relied on this behavior.

host-contracts/contracts/FHEVMExecutor.sol
antoniupop: I think it's already included in the seed (same as chainId)

host-contracts/contracts/FHEVMExecutor.sol
antoniupop: same as above

host-contracts/contracts/FHEVMExecutor.sol
antoniupop: Yes, good point - done

antoniupop:
> I put few comments already in the code, noticed few small issues. Also I had a more general comment here, but since I am missing some context I am asking directly here: if you use blockhash in the handles preimage to avoid collisions due to reorg, why not also use block.timestamp (like I already did for the oprf seed)? This would mitigate a lot of reorgs collisions cases : 99.4% of reorgs are due to missed slots, those would be caught by this method, to reduce collision risk even further.
> 
> Mocked mode in the hardhat plugin has to be updated obviously. Could you also share what separator do you use for user encrypted inputs (aka "input handles") to make the mocked mode more realistic, and share it with Alex B and me?
> 
> Also please note that this PR introduces a big change: previously, all trivially encrypted handles for a given plaintext gave the same handle as result, but this is no longer the case, you will get different handles for the same trivially encrypted value except if in the same block. Hopefully no already deployed contract relied on this behavior.

Done.

host-contracts/contracts/FHEVMExecutor.sol
jatZama: Ok, then I agree it is not needed to put it here again, if you also think it is better this way, fine.

host-contracts/contracts/FHEVMExecutor.sol
jatZama: I agree it is not needed to put it here again, if you also think it is better this way, fine.

jatZama (approved):

antoniupop:
The rationale for this change is detailed in the [RFC 10](https://github.com/zama-ai/tech-spec/blob/rfc-010/rfcs/010-block-dependent-handles.md#rfc-010-block-dependent-handles)

antoniupop:
@mergifyio queue

PR #2020, opened 2026-02-23
feat(cli): rework fhevm-cli as typed boot pipeline

Eikix:
## Summary

Replaces the legacy bash-based `deploy-fhevm-stack.sh` orchestration with a typed TypeScript/Bun CLI (`fhevm-cli`) that manages the full local development stack.

### What's new

- **Configuration engine** — HD key derivation from mnemonic, typed config model, per-service env generation, version resolution from GitHub releases
- **Docker orchestration layer** — compose wrapper, container inspection, readiness polling (HTTP, log-sentinel, exit-code, docker-state), service topology map
- **13-step boot pipeline** — ordered dependency graph with resume (`--from N`), parallel step support, MinIO/KMS signer runtime discovery
- **Operational commands** — `up`, `down`, `clean`, `restart`, `status`, `logs`, `test`, `pause`, `unpause`, `doctor`
- **Multi-coprocessor support** — configurable topology (N coprocessors, KMS nodes, custodians)
- **CI mode** — `--ci` flag with NDJSON output, BuildKit cache, diagnostic log capture, key cache snapshots
- **Relayer config generation** — template-based YAML patching for relayer v0.9.0 compatibility

### What's removed

- `test-suite/fhevm/scripts/deploy-fhevm-stack.sh` (926 lines of bash)
- `test-suite/fhevm/scripts/setup-kms-signer-address.sh`
- `test-suite/fhevm/scripts/debug-container.sh`
- `test-suite/fhevm/fhevm-cli` (old binary stub)
- Hardcoded `.env.*` staging files (now generated from config model)

## Test plan

- [x] 246 unit tests pass (`bun test` in `cli/`)
- [x] Full 13-step boot pipeline completes end-to-end on local Docker
- [ ] CI workflow integration (separate PR)


Eikix:
This rework fixes a big chunk of the local deploy pain we hit, but not all of it.

What this PR already improves a lot:
- Boot ordering is much safer now (gateway deploy -> discover addresses -> host deploy -> rediscover -> regen downstream env files).
- The stale-address issue that was breaking `gateway-sc-trigger-keygen` is handled by the explicit discovery + env regeneration flow.
- `docker compose up` is now scoped per step/service instead of blindly bringing everything up all the time.
- Resume/state/step diagnostics are much better than the old bash flow.

What still looks like a blind spot vs the failures we saw:
1. Version source is still not aligned to gitops/devnet by default.
   - `cli/src/config/versions.ts` uses pinned `2458fa9` for multiple groups and otherwise "latest release/tag".
   - This can still drift from the stack actually running on devnet.
   - We need an explicit "version profile" mode (ex: gitops snapshot / lockfile) so local validation can match deployed infra deterministically.

2. Readiness gates are still mostly service-level, not consensus-level.
   - The pipeline waits for services and discovers keys/addresses, but it does not prove the full multi-coproc path is ready for workload (especially threshold / sns-worker effective readiness).
   - We should add a post-boot health gate that validates the consensus-critical path before tests run.

3. Missing preflight/doctor for common local failure modes.
   - A lot of time is still lost on "why this run is broken" (orphan containers, stale state, wrong env mix, resource exhaustion).
   - A lightweight `doctor` command (or preflight stage) would make this much less flaky in practice.

Net: strong direction and definitely better than current bash deploy, but I wouldn’t call it sufficient yet for "reproducible local == devnet-like" until version profile + stronger readiness checks are in.

Eikix:
Ran a full local multicopro regression attempt from `antoniu/test-multi-copro` and `antoniu/cosensus-decompression` and collected what actually made runs flaky.

The short version: yes, the "dirty" files were useful. They basically encode the exact runtime pain points we had to patch around manually.

## What the dirty files told us

They’re not random drift. They cluster around a few failure classes:

- boot ordering / idempotency hardening
  - `gateway-contracts/tasks/addHostChains.ts`
  - `gateway-contracts/tasks/addPausers.ts`
  - `host-contracts/tasks/addPausers.ts`
- local topology + deploy orchestration gaps
  - `test-suite/fhevm/fhevm-cli`
  - `test-suite/fhevm/scripts/deploy-fhevm-stack.sh`
  - `test-suite/fhevm/docker-compose/*.yml`
- relayer / test env drift after deploy
  - `test-suite/fhevm/env/staging/.env.relayer`
  - `test-suite/fhevm/env/staging/.env.test-suite`
- KMS key/signer bootstrap resolution issues
  - `test-suite/fhevm/scripts/setup-kms-signer-address.sh`
  - `test-suite/fhevm/env/staging/.env.kms-connector`
- resource limit symptom under multicopro
  - `database-docker-compose.yml` (`max_connections=300`)

So yes, these edits gave us signal about where flakiness came from.

## Flakiness report (observed)

1. Relayer key URL and endpoint shape drift
- Symptom: relayer/test path mismatches (`/v2` vs inferred key URL path), 404/Bad JSON style failures.
- Impact: tests fail before consensus path is even exercised.
- Need in CLI: explicit post-boot key URL validation (HTTP 200 on FHE public key + CRS), fail fast with actionable error.

2. KMS bootstrap race (events missed)
- Symptom: listener/poller starts from `latest` when no previous cursor; keygen/crsgen can be missed if timing is unlucky.
- Impact: relayer never gets valid key material for the current run.
- Need in CLI: a deterministic bootstrap gate, not just service-up checks. Verify key material exists and is fresh for the current deployment window.

3. Address propagation drift to test env
- Symptom: required test vars missing or stale (`CIPHERTEXT_COMMITS_ADDRESS` notably).
- Impact: e2e tests fail with config errors or run against wrong contracts.
- Need in CLI: single source-of-truth address export after deploy, plus contract-address completeness check before `test`.

4. One-shot step semantics not robust enough
- Symptom: some deploy steps complete work but container state is not interpreted as completed in a stable way.
- Impact: pipeline stalls or requires manual intervention.
- Need in CLI: first-class "oneshot completed" semantics via explicit completion markers + exit-state policy.

5. Multicopro readiness was service-level, not consensus-level
- Symptom: containers are "running" but effective threshold path is not ready.
- Impact: intermittent early test failures.
- Need in CLI: consensus-critical readiness gate (e.g. key bootstrap seen by enough coprocessors for threshold).

6. Resource defaults too low for multicopro stress
- Symptom: DB connection pressure under multicopro workloads.
- Impact: random downstream failures that look unrelated.
- Need in CLI: profile-aware DB sizing defaults (or automatic warning if pool config > DB cap).

7. Generated topology artifacts leak between runs
- Symptom: generated compose files linger (`coprocessor-*.generated.*.yml`).
- Impact: run-to-run contamination and hard-to-debug behavior.
- Need in CLI: deterministic generated-file lifecycle (create in temp dir + guaranteed cleanup).

## What PR #2020 already improves

- Typed pipeline + step graph is a big improvement over the bash flow.
- Better resume/diagnostics shape.
- Better env generation model than static hand-edited `.env.*`.

## What still needs to be explicit before we can call local runs reliable

- version profile mode aligned with deployed infra (gitops snapshot/lock), not implicit "latest" behavior
- consensus-path readiness gate (not only service liveness)
- key material freshness/availability gate
- address completeness gate for test mode
- generated artifact cleanup guarantees
- multicopro resource preflight (DB + pool sanity)

If useful I can turn this into a concrete checklist in `doctor` so failures become deterministic + self-explanatory.

PR #2028, opened 2026-02-25
feat(host-contracts): implement context-aware KMSVerifier

obatirou:
Closes https://github.com/zama-ai/fhevm-internal/issues/1062

Add per-context signer sets and thresholds so that old KMS contexts remain verifiable after governance rotates signers via defineNewContext. Includes destroyKmsContext, v0/v1 extraData parsing, reinitializeV2 migration path, legacy view redirection, and IKMSVerifier interface updates.


obatirou:
@claude /pr-review

obatirou:
@claude /pr-review

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -305,37 +401,116 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
...
+            KMSVerifierStorage storage $ = _getKMSVerifierStorage();
+            return $.currentKmsContextId;
+        }
+        uint8 version = uint8(extraData[0]);
+        if (version == 0x01) {
+            // v1 (0x01 prefix): reads a 32-byte context ID starting at byte 1.
+            // Trailing bytes after byte 33 are ignored for forward-compatibility
+            // with potential v1 extensions.
+            if (extraData.length < 33) {
+                revert DeserializingDecryptionProofFail();
```
isaacdecoded: Not sure if the `DeserializingDecryptionProofFail` error is the most accurate one here since the deserialization is specifically in the `extraData`. I mean, the `decryptionProof` is the full bytes containing this other `extraData` bytes.

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -248,7 +292,7 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
      */
     function getThreshold() public view virtual returns (uint256) {
         KMSVerifierStorage storage $ = _getKMSVerifierStorage();
-        return $.threshold;
+        return $.contextThreshold[$.currentKmsContextId];
```
isaacdecoded, line 287: Maybe worth specifying in the NatSpec comments that these methods (`setThreshold`, `getKmsSigners`, `getThreshold`) are directly considering the current KMS context. Still, I know we're still figuring out if we should deprecate these methods instead 🤔

host-contracts/contracts/KMSVerifier.sol
obatirou: Indeed good point 
Will use `DeserializingExtraDataFail`

host-contracts/contracts/KMSVerifier.sol
obatirou: How could we decide?
Who has the final word on this?

host-contracts/contracts/KMSVerifier.sol
obatirou: Done

host-contracts/contracts/KMSVerifier.sol
isaacdecoded: Thinking twice, I think the best is to keep them, or at least the `getKmsSigners` and `getThreshold` methods. Since these methods are used by the FHE.sol library. Meaning that some apps could be using an old version of this library, and after we upgrade the KMSVerifier contract, these outdated apps will try to call these methods and fail if we remove them.

The only one I believe we can fully replace is the `setThreshold` method because it's a governance calling method.

Makes sense to you? @obatirou

obatirou: Yes make sense 
Let's remove only `setThreshold` then
Will make the change

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -120,16 +157,23 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
...
         __EIP712_init(CONTRACT_NAME_SOURCE, "1", verifyingContractSource, chainIDSource);
+        KMSVerifierStorage storage $ = _getKMSVerifierStorage();
+        $.currentKmsContextId = KMS_CONTEXT_COUNTER_BASE;
         defineNewContext(initialSigners, initialThreshold);
     }
 
     /**
-     * @notice Re-initializes the contract from V1.
-     * @dev Define a `reinitializeVX` function once the contract needs to be upgraded.
+     * @notice Re-initializes the contract to V3 with context-aware KMS support.
```
jatZama: `Re-initializes the contract from V1` would be more correct here.

jatZama: or `to V2` maybe.
But I prefer to also mention it is going from V1.
So maybe` Re-initializes the contract from V1 to V2`

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -120,16 +157,23 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
...
+     * @dev Migrates existing signers into the first context. The legacy mapping retains its pre-migration
+     *      values and should be considered stale after this call.
```
jatZama: i think there is a typo here ` values and should be..`?

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -174,9 +202,27 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
      * @param threshold    The threshold to be set. Threshold should be non-null and less than the number of signers.
      */
     function setThreshold(uint256 threshold) public virtual onlyACLOwner {
-        _setThreshold(threshold);
         KMSVerifierStorage storage $ = _getKMSVerifierStorage();
-        emit NewContextSet($.signers, threshold);
+        _setContextThreshold($.currentKmsContextId, threshold);
+        emit NewContextSet($.currentKmsContextId, $.contextSigners[$.currentKmsContextId], threshold);
```
jatZama: Is it really expected to keep currentKmsContextId unchanged when changing the threshold? I find it a bit counter-intuitive, since in my mind a context is uniquely defined by a set of signers AND a threshold, but here you loose unicity of the id when changing threshold...

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -305,37 +404,116 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
...
+
+    /**
+     * @notice              Extracts the KMS context ID from extra data.
+     * @param extraData     The extra data bytes from the decryption proof.
+     * @return contextId    The extracted KMS context ID.
+     */
+    function _extractKmsContextId(bytes memory extraData) internal view virtual returns (uint256) {
+        // v0 (0x00 prefix or empty): uses the current context. Trailing bytes are
+        // ignored for forward-compatibility with potential v0 extensions.
+        if (extraData.length == 0 || uint8(extraData[0]) == 0x00) {
```
jatZama, line 450: do we really need this check "extraData.length == 0" ? i thoguht currently deployed version alwways expected already extraData="0x00" , so I don't see any case where length would be null?

host-contracts/contracts/KMSVerifier.sol
isaacdecoded: Good point, indeed 🤔 Also, I'm wondering if it's correct to emit the `NewContextSet` event here since it will subsequently trigger the key resharing process, and the set of signers is the same in this case. So maybe we should emit a different event, e.g., `NewContextThreshold`.

I'll summon @jot2re and @eudelins-zama 🙏

host-contracts/contracts/KMSVerifier.sol
isaacdecoded: Yes, I think the same, we shouldn't receive a null value for the extraData argument. But maybe still worth having it checked, otherwise this function could panic at line 474 when trying to initialize the version variable, no?

host-contracts/contracts/KMSVerifier.sol
jot2re: There is no issue in doing a resharing for a context change even when the set of parties remain the same. For now, we accepted to do that whenever we change a context, even if it is not strictly needed. This is because it enhances the security and is a pretty cheap operation (i.e. takes a couple of minutes).
So when changing the threshold would indeed mean adding a new context.

obatirou: Thank you for those info and comments 
In the end, if not mistaken, @isaacdecoded we decided to remove `setThreshold` which is a legacy function and `defineNewContext` should be used instead ?
Is there a realistic scenario where a threshold should be changed without updating the parties for the current context?

jatZama: In any case, if you decide to keep `setThreshold`, then since it is considered a new context, you should change current implementation to bump the `currentKmsContextId`.

host-contracts/contracts/KMSVerifier.sol
jatZama: For the sake of calrity, when you say :

>  we shouldn't receive a null value

 You meant an empty array value right? Because AFAIU we currently always send a "null single byte" value , ie extraData=0x00 (length is 1)

host-contracts/contracts/KMSVerifier.sol
obatirou: I removed it. It will be clearer to only use `defineNewContext` and avoid confusion

host-contracts/contracts/KMSVerifier.sol
isaacdecoded: Yes, I meant an empty byte array. Probably you're right, and I'm thinking too YAGNI about this validation. I'll also remove it from the gateway-contract PR if you agree on this too @obatirou. Let me know.

host-contracts/contracts/KMSVerifier.sol
obatirou: Done with [cf8b36e](https://github.com/zama-ai/fhevm/pull/2028/commits/cf8b36efbc2cb27c64731286b0424fa5d5248f88)

host-contracts/contracts/KMSVerifier.sol
obatirou: Done with [cf8b36e](https://github.com/zama-ai/fhevm/pull/2028/commits/cf8b36efbc2cb27c64731286b0424fa5d5248f88)

host-contracts/contracts/KMSVerifier.sol
obatirou: Done with [1d2d7e9](https://github.com/zama-ai/fhevm/pull/2028/commits/1d2d7e94ba8de372e8a508b952925b49f6030572)

host-contracts/contracts/KMSVerifier.sol
obatirou: If we can guarantee that we cannot have an empty byte array sure we can remove it.
But note that `_extractKmsContextId`  is used by user facing functions `getContextSignersAndThresholdFromExtraData` and `verifyDecryptionEIP712KMSSignatures`.
They should be called by FHE lib but could also be used directly.
IMO, this is a cheap check that ensure a correct returned error in case of misusage/bug.
Let me know if you want to remove it. I do not have strong opinion on this.

isaacdecoded: > this is a cheap check that ensure a correct returned error in case of misusage/bug

I also agree on this.

obatirou: let's keep it then

obatirou: Closing thread
if needed you can reopen

isaacdecoded (approved):
LGTM 👍

obatirou:
Just for posterity: `isValidKmsContext` was removed following the fact that Connector will consider a new context activated as soon as it sees the request and not rely on a KMSVerifier call for verification
* https://github.com/zama-ai/tech-spec/pull/389/
* https://zama-ai.slack.com/archives/C0A8BJPD5UG/p1772201562185559?thread_ts=1772121420.401659&cid=C0A8BJPD5UG

PR #2032, opened 2026-02-27
feat(kms-connector): context-aware extraData handling for decryption

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/1071

## Summary

The two first commits are just some refactoring/cleaning a bit the way for the upcoming work.

The other commits add KMS context validation to the decryption pipeline. Before sending a decryption request to KMS Core, the kms-connector now extracts a context_id from the request's extraData field, checks it against a local DB table, and rejects requests with unknown or invalidated contexts.

## Key changes
1. New `kms_context` DB table (`migrations/20260226161117_add_kms_context.sql`) — stores context IDs with an `is_valid` boolean flag.
2. New `EthereumListener` (`gw-listener/src/core/ethereum.rs`) — connects to the Ethereum L1 `KMSVerifier` contract, reads the current context ID via `getCurrentKmsContextId()`, and stores it in the DB. Runs alongside the existing `GatewayListener` inside a new parent `EventListener` struct. No live subscription yet (deferred to RFC-005).
3. `extraData` parsing (`utils/src/types/extra_data.rs`) — versioned format: byte 0 = version (0x01), bytes 1-32 = context ID (big-endian U256). Empty or `[0x00]` treated as no-context for backward compat

### DIFF
--- a/kms-connector/crates/utils/src/types/extra_data.rs
+++ b/kms-connector/crates/utils/src/types/extra_data.rs
@@ -1,0 +1,103 @@
+use alloy::primitives::U256;
+use anyhow::anyhow;
+
+/// The expected length of the `extra_data` bytes.
+const EXPECTED_EXTRA_DATA_LENGTH: usize = 33;
+
+/// The version of the `extra_data` format.
+const EXTRA_DATA_VERSION: u8 = 0x01;
+
+/// Parses the `extra_data` bytes to extract an optional context ID.
+///
+/// Format (v1):
+/// - Byte 0: version (`0x01`)
+/// - Bytes 1..33: context ID (32 bytes, big-endian U256)
+/// - Bytes 33..: optional additional data (ignored)
+///
+/// Special cases for backward compatibility:
+/// - Empty `extra_data` → `Ok(None)`
+/// - `extra_data == [0x00]` → `Ok(None)` (legacy sentinel)
+pub fn parse_extra_data_context(extra_data: &[u8]) -> anyhow::Result<Option<U256>> {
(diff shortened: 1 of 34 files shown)


eudelins-zama:
@claude /pr-review

eudelins-zama:
@claude /pr-review

eudelins-zama:
@claude /pr-review

kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -0,0 +1,103 @@
+use alloy::primitives::U256;
+use anyhow::anyhow;
+
+/// The expected length of the `extra_data` bytes.
+const EXPECTED_EXTRA_DATA_LENGTH: usize = 33;
+
+/// The version of the `extra_data` format.
+const EXTRA_DATA_VERSION: u8 = 0x01;
```
maksymsur, line 8: shall these values be configurable as env vars for better transition to other versions?

maksymsur (approved):
LGTM!

In addition to comment above - can you also consider adding in-memory cache for the active context? Right now every decryption does a DB lookup. Under real load this may become a real hot path.

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: We have already plenty of configurable var, that we end up never using.

I think it's better to update these "hardcoded" value across time, to avoid having to sync with every other partners running the protocol to tell them to update their config.
They would just upgrade the software version, and would be easier for them IMO

PR #2036, opened 2026-03-02
fix(coprocessor): force compress/decompress for all ciphertexts

antoniupop:
This is needed to ensure ciphertexts that could be fetched from DB are identical to those re-used within a tfhe-worker partition for consensus across coprocessors.

### DIFF
--- a/coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
+++ b/coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
@@ -251,21 +251,6 @@ impl<'a> Scheduler<'a> {
     }
 }
 
-fn decompress_transaction_inputs(
-    inputs: &mut HashMap<Handle, Option<DFGTxInput>>,
-    gpu_idx: usize,
-) -> Result<usize> {
-    let mut count = 0;
-    for txinput in inputs.values_mut() {
-        if let Some(DFGTxInput::Compressed(((t, c), allowed))) = txinput {
-            let decomp = SupportedFheCiphertexts::decompress(*t, c, gpu_idx)?;
-            *txinput = Some(DFGTxInput::Value((decomp, *allowed)));
-            count += 1;
-        }
-    }
-    Ok(count)
-}
-
 fn re_randomise_operation_inputs(
     cts: &mut [SupportedFheCiphertexts],
     opcode: i32,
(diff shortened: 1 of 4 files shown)


coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -465,12 +419,15 @@ fn try_execute_node(
...
     let mut cts = Vec::with_capacity(node.inputs.len());
     for i in std::mem::take(&mut node.inputs) {
-        if let DFGTaskInput::Value(i) = i {
-            cts.push(i);
-        } else {
-            // That should not be possible as we called the checker.
-            error!(target: "scheduler", { handle = ?hex::encode(&node.result_handle) }, "Computation missing inputs");
-            return Err(SchedulerError::MissingInputs.into());
+        match i {
+            DFGTaskInput::Value(v) => cts.push(v),
```
Eikix: NB for when we want to merge on main -> I wonder if for later we should have some kind of way to force `DFGTaskInput::Value` here to be an "immediate" (in some way, make sure it's always a Scalar in the current code structure)

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -491,26 +448,23 @@ fn try_execute_node(
...
-        is_allowed,
         gpu_idx,
         transaction_id,
         tracer,
         ctx,
     ))
 }
 
-type OpResult = Result<(SupportedFheCiphertexts, Option<(i16, Vec<u8>)>)>;
+type OpResult = Result<(i16, Vec<u8>)>;
```
Eikix: NB main merge -> I wonder if (i16, Vec<u8>) would benefit to become a struct CompressedCiphertext for more readability

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -420,7 +420,13 @@ fn try_execute_node(
     let mut cts = Vec::with_capacity(node.inputs.len());
     for i in std::mem::take(&mut node.inputs) {
         match i {
-            DFGTaskInput::Value(v) => cts.push(v),
+            DFGTaskInput::Value(v) => {
+                if matches!(v, SupportedFheCiphertexts::Scalar(_)) {
+                    cts.push(v);
+                } else {
+                    error!(target: "scheduler", { handle = ?hex::encode(&node.result_handle) }, "Consensus risk: non-scalar uncompressed ciphertext");
```
Eikix: do we return an error or just log & skip?

antoniupop: I've hesitated on this one - and I see it's wrong, it should still push the CT. Technically if one slips through it's a consensus risk rather than an error (it's not missing input) so we should allow execution rather than give up I think - wdyt?
In any case well spotted, this is wrong as is, should either return error or push CT.

Eikix:
fixes https://github.com/zama-ai/fhevm-internal/issues/1086
and https://github.com/zama-ai/fhevm-internal/issues/1087

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
Eikix: Yeah we can still push it. I wonder if we can enforce it via the type system when we create DFGTaskInput::Value in the first place.

antoniupop:
@claude /pr-review

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -460,12 +413,25 @@ fn try_execute_node(
...
+			   "Consensus risk: non-scalar uncompressed ciphertext");
+                }
+                cts.push(v);
+            }
+            DFGTaskInput::Compressed(cct) => {
+                cts.push(SupportedFheCiphertexts::decompress(
+                    cct.ct_type,
+                    &cct.ct_bytes,
+                    gpu_idx,
+                )?);
```
Eikix: We might want to do something like:
```
DFGTaskInput::Compressed(cct) => {
    let decompressed = SupportedFheCiphertexts::decompress(
        cct.ct_type,
        &cct.ct_bytes,
        gpu_idx,
    )
    .map_err(|e| {
        error!(
            target: "scheduler",
            { handle = ?hex::encode(&node.result_handle), ct_type = cct.ct_type, error = ?e },
            "Error while decompressing op input"
        );
        telemetry::set_current_span_error(&e);
        SchedulerError::DecompressionError
    })?;

    cts.push(decompressed);
}
```

Eikix (approved):
lgtm

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

PR #2066, opened 2026-03-09
feat(test-suite): replace fhevm bash cli with Bun runtime

Eikix:
Closes https://github.com/zama-ai/fhevm-internal/issues/1042

## Summary
Replace the legacy `test-suite/fhevm/fhevm-cli` bash workflow with a Bun-native TypeScript CLI.

The new CLI makes target selection, runtime generation, resume, local builds, upgrade flows, and multicopro scenarios explicit and reproducible instead of relying on bash defaults and workflow-specific behavior.

## User-facing changes
- boot explicit targets:
  - `latest-supported`
  - `latest-main`
  - `sha --sha <git-sha>`
  - `devnet`
  - `testnet`
  - `mainnet`
- treat `latest-supported` as a tracked maintained bundle profile:
  - `test-suite/fhevm/profiles/latest-supported.json`
- keep `latest-main` as the moving mainline baseline:
  - newest complete repo-owned main SHA bundle at or after the simple-ACL floor
  - non-repo companions come from the mainline defaults in `src/presets.ts`
- lock every resolved bundle under `.fhevm/locks`
- persist runtime state under `.fhevm/state.json`
- support full-group local overrides for:
  - `coprocessor`
  - `kms-connector`
  - `gateway-contracts`
  - `host-contracts`
  - `test-suite`
- support per-service overrides only for runtime groups:
  - `coprocessor`
  - `kms-connector`
  - `test-suite`
- support `--build` as the full local-workspace path:
  - normal stacks: full local workspace, including coprocessor
  - scenario runs: full local workspace except coprocessor, which stays scenario-driven
- support `upgrade <group>` for active runtime overrides:
  - `coprocessor`
  - `kms-connector`
  - `test-suite`
- replace the old multicopro CLI flags with explicit scenario files:
  - `fhevm-cli up --scenario ./scenarios/two-of-two.yaml`
- keep `--override coprocessor` as the fast local-dev shorthand
- keep `deploy` as an alias for `up`
- remove `doctor` in favor of `up --dry-run`

## Architecture
Tracked inputs are now split by ownership:
- `test-suite/fhevm/docker-compose/`: base compose truth
- `test-suite/fhevm/templates/`: tracked templates rendered by the CLI
- `test-suite/fhevm/static/`: tracked config mounted as-is
- `test-suite/fhevm/scenarios/`: checked-in consensus/matrix inputs
- `test-suite/fhevm/profiles/`: tracked maintained bundle profiles

Generated runtime output lives under `.fhevm/`:
- `.fhevm/env/`
- `.fhevm/config/`
- `.fhevm/compose/`
- `.fhevm/addresses/`
- `.fhevm/locks/`
- `.fhevm/state.json`

The runtime flow is now:
1. resolve the target/profile bundle
2. apply version env overrides
3. resolve the coprocessor scenario or `--override coprocessor` shorthand
4. build a normalized runtime plan
5. render env/config/compose artifacts
6. run the step-based pipeline with resume support

The code is split around that model:
- `src/resolve.ts`
- `src/scenario.ts`
- `src/runtime-plan.ts`
- `src/render-env.ts`
- `src/render-config.ts`
- `src/render-compose.ts`
- `src/pipeline.ts`

Tracked compose files are the default runtime truth. `.fhevm/compose/` now contains selective overrides instead of full generated replacements for every component. Coprocessor topology remains the structural exception.

## Compatibility and targeting
- `latest-main` and `sha` are hard-cut to the simple-ACL era and fail before the supported floor
- exact `sha` resolution requires a complete repo-owned image set and a commit that exists on `main`
- `latest-main` resolves repo-owned services from `main`; `src/presets.ts` supplies the non-repo companion baseline for the common mainline path
- `src/compat.ts` is kept focused on cross-era protection:
  - known-invalid bundle pairs
  - legacy runtime shims
  - target resolution anchors

## CI contract
- reusable e2e now always boots `latest-main` with the checked-in `two-of-two` scenario
- manual workflow dispatch defaults to `build=true`, so a human dispatch naturally validates the checked out branch from source
- merge-queue orchestration passes `build=false` explicitly and overlays repo-owned `*_VERSION` env vars only for components whose Docker build succeeded
- skipped component builds leave those overrides unset, so they naturally stay on the `latest-main` baseline
- merge queue fails before e2e if a required repo-owned build reports a non-skipped failure
- operators e2e uses the same `latest-main` baseline and scenario model

## Quality and reliability improvements
- replace the monolithic bash runtime with typed services and tests
- make runtime artifact regeneration deterministic for resume, upgrade, and teardown
- make Docker/preflight failures surface clearly instead of degrading into misleading output
- fix lock-file boots, exact SHA handling, runtime error rendering, and stale state handling
- add scenario parsing/validation for consensus workflows
- keep state persistence normalized by deriving topology from scenario on load instead of persisting both authorities
- port the ciphertext-drift path off bash into shared TS/Bun code

## Validation
Verified on the current branch state:
- `bun run check`
- `bun test` (`322` tests)
- repeated local QA and live-stack validation across:
  - baseline lifecycle
  - resume / regenerate paths
  - local override flows
  - full local-workspace `--build` dry-runs
  - `upgrade coprocessor`
  - scenario-driven multicopro boot paths
  - `latest-main` dry-runs and scenario dry-runs

## Notes
- `latest-supported` is now a tracked maintained profile, not a dynamic “latest release” resolver.
- `latest-main` is the moving baseline for PR validation and local mainline testing.
- `--scenario` remains the explicit consensus/matrix interface.
- `--override coprocessor` remains the fast local e2e path.
- `--build` is the full local-workspace path.


Eikix:
@claude /pr-review

coprocessor/fhevm-engine/Dockerfile.workspace
```diff
@@ -62,6 +62,7 @@ RUN npm ci && \
 FROM ghcr.io/zama-ai/fhevm/gci/rust-glibc:1.91.0 AS builder
 
 ARG CARGO_PROFILE=release
+ARG BUILD_ID=unknown
```
Eikix, line 65: This is probably used in many places, can .git copy be a default and if it's not available then we take it from an argument?

eudelins-zama (commented):
Just a small note to say it would be nice to merge this [feature branch](https://github.com/zama-ai/fhevm/pull/2072) first to avoid handling conflicts with this one 😇

Eikix:
@claude /pr-review

PanGan21:
@cla-bot check

Eikix:
Opened a sacrificial draft PR to exercise the merge-queue-style E2E orchestration on the same head commit as this PR: #2087

It is draft-only and should not be merged. Once we have the workflow signal, we can close it.

.github/workflows/test-suite-e2e-operators-tests.yml
```diff
@@ -142,16 +151,16 @@ jobs:
 
       - name: Show logs on test failure
         working-directory: test-suite/fhevm
-        if: always()
+        if: failure()
         run: |
           echo "::group::Relayer Logs"
-          ./fhevm-cli logs relayer
+          timeout 20 ./fhevm-cli logs relayer || true
```
Eikix: why timeout 20 here? do we expect it to stop on its own fast or do we expect the CI to always wait until timeout?

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -242,28 +288,28 @@ jobs:
 
       - name: Show logs on test failure
         working-directory: test-suite/fhevm
-        if: always()
+        if: failure()
         run: |
           echo "::group::Relayer Logs"
-          ./fhevm-cli logs fhevm-relayer
+          timeout 20 ./fhevm-cli logs fhevm-relayer || true
```
Eikix: same question here

test-suite/fhevm/docker-compose/coprocessor-docker-compose.yml
```diff
@@ -24,7 +24,7 @@ services:
       dockerfile: coprocessor/fhevm-engine/Dockerfile.workspace
       target: db-migration
       args:
-        CARGO_PROFILE: ${FHEVM_CARGO_PROFILE:-release}
+        CARGO_PROFILE: release
```
Eikix: what is this cargo_profile arg? what is its function

Eikix:
### Code review

Found 2 issues:

1. **Compat flags with literal values are silently dropped for pre-v0.12.0 transaction-sender.** The `legacy-tx-sender-gateway-flags` profile in `compat.ts` uses literal numeric strings as the second tuple element (e.g., `["--delegation-fallback-polling", "30"]`), but `applyInstanceAdjustments` in `artifacts.ts` treats the second element as an env var key and looks it up via `envVars[envKey]`. Since `envVars["30"]`, `envVars["100000"]`, and `envVars["2"]` are always `undefined`, the condition is falsy and these three flags are never injected. Only `--multichain-acl-address` (backed by real env var `MULTICHAIN_ACL_ADDRESS`) actually works. This means pre-v0.12.0 `transaction-sender` silently starts without the delegation fallback polling, max retry, and nonce-error retry tuning.

https://github.com/zama-ai/fhevm/blob/bb78ce0d1417fb26017605e0d26b4f9bb78cf22f/test-suite/fhevm/src/compat.ts#L32-L42

https://github.com/zama-ai/fhevm/blob/bb78ce0d1417fb26017605e0d26b4f9bb78cf22f/test-suite/fhevm/src/artifacts.ts#L219-L223

2. **E2E tests workflow regresses to single-coprocessor mode.** Commit `a748a31e` on main explicitly added `--coprocessors 2 --coprocessor-threshold 2` to the e2e-tests workflow for 2-of-2 consensus testing. This PR drops those flags — the "Resolve CLI up args" step only writes `--target latest-release` to the args file. The operators workflow correctly preserves `--coprocessors 2 --threshold 2` (line 140), but the e2e-tests workflow does not, silently reverting to single-coprocessor mode.

https://github.com/zama-ai/fhevm/blob/bb78ce0d1417fb26017605e0d26b4f9bb78cf22f/.github/workflows/test-suite-e2e-tests.yml#L155-L165

Compare with operators workflow which correctly includes the flags:

https://github.com/zama-ai/fhevm/blob/bb78ce0d1417fb26017605e0d26b4f9bb78cf22f/.github/workflows/test-suite-e2e-operators-tests.yml#L137-L143

Eikix:
### Code review — re-review after fix commit `7854605`

**Previous review found 4 P1s and 12 P2s. Fix commit addressed 14 of 16, but introduced 1 new P1.**

---

#### New finding

**P1 — `workspace_all` unbound variable crashes every `registry`-mode CI run.** The "Resolve CLI up args" step has `set -euo pipefail`. `workspace_all` is only initialized inside the `workspace)` case branch (line 185). In the `registry)` branch it is never set. Line 206 then references `$workspace_all` without a `:-` default. With `-u` in effect, bash exits with `workspace_all: unbound variable`. Since the orchestrate workflow passes `image_mode: registry` (line 165 of `test-suite-orchestrate-e2e-tests.yml`), **every merge-queue CI run will fail at this step**.

Fix: add `workspace_all=false` before the `case` block.

https://github.com/zama-ai/fhevm/blob/7854605cc9c278c26f17dd8187a2e875bc724f51/.github/workflows/test-suite-e2e-tests.yml#L168-L210

---

#### Still open from previous review

**P1 — Schema-drift guard still only fires on partial (per-service) overrides.** `partialSchemaOverrides` at runtime.ts:319-321 still filters on `!!item.services?.length`. Full-group overrides (`--override coprocessor`, `--override all`) have `services: undefined` and bypass the guard entirely. The dangerous case (local code builds all services including db-migration) remains unchecked.

https://github.com/zama-ai/fhevm/blob/7854605cc9c278c26f17dd8187a2e875bc724f51/test-suite/fhevm/src/runtime.ts#L319-L333

**P2 — `probeBootstrap` still swallows permanent infrastructure failures as transient.** `ensureMaterialUrl` exhaustion (MinIO down) is caught by `probeBootstrap`'s outer catch and converted to `return false`, causing 4+ minutes of retries before a generic timeout. (runtime.ts:593-605, unchanged)

**P2 — `runDown` throw inside `runClean` still abandons `STATE_DIR` removal.** No try/catch was added. A single `compose down` failure leaves the workspace permanently dirty. (runtime.ts:1295-1317, unchanged)

**P2 — `gateway-set-relayer-mocked-payment` step still has no behavioral test.** The ordering fix from "approve mocked payment after deploy" is only covered by the env-value assertion, not the execution sequence.

---

#### Fixed (14 items confirmed)

| Previous finding | Status |
|---|---|
| P1-1: Compat flag literal values dropped | Fixed — discriminated union `CompatArgValue` with `{ env }` / `{ value }` |
| P1-2: Missing `--coprocessors 2 --threshold 2` | Fixed — flags in `printf` at line 173 |
| P1-4: Relayer v0.9.0 for modern targets | Fixed — `MODERN_COMPANIONS` with `sha-29b0750` |
| P2-5: Resume silently discards overrides | Fixed — `ensureResumeOptions` rejects conflicts |
| P2-6: `coprocessor2-*` container names | Fixed — changed to `coprocessor1-*` |
| P2-9: simpleAclFloor 1000-commit window | Partially fixed — window increased to 5000 |
| P2-10: Compat rules using wrong version key | Fixed — per-rule `versionKey` field |
| P2-11: Relayer config port 5433 | Fixed — changed to 5432 |
| P2-12: Operators missing primary containers | Fixed — all containers now logged |
| P2-13: `chain_id` hardcoded | Fixed — reads from env with fallback |
| P2-14: kms-connector schema guard untested | Fixed — test added |
| P2-15: rewriteRelayerConfig modern path untested | Fixed — test added |

---

#### Verdict: not merge-ready

The new unbound-variable bug (new P1) will break every orchestrate-dispatched CI run. The schema guard inversion (carried P1) leaves full-group overrides unchecked. Both need fixes before merge.

Eikix:
### Code review

Found 3 high-confidence issues and 7 notable concerns.

---

**High confidence (score >= 80):**

**1.** `COPROCESSOR_TX_SENDER_VERSION` extracted from wrong GitOps doc — uses `coprocessorGateway` but tx-sender is a coprocessor-side worker, not a gateway component. The sibling entries (`TFHE_WORKER`, `ZKPROOF_WORKER`, `SNS_WORKER`) all use `coprocessorWorkers`. If the real GitOps YAML for devnet/testnet/mainnet doesn't include tx-sender in the gateway values file, `findTag` throws and aborts resolution for all network targets.

https://github.com/zama-ai/fhevm/blob/512978db0a767ed9006da894c45c23f1766ed95b/test-suite/fhevm/src/resolve.ts#L248-L253

**2.** CI "Resolve CLI up args" step (52 lines of bash) implements business logic the CLI should own: relayer version coercion based on `workspace_all`/`STACK_ERA`, comma-split-and-trim for `WORKSPACE_OVERRIDE`, and three separate conditional arms for `RELAYER_VERSION`/`RELAYER_MIGRATE_VERSION` injection. The compat matrix knowledge is split between `compat.ts` (defines the pin) and this workflow step (decides when to apply it). A future developer updating the compat matrix has no way to know the workflow has its own triggering conditions. The orchestrate workflow also adds `fetch-depth: 0`, 3 extra job steps (Setup Bun, Install CLI deps, Read compat anchors), and JavaScript git-ancestor logic for Simple ACL — all net-new CI functionality, not a bash-to-Bun translation.

https://github.com/zama-ai/fhevm/blob/512978db0a767ed9006da894c45c23f1766ed95b/.github/workflows/test-suite-e2e-tests.yml#L165-L216

**3.** `logs` command changed from `docker logs <container>` (dumps and exits) to `docker logs --follow --tail 200` (follows indefinitely). Any script piping `fhevm-cli logs X | grep Y` will hang. The old bash CLI ran `docker logs` without `--follow`.

https://github.com/zama-ai/fhevm/blob/512978db0a767ed9006da894c45c23f1766ed95b/test-suite/fhevm/src/commands/logs.ts#L56-L63

---

**Notable concerns (score 75, borderline but worth flagging):**

**4.** `delegatedUserDecryption.ts` + `utils.ts` changes add a new `awaitDelegatedDecrypt` polling helper — a standalone flakiness fix completely unrelated to the CLI rewrite. Should be its own PR for bisectability.

**5.** `ensureResumeOptions` in `commands/up.ts:50-54` unconditionally rejects `--resume` when any `--override` is present, instead of comparing against `state.overrides`. A user who ran `up --override coprocessor` cannot resume with `--resume --override coprocessor` even though the override matches.

**6.** `pipeline.ts:625-634` mutates `state.discovery.actualFheKeyId`/`actualCrsKeyId` before the "predicted ids drifted" error check. If the check fails, the corrupted state is already in memory and gets persisted by subsequent `markStep`, poisoning `--resume` runs.

**7.** `pipeline.ts:844-846` early-exit bootstrap check uses `Effect.catchAll(() => Effect.succeed(false))`, swallowing real `MinioError`/`RpcError` failures as "bootstrap not done" — can trigger redundant keygen/crsgen.

**8.** Operators test profile lost `--parallel` flag. Old `run-tests.sh` passed `--parallel` for operators; new `commands/test.ts` does not. Sequential execution increases CI time and may expose ordering bugs.

**9.** `Dockerfile.workspace` `BUILD_ID` will always be `"unknown"` — nothing passes `--build-arg BUILD_ID=...`. Also, `CARGO_PROFILE` override removed from compose files, so `--local` Cargo profile optimization no longer works.

**10.** `snapshot_logs()` bash function copy-pasted between both CI workflows with hardcoded container names that bypass the CLI's `logs` command. If the CLI changes naming, log dumps silently produce empty output (`|| true` on every call).

Eikix:
### Effect-TS Best Practices Review

Reviewed against Effect 3.19.19 conventions.

**What's done well:** 13 specific `Data.TaggedError` subtypes, `Effect.gen` + `return yield*` used consistently, `Effect.log` for structured logging, clean `Layer` composition with transitive test stubs, proper `Schedule` composition for retries, appropriate concurrency annotations.

---

**HIGH — `throw` inside `Effect.gen` (creates defects, bypasses typed error channel)**

| File | Line | Code |
|------|------|------|
| `contract-task.ts` | 22 | `throw new Error("Stack is not running...")` |
| `pipeline.ts` | 151-184 | `validateDiscovery()` uses `throw` inside Effect.gen callers |
| `pipeline.ts` | 516-517 | `resolveUpgradePlan()` uses `throw` |
| `EnvWriter.ts` | 48-56 | `resolveEnvMap()` uses `throw` |

These should be `Effect.fail(new PreflightError({...}))` or similar. Pure helpers called outside Effect are less concerning, but `validateDiscovery` and `resolveEnvMap` are called within `Effect.gen` blocks.

**HIGH — Plain `Error` in `Effect.fail` (loses type info, forces `unknown` error channel)**

| File | Line |
|------|------|
| `pipeline.ts` | 297, 503, 632 |
| `down.ts` | 29 |

Each `Effect.fail(new Error(...))` escapes into the `unknown` error channel, forcing `Effect.Effect<void, unknown, never>` at the top level. These should use specific `TaggedError` subtypes.

**MEDIUM — `Context.Tag` everywhere instead of `Effect.Service`**

All 9 services use `Context.Tag` + static `Live` layer. `Effect.Service` with `accessors: true` and `dependencies: [...]` would eliminate manual `Layer.provide` wiring in `layers.ts`, generate accessors automatically, and bundle a `Default` layer per service. `Context.Tag` is fine for pure infrastructure (`CommandRunner`), but business logic services (`ImageBuilder`, `EnvWriter`, `ContainerProbe`) would benefit.

**MEDIUM — No `Effect.fn` for tracing**

No function uses `Effect.fn("ServiceName.method")`. While tracing isn't critical for a CLI, `Effect.fn` also provides better stack traces and call-site capture for debugging.

**LOW — String literal errors as retry sentinels**

`ContainerProbe` and `MinioClient` use `"not-ready" as const`. A `class NotReady extends Data.TaggedError("NotReady")<{}>` would integrate better with `catchTag`.

**LOW — `as` type assertions in CLI parsing**

`cli.ts` has many `as string`, `as boolean` casts from `node:util/parseArgs`. Hard to avoid without a Schema-based parser.

---

| Category | Verdict |
|----------|---------|
| Error types | Good specific TaggedErrors, but 4 sites use plain `Error` in `fail` |
| Service pattern | Older `Context.Tag` style; `Effect.Service` would modernize |
| Generators | Excellent — `Effect.gen` + `return yield*` everywhere |
| Layer composition | Solid manual wiring; `Effect.Service` would simplify |
| Logging | Good — `Effect.log` in Effect code, `console.*` only at boundary |
| Retry/scheduling | Well-implemented with `Schedule` composition |
| Critical violations | `throw` in gen (4 sites), plain `Error` in `Effect.fail` (4 sites) |

test-suite/fhevm/docker-compose/gateway-sc-docker-compose.yml
```diff
@@ -27,16 +27,16 @@ services:
       cache_to:
         - ${FHEVM_CACHE_TO_GATEWAY_SC_ADD_NETWORK:-type=gha,mode=max}
     env_file:
-      - ../env/staging/.env.gateway-sc.local
+      - ../../../.fhevm/env/gateway-sc.env
     command:
-      - npx hardhat task:addHostChainsToGatewayConfig
+      - npx hardhat task:addHostChainsToGatewayConfig --use-internal-proxy-address true
```
Eikix, line 32: why add  `--use-internal-proxy-address true` here and below

test-suite/fhevm/scenarios/one-local-outlier.yaml
```diff
@@ -0,0 +1,14 @@
...
+  threshold: 2
+instances:
+  - index: 1
+    source:
+      mode: local
+    env:
+      OTEL_SERVICE_NAME: coprocessor-1-local
+    args:
+      tfhe-worker:
+        - --coprocessor-fhe-threads=4
```
Eikix: why are we setting this flag?

test-suite/fhevm/scenarios/one-registry-outlier.yaml
```diff
@@ -0,0 +1,10 @@
+version: 1
+kind: coprocessor-consensus
+topology:
+  count: 3
+  threshold: 2
+instances:
+  - index: 2
+    source:
+      mode: registry
+      tag: abc1234
```
Eikix: what is this? looks like a dummy example but that's not good to keep.

test-suite/fhevm/scenarios/two-of-two.yaml
```diff
@@ -0,0 +1,5 @@
+version: 1
```
Eikix, line 1: good example 
why versioning btw?

test-suite/fhevm/docker-compose/gateway-sc-docker-compose.yml
Eikix: The flag is there to force the task to use the locally generated addresses from addresses/, not whatever comes from the normal deployment env.

Eikix:
@claude /pr-review

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -252,41 +261,36 @@ jobs:
 
       - name: Show logs on test failure
         working-directory: test-suite/fhevm
-        if: always()
+        if: failure()
```
Eikix, line 182: Go back to always()

.github/workflows/test-suite-e2e-operators-tests.yml
```diff
@@ -111,24 +140,15 @@ jobs:
...
-          GATEWAY_VERSION: ${{ inputs.gateway_version }}
-          HOST_LISTENER_VERSION: ${{ inputs.host_listener_version }}
-          GW_LISTENER_VERSION: ${{ inputs.gateway_listener_version }}
-          TX_SENDER_VERSION: ${{ inputs.tx_sender_version }}
-          TFHE_WORKER_VERSION: ${{ inputs.tfhe_worker_version }}
-          SNS_WORKER_VERSION: ${{ inputs.sns_worker_version }}
-          ZKPROOF_WORKER_VERSION: ${{ inputs.zkproof_worker_version }}
         run: |
-          ./fhevm-cli deploy --coprocessors 2 --coprocessor-threshold 2
+          ./fhevm-cli up --target latest-main --scenario ./scenarios/two-of-two.yaml
```
Eikix, line 151: can we use the short form here `./fhevm-cli up --target latest-main --scenario two-of-two`

.github/workflows/test-suite-e2e-operators-tests.yml
```diff
@@ -142,32 +162,39 @@ jobs:
 
       - name: Show logs on test failure
         working-directory: test-suite/fhevm
-        if: always()
+        if: failure()
```
Eikix, line 165: use always()

PR #2070, opened 2026-03-09
fix(test-suite): fix delegated user decrypt assertion

tawadaa:


test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
```diff
@@ -243,9 +243,7 @@ describe('Delegated user decryption', function () {
         this.signers.bob,
         privateKey,
         publicKey,
-      )
-    ).to.be.rejectedWith(
-      new RegExp(USER_DECRYPTION_NOT_DELEGATED_SELECTOR),
-    );
+      ),
+    ).to.be.rejectedWith('Execution reverted');
```
isaacdecoded, line 247: @tawadaa do you know why the error changed? Maybe something has changed on the Relayer side?

In this case, I think the [const declared at line 8](https://github.com/zama-ai/fhevm/pull/2070/changes#diff-ace0f1b1a0aeb31e50abb3b2d152a5572c399f51ddfe8d8e3a00090511bf1a8cR8) is not needed anymore.

test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
```diff
@@ -58,9 +58,9 @@ describe('Delegated user decryption', function () {
       );
     await delegateTx.wait();
 
-    // Wait for 15 blocks to ensure delegation is propagated by the coprocessor.
+    // Wait for 2 blocks to ensure delegation is propagated by the coprocessor.
     const currentBlock = await ethers.provider.getBlockNumber();
-    await waitForBlock(currentBlock + 15);
+    await waitForBlock(currentBlock + 2);
```
isaacdecoded: Do you know why this got decremented to 2 blocks? Is the simple ACL already working maybe?

tawadaa: With 15 blocks we hit the timeout for the last test, as we wait for 15 blocks twice.
Not sure for simple ACL, as this tested for v0.11.

eudelins-zama: I would appreciate if we could merge https://github.com/zama-ai/fhevm/pull/2072 first, to avoid dealing with conflicts on a big long living feature branch :innocent:

test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
tawadaa: Not sure if it changed, but the error we observed is an Execution reverted with and error 500

test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
manoranjith: Simple ACL changes are not yet active on main.

The error matching would again change with simple ACL, as the check will be have a new error label for Host ACL related errors.

manoranjith: I agree with @eudlins-zama that we should merge the feature branch sooner.

test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
manoranjith: The error from relayer didn't change. Relater sends only an abstracted information (labels) and a message intended for user. Revert selector is not there, since we do the ACL as boolena check (isDelegatedUserDecryptReady).

However, the message will also change in the incoming 0.12 release due to Host ACL feature.

manoranjith (approved):
LGTM.

isaacdecoded (approved):

test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
Eikix: Since simple ACL is on main, what is the next step on this PR?

tawadaa: @Eikix There were 2 main issues that this PR is resolving:
1. Enough timeout for real blockchain networks
2. Reject with the correct Error message

With simple ACL, we still need 1, but not sure if 2 was fixed

Eikix: So the PR is good to merge as is? or it needs rework?

tawadaa: No, no longer up to date

tawadaa:
Replaced by https://github.com/zama-ai/fhevm/pull/2250

PR #2086, opened 2026-03-10
feat(coprocessor): add cargo-llvm-cov code coverage to Rust CI workflows

PanGan21:
## Summary
- Adds `cargo-llvm-cov` (LLVM source-based instrumentation) to coprocessor Rust CI test workflow
- Coverage replaces `cargo test` (not in addition to it) so overhead is minimal (~10% compile time)
- Coverage results are printed to the GitHub Actions job summary tab (non-blocking, no artifacts uploaded)
- Posts comment on pr with test coverage of added/modified lines of code

Related: https://github.com/zama-ai/fhevm-internal/issues/1146


Eikix:
@claude /pr-review

Eikix:
@claude /pr-review

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -101,7 +107,26 @@ jobs:
       run: |
         docker run --rm -d -p 4566:4566 --name localstack localstack/localstack:latest
 
-    - name: Run tests
+    - name: Run tests with coverage
```
PanGan21, line 115: Do we want to run with coverage only when merging to main branch in order to speed up?

Eikix: I think it can be good to run just as a report for the reviewer and PR person (non-blocking during PR time) 
+ strong invitation to run it locally before creating a PR

and then yes it can run on main to give us visibility and create follow up issues if we have holes

PanGan21:
@claude /pr-review

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -101,7 +107,30 @@ jobs:
...
+      run: cargo llvm-cov clean --workspace --profile coverage
+      working-directory: coprocessor/fhevm-engine
+
+    - name: Run tests with coverage
+      run: |
+        DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor \
+        TEST_GLOBAL_LOCALSTACK=1 \
+        cargo llvm-cov --no-report --workspace --profile coverage -- --test-threads=1
+      working-directory: coprocessor/fhevm-engine
+
```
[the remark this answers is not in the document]
PanGan21: `--workspace` is not supported with `report` so I am resolving the comment. Here is the error:
`error: --workspace is specific to [test,nextest,nextest-archive,show-env,clean,no subcommand] and not supported for subcommand 'report'`

PanGan21:
@claude /pr-review

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -101,7 +107,36 @@ jobs:
...
       run: |
-        DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor TEST_GLOBAL_LOCALSTACK=1 cargo test --release -- --test-threads=1
+        if cargo llvm-cov report --profile coverage > /tmp/cov-report.txt 2>&1; then
+          REPORT=$(cat /tmp/cov-report.txt)
+        else
+          echo "cargo llvm-cov report failed:"
+          cat /tmp/cov-report.txt
+          REPORT=""
+        fi
+        {
```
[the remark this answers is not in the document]
PanGan21: `--workspace` is not supported with `report` so I am resolving the comment. Here is the error:
`error: --workspace is specific to [test,nextest,nextest-archive,show-env,clean,no subcommand] and not supported for subcommand 'report'`

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -101,7 +107,36 @@ jobs:
...
-    - name: Run tests
+    - name: Clean previous coverage data
+      run: cargo llvm-cov clean --workspace --profile coverage
+      working-directory: coprocessor/fhevm-engine
+
+    - name: Run tests with coverage
+      run: |
+        DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor \
+        TEST_GLOBAL_LOCALSTACK=1 \
+        cargo llvm-cov --no-report --workspace --profile coverage -- --test-threads=1
```
Eikix, line 119: what does the --no-report flag do? since we want to use the report in the next step

PanGan21: With --no-report, the files are kept, so the next step can call cargo llvm-cov report to generate the report from the existing data. We just need two steps to achieve this functionality on ci

PanGan21:
@claude /pr-review

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -53,7 +54,13 @@ jobs:
     - name: Setup Rust
       uses: dtolnay/rust-toolchain@a54c7afa936fefeb4456b2dd8068152669aa8203
```
rudy-6-4: Maybe using actions-rust-lang/setup-rust-toolchain to rely on rust-toolchain.toml .
Or https://github.com/dsherret/rust-toolchain-file which is a fork of dtolnay/rust-toolchain

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -101,7 +108,90 @@ jobs:
...
+    - name: Clean previous coverage data
+      run: cargo llvm-cov clean --workspace --profile coverage
+      working-directory: coprocessor/fhevm-engine
+
+    - name: Run tests with coverage
       run: |
-        DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor TEST_GLOBAL_LOCALSTACK=1 cargo test --release -- --test-threads=1
+        DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor \
+        TEST_GLOBAL_LOCALSTACK=1 \
+        cargo llvm-cov --no-report --workspace --profile coverage -- --test-threads=1
```
rudy-6-4, line 119: Is test-threads=1 still needed ? Test that need to be sequential are tagged specifically so it could work.
(I see it was there before)

rudy-6-4 (dismissed):

.github/workflows/coprocessor-cargo-tests.yml
PanGan21: Done! However for consistency we should consider applying it to the other workflow files as well

.github/workflows/coprocessor-cargo-tests.yml
PanGan21: I see that tests are explicitly tagged with #[serial(db)], as you mentioned. Do you happen to know why this was originally added?

I'm a bit concerned about removing --test-threads=1, because it could introduce race conditions if some tests that interact with the database are not tagged with #[serial(db)] even though they should be. I’d like to understand the original reason for this setup before changing it.

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -101,7 +108,90 @@ jobs:
...
+        pip install --quiet diff-cover
+        git fetch --unshallow origin
+        git fetch origin "${BASE_REF}:refs/remotes/origin/${BASE_REF}"
+        diff-cover /tmp/lcov.info \
+          --compare-branch="origin/${BASE_REF}" \
+          --diff-range-notation='...' \
+          --format "json:/tmp/diff-cover.json,markdown:/tmp/diff-cover-report.md" \
+          --fail-under 0 || true
+        if [ -f /tmp/diff-cover.json ]; then
+          DIFF_PCT=$(python3 -c "import json; d=json.load(open('/tmp/diff-cover.json')); print(f\"{d['total_percent_covered']:.1f}%\")" 2>/dev/null || echo "N/A")
```
Eikix: This ends up posting `100.0%` when the diff has no coverable lines. The current run does exactly that: `diff-cover` says `No lines with coverage information in this diff`, but `total_percent_covered` still becomes `100.0%` and the sticky comment reads as full diff coverage. I'd map the empty-diff case to `N/A` or skip the comment entirely.

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -101,7 +108,90 @@ jobs:
...
+            echo ''
+            cat /tmp/diff-cover-report.md
+            echo ''
+            echo '</details>'
+          fi
+        } | tee -a "$GITHUB_STEP_SUMMARY" > /tmp/coverage-comment.md
+
+    - name: Post coverage comment on PR
+      if: ${{ !cancelled() }}
+      uses: marocchino/sticky-pull-request-comment@773744901bac0e8cbb5a0dc842800d45e9b2b405 # v2.9.4
```
Eikix, line 195: This is on `pull_request`, so fork PRs on a public repo only get a read-only `GITHUB_TOKEN`. The sticky comment won't be able to write there, and depending on the action response that can fail the job after the tests already ran. I'd gate this to same-repo PRs or move the commenting bit into a safe `pull_request_target` job.

.github/workflows/coprocessor-cargo-tests.yml
PanGan21: Good catch. The Python script now checks `total_num_lines` from the diff-cover JSON and outputs N/A when it's 0 instead of 100.0%.

.github/workflows/coprocessor-cargo-tests.yml
PanGan21: Fixed. Added a check to see if condition so the sticky comment step is skipped entirely on fork PRs.

.github/workflows/coprocessor-cargo-tests.yml
PanGan21: I created this ticket to investigate this separately because it seems it requires more changes than just the ci:
https://github.com/zama-ai/fhevm-internal/issues/1156

Eikix: > I see that tests are explicitly tagged with #[serial(db)], as you mentioned. Do you happen to know why this was originally added?
> 
> I'm a bit concerned about removing --test-threads=1, because it could introduce race conditions if some tests that interact with the database are not tagged with #[serial(db)] even though they should be. I’d like to understand the original reason for this setup before changing it.

We can try it and see, normally serial tests are marked as such. Iirc, it's because some tests can cross pollute one another if they run at the same tiem with the same db

Eikix (approved):
gg!

PanGan21:
@Mergifyio queue

PR #2096, opened 2026-03-11
feat(coprocessor): drift detection in gw-listener

Eikix:
## Summary

- Add opt-in drift detection to `gw-listener` by polling `CiphertextCommits` events
- Fetch the expected coprocessor tx-sender set from `GatewayConfig` when drift detection is enabled
- Track per-handle submissions in memory until the handle is complete or stale
- Snapshot the expected sender set per handle so open handles are evaluated against the membership they started with
- Persist the earliest open `CiphertextCommits` block and rebuild open-handle state from chain logs after restart
- Detect multiple digest variants for a handle and log the sender breakdown
- Compare consensus digests against the local `ciphertext_digest` row, retrying later if local digests are not ready yet
- Alert if all expected coprocessors submitted but no consensus event was observed
- Alert if consensus was reached but some expected coprocessors never submitted within a grace window
- Metrics: `drift_detected`, `consensus_timeout`, `missing_submission`

## Scope Notes

- Restart recovery persists only one watermark: the earliest open `CiphertextCommits` block
- Open-handle state is rebuilt from chain logs on startup; it is not stored in Postgres as structured detector state
- Rebuild is silent while historical logs are replayed, then open handles are evaluated once after alerts are re-enabled
- `GatewayConfig` updates refresh the current expected sender set for new handles; existing open handles keep their original membership snapshot
- This PR does not pause processing, fetch canonical ciphertexts, or wipe/recompute state

## Test Plan

- [x] `SQLX_OFFLINE=true cargo build -p gw-listener --manifest-path coprocessor/fhevm-engine/Cargo.toml`
- [x] `SQLX_OFFLINE=true cargo test -p gw-listener --manifest-path coprocessor/fhevm-engine/Cargo.toml`
- [x] `./test-suite/fhevm/fhevm-cli test ciphertext-drift`
- [x] CI: `test-suite-e2e-tests` with `deploy-build=true`
- [ ] Manual: restart `gw-listener` while a handle is still open and verify tracking resumes from the persisted watermark
- [ ] Manual: verify `/metrics` exposes `drift_detected`, `consensus_timeout`, and `missing_submission`

## E2E Drift Scenario

- The e2e drift test uses the existing multi-coprocessor stack
- One coprocessor DB installs a temporary one-shot Postgres trigger on `ciphertext_digest`
- The trigger flips one byte in `ciphertext` the first time a row becomes fully ready
- `transaction-sender` then submits the already-polluted digest, which should trigger the Rust drift detector

closes https://github.com/zama-ai/fhevm-internal/issues/1147


coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,450 @@
...
+
+use crate::metrics::{
+    CONSENSUS_CONFIRMED_COUNTER, CONSENSUS_HANDLE_NOT_FOUND_COUNTER, DRIFT_DETECTED_COUNTER,
+    DRIFT_EARLY_WARNING_COUNTER,
+};
+
+use fhevm_gateway_bindings::ciphertext_commits::CiphertextCommits;
+
+struct PerCoprocessorDigest {
+    #[allow(dead_code)] // needed for M2 recovery
```
Eikix: Yagni. Implement it when needed.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,450 @@
...
+    drift_early_warning: u64,
+    consensus_handle_not_found: u64,
+}
+
+impl DriftDetector {
+    pub(crate) fn new() -> Self {
+        Self {
+            submissions: HashMap::new(),
+            drift_detected: 0,
+            consensus_confirmed: 0,
```
Eikix: Yagni

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,450 @@
...
+            drift_early_warning: 0,
```
Eikix: We can regroup drift_early_warning & drift_detected in my opinion. A drift is a drift, whether or not there is a consensus.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,450 @@
...
+            consensus_handle_not_found: 0,
```
Eikix: Lagging behind consensus is out of scope

Eikix:
@claude /pr-review

Eikix:
@claude /pr-review

Eikix:
@claude /pr-review

Eikix:
### Code review

Found 1 issue:

1. **Handle leak in `drift_detector.rs`**: `observe_submission` never calls `finish_if_complete`. When consensus arrives before all submissions, `try_check_local_consensus` sets `local_consensus_checked = true` but `finish_if_complete` returns early (line 507-508) because `submissions.len() != expected_senders.len()`. When late submissions arrive via `observe_submission`, nothing calls `finish_if_complete`. In `evict_stale`, the branch at line 372 hits `continue` without evicting when consensus is present, `local_consensus_checked` is true, and all submissions are in. These handles accumulate indefinitely in `open_handles`, and `earliest_open_block()` returns their stale `first_seen_block` forever, preventing the rebuild watermark from advancing — on every restart the listener replays from a stuck block.

https://github.com/zama-ai/fhevm/blob/84c4869a3f4c9719762eb2aad435ac369a233e86/coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs#L103-L174

https://github.com/zama-ai/fhevm/blob/84c4869a3f4c9719762eb2aad435ac369a233e86/coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs#L502-L517

https://github.com/zama-ai/fhevm/blob/84c4869a3f4c9719762eb2aad435ac369a233e86/coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs#L370-L373

charts/coprocessor/values.yaml
```diff
@@ -417,12 +417,31 @@ gwListener:
   replicas: 1
 
   env:
-    # =========================================================================
-    # NEW ENVIRONMENT VARIABLES
-    # =========================================================================
-    # The address of the KMSGeneration contract
+    - name: DATABASE_URL
```
Eikix, line 420: gw listener already had DB access, we're just adding it to the named env

Eikix:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance (no CLAUDE.md files in repo). Reviewed commits 84c4869a and 92f2b645 (changes since prior review).

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1198 @@
...
+#[derive(Default)]
+struct DeferredMetrics {
+    drift_detected: u64,
+    consensus_timeout: u64,
+    missing_submission: u64,
+}
+
+pub(crate) struct DriftDetector {
+    current_expected_senders: Vec<Address>,
+    open_handles: HashMap<FixedBytes<32>, HandleState>,
```
PanGan21: Not sure if we risk a memory leak when loading all open handles in memory. Of course it is bounded by the watermark block and the `drift_post_consensus_grace_blocks`, but in case of big load this can be risky.

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
```diff
@@ -300,6 +366,175 @@ impl<P: Provider<Ethereum> + Clone + 'static, A: AwsS3Interface + Clone + 'stati
...
+        }
+
+        info!(
+            from_block,
+            to_block, "Rebuilding drift detector from persisted watermark"
+        );
+        drift_detector.set_alerts_enabled(false);
+
+        let mut batch_from = from_block;
+        while batch_from <= to_block {
```
PanGan21, line 448: I think we don't have a test case that validates recovery with a persisted watermark that spans multiple batches. Maybe I am missing it but if it is there let me know
Example
earliest_open_ct_commits_block (the watermark) = block 1000
last_processed_block_num = block 2000
get_logs_block_batch_size = 100

Then logs are fetched:
Batch 1: blocks 1000-1099
Batch 2: blocks 1100-1199
Batch 3: blocks 1200-1299

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1198 @@
...
+                    log_index = ?context.log_index,
+                    sender = %event.coprocessorTxSender,
+                    previous_ciphertext_digest = %existing.digests.ciphertext_digest,
+                    previous_ciphertext128_digest = %existing.digests.ciphertext128_digest,
+                    new_ciphertext_digest = %digests.ciphertext_digest,
+                    new_ciphertext128_digest = %digests.ciphertext128_digest,
+                    "Same coprocessor submitted different digests for one handle"
+                );
+            }
+            return;
```
PanGan21, line 170: Should't we increment `drift_detected` here?

Eikix: I'm not sure since here it seems more like equivocation, a single coprocessor submitted different digests for the same handle
Not even sure what this case requires in terms of alerting

PanGan21: Under which situation can this happen? Or even can this happen at all?

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
Eikix: So the scaling is approximately:

  - 1k open handles: ~1-1.5 MB
  - 10k open handles: ~10-15 MB
  - 100k open handles: ~100-150 MB
  
Since activity is low on mainnet we can take this time to figure out the proper knob on the grace blocks. I added metrics to make sure we can choose the right cursor

Eikix:
@claude /pr-review

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
Eikix: I don't know. Normally it's not possible afaik

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1645 @@
...
+            context,
+            digests: DigestPair {
+                ciphertext_digest: event.ciphertextDigest,
+                ciphertext128_digest: event.snsCiphertextDigest,
+            },
+            senders: event.coprocessorTxSenders.clone(),
+        });
+        state.local_consensus_checked = false;
+        CONSENSUS_LATENCY_BLOCKS_HISTOGRAM
+            .observe(context.block_number.saturating_sub(state.first_seen_block) as f64);
```
Eikix: [AUTOMATED] **Bug (Medium):** `CONSENSUS_LATENCY_BLOCKS_HISTOGRAM` is observed unconditionally during rebuild replay.

During `rebuild_drift_detector`, `alerts_enabled` is `false` and historical events are replayed. But this histogram observation has no `alerts_enabled` guard, so every replayed `AddCiphertextMaterialConsensus` event pollutes the histogram with stale historical data.

Compare with the deferred counters (`drift_detected`, `consensus_timeout`, `missing_submission`) which are correctly gated on `alerts_enabled`, and with `try_check_local_consensus` which returns early at line 305 when `!alerts_enabled`. This histogram is the only metric that bypasses the rebuild guard.

After a restart with a large rebuild window, the histogram will be filled with historical block distances, skewing the distribution that operators use to tune `--drift-no-consensus-timeout-blocks`.

Suggested fix:
```rust
if self.alerts_enabled {
    CONSENSUS_LATENCY_BLOCKS_HISTOGRAM
        .observe(context.block_number.saturating_sub(state.first_seen_block) as f64);
}
```

Confidence: 90/100

test-suite/fhevm/scripts/run-ciphertext-drift-e2e.sh
```diff
@@ -0,0 +1,64 @@
...
+"${SCRIPT_DIR}/inject-coprocessor-drift.sh" "$FAULTY_INSTANCE_INDEX" > "$HANDLE_FILE" &
+injector_pid=$!
+
+test_exit=0
+docker exec \
+  -e GATEWAY_RPC_URL= \
+  "$TEST_CONTAINER" \
+  ./run-tests.sh -n staging -g "$GREP_PATTERN" || test_exit=$?
+
+wait "$injector_pid"
```
Eikix: [AUTOMATED] **Bug (Medium):** Under `set -euo pipefail`, `wait "$injector_pid"` propagates the background process's exit code. If `inject-coprocessor-drift.sh` times out (exits 1), `set -e` terminates the script here. Lines 52-53 (`injector_pid=""` and reading the handle file) never execute, and the test suite's exit code captured in `test_exit` is silently lost.

The operator sees a generic exit code 1 with no diagnostic output about whether the failure was the test suite or the injector.

Suggested fix:
```bash
injector_exit=0
wait "$injector_pid" || injector_exit=$?
injector_pid=""
```
Then check `injector_exit` explicitly before proceeding.

Confidence: 90/100

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1645 @@
...
+                consensus_variant_sender_count,
+                observed_variant_count = observed_variants.len(),
+                observed_variants = ?observed_variants,
+                source = "consensus",
+                "Drift detected: local digest does not match consensus"
+            );
+            self.deferred_metrics.drift_detected += 1;
+        }
+
+        if !self.alerts_enabled {
```
Eikix: [AUTOMATED] **Nit (Low):** During rebuild (`alerts_enabled=false`), `handle_consensus` calls `try_check_local_consensus` which executes the full DB query but then returns early here without setting `local_consensus_checked = true`. After rebuild completes and `refresh_pending_consensus_checks` runs, every handle with consensus is re-queried — doubling the DB load proportional to the number of consensus events during rebuild.

Consider either:
- Skipping the DB query entirely when `!alerts_enabled` (move the guard to the top of the function), or
- Setting `local_consensus_checked = true` even when alerts are disabled (the flag just means "we verified against the DB", not "we alerted")

The drift metric is NOT double-counted (the increment at line 302 is correctly gated on `alerts_enabled`), so this is purely an efficiency concern.

Confidence: 85/100

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1645 @@
...
+                    previous_ciphertext128_digest = %existing.digests.ciphertext128_digest,
+                    new_ciphertext_digest = %digests.ciphertext_digest,
+                    new_ciphertext128_digest = %digests.ciphertext128_digest,
+                    "Same coprocessor submitted different digests for one handle"
+                );
+            }
+            return;
+        }
+
+        state.submissions.push(Submission {
```
Eikix, line 173: [AUTOMATED] **Nit (Low):** `observe_submission` accepts submissions from any sender address, including addresses not in `expected_senders`. The completion checks in `finish_if_complete` (line 517) and `finalize_completed_without_consensus` (line 483) compare `submissions.len() == expected_senders.len()`. If an unexpected sender submits:
- With N-1 expected + 1 unexpected sender: handle is prematurely marked complete (one expected sender was never seen)
- With N expected + 1 unexpected: `len` is N+1, equality is permanently false, handle can only be evicted by timeout

In practice, the on-chain contract likely restricts who can call `AddCiphertextMaterial`, so this is defense-in-depth. Consider either filtering out unexpected senders or changing the completion check to count only submissions from `expected_senders`.

Confidence: 75/100

Eikix (commented):
[AUTOMATED]

## Code Review — Ciphertext Drift Detection

Overall this is a well-structured addition. The state machine design is sound, the batch-then-finalize pattern in `gw_listener.rs` correctly handles the consensus-arrives-in-same-batch case, and the deferred metrics flush ordering is correct. The test coverage is solid with 24 unit tests plus an E2E script.

Found 2 confirmed issues (medium), 1 efficiency concern (low), and 1 defense-in-depth suggestion (low). 6 other reported issues were investigated and confirmed as false positives.

**Checked for:** state machine correctness, security (timing attacks, submission ordering), SQL safety, metric correctness, error handling, edge cases, guideline compliance (telemetry style guide — no violations found).

charts/coprocessor/values.yaml
dartdart26: Is this variable used? If no, can we remove it?

dartdart26: Is this variable used? If not, can we remove it?

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
```diff
@@ -89,6 +89,32 @@ struct Conf {
...
+        long,
+        default_value_t = 50,
+        requires = "ciphertext_commits_address",
+        help = "Drop a handle and alert if no consensus arrives after this many blocks"
+    )]
+    drift_no_consensus_timeout_blocks: u64,
+
+    #[arg(
+        long,
+        default_value_t = 10,
```
dartdart26: As my previous comment, maybe this value should be bigger.

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
```diff
@@ -89,6 +89,32 @@ struct Conf {
...
+    #[arg(
+        long,
+        requires = "ciphertext_commits_address",
+        help = "GatewayConfig contract address used to fetch coprocessor tx-senders"
+    )]
+    gateway_config_address: Option<Address>,
+
+    #[arg(
+        long,
+        default_value_t = 50,
```
dartdart26: I would put a bigger default value as the current instantiation of the GW can have block times of 100ms, making 5 second a relatively small period of time.

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
```diff
@@ -156,6 +182,11 @@ async fn main() -> anyhow::Result<()> {
     else {
         anyhow::bail!("--host-chain-id or CHAIN_ID env var is missing.")
     };
+    if conf.ciphertext_commits_address.is_some() && conf.gateway_config_address.is_none() {
+        anyhow::bail!(
+            "--gateway-config-address is required when --ciphertext-commits-address is set."
+        );
+    }
```
dartdart26: Why don't we use `requires` from `clap` in this direction too?

coprocessor/fhevm-engine/gw-listener/src/metrics.rs
dartdart26, line 1: Could we put explanations of these in docs/metrics/metrics.md?

PanGan21 (dismissed):

coprocessor/fhevm-engine/gw-listener/src/metrics.rs
```diff
@@ -96,3 +96,54 @@ pub(crate) static KEY_DIGEST_MISMATCH_COUNTER: LazyLock<IntCounter> = LazyLock::
     )
     .unwrap()
 });
+
+pub(crate) static DRIFT_DETECTED_COUNTER: LazyLock<IntCounter> = LazyLock::new(|| {
+    register_int_counter!(
+        "coprocessor_gw_listener_drift_detected_counter",
+        "Number of detected drifts from peer submissions or consensus mismatch"
```
dartdart26: I am not sure I understand the "consensus mismatch" description here? What does it mean, why is it "or"?

dartdart26 (commented):
I will continue with the review as it is a lot of code.

Is the PR AI-generated? I want to know such that I try to verify everything in it.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1724 @@
...
+impl DriftDetector {
+    pub(crate) fn new(
+        expected_senders: Vec<Address>,
+        host_chain_id: ChainId,
+        no_consensus_timeout_blocks: u64,
+        post_consensus_grace_blocks: u64,
+    ) -> Self {
+        Self {
+            current_expected_senders: expected_senders,
+            open_handles: HashMap::new(),
```
dartdart26, line 111: Is there an upper bound on open handles? I guess there is an implicit one based on how many blocks we wait for consensus and how many deferred missing submissions?

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1724 @@
...
+        state.consensus = Some(ConsensusState {
+            context,
+            digests: DigestPair {
+                ciphertext_digest: event.ciphertextDigest,
+                ciphertext128_digest: event.snsCiphertextDigest,
+            },
+            senders: event.coprocessorTxSenders.clone(),
+        });
+        state.local_consensus_checked = false;
+        if self.alerts_enabled {
```
dartdart26: I was wondering - if alerts are not enabled, what would the DriftDetector do, just log? I am thinking of alerts are not enabled, do we need a DriftDetector instance at all?

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1724 @@
...
+pub(crate) struct EventContext {
+    pub(crate) block_number: u64,
+    pub(crate) block_hash: Option<B256>,
+    pub(crate) tx_hash: Option<B256>,
+    pub(crate) log_index: Option<u64>,
+}
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+struct DigestPair {
+    ciphertext_digest: FixedBytes<32>,
```
dartdart26: I think we should use a type here to not use the literal `32`. We use it elsewhere too and it would be better to have one type we can reuse and not specify 32 everywhere.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
dartdart26: If that is the case, we might want to put a comment or maybe be explicit.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1724 @@
...
+        }
+    }
+
+    pub(crate) fn flush_metrics(&mut self) {
+        if self.deferred_drift_detected == 0
+            && self.deferred_consensus_timeout == 0
+            && self.deferred_missing_submission == 0
+        {
+            return;
+        }
```
dartdart26: This if statement seems redundant.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1724 @@
...
+        MISSING_SUBMISSION_COUNTER.inc_by(self.deferred_missing_submission);
+        self.deferred_drift_detected = 0;
+        self.deferred_consensus_timeout = 0;
+        self.deferred_missing_submission = 0;
+    }
+
+    fn evaluate_open_handles(&mut self, current_block: u64) {
+        if !self.alerts_enabled {
+            return;
+        }
```
dartdart26: Would be nice to remove the `alerts_enabed` flag as it is everywhere. We could maybe have an optional DriftDetector.

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
```diff
@@ -300,6 +369,173 @@ impl<P: Provider<Ethereum> + Clone + 'static, A: AwsS3Interface + Clone + 'stati
         Ok(())
     }
 
+    async fn fetch_expected_coprocessor_tx_senders(
+        &self,
+        at_block: Option<u64>,
+    ) -> anyhow::Result<Vec<Address>> {
+        let Some(gateway_config_address) = self.conf.gateway_config_address else {
+            return Ok(Vec::new());
+        };
```
dartdart26: Wouldn't it be better to Result an Option? Or not call this function at all if there is no GW address? It would make the code easier to reason about and simpler.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
Eikix: alerts_enabled is only temporarily false during rebuild replay
could rename it to is_rebuild_mode

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
Eikix: do you mean like type Digest = FixedBytes<32> or something else?

dartdart26 (dismissed):
Thank you for working on it so quickly! I know it's been rushed and I understand the reasoning and the extensive use of AI.

That said, it's been a bit hard for me to verify all works as expected.

I am approving it, knowing that this is just for detection and won't change coprocessor state in any way. If we observe issues, we can fix them, but it won't affect coprocessor functionality.

Thanks @Eikix, good job!

P.S. Maybe the things that could affect the coprocessor are changes in gw_listener.rs, we could double check these, but they do look ok.

Eikix (commented):
[AUTOMATED]

## Code Review

Reviewed for bugs, guideline compliance, and error handling. Found 2 issues.

### Summary
- **Bug (low severity):** Histogram metrics are observed eagerly, breaking the deferred-metric invariant stated in the main loop comment.
- **Bug (low severity):** `finalize_completed_without_consensus` uses `==` instead of `>=`, missing handles when unexpected senders submit.

No guideline violations found. Error handling for drift detection paths correctly uses log-and-continue.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1713 @@
...
+            context,
+            digests: DigestPair {
+                ciphertext_digest: event.ciphertextDigest,
+                ciphertext128_digest: event.snsCiphertextDigest,
+            },
+            senders: event.coprocessorTxSenders,
+        });
+        state.local_consensus_checked = false;
+        if !self.replaying {
+            CONSENSUS_LATENCY_BLOCKS_HISTOGRAM
```
Eikix, line 238: [AUTOMATED] `CONSENSUS_LATENCY_BLOCKS_HISTOGRAM` (here) and `POST_CONSENSUS_COMPLETION_BLOCKS_HISTOGRAM` (line 549) are observed eagerly — directly when the event is processed. The counter metrics (`deferred_drift_detected`, etc.) are correctly deferred via `flush_metrics()`, which only runs after the DB persist in `update_listener_progress`. If that persist fails and the batch retries, counters are not double-counted but histograms are.

This violates the invariant stated at gw_listener.rs:369: *"Update metrics only after a successful DB update as we don't want to consider events that will be processed again."*

Low severity — only matters on DB persist failure + retry, and the effect is minor histogram distortion. Consider deferring histograms the same way counters are deferred.

Confidence: 90/100

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1713 @@
...
+    fn finalize_completed_without_consensus(&mut self) {
+        // Invariant: the gateway emits consensus as part of processing the final
+        // agreeing submission. Once every expected sender has submitted, the
+        // absence of a consensus event is already anomalous, so we alert
+        // immediately instead of waiting for `no_consensus_timeout_blocks`.
+        let completed_without_consensus = self
+            .open_handles
+            .iter()
+            .filter_map(|(handle, state)| {
+                (state.submissions.len() == state.expected_senders.len()
```
Eikix, line 507: [AUTOMATED] This uses `==` but should be `>=`. `observe_submission` accepts submissions from any sender — not just those in `expected_senders`. If an unexpected sender submits (as exercised by the `unexpected_sender_does_not_block_completion` test), `submissions.len()` can exceed `expected_senders.len()`, and `==` misses the handle.

The handle is not leaked — it eventually gets caught by `evict_stale` via `GatewayNeverReachedConsensus` after the full timeout. But the operator gets a generic timeout alert instead of the specific "all submitted but no consensus" alert, and the handle sits in memory longer than necessary.

`finish_if_complete` (line 540) correctly uses `<` (equivalent to proceeding on `>=`). This check should match: `>=` instead of `==`.

Confidence: 85/100

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
```diff
@@ -89,6 +89,33 @@ struct Conf {
...
+    /// the default of 3000 blocks ≈ 5 minutes, which accommodates coprocessors
+    /// that may be stuck for a few minutes.
+    #[arg(long, default_value_t = 3000, requires = "ciphertext_commits_address")]
+    drift_no_consensus_timeout_blocks: u64,
+
+    /// After consensus, how many additional blocks to wait for remaining
+    /// coprocessors to submit their ciphertext material. Measured in blocks —
+    /// with 100ms block times the default of 3000 blocks ≈ 5 minutes.
+    #[arg(long, default_value_t = 3000, requires = "ciphertext_commits_address")]
+    drift_post_consensus_grace_blocks: u64,
```
dartdart26: Sorry for going back at this again - now that I think about it, if there's no traffic in the system, 3K blocks might be a lot. So am wondering if that's a good value.

Makes me think if we should try to somehow instead measure wall clock time locally? Wdyt?

dartdart26 (commented):
Just put one comment re grace periods, please let me know what you think of it.

coprocessor/fhevm-engine/gw-listener/src/bin/gw_listener.rs
dartdart26: Reasoning is that catching up depends on host block actually, not on GW blocks. So maybe wall clock time is what we want.

Eikix: I'm not sure I follow the reasoning. What would be the difference between 3000 blocks = 5mn vs. a duration of 5mn or less / more?

dartdart26: Blocks are not produced on the GW when there are no txns.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1779 @@
...
+    current_expected_senders: Vec<Address>,
+    /// Handles waiting for consensus or post-consensus grace. Bounded implicitly:
+    /// `evict_stale` removes entries after `no_consensus_timeout` (no consensus)
+    /// or `post_consensus_grace` (consensus reached). Steady-state size is
+    /// proportional to handle throughput * timeout duration.
+    open_handles: HashMap<CiphertextDigest, HandleState>,
+    host_chain_id: ChainId,
+    local_node_id: String,
+    no_consensus_timeout: Duration,
+    post_consensus_grace: Duration,
```
dartdart26: Nit: Maybe this can be prefixed with `drift_` as is the config, for consistency.

You can merge without addressing this if you want, it's a nit.

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -0,0 +1,1779 @@
...
+    /// all open handles against the current chain tip. Called by
+    /// `rebuild_drift_detector` in `gw_listener.rs` after log replay completes.
+    pub(crate) async fn end_of_rebuild(&mut self, db_pool: &Pool<Postgres>) -> anyhow::Result<()> {
+        self.refresh_pending_consensus_checks(db_pool).await?;
+        self.evaluate_open_handles(Instant::now());
+        Ok(())
+    }
+}
+
+fn classify_handle(
```
dartdart26: Nit: Why not make this accept `&self` to avoid passing the durations?

You can merge without addressing this if you want, it's a nit.

dartdart26 (dismissed):
Added some nits, will let you decide if you want to address them or not. Otherwise, approved.

dartdart26 (approved):

Eikix:
@mergifyio queue

PR #2098, opened 2026-03-11
fix(coprocessor): use block hash to identify transaction

rudy-6-4:


rudy-6-4:
@claude /pr-review

coprocessor/fhevm-engine/db-migration/migrations/20260309163043_concurrent_index_computations_block_hash_2.sql
```diff
@@ -0,0 +1,3 @@
+-- no-transaction
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_computations_handle_transaction
+    ON computations (output_handle, transaction_id, block_hash);
```
PanGan21: If `transaction_id` is the transaction hash then might be more future proof to also restrict this per `chain_id`. However this is very edge and rare case.

rudy-6-4:
E2E: https://github.com/zama-ai/fhevm/actions/runs/23196828271/job/67407480558

coprocessor/fhevm-engine/db-migration/migrations/20260309163043_concurrent_index_computations_block_hash_2.sql
rudy-6-4: output_handle is almost unique, transaction id is almost unique, block hash is with very high probability unique.  I am not sure it's worth it to add chain id which is constant for multi-chain. Also it's just the index here so there is not risks except a slower selection case of collision.
I think the chain_id is/will be in the primary key (current one use tenant_id). 

But maybe you had in mind the risk of bad logic due to collision ?
Like here https://github.com/zama-ai/fhevm/pull/2098/changes#diff-ee9d15ae012b277f2f2380d60b93a8df3978b2694aaa0b8d3ce9b4b52bd18e6dR330 maybe the chain_id should be added ?

PanGan21: > I think the chain_id is/will be in the primary key (current one use tenant_id).

This solves my comment actually! So it is indeed safe to use it like this even with multiple chains

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
E2E success: https://github.com/zama-ai/fhevm/actions/runs/23753602589/job/69202207211

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

PR #2100, opened 2026-03-11
feat(host-contracts): implement epoch-based lifecycle for KMSVerifier

obatirou:
Closes https://github.com/zama-ai/fhevm-internal/issues/1130


obatirou:
Will be startup again from scratch when getting back on this

PR #2103, opened 2026-03-12
chore(coprocessor): update tfhe-rs to 1.5.4

dartdart26:
Also, treat `get_kind_of()` returning None for a correct index as an error instead of skipping the index. The skipping was there to workaround a bug in tfhe-rs that is now fixed in 1.5.4.

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
@@ -291,10 +291,10 @@ pub fn extract_ct_list(
 ) -> Result<Vec<SupportedFheCiphertexts>, FhevmError> {
     let mut res = Vec::new();
     for idx in 0..expanded.len() {
-        let Some(data_kind) = expanded.get_kind_of(idx) else {
+        let data_kind = expanded.get_kind_of(idx).ok_or_else(|| {
             tracing::error!(len = expanded.len(), idx, "get_kind_of returned None");
-            continue;
-        };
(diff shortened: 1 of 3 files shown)


coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -291,10 +291,10 @@ pub fn extract_ct_list(
 ) -> Result<Vec<SupportedFheCiphertexts>, FhevmError> {
     let mut res = Vec::new();
     for idx in 0..expanded.len() {
-        let Some(data_kind) = expanded.get_kind_of(idx) else {
+        let data_kind = expanded.get_kind_of(idx).ok_or_else(|| {
             tracing::error!(len = expanded.len(), idx, "get_kind_of returned None");
-            continue;
-        };
+            FhevmError::MissingTfheRsData
```
goshawk-3, line 296: Is MissingTfheRsData suitable for this error? Shall we  return a new type of error?

goshawk-3 (approved):

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
dartdart26: I thought about it, but we never handle these anyway anywhere. And it is kind of right. If you feel a new one is needed, I can add one.

dartdart26:
E2E pass: https://github.com/zama-ai/fhevm/actions/runs/23008064754

dartdart26:
@mergifyio queue

PR #2113, opened 2026-03-13
chore(coprocessor): improve host-listener tests performance

PanGan21:
## Summary

- Reduce `NB_EVENTS_PER_WALLET` from 200 to 50 in restart/reorg tests — total events are sufficient to test catchup, restart, and reorgs
- Replace hardcoded sleep waits with poll-based approach that exit as soon as the expected condition is met
- Reduce slow-lane test block time from 3.0s to 1.0s and replace sleeps with Anvil block number polling

## Performance

- main branch: 14m 45s
- current branch: 7m 42s

Ref: https://github.com/zama-ai/fhevm-internal/issues/1155


rudy-6-4 (approved):

PanGan21:
@mergify queue

PR #2119, opened 2026-03-16
refactor(coprocessor): improves test coverage on host-listener

PanGan21:
## Summary

Adds unit tests on host-listener files:
reorgs
- `test_get_missing_ancestors_shallow_reorg` — walk back to a known ancestor
- `test_get_missing_ancestors_deep_reorg` — walk back blocks through a long gap
- `test_get_missing_ancestors_beyond_max_depth` — stop at max depth limit
- `test_check_missing_ancestors_not_ready` — skip reorg detection when history < configured blocks

dependence chains
- `test_dependence_chains_empty_logs` — empty input returns no chains
- `test_dependence_chains_across_blocks_false` — past dependency ignored when `across-blocks` flag is off
- `test_dependence_chains_connex_two_past_chains_merge` — two past chains merge in connex mode

slow lane
- `classify_slow_disconnected_components_at_threshold_are_fast` — exactly at-threshold chains are not slow
- `classify_slow_single_chain_at_boundary` — slow lane condition is respected
- `propagate_slow_lane_non_linear_dependency` — non linear DAG

health check
- `is_alive_after_blockchain_tick_update` — alive after blockchain tick
- `is_alive_after_timeout_tick_update` — alive after timeout tick
- `is_alive_after_database_tick_update` — alive after database tick
- `is_alive_after_all_ticks_update` — alive after all ticks
- `is_alive_at_creation` — alive at creation

Closes: https://github.com/zama-ai/fhevm-internal/issues/1155

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/utils.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/utils.rs
@@ -115,6 +115,17 @@ impl HeartBeat {
         let elapsed = self.now_timestamp() - self.timestamp.load(Ordering::Relaxed);
         elapsed <= freshness.as_secs()
     }
+
+    pub fn with_elapsed_secs(elapsed_secs: u64) -> Self {
+        let now = std::time::Instant::now();
+        let timestamp_origin = now
+            .checked_sub(Duration::from_secs(elapsed_secs))
+            .unwrap_or(now);
+        Self {
+            timestamp_origin,
+            timestamp: Arc::new(AtomicU64::new(0)),
+        }
+    }
 }
 
 impl Default for HeartBeat {
--- a/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
@@ -1241,6 +1241,9 @@ mod tests {
         let missing = iter.get_missing_ancestors(blocks[10]).await;
 
         assert_eq!(missing.len(), 3);
+        assert_eq!(missing[0].number, 7);
+        assert_eq!(missing[1].number, 8);
+        assert_eq!(missing[2].number, 9);
     }
 
     // Skips reorg detection when history has fewer than 2 blocks and just adds the block.
--- a/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
@@ -1136,3 +1136,138 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
     cancel_token.cancel();
     anyhow::Result::Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use std::collections::VecDeque;
+    use std::sync::Arc;
+
+    use alloy::node_bindings::Anvil;
+    use alloy::providers::ext::AnvilApi;
+    use alloy::providers::{Provider, ProviderBuilder, WsConnect};
+    use tokio::sync::RwLock;
+
+    use fhevm_engine_common::utils::HeartBeat;
+
+    use super::*;
+
+    fn new_test_iter(reorg_max: u64) -> InfiniteLogIter {
+        InfiniteLogIter {
+            url: String::new(),
+            block_time: 12,
+            contract_addresses: vec![],
+            catchup_blocks: None,
+            next_blocklogs: VecDeque::new(),
+            stream: None,
+            provider: Arc::new(RwLock::new(None)),
+            last_valid_block: None,
+            start_at_block: None,
+            end_at_block: None,
+            absolute_end_at_block: None,
+            catchup_margin: 5,
+            catchup_paging: 100,
+            tick_timeout: HeartBeat::new(),
+            tick_block: HeartBeat::new(),
+            reorg_maximum_duration_in_blocks: reorg_max,
+            block_history: BlockHistory::new(reorg_max as usize),
+            catchup_finalization_in_blocks: 20,
+            timeout_request_websocket: 15,
+        }
+    }
+
+    async fn setup_iter_with_chain(
+        num_blocks: u64,
+        reorg_max: u64,
+        known: std::ops::RangeInclusive<usize>,
+    ) -> (
+        alloy::node_bindings::AnvilInstance,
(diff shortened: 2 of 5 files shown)


coprocessor/fhevm-engine/host-listener/src/health_check.rs
```diff
@@ -62,3 +62,74 @@ impl HealthCheckService for HealthCheck {
...
+            blockchain_tick: HeartBeat::new(),
+            blockchain_provider: Arc::new(RwLock::new(None)),
+            database_pool: Arc::new(RwLock::new(pool)),
+            database_tick: HeartBeat::new(),
+        }
+    }
+
+    #[tokio::test]
+    async fn is_alive_after_blockchain_tick_update() {
+        let health_check = build_test_health_check();
```
Eikix: This test might pass even without the `update()`. 
I think HeartBeat::new() starts out fresh, so a newly created HealthCheck is already alive here. That means we’re not proving that blockchain_tick.update() causes liveness, only that the instance was created recently. 

If that's true (im not sure), try to make the initial blockchain ticks stale and then assert that updating blockchain_tick is what turns is_alive() on

Eikix: might need something like 

```
#[cfg(test)]
impl HeartBeat {
    pub fn with_elapsed_secs(elapsed_secs: u64) -> Self {
        let now = std::time::Instant::now();
        let timestamp_origin = now
            .checked_sub(Duration::from_secs(elapsed_secs))
            .unwrap_or(now);

        Self {
            timestamp_origin,
            timestamp: Arc::new(AtomicU64::new(0)),
        }
    }
}

fn build_test_health_check(
        blockchain_timeout_tick: HeartBeat,
        blockchain_tick: HeartBeat,
        database_tick: HeartBeat,
    ) -> HealthCheck {
        let db_url = "postgres://test:test@localhost:5432/test";
        let pool = sqlx::postgres::PgPoolOptions::new()
            .connect_lazy(db_url)
            .unwrap();

        HealthCheck {
            blockchain_timeout_tick,
            blockchain_tick,
            blockchain_provider: Arc::new(RwLock::new(None)),
            database_pool: Arc::new(RwLock::new(pool)),
            database_tick,
        }
    }

```

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -1136,3 +1136,138 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
...
+
+    // Stops walking at reorg_maximum_duration_in_blocks even if more unknown ancestors remain.
+    // Tests that the function doesn't walk forever and respects the configured max depth.
+    #[tokio::test]
+    async fn test_get_missing_ancestors_beyond_max_depth() {
+        let (_anvil, iter, blocks) = setup_iter_with_chain(10, 3, 0..=0).await;
+
+        let missing = iter.get_missing_ancestors(blocks[10]).await;
+
+        assert_eq!(missing.len(), 3);
```
Eikix, line 1243: can also assert on the missing ancestors block number

Eikix: smth like 
assert_eq!(missing[0].number, 7);
assert_eq!(missing[1].number, 8);
assert_eq!(missing[2].number, 9);

if that passes

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
```diff
@@ -1048,4 +1048,65 @@ mod tests {
...
+        cache.write().await.put(past_handle2, past_chain_hash2);
+
+        let mut logs = vec![];
+        let tx1 = TransactionHash::with_last_byte(2);
+        let _v = op2(past_handle1, past_handle2, &mut logs, tx1);
+
+        let chains = dependence_chains(&mut logs, &cache, true, true).await;
+
+        assert_eq!(chains.len(), 1);
+        assert_eq!(cache.read().await.len(), 3);
```
Eikix, line 1111: can you assert on the chain hash too if possible?

coprocessor/fhevm-engine/host-listener/src/health_check.rs
PanGan21: You are right on this the tests pass without calling `update()`. I refactored to be more realistic following your suggestion

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
PanGan21: done

coprocessor/fhevm-engine/host-listener/src/database/dependence_chains.rs
PanGan21: done

coprocessor/fhevm-engine/fhevm-engine-common/src/utils.rs
```diff
@@ -115,6 +115,17 @@ impl HeartBeat {
         let elapsed = self.now_timestamp() - self.timestamp.load(Ordering::Relaxed);
         elapsed <= freshness.as_secs()
     }
+
+    pub fn with_elapsed_secs(elapsed_secs: u64) -> Self {
```
Eikix, line 119: perhaps make it test only so it's not bundled into production code if thats possible

PanGan21: Hmm test only will not work because it should be used from a different crate.
Would you agree if I make a cargo feature `test-utils` and gate this function behind this feature and then add `fhevm-engine-common` in host listener dev dependencies with this feature enabled?

Eikix: what do you think is best?
perhaps then letting this code bleed into production code is not that bad

PanGan21: The most common standard is to have a separate feature for test utils which is enabled only when testing but it is not worthy if it is only for one function so we could keep the function for now and in case we see similar situations we add the feature?
What do you think?

Eikix (approved):
lgtm

PanGan21:
@mergify queue

PR #2122, opened 2026-03-17
feat(coprocessor): add ability to revert to a previous state

dartdart26:
If a coprocessor is in a bad state, it can be useful to revert to a previous state and redo the computations and other operations with the hope that it reaches the correct state.

Reversal can be initiated by providing a host chain ID and a block number to revert to. The reversal process expects that all coprocessor components are stopped before running it.

In order to delete data by block number and chain ID, we have to add a DB migration to keep track of them for all  elevant tables. For blocks before this migration, we won't be able to revert, but we assume this is acceptable, also given that there isn't much we can do.

Reasoning for the way we implemented reversal is that we assumes we would be able to detect a state drift with a ciphertext granularity. Consequently, the first such drift that we observe means that no prior state was affected. That is why we delete all data for blocks including the offending one and the ones after it.

Note that above assumes all ciphertext have been computed up to the offending block, but in reality, due to out-of-order processing, it might not be the case. That means some prior ciphertexts might also drift. Therefore, operators might choose to go a bit further back than the offending block to make it even more likely to capture the root cause of the drift.

For now, we cannot revert across a key rotation boundary. The tfhe-worker always uses the latest key from the `keys` table. If a new key was activated after to_block_number, re-processed computations will use the new key instead of the original one, producing incorrect ciphertexts.

This commit adds a test that checks the DB after the reversal script has run, an integration test for the host listener, and an E2E test.


dartdart26:
@Eikix Please don't review yet, still working on it.

dartdart26:
E2E success here: https://github.com/zama-ai/fhevm/actions/runs/23251441988/job/67594233162

dartdart26:
Success E2E: https://github.com/zama-ai/fhevm/actions/runs/23295863890

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
```diff
@@ -0,0 +1,170 @@
...
+-- Reasoning for this script is that it assumes we would be able to detect a state drift
+-- with a ciphertext granularity. Consequently, the first such drift that we observe means
+-- that no prior state was affected. To revert, pass the block before the offending one
+-- as to_block_number (i.e. offending_block - 1). The script deletes all data for blocks
+-- strictly greater than to_block_number.
+--
+-- Usage:
+--   psql -v chain_id=<CHAIN_ID> -v to_block_number=<BLOCK_NUMBER> -f revert_coprocessor_db_state.sql
+--
+-- WARNING: For now, do not revert across a key rotation boundary. The tfhe-worker always
```
Eikix: Should there be a way to detect this & fail fast? We can just treat it as a case we won't code for, and won't code for, for a long time. But it can be safer to fail fast so as to not forget about it (if it's possible to fail fast here)

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
```diff
@@ -0,0 +1,170 @@
...
+-- ===========================================================================
+-- Delete from tables that have their own block_number column
+-- ===========================================================================
+
+DELETE FROM allowed_handles
+ WHERE block_number > :'to_block_number'
+   AND host_chain_id = :'chain_id';
+
+DELETE FROM ciphertext_digest
+ WHERE txn_block_number > :'to_block_number'
```
Eikix: is `txn_block_number` the gateway block number of the txn where the digest is posted? if so, is it being compared to a host block number?

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
```diff
@@ -0,0 +1,170 @@
...
+\set ON_ERROR_STOP on
+
+BEGIN;
+
+-- ===========================================================================
+-- Validate parameters
+-- ===========================================================================
+
+CREATE TEMP TABLE _param_check AS
+SELECT :'chain_id'::bigint AS chain_id, :'to_block_number'::bigint AS to_block_number;
```
PanGan21, line 46: Should there be some validation about (chain_id, block_number) or we trust the person executing this?

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: Great catch! I missed that - I think we need to delete by handle here, let me look into it.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: We could check, but I am not sure what to check against. We mostly trust, yeah. Any ideas on what to check against?

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: Let me think about how exactly - I guess we could try see if a ciphertext for the last block is from key X and if a ciphertext from the first one (the to block) is from key Y. Issue is hosts are not tied to the GW, so that might be a bit tricky.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
PanGan21: Maybe making sure that there are indeed ciphertexts linked to this block number, however this seems that we still need to trust the person executing. 
If that was not only sql script we could validate with gateway about the incorrect commitment potentially but most likely this is not possible now

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: @Eikix I think this goes back to the offline discussion we had about the GW chain not in sync with the host chains. AFAICT, there is not reliable way to say a key rotation hasn't happened for given host chain C in the range we are about to remove.

We could look at the `ciphertext_digest` table, but we don't know if it has populated fully with all the digests that would detect a different key (it's a race between the sns-worker and the operator stopping the system to run the reversal script). Maybe still worth adding that check, but keep the warning and be explicit that we can't currently detect it.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
PanGan21: We can also make clear that we trust the operator and do not check something. Just wanted to emphasize that wrong block number can potentially harm

dartdart26: Yeah, I will add a comment in the script that we don't check and trust.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: I added a best-effort check.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: Done.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
```diff
@@ -0,0 +1,210 @@
...
+-- ***WARNING***: Do not revert across a key rotation boundary. The tfhe-worker always
+-- uses the latest key from the `keys` table. If a new key was activated after
+-- to_block_number, re-processed computations will use the new key instead of
+-- the original one, producing incorrect ciphertexts. The script checks
+-- ciphertext_digest.key_id_gw for affected handles and fails if a mismatch
+-- is detected, but this check is BEST-EFFORT ONLY (rows may not exist yet if the
+-- sns-worker hasn't processed them).
+--
+-- Reasoning for this script is that it assumes we would be able to detect a state drift
+-- with a ciphertext granularity. Consequently, the first such drift that we observe means
```
antoniupop: As discussed earlier, the meaning of "first" should be in L1 order rather than "observation order" being when we first detect out-of-consensus state. We don't enforce L1 order on CT committment.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
```diff
@@ -0,0 +1,210 @@
...
+DELETE FROM computations
+ WHERE host_chain_id = :'chain_id'
+   AND block_number > :'to_block_number';
+
+-- ===========================================================================
+-- Delete dependence_chain entries
+-- ===========================================================================
+
+DELETE FROM dependence_chain
+ WHERE dependence_chain_id IN (SELECT dc_id FROM _affected_dc_ids);
```
antoniupop: It's not clear to me that this is either safe or needed. 
If there's computations that are not covered by reset still uncompleted in those DCs there could be an edge case. Conversely if we leave the DCs and we just clear the computations, I would think they would be marked completed then new ones created when HL catches up.

antoniupop (commented):
Not sure it's relevant/needed, but is this handled on GW side to account for multiple commitments for the same handle and same copro? These could be the same or different CTs whether drifted or not. I guess ACL propagation (and delegation) too since you drop from allowed_handles so will be re-sent on catch-up or is that flow already removed?

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: Yes, I put a comment below to make it clear. But I can reword it if you think is needed.

dartdart26:
> Not sure it's relevant/needed, but is this handled on GW side to account for multiple commitments for the same handle and same copro? These could be the same or different CTs whether drifted or not. I guess ACL propagation (and delegation) too since you drop from allowed_handles so will be re-sent on catch-up or is that flow already removed?

I am not exactly sure if I get it, but if you refer to sending different ciphertext digests to the GW for the same handle, the assumption here is that a minority of coprocessors drift, with the majority being able to reach consensus. Therefore, if CoprocessorX is being reverted and then submits digests again, it will get `CoprocessorAlreadyAdded` revert irrespective of whether consensus has been reached or not. CoprocessorX will move on, even if the bad digest it originally sent stays in the GW contract - arguably we want that.

ACL propagation is no longer happening on main.

antoniupop:
> I am not exactly sure if I get it, but if you refer to sending different ciphertext digests to the GW for the same handle, the assumption here is that a minority of coprocessors drift, with the majority being able to reach consensus. Therefore, if CoprocessorX is being reverted and then submits digests again, it will get `CoprocessorAlreadyAdded` revert irrespective of whether consensus has been reached or not. CoprocessorX will move on, even if the bad digest it originally sent stays in the GW contract - arguably we want that.

So we assume we don't actually break out of consensus here? I think this is risky, we would not be able to restore consensus if it happens. I can't quite see the benefit of keeping the old discarded digest in GW if a copro has re-computed and updated its DB (what bout S3? re-upload or stay consistent with GW and keep old one?). I'd be in favour of updating I think. Am I missing something?

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: I see, good point. So are you suggesting we just don't delete from the `dependence_chain` table, but keep the delete from `computations`?

dartdart26:
> > I am not exactly sure if I get it, but if you refer to sending different ciphertext digests to the GW for the same handle, the assumption here is that a minority of coprocessors drift, with the majority being able to reach consensus. Therefore, if CoprocessorX is being reverted and then submits digests again, it will get `CoprocessorAlreadyAdded` revert irrespective of whether consensus has been reached or not. CoprocessorX will move on, even if the bad digest it originally sent stays in the GW contract - arguably we want that.
> 
> So we assume we don't actually break out of consensus here? I think this is risky, we would not be able to restore consensus if it happens. I can't quite see the benefit of keeping the old discarded digest in GW if a copro has re-computed and updated its DB (what bout S3? re-upload or stay consistent with GW and keep old one?). I'd be in favour of updating I think. Am I missing something?

We can't change the GW contracts. The main assumption of the whole system is that a majority is always honest. We can't break that and expect the system to behave nicely. Another problem related to that is we have one implementation only, leading to correlated bugs on all instances.

Re S3, the data is keyed on the digest, so we will keep the old ciphertexts hanging for now. If we want garbage collection on S3, I suggest we do it in another PR.

antoniupop:
> We can't change the GW contracts. The main assumption of the whole system is that a majority is always honest. We can't break that and expect the system to behave nicely. Another problem related to that is we have one implementation only, leading to correlated bugs on all instances.

I'm talking about errors, not malicious instances. These can accumulate (contamination) so could end up there if the reconciliation process is not fast enough (or we're very unlucky).

> Re S3, the data is keyed on the digest, so we will keep the old ciphertexts hanging for now. If we want garbage collection on S3, I suggest we do it in another PR.

Makes sense yes. In any case I would still upload corrected CTs as well even if not GCing old ones.

dartdart26:
> > We can't change the GW contracts. The main assumption of the whole system is that a majority is always honest. We can't break that and expect the system to behave nicely. Another problem related to that is we have one implementation only, leading to correlated bugs on all instances.
> 
> I'm talking about errors, not malicious instances. These can accumulate (contamination) so could end up there if the reconciliation process is not fast enough (or we're very unlucky).
> 
> > Re S3, the data is keyed on the digest, so we will keep the old ciphertexts hanging for now. If we want garbage collection on S3, I suggest we do it in another PR.
> 
> Makes sense yes. In any case I would still upload corrected CTs as well even if not GCing old ones.

Re GW - not sure what we can do. Any suggestions?

Re upload of corrected CT to S3 - yes, that will happen automatically.

antoniupop:
> Re GW - not sure what we can do. Any suggestions?

I'm no expert on GW contracts, I'd think we want to:
- keep any achieved consensus (so if it was reached already, we don't change anything)
- update a coprocessor's "vote" if consensus was not yet reached

edit: assuming indeed that a majority is honest, this should not break any guarantees, just allow correcting errors after a restore

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: As discussed offline, I've removed the delete of data from `dependence_chain`.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: I reworded it.

antoniupop (dismissed):
LGTM!

dartdart26:
E2E pass: https://github.com/zama-ai/fhevm/actions/runs/23349452140

dartdart26:
@mergifyio queue

dartdart26:
@mergifyio queue

dartdart26:
@mergifyio queue

antoniupop (approved):

dartdart26:
includes changes in https://github.com/zama-ai/fhevm/pull/2147

dartdart26:
E2E: https://github.com/zama-ai/fhevm/actions/runs/23448952085

antoniupop (approved):

eudelins-zama (approved):

PR #2132, opened 2026-03-19
docs: update integration guide to add details about wrapping/unwrapping

ankurdotb:
This guide should help wallets and CEXs evaluate integrations at a high level.

This is meant to be just a basic guide to provide that high level info. When the new Zama SDK is ready/released, the guide will be updated to talk through the specific high-level functions necessary at an SDK level.


docs/examples/integration-guide.md
```diff
@@ -0,0 +1,272 @@
...
+- **Wrapped confidential token**: A standard ERC-20 token that has been wrapped into an ERC-7984 confidential form via a wrapper contract. The underlying ERC-20 remains unchanged.
+- **Confidential Token Wrappers Registry**: An on-chain registry that maps ERC-20 tokens to their corresponding ERC-7984 confidential token wrappers.
+
+## Wallet and exchange integration at a glance
+
+At a high-level, to integrate Zama Protocol into a wallet or exchange, you do **not** need to run FHE infrastructure. You can interact with the Zama Protocol using [Relayer SDK](https://docs.zama.org/protocol/relayer-sdk-guides) in your wallet or app. These are the steps at a high-level:
+
+1. **Relayer SDK initialization** in web app, browser extension, or mobile app. Follow the [setup guide for Relayer SDK](https://docs.zama.org/protocol/relayer-sdk-guides/development-guide/webapp). In browser contexts, importing the library via the CDN links is easiest. Alternatively, do this by importing the `@zama-fhe/relayer-sdk` NPM package.
```
poppyseedDev, line 32: The only thing I am not too sure about is how links should be referenced, usually we try to reference them as a relative link instead of full link throughout the doc. 

Instead of:
> https://docs.zama.org/protocol/relayer-sdk-guides/development-guide/webapp

It would be:
> ../relayer-sdk-guides/development-guide/webapp

but otherwise looking good!

poppyseedDev (approved):

poppyseedDev:
@mergifyio queue

docs/examples/integration-guide.md
ankurdotb: I've double-checked this: yes, I'd prioritise relative links as well, as Gitbook then automatically links it correctly. 

However, Relayer SDK docs aren't in this repo (yet), they are still linked to from [the Relayer SDK repo](https://github.com/zama-ai/relayer-sdk/tree/main/docs). So in this case, the relative link would break because Gitbook would not see it in _this_ repo.

PR #2134, opened 2026-03-19
refactor(test-suite): replace fhevm bash cli with lean Bun TS

Eikix:
Closes https://github.com/zama-ai/fhevm-internal/issues/1042

Supersedes #2066.

`#2066` keeps the iteration history and design exploration. This PR is the same final direction with a clean atomic history for review.

## Summary
Replace the legacy `test-suite/fhevm/fhevm-cli` bash flow with a lean Bun/TypeScript CLI.

The new CLI makes target resolution, runtime generation, scenarios, local overrides, resume, upgrade, and test execution explicit instead of spreading that behavior across bash scripts and CI glue.

## What changed
- replace the bash CLI/runtime scripts with a Bun/TS implementation under `test-suite/fhevm/src`
- move tracked runtime inputs into clearer buckets:
  - `docker-compose/`
  - `templates/`
  - `static/`
  - `profiles/`
  - `scenarios/`
- keep generated runtime state under `.fhevm/` with a hard split between:
  - `.fhevm/state`
  - `.fhevm/runtime`
- support explicit targets:
  - `latest-supported`
  - `latest-main`
  - `sha --sha <git-sha>`
  - `devnet`
  - `testnet`
  - `mainnet`
- support local overrides and scenario-driven coprocessor topologies
- keep `deploy` as an alias for `up`
- replace `doctor` with `up --dry-run`

## CI contract
- direct PR e2e boots `latest-main --build` with the checked-in `two-of-two` scenario
- PR e2e now runs the bundled `./fhevm-cli test light` suite instead of copying those steps in YAML
- orchestrated merge-queue/release validation calls the reusable e2e workflow with `build=false`
- the reusable e2e workflow is explicitly told when it is running in orchestrated mode
- the workflow now also gates the CLI itself with:
  - `bun run check`
  - `bun test src`

## Validation
Verified locally on this branch:
- focused live-stack and CI-path validation during the original review cycle

## Notes
- `latest-supported` is now a tracked maintained profile
- `latest-main` is the moving baseline for local and PR validation
- `two-of-two` is the canonical checked-in multicopro scenario preset


.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -83,18 +80,23 @@ on:
...
+    # Run on manual/reusable invocations, and on ordinary in-repo PRs.
+    # Skip only the direct PR path for sacrificial/merge-queue branches.
+    if: ${{ github.event_name != 'pull_request' || inputs.orchestrated || (github.event.pull_request.head.repo.full_name == github.repository && !startsWith(github.head_ref, 'mergify/merge-queue/')) }}
     permissions:
       contents: 'read' # Required to checkout repository code
       id-token: 'write' # Required for OIDC authentication
       packages: 'read' # Required to read GitHub packages/container registry
     env:
+      GH_TOKEN: ${{ secrets.GHCR_READ_TOKEN || github.token }}
+      BUILD: ${{ github.event_name == 'pull_request' && 'true' || (inputs.build && 'true' || 'false') }}
```
Eikix: [AUTOMATED] **Bug**: `BUILD` ignores `inputs.build` in the orchestrated path.

In a reusable workflow called via `workflow_call`, `github.event_name` is inherited from the caller (not set to `"workflow_call"`). When the orchestrator fires on `pull_request` and calls this workflow with `build: false`, the expression short-circuits:

```
'pull_request' == 'pull_request' && 'true'  →  'true'
```

`inputs.build = false` is never consulted. This causes the e2e job to pass `--build` to `fhevm-cli up`, triggering a redundant local image rebuild even though the orchestrator already built and published the images.

Suggested fix — gate on `inputs.orchestrated`:
```yaml
BUILD: ${{ inputs.orchestrated && 'false' || (github.event_name == 'pull_request' && 'true' || (inputs.build && 'true' || 'false')) }}
```

Confidence: 92/100

Eikix (commented):
[AUTOMATED]

## Code Review Summary

Reviewed for bugs, guideline compliance, and error handling across all 39 changed files.

### Findings

| Category | Count | Severity |
|----------|-------|----------|
| Bug | 1 | High — CI correctness |
| Naming (guideline) | 3 | Nit |
| Error handling | 3 | Medium |

The BUILD env var issue in the e2e workflow is the most impactful finding — it causes unnecessary image rebuilds in the orchestrated CI path. The naming and error handling items are non-blocking improvement suggestions.

Overall the PR is well-structured: the compose path migration is consistent, the CI consolidation is clean, and the documentation is thorough.

test-suite/fhevm/src/generate/env.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  envs["gateway-sc"].NUM_COPROCESSORS = String(plan.topology.count);
+  envs["gateway-sc"].COPROCESSOR_THRESHOLD = String(plan.topology.threshold);
+  envs["host-sc"].NUM_COPROCESSORS = String(plan.topology.count);
+  envs["host-sc"].COPROCESSOR_THRESHOLD = String(plan.topology.threshold);
+  envs["coprocessor"].DATABASE_URL = `postgresql://${envs.database.POSTGRES_USER}:${envs.database.POSTGRES_PASSWORD}@${POSTGRES_HOST}/coprocessor`;
+  envs["coprocessor"].TENANT_API_KEY = DEFAULT_TENANT_API_KEY;
+  envs["coprocessor"].COPROCESSOR_API_KEY = DEFAULT_TENANT_API_KEY;
+  envs["coprocessor"].AWS_ENDPOINT_URL = state.discovery?.endpoints.minioExternal ?? MINIO_INTERNAL_URL;
+  const kp = state.discovery?.minioKeyPrefix ?? "PUB";
```
Eikix: [AUTOMATED] **Nit: naming** — `kp` and `minioInt` (line 81) are abbreviations.

Global guideline: *"Prefer explicit verbose names over short clever ones. Avoid abbreviations."*

The source properties are already explicit (`minioKeyPrefix`, `minioInternal`), so these could just be:
```ts
const minioKeyPrefix = state.discovery?.minioKeyPrefix ?? "PUB";
const minioInternalUrl = state.discovery?.endpoints.minioInternal ?? MINIO_INTERNAL_URL;
```

Same pattern at `stack.ts:690` (`kp`).

Confidence: 93/100

test-suite/fhevm/src/generate/addresses.ts
```diff
@@ -0,0 +1,71 @@
...
+    ["FHEVM_EXECUTOR_CONTRACT_ADDRESS", state.discovery?.host.FHEVM_EXECUTOR_CONTRACT_ADDRESS],
+    ["KMS_VERIFIER_CONTRACT_ADDRESS", state.discovery?.host.KMS_VERIFIER_CONTRACT_ADDRESS],
+    ["INPUT_VERIFIER_CONTRACT_ADDRESS", state.discovery?.host.INPUT_VERIFIER_CONTRACT_ADDRESS],
+    ["HCU_LIMIT_CONTRACT_ADDRESS", state.discovery?.host.HCU_LIMIT_CONTRACT_ADDRESS],
+    ["PAUSER_SET_CONTRACT_ADDRESS", state.discovery?.host.PAUSER_SET_CONTRACT_ADDRESS],
+  ]);
+
+export const renderHostAddressesSolidity = (state: Pick<State, "discovery">) =>
+  renderSolidityFile([
+    ["aclAdd", state.discovery?.host.ACL_CONTRACT_ADDRESS],
```
Eikix: [AUTOMATED] **Nit: naming** — `aclAdd`, `fhevmExecutorAdd`, `kmsVerifierAdd`, etc. abbreviate "Address" as "Add".

Global guideline: *"Prefer explicit verbose names over short clever ones. Avoid abbreviations."*

The gateway section of this same file uses unabbreviated names (`gatewayConfigAddress`, `inputVerificationAddress`), making this inconsistent within the file.

Suggested: `aclAddress`, `fhevmExecutorAddress`, `kmsVerifierAddress`, `inputVerifierAddress`, `hcuLimitAddress`, `pauserSetAddress`.

Confidence: 90/100

test-suite/fhevm/src/resolve/bundle-store.ts
```diff
@@ -0,0 +1,150 @@
...
+    throw new GitHubApiError("Lock file must include an env object with every version key");
+  }
+  const missing = VERSION_KEYS.filter((key) => typeof candidate.env?.[key] !== "string" || !candidate.env[key]?.length);
+  if (missing.length) {
+    throw new GitHubApiError(`Lock file is missing required version keys: ${missing.join(", ")}`);
+  }
+  return candidate as VersionBundle;
+};
+
+const validateLockedRuntimeCompat = async (bundle: VersionBundle) => {
```
Eikix: [AUTOMATED] **Error handling** — `validateLockedRuntimeCompat` is a no-op stub that returns `bundle` unchanged, but is called in two places as though it validates the bundle.

A function named `validate...` that silently does nothing is confusing — callers assume compat was checked. Consider either implementing the body, or removing the function and inlining the pass-through so the code reads what it does.

Confidence: 85/100

test-suite/fhevm/src/stack.ts
```diff
@@ -0,0 +1,1602 @@
...
+        body: JSON.stringify({
+          jsonrpc: "2.0",
+          id: 1,
+          method: "eth_call",
+          params: [{ to: withHexPrefix(discovery.gateway.KMS_GENERATION_ADDRESS), data }, "latest"],
+        }),
+      });
+      if (!response.ok) return 0n;
+      const payload = (await response.json()) as { result?: string };
+      return payload.result ? BigInt(payload.result) : 0n;
```
Eikix: [AUTOMATED] **Error handling** — `BigInt(payload.result)` throws `SyntaxError` on non-numeric strings (e.g. `"0x"`, revert reasons). The outer `catch` (line 725) returns `null`, so `waitForBootstrap` silently retries for up to 240 seconds before timing out.

A persistent RPC misconfiguration or wrong contract address would cause a 4-minute hang with no diagnostic output.

Consider guarding:
```ts
if (payload.result && !/^0x[0-9a-fA-F]+$/.test(payload.result)) {
  console.log(`[warn] unexpected eth_call result: ${payload.result}`);
  return 0n;
}
return payload.result ? BigInt(payload.result) : 0n;
```

Confidence: 88/100

test-suite/fhevm/src/stack.ts
```diff
@@ -0,0 +1,1602 @@
...
+    (item): item is LocalOverride & { services: string[] } =>
+      !!item.services?.length && SCHEMA_COUPLED_GROUPS.includes(item.group),
+  );
+
+const minioIp = async () => {
+  const result = await run(["docker", "inspect", "fhevm-minio"], { allowFailure: true });
+  if (result.code !== 0) {
+    throw new PreflightError("Could not determine MinIO IP");
+  }
+  const inspected = JSON.parse(result.stdout) as Array<{
```
Eikix: [AUTOMATED] **Error handling** — `JSON.parse(result.stdout)` here lacks a try/catch, unlike the identical pattern in `dockerInspect()` at line 534 which wraps it in a guard that throws `PreflightError` with context.

If `docker inspect` returns invalid JSON, this surfaces as a raw `SyntaxError` with no indication of which operation failed.

Suggested: apply the same guard used at line 534.

Confidence: 90/100

test-suite/fhevm/src/commands/test.ts
```diff
@@ -0,0 +1,654 @@
...
+  HEAVY_TEST_PROFILES,
+  LIGHT_TEST_PROFILES,
+  STANDARD_TEST_PROFILES,
+  TEST_GREP,
+  TEST_PARALLEL,
+  TEST_SUITE_CONTAINER,
+} from "../layout";
+import type { TestOptions } from "../types";
+
+const DRIFT_WARNING = '"message":"Drift detected: observed multiple digest variants for handle"';
```
Eikix, line 29: I guess this is quite flaky since drift warning log can change over time. But then test should fail and catch it.

test-suite/fhevm/src/commands/test.ts
```diff
@@ -0,0 +1,654 @@
...
+const DB_REVERT_CONTAINERS = [
+  "host-listener",
+  "host-listener-poller",
+  "gw-listener",
+  "tfhe-worker",
+  "sns-worker",
+  "transaction-sender",
+  "zkproof-worker",
+] as const;
+const DEFAULT_DB_REVERT_CHAIN_ID = "12345";
```
Eikix: Why do we need a magic number here?

test-suite/fhevm/src/commands/test.ts
```diff
@@ -0,0 +1,654 @@
...
+const DEFAULT_DB_REVERT_TESTS = "test add 42 to uint64 input and decrypt";
```
Eikix: Why do we hard code this?

test-suite/fhevm/src/flow/up-flow.ts
```diff
@@ -0,0 +1,1863 @@
...
+      throw new PreflightError(
+        `Predicted bootstrap ids drifted: expected ${discovery.fheKeyId}/${discovery.crsKeyId}, got ${actualFheKeyId}/${actualCrsKeyId}`,
+      );
+    }
+    return { actualFheKeyId, actualCrsKeyId };
+  } catch (error) {
+    if (error instanceof MinioError || error instanceof PreflightError) {
+      throw error;
+    }
+    return null;
```
Eikix: [AUTOMATED]

**Error handling**: This catch block silently discards unexpected errors (returning `null` = "not ready yet") with zero logging.

`ethCallRaw` already returns `0n` for expected "not ready" cases (`!response.ok`, `!payload.result`), so errors reaching this catch represent genuinely unexpected failures — `RpcError` from malformed responses, `TypeError` from persistent network issues, etc.

A user who misconfigures a gateway RPC URL will wait 240s polling, seeing only `[wait] bootstrap materials (10s elapsed)` logs, then get `BootstrapTimeout` with no hint that the endpoint was unreachable the entire time.

Consider logging the swallowed error so persistent failures are diagnosable:
```ts
} catch (error) {
  if (error instanceof MinioError || error instanceof PreflightError) {
    throw error;
  }
  console.log(`[warn] bootstrap probe error (will retry): ${error instanceof Error ? error.message : String(error)}`);
  return null;
}
```

Nit: not blocking — the current behavior is functional, just hard to debug when things go wrong.

Confidence: 88/100

Eikix (commented):
[AUTOMATED]

## Code Review

Reviewed for bugs, security issues, guideline compliance, and error handling.

**Result**: 1 issue found (error handling). No bugs or security issues detected.

The CLI is well-structured with thorough input validation, proper async/await patterns, correct shell escaping for `sh -lc` commands, and a solid typed error hierarchy. The 15-step boot pipeline, state persistence, and resume logic all look correct.

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -1,8 +1,17 @@
 name: test-suite-e2e-tests
 
 on:
+  pull_request:
```
Eikix, line 4: TBD: for small PRs, do we want a way to skip it?
e.g. label e2e to run it on the PR

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -1,8 +1,17 @@
...
   workflow_dispatch:
     inputs: &workflow_inputs
+      orchestrated:
+        description: "Run from the orchestrated image-validation path."
+        default: false
+        type: boolean
+      build:
```
Eikix, line 17: From mob review: `build` on PR is too aggressive. TBD what to do. 
Until we have a 'lighter' way to run it

Eikix:
Mob review comment: print each version of the stack at start

.github/workflows/test-suite-e2e-tests.yml
Eikix: For now gating behind a e2e label

PanGan21 (dismissed):

PanGan21 (dismissed):

Eikix (commented):
[AUTOMATED]

## Code Review

Reviewed for bugs (2 independent passes), guideline compliance, and error handling.

**Bugs**: None found. Shell commands use `Bun.spawn` with argv arrays (no shell injection). SQL uses validated numeric chain IDs and `sqlLiteral()` escaping. State writes use atomic tmp+rename. Async/await patterns are correct.

**Guideline issues**: 4 naming nits (abbreviations vs CLAUDE.md naming convention).

**Error handling**: 3 observability gaps in teardown, cache validation, and container probe paths.

Overall this is solid, well-structured code. No blocking issues.

test-suite/fhevm/src/generate/env.ts
```diff
@@ -0,0 +1,340 @@
...
+  const envs = structuredClone(templateEnvs);
+  applyTopologyEnv(envs, plan);
+  applyBaseRuntimeEnv(envs, state);
+  applyCompatEnv(envs, plan);
+  applyDiscoveryEnv(envs, state, plan);
+  const instanceEnvs = await buildInstanceEnvs(envs, plan, deriveWallet);
+
+  // Uniform per-chain gateway-sc indexed vars for ALL host chains.
+  envs["gateway-sc"].NUM_HOST_CHAINS = String(plan.hostChains.length);
+  for (let ci = 0; ci < plan.hostChains.length; ci += 1) {
```
Eikix: [AUTOMATED] Nit: `ci` abbreviates "chain index" — the second loop (line 279) derives `const chainIndex = ci + 1`, which signals `chainIndex` is the right name for the loop variable itself.

CLAUDE.md: "Avoid abbreviations or ambiguous names — use descriptive names that say what the thing does."

Suggestion: rename `ci` → `chainIndex` in both loops.

Confidence: 92/100

test-suite/fhevm/src/cli.ts
```diff
@@ -0,0 +1,284 @@
...
+  up,
+  upDryRun,
+  upgrade,
+} from "./flow/up-flow";
+import { parseUpInput } from "./input/up";
+import { asBool, asString } from "./input/shared";
+const HELP_FLAGS = new Set(["--help", "-h"]);
+
+/** Collects the accepted flag spellings for one command definition. */
+const commandOptionForms = async (cmd: any) => {
```
Eikix: [AUTOMATED] Nit: `cmd: any` — citty exports `CommandDef<ArgsDef>` which is the correct type here (and at lines 47, 242, 244). These functions access `.args` and `.subCommands`, both typed on `CommandDef<T>`. With strict mode enabled, the explicit `any` opts out of type checking unnecessarily.

Confidence: 82/100

test-suite/fhevm/src/flow/up-flow.ts
```diff
@@ -0,0 +1,1934 @@
...
+  } catch {
+    console.log(`[warn] compose down failed for ${component}`);
+    return false;
+  }
+};
+
+const removeProjectResources = async (kind: "container" | "volume" | "network", format: string) => {
+  const listed = await run(
+    ["docker", kind, "ls", ...(kind === "container" ? ["-a"] : []), "--filter", `label=com.docker.compose.project=${PROJECT}`, "--format", format],
+    { allowFailure: true },
```
Eikix: [AUTOMATED] `removeProjectResources`: when `docker <kind> ls` fails (code !== 0), the function returns silently with no log (line 960). The removal command (line 972) also uses `allowFailure: true` with its result discarded — partial removal of containers/volumes/networks is invisible.

Contrast with `composeDown` nearby which logs `[warn] compose down failed for ${component}` on failure.

Suggestion: log a warning when listing fails or removal partially fails, so `down`/`clean` don't silently leave orphaned resources.

Confidence: 90/100

test-suite/fhevm/src/resolve/bundle-store.ts
```diff
@@ -0,0 +1,163 @@
...
+  if (!options.reset) {
+    try {
+      const cached = await readJson<VersionBundle>(cachePath);
+      if (cached.target === options.target) {
+        console.log(`[resolve] using cached ${options.target} bundle`);
+        const validated = validateLockBundleShape(cached);
+        return await validateLockedRuntimeCompat(validated);
+      }
+    } catch {
+      // cache miss
```
Eikix: [AUTOMATED] The `// cache miss` catch block (lines 130-132) also catches `GitHubApiError` from `validateLockBundleShape` when the cached JSON is structurally corrupt. The fallback to a live fetch is correct behavior, but the silent swallow makes it impossible to distinguish a normal cache miss from a corrupt cache file.

Suggestion: narrow the catch to file-not-found, or log a warning for non-ENOENT errors:
```ts
} catch (error) {
  if ((error as NodeJS.ErrnoException).code !== 'ENOENT') {
    console.log(`[resolve] warn: cache invalid (${error instanceof Error ? error.message : error}), re-resolving`);
  }
}
```

Confidence: 85/100

test-suite/fhevm/src/flow/up-flow.ts
```diff
@@ -0,0 +1,1934 @@
...
+      throw new ProbeTimeout(container, 180);
+    }
+    await Bun.sleep(2_000);
+  }
+};
+
+/** Waits until container logs contain the requested pattern. */
+export const waitForLog = async (container: string, pattern: RegExp) => {
+  for (let attempt = 0; attempt <= 90; attempt += 1) {
+    const logs = await run(["docker", "logs", container], { allowFailure: true });
```
Eikix: [AUTOMATED] `waitForLog`: when the container is removed or crashes between the `waitForContainer` check and this polling loop, `docker logs` returns code !== 0 with `allowFailure: true`, so the error is absorbed. The loop spins for ~3 minutes before throwing `ProbeTimeout` — which says "not ready after 180s" rather than "container is gone."

Contrast with `waitForContainer` (line 641) which inspects container state and throws `ContainerCrashed` immediately.

Suggestion: check `logs.code !== 0` and either throw immediately or add a `docker inspect` check inside the loop.

Confidence: 92/100

tawadaa (dismissed):
LGTM, excellent work, thanks!

test-suite/fhevm/src/flow/repair.ts
```diff
@@ -0,0 +1,129 @@
...
+export const resumeSteadyStateServices = (state: State) => {
+  const chains = hostChainsForState(state);
+  const chainSuffixes = chains.map((chain) => chain.suffix);
+  const topology = topologyForState(state);
+  return {
+    "base": ["fhevm-minio", "coprocessor-and-kms-db", KMS_CORE_CONTAINER, "gateway-node", ...chains.map((chain) => chain.node)],
+    "coprocessor": chainSuffixes.flatMap((suffix) =>
+      Array.from({ length: topology.count }, (_, index) => {
+        const prefix = index === 0 ? "coprocessor-" : `coprocessor${index}-`;
+        return COPROCESSOR_RUNTIME_SUFFIXES.map((service) => `${prefix}${service}${suffix}`);
```
Eikix: [AUTOMATED] **Bug (confidence 98/100)**: For multi-chain scenarios, `chainSuffixes` includes entries like `"-chain-b"`. This line applies *all* `COPROCESSOR_RUNTIME_SUFFIXES` (7 services) to every chain suffix, but `buildExtraCoprocessorListenerOverride` in `generate/compose.ts` only creates `host-listener` and `host-listener-poller` for extra chains.

This generates phantom names like `coprocessor-tfhe-worker-chain-b` that never exist as containers. `resumeRepairStep` will always think coprocessor needs repair on multi-chain setups, causing `up --resume` to re-run the entire coprocessor pipeline even when the stack is healthy.

**Suggested fix**: Only apply extra-chain suffixes to listener services:
```typescript
const LISTENER_ONLY_SUFFIXES = ["host-listener", "host-listener-poller"];

"coprocessor": [
  // Default chain: all runtime services
  ...Array.from({ length: topology.count }, (_, index) => {
    const prefix = index === 0 ? "coprocessor-" : `coprocessor${index}-`;
    return COPROCESSOR_RUNTIME_SUFFIXES.map((s) => `${prefix}${s}`);
  }).flat(),
  // Extra chains: only listener services
  ...chainSuffixes.filter(Boolean).flatMap((suffix) =>
    Array.from({ length: topology.count }, (_, index) => {
      const prefix = index === 0 ? "coprocessor-" : `coprocessor${index}-`;
      return LISTENER_ONLY_SUFFIXES.map((s) => `${prefix}${s}${suffix}`);
    }).flat(),
  ),
],
```

test-suite/fhevm/src/generate/env.ts
```diff
@@ -0,0 +1,356 @@
...
+export type WalletMaterial = {
+  address: string;
+  privateKey: string;
+};
+
+
+const HAS_PLACEHOLDER = /(?<!\$)\$\{[A-Z0-9_]+\}/;
+
+/** Resolves env interpolation until placeholders disappear or a fixed point is reached. */
+export const resolveEnvMap = (env: Record<string, string>) => {
```
Eikix: [AUTOMATED] **Nit: Guideline (confidence 92/100)**: `resolveEnvMap` implements a multi-pass fixed-point resolution loop (up to 4 passes) over `${VAR}` placeholders. The project guidelines state: *"No multi-pass interpolation, YAML-rewriting engines."*

Docker Compose resolves `${VAR}` natively from env files via `--env-file`. Could the template env files be restructured to avoid chained variable references, so each value is either a literal or resolved imperatively in a single pass by the caller?

Non-blocking — flagging for design discussion.

test-suite/fhevm/src/utils/process.ts
```diff
@@ -0,0 +1,179 @@
...
+    });
+  } catch (error) {
+    if (options.allowFailure) {
+      return 1;
+    }
+    throw new CommandError(argv, 1, error instanceof Error ? error.message : String(error));
+  }
+  const code = await proc.exited;
+  if (code !== 0 && !options.allowFailure) {
+    throw new CommandError(argv, code, "");
```
Eikix: [AUTOMATED] **Error handling (confidence 95/100)**: `runStreaming` throws `new CommandError(argv, code, "")` with an empty stderr string. The resulting error message is just `"docker compose ... failed (1)"` with no diagnostic content.

Since stderr is inherited (streams to the terminal live), the actual error is visible while running but lost from the `CommandError` that propagates to the top-level catch. Users must scroll up to find the real error.

Consider including a hint: `"(see output above)"` or capturing stderr alongside streaming, similar to how `runWithHeartbeat` handles this.

test-suite/fhevm/src/commands/test.ts
```diff
@@ -0,0 +1,789 @@
...
+      });
+    };
+
+    const pauseScope = PAUSE_PROFILE_SCOPE[name];
+    if (pauseScope) {
+      await pause(pauseScope);
+      try {
+        return await runGrep();
+      } finally {
+        await unpause(pauseScope).catch(() => undefined);
```
Eikix: [AUTOMATED] **Error handling (confidence 95/100)**: `unpause` failure is silently swallowed with `.catch(() => undefined)`. If `unpause` fails (container unreachable, contract call reverts), contracts are left in a paused state with no warning. Subsequent test runs will fail with cryptic revert errors.

**Suggested fix**: At minimum log the failure:
```typescript
await unpause(pauseScope).catch((e) =>
  console.error(`[warn] unpause ${pauseScope} failed: ${e instanceof Error ? e.message : String(e)}`)
);
```

test-suite/fhevm/src/commands/test.ts
```diff
@@ -0,0 +1,789 @@
...
+        }
+      }
+      if (attempt === attempts) {
+        throw new PreflightError(`timed out waiting for drift injection trigger to fire in ${dbName}`);
+      }
+      await Bun.sleep(options.pollIntervalSeconds * 1000);
+    }
+    throw new PreflightError(`timed out waiting for drift injection trigger to fire in ${dbName}`);
+  } finally {
+    await psql(dbName, [], options, DRIFT_CLEANUP_SQL).catch(() => undefined);
```
Eikix: [AUTOMATED] **Error handling (confidence 92/100)**: Drift cleanup SQL failure is silently swallowed with `.catch(() => undefined)`. If this fails (DB connection issue, permission error), the drift injection state table and trigger remain installed in the coprocessor database, potentially causing spurious drift detection in future test runs.

**Suggested fix**:
```typescript
await psql(dbName, [], options, DRIFT_CLEANUP_SQL).catch((e) =>
  console.log(`[warn] drift cleanup failed: ${e instanceof Error ? e.message : String(e)}`)
);
```

Eikix (commented):
[AUTOMATED]

## Code Review — PR #2134

Reviewed for bugs, guideline compliance, and error handling.

### Summary

- **1 bug** in multi-chain resume logic (phantom container names cause unnecessary re-runs)
- **1 guideline note** about the multi-pass env interpolation engine
- **4 error handling issues** where silent `.catch(() => undefined)` or bare `catch {}` swallow diagnostics

Overall this is a strong, well-structured rewrite. The issues below are all fixable without major refactoring.

test-suite/fhevm/src/flow/runtime-compose.ts
```diff
@@ -0,0 +1,330 @@
...
+      env: await composeEnv(component),
+      allowFailure: true,
+    });
+    if (code !== 0) {
+      console.log(`[warn] compose down failed for ${component} (${code})`);
+      return false;
+    }
+    return true;
+  } catch {
+    console.log(`[warn] compose down failed for ${component}`);
```
Eikix: [AUTOMATED] **Error handling (confidence 90/100)**: Bare `catch {}` with no error binding discards all diagnostic information. If `composeEnv` throws (missing env file, JSON parse error) or `runStreaming` fails to spawn, the log shows only `"compose down failed for <component>"` with no reason.

**Suggested fix**: Include the error message:
```typescript
} catch (error) {
  console.log(`[warn] compose down failed for ${component}: ${error instanceof Error ? error.message : String(error)}`);
  return false;
}
```

Same pattern applies to the similar catch block in `multiChainComposeDown` (line 293).

tawadaa (approved):

Eikix:
@mergify queue

Eikix:
@mergifyio queue

Eikix:
@mergifyio refresh

Eikix:
@mergifyio dequeue

Eikix:
@mergifyio queue

Eikix:
@mergifyio queue

Eikix:
@mergifyio queue

eudelins-zama:
@mergifyio refresh

eudelins-zama (approved):

eudelins-zama:
@mergifyio refresh

PanGan21 (approved):

PR #2135, opened 2026-03-19
feat(test-suite): add multi-chain e2e testing infrastructure

PanGan21:
## Summary
- Add multi-chain deploy support (`--multi-chain` flag) that spins up a second host chain (Chain B) using parameterized docker compose files
- Register Chain B in the coprocessor `host_chains` table and restart the zkproof-worker so it can resolve Chain B's ACL address
- Add 7 multi-chain state isolation e2e tests: ciphertext handle isolation, ACL permission isolation, block reorg isolation, chain halt isolation, simultaneous decryption, random number independence, and cross-chain user input
- Propagate KMS signer address to Chain B host-sc env files

Closes: https://github.com/zama-ai/fhevm-internal/issues/1163


PanGan21:
Multi chain E2E tests passing: https://github.com/zama-ai/fhevm/actions/runs/23339565119/job/67893253000

PanGan21:
Old single chain e2e tests still pass: https://github.com/zama-ai/fhevm/actions/runs/23341768810/job/67896862476

antoniupop:
Does this work with multi-chain and multi-copro? It looks like it would just set up multi-chain in one coprocessor no? `run_additional_coprocessor_instance` won't register this additional chain. 

Also not sure why the name `cleanup_single_step` (of course was there, not added in this PR - thought this was actually cleanup at first.

PanGan21:
> Does this work with multi-chain and multi-copro? It looks like it would just set up multi-chain in one coprocessor no? `run_additional_coprocessor_instance` won't register this additional chain.
> 
> Also not sure why the name `cleanup_single_step` (of course was there, not added in this PR - thought this was actually cleanup at first.

Good catch. Currently `--multi-chain` only registers Chain B in the primary coprocessor's database and only starts Chain B host-listeners for the primary instance. I will address this by making it work with multi coprocessor setup and renaming `cleanup_single_step` to something more meaningful.

Just to clarify multi chain doesn't necessarily mean multi coprocessor right? The infrastructure can be one coprocessor with a host listener for each chain or multiple coprocessors and each one with a host listener for each chain. Is this correct?

antoniupop:
> Just to clarify multi chain doesn't necessarily mean multi coprocessor right? The infrastructure can be one coprocessor with a host listener for each chain or multiple coprocessors and each one with a host listener for each chain. Is this correct?

Absolutely - multi-copro is for security (run multiple instances and validate consensus e.g. 3 out of 5) and is orthogonal to multi-chain. Each copro has its own listeners for each chain (they're completely independent)

PanGan21:
Closing in favor of: https://github.com/zama-ai/fhevm/pull/2150

PR #2137, opened 2026-03-19
fix(gateway-contracts):  overload `isUserDecryptionReady` with old signature

zmalatrax:
For v0.12, with the removal of `MultichainACL.sol` we decided to update the signature of the view function `isUserDecryptionReady`, from `isUserDecryptionReady(address, (bytes32, address)[], bytes)` to `isUserDecryptionReady((bytes32, address)[], bytes)`. See https://github.com/zama-ai/fhevm/pull/1904

Though, it actually breaks the relayer `main` until the upgrade is done on testnet and mainnet ; as updating the rust bindings `fhevm_gateway_bindings` would remove the `isUserDecryptionReady(address, (bytes32, address)[], bytes)` selector currently deployed on testnet and mainnet.

Proposed solution is to overload the `isUserDecryptionReady` view function with its old signature, with a deprecation notice (custom label as not supported by NatSpec), which would allow backward compatibility during the upgrade window.

Then, once mainnet has been upgraded, the old signature overload can be removed in the next contract upgrade


jatZama (dismissed):

gateway-contracts/contracts/Decryption.sol
```diff
@@ -681,7 +681,7 @@ contract Decryption is
     function isUserDecryptionReady(
         CtHandleContractPair[] calldata ctHandleContractPairs,
         bytes calldata /* extraData */
-    ) external view virtual returns (bool) {
+    ) public view virtual returns (bool) {
```
isaacdecoded, line 684: Just wondering why this has been changed to `public`?

zmalatrax: The overloaded function `isUserDecryptionReady` (old signature) is calling the `isUserDecryptionReady` (new signature), so the called function must be `public`

leventdem (approved):
OK

zmalatrax:
@mergify queue

PR #2138, opened 2026-03-19
ci(kms-connector): fix check-changes of bindings

eudelins-zama:
Fix invalid path of bindings for `check-changes` block in `kms-connector`'s CI


isaacdecoded (approved):
Good catch!

eudelins-zama:
@mergifyio queue

PR #2146, opened 2026-03-20
feat(host-contracts): add executor prepare-upgrade task

Eikix:
## Summary
- add a host-contracts task to deploy a new `FHEVMExecutor` implementation without upgrading the proxy
- keep the existing deploy/upgrade tasks unchanged
- keep the new implementation source in the checked-out tree and use `--upgrade-from-ref` only for old-side OZ validation
- auto-create `host-contracts/addresses` in the existing setter tasks so the prepare flow does not fail on a fresh checkout
- document the expected setup and command shape in `host-contracts/README.md`

## Validation
- `npx --no-install hardhat help task:prepareUpgradeFHEVMExecutor`
- `git diff --check`

## Notes
- this is additive only and is meant to support testnet/mainnet DAO upgrade flows
- the task still expects the checked-out tree to contain the exact new host-contract code being deployed
- the generated `addresses/.env.host` and `addresses/FHEVMHostAddresses.sol` must be populated before running the task

Closes https://github.com/zama-ai/fhevm-internal/issues/1177


host-contracts/tasks/taskDeploy.ts
```diff
@@ -8,6 +8,10 @@ import path from 'path';
 
 import { getRequiredEnvVar } from './utils/loadVariables';
 
+function ensureAddressesDirectoryExists() {
+  fs.mkdirSync(path.join(__dirname, '../addresses'), { recursive: true });
```
Eikix: why here the address directory is under `__dirname, '../addresses'`
but below, we see `const envFilePath = path.join(__dirname, '../addresses/.env.host');`

something is weird / hard to reason about

host-contracts/tasks/upgradeContracts.ts
```diff
@@ -67,6 +100,56 @@ async function upgradeCurrentToNew(
   }
 }
 
+async function prepareNewImplementation(
```
Eikix: this is hardly a `prepare`, it's a deploy
it should be called deployNewImplementation and in the doc comments, we should say that it is specifically meant for upgrades where we can't call directly the UUPS upgrade entrypoint in case the dao holds the rights to upgrade.

you can succinctly take some of those precisions if they already in readme, if not, then dont

Eikix:
@claude /pr-review

tawadaa (dismissed):

antoniupop (dismissed):

antoniupop (approved):

Eikix:
@mergifyio queue

Eikix:
@mergifyio refresh

leventdem (approved):
OK.

PR #2150, opened 2026-03-23
feat(test-suite): add multi-chain e2e testing support

PanGan21:
## Summary

- Adds multi-chain (2 host chains) e2e testing to the TypeScript CLI, with full support for multi-coprocessor scenarios
- Introduces 7 multi-chain isolation tests covering cross-chain ciphertext isolation, ACL isolation, independent decryption, chain halt resilience, and random number independence
- Extends compose generation, env rendering, relayer config, and stack orchestration to deploy a second host chain (Chain B) with its own smart contracts and coprocessor listeners

## Details

**New scenarios:**
- `multi-chain.yaml` — 1 coprocessor, 2 host chains
- `two-of-two-multi-chain.yaml` — 2 coprocessors, 2 host chains

Closes: https://github.com/zama-ai/fhevm-internal/issues/1163

### DIFF
--- a/.github/workflows/test-suite-e2e-tests.yml
+++ b/.github/workflows/test-suite-e2e-tests.yml
@@ -76,6 +76,10 @@ on:
         description: "KMS Core version"
         default: ""
         type: string
+      multi-chain:
+        description: "Enable multi-chain mode (deploys a second host chain)"
+        default: false
+        type: boolean
   workflow_call:
     secrets:
       GHCR_READ_TOKEN:
@@ -176,8 +180,14 @@ jobs:
 
       - name: Dry run
         working-directory: test-suite/fhevm
+        env:
+          MULTI_CHAIN: ${{ inputs.multi-chain && 'true' || 'false' }}
         run: |
-          args=(--target latest-main --scenario two-of-two)
+          scenario="two-of-two"
+          if [ "$MULTI_CHAIN" = "true" ]; then
+            scenario="two-of-two-multi-chain"
+          fi
+          args=(--target latest-main --scenario "$scenario")
           if [ "$BUILD" = "true" ]; then
             args+=(--build)
           fi
@@ -185,8 +195,14 @@ jobs:
 
       - name: Boot fhevm Stack
         working-directory: test-suite/fhevm
+        env:
+          MULTI_CHAIN: ${{ inputs.multi-chain && 'true' || 'false' }}
         run: |
-          args=(--target latest-main --scenario two-of-two)
+          scenario="two-of-two"
+          if [ "$MULTI_CHAIN" = "true" ]; then
+            scenario="two-of-two-multi-chain"
+          fi
+          args=(--target latest-main --scenario "$scenario")
           if [ "$BUILD" = "true" ]; then
             args+=(--build)
           fi
@@ -196,6 +212,18 @@ jobs:
         working-directory: test-suite/fhevm
         run: |
           ./fhevm-cli test standard
+
+      - name: Heavy e2e suite
+        working-directory: test-suite/fhevm
+        run: |
+          # Heavy is the long-running operators lane.
+          ./fhevm-cli test heavy
+
+      - name: Multi-chain isolation tests
+        if: ${{ inputs.multi-chain }}
+        working-directory: test-suite/fhevm
+        run: |
+          ./fhevm-cli test multi-chain-isolation
 
       - name: Show logs
         working-directory: test-suite/fhevm
--- a/.github/workflows/test-suite-e2e-tests.yml
+++ b/.github/workflows/test-suite-e2e-tests.yml
@@ -76,10 +76,10 @@ on:
         description: "KMS Core version"
         default: ""
         type: string
-      multi-chain:
-        description: "Enable multi-chain mode (deploys a second host chain)"
-        default: false
-        type: boolean
+      scenario:
+        description: "Deployment scenario (e.g. two-of-two, two-of-two-multi-chain)"
+        default: "two-of-two"
+        type: string
   workflow_call:
     secrets:
       GHCR_READ_TOKEN:
@@ -125,6 +125,7 @@ jobs:
       TEST_SUITE_VERSION: ${{ inputs.test-suite-version }}
       RELAYER_VERSION: ${{ inputs.relayer-version }}
       CORE_VERSION: ${{ inputs.kms-core-version }}
+      SCENARIO: ${{ inputs.scenario || 'two-of-two' }}
     runs-on: large_ubuntu_32
     steps:
       - name: Checkout code
@@ -180,14 +181,8 @@ jobs:
 
       - name: Dry run
         working-directory: test-suite/fhevm
-        env:
-          MULTI_CHAIN: ${{ inputs.multi-chain && 'true' || 'false' }}
         run: |
-          scenario="two-of-two"
-          if [ "$MULTI_CHAIN" = "true" ]; then
-            scenario="two-of-two-multi-chain"
-          fi
-          args=(--target latest-main --scenario "$scenario")
+          args=(--target latest-main --scenario "$SCENARIO")
           if [ "$BUILD" = "true" ]; then
             args+=(--build)
           fi
@@ -195,14 +190,8 @@ jobs:
 
       - name: Boot fhevm Stack
         working-directory: test-suite/fhevm
-        env:
-          MULTI_CHAIN: ${{ inputs.multi-chain && 'true' || 'false' }}
         run: |
-          scenario="two-of-two"
-          if [ "$MULTI_CHAIN" = "true" ]; then
-            scenario="two-of-two-multi-chain"
-          fi
-          args=(--target latest-main --scenario "$scenario")
+          args=(--target latest-main --scenario "$SCENARIO")
           if [ "$BUILD" = "true" ]; then
             args+=(--build)
           fi
@@ -212,18 +201,6 @@ jobs:
         working-directory: test-suite/fhevm
         run: |
           ./fhevm-cli test standard
-
-      - name: Heavy e2e suite
-        working-directory: test-suite/fhevm
-        run: |
-          # Heavy is the long-running operators lane.
-          ./fhevm-cli test heavy
-
-      - name: Multi-chain isolation tests
-        if: ${{ inputs.multi-chain }}
-        working-directory: test-suite/fhevm
-        run: |
-          ./fhevm-cli test multi-chain-isolation
 
       - name: Show logs
         working-directory: test-suite/fhevm
--- a/test-suite/e2e/test/multiChain/multiChainHelper.ts
+++ b/test-suite/e2e/test/multiChain/multiChainHelper.ts
@@ -1,0 +1,127 @@
+import { ethers as hardhatEthers } from 'hardhat';
+import { ethers } from 'ethers';
+import { createInstance as createFhevmInstance } from '@zama-fhe/relayer-sdk/node';
+import { vars } from 'hardhat/config';
+
+const defaultMnemonic =
+  'adapt mosquito move limb mobile illegal tree voyage juice mosquito burger raise father hope layer';
+const mnemonic: string = process.env.MNEMONIC ?? vars.get('MNEMONIC', defaultMnemonic);
+
+const aclAddress = process.env.ACL_CONTRACT_ADDRESS!;
+const kmsVerifierAddress = process.env.KMS_VERIFIER_CONTRACT_ADDRESS!;
+const inputVerifierAddress = process.env.INPUT_VERIFIER_CONTRACT_ADDRESS!;
+const decryptionAddress = process.env.DECRYPTION_ADDRESS!;
+const inputVerificationAddress = process.env.INPUT_VERIFICATION_ADDRESS!;
+const relayerUrl = process.env.RELAYER_URL!;
+const gatewayChainId = Number(process.env.CHAIN_ID_GATEWAY!);
+
+export interface ChainConfig {
+  rpcUrl: string;
+  chainId: number;
+}
+
+export const CHAIN_A: ChainConfig = {
+  rpcUrl: process.env.RPC_URL || 'http://localhost:8545',
+  chainId: Number(process.env.CHAIN_ID_HOST || 12345),
+};
+
+export const CHAIN_B: ChainConfig = {
+  rpcUrl: process.env.RPC_URL_CHAIN_B || 'http://localhost:8547',
+  chainId: Number(process.env.CHAIN_ID_HOST_B || 67890),
+};
+
+const providers = new Map<string, ethers.JsonRpcProvider>();
+
+export function getProvider(chain: ChainConfig): ethers.JsonRpcProvider {
+  if (!providers.has(chain.rpcUrl)) {
+    providers.set(chain.rpcUrl, new ethers.JsonRpcProvider(chain.rpcUrl));
+  }
+  return providers.get(chain.rpcUrl)!;
+}
+
+export type ManagedWallet = ethers.NonceManager & { address: string; reset: () => void };
+
+function wrapWithNonceManager(wallet: ethers.Wallet): ManagedWallet {
+  const nm = new ethers.NonceManager(wallet);
+  (nm as any).address = wallet.address;
+  return nm as ManagedWallet;
+}
+
+export interface NamedSigners {
+  alice: ManagedWallet;
+  bob: ManagedWallet;
+  carol: ManagedWallet;
+  dave: ManagedWallet;
+  eve: ManagedWallet;
+}
+
+const signersCache = new Map<string, NamedSigners>();
+
+export function getSigners(chain: ChainConfig): NamedSigners {
+  if (!signersCache.has(chain.rpcUrl)) {
+    const provider = getProvider(chain);
+    const hdNode = ethers.HDNodeWallet.fromMnemonic(
+      ethers.Mnemonic.fromPhrase(mnemonic),
+      "m/44'/60'/0'/0",
+    );
+    const mkWallet = (i: number) => wrapWithNonceManager(
+      new ethers.Wallet(hdNode.deriveChild(i).privateKey, provider),
+    );
+    signersCache.set(chain.rpcUrl, {
+      alice: mkWallet(0),
+      bob: mkWallet(1),
+      carol: mkWallet(2),
+      dave: mkWallet(3),
+      eve: mkWallet(4),
+    });
+  }
+  return signersCache.get(chain.rpcUrl)!;
+}
+
+export function getWallet(chain: ChainConfig, index: number): ManagedWallet {
+  const provider = getProvider(chain);
+  const hdNode = ethers.HDNodeWallet.fromMnemonic(
+    ethers.Mnemonic.fromPhrase(mnemonic),
+    "m/44'/60'/0'/0",
+  );
+  return wrapWithNonceManager(
+    new ethers.Wallet(hdNode.deriveChild(index).privateKey, provider),
+  );
+}
+
+export async function createInstance(chain: ChainConfig) {
+  return createFhevmInstance({
+    verifyingContractAddressDecryption: decryptionAddress,
+    verifyingContractAddressInputVerification: inputVerificationAddress,
+    kmsContractAddress: kmsVerifierAddress,
+    inputVerifierContractAddress: inputVerifierAddress,
+    aclContractAddress: aclAddress,
+    network: chain.rpcUrl,
+    relayerUrl,
+    gatewayChainId,
+    chainId: chain.chainId,
+  });
+}
+
+export async function deployContract(
+  contractName: string,
+  deployer: ethers.Signer,
+  ...constructorArgs: unknown[]
+): Promise<ethers.Contract> {
+  const artifact = await hardhatEthers.getContractFactory(contractName);
+  const factory = new ethers.ContractFactory(artifact.interface, artifact.bytecode, deployer);
+  const contract = await factory.deploy(...constructorArgs, { gasLimit: 10_000_000 });
+  await contract.waitForDeployment();
+  return contract as ethers.Contract;
+}
+
+export async function evmSnapshot(provider: { send: (method: string, params: unknown[]) => Promise<unknown> }): Promise<string> {
+  return provider.send('evm_snapshot', []) as Promise<string>;
+}
+
+export async function evmRevert(
+  provider: { send: (method: string, params: unknown[]) => Promise<unknown> },
+  snapshotId: string,
+): Promise<boolean> {
+  return provider.send('evm_revert', [snapshotId]) as Promise<boolean>;
+}
--- a/test-suite/e2e/test/multiChain/multiChainHelper.ts
+++ b/test-suite/e2e/test/multiChain/multiChainHelper.ts
@@ -54,12 +54,6 @@ function parseHostChains(): ChainConfig[] {
 /** All discovered host chains. Index 0 is the primary chain. */
 export const HOST_CHAINS: ChainConfig[] = parseHostChains();
 
-/** Primary host chain (always present). */
-export const CHAIN_A: ChainConfig = HOST_CHAINS[0];
-
-/** First extra host chain, or undefined in single-chain mode. Guard with `if (!CHAIN_B) this.skip()`. */
-export const CHAIN_B: ChainConfig | undefined = HOST_CHAINS[1];
-
 const providers = new Map<string, ethers.JsonRpcProvider>();
 
 export function getProvider(chain: ChainConfig): ethers.JsonRpcProvider {
--- a/test-suite/e2e/test/multiChain/multiChainHelper.ts
+++ b/test-suite/e2e/test/multiChain/multiChainHelper.ts
@@ -20,15 +20,36 @@ export interface ChainConfig {
   chainId: number;
 }
 
-export const CHAIN_A: ChainConfig = {
-  rpcUrl: process.env.RPC_URL || 'http://localhost:8545',
-  chainId: Number(process.env.CHAIN_ID_HOST || 12345),
-};
+function requireEnv(name: string): string {
+  const value = process.env[name];
+  if (!value) throw new Error(`Missing required env var: ${name}`);
+  return value;
+}
 
-export const CHAIN_B: ChainConfig = {
-  rpcUrl: process.env.RPC_URL_CHAIN_B || 'http://localhost:8547',
-  chainId: Number(process.env.CHAIN_ID_HOST_B || 67890),
-};
+/** Discovers all host chains from indexed env vars (HOST_CHAIN_1_*, HOST_CHAIN_2_*, …). */
+function parseHostChains(): ChainConfig[] {
+  const primary: ChainConfig = {
+    rpcUrl: requireEnv('RPC_URL'),
+    chainId: Number(requireEnv('CHAIN_ID_HOST')),
+  };
+  const chains: ChainConfig[] = [primary];
+  for (let i = 1; ; i++) {
+    const rpcUrl = process.env[`HOST_CHAIN_${i}_RPC_URL`];
+    const chainId = process.env[`HOST_CHAIN_${i}_CHAIN_ID`];
+    if (!rpcUrl || !chainId) break;
+    chains.push({ rpcUrl, chainId: Number(chainId) });
+  }
+  return chains;
+}
+
+/** All discovered host chains. Index 0 is the primary chain. */
+export const HOST_CHAINS: ChainConfig[] = parseHostChains();
+
+/** Primary host chain (always present). */
+export const CHAIN_A: ChainConfig = HOST_CHAINS[0];
+
+/** First extra host chain, or undefined in single-chain mode. Guard with `if (!CHAIN_B) this.skip()`. */
+export const CHAIN_B: ChainConfig | undefined = HOST_CHAINS[1];
 
 const providers = new Map<string, ethers.JsonRpcProvider>();
 
--- a/test-suite/e2e/test/multiChain/multiChainHelper.ts
+++ b/test-suite/e2e/test/multiChain/multiChainHelper.ts
@@ -7,9 +7,6 @@ const defaultMnemonic =
   'adapt mosquito move limb mobile illegal tree voyage juice mosquito burger raise father hope layer';
 const mnemonic: string = process.env.MNEMONIC ?? vars.get('MNEMONIC', defaultMnemonic);
 
-const aclAddress = process.env.ACL_CONTRACT_ADDRESS!;
-const kmsVerifierAddress = process.env.KMS_VERIFIER_CONTRACT_ADDRESS!;
-const inputVerifierAddress = process.env.INPUT_VERIFIER_CONTRACT_ADDRESS!;
 const decryptionAddress = process.env.DECRYPTION_ADDRESS!;
 const inputVerificationAddress = process.env.INPUT_VERIFICATION_ADDRESS!;
 const relayerUrl = process.env.RELAYER_URL!;
@@ -18,6 +15,9 @@ const gatewayChainId = Number(process.env.CHAIN_ID_GATEWAY!);
 export interface ChainConfig {
   rpcUrl: string;
   chainId: number;
+  aclAddress: string;
+  kmsVerifierAddress: string;
+  inputVerifierAddress: string;
 }
 
 function requireEnv(name: string): string {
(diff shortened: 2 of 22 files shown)


test-suite/e2e/test/multiChain/multiChainIsolation.ts
```diff
@@ -0,0 +1,335 @@
...
+      expect(handleA).to.not.eq(handleB);
+
+      try {
+        const crossChainTx = await (erc20B.connect(this.deployerB) as EncryptedERC20)[
+          'transfer(address,bytes32,bytes)'
+        ](this.signersB.bob.address, handleA, '0x', { gasLimit: 10_000_000 });
+        await crossChainTx.wait();
+      } catch {
+        // Expected: cross-chain handle should be rejected
+      }
```
PanGan21, line 141: [AUTOMATED] **Bug: Test passes whether cross-chain isolation holds or not.**

The `try/catch` at line 131-138 swallows the error if the cross-chain transfer reverts (expected), but also lets execution continue silently if it **succeeds** (isolation broken). The only assertion after the catch (line 140-141) checks Chain A's deployer balance — which is unaffected because the transfer targets Chain B's contract.

If isolation were broken and the transfer succeeded on Chain B, this test would still pass green.

Suggested fix — replace the bare try/catch with an explicit revert assertion:
```typescript
await expect(
  (erc20B.connect(this.deployerB) as EncryptedERC20)[
    'transfer(address,bytes32,bytes)'
  ](this.signersB.bob.address, handleA, '0x', { gasLimit: 10_000_000 })
).to.be.reverted;
```
Or add `expect.fail('cross-chain transfer should have reverted')` after `crossChainTx.wait()` inside the try block.

Confidence: 95/100

test-suite/fhevm/src/flow/up-flow.ts
```diff
@@ -988,6 +992,42 @@ const stepComposeUp = async (
...
+    );
+    return code === 0;
+  } catch {
+    return false;
+  }
+};
+
 /** Lists the coprocessor containers whose health determines coprocessor readiness. */
 const coprocessorHealthContainers = (state: Pick<State, "scenario">) => {
   const topology = topologyForState(state);
```
PanGan21, line 1086: [AUTOMATED] **Bug: `multiChainComposeDown` swallows errors + caller ignores return values.**

The bare `catch { return false; }` here discards all errors without logging. The caller at the `down` function (line ~1574) uses `Promise.all(...).map(multiChainComposeDown)` but discards the returned booleans — failures are never tracked in the `failed[]` array.

Compare with single-chain `composeDown` which logs `[warn] compose down failed for ${component}` and tracks failures. If any Chain B container fails to tear down, the user gets no warning and may hit port conflicts or stale state on the next `up`.

Suggested fix — either:
1. Log a warning in the catch block and track failures like `composeDown` does, or
2. Collect return values and push failed names into the `failed[]` array.

Confidence: 97/100

test-suite/fhevm/src/flow/up-flow.ts
```diff
@@ -988,6 +992,42 @@ const stepComposeUp = async (
   await composeUp(component, services, options);
```
PanGan21, line 1042: [AUTOMATED] **Bug: `resetAfterStep` does not tear down multi-chain containers.**

`resetAfterStep` iterates `COMPONENT_BY_STEP` to tear down containers when resuming with `--from-step`, but `COMPONENT_BY_STEP` in `layout.ts` does not include `host-node-b`, `host-sc-b`, or `coprocessor-host-b`.

When a user runs `--resume --from-step base` on a multi-chain stack, Chain B containers are left orphaned while base components they depend on are torn down and restarted. The full `down` function correctly handles this (line ~1574), but `resetAfterStep` does not.

Suggested fix — add multi-chain teardown to `resetAfterStep`, guarded by `stackSpecForState(state).multiChain`.

Confidence: 90/100

PanGan21 (commented):
[AUTOMATED]

## Code Review

Reviewed for bugs, guideline compliance, and error handling. Found 3 confirmed issues after validation (filtered from 10 initial candidates — 7 were false positives or not actionable).

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | High | `multiChainIsolation.ts` | Test passes regardless of whether cross-chain isolation holds |
| 2 | High | `up-flow.ts` | `resetAfterStep` does not tear down multi-chain containers |
| 3 | High | `up-flow.ts` | `multiChainComposeDown` failures silently discarded |

All three are actionable. See inline comments for details.

test-suite/e2e/test/multiChain/multiChainIsolation.ts
PanGan21: fixed

test-suite/fhevm/src/flow/up-flow.ts
PanGan21: fixed

test-suite/fhevm/src/flow/up-flow.ts
PanGan21: fixed

test-suite/fhevm/src/flow/up-flow.ts
```diff
@@ -1464,6 +1584,12 @@ export const down = async () => {
     await ensureRuntimeArtifacts(state, "teardown");
   }
   const failed: string[] = [];
+  const multiChainResults = await Promise.all(
+    ["coprocessor-host-b", "host-sc-b", "host-node-b"].map(async (name) => ({ name, ok: await multiChainComposeDown(name) })),
+  );
```
PanGan21: [AUTOMATED] **Bug: `down()` unconditionally tears down multi-chain containers, breaking single-chain deployments.**

The multi-chain teardown block runs for all deployments. For non-multi-chain deployments, the compose files don't exist (explicitly removed in `generateComposeOverrides`), so `docker compose` fails, returns `false`, gets pushed to `failed[]`, and throws `PreflightError`.

Compare with `resetAfterStep` which correctly guards with `if (state.scenario.multiChain)`.

Fix: wrap this block in `if (state?.scenario?.multiChain)`.

Confidence: 98/100

PanGan21 (commented):
[AUTOMATED]

## Code Review (re-review after fixes)

Re-reviewed for bugs, guideline compliance, and error handling. Found 1 confirmed issue after validation (filtered from 4 candidates — 3 were false positives).

Previous 3 issues from the first review have been fixed correctly.

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | High | `up-flow.ts` | `down()` unconditionally tears down multi-chain containers, breaking single-chain teardown |

test-suite/fhevm/src/flow/up-flow.ts
PanGan21: fixed

test-suite/fhevm/src/flow/up-flow.ts
```diff
@@ -476,9 +480,17 @@ const ensureRuntimeArtifacts = async (state: State, reason: string) => {
   await generateRuntime(state, stackSpecForState(state));
 };
 
-const resetAfterStep = async (step: StepName) => {
+const resetAfterStep = async (step: StepName, state: Pick<State, "scenario">) => {
   const start = stateStepIndex(step);
   const failed: string[] = [];
+  if (state.scenario.multiChain) {
+    const multiChainResults = await Promise.all(
+      ["coprocessor-host-b", "host-sc-b", "host-node-b"].map(async (name) => ({ name, ok: await multiChainComposeDown(name) })),
```
PanGan21: [AUTOMATED] **Bug: `resetAfterStep` unconditionally tears down ALL multi-chain containers regardless of `fromStep`**

When a user runs `--resume --from-step coprocessor`, `start` corresponds to the `coprocessor` step index. The main loop (line 494) correctly only tears down components at or after that step. However, the multi-chain block (lines 486-492) unconditionally destroys `host-node-b` (created during `base`), `host-sc-b` (created during `host-deploy`), and `coprocessor-host-b` (created during `coprocessor`) — all three, regardless of `start`.

This means `host-node-b` gets destroyed but the `base` step never re-executes, leaving Chain B's RPC node offline for the rest of the pipeline.

**Fix:** Map each multi-chain compose file to its parent pipeline step and filter by `stateStepIndex(parentStep) >= start` before tearing down:
```ts
const MULTI_CHAIN_STEP: Record<string, StepName> = {
  "host-node-b": "base",
  "host-sc-b": "host-deploy",
  "coprocessor-host-b": "coprocessor",
};

const toTearDown = Object.entries(MULTI_CHAIN_STEP)
  .filter(([, parentStep]) => stateStepIndex(parentStep) >= start)
  .map(([name]) => name);
```

Confidence: 95/100

PanGan21 (commented):
[AUTOMATED]

## Code Review (Round 3)

One confirmed bug found in `resetAfterStep` — multi-chain teardown is not step-aware.

Fix has been applied locally (not yet committed/pushed).

test-suite/fhevm/src/flow/up-flow.ts
PanGan21: fixed

test-suite/fhevm/src/layout.ts
```diff
@@ -237,6 +241,7 @@ export const hostAddressesSolidityPath = path.join(
   "host",
   "FHEVMHostAddresses.sol",
 );
+export const hostBAddressesPath = path.join(ADDRESS_DIR, "host-b", ".env.host-b");
```
PanGan21: [AUTOMATED] **Nit: `hostBAddressesPath` filename mismatch with deploy container output**

`hostBAddressesPath` points to `.env.host-b`, but the `host-contracts` deploy task always writes to `addresses/.env.host` (hardcoded in `taskDeploy.ts:278`). The compose volume remap (`/addresses/host:` → `/addresses/host-b:`) means the deployed file lands at `addresses/host-b/.env.host`, not `addresses/host-b/.env.host-b`.

Currently masked because:
1. `generateRuntime` writes an empty placeholder at this path during `base`
2. `discoverChainBContracts()` reads it back → empty `{}`
3. Fallback to Chain A addresses works because both chains use the same deterministic deployer → identical addresses

Would break if Chain A and Chain B ever produce different contract addresses.

**Fix:** `path.join(ADDRESS_DIR, "host-b", ".env.host")` to match the deploy container output.

Confidence: 95/100

test-suite/fhevm/src/generate/env.ts
```diff
@@ -243,6 +246,74 @@ export const renderEnvMaps = async (
...
   applyDiscoveryEnv(envs, state, plan);
   const instanceEnvs = await buildInstanceEnvs(envs, plan, deriveWallet);
+
+  if (plan.multiChain) {
+    const hostBHttp = `http://host-node-b:${CHAIN_B_PORT}`;
+    const hostBWs = `ws://host-node-b:${CHAIN_B_PORT}`;
+
+    envs["gateway-sc"].NUM_HOST_CHAINS = "2";
+    envs["gateway-sc"].HOST_CHAIN_CHAIN_ID_1 = CHAIN_B_ID;
+    envs["gateway-sc"].HOST_CHAIN_FHEVM_EXECUTOR_ADDRESS_1 =
```
PanGan21: [AUTOMATED] **Nit: Gateway-sc Chain B addresses copied from Chain A instead of `discovery.hostB`**

`HOST_CHAIN_FHEVM_EXECUTOR_ADDRESS_1` and `HOST_CHAIN_ACL_ADDRESS_1` are copied from `_0` (Chain A). Other multi-chain paths in this file correctly use `state.discovery.hostB` (e.g., line 301 for `kms-connector`). This is inconsistent and only works because deterministic addresses match.

Suggested fix (after fixing the `hostBAddressesPath` issue):
```ts
envs["gateway-sc"].HOST_CHAIN_FHEVM_EXECUTOR_ADDRESS_1 =
  state.discovery?.hostB?.FHEVM_EXECUTOR_CONTRACT_ADDRESS ??
  envs["gateway-sc"].HOST_CHAIN_FHEVM_EXECUTOR_ADDRESS_0 ?? "";
envs["gateway-sc"].HOST_CHAIN_ACL_ADDRESS_1 =
  state.discovery?.hostB?.ACL_CONTRACT_ADDRESS ??
  envs["gateway-sc"].HOST_CHAIN_ACL_ADDRESS_0 ?? "";
```

Confidence: 80/100

PanGan21 (commented):
[AUTOMATED]

## Code Review (Round 4)

3 confirmed issues found — all latent, currently masked by deterministic contract addresses but will break if Chain A/B ever diverge. Checked for bugs, guideline compliance, error handling, and silent failures.

### Issue not attached to a diff line:

**`ensureRuntimeArtifacts` doesn't track multi-chain files** (`up-flow.ts:455`)

`requiredPaths` doesn't include `hostBAddressesPath` or multi-chain compose/env files. If only multi-chain artifacts are deleted between runs, `allExist` returns `true` and `generateRuntime` is skipped, causing `multiChainComposeUp` to fail with missing file errors on resume.

Suggested fix: when `state.scenario.multiChain` is true, add multi-chain paths to `requiredPaths`.

**False positives filtered out**: zkproof-worker ordering (worker caches at startup, no race), mintTx.wait() (ethers v6 throws on revert), multiChainComposeDown error swallowing (callers do re-throw), env-without-discovery (renderEnvFile filters undefined values).

test-suite/fhevm/src/layout.ts
PanGan21: fixed

test-suite/fhevm/src/generate/env.ts
PanGan21: fixed

PanGan21:
[AUTOMATED]

## Code Review (Round 5)

No new runtime bugs found. Checked for bugs, guideline compliance, error handling, and silent failures.

**One additional latent issue (same class as round 4):** `coprocessor-b.X` env maps in `env.ts:287-296` inherit Chain A's contract addresses (`ACL_CONTRACT_ADDRESS`, `FHEVM_EXECUTOR_CONTRACT_ADDRESS`, `INPUT_VERIFIER_ADDRESS`) without overriding from `discovery.hostB`. Works today due to deterministic addresses — same class as the gateway-sc and hostBAddressesPath issues already flagged.

**False positives filtered out:** readEnvFile ENOENT (env files written before compose generation), ensureRuntimeArtifacts gating (regenerate always overwrites), teardown naming (convention applies to function names only), parallel teardown ordering (Docker handles graceful shutdown).

PanGan21:
[AUTOMATED]

## Code Review (Round 6 — Final)

No new issues found. All 11 previously reported issues are confirmed fixed in local code. Checked for bugs, guideline compliance, error handling, and silent failures across all files including YAML scenarios, CI workflow, test helpers, fixture code, and README.

The PR is clean and ready for merge once the fix commits are pushed.

test-suite/fhevm/src/generate/compose.ts
```diff
@@ -324,6 +325,99 @@ const buildComposeOverride = async (component: string, plan: StackSpec) => {
...
+      return arg;
+    });
+  }
+  return { services: { [container]: clone } };
+};
+
+/** Builds a host-sc compose override for an extra host chain. */
+const buildExtraHostScOverride = async (chain: HostChainScenario): Promise<ComposeDoc> => {
+  const doc = rewriteComposePaths(await loadComposeDoc("host-sc"));
+  const suffixId = hostChainSuffixId(chain.key); // "host-b" → "b"
```
Eikix: The refactor is much better now that `hostChains` is real data, but runtime naming is still partly driven by host-specific string surgery. I’d consider making `hostChains[].key` the explicit runtime slug and deriving names uniformly from that, e.g. `host-node-${key}`, `host-sc-${key}`, `coprocessor-host-${key}`, `addresses/host-${key}`. That would remove the remaining implicit `host` / `host-b` contract from the lower-level implementation.

test-suite/fhevm/src/generate/index.ts
```diff
@@ -110,6 +112,9 @@ export const generateRuntime = async (state: State, plan: StackSpec) => {
   );
   await writeWritableFile(hostAddressesPath, renderHostAddressesEnv(state));
   await writeWritableFile(hostAddressesSolidityPath, renderHostAddressesSolidity(state));
+  for (const chain of plan.hostChains.slice(1)) {
+    await writeWritableFile(hostChainAddressesPath(chain.key), renderHostChainAddresses(state, chain.key));
```
Eikix, line 112: Extra host chains still only get `.env.host`, but not `FHEVMHostAddresses.sol`. Those extra `host-sc-*` services mount their own chain-specific address dir into `/app/addresses`, and host contract tasks compile contracts that import `addresses/FHEVMHostAddresses.sol`. That means the primary host chain gets the full contract artifact set, while extra host chains only get the dotenv file. I’d make host-chain address generation symmetric and write the Solidity file per host chain too, not just for the primary one.

test-suite/fhevm/src/scenario/resolve.ts
```diff
@@ -88,13 +90,57 @@ const scenarioCandidatePaths = (value: string) => {
...
+    if (!Array.isArray(parsed.hostChains)) {
+      throw new Error(`${sourceLabel}: hostChains must be an array`);
+    }
+    const seen = new Set<string>();
+    return parsed.hostChains.map((entry, index) => {
+      if (!entry || typeof entry !== "object") {
+        throw new Error(`${sourceLabel}: hostChains[${index}] must be an object`);
+      }
+      const chain = entry as Record<string, unknown>;
+      const key = normalizeScalar(chain.key, `${sourceLabel}: hostChains[${index}].key`);
```
Eikix, line 108: `hostChains[].key` still parses as if it were arbitrary identity, but the runtime treats it like a naming token. At this point the implementation is deriving service/env/address names from `key`, so I’d make that explicit and validate it as a safe runtime slug instead of accepting any string shape here. That keeps the schema honest and avoids keys that parse successfully but generate broken runtime names later.

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -203,6 +219,12 @@ jobs:
           # Heavy is the long-running operators lane.
           ./fhevm-cli test heavy
 
+      - name: Multi-chain isolation tests
```
Eikix: `multi-chain-isolation` is still wired as a separate workflow-only step instead of being part of the `standard` profile itself. If the intended outcome is that multi-chain belongs to the standard CI lane, I’d push that into `test standard` rather than keeping it as an extra workflow add-on behind `inputs.multi-chain`.

test-suite/fhevm/src/scenario/resolve.ts
```diff
@@ -88,13 +90,57 @@ const scenarioCandidatePaths = (value: string) => {
       ];
 };
 
+const DEFAULT_HOST_CHAIN: HostChainScenario = { key: "host", chainId: DEFAULT_CHAIN_ID, rpcPort: 8545 };
```
Eikix: /  still encode the old built-in chain layout in parser code.

If the parser is now centered on , these legacy constants are another sign that the old model is still partially preserved. I’d keep only the default single-chain case and remove the legacy Chain B fallback entirely.

test-suite/e2e/test/multiChain/multiChainHelper.ts
```diff
@@ -0,0 +1,127 @@
...
+const inputVerificationAddress = process.env.INPUT_VERIFICATION_ADDRESS!;
+const relayerUrl = process.env.RELAYER_URL!;
+const gatewayChainId = Number(process.env.CHAIN_ID_GATEWAY!);
+
+export interface ChainConfig {
+  rpcUrl: string;
+  chainId: number;
+}
+
+export const CHAIN_A: ChainConfig = {
```
Eikix: The test helper still hardcodes  /  and the corresponding env names.

That keeps the e2e layer tied to the old A/B model even though the CLI moved to . I’d have the helper consume a normalized host-chain config from env instead of baking in exactly two named chains here.

test-suite/e2e/test/multiChain/multiChainHelper.ts
```diff
@@ -0,0 +1,127 @@
...
+  rpcUrl: process.env.RPC_URL || 'http://localhost:8545',
```
Eikix: The fallback RPC URLs and chain IDs here are very forgiving for infra-oriented tests.

If these env vars are part of the expected multi-chain runtime contract, missing values should probably fail fast rather than quietly defaulting to localhost/12345/67890. Otherwise a miswired setup can still appear valid.

test-suite/fhevm/src/generate/compose.ts
```diff
@@ -341,4 +435,30 @@ export const generateComposeOverrides = async (_state: State, plan: StackSpec) =
     const doc = await buildComposeOverride(component, plan);
     await fs.writeFile(target, YAML.stringify(doc));
   }
+
+  // Extra host chain compose overrides
+  const extraChains = plan.hostChains.slice(1);
+  const primaryChainId = plan.hostChains[0]?.chainId ?? "12345";
+  const extraChainFileNames: string[] = [];
+  for (const chain of extraChains) {
+    const suffixId = hostChainSuffixId(chain.key); // "host-b" → "b"
```
Eikix: This still hardcodes the runtime naming scheme around  plus  /  / .

If  is meant to be the runtime slug, I’d derive names directly from that in one consistent helper instead of reconstructing them ad hoc in multiple places.

test-suite/fhevm/src/generate/env.ts
```diff
@@ -243,6 +255,72 @@ export const renderEnvMaps = async (
...
+        coproChain.CHAIN_ID = chain.chainId;
+        if (hostAddresses.ACL_CONTRACT_ADDRESS) {
+          coproChain.ACL_CONTRACT_ADDRESS = hostAddresses.ACL_CONTRACT_ADDRESS;
+          coproChain.FHEVM_EXECUTOR_CONTRACT_ADDRESS = hostAddresses.FHEVM_EXECUTOR_CONTRACT_ADDRESS;
+          coproChain.INPUT_VERIFIER_ADDRESS = hostAddresses.INPUT_VERIFIER_CONTRACT_ADDRESS;
+        }
+        instanceEnvs[`coprocessor-${suffixId}.${index}`] = coproChain;
+      }
+
+      envs["test-suite"][`RPC_URL_CHAIN_${suffixId.toUpperCase()}`] = hostHttp;
```
Eikix: The generated test-suite env still hardcodes the old  naming pattern via  / .

If the topology model is now host-chain-driven, this env surface should also come from a normalized chain config rather than preserving the old A/B-style variable family.

.github/workflows/test-suite-e2e-tests.yml
Eikix: The workflow still treats multi-chain as a separate boolean mode with a separate test step, which keeps some of the old special-case shape alive at the CI layer too.

If multi-chain is now a first-class topology, the cleaner end state is for the CLI profiles and scenario selection to own that behavior, rather than the workflow carrying a separate multi-chain branch.

test-suite/fhevm/src/scenario/resolve.ts
```diff
@@ -88,13 +90,57 @@ const scenarioCandidatePaths = (value: string) => {
...
+      }
+      return {
+        key,
+        chainId,
+        rpcPort,
+        name: normalizeOptionalText(chain.name, `${sourceLabel}: hostChains[${index}].name`),
+      };
+    });
+  }
+  if (parsed.multiChain === true) {
```
Eikix: `parseHostChains()` still carries `multiChain: true` fallback even though the new model is `hostChains`.

If the intent is a hard cutover to explicit topology, this keeps the old model alive in the new parser and adds unnecessary branching. I’d drop the legacy `multiChain` path and require scenarios to declare `hostChains` directly.

test-suite/fhevm/src/scenario/resolve.ts
Eikix: `DEFAULT_HOST_CHAIN` / `LEGACY_CHAIN_B` still encode the old built-in chain layout in parser code.

If the parser is now centered on `hostChains`, these legacy constants are another sign that the old model is still partially preserved. I’d keep only the default single-chain case and remove the legacy Chain B fallback entirely.

test-suite/e2e/test/multiChain/multiChainHelper.ts
Eikix: The test helper still hardcodes `CHAIN_A` / `CHAIN_B` and the corresponding env names.

That keeps the e2e layer tied to the old A/B model even though the CLI moved to `hostChains`. I’d have the helper consume a normalized host-chain config from env instead of baking in exactly two named chains here.

test-suite/e2e/test/multiChain/multiChainHelper.ts
Eikix: The fallback RPC URLs and chain IDs here are very forgiving for infra-oriented tests.

If these env vars are part of the expected multi-chain runtime contract, missing values should probably fail fast rather than quietly defaulting to localhost/12345/67890. Otherwise a miswired setup can still appear valid.

test-suite/fhevm/src/generate/compose.ts
Eikix: This still hardcodes the runtime naming scheme around `hostChainSuffixId(chain.key)` plus `host-node-*` / `host-sc-*` / `coprocessor-host-*`.

If `key` is meant to be the runtime slug, I’d derive names directly from that in one consistent helper instead of reconstructing them ad hoc in multiple places.

test-suite/fhevm/src/layout.ts
```diff
@@ -242,7 +244,12 @@ export const paymentBridgingAddressesSolidityPath = path.join(
   "gateway",
   "PaymentBridgingAddresses.sol",
 );
-export const hostAddressesPath = path.join(ADDRESS_DIR, "host", ".env.host");
+export const hostChainAddressesPath = (key: string) =>
+  path.join(ADDRESS_DIR, key, ".env.host");
+export const hostAddressesPath = hostChainAddressesPath("host");
+
+/** Extracts the suffix ID from a host chain key. "host" → "", "host-b" → "b", "host-foo" → "foo". */
```
Eikix: `hostChainSuffixId()` still assumes keys look like `host` / `host-*`.

That means the schema looks more flexible than the runtime actually is. Either validate keys to that exact shape, or make `key` the explicit runtime slug and stop depending on `host-*` string conventions here.

test-suite/fhevm/src/generate/env.ts
Eikix: The generated test-suite env still hardcodes the old `CHAIN_B` naming pattern via `RPC_URL_CHAIN_${suffix}` / `CHAIN_ID_HOST_${suffix}`.

If the topology model is now host-chain-driven, this env surface should also come from a normalized chain config rather than preserving the old A/B-style variable family.

.github/workflows/test-suite-e2e-tests.yml
Eikix: The workflow still treats multi-chain as a separate boolean mode with a separate test step, which keeps some of the old special-case shape alive at the CI layer too.

If multi-chain is now a first-class topology, the cleaner end state is for the CLI profiles and scenario selection to own that behavior, rather than the workflow carrying a separate multi-chain branch.

test-suite/fhevm/src/generate/compose.ts
PanGan21: fixed

test-suite/fhevm/src/generate/index.ts
PanGan21: fixed

.github/workflows/test-suite-e2e-tests.yml
PanGan21: fixed

test-suite/fhevm/src/scenario/resolve.ts
PanGan21: fixed

test-suite/e2e/test/multiChain/multiChainHelper.ts
PanGan21: fixed

test-suite/e2e/test/multiChain/multiChainHelper.ts
PanGan21: fixed

test-suite/fhevm/src/generate/compose.ts
PanGan21: fixed

test-suite/fhevm/src/generate/env.ts
PanGan21: fixed

.github/workflows/test-suite-e2e-tests.yml
PanGan21: fixed seems like duplicate comment so resolving

test-suite/fhevm/src/scenario/resolve.ts
PanGan21: fixed

test-suite/e2e/test/multiChain/multiChainHelper.ts
PanGan21: fixed seems like duplicate comment so resolving

test-suite/fhevm/src/generate/compose.ts
PanGan21: fixed seems like duplicate comment so resolving

test-suite/fhevm/src/generate/env.ts
PanGan21: fixed seems like duplicate comment so resolving

.github/workflows/test-suite-e2e-tests.yml
PanGan21: fixed and seems like duplicate

test-suite/fhevm/src/layout.ts
PanGan21: fixed

test-suite/e2e/test/multiChain/multiChainHelper.ts
PanGan21: fixed, i think it is duplicate comment

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -197,6 +202,12 @@ jobs:
         run: |
           ./fhevm-cli test standard
 
+      - name: Heavy e2e suite
```
Eikix: Don't put the heavy tests in the e2e tests! they take over 16 hours

PanGan21: Ooops sorry will remove indeed

test-suite/fhevm/src/scenario/resolve.ts
```diff
@@ -88,13 +90,58 @@ const scenarioCandidatePaths = (value: string) => {
...
+      if (!/^host(-[a-z0-9]+)?$/.test(key)) {
+        throw new Error(
+          `${sourceLabel}: hostChains[${index}].key "${key}" must match "host" or "host-{id}" where id is lowercase alphanumeric`,
+        );
+      }
+      if (seen.has(key)) {
+        throw new Error(`${sourceLabel}: duplicate hostChains key "${key}"`);
+      }
+      seen.add(key);
+      const chainId = normalizeScalar(chain.chainId, `${sourceLabel}: hostChains[${index}].chainId`);
```
Eikix, line 118: `hostChains[].chainId` is still only normalized to a string here, but later it is treated as an integer in SQL/config generation.

This should fail fast in parsing. Right now a non-numeric `chainId` can parse successfully and only break later in much messier ways.

test-suite/fhevm/src/flow/up-flow.ts
```diff
@@ -434,8 +439,9 @@ export const validateDiscovery = (state: Pick<State, "target" | "versions" | "di
       throw new PreflightError(`Missing gateway discovery value ${key}`);
     }
   }
+  const primaryHost = discovery.hosts["host"] ?? {};
```
Eikix: Extra host-chain discovery is still allowed to silently fall back to primary-chain addresses.

That means a broken extra host deployment can still boot with Chain A ACL / executor data wired into Chain B. Missing per-chain discovery should fail fast instead of degrading to primary-chain values.

.github/workflows/test-suite-e2e-tests.yml
PanGan21: removed

test-suite/fhevm/src/scenario/resolve.ts
PanGan21: fixed

test-suite/fhevm/src/flow/up-flow.ts
PanGan21: fixed

PR #2158, opened 2026-03-23
test(gateway-contracts): refactor variable declaration into scoped blocks

isaacdecoded:
Closes https://github.com/zama-ai/fhevm-internal/issues/322

### DIFF
--- a/gateway-contracts/test/GatewayConfig.ts
+++ b/gateway-contracts/test/GatewayConfig.ts
@@ -58,26 +58,6 @@ describe("GatewayConfig", function () {
   const fakeTxSender = createRandomWallet();
   const fakeSigner = createRandomWallet();
 
-  let gatewayConfig: GatewayConfig;
-  let owner: Wallet;
-  let pauser: Wallet;
-  let nKmsNodes: number;
-  let kmsNodes: KmsNodeStruct[];
-  let kmsTxSenders: HardhatEthersSigner[];
-  let kmsSigners: HardhatEthersSigner[];
-  let coprocessors: CoprocessorStruct[];
-  let nCoprocessors: number;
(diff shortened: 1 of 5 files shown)


gateway-contracts/test/GatewayConfig.ts
zmalatrax: Should it follow the same pattern refactorization? (avoid shared global variables)

The `getInputsForDeployFixture` mutates global variables rather than returning them locally

> "Deployment" (lines 153–570, ~30 tests) has zero local initialization. It relies entirely on the parent before() hook for owner, gatewayConfig, kmsNodes, coprocessors, custodians, and the high*Threshold values.  
> 
> - Move the let declarations into each sub-suite that uses them
> - Make getInputsForDeployFixture() return kmsNodes/coprocessors/custodians as return values instead of mutating parent scope
> - Give "Deployment" its own beforeEach with a local fixture load

isaacdecoded: Yes, you're right. In the beginning, I hesitated because for these GatewayConfig tests, there are multiple "big" describe blocks, so I thought it was good to reuse the same global vars. But now you pointed out, the main goal for this PR is to precisely avoid that. Changes done [here](https://github.com/zama-ai/fhevm/pull/2158/changes/5a675731aaffd8b9a365f7ec7d4c9d8953713d04)

zmalatrax: Yeah, it's a bit tricky as it inflates the various describe blocks

We could think of a broader refactor with another fixture and a destructuring pattern within the tests to avoid the beforeEach ; but it requires modifying all tests and might end up being harder to reason about 🤔

zmalatrax (approved):
lgtm

isaacdecoded:
@mergifyio queue

PR #2164, opened 2026-03-24
feat(coprocessor): parallelize CI tests per service and reduce operator tests for prs

PanGan21:
## Summary

- Parallelize coprocessor CI tests across 7 service-specific matrix jobs using per-package
- Add dedicated `coverage-report` job that merges per-service LCOV artifacts with `lcov`
- Add `cargo-tests-status` aggregator job to preserve the `coprocessor-cargo-test/cargo-tests (bpr)` branch protection check name
- Runs the operator tests up to FHEUint64 on pr and everything on merge-queue
- Runs the operator gpu tests only for FHEUint64 on pr and everything on merge-queue
- Make test harness DB setup generic: extract DB name from URL instead of hardcoding `"coprocessor"`, enabling isolated per-service test databases
- Replace manual table truncation in `transaction-sender` tests with `setup_test_db()` for consistency
- Removes unused postgres setup in docker from CI

Closes: https://github.com/zama-ai/fhevm-internal/issues/1186

### DIFF
--- a/coprocessor/fhevm-engine/test-harness/src/instance.rs
+++ b/coprocessor/fhevm-engine/test-harness/src/instance.rs
@@ -3,7 +3,8 @@ use std::sync::Arc;
 use crate::db_utils::setup_test_key;
 use fhevm_engine_common::utils::DatabaseURL;
 use sqlx::postgres::types::Oid;
-use sqlx::Row;
+use sqlx::postgres::PgConnectOptions;
+use sqlx::{ConnectOptions, Row};
 use testcontainers::{core::WaitFor, runners::AsyncRunner, GenericImage, ImageExt};
 use tokio_util::sync::CancellationToken;
 use tracing::info;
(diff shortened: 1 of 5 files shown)


coprocessor/fhevm-engine/transaction-sender/tests/common.rs
```diff
@@ -80,25 +80,14 @@ impl TestEnvironment {
...
+        let db_instance = setup_test_db(ImportMode::None)
+            .await
+            .map_err(|e| anyhow::anyhow!("{e}"))?;
         let db_pool = PgPoolOptions::new()
             .max_connections(10)
-            .connect(DatabaseURL::default().as_str())
+            .connect(db_instance.db_url())
             .await?;
 
-        Self::truncate_tables(
```
Eikix, line 88: why do we stop truncating table?

PanGan21: I think truncating was there because the db instance was shared between services.
Now a new/clean db instance is used for transaction sender with a separate db url therefore there is no need to truncate.
Additionally if a new table would be added with a migration then this table should be added in truncate as well which imo is not optimal to do twice.
Do you think I should revert this part?

antoniupop:
You may need to rename if not using nextest

antoniupop (dismissed):

coprocessor/fhevm-engine/test-harness/src/instance.rs
```diff
@@ -55,6 +55,33 @@ pub async fn setup_test_db(mode: ImportMode) -> Result<DBInstance, Box<dyn std::
...
+        !name.is_empty()
+            && name
+                .chars()
+                .all(|c| c.is_alphanumeric() || c == '_' || c == '-'),
+        "invalid database name extracted from URL: {name}"
+    );
+    name
+}
+
+fn admin_url_from(db_url: &str) -> String {
```
Eikix, line 70: `DATABASE_URL` is being parsed here with raw string splitting. That is a regression from the previous implementation because valid Postgres URLs may contain `/` in query values, for example `...?sslrootcert=/tmp/root.crt`. In that case `admin_url_from()` will split on the wrong slash and produce an invalid admin URL in the localhost reset path. It would be safer to parse the URL structurally and change only the database name while preserving the rest of the connection options.

PanGan21: Done

Eikix (dismissed):
LGTM

coprocessor/fhevm-engine/transaction-sender/tests/common.rs
PanGan21: I restored it, it was used to run the tests with a specific flag

Eikix (approved):
lgtm

antoniupop (approved):

PanGan21:
@mergify queue

PR #2166, opened 2026-03-24
feat(kms-connector): event listening via eth_getLogs

eudelins-zama:
## Description of changes

Replaces the per-event-type subscription model (`eth_newFilter` + `eth_getFilterChanges`) with a simpler, batched `eth_getLogs` polling approach. This improves reliability (no more stale server-side filters).

Closes https://github.com/zama-ai/fhevm-internal/issues/1189

### Core architecture change

- **Before:** One async task per `EventType` (7 total), each creating a server-side filter via `eth_newFilter` and polling events via `eth_getFilterChanges`. A separate catchup phase was using `eth_getFilterLogs` to recover missed events on startup.
- **After:** Two polling tasks (`Decryption` and `KmsGeneration`), each running a `run_poll_loop` that calls `eth_getLogs` with explicit `fromBlock`/`toBlock` ranges. The catchup phase is "merged" with the core listening phase.

### Key changes

- **Batched DB insertion:** Events from a single `eth_getLogs` response are published in one DB transaction (`publish_batch`), replacing individual `publish_event` calls.
  - The batch size is configurable.
  - When more blocks remain, `ticker.reset_immediately()` triggers the next poll without waiting.
- **Error handling:** New `max_consecutive_polling_errors` config (default 20) stops the listener after repeated failures instead of failing on the first error.
  - Errors reset on any successful poll.
  - Replaces the previous mechanism which was crashing the listener when the filter was dropped by the RPC node.
  - This change is open to debate ofc.
- **Metric rename:** `event_received_errors` → `event_listening_errors`, now labeled by `contract` instead of `event_type`.
- **Default decryption polling:** Changed from 1s to 500ms.


eudelins-zama:
@claude /pr-review

eudelins-zama:
@claude /pr-review

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -273,36 +263,36 @@ async fn publish_key_reshare_same_set(
...
+        WHERE event_type = ANY($1::event_type[]) AND (block_number IS NULL OR block_number < $2)",
+        event_types as &[EventType],
         last_block_polled.map(|n| n as i64),
         Utc::now(),
     )
     .execute(db_pool)
     .await?;
 
-    if query_result.rows_affected() == 1 {
+    if query_result.rows_affected() == event_types.len() as u64 {
```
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,174 @@ where
...
 
-        info!(
-            "Successfully caught {event_count} {event_type} events from block {catchup_from_block}!"
-        );
-        if let Err(e) = provider.uninstall_filter(event_filter_id).await {
-            warn!("Failed to uninstall {event_type} event catchup filter: {e}");
-        }
-        Ok(())
+    /// Decodes a log and dispatches it to the appropriate event handler.
+    async fn process_log(&self, group: MonitoredContract, log: Log) {
```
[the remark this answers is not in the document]
eudelins-zama: Fixed 28d6226944f9655b3d3899ea394d9a6ce1091e71

eudelins-zama:
@claude /pr-review

eudelins-zama:
@claude /pr-review

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,185 @@ where
...
+                Ok(event) => match event.data {
+                    DecryptionEvents::PublicDecryptionRequest(e) => e.into(),
+                    DecryptionEvents::UserDecryptionRequest(e) => e.into(),
+                    _ => return trace!("Ignoring Decryption event: {log:?}"),
+                },
+                Err(e) => {
+                    EVENT_RECEIVED_ERRORS
+                        .with_label_values(&[&contract_label])
+                        .inc();
+                    return warn!("Failed to decode Decryption event: {e}");
```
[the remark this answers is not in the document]
eudelins-zama: Known tradeoff to limit changes in this PR

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,185 @@ where
...
+                    KMSGenerationEvents::CrsgenRequest(e) => e.into(),
+                    KMSGenerationEvents::PRSSInit(e) => e.into(),
+                    KMSGenerationEvents::KeyReshareSameSet(e) => e.into(),
+                    _ => return trace!("Ignoring KMSGeneration event: {log:?}"),
+                },
+                Err(e) => {
+                    EVENT_RECEIVED_ERRORS
+                        .with_label_values(&[&contract_label])
+                        .inc();
+                    return warn!("Failed to decode KMSGeneration event: {e}");
```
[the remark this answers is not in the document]
eudelins-zama: Same

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,185 @@ where
...
+    /// Returns the last **completed** block, so that the polling loop starts from `last + 1`.
+    async fn get_start_block(
+        &self,
+        from_block_config: Option<u64>,
+        event_types: &[EventType],
+    ) -> anyhow::Result<Option<u64>> {
+        if let Some(from_block) = from_block_config {
+            info!("Found configured from_block_number ({from_block}) for polling");
+            // Subtract 1 because the polling loop will do `last + 1` to get the first block.
+            return Ok(Some(from_block.saturating_sub(1)));
```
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/gateway.rs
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/gateway.rs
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,185 @@ where
...
+            for log in logs {
+                self.process_log(contract, log).await;
+            }
 
-        for event in past_events {
-            self.spawn_event_handling(event_type, event, last_block_polled)
-                .await;
-        }
+            last_processed_block = Some(to_block);
+            self.update_block_tracking(event_types, to_block).await?;
```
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -273,36 +263,39 @@ async fn publish_key_reshare_same_set(
...
-            "Last block polled for {event_type} was successfully updated!"
+            "Last block polled updated for {}/{} event types in {event_types:?}",
+            rows_affected,
+            event_types.len()
         );
     } else {
         debug!(
             last_block_polled,
-            "Last block polled for {event_type} was not updated: {query_result:?}"
+            "Last block polled for {event_types:?} were not updated: {query_result:?}"
```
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +106,183 @@ where
...
+                Ok((new_from_block, has_more)) => {
+                    consecutive_errors = 0;
+                    from_block = new_from_block;
+                    if has_more {
+                        ticker.reset_immediately();
+                    }
+                }
+                Err(e) => {
+                    EVENT_LISTENING_ERRORS
+                        .with_label_values(&[contract.to_string().to_lowercase()])
```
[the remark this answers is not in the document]
eudelins-zama: Hallucination

kms-connector/crates/gw-listener/tests/common/mod.rs
```diff
@@ -191,6 +191,26 @@ pub async fn fetch_from_db(db: &Pool<Postgres>, event_type: EventType) -> sqlx::
...
+    let start = std::time::Instant::now();
+    loop {
+        let rows = fetch_from_db(db, event_type).await?;
+        if check_event_in_db(&rows, expected_event.clone()).is_ok() {
+            return Ok(());
+        }
+        if start.elapsed() > timeout {
+            anyhow::bail!("Timed out waiting for {event_type} event in DB");
+        }
+        tokio::time::sleep(poll_interval).await;
```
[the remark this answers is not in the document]
eudelins-zama: Nitpicking for a test fn, ignoring

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -346,88 +305,54 @@ where
...
             data: None,
         });
 
-        let event_type = EventType::KeygenRequest;
-        tokio::spawn(gw_listener.subscribe(event_type));
-        test_instance.wait_for_log("Failed to catch up").await;
-        test_instance
-            .wait_for_log(&format!("Waiting for next {event_type}"))
-            .await;
+        gw_listener.poll_events(MonitoredContract::Decryption).await;
```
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -346,88 +305,54 @@ where
...
+
+        // Both tasks will fail on get_block_number or get_logs
+        for _ in 0..2 {
+            asserter.push_success(&100_u64);
+            asserter.push_failure(ErrorPayload {
+                code: -32000,
+                message: "rpc error".into(),
+                data: None,
+            });
+        }
```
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/config/gw-listener.toml
```diff
@@ -50,6 +50,14 @@ database_url = "postgres://postgres:postgres@localhost/kms-connector"
 # ENV: KMS_CONNECTOR_KMS_GENERATION_POLLING (format: https://docs.rs/humantime/latest/humantime/)
 # key_management_polling = "30s"
 
+# Maximum number of blocks per eth_getLogs RPC call (optional, defaults to 100)
+# ENV: KMS_CONNECTOR_GET_LOGS_BATCH_SIZE
+# get_logs_batch_size = 100
+
+# Maximum consecutive polling errors before the listener stops (optional, defaults to 10)
+# ENV: KMS_CONNECTOR_MAX_CONSECUTIVE_POLLING_ERRORS
+# max_consecutive_polling_errors = 10
```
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +106,183 @@ where
...
-        };
+        let to_block = std::cmp::min(
+            from_block.saturating_add(self.config.get_logs_batch_size.saturating_sub(1)),
+            current_block,
+        );
 
+        let signatures = event_types
+            .iter()
+            .map(|e| e.signature_hash())
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/config.rs
```diff
@@ -62,6 +62,14 @@ pub struct Config {
     #[serde(with = "humantime_serde", default = "default_key_management_polling")]
     pub key_management_polling: Duration,
 
+    /// The maximum number of blocks to fetch per `eth_getLogs` request.
+    #[serde(default = "default_get_logs_batch_size")]
+    pub get_logs_batch_size: u64,
+
+    /// Maximum number of consecutive polling errors before stopping the loop.
+    #[serde(default = "default_max_consecutive_polling_errors")]
+    pub max_consecutive_polling_errors: u8,
```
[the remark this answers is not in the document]
eudelins-zama: Fixed

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -89,12 +83,11 @@ async fn publish_event_inner(
         warn!("Unexpected query result while publishing event: {query_result:?}");
     }
```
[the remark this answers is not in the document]
eudelins-zama: We're using a single listener replica for now, but indeed we could change this in the future

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,183 @@ where
...
+                    consecutive_errors = 0;
+                    from_block = new_from_block;
+                    if has_more {
+                        ticker.reset_immediately();
+                    }
+                }
+                Err(e) => {
+                    EVENT_LISTENING_ERRORS
+                        .with_label_values(&[contract.to_string().to_lowercase()])
+                        .inc();
```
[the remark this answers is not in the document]
eudelins-zama: Hallucination

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,183 @@ where
...
-            Some(block) => *block,
-        };
+        contract: MonitoredContract,
+        base_filter: Filter,
+        event_types: &[EventType],
+        from_block: u64,
+    ) -> anyhow::Result<(u64, bool)> {
+        let current_block = self.provider.get_block_number().await?;
+
+        if from_block >= current_block {
```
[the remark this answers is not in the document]
eudelins-zama: Fixed 92ab2b92f7376c7d6fc4a24d687633de16d294f0

kms-connector/crates/gw-listener/tests/common/mod.rs
```diff
@@ -191,6 +191,26 @@ pub async fn fetch_from_db(db: &Pool<Postgres>, event_type: EventType) -> sqlx::
     sqlx::query(query).fetch_all(db).await
 }
 
+pub async fn poll_db_for_event(
+    db: &Pool<Postgres>,
+    event_type: EventType,
+    expected_event: &GatewayEventKind,
+) -> anyhow::Result<()> {
+    let timeout = Duration::from_secs(30);
+    let poll_interval = Duration::from_millis(200);
```
maksymsur, line 200: Do we need to make here and in other places of the code timings configurable via env vars? Or we are pretty sure that these are good and no experimentation is needed to find the golden needle?

maksymsur (dismissed):
LGTM! Only one small clarification

kms-connector/crates/gw-listener/tests/common/mod.rs
eudelins-zama: These are values for the integration tests, so no need to make them configurable (for now at least).

We could do later if needed though! :slightly_smiling_face:

kms-connector/crates/gw-listener/src/core/config.rs
dvdplm: Is `0` allowed here? What would the effect be?

Total nitpick, but I'd probably use `usize` here (and above as well).

kms-connector/crates/gw-listener/src/core/config.rs
```diff
@@ -75,13 +83,21 @@ fn default_service_name() -> String {
...
 }
 
+fn default_get_logs_batch_size() -> u64 {
+    100
+}
+
+fn default_max_consecutive_polling_errors() -> u8 {
+    20
+}
+
```
dvdplm, line 100: Do you really need these to make `serde` do the right thing? What happens if you just put in these values in the `Default` impl?

kms-connector/crates/gw-listener/src/core/config.rs
```diff
@@ -125,6 +143,8 @@ mod tests {
             env::remove_var("KMS_CONNECTOR_DECRYPTION_CONTRACT__ADDRESS");
             env::remove_var("KMS_CONNECTOR_KMS_GENERATION_CONTRACT__ADDRESS");
             env::remove_var("KMS_CONNECTOR_SERVICE_NAME");
+            env::remove_var("KMS_CONNECTOR_GET_LOGS_BATCH_SIZE");
+            env::remove_var("KMS_CONNECTOR_MAX_CONSECUTIVE_POLLING_ERRORS");
```
dvdplm, line 147: Wow, this is an `unsafe` function! I had no idea...

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -1,33 +1,56 @@
...
+    EventType::PrssInit,
+    EventType::KeyReshareSameSet,
+];
+
+/// Identifies which contract is being polled.
+#[derive(Clone, Copy)]
+enum MonitoredContract {
+    Decryption,
+    KmsGeneration,
+}
```
dvdplm, line 46: Monitoring a new contract means we have to make a new release. That's probably fine, but something worth pointing out in the docs I think? Alternatively consider making this a configurable item I guess, but that seems out of scope for this PR (add issue if worth doing?).

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -61,33 +81,21 @@ where
...
-
-        tasks.spawn(self.clone().subscribe(EventType::PublicDecryptionRequest));
-        tasks.spawn(self.clone().subscribe(EventType::UserDecryptionRequest));
-        tasks.spawn(self.clone().subscribe(EventType::PrepKeygenRequest));
-        tasks.spawn(self.clone().subscribe(EventType::KeygenRequest));
-        tasks.spawn(self.clone().subscribe(EventType::CrsgenRequest));
-        tasks.spawn(self.clone().subscribe(EventType::PrssInit));
-        tasks.spawn(self.subscribe(EventType::KeyReshareSameSet));
+        tasks.spawn(self.clone().poll_events(MonitoredContract::Decryption));
+        tasks.spawn(self.poll_events(MonitoredContract::KmsGeneration));
```
dvdplm, line 101: The `strum` crate is in the `Cargo.lock` so you could do something like `for contract in MonitoredContract::iter()` to avoid hard coding each variant.

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,183 @@ where
...
-            }
-            EventType::PrssInit => {
-                let filter = self.kms_generation_contract.PRSSInit_filter();
-                self.subscribe_inner(event_type, filter, polling).await
-            }
-            EventType::KeyReshareSameSet => {
-                let filter = self.kms_generation_contract.KeyReshareSameSet_filter();
-                self.subscribe_inner(event_type, filter, polling).await
-            }
+    /// Shared `eth_getLogs` polling loop.
```
dvdplm: In what sense is it "shared"?

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,183 @@ where
...
+        info!("Started {contract} polling from block {from_block}");
 
-        select! {
-            _ = self.process_events(event_type, event_poller, &mut last_block_polled) => (),
-            _ = self.cancel_token.cancelled() => info!("{event_type} subscription cancelled..."),
+        let mut ticker = tokio::time::interval(poll_interval);
+        let max_errors = self.config.max_consecutive_polling_errors;
+        let mut consecutive_errors: u8 = 0;
+        loop {
+            ticker.tick().await;
```
dvdplm, line 157: I'm not good with tokio stuff, so this is more of a curiosity: what is advantage/disadvantage of doing it with a loop + ticker like this vs using `tokio::select!`?

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,183 @@ where
...
-        let contract_address = match event_type {
-            EventType::PublicDecryptionRequest | EventType::UserDecryptionRequest => {
-                self.decryption_contract.address()
-            }
-            _ => self.kms_generation_contract.address(),
-        };
+        let to_block = std::cmp::min(
+            from_block.saturating_add(self.config.get_logs_batch_size.saturating_sub(1)),
+            current_block,
+        );
```
dvdplm, line 202: What happens if we're miscvonfigured and `self.config.get_logs_batch_size` is, say, 1? Wouldn't `to_block` and `from_block` end up being equal? would that be bad?

dvdplm: I guess you check that `from_block` is strictly greater than `current_block` above, so it's fine?

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -97,233 +105,183 @@ where
...
-            });
+        let filter = base_filter.from_block(from_block).to_block(to_block);
 
-        for event in past_events {
-            self.spawn_event_handling(event_type, event, last_block_polled)
-                .await;
-        }
+        let logs = self.provider.get_logs(&filter).await?;
+        let events = Self::prepare_events(contract, logs)?;
+        publish_batch(&self.db_pool, events, event_types, to_block).await?;
```
dvdplm, line 208: Do we have to `await` here? Or could we let the publication happen asynchronously?

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -14,71 +14,65 @@ use fhevm_gateway_bindings::{
...
-            Ok(()) => return Ok(()),
-            Err(e) => error!("Insertion attempt #{i}/{INSERTION_RETRY_LIMIT} failed: {e}"),
-        }
-        if i != INSERTION_RETRY_LIMIT {
-            tokio::time::sleep(INSERTION_RETRY_INTERVAL).await;
-        }
+    let mut tx = db_pool.begin().await?;
+    for event in events {
+        publish_event_inner(&mut *tx, event).await?;
     }
```
dvdplm, line 35: I guess this could be a `tokio::join!` to run all queries in parallel, but not sure that's an actual win.

dvdplm (commented):
LGTM, left  some questions/suggestions but nothing required/major imo.

A curiosity: do you know how filtering is implemented on the server side by any chance? I bet they're typically polling as well, so this change is probably more efficient on that side as well.

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: I don't think we want to make this configurable.
The kms-connector is meant to listen to a very specific set of events (and thus contracts). Listening to new events/contracts **should** (IMHO) imply a new release!

I'll update the doc as you suggested though :slightly_smiling_face:

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: Good suggestion, I was avoiding `strum` to avoid having one more external deps.
But since it's already a sub-dep, I guess it would be fine!

Though for this specific suggestion, I prefer not to apply it as with [RFC-013](https://github.com/zama-ai/tech-spec/pull/408), this will move soon

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: Used to listen to both `Decryption` and `KMSGeneration` contracts

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: Not sure to see what exact pattern you have in mind with the `select!`.

But `select!` picks the first of the given futures that is `Ready`, and process with the associated code block.

Here we want to wait for a specific interval of time before polling again, so that's why we use the ticker and not `select!`.

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: We could probably spend some time to add checks on the config.

But to give you a bit of context here: we tend to not ever use optional fields of the config, and use only whatever is `Default`.
Because syncing with partners on setting specific config values is tedious. (I think it's easier to change the hard-coded default and ask for a version bump than having to sync on setting up specific values)

So the configurable aspect is mainly here "just-in-case", and should barely be used, so Idk if it's worth spending too much time on setting strong guardrails

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: To avoid missing events, I prefer not to go processing the next batch until we're sure the previous one is fully stored in DB.

That way, if we get an error while storing the events, we'll retry this specific batch before processing the next one.
There may be another (cleaner) way to achieve this without redoing the RPC calls and so on though.

Also, I think this is fine right now, but under high load, we might have to switch to async publication indeed

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Probably indeed!

Will take a look

eudelins-zama:
> A curiosity: do you know how filtering is implemented on the server side by any chance?

Tbh, I have not looked into it :upside_down_face: @dvdplm

kms-connector/crates/gw-listener/src/core/config.rs
eudelins-zama: See related comment https://github.com/zama-ai/fhevm/pull/2166#discussion_r2994631504

> Total nitpick, but I'd probably use usize here (and above as well).

Any specific reason for using `usize`?

kms-connector/crates/gw-listener/src/core/config.rs
eudelins-zama: Will double check, not sure if `serde` uses the `Default` of the type (`u8` here) or the struct (`Config`) in such case

eudelins-zama: I tested and it seems to use the `Default` of the type (so `0` for `u8`) and not the `20` that is set in the `Default` impl of the `Config`. Thus, using a fn is the only way to do what we want AFAIU.

Related `serde` issues:
- https://github.com/serde-rs/serde/issues/368
- https://github.com/serde-rs/serde/issues/2254

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Hm, not sure it's possible because of the `&mut *tx`, right?

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: Done in 15bc3ecfcbab603ade6b474aa20384869f6f7f0e

kms-connector/crates/gw-listener/src/core/config.rs
eudelins-zama: `0` is currently handled the same as `1` from AFAICT: https://github.com/zama-ai/fhevm/blob/15bc3ecfcbab603ade6b474aa20384869f6f7f0e/kms-connector/crates/gw-listener/src/core/gateway.rs#L158-L179

kms-connector/crates/gw-listener/src/core/publish.rs
dvdplm: oh, right, well, they'd each need their own channel. Either way: given the call is doing IO I don't think it matters?

kms-connector/crates/gw-listener/src/core/gateway.rs
dvdplm: Fair!

kms-connector/crates/gw-listener/src/core/gateway.rs
dvdplm: I think I'm confused with the pattern where you `select!` on a future + a timer (aka deadline) to implement timeouts. Ignore me.

kms-connector/crates/gw-listener/src/core/gateway.rs
dvdplm: Ah, I see. 

```suggestion
    /// Polling loop to listen to both [`Decryption`] and [`KMSGeneration`] contracts.
```

…maybe?

kms-connector/crates/gw-listener/src/core/config.rs
dvdplm: Ha, had no idea `serde` was limited like that. Ty for checking!

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: Done 1fde9681643aaf513862e21c46edeac620b02504 :slightly_smiling_face:

kms-connector/crates/gw-listener/src/core/config.rs
dvdplm: > Any specific reason for using `usize`?

To me, when I read `usize` in code like this my mind goes "ok, any number is fine here", whereas both `u64` and `u8` kiiiiind of signals that the author meant something special, which in turns makes me, the reader, take a longer look. If the intention really is "just a number, no special meaning" I tend to go with `usize`. But yeah, very nitpicky.

dvdplm (commented):
LGTM

kms-connector/crates/gw-listener/src/core/config.rs
eudelins-zama: Fixed in e2b787c47329459b9189b83e1284fd9aed626ef7 :wink: (only for the newly added field, other fields are out of scope IMO)

dvdplm (approved):
LGTM (this time for reals!)

eudelins-zama:
@mergifyio queue

eudelins-zama:
@mergifyio refresh

isaacdecoded (approved):

eudelins-zama:
@mergifyio refresh

PR #2167, opened 2026-03-24
feat(common): add ABI compatibility checks and reorganize CI scripts

Eikix:
## Summary
- Add a shared ABI compatibility checker plus a CI entrypoint for host and gateway contracts
- Add a local wrapper command to compare two refs with the same core logic used in CI
- Reorganize `ci/` into subsystem folders: `abi-compat/`, `upgrade-check/`, `shared/`
- Fix code smells: broken import in exceptions.ts, missing error handling in upgrade-check list.ts, incorrect `repoRoot` computation

## Structure

```
ci/
├── abi-compat/     check.ts, config.ts, exceptions.ts, lib.ts, list.ts, README.md
├── upgrade-check/  check.ts, config.ts, hints.ts, lib.ts, list.ts
├── shared/         merge-address-constants.ts
└── README.md
```

Both subsystems follow the same pattern:
- `check.ts` — CI entrypoint (single package, GitHub Actions annotations)
- `lib.ts` — core comparison logic
- `list.ts` — local multi-package report (creates worktrees, installs deps, runs comparison)
- `config.ts` — package names, contract lists, build deps

## What ABI compat checks
- Compares deployed contract ABIs against a baseline tag
- Matches by full canonical ABI signature (overloads supported)
- Filters out initializer/reinitializer and gated admin/protocol-only functions by modifier pattern
- Allows additions and fails when a stable signature disappears

## Local usage

```bash
bun ci/abi-compat/list.ts --from v0.11.1 --to v0.12.0-0
bun ci/abi-compat/list.ts --from v0.11.1 --package host-contracts
```

```bash
bun ci/upgrade-check/list.ts --from v0.11.1 --to v0.12.0-0
```

## Code smell fixes
- `ci/abi-compat/exceptions.ts`: import from `./abi-compat-config` (nonexistent) fixed to `./config`
- `ci/upgrade-check/list.ts`: `run()` now captures stderr/stdout on failure instead of swallowing errors; worktrees cleaned up explicitly instead of relying on `git worktree prune`
- `ci/upgrade-check/list.ts`: `repoRoot` used `dirname(import.meta.dir)` which would resolve to wrong directory after move; fixed to `resolve(import.meta.dir, "../..")`
- `ci/upgrade-check/list.ts`: inline `PackageName`/`PACKAGE_CONFIG` extracted to `config.ts`

## Validation
- All 5 entrypoints verified with `bun build --no-bundle`
- No stale references to old paths (grep verified)
- zizmor clean on both workflow files


README.md
```diff
@@ -139,6 +139,16 @@ There are two ways to contribute to FHEVM:
...
+### Contract ABI checks
+
+To compare the stable contract ABI surface between two refs locally, use:
+
+```bash
+bun ci/list-abi-compat.ts --from v0.11.1 --to v0.12.0-0
+```
+
+Pass `--package host-contracts` or `--package gateway-contracts` to limit the scope.
+
```
eudelins-zama: Nit: not sure we want this in the main `README.md`?

Eikix: agreed

eudelins-zama:
Would be nice to put your scripts in a dedicated folder within the `ci` folder

ci/abi-compat-config.ts
```diff
@@ -0,0 +1,40 @@
+export type PackageName = "host-contracts" | "gateway-contracts";
+
+export const EXCLUDED_MODIFIERS = new Set([
+  "onlyOwner",
+  "onlyACLOwner",
+  "onlyGatewayOwner",
+  "onlyFromEmptyProxy",
+  "onlyPauser",
+  "onlyCoprocessorTxSender",
+  "onlyKmsTxSender",
```
Eikix, line 10: Not sure we want to exclude modifiers that target our internal infra, since that may still result in breaking changes, just not user facing.
Wdyt @eudelins-zama

eudelins-zama: As this is internal to contracts logic (i.e. not directly used by services like copro or kms-connector), I cannot answer this :thinking: 
Maybe @isaacdecoded opinion is more relevant

isaacdecoded: I'm trying to find a case where changing these modifiers can result in a breaking change. Maybe adding new ones could, but changing/removing the current ones, I don't think so, since they're running validations over values stored in the contracts' state 🤔

ci/abi-compat/list.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  if (packageFailures > 0) {
+    return packageFailures;
+  }
+
+  return 0;
+}
+
+const { fromRef, toRef, packages } = parseArgs();
+const repoRoot = resolve(import.meta.dir, "..");
```
PanGan21: Is this correct path since it is nested?

ci/abi-compat/lib.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  return null;
+}
+
+function stripComments(source: string) {
+  return source.replace(/\/\*[\s\S]*?\*\//g, "").replace(/\/\/.*$/gm, "");
+}
+
+function excludedFunctionNames(source: string) {
```
PanGan21: Isn't it better to exclude by function signature? Since solidity supports overloads this can result to inconsistency

Eikix: Will save it for a follow-up

Eikix:
Superseded by a refreshed PR opened from a new head branch to avoid stale bot review state and Mergify issues.

PR #2170, opened 2026-03-25
fix(coprocessor): scheduler, catch panic

rudy-6-4:


rudy-6-4:
@claude /pr-review

coprocessor/fhevm-engine/tfhe-worker/src/tests/operators_from_events.rs
```diff
@@ -319,6 +321,93 @@ async fn test_fhe_binary_operands_events() -> Result<(), Box<dyn std::error::Err
...
+
+    wait_until_all_allowed_handles_computed(&app).await?;
+    let errors = errors_on_allowed_handles(&app).await?;
+    for msg in &errors {
+        assert!(
+            msg.contains("ExecutionPanic"),
+            "Expected error message to contain 'ExecutionPanic', but got: {msg}"
+        );
+    }
+    assert_eq!(
```
Eikix, line 403: do you need to assert cases.len() >= 1?

Eikix:
clippy fails but otherwise, nice! lgtm

coprocessor/fhevm-engine/tfhe-worker/src/tests/operators_from_events.rs
rudy-6-4: yes, if all cases are computed without error (there should all be with error) for instance the test would be green instead of red. Because the loop before only loop on errors .

Eikix: but what if the cases are empty somehow? the loop will never happen right and the test will be green

antoniupop (approved):
LGTM

PanGan21 (approved):

coprocessor/fhevm-engine/tfhe-worker/src/tests/operators_from_events.rs
rudy-6-4: yes if the "cases" array is empty, the test does nothing. I'll merge PR and will do a trivial PR fix.

rudy-6-4:
@mergifyio queue

PR #2174, opened 2026-03-25
feat(coprocessor): speed up CI

antoniupop:
To speed up compilation times in CI

### DIFF
--- a/.github/workflows/coprocessor-cargo-tests.yml
+++ b/.github/workflows/coprocessor-cargo-tests.yml
@@ -41,6 +41,9 @@ jobs:
       packages: 'read' # Required to read GitHub packages/container registry
       pull-requests: 'write' # Required to post coverage comment on PR
     runs-on: large_ubuntu_16
+    env:
+      CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER: clang
+      RUSTFLAGS: "-C link-arg=-fuse-ld=mold"
     steps:
     - name: Checkout code
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -65,7 +68,7 @@ jobs:
     - name: Install cargo dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y protobuf-compiler && \
+        sudo apt-get install -y protobuf-compiler mold clang && \
         cargo install sqlx-cli --version 0.7.2 --no-default-features --features postgres --locked
 
     - name: Install foundry
--- a/.github/workflows/coprocessor-cargo-tests.yml
+++ b/.github/workflows/coprocessor-cargo-tests.yml
@@ -117,8 +117,8 @@ jobs:
 
     - name: Compile tests with coverage instrumentation
       run: |
-        # shellcheck disable=SC1090
-        source <(cargo llvm-cov show-env --sh)
+        cargo llvm-cov show-env --sh > /tmp/llvm-cov-env.sh
+        source /tmp/llvm-cov-env.sh
         DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor \
         SQLX_OFFLINE=true \
         cargo test --no-run --workspace --profile coverage
--- a/.github/workflows/coprocessor-cargo-tests.yml
+++ b/.github/workflows/coprocessor-cargo-tests.yml
@@ -147,10 +147,11 @@ jobs:
 
     - name: Run tests with coverage
       run: |
+        source /tmp/llvm-cov-env.sh
         DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor \
         TEST_GLOBAL_LOCALSTACK=1 \
         SQLX_OFFLINE=true \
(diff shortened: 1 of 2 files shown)


antoniupop:
@claude /pr-review

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -95,27 +103,53 @@ jobs:
...
 
     - name: Clean previous coverage data
       run: cargo llvm-cov clean --workspace --profile coverage
       working-directory: coprocessor/fhevm-engine
 
+    - name: Compile tests with coverage instrumentation
+      run: |
+        # shellcheck disable=SC1090
+        source <(cargo llvm-cov show-env --sh)
+        DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor \
```
PanGan21, line 122: Do we still `DATABASE_URL` if `SQLX_OFFLINE=true` in compilation step?

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -81,41 +103,53 @@ jobs:
...
-        docker run --rm -d -p 4566:4566 --name localstack localstack/localstack:4.14.0
+      run: docker run --rm -d -p 4566:4566 --name localstack localstack/localstack:4.14.0
 
     - name: Clean previous coverage data
       run: cargo llvm-cov clean --workspace --profile coverage
       working-directory: coprocessor/fhevm-engine
 
+    - name: Compile tests with coverage instrumentation
+      run: |
+        # shellcheck disable=SC1090
```
PanGan21: Instead of disabling we can do here the following:
```
cargo llvm-cov show-env --sh > /tmp/llvm-cov-env.sh
source /tmp/llvm-cov-env.sh
```

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -81,41 +103,53 @@ jobs:
...
+          cat /tmp/db-init.log
+          exit 1
+        fi
+        echo "Database migration completed"
+
     - name: Run tests with coverage
       run: |
         DATABASE_URL=postgresql://postgres:postgres@localhost:5432/coprocessor \
         TEST_GLOBAL_LOCALSTACK=1 \
+        SQLX_OFFLINE=true \
```
PanGan21, line 153: I think since we run the tests in this step then `SQLX_OFFLINE=true` is not correct, but correct me if I am wrong

antoniupop: Yes - it's not needed for execution, but it's there to ensure we don't re-trigger compilation.

.github/workflows/coprocessor-cargo-tests.yml
antoniupop: No, but I am trying to ensure compilation is not re-triggered later if env changed.

.github/workflows/coprocessor-cargo-tests.yml
antoniupop: yes, let me try this.

.github/workflows/coprocessor-cargo-tests.yml
```diff
@@ -81,41 +103,53 @@ jobs:
...
     - name: Start localstack
-      run: |
-        docker run --rm -d -p 4566:4566 --name localstack localstack/localstack:4.14.0
+      run: docker run --rm -d -p 4566:4566 --name localstack localstack/localstack:4.14.0
 
     - name: Clean previous coverage data
       run: cargo llvm-cov clean --workspace --profile coverage
       working-directory: coprocessor/fhevm-engine
 
+    - name: Compile tests with coverage instrumentation
```
PanGan21, line 118: It looks like this step is not really helping on avoiding recompilation.
I see in the ci tests compiled two times. 
In https://github.com/zama-ai/fhevm/pull/2164 the compilation will happen anyway inside the parallel runners dedicated for each service so we can remove this step if you agree

antoniupop: This step should be in each runner instead in #2164 , but each runner needs to wait for DB init/migration as well after. 
You are right that this is currently broken, must be a compile option or env difference triggering re-compile, so it needs fixing rather than removing. If we remove, then essentially this PR does nothing at all.

PanGan21: > If we remove, then essentially this PR does nothing at all.

I think though that mold is fair addition but let's indeed try to fix it. I will also check what is causing it

PanGan21: Nice it seems fixed with the latest addition!

antoniupop: Yes, I think this should be it now - just needs adjusting so you only build each crate in separate runners and hopefully we get under 30 min

antoniupop: I suspect the cargo llvm-cov was setting some different env than show-env did and that was throwing off the compiler check.

PanGan21 (approved):

antoniupop:
@mergifyio queue

PR #2205, opened 2026-03-29
chore(coprocessor): cleanup ACL allow settings and metrics

dartdart26:
Since ACL allow settings are no longer used, remove related code, metrics and DB triggers from txn-sender.

Furthermore, remove the `AddCiphertextMaterial` event from the mock contract in txn-sender as it is not being used.


antoniupop (approved):

dartdart26:
@mergifyio queue

PR #2206, opened 2026-03-29
fix(coprocessor): do not retry on `VerifyProofNotRequested`

dartdart26:
If `VerifyProofNotRequested` is returned, the txn-sender should not retry as it can never succeed. This error is not supposed to happen, but it is better if we handle it explicitly.


PanGan21 (approved):

antoniupop (approved):

dartdart26:
@mergifyio queue

PR #2218, opened 2026-03-30
feat(contracts): add remove and swap pausers tasks

zmalatrax:
Closes https://github.com/zama-ai/fhevm-internal/issues/467

## Summary

This PR adds singular `removePauser` and `swapPauser` Hardhat tasks for the host and gateway
contracts, refactors all pauser tasks to use CLI arguments (instead of environment variables)
where applicable, and introduces a shared `getPauserSetContract` helper to reduce duplication.

### Tasks

- `task:removeGatewayPauser` / `task:removeHostPauser` — removes a single pauser via `--pauser-address`
- `task:swapGatewayPauser` / `task:swapHostPauser` — swaps a single pauser via `--old-pauser-address` and `--new-pauser-address`
- `task:addGatewayPausers` / `task:addHostPausers` — unchanged interface (still uses `NUM_PAUSERS` / `PAUSER_ADDRESS_{i}` env vars), but refactored to use the shared helper

### Refactoring

- Extracted `getPauserSetContract(useInternalProxyAddress, hre)` helper in each package's `tasks/utils/loadVariables.ts` — handles compilation, deployer wallet, address loading, and contract instantiation
- All pauser tasks now call `tx.wait()` after each contract call for proper tx confirmation
- Simplified `addPausers` tasks in all three packages to use the helper
- Removed unused `removePauser` and `swapPauser` tasks from `library-solidity` (only `addHostPausers` is used there, during test setup)
- Cleaned up `.env.example` files and gateway docs — removed `OLD_PAUSER_ADDRESS_*` / `NEW_PAUSER_ADDRESS_*` env vars (no longer needed)

### Tests

- Added integration tests for add, remove, and swap pauser tasks in both `host-contracts` and `gateway-contracts`
- Tests use `Wallet.createRandom().address` for pauser addresses
- Gateway tests use `loadFixture(loadTestVariablesFixture)` for EVM state restoration (required because the ownership tests transfer GatewayConfig ownership before pauser tests run alphabetically)

### Verification

```bash
cd host-contracts && DOTENV_CONFIG_PATH=.env.example npx hardhat test test/tasks/pausers.ts
cd gateway-contracts && DOTENV_CONFIG_PATH=.env.example npx hardhat test test/tasks/pausers.ts
```


gateway-contracts/tasks/removePausers.ts
```diff
@@ -0,0 +1,41 @@
...
+    await hre.run("compile:specific", { contract: "contracts/immutable" });
+    console.log("Removing pausers from PauserSet contract");
+
+    const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
+    const numPausers = parseInt(getRequiredEnvVar("NUM_PAUSERS"));
+    const deployer = new hre.ethers.Wallet(deployerPrivateKey).connect(hre.ethers.provider);
+
+    const pausers = [];
+    for (let idx = 0; idx < numPausers; idx++) {
+      pausers.push(getRequiredEnvVar(`PAUSER_ADDRESS_${idx}`));
```
isaacdecoded: I think we want to receive the pauser address to be removed as an argument, instead of iterating/removing all the initial pausers, i.e., these tasks should be provided as operational commands.

gateway-contracts/tasks/swapPausers.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
+    const numPausers = parseInt(getRequiredEnvVar("NUM_PAUSERS"));
+    const deployer = new hre.ethers.Wallet(deployerPrivateKey).connect(hre.ethers.provider);
+
+    const pauserSwaps = [];
+    for (let idx = 0; idx < numPausers; idx++) {
+      pauserSwaps.push({
+        oldPauser: getRequiredEnvVar(`OLD_PAUSER_ADDRESS_${idx}`),
+        newPauser: getRequiredEnvVar(`NEW_PAUSER_ADDRESS_${idx}`),
```
isaacdecoded: Or maybe pass them, the old and new pauser addresses, as arguments. And I'd say just a pair of addresses instead of a list of `(old, new)[]`. I mean, using a list implicitly maps the addresses one-to-one, which I'm not sure is a usage logic we want to have (?).

Maybe you're implementing like this to avoid calling the same task multiple times if the "operator" needs to swap multiple pauser addresses?

gateway-contracts/tasks/removePausers.ts
zmalatrax: indeed, will change

gateway-contracts/tasks/swapPausers.ts
zmalatrax: agree with using arguments rather than env vars ; 
it depends if we want to swap multiple pausers at the same time or one by one, we could provide arrays as arguments

host-contracts/.env.example
```diff
@@ -30,6 +30,11 @@ PAUSER_ADDRESS_0="0xa44366bAA26296c1409AD1e284264212029F02f1" # accounts[2] (add
 # Pauser 2
 PAUSER_ADDRESS_1="0xc1d91b49A1B3D1324E93F86778C44a03f1063f1b" # accounts[3] (address)
 
+# Pauser swap inputs
+# These are used by the pauser management tasks when swapping pausers.
+OLD_PAUSER_ADDRESS_0="0xa44366bAA26296c1409AD1e284264212029F02f1" # accounts[2] (address)
+NEW_PAUSER_ADDRESS_0="0xc1d91b49A1B3D1324E93F86778C44a03f1063f1b" # accounts[3] (address)
```
obatirou: Are they used?
My understanding is that cli args are used instead now

gateway-contracts/tasks/removePauser.ts
```diff
@@ -0,0 +1,34 @@
...
+    const deployerPrivateKey = getRequiredEnvVar("DEPLOYER_PRIVATE_KEY");
+    const deployer = new hre.ethers.Wallet(deployerPrivateKey).connect(hre.ethers.provider);
+
+    if (useInternalProxyAddress) {
+      loadGatewayAddresses();
+    }
+    const pauserSetAddress = getRequiredEnvVar("PAUSER_SET_ADDRESS");
+
+    const pauserSet = await hre.ethers.getContractAt("PauserSet", pauserSetAddress, deployer);
+    await pauserSet.removePauser(pauserAddress);
```
obatirou: Shouldn't there is a wait() at some point for confirmation that the TX passed?
Especially as they are critical operations

obatirou:
I am not sure we need to add removePauser and swapPauser for library solidity 🤔 
Also seems some helpers could be implemented to avoid copy paste and having to maintain several versions across gateway and host

host-contracts/.env.example
zmalatrax: forgotten while cleaning up, ty

zmalatrax:
> I am not sure we need to add removePauser and swapPauser for library solidity 

blindly added remove/swapPauser where addPauser was added ; addPauser is used in in test setup of library-solidity, but remove/swap are not required ; I'll remove them

> Also seems some helpers could be implemented to avoid copy paste and having to maintain several versions across gateway and host

Sure, added an helper to reduce code duplication between add/remove/swapPauser tasks ; though host and gateway cannot share a common helper (different npm workspaces)

obatirou:
> Sure, added an helper to reduce code duplication between add/remove/swapPauser tasks ; though host and gateway cannot share a common helper (different npm workspaces)

Oh right indeed

gateway-contracts/tasks/removePauser.ts
zmalatrax: definitely, an oversight ; also fixed add*Pausers

zmalatrax:
fyi: having a closer look at some tasks in hardhat.config.ts of the gateway while addressing comments, I realized that there is a mismatch between the parameters name during invocation and actual arguments expected by hh tasks, it somehow worked thanks to `DOTENV_CONFIG_PATH=.env.example`, but it's brittle ; opened an issue: https://github.com/zama-ai/fhevm-internal/issues/1223

gateway-contracts/test/tasks/pausers.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    expect(await pauserSet.isPauser(pauserToRemove)).to.eq(false);
+  });
+
+  it("Should swap a pauser through the task", async function () {
+    const fixtureData = await loadFixture(loadTestVariablesFixture);
+    const owner = fixtureData.owner;
+    const oldPauser = createRandomAddress();
+    const newPauser = createRandomAddress();
+    await pauserSet.connect(owner).addPauser(oldPauser);
```
obatirou: Why tests are different in host contracts?
Here helpers are used but host use hardcoded addresses for example, they should almost be copy paste no?

gateway-contracts/test/tasks/pausers.ts
```diff
@@ -0,0 +1,64 @@
...
+    process.env.NUM_PAUSERS = "1";
+    process.env.PAUSER_ADDRESS_0 = newPauser;
+
+    await hre.run("task:addGatewayPausers", { useInternalProxyAddress: true });
+
+    expect(await pauserSet.isPauser(newPauser)).to.eq(true);
+  });
+
+  it("Should remove a pauser through the task", async function () {
+    const fixtureData = await loadFixture(loadTestVariablesFixture);
```
obatirou: It is already called in before(async function (), this could be removed?

obatirou:
LGTM, minor comments
The description of the PR should be updated also

isaacdecoded (dismissed):
LGTM! ✅

isaacdecoded (approved):

gateway-contracts/test/tasks/pausers.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  it("Should swap a pauser through the task", async function () {
+    const oldPauser = Wallet.createRandom().address;
+    const newPauser = Wallet.createRandom().address;
+    await pauserSet.connect(owner).addPauser(oldPauser);
+
+    expect(await pauserSet.isPauser(oldPauser)).to.eq(true);
+    expect(await pauserSet.isPauser(newPauser)).to.eq(false);
+
+    await hre.run("task:swapGatewayPauser", {
```
obatirou: nit: hre.run vs run in host

gateway-contracts/test/tasks/pausers.ts
```diff
@@ -0,0 +1,65 @@
...
+      pauserAddress: pauserToRemove,
+    });
+
+    expect(await pauserSet.isPauser(pauserToRemove)).to.eq(false);
+  });
+
+  it("Should swap a pauser through the task", async function () {
+    const oldPauser = Wallet.createRandom().address;
+    const newPauser = Wallet.createRandom().address;
+    await pauserSet.connect(owner).addPauser(oldPauser);
```
obatirou, line 51: nit: why connect(owner) is necessary but not in host?

obatirou (approved):
lgtm
nit comments

gateway-contracts/test/tasks/pausers.ts
zmalatrax: Need to use a different fixture to have the current owner, which is not the actual deployer of the contract when running al gateway-contracts tests (as in CI, job was failing)
> The root cause: test files run alphabetically (ownership.ts before pausers.ts). The ownership tests transfer GatewayConfig ownership away from the deployer. Without loadFixture, the pauser tests inherit that
  state and the deployer is no longer authorized.
>
>  The fix: gateway test uses loadFixture(loadTestVariablesFixture) which restores the EVM snapshot to a clean state where the deployer is still the owner. The host test can keep using getPauserSetContract
  directly since host-contracts' PauserSet access control doesn't have this cross-test dependency.

obatirou: ok thanks!

isaacdecoded:
@zmalatrax just pinging you since I can't remember what happened with this PR and why it wasn't merged 🥲

obatirou:
> @zmalatrax just pinging you since I can't remember what happened with this PR and why it wasn't merged 🥲

Ouch, does this means those are not available in 0.13 or 0.14?

gateway-contracts/test/tasks/pausers.ts
zmalatrax: fixed in https://github.com/zama-ai/fhevm/pull/2218/changes/d7fbef7874252a7e05f86737e477cb1493ab335d

isaacdecoded (dismissed):
LGTM! 👍

isaacdecoded (approved):

PR #2226, opened 2026-04-01
docs: add contributing guide and fix stale contribution links

ming1523:
## Summary

This PR fixes the repository's contribution entrypoints and removes a few stale contribution references.

## Changes

- add a root CONTRIBUTING.md as the canonical contribution guide
- fix broken contribution links in the relayer docs
- update the coprocessor contribution page to point to the current zama-ai/fhevm issue tracker
- fix a small typo in the CLA sentence

## Why

At the moment:
- the relayer docs point to a non-existent root CONTRIBUTING.md
- .github/CONTRIBUTING.md is about GitHub workflows, not contributor onboarding
- the coprocessor contribution page still references the old fhevm-backend issue tracker

This makes the contribution path harder to discover for new external contributors.

## Validation

- verified the updated contribution links point to existing files
- verified the stale fhevm-backend issue link is removed from the touched docs

PR #2230, opened 2026-04-01
docs: clarify fromExternal view limitations

### LINKED ISSUE #2026: ```view``` Modifier Incompatibility with ```FHE.fromExternal``` and FHE Precompiles (Undocumented DX Issue)
```markdown
## Problem

When writing a function that logically feels read-only — such as checking whether a submitted encrypted value meets a threshold — developers will naturally reach for the `view` modifier:

```solidity
function meetsThreshold(
    uint256 id,
    externalEuint32 req
) external view returns (ebool) {
    euint32 val = FHE.fromExternal(req, permission);
    return FHE.ge(val, thresholds[id]);
}
```

This **fails to compile** with:

```
TypeError: Function declared as view, but this expression (potentially) modifies the state.
```

The error is technically correct but non-obvious and currently undocumented.

---

## Root Cause

`FHE.fromExternal` internally calls `Impl.verify()`, which validates the external ciphertext proof against the FHE coprocessor. This writes state at the EVM level, making it incompatible with `view` — even though it doesn't mutate any application-level contract state.

The same applies to other FHE precompile operations including `FHE.add`, `FHE.sub`, `FHE.mul`, `FHE.ge`, `FHE.le`, and `FHE.eq` in certain contexts.

---

## Proposed Fix

No code change needed. This is purely a **documentation improvement**.

The Getting Started guide and `FHE.fromExternal` API reference should explicitly state:

> Functions using `FHE.fromExternal` or any FHE precompile operation cannot be marked as `view`. Although these operations do not modify your contract's application state, they perform internal proof validation that modifies EVM state under the hood. Declare such functions without `view` and invoke them as regular transactions or via `eth_call`.

Additionally, a short note on the recommended pattern for developers wanting off-chain query-like behavior would be helpful. For example, separating proof submission and threshold evaluation into two distinct steps.

---

## Environment

- fhEVM version: latest
- Toolchain: Hardhat / Foundry (standard `solc` setups)
- Network: SepoliaETH

```

ming1523:
## Summary

This PR documents that functions using `FHE.fromExternal()` should not be marked as `view`.

## Why

`FHE.fromExternal()` often looks like a read-only conversion step from a developer's perspective, but in practice it performs proof validation and is not compatible with Solidity's `view` modifier.

Without an explicit note, this can be a confusing DX trap when developers try to write helper functions that feel read-only and then hit compiler errors.

This PR adds that guidance in the main places where developers are introduced to encrypted input handling.

Fixes #2026.

## Changes

- add a warning note to `docs/sdk-guides/input.md`
- add a warning note to `docs/solidity-guides/inputs.md`
- add a warning note to the quick start tutorial step that introduces `FHE.fromExternal()`

## Validation

- verified the change is limited to documentation
- verified all added notes are colocated with existing `FHE.fromExternal()` guidance

### DIFF
--- a/docs/sdk-guides/input.md
+++ b/docs/sdk-guides/input.md
@@ -45,6 +45,9 @@ contract MyContract {
 }
 ```
 
+> Important: Functions that call `FHE.fromExternal()` or other FHE operators such as `FHE.add()` cannot be marked as `view`.
+> These operations trigger proof verification / FHE execution under the hood, so they must be executed as regular state-changing calls rather than read-only `eth_call` view functions.
+
 With `my_contract` the contract in question using `ethers` it is possible to call the add function as following.
 
 ```js
--- a/docs/solidity-guides/getting-started/quick-start-tutorial/turn_it_into_fhevm.md
+++ b/docs/solidity-guides/getting-started/quick-start-tutorial/turn_it_into_fhevm.md
@@ -220,6 +220,12 @@ FHE.fromExternal(inputEuint32, inputProof);
 
 This method verifies the zero-knowledge proof and returns a usable encrypted value within the contract.
 
+{% hint style="warning" %}
+Functions using `FHE.fromExternal()` should not be marked as `view`.
+
+Although this does not mutate your application-level contract state, the proof verification step is not compatible with Solidity's `view` modifier. Keep these functions as regular state-changing functions.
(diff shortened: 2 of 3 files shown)


ming1523:

aquint-zama:
@cla-bot check

PR #2231, opened 2026-04-01
test: guard relayer event id allocations

ming1523:
## Summary

This PR adds regression tests for relayer event ID allocation in `relayer/src/core/event.rs`.

## Why

The file already contains a TODO to add coverage for event ID conflicts.

These tests make two invariants explicit:
- non-delegated event categories should not accidentally reuse each other's event IDs
- delegated user decrypt events should remain intentionally aligned with the user decrypt event ID space

This helps protect the routing contract encoded in `event_id()` without changing runtime behavior.

## Changes

- add a test that checks the non-delegated event ID spaces are disjoint
- add a test that checks delegated user decrypt events stay mapped to the corresponding user decrypt IDs

## Validation

- verified the change is limited to test code in `relayer/src/core/event.rs`
- `cargo test` was not run in this environment because the current machine does not have a usable Rust toolchain (`cargo` not found in PATH)

### DIFF
--- a/relayer/src/core/event.rs
+++ b/relayer/src/core/event.rs
@@ -969,6 +969,7 @@ fn parse_chain_id(chain_id: &str) -> Result<u64, ParseIntError> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::collections::HashSet;
     use std::convert::TryFrom;
     use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


ming1523:

aquint-zama:
@cla-bot check

PR #2232, opened 2026-04-01
refactor: remove unwrap from advisory lock id helper

ming1523:
## Summary

This PR removes an unnecessary `unwrap()` from the relayer advisory lock ID helper.

## Why

`compute_advisory_lock_id()` derives its bytes from `Sha256::digest()`, which always returns a 32-byte digest.

That means converting the first 8 bytes into a fixed array is guaranteed to succeed in practice, so keeping an `unwrap()` there adds a panic surface without any real benefit.

Using `copy_from_slice()` keeps the behavior identical while making the helper explicitly non-panicking.

## Validation

- verified the change is behavior-preserving
- verified the existing unit tests for the helper remain applicable


ming1523:

aquint-zama:
@cla-bot check

PR #2234, opened 2026-04-01
fix(coprocessor): don't insert test host chain in db-migration image

dartdart26:
Test host chain insert should only happen when on CI or testing, it doesn't have to run on the prod invocation of the db-migration image.


dartdart26:
E2E: https://github.com/zama-ai/fhevm/actions/runs/23844284344/job/69507572453

PanGan21 (approved):

dartdart26:
@mergifyio queue

PR #2239, opened 2026-04-01
feat(relayer): initialize Helm chart for relayer with configuration files and templates

maxnovawind:


charts/relayer/values.yaml
```diff
@@ -0,0 +1,486 @@
...
+          cron_startup_delay_after_recovery: "30s"
+      devnet:
+        verbose_errors: true
+  volumeMount:
+    mountPath: "/app/config"
+    subPath: ""
+    readOnly: true
+
+# Environment variables for the deployment
+env:
```
deasydoesit, line 215: Should env handling be aligned with our approach for `kms-connector`, `coprocessor`, etc. so as not to rely on env variables for general configuration/required flag?

charts/relayer/values.yaml
```diff
@@ -0,0 +1,486 @@
...
+  args: []
+  envFrom: []
+  image:
+    repository: ""
+    tag: ""
+    pullPolicy: ""
+  resources: {}
+
+# Dynamic resources - define arbitrary Kubernetes resources
+additionalResources: []
```
deasydoesit, line 486: Worth providing a quick example

charts/relayer/values.yaml
```diff
@@ -0,0 +1,486 @@
...
+command: ["/bin/server", "--config-file", "/app/config/local.yaml"]
+
+args: []
+
+# This will set the replicaset count
+replicaCount: 1
+
+# Container image configuration
+image:
+  repository: ghcr.io/zama-ai/fhevm-relayer
```
deasydoesit, line 60: ghcr.io?

charts/relayer/values.yaml
```diff
@@ -0,0 +1,486 @@
...
+            max_concurrency: 250
+            capacity: 11000
+            safety_margin: 1000
+        contracts:
+          user_decrypt_shares_threshold: 9
+        gw_event_not_found_retry:
+          max_retries: 5
+          retry_delay_ms: 1000
+      http:
+        endpoint: "0.0.0.0:3000"
```
deasydoesit, line 139: Probably worth enforcing port standardization between here, service, and probes

charts/relayer/values.yaml
maxnovawind: Can't do this right now, the relayer is not built in the same way as the coprocessor.

charts/relayer/values.yaml
maxnovawind: Yes normally we should use this image since it's oss

charts/relayer/values.yaml
maxnovawind: Yes

PR #2240, opened 2026-04-02
chore(charts): support configurable command in db-migration job template

tawadaa:


antoniupop (approved):

piizama (approved):

tawadaa:
There's a problem with chart linter, this fix is on 0.8.x and there's an update in main that bumped the version to 0.9.x, should we skip it? @piizama

tawadaa:
closing this PR, as it's not needed for v0.12.x and was used only for experimentation

PR #2252, opened 2026-04-07
test(test-suite): add live-safe e2e variants

Eikix:
## Summary
- keep the deterministic local-only HCU block-cap tests as the strong coverage path
- make the live HCU path non-mutating by checking that standard encrypted transfers stay within the deployed caps
- make delegated-expiry work on both local and live networks with a short real expiry instead of time travel

## Details
This PR keeps the split small and explicit:
- local: strong deterministic tests where we need them
- live networks: weaker but still useful non-mutating checks where mutation is not acceptable

Changes in this PR:
- add a tiny `test/network.ts` helper to classify live networks
- for HCU:
  - local-only block-cap tests still mutate local config and use local mining controls
  - live-network coverage no longer writes shared contract state
  - live-network coverage verifies that real encrypted `transfer` / `transferFrom` calls stay within the currently deployed caps
- for delegated expiry:
  - remove the stale 1-hour/time-travel assumption
  - use a short real expiry and wait for chain time to pass on both local and live networks

## Validation
- `bash -n test-suite/e2e/run-tests.sh`
- `git diff --check`

## Notes
The full e2e package toolchain is not installed in this checkout, so Hardhat-based execution and package-level formatter/typechecker runs could not be completed locally here.

closes https://github.com/zama-ai/fhevm-internal/issues/1246

### DIFF
--- a/test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
+++ b/test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
@@ -9,6 +9,13 @@ const NOT_ALLOWED_ON_HOST_ACL = 'not_allowed_on_host_acl';
 const DELEGATION_EXPIRY_SECONDS = 75;
 const DELEGATION_EXPIRY_POLL_MS = 2_000;
 
+const relayerErrorLabel = (error: unknown): string | undefined => {
+  if (typeof error !== 'object' || error === null || !('relayerApiError' in error)) {
+    return undefined;
+  }
+  return (error as { relayerApiError?: { label?: string } }).relayerApiError?.label;
+};
+
 const waitForDelegationExpiry = async (expirationTimestamp: number) => {
   while (true) {
     const latestBlock = await ethers.provider.getBlock('latest');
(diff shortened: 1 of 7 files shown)


test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
```diff
@@ -307,6 +308,9 @@ describe('Delegated user decryption', function () {
     });
 
     it('should reject when delegation has expired', async function () {
+      if (isLiveNetwork()) {
```
PanGan21: We might want to do the same in this [test](https://github.com/zama-ai/fhevm/blob/2b2363a4aa496d92278fe5c0e6e53aab58f83128/test-suite/e2e/test/multiChain/multiChainIsolation.ts#L190) since I don't think we can force a reorg on live chain

Eikix: Done!

PanGan21: Thank you!

Eikix:
@claude /pr-review

PanGan21 (dismissed):
lgtm!

immortal-tofu (approved):

Eikix:
@mergifyio queue

PanGan21 (approved):

PR #2254, opened 2026-04-08
feat(coprocessor): implement IAM authentication for coprocessor RDS

antoniupop:


antoniupop:
@deasydoesit We need to deploy to a staging environment with an RDS instance that has IAM auth enabled -- this is the only way to test the full end-to-end flow (token generation, TLS handshake, connection pooling, token refresh). 

I've tested locally and looks like hte signed url generated is fine.

charts/coprocessor/templates/_helpers.tpl
```diff
@@ -37,3 +37,25 @@
...
+{{- end -}}
+
+{{- define "coprocessorDatabaseEnv" -}}
+{{- $authMode := include "coprocessorDatabaseAuthMode" . -}}
+- name: DATABASE_URL
+  value: {{ .Values.commonConfig.databaseUrl | quote }}
+{{- if eq $authMode "iam" }}
+- name: DATABASE_IAM_AUTH_ENABLED
+  value: "true"
+- name: DATABASE_IAM_REGION
```
piizama, line 109: The region shouldn't be explicitly required by AWS libraries, it is able to retrieve it from the environment variables (`AWS_REGION` automatically set when using an IRSA). So there should be a conditional around setting this, it could be useful only for edge cases where you connect to a database in another region.

piizama (changes requested):

charts/coprocessor/templates/_helpers.tpl
antoniupop: done

coprocessor/fhevm-engine/db-migration/prepare_database_url.sh
```diff
@@ -0,0 +1,44 @@
...
+    *)
+      return 1
+      ;;
+  esac
+}
+
+resolve_database_url_cmd() {
+  if [[ -x /usr/local/bin/resolve_database_url ]]; then
+    /usr/local/bin/resolve_database_url
+  elif [[ -x "${workspace_dir}/target/release/resolve_database_url" ]]; then
```
PanGan21, line 21: Consider also checking `target/debug/resolve_database_url`, since that is the default output path for a normal cargo build, otherwise we may miss an already built local binary and fall back to recompiling.

antoniupop: done

PanGan21 (dismissed):

deasydoesit (dismissed):

antoniupop:
@mergifyio queue

piizama (dismissed):

coprocessor/fhevm-engine/db-migration/Dockerfile
```diff
@@ -19,12 +19,20 @@ COPY coprocessor/fhevm-engine/db-migration/migrations ./migrations
 
 WORKDIR /app/coprocessor/fhevm-engine
 
+RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked \
+    --mount=type=cache,target=/app/coprocessor/fhevm-engine/target,sharing=locked \
```
PanGan21, line 24: I think we need to add the rust bindings from gateway contracts otherwise they are missing during docker build: https://github.com/zama-ai/fhevm/actions/runs/25011054567/job/73249134834?pr=2379

antoniupop: You're right - and I should have tested builds before pushing on queue...

PanGan21 (approved):

coprocessor/fhevm-engine/db-migration/Dockerfile
antoniupop: Looks fixed now (https://github.com/zama-ai/fhevm/actions/runs/25039151502).

antoniupop:
@mergifyio queue

PR #2264, opened 2026-04-08
chore(coprocessor): multi copro listener chart updates

deasydoesit:
This PR goes to supporting multi-chain host listeners:
   
  - Updates Host-listener deployments (`host-listener`, `host-listener-poller`, `host-listener-catchup-only`) to iterate over a per-chain config so each chain gets its own `deployment`, `service`, and `ServiceMonitor` for each listener type.   
  - Adds validation (`coprocessor-host-listener-validation.yaml`) to check for the new multi-chain values structure.                                                                                                   
  - Updates `values.yaml` to restructure `host-listener` values to a multi-chain shape.

### DIFF
--- a/charts/coprocessor/Chart.yaml
+++ b/charts/coprocessor/Chart.yaml
@@ -1,6 +1,6 @@
 name: coprocessor
 description: A helm chart to distribute and deploy Zama fhevm Co-Processor services
-version: 0.9.0
+version: 0.10.0
 apiVersion: v2
 keywords:
   - fhevm
--- a/charts/coprocessor/templates/_helpers.tpl
+++ b/charts/coprocessor/templates/_helpers.tpl
@@ -5,17 +5,69 @@
 
 {{- define "hostListenerName" -}}
 {{- $hostListenerNameDefault := printf "%s-%s" .Release.Name "host-listener" }}
-{{- default $hostListenerNameDefault .Values.hostListener.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- default $hostListenerNameDefault .Values.hostListenerShared.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
 {{- define "hostListenerPollerName" -}}
 {{- $hostListenerPollerNameDefault := printf "%s-%s" .Release.Name "host-listener-poller" }}
-{{- default $hostListenerPollerNameDefault .Values.hostListenerPoller.nameOverride | trunc 63 | trimSuffix "-" -}}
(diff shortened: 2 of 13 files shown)


maxnovawind:
### Code review

Found 2 issues:

1. `args.serviceName` is defined in `values.yaml` for all three host-listener variants but silently ignored by the templates, which hardcode `--service-name=host-listener-{{ $chain.name }}` instead. The `gw-listener` template correctly uses `{{ .Values.gwListener.args.serviceName }}`, showing this is an inconsistency rather than intentional. Any operator who customizes `serviceName` for host-listeners gets no effect, and existing deployments will see their OTLP trace service names change on upgrade.

`values.yaml` defines the unused field:
https://github.com/zama-ai/fhevm/blob/8fef7da51d9ffce0fbd3b9d8e55b88aebb1a2c31/charts/coprocessor/values.yaml#L195-L199

Template hardcodes instead of reading from values:
https://github.com/zama-ai/fhevm/blob/8fef7da51d9ffce0fbd3b9d8e55b88aebb1a2c31/charts/coprocessor/templates/coprocessor-host-listener-deployment.yaml#L66-L68

Contrast with gw-listener which does it correctly:
https://github.com/zama-ai/fhevm/blob/8fef7da51d9ffce0fbd3b9d8e55b88aebb1a2c31/charts/coprocessor/templates/coprocessor-gw-listener-deployment.yaml#L62-L64

2. Top-level `hostListener.enabled`, `hostListenerPoller.enabled`, and `hostListenerCatchupOnly.enabled` flags are dead config -- templates only gate on `$chain.hostListener.enabled` inside the `chains[]` range loop. Combined with `chains: []` as the default, an existing deployment that had `hostListener.enabled: true` but no `chains` entries will silently render zero host-listener Deployments on upgrade. Helm will delete the running pods with no error or warning.

Dead `enabled` flag in values.yaml:
https://github.com/zama-ai/fhevm/blob/8fef7da51d9ffce0fbd3b9d8e55b88aebb1a2c31/charts/coprocessor/values.yaml#L164-L166

Template gates on per-chain enabled only:
https://github.com/zama-ai/fhevm/blob/8fef7da51d9ffce0fbd3b9d8e55b88aebb1a2c31/charts/coprocessor/templates/coprocessor-host-listener-deployment.yaml#L1-L3

Empty default that silently produces zero resources:
https://github.com/zama-ai/fhevm/blob/8fef7da51d9ffce0fbd3b9d8e55b88aebb1a2c31/charts/coprocessor/values.yaml#L517-L519

piizama (approved):

eudelins-zama (approved):

deasydoesit:
@mergifyio queue

PR #2243, opened 2026-04-03
feat(host-contracts): add ProtocolConfig and KMSGeneration contracts

obatirou:
Introduce host-side `ProtocolConfig` (KMS context management, thresholds, node registry) and `KMSGeneration` (keygen/CRS lifecycle with extraData replay protection). Includes interfaces, deploy tasks for both host-contracts and library-solidity, Hardhat upgrade tests with real initialization and state preservation checks, Foundry deploy helpers and tests, rust bindings, and contract selectors.         

### ProtocolConfig host-port notes                                                     
  
ProtocolConfig intentionally drops 3 view functions present in GatewayConfig: `getDecryptionThresholdForContext`, `getKmsGenThresholdForContext`, and `getKmsTxSendersForContext`. None have an on-chain consumer. The `contextTxSenders` storage mapping was also removed to avoid writing to storage that nothing reads.                  

### KMSGeneration host-port notes                                                      
Use the following command to view the diff 
```bash
diff -u gateway-contracts/contracts/KMSGeneration.sol host-contracts/contracts/KMSGeneration.sol
diff -u gateway-contracts/contracts/interfaces/IKMSGeneration.sol host-contracts/contracts/interfaces/IKMSGeneration.sol
diff -u gateway-contracts/tasks/generateKmsMaterials.ts host-contracts/tasks/generateKmsMaterials.ts
```
This PR keeps host-contracts/contracts/KMSGeneration.sol close to  gateway-contracts/contracts/KMSGeneration.sol, with the following intentional host-specific differences:
- ProtocolConfig replaces GatewayConfig for KMS node, threshold, and context lookups.
- ACLOwnable replaces GatewayOwnable: onlyKmsTxSender is implemented locally.
- EIP-712 payloads for keygen, and crsgen include bytes extraData (0x02 || contextId || epochId)
- requestExtraData is stored per request ID and is used both for signature verification and for resolving the original KMS context when building storage URLs after a context rotation. Destroyed contexts are no longer queryable.                                                    
- For the key lifecycle, paramsType and extraData are stored under both prepKeygenId and keyId.
- Host-only controls add abortKeygen(), abortCrsgen(), and hasPendingKeyManagementRequest().
- Gateway-only temporary flows prssInit() and keyReshareSameSet() are not migrated.  
- Migration support is provided via initializeFromMigration(MigrationState).         
  
### Migration review notes                                                           

- To keep steady-state review close to the gateway contract, migration-only code is isolated in a dedicated section at the end of the contract.
[...]

PR #2267, opened 2026-04-08
feat(kms-connector): keygen on ethereum

eudelins-zama:
## Description of changes

This PR migrates key generation event listening and transaction sending from the Gateway chain to Ethereum (L1), implementing dual-chain routing in the KMS Connector.

**Event Listening (`gw-listener`)**
- `EthereumListener` now polls `KMSGeneration` events (`PrepKeygenRequest`, `KeygenRequest`, `CrsgenRequest`) directly from Ethereum using finalized blocks, with its own cancellation token and error handling.
- `PrssInit` and `KeyReshareSameSet` handling have been removed (see RFC-005 and RFC-013) for more context
- `GatewayListener` is simplified to only handle `Decryption` events — the `MonitoredContract` enum and `JoinSet`-based multi-contract polling are removed.
- Healthcheck now monitors both Gateway and Ethereum RPC providers.

**Transaction Sending (`tx-sender`)**
- The monolithic `TransactionSenderInner` is split into `GatewayTransactionSender` (decryption responses) and `EthereumTransactionSender` (keygen responses), each with its own provider and config.
- `EthereumTransactionSender` waits for finalized confirmations (configurable, default 64) with a receipt timeout before considering a tx confirmed.
- `TransactionSender::forward_response` routes responses to the correct chain based on the response kind.
- New config fields: `ethereum_url`, `ethereum_chain_id`, `ethereum_tx_required_confirmations`, `ethereum_tx_get_receipt_timeout`.

**Event Model Cleanup**
- Removed `PrssInit` and `KeyReshareSameSet` event types, DB queries, gRPC handlers, and test cases throughout the stack.
- Added `extra_data` column to `prep_keygen_requests`, `keygen_requests`, and `crsgen_requests` tables via a new SQL migration.
- Keygen event bindings now come from `fhevm_host_bindings` (Ethereum contracts) instead of `fhevm_gateway_bindings`.

**Testing**
- Added `KMSGenerationMock.sol` and `KMSVerifierMock.sol` test contracts for Ethereum-side integration tests.

Closes https://github.com/zama-ai/fhevm-internal/issues/1229
Closes https://github.com/zama-ai/fhevm-internal/issues/1119
Closes https://github.com/zama-ai/fhevm-internal/issues/1167


kms-connector/config/tx-sender.toml
```diff
@@ -12,6 +12,24 @@ gateway_url = "http://localhost:8545"
 # ENV: KMS_CONNECTOR_GATEWAY_CHAIN_ID
 gateway_chain_id = 54321
 
+# Ethereum RPC nod endpoint (required)
```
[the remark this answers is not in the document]
eudelins-zama: Fixed c2da1437d9850908cffd20e90a6f7b95ffd20a92

kms-connector/config/tx-sender.toml
```diff
@@ -12,6 +12,24 @@ gateway_url = "http://localhost:8545"
...
+
+# Number of confirmations required for Ethereum transactions to be considered confirmed
+# (optional, defaults to 64)
+# ENV: KMS_CONNECTOR_ETHEREUM_TX_REQUIRED_CONFIRMATIONS
+# ethereum_tx_required_confirmations = 64
+
+# Timeout for getting Ethereum transaction receipts with the required number of confirmations
+# (optional, defaults to 5m)
+# ENV: KMS_CONNECTOR_ETHEREUM_TX_GET_RECEIPT_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
+# ethereum_tx_get_receipt_timeout = "5m"
```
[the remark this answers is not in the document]
eudelins-zama: Fixed c2da1437d9850908cffd20e90a6f7b95ffd20a92

kms-connector/crates/tx-sender/src/core/ethereum.rs
```diff
@@ -0,0 +1,350 @@
...
+            self.config.gas_multiplier_percent,
+            &mut call,
+        )
+        .await?;
+
+        let tx = self.provider.send_transaction(call).await?;
+        info!("Tx sent to RPC node. Waiting for finalized receipt (~64 confirmations)...");
+        let receipt = tx
+            .with_required_confirmations(self.config.tx_required_confirmations)
+            .with_timeout(Some(self.config.get_receipt_timeout))
```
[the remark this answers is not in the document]
eudelins-zama: Fixed c2da1437d9850908cffd20e90a6f7b95ffd20a92

kms-connector/crates/tx-sender/src/core/ethereum.rs
```diff
@@ -0,0 +1,350 @@
...
+        // Mock out of gas tx
+        let tx_sender = EthereumTransactionSender::new(
+            mock_provider.clone(),
+            KMSGenerationInstance::new(Address::default(), mock_provider.clone()),
+            EthereumSenderConfig {
+                tx_retries: 1,
+                trace_reverted_tx: true,
+                gas_multiplier_percent: 105,
+                ..Default::default()
+            },
```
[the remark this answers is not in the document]
eudelins-zama: Fixed c2da1437d9850908cffd20e90a6f7b95ffd20a92

kms-connector/connector-db/migrations/20260407130028_key_generation_ethereum_migration.sql
```diff
@@ -0,0 +1,16 @@
...
+
+ALTER TABLE keygen_requests ADD COLUMN extra_data BYTEA;
+UPDATE keygen_requests SET extra_data = NULL;
+
+ALTER TABLE crsgen_requests ADD COLUMN extra_data BYTEA;
+UPDATE crsgen_requests SET extra_data = NULL;
+
+-- prss_init and key_reshare_same_set tables will be remove in next release (v0.14.0).
+-- We keep them for now for compatibility of the DB migration with v0.12 release.
+-- Tracking issue: TODO.
```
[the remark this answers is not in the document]
eudelins-zama: Fixed c2da1437d9850908cffd20e90a6f7b95ffd20a92

eudelins-zama:
@claude /pr-review

kms-connector/crates/tx-sender/src/core/ethereum.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    #[tracing::instrument(skip_all)]
+    pub async fn send_to_ethereum(&self, response: KmsResponseKind) -> Result<(), Error> {
+        info!("Sending response to Ethereum: {response:?}");
+        let tx_result = match response {
+            KmsResponseKind::PrepKeygen(response) => self.send_prep_keygen_response(response).await,
+            KmsResponseKind::Keygen(response) => self.send_keygen_response(response).await,
+            KmsResponseKind::Crsgen(response) => self.send_crsgen_response(response).await,
+            _ => unreachable!("Only keygen responses should be sent to Ethereum"),
```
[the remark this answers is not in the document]
eudelins-zama: This is on purpose. We expect very few tx to be sent to Ethereum (at the beginning at least), we are talking about few txs per year.
So no need for metrics and so on right now

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -42,18 +93,192 @@ where
...
+                .bind(event_type)
+                .fetch_one(&self.db_pool)
+                .await?
+                .try_get::<Option<i64>, _>("block_number")?;
+
+        let Some(block_number) = query_result else {
+            info!("No block number stored in DB yet for {event_type}");
+            return Ok(None);
+        };
+        Ok(Some(block_number as u64))
```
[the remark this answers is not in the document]
eudelins-zama: As it's the (trusted) gw-listener that inserted the `block_number` in the first place (and that was an `u64`), I think we're good here.

AFAIK, only thing that could go wrong is if the initial `u64` is bigger than `i64::MAX`, but from what I remember, we ruled out this was possible

kms-connector/config/tx-sender.toml
```diff
@@ -12,6 +12,24 @@ gateway_url = "http://localhost:8545"
 # ENV: KMS_CONNECTOR_GATEWAY_CHAIN_ID
 gateway_chain_id = 54321
 
+# Ethereum RPC node endpoint (required)
+# ENV: KMS_CONNECTOR_ETHEREUM_URL
+ethereum_url = "http://localhost:8545"
+
+# Chain ID for the Ethereum network (required)
+# ENV: KMS_CONNECTOR_ETHEREUM_CHAIN_ID
+ethereum_chain_id = 11155111
```
dvdplm, line 21: Where does `111555111` come from? Isn't the eth chain id `0x1`?

kms-connector/crates/gw-listener/src/monitoring/health.rs
```diff
@@ -1,35 +1,54 @@
...
 
-impl<P: Provider> Healthcheck for State<P> {
+impl<GP: Provider, EP: Provider> Healthcheck for State<GP, EP> {
     async fn healthcheck(&self) -> actix_web::HttpResponse {
         let mut errors = vec![];
         let database_connected =
             database_healthcheck(&self.db_pool, self.healthcheck_timeout, &mut errors).await;
-        let gateway_connected =
-            gateway_healthcheck(&self.provider, self.healthcheck_timeout, &mut errors).await;
+        let gateway_connected = rpc_node_healthcheck(
```
dvdplm, line 40: Not your code, but the errors in `rpc_node_healthcheck` references "Gateway" – now that they are (potentially) separate it might be confusing if both are down at the same time? Seems like `Provider` does not impl `Debug` but perhaps there's a way to make it nice? (`Network` does impl `Debug` for instance).

kms-connector/crates/gw-listener/src/monitoring/health.rs
```diff
@@ -1,35 +1,54 @@
...
+            &self.gateway_provider,
+            self.healthcheck_timeout,
+            &mut errors,
+        )
+        .await;
+        let ethereum_connected = rpc_node_healthcheck(
```
dvdplm, line 47: I guess these two could run in parallel, with `tokio::join` or some such. Premature optimization?

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -204,47 +148,28 @@ where
...
-                }
-            }
+    fn decode_log(log: &Log) -> anyhow::Result<ProtocolEventKind> {
+        let event = DecryptionEvents::decode_log(&log.inner)
+            .map_err(|e| anyhow!("Failed to decode Decryption event: {e}"))?;
+        match event.data {
+            DecryptionEvents::PublicDecryptionRequest(e) => Ok(e.into()),
+            DecryptionEvents::UserDecryptionRequest(e) => Ok(e.into()),
+            _ => Err(anyhow!("Unexpected Decryption event: {log:?}")),
         }
```
dvdplm: This PR is very clean so I'm struggling to come up with any decent nitpicks! :) Here, we *could* add a `TryFrom<DecryptionEvent>` impl so that you'd end up with:

```rust
let event = DecryptionEvent::decode_log(…);
event.data.try_into()
```

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -17,18 +54,32 @@ where
...
             kms_verifier_contract,
         }
     }
 
     /// Starts the `EthereumListener`.
     pub async fn start(self) {
-        // No listening for now, will be done when implementing RFC-005.
-
+        select! {
+            biased;
```
dvdplm, line 76: I had no idea this existed. TIL!

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -42,18 +93,192 @@ where
...
+                        .inc();
+                    consecutive_errors = consecutive_errors.saturating_add(1);
+                    warn!("KMSGeneration listening error: {e} ({consecutive_errors}/{max_errors})");
+                    if consecutive_errors >= max_errors {
+                        anyhow::bail!("Too many consecutive errors for KMSGeneration polling");
+                    }
+                }
+            }
+        }
+    }
```
dvdplm, line 143: This is very similar to the Gateway version. Is that a concern?

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -42,18 +93,192 @@ where
...
+
+    /// Fetches logs for a block range, decodes them, and publishes them in a single transaction.
+    ///
+    /// Returns `(new_from_block, has_more_blocks)`.
+    async fn fetch_and_publish(
```
dvdplm, line 148: Also very similar. Maybe there's an opportunity to share some code?

kms-connector/crates/gw-listener/src/core/ethereum.rs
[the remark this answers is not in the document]
dvdplm: I think this is fine, `9_223_372_036_854_775_807` is large enough to feel safe here.

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -83,4 +313,90 @@ mod tests {
 
         assert!(row.try_get::<bool, _>("is_valid").unwrap());
     }
+
+    #[rstest::rstest]
+    #[timeout(Duration::from_secs(90))]
```
dvdplm, line 290: This is a very generous timeout... sure you don't want to abort this sooner if things are broken?

kms-connector/Cargo.toml
```diff
@@ -36,8 +38,8 @@ alloy = { version = "=1.0.38", default-features = false, features = [
     "signer-aws",
     "std",
 ] }
-anyhow = { version = "=1.0.98", default-features = false }
-async-trait = { version = "=0.1.88", default-features = false }
+anyhow = { version = "=1.0.100", default-features = false }
```
dvdplm, line 39: Could go all the way to 1.0.102 here!

kms-connector/crates/utils/src/tests/setup/blockchain.rs
```diff
@@ -37,19 +37,22 @@ pub struct GatewayInstance {
...
     pub gateway_config_contract: GatewayConfigInstance<WalletProvider>,
     pub kms_generation_contract: KMSGenerationInstance<WalletProvider>,
+    pub kms_verifier_address: Address,
     pub anvil: ContainerAsync<GenericImage>,
     pub anvil_host_port: u16,
     pub block_time: u64,
 }
 
 impl GatewayInstance {
+    #[allow(clippy::too_many_arguments)]
```
dvdplm: I like the new `#[expect(clippy::some_lint)]` so that if/when the code changes so that the lint no longer triggers, you get a compilation warning.

kms-connector/crates/utils/src/tests/setup/blockchain.rs
```diff
@@ -173,6 +192,8 @@ async fn setup_anvil(block_time: u64) -> anyhow::Result<ContainerAsync<GenericIm
             TEST_MNEMONIC,
             "--block-time",
             &format!("{block_time}"),
+            "--slots-in-an-epoch",
```
dvdplm, line 197: What does this do? Worth a comment you think?

kms-connector/crates/utils/src/monitoring/health.rs
```diff
@@ -43,10 +43,12 @@ pub async fn database_healthcheck(
     }
 }
 
-/// Performs the Gateway healthcheck.
+/// Performs the healthcheck of a blockchain RPC node.
+///
+/// Uses `eth_getBlockNumber` for this.
```
dvdplm: I think it is `eth_BlockNumber`?

kms-connector/crates/tx-sender/tests/integration_tests.rs
```diff
@@ -354,29 +352,55 @@ async fn start_test_tx_sender(
...
-            gas_multiplier_percent: 130,
-        },
+    // Ethereum provider + KMSGeneration contract (same Anvil in tests)
+    let eth_sender_config = EthereumSenderConfig {
+        tx_retries: 3,
+        tx_retry_interval: Duration::from_millis(100),
+        trace_reverted_tx: true,
+        gas_multiplier_percent: 130,
+        tx_required_confirmations: 2,
+        get_receipt_timeout: Duration::from_mins(1),
```
dvdplm, line 382: That is a very generous timeout! Is it ok to wait for a full minute?

kms-connector/crates/tx-sender/src/monitoring/health.rs
```diff
@@ -34,8 +37,18 @@ impl Healthcheck for State {
...
+            self.healthcheck_timeout,
+            &mut errors,
+        )
+        .await;
+        let ethereum_connected = rpc_node_healthcheck(
+            &self.ethereum_provider,
+            self.healthcheck_timeout,
+            &mut errors,
+        )
+        .await;
```
dvdplm, line 54: Could run in parallel perhaps?

kms-connector/crates/tx-sender/src/core/tx_sender.rs
```diff
@@ -98,34 +95,52 @@ where
...
+    /// Spawns a new task to forward each response to the appropriate chain.
     async fn spawn_response_forwarding_tasks(
         &self,
         responses: Vec<KmsResponse>,
         cancel_token: &CancellationToken,
     ) {
         for response in responses {
-            let inner = self.inner.clone();
+            let gw_sender = self.gw_sender.clone();
+            let eth_sender = self.eth_sender.clone();
```
dvdplm, line 106: Are these cheap to clone? At a glance they seem somewhat heavy. OTOH it doesn't seem obvious to avoid without also working around `spawn_with_limit`'s requirements. How big is `responses`? If it's not super big, maybe you could call `send_to_gateway`/`send_to_ethereum` directly here? Not sure it's worth it, just a random idea.

kms-connector/crates/tx-sender/src/core/tx_sender.rs
```diff
@@ -143,301 +158,63 @@ where
...
-        let provider =
-            connect_to_rpc_node_with_wallet(config.gateway_url, config.gateway_chain_id, wallet)
-                .await?;
+        // Gateway provider + Decryption contract
+        let gw_sender_config = GatewaySenderConfig {
+            tx_retries: config.tx_retries,
+            tx_retry_interval: config.tx_retry_interval,
+            trace_reverted_tx: config.trace_reverted_tx,
+            gas_multiplier_percent: config.gas_multiplier_percent,
+        };
```
dvdplm: Could add a `GatewaySenderConfig::from(Config)` perhaps?

kms-connector/crates/tx-sender/src/core/tx_sender.rs
```diff
@@ -143,301 +158,63 @@ where
...
-                self.send_public_decryption_response(response).await
-            }
-            KmsResponseKind::UserDecryption(response) => {
-                self.send_user_decryption_response(response).await
-            }
-            KmsResponseKind::PrepKeygen(response) => self.send_prep_keygen_response(response).await,
-            KmsResponseKind::Keygen(response) => self.send_keygen_response(response).await,
-            KmsResponseKind::Crsgen(response) => self.send_crsgen_response(response).await,
+        // Ethereum provider + KMSGeneration contract
+        let eth_sender_config = EthereumSenderConfig {
```
dvdplm: Same here? A `impl From<Config> for EthereumSenderConfig`?

kms-connector/crates/tx-sender/src/core/config.rs
```diff
@@ -174,6 +189,21 @@ fn default_gc_decryption_under_process_limit() -> PgInterval {
     PgInterval::try_from(Duration::from_mins(6)).unwrap()
 }
 
+fn default_ethereum_tx_get_receipt_timeout() -> Duration {
+    // Max time required for a transaction to be considered finalized on Ethereum.
+    // Ethereum uses 12-second slots and 32 slots per epoch (about 6.4 minutes). So 2 epochs take
+    // usually 12-13 minutes. Thus, a 20 minutes timeout is reasonable.
```
dvdplm, line 195: 👍

kms-connector/crates/tx-sender/src/core/config.rs
```diff
@@ -184,6 +214,10 @@ impl Default for Config {
             gateway_url: Url::from_str("http://localhost:8545").unwrap(),
             gateway_chain_id: 54321,
             decryption_contract: default_decryption_contract_config(),
+            ethereum_url: Url::from_str("http://localhost:8545").unwrap(),
+            ethereum_chain_id: 11155111,
```
dvdplm:
```suggestion
            ethereum_chain_id: 11155111, // Sepolia
```

dvdplm (commented):
Partial review. Will be back.

kms-connector/crates/tx-sender/src/core/tx_sender.rs
eudelins-zama: Yes it should be cheap to clone a Provider: https://alloy.rs/guides/rpc-provider-abstraction#provider-does-not-require-arc

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Two main diff:
- We will soon update `ethereum.rs` to support listening to `KMSVerifier` contract as well. So there will be more differences introduced soon.
- Main difference is the listening from finalized block only to avoid re-org handling

kms-connector/Cargo.toml
eudelins-zama: No, we must align with `kms-grpc` unfortunately (as discussed in DM last time), I don't have this luxury! :smile:

kms-connector/crates/utils/src/tests/setup/blockchain.rs
eudelins-zama: Good idea indeed, did not know about it, thx!

kms-connector/crates/utils/src/tests/setup/blockchain.rs
eudelins-zama: Yes indeed. As we listen to finalized block in `ethereum.rs`, we need to reduce the slots in an epoch (defaults to 32) if we don't want to wait 12min for events to be caught by the listener.

Will comment, and update this param in the e2e setup as well!

kms-connector/crates/gw-listener/src/monitoring/health.rs
eudelins-zama: Makes sense to me, will do

kms-connector/config/tx-sender.toml
eudelins-zama: Discussed in DM

kms-connector/crates/gw-listener/src/monitoring/health.rs
eudelins-zama: Indeed, will try to improve the error

kms-connector/crates/utils/src/monitoring/health.rs
eudelins-zama: Good catch!

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Will reduce to 30. But as we spawn containers for these tests I prefer to be generous to avoid flaky tests

kms-connector/crates/gw-listener/src/monitoring/health.rs
eudelins-zama: Created this tracking issue instead: https://github.com/zama-ai/fhevm-internal/issues/1262

Cause it concerns all 3 services of the kms-connector AFAIK, so I want to avoid making this PR even bigger

kms-connector/crates/utils/src/tests/setup/blockchain.rs
eudelins-zama: Done 972dde96ac67163cb13444a00abd1cab3f168a32

kms-connector/crates/utils/src/tests/setup/blockchain.rs
eudelins-zama: Fixed ee20fcc48ea2e77dd26839de0b4743abaa2400ad

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Fixed ee20fcc48ea2e77dd26839de0b4743abaa2400ad

kms-connector/crates/tx-sender/tests/integration_tests.rs
eudelins-zama: Fixed ee20fcc48ea2e77dd26839de0b4743abaa2400ad

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Now tests are failing with a timeout :joy: https://github.com/zama-ai/fhevm/actions/runs/24190776989/job/70607367912?pr=2267

Will revert to 90

eudelins-zama: Done ae6bd45c0209de0a2ef5ddca7edfdd6f19ce74ad

kms-connector/crates/utils/src/monitoring/health.rs
eudelins-zama: Fixed ee20fcc48ea2e77dd26839de0b4743abaa2400ad

kms-connector/crates/tx-sender/src/monitoring/health.rs
eudelins-zama: Created this tracking issue instead: https://github.com/zama-ai/fhevm-internal/issues/1262

Cause it concerns all 3 services of the kms-connector AFAIK, so I want to avoid making this PR even bigger

kms-connector/crates/tx-sender/src/core/tx_sender.rs
eudelins-zama: Fixed 132aad1d64619921545db4b8fcefe66c6a936b1f

kms-connector/crates/tx-sender/src/core/tx_sender.rs
eudelins-zama: Fixed 132aad1d64619921545db4b8fcefe66c6a936b1f

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Would like to wait for the upcoming update of the `ethereum.rs` (mentioned [here](https://github.com/zama-ai/fhevm/pull/2267#discussion_r3057624851)) to see what can easily be de-duplicated once the changes are made.

But wouldn't rush for this and then realize that it lead to ugly code that is better to be "duplicated" (with slight difference ofc).

Does it make sense to you?

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: Fixed ee20fcc48ea2e77dd26839de0b4743abaa2400ad

kms-connector/crates/gw-listener/src/monitoring/health.rs
eudelins-zama: Fixed f877a717449260be50af5d2625e6bd53e88d1a5c

kms-connector/crates/tx-sender/src/core/config.rs
eudelins-zama: Fixed 132aad1d64619921545db4b8fcefe66c6a936b1f

kms-connector/Cargo.toml
dvdplm: Sigh, true.

kms-connector/crates/tx-sender/src/core/tx_sender.rs
dvdplm: Cool, one headache less.

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -42,18 +93,183 @@ where
...
+        }
+
+        match min_last_processed_block {
+            Some(last_block_polled) => Ok(last_block_polled.saturating_add(1)),
+            None => {
+                let finalized = self
+                    .provider
+                    .get_block_by_number(BlockNumberOrTag::Finalized)
+                    .await?
+                    .ok_or_else(|| anyhow!("Finalized block not available"))?
```
dvdplm: What happens here if the last block in the DB *changes* during the for-loop? Isn't it safer (and more efficient) to fetch the block from the DB outside the loop for all `EventType`s in one query, and then match the SQL result rows to the right `EventType`? More complex for sure, but safer?

kms-connector/crates/tx-sender/src/core/gateway.rs
```diff
@@ -0,0 +1,349 @@
...
+        );
+        debug!("Calldata length {}", call_builder.calldata().len());
+
+        let call = call_builder.into_transaction_request();
+        self.send_tx_sync_with_retry(call).await
+    }
+
+    /// Sends the requested transaction with retries.
+    ///
+    /// The `gas_limit` is increased at each attempts.
```
dvdplm: It'd be good to expand on this; the reader will wonder "by how much"...

How about: "The `gas_limit` is re-estimated and (likely) increased at each attempt"?

kms-connector/crates/tx-sender/src/core/ethereum.rs
```diff
@@ -0,0 +1,367 @@
...
+use anyhow::anyhow;
+use connector_utils::{
+    provider::NonceManagedProvider,
+    types::{CrsgenResponse, KeygenResponse, KmsResponseKind, PrepKeygenResponse},
+};
+use fhevm_host_bindings::kms_generation::KMSGeneration::KMSGenerationInstance;
+use std::time::Duration;
+use tracing::{debug, error, info, warn};
+
+/// The internal struct used to send keygen transactions to Ethereum.
```
dvdplm: Why "internal"? It's `pub` after all?

kms-connector/crates/tx-sender/src/core/ethereum.rs
```diff
@@ -0,0 +1,367 @@
...
+
+#[derive(Clone, Default)]
+pub struct EthereumSenderConfig {
+    pub tx_retries: u8,
+    pub tx_retry_interval: Duration,
+    pub trace_reverted_tx: bool,
+    pub gas_multiplier_percent: usize,
+    pub tx_required_confirmations: u64,
+    pub get_receipt_timeout: Duration,
+}
```
dvdplm, line 35: Some docs here?

kms-connector/crates/tx-sender/src/core/config.rs
```diff
@@ -116,8 +131,8 @@ where
 impl DeserializeConfig for Config {}
 
 impl Config {
-    pub async fn build_wallet(&self) -> Result<KmsWallet, Error> {
-        let chain_id = Some(self.gateway_chain_id);
+    pub async fn build_wallet(&self, chain_id: u64) -> Result<KmsWallet, Error> {
+        let chain_id = Some(chain_id);
```
dvdplm, line 135: Not your code but I find it a bit odd that the wallet creation methods take an `Option`. They do because they are calling `alloy` methods that take `Option`, but why? What does it even mean to create a wallet with chain ID of `None`?

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: > Isn't it safer (and more efficient) to fetch the block from the DB outside the loop

I think this is what we already do right? :sweat_smile: 
Or did I miss something?

kms-connector/crates/tx-sender/src/core/ethereum.rs
eudelins-zama: Will be duplicated from the main `Config`, not sure if it worth the duplication. Maybe?

kms-connector/crates/tx-sender/src/core/config.rs
eudelins-zama: `alloy` can "guess" the chainId by making a RPC call `eth_chainId`. Thus they make it optional AFAIU.

But to avoid this extra call at each tx, we provide it ourselves.

kms-connector/crates/tx-sender/src/core/ethereum.rs
eudelins-zama: Fixed 15961ed154688d3f0f2ed8b6dc65ef4384a0d613

kms-connector/crates/tx-sender/src/core/gateway.rs
eudelins-zama: Fixed 15961ed154688d3f0f2ed8b6dc65ef4384a0d613

kms-connector/crates/tx-sender/src/core/ethereum.rs
dvdplm: Good point. Hmm. Fair enough, perhaps it's actually better to have docs in a single place.

kms-connector/crates/gw-listener/src/core/ethereum.rs
dvdplm: My comment ended up on the wrong line; I use vscode to do code reviews and I think it might have failed to update the branch or something? 

Anyway, my comment should have been [here](https://github.com/zama-ai/fhevm/pull/2267/changes/BASE..132aad1d64619921545db4b8fcefe66c6a936b1f#diff-e09c0331b9004069190fb9c30b036b5dcedba7ebc9b9361a89aa8ee9b29d281dR215).

```rust
        for &event_type in event_types {
            if let Some(last) = self.get_last_block_polled_from_db(event_type).await? { // <–– DB fetch, once per event type
                min_last_processed_block = match min_last_processed_block {
                    Some(current) => Some(std::cmp::min(current, last)),
                    None => Some(last),
                };
            }
        }
```

My question was: the `get_last_block_polled_from_db` call fetches a single row from the DB, so in between iterations in the `for-loop` the data  in the DB could  change. If it does, is it bad? Maybe? Maybe not, given they are different event types? 

But either way, I was thinking, maybe we can think of some clever SQL that gets you the earliest processed block for all event types in one query, that way we don't even need to think about pesky potential data race issues.

eudelins-zama: Unless we use multiple replicas of listeners (which is not the case right now), this should not happen as we update the DB only in the main event listening loop that start after this one.

But if that happen, since we're computing the min, the worst case of a race is that we re-scan a few blocks we have already processed. Should be OK IMO.

That being said, it was fairly easy to update, and we end up with a cleaner code, so thx for the suggestion!

Fixed in db32fefbb129928a6ce63d5aa8b141a55379beff

dvdplm (approved):
LGTM

kms-connector/tests/contracts/KMSGenerationMock.sol
```diff
@@ -0,0 +1,82 @@
...
+        Default,
+        Test
+    }
+
+    enum KeyType {
+        Server,
+        Public
+    }
+
+    event PrepKeygenRequest(uint256 prepKeygenId, uint256 epochId, ParamsType paramsType, bytes extraData);
```
obatirou: @eudelins-zama https://github.com/zama-ai/fhevm/pull/2243#discussion_r3056696572
Do you agree with this?
cc @isaacdecoded

eudelins-zama: Yes I agree :slightly_smiling_face: 

Once you will have merged your PR on the feat branch, I'll use `path` dependencies again in the `Cargo.toml` (instead of `git` currently).

So such diff should make tests break, I'll update the mocks at that time :wink:

obatirou: Perfect, thanks!

eudelins-zama: Done

kms-connector/tests/contracts/KMSGenerationMock.sol
```diff
@@ -0,0 +1,81 @@
...
+
+        emit PrepKeygenRequest(prepKeygenId, paramsType, "");
+    }
+
+    function prepKeygenResponse(uint256 prepKeygenId, bytes calldata signature) external {
+        address kmsTxSender;
+        keyCounter++;
+        uint256 keyId = keyCounter;
+
+        emit PrepKeygenResponse(prepKeygenId, signature, kmsTxSender);
```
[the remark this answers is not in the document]
eudelins-zama: Not relevant

kms-connector/tests/contracts/KMSGenerationMock.sol
```diff
@@ -0,0 +1,81 @@
...
+    uint256 prepKeygenCounter = 3 << 248;
+    uint256 keyCounter = 4 << 248;
+    uint256 crsCounter = 5 << 248;
+
+    function keygen(ParamsType paramsType) external {
+        prepKeygenCounter++;
+        uint256 prepKeygenId = prepKeygenCounter;
+
+        emit PrepKeygenRequest(prepKeygenId, paramsType, "");
+    }
```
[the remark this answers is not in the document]
eudelins-zama: Good enough for our current testing setup for now

kms-connector/connector-db/migrations/20260407130028_key_generation_ethereum_migration.sql
```diff
@@ -0,0 +1,20 @@
...
+
+-- Add extra_data column to key generation event tables.
+ALTER TABLE prep_keygen_requests ADD COLUMN extra_data BYTEA;
+UPDATE prep_keygen_requests SET extra_data = NULL;
+
+ALTER TABLE keygen_requests ADD COLUMN extra_data BYTEA;
+UPDATE keygen_requests SET extra_data = NULL;
+
+ALTER TABLE crsgen_requests ADD COLUMN extra_data BYTEA;
+UPDATE crsgen_requests SET extra_data = NULL;
```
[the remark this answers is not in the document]
eudelins-zama: On purpose for forward compatibility with `kms-connector` v0.12

eudelins-zama:
Closed in favor of https://github.com/zama-ai/fhevm/pull/2305 for conflict management

PR #2272, opened 2026-04-09
feat(test-suite): add rollout generator for compatibility matrices

Eikix:
## Summary
Add a deterministic `fhevm-cli rollout` command that generates cumulative mixed-version lock files and `matrix.json` from two pinned lock bundles and an explicit rollout order.
Add a manual compatibility-matrix workflow that resolves `from` and `to`, uploads the generated rollout artifacts once, and fans out reusable e2e jobs per generated lock.
Extend the reusable e2e workflow to select a named lock file from an uploaded artifact and to accept an optional test-profile override.

## Why
This implements the rollout-based compatibility validation described in zama-ai/fhevm-internal#1252 without committing rollout lock files to the repo.
QA can now choose the upgrade order through one dispatch input and get one reproducible e2e result per cumulative rollout step.

## Validation
- `bun run check`
- `bun test src`


Eikix:
Closing this PR and restarting on a fresh branch.

This branch was useful to validate the rollout direction, but it also showed that `v0.11 -> v0.12` is the wrong path to land the feature cleanly.

Main learnings:
- `relayer-sdk` has protocol behavior and cannot be treated as an invisible test dependency
- some intermediate rollout states are invalid and should not appear in the matrix at all
- the test-suite Docker image should be pinned to the target release tag instead of drifting with `main`
- rollout planning needs explicit semantics such as `skip`, `upgrade`, and `deploy-new`
- some dependent components need to move in the same rollout step instead of being split artificially

So the next step is not to keep iterating on this branch. A new branch will be opened with the same goal, but using those learnings from the start:
- pin the test-suite Docker image to the target release tag
- model rollout-controlled components explicitly, including `relayer-sdk`
- remove invalid intermediate states from the matrix
- focus on the supported adjacent upgrade path (`v0.12 -> v0.13` / `main`) instead of forcing `v0.11 -> v0.12` to fit

This PR was useful for discovery, but it is no longer the right vehicle for the implementation.

PR #2274, opened 2026-04-09
feat(coprocessor): add `FHE.sum` native opcode (op 28) end-to-end

PanGan21:
## Summary

- Adds `FHE.sum` as a native coprocessor opcode (op 28). Solidity emits a single `FheSum` event carrying the full input array; the coprocessor executes the sum using tfhe-rs.
- Supported types: `euint8`, `euint16`, `euint32`, `euint64`, `euint128`. 
- Array size limits: 100 elements for Uint8–Uint32, 60 for Uint64–Uint128.

Closes: https://github.com/zama-ai/fhevm-internal/issues/1226


coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -3126,6 +3174,67 @@ pub fn perform_fhe_operation_impl(
             fhe_operation: format!("{:?}", fhe_operation),
             input_types: input_operands.iter().map(|i| i.type_name()).collect(),
         }),
+        SupportedFheOperations::FheSum => match &input_operands[0] {
```
PanGan21, line 3202: This match arm is great for abstracting with a  macro. However I see in the codebase it is not used much often. What do you think about it?

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -699,6 +701,52 @@ pub fn check_fhe_operand_types(
 
                     Ok(())
                 }
+                SupportedFheOperations::FheSum => {
+                    const FHE_SUM_MIN_INPUTS: usize = 2;
+                    // Wide types require more GPU memory per element; matches FHEVMExecutor.sol limits.
```
obatirou: nit: not sure this comment regarding GPU is necessary?

PanGan21: fixed

host-contracts/examples/tests/FHEVMManualTestSuite.sol
```diff
@@ -210,4 +210,23 @@ contract FHEVMManualTestSuite {
...
+        euint8[] memory values = new euint8[](3);
+        values[0] = FHE.fromExternal(a, inputProof);
+        values[1] = FHE.fromExternal(b, inputProof);
+        values[2] = FHE.fromExternal(c, inputProof);
+        euint8 result = FHE.sum(values);
+        FHE.allowThis(result);
+        resEuint8 = result;
+    }
+
+    function test_sum_euint64(externalEuint64 a, externalEuint64 b, bytes calldata inputProof) public {
```
obatirou, line 254: could you add the tests for the other types too?

host-contracts/test/fhevmOperations/manual.ts
```diff
@@ -739,4 +739,36 @@ describe('FHEVM manual operations', function () {
...
+      encryptedAmount.handles[1],
+      encryptedAmount.handles[2],
+      encryptedAmount.inputProof,
+    );
+    await tx.wait();
+    const res = await decrypt8(await this.contract.resEuint8());
+    expect(res).to.equal(60);
+  });
+
+  it('sum euint64 - two elements', async function () {
```
obatirou, line 794: Same here with other types?

host-contracts/examples/tests/FHEVMManualTestSuite.sol
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2274/changes/b550cc913461776af0682116d79898d2f156a0f8

host-contracts/test/fhevmOperations/manual.ts
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2274/changes/b550cc913461776af0682116d79898d2f156a0f8

obatirou:
Were make update-conformance && make prettier run?

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
antoniupop: I have mixed feelings about macros. Some places like here look like a good match, but I find that readability/maintainability can go down with too much macro use. Use your best judgement, but I would be against too much usage if it just saves a few lines of code. I find the code fine as is really...

PanGan21: Thanks for the feedback, I will skip this refactor for now, since this applies to many other cases in this file.
If we choose to refactor to macros we can do it consistently for the whole file.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -650,6 +656,41 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
         emit FheRandBounded(msg.sender, upperBound, randType, seed, result);
     }
 
+    /**
+     * @notice          Computes FHESum operation over an array of ciphertexts of the same type.
+     * @param values    Array of ciphertext handles. All must be the same FheType.
+     * @return result   Result handle of the same FheType as the inputs.
+     */
+    function fheSum(bytes32[] calldata values) public virtual returns (bytes32 result) {
```
obatirou, line 664: as this function is added to host-contracts/contracts/FHEVMExecutor.sol. We will need to upgrade it.
You will need to bump the `REINITIALIZER_VERSION` and add a new `reinitializeVX()` function.

Same for HCULimit.sol

PanGan21: Good catch, this was actually the meaning of my question about the contract upgrade

PanGan21:
> Were make update-conformance && make prettier run?

Done

host-contracts/contracts/FHEVMExecutor.sol
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2274/changes/55bbcd67cfe86bb339d537b035df30977af9f941
Would appreciate to have a look for correctness since it is first time

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
```diff
@@ -1796,6 +1796,28 @@ pub fn get_op_size_on_gpu(
...
+            }
+            let n = input_operands.len() as u64;
+            // No dedicated get_sum_size_on_gpu API exists in tfhe-rs; using N * ciphertext_size
+            // as an approximation.
+            match &input_operands[0] {
+                SupportedFheCiphertexts::FheUint8(v) => Ok(v.get_size_on_gpu() * n),
+                SupportedFheCiphertexts::FheUint16(v) => Ok(v.get_size_on_gpu() * n),
+                SupportedFheCiphertexts::FheUint32(v) => Ok(v.get_size_on_gpu() * n),
+                SupportedFheCiphertexts::FheUint64(v) => Ok(v.get_size_on_gpu() * n),
+                SupportedFheCiphertexts::FheUint128(v) => Ok(v.get_size_on_gpu() * n),
```
rudy-6-4: There is no 160, 256 bits ?

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
```diff
@@ -1796,6 +1796,28 @@ pub fn get_op_size_on_gpu(
                 }),
             }
         }
+        SupportedFheOperations::FheSum => {
+            if input_operands.is_empty() {
```
rudy-6-4: this function is already insanely long. the sum case could be in a separate function.
`SupportedFheOperations::FheSum => bla_fhe_sum(input_operands),
`
This set the bar for the next additions.

coprocessor/proto/common.proto
```diff
@@ -36,4 +36,5 @@ enum FheOperation {
   FHE_RAND = 26;
   FHE_RAND_BOUNDED = 27;
   FHE_GET_CIPHERTEXT = 32;
```
rudy-6-4, line 39: If there are other planned value this could be commented
// FHE_XXX = 33
or
// 33, 34 reserved for other op (not decided yet)

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -699,6 +701,51 @@ pub fn check_fhe_operand_types(
 
                     Ok(())
                 }
+                SupportedFheOperations::FheSum => {
+                    const FHE_SUM_MIN_INPUTS: usize = 2;
+                    const FHE_SUM_MAX_INPUTS_WIDE: usize = 60;
+                    const FHE_SUM_MAX_INPUTS_NARROW: usize = 100;
+
+                    if input_handles.len() < FHE_SUM_MIN_INPUTS {
```
rudy-6-4, line 709: is it really useful ? doesn't the solidity code handle that ? (eg identity or add for 0 and 1)

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -699,6 +701,51 @@ pub fn check_fhe_operand_types(
...
+                            fhe_operation_name: format!("{:?}", fhe_op),
+                            expected_operands: FHE_SUM_MIN_INPUTS,
+                            got_operands: input_handles.len(),
+                        });
+                    }
+
+                    let first_type = get_ct_type(&input_handles[0])?;
+
+                    let fhe_sum_max_inputs = match first_type {
+                        5 | 6 => FHE_SUM_MAX_INPUTS_WIDE, // Uint64 | Uint128
```
rudy-6-4: This should be a "if" ? what about 7 and 8 ?

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -699,6 +701,51 @@ pub fn check_fhe_operand_types(
 
                     Ok(())
                 }
+                SupportedFheOperations::FheSum => {
```
rudy-6-4, line 704: should this reject unsupported type immediatly , like 160 or 256 ?
The type is used for fhe_sum_max_inputs. This would be clearer that it's not a bug that 7 and 8 are not matched.

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -3126,6 +3173,67 @@ pub fn perform_fhe_operation_impl(
             fhe_operation: format!("{:?}", fhe_operation),
             input_types: input_operands.iter().map(|i| i.type_name()).collect(),
         }),
+        SupportedFheOperations::FheSum => match &input_operands[0] {
+            SupportedFheCiphertexts::FheUint8(_) => input_operands
```
rudy-6-4: This could be a function. Will be used by is_in and max etc

coprocessor/proto/common.proto
```diff
@@ -36,4 +36,5 @@ enum FheOperation {
   FHE_RAND = 26;
```
rudy-6-4, line 36: Is this file still relevant ? @Eikix

host-contracts/contracts/HCULimit.sol
```diff
@@ -1475,6 +1475,65 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
+        FheType resultType,
+        bytes32[] calldata values,
+        bytes32 result,
+        address caller
+    ) external virtual {
+        if (msg.sender != fhevmExecutorAddress) revert CallerMustBeFHEVMExecutorContract();
+        uint256 n = values.length;
+        uint256 opHCU;
+        if (resultType == FheType.Uint8) {
+            if (n <= 10) opHCU = 90900;
```
rudy-6-4, line 1495: Ternary could a bit better aligned and easier to eyes parse

```
if (resultType == FheType.Uint8) {
    opHCU = (n <= 10) ? 90900 :
            (n <= 30) ? 127000 :
            (n <= 60) ? 148000 : 159000;
} else if (resultType == FheType.Uint16) {
    opHCU = (n <= 10) ? 95000 :
            (n <= 30) ? 136000 :
            (n <= 60) ? 162000 : 184000;
} else if (resultType == FheType.Uint32) {
    opHCU = (n <= 10) ? 116000 :
            (n <= 30) ? 164000 :
            (n <= 60) ? 205000 : 281000;
} else if (resultType == FheType.Uint64) {
    opHCU = (n <= 10) ? 139000 :
            (n <= 30) ? 216000 : 306000;  // no distinct <=60 tier
} else if (resultType == FheType.Uint128) {
    opHCU = (n <= 10) ? 219000 :
            (n <= 30) ? 355000 : 552000;  // no distinct <=60 tier
} else {
    // revert or handle invalid type as needed
    revert("Invalid FheType");
}

```

host-contracts/lib/Impl.sol
```diff
@@ -705,6 +712,11 @@ library Impl {
         result = IFHEVMExecutor($.CoprocessorAddress).fheRandBounded(upperBound, randType);
     }
 
+    function sum(bytes32[] memory values) internal returns (bytes32 result) {
+        CoprocessorConfig storage $ = getCoprocessorConfig();
+        result = IFHEVMExecutor($.CoprocessorAddress).fheSum(values);
```
rudy-6-4, line 717: here it could degenerate case (size 1 or 2)

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: This just validates inputs independently, just defensive programming to avoid issues - remember the case with div which would be okay if we were doing the same..

coprocessor/proto/common.proto
PanGan21: I think it is because it is use here: https://github.com/zama-ai/fhevm/blob/d517efbd42e4fe7a62a5265d82c5265bc78e7b50/coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs#L15

coprocessor/proto/common.proto
PanGan21: It doesn't seem reserved so I refactored to use opcode 28.

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2274/commits/883956c6544bbd84f083f4d56d6ba5082941bf2b

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2274/commits/882a523efe355a1b6700584a8a3b135c1540ba8c

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2274/commits/4dc1851eb8a321afe316fc1679632e91f7616565. 
Refactored into a reusable function that can be used with is_in as well

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Fixed, types 7 and 8 now hit the `UnsupportedFheTypes` error

host-contracts/contracts/HCULimit.sol
PanGan21: I see this might be better but this file is autogenerated by the solidity library and it doesn't use ternary operators. I am not sure if it is worthy to add an exception for that on it.
Also prettier will format this in a much uglier format. WDYT?

host-contracts/lib/Impl.sol
PanGan21: The check is already happening here: https://github.com/zama-ai/fhevm/blob/c69a30da739301e5dcddcc2d22d53b9e01f5f8a2/host-contracts/contracts/FHEVMExecutor.sol#L665-L680

I think there is no need to add it twice in the function body. It will also duplicate maintenance if we would like to change something

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
PanGan21: Both are supported by tfhe-rs but:
- `FheUint160` maps to eaddress in fhevm — summing Ethereum addresses has no semantic meaning.
- `FheUint256` is excluded for the same reason `FHE.add`doesn't support `euint256` — arithmetic operations in fhevm are defined up to euint128, and FheSum follows the same semantics.

Let me know if you think differently about it

host-contracts/contracts/FHEVMExecutor.sol
obatirou: Seems ok to me 👍

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
rudy-6-4: Ok I though 256 bits were supported for arithmetic (since the operation are present in fhevm 
https://github.com/zama-ai/fhevm/blob/main/coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs#L829
)

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
rudy-6-4: Ok

host-contracts/contracts/HCULimit.sol
rudy-6-4: Ok, let it that way.

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
PanGan21: I don't see an overload in Solidity. Might be good to remove it from coprocessor in that case..

host-contracts/lib/Impl.sol
rudy-6-4: Sorry the comment was unclear, I think we should handle degenerate case 0 (sum([]) = 0), 1 and 2. S
E.g. a contract handling a table that start empty and grows and does a sum on it, it forces a developer to handle size 0 1 and 2 manually but there is no real reason for that.

PanGan21: Hey @obatirou we need a second opinion here.
Would you say it is best to handle empty array or array with one element or throw error?
@rudy-6-4  has a point here that when writing generic code these are valid operations.

```
python
Python 3.13.5 (main, Jun 25 2025, 18:55:22) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> sum([])
0
>>> sum([1])
1
>>> sum([2])
2
>>> exit()
```

Additionally to that in case of 0 do you think we should return trivially encrypted 0?

PanGan21: Thanks for flagging it @rudy-6-4 . Discussed this offline and agreed the following:

**Uninitialized element in the array**: Substitute with trivialEncrypt(0) — consistent with how binary ops handle uninitialized inputs, but leaks metadata
**Array length of 0 or 1**:  Substitute — length 0 substitute with trivialEncrypt(0) and length 1 substitute with trivialEncrypt(val[0])

Let me know what you think. A small drawback to this approach is that we send computation to copro for handles that are already predicted.

PanGan21: Implementation changed:
- Empty array is being handled by the coprocessor now
- Result type has been added to preimage so coprocessor knows the output type for empty array to use for trivialEncrypt(0)
- Uninitialized items in array are still being replaced with trivialEncrypt(0)

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -3126,6 +3193,61 @@ pub fn perform_fhe_operation_impl(
             fhe_operation: format!("{:?}", fhe_operation),
             input_types: input_operands.iter().map(|i| i.type_name()).collect(),
         }),
+        SupportedFheOperations::FheSum => {
+            if input_operands.is_empty() {
+                if !matches!(output_type, 2..=6) {
```
rudy-6-4, line 3198: not sure you should do the check at every level or maybe the check should be in a separate function in that case. eg fhe_sum_check_output_type(...)

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -650,6 +656,45 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
+
+        uint256 maxSize = (resultType == FheType.Uint64 || resultType == FheType.Uint128) ? 60 : 100;
+        if (values.length > maxSize) revert FHECollectionSizeInvalid(values.length, maxSize);
+
+        for (uint256 i = 0; i < values.length; i++) {
+            if (!acl.isAllowed(values[i], msg.sender)) revert ACLNotAllowed(values[i], msg.sender);
+            if (_typeOf(values[i]) != resultType) revert IncompatibleTypes();
+        }
+
+        result = keccak256(
```
rudy-6-4: given that isIn is coming (and max etc), it could be a generic _naryOp function (but can be done in isIn PR)

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -699,6 +701,55 @@ pub fn check_fhe_operand_types(
...
+
+                    if input_handles.is_empty() {
+                        return Ok(());
+                    }
+
+                    let first_type = get_ct_type(&input_handles[0])?;
+
+                    // FheUint160 (7) and FheUint256 (8) are not supported for FheSum.
+                    let fhe_sum_max_inputs = match first_type {
+                        5 | 6 => FHE_SUM_MAX_INPUTS_WIDE,   // Uint64 | Uint128
```
rudy-6-4, line 716: Are this integer available with a name/construct. (Comments only are ok if not)

PanGan21: There are no standalone named constants for these values. I can add some constants but it seems they will be used only here. I can generalize this in isIn if needed

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2274/commits/d006829183060949758177a91d8846b059455373

host-contracts/contracts/FHEVMExecutor.sol
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2274/commits/a711b6e6591fca61bb001c6c6c525091302dbd6b

PanGan21: I kept building the preimage out of the function because it is probably different for each operation

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
rudy-6-4, line 716: Arh sorry I though the fhe_sum_check_output_type could be used here. But i read too quick.

PanGan21: Oh ok, shall i revert this? Doesn't add something tbh

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -672,12 +672,7 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
         uint256 maxSize = (resultType == FheType.Uint64 || resultType == FheType.Uint128) ? 60 : 100;
         if (values.length > maxSize) revert FHECollectionSizeInvalid(values.length, maxSize);
 
-        for (uint256 i = 0; i < values.length; i++) {
-            if (!acl.isAllowed(values[i], msg.sender)) revert ACLNotAllowed(values[i], msg.sender);
-            if (_typeOf(values[i]) != resultType) revert IncompatibleTypes();
-        }
-
-        result = keccak256(
+        bytes32 preimage = keccak256(
```
rudy-6-4, line 675: One of the goal was to share the handle generation between all nary op. (like other ops)

rudy-6-4 (dismissed):

host-contracts/contracts/FHEVMExecutor.sol
PanGan21: Well it will not be the same preimage, because the resultType used for sum is not needed for other other opearations (i.e. isIn).
For sum the resultType is used so corpocessor knows what will be the result, however this is not needed to be known for bolean operations.

Do you think it is better to have common preimage although it is not used always?

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
rudy-6-4: I let you choose. But it could be accept fhe_sum => return Result<()> with Error about bad type of bad size. Or nothing.

PanGan21: I reverted it for now.

PanGan21:
@claude /pr-review

host-contracts/contracts/FHEVMExecutor.sol
PanGan21: Updated in [this commit](https://github.com/zama-ai/fhevm/pull/2274/commits/35750154a2d4dfb0a61144f5ddb6b538083d14aa) to have the same preimage and getting the result type from the handle metadata

PanGan21:
@claude /pr-review

rudy-6-4 (dismissed):

PanGan21:
@mergify queue

PanGan21:
@Mergifyio dequeue

PanGan21:
@mergify queue

rudy-6-4 (dismissed):

PanGan21:
@mergify queue

rudy-6-4 (approved):

PanGan21:
@mergify queue

host-contracts/contracts/HCULimit.sol
```diff
@@ -139,14 +139,14 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
...
-     * @notice Re-initializes the contract from V1.
+     * @notice Re-initializes the contract from V2.
      * @param hcuCapPerBlock New global HCU cap per block.
      * @param maxHCUDepthPerTx Maximum sequential HCU depth per transaction.
      * @param maxHCUPerTx Maximum total HCU per transaction.
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    function reinitializeV2(
+    function reinitializeV3(
```
Eikix, line 149: Forgot to mention but perhaps you don't need the storage var setting anymore

Since for fresh deploys initializeFromEmptyProxy is used (which needs to accumulate storage var setups)
But for migration i dont think it'll be required (let me know what you think)

Ofc it can be done in a follow-up PR or even ignored for now. It's mainly a nit.

PanGan21: Not sure if i fully understand. does this mean the body should be empty or to not bump?

PanGan21: Also what changed and we don't need the storage var anymore?

PanGan21:
@mergify queue

PR #2275, opened 2026-04-09
fix(test-suite): retry GitHub rate limits in resolver

Eikix:
## Summary
- retry GitHub rate-limit responses in the fhevm resolver instead of failing immediately
- increase retry attempts and use capped exponential backoff with light jitter
- extend resolver tests to cover 429 and rate-limit 403 responses

## Verification
- `bun test src/resolve.test.ts`
- `bun run check`

### DIFF
--- a/test-suite/fhevm/src/resolve/github.ts
+++ b/test-suite/fhevm/src/resolve/github.ts
@@ -61,6 +61,18 @@ const retryDelayMs = (attempt: number) => {
   return base + Math.floor(base * GH_API_RETRY_JITTER_RATIO * Math.random());
 };
 
+const retryReason = (message: string) => {
+  const lower = message.toLowerCase();
+  if (lower.includes("secondary rate limit")) return "secondary rate limit";
+  if (lower.includes("rate limit") || /\bhttp 429\b/.test(lower)) return "rate limit";
+  if (/\bhttp 5\d\d\b/.test(lower)) return "GitHub API 5xx";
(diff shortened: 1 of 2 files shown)


Eikix:
@claude /pr-review

test-suite/fhevm/src/resolve/github.ts
```diff
@@ -51,19 +62,77 @@ export const shouldRetryGitHubCliError = (message: string) => {
...
+      .map((line) => {
+        const index = line.indexOf(":");
+        return index < 0 ? undefined : [line.slice(0, index).trim().toLowerCase(), line.slice(index + 1).trim()] as const;
+      })
+      .filter((entry): entry is readonly [string, string] => Boolean(entry)),
+  );
+  return { headers, body: bodyParts.join("\n\n").trim() };
+};
+
+export const rateLimitRetryDelayMs = (headers: Record<string, string>, attempt: number, now = Date.now()) => {
```
PanGan21, line 93: Suggestion from codex:

This function always uses `x-ratelimit-reset` when it exists, but [GitHub’s guidance](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit) for secondary limits says to use that header only when `x-ratelimit-remaining == 0`, otherwise use Retry-After or wait at least one minute. That means a secondary-limit response can now block the resolver until the primary bucket resets, which is often far longer than intended.

PanGan21 (approved):

Eikix:
@mergifyio queue

PR #2278, opened 2026-04-09
fix(test-suite): freeze orchestrated fhevm baseline

Eikix:
## Summary
Freeze orchestrated `fhevm` e2e baselines from the PR base SHA instead of resolving floating `latest-main` during the reusable workflow boot path.

Closes zama-ai/fhevm-internal#1263.

## Why
Merge-queue / orchestrated e2e was still resolving `latest-main` live when no lock artifact was provided. That made the boot path depend on live GitHub and GHCR state and caused flaky failures from:
- package metadata scope/auth issues
- GitHub API / fetch failures and rate limiting
- image sets where `latest-main` points at a commit whose full orchestrated stack is not available yet

The goal of this PR is narrow:
- resolve one baseline lock from `github.event.pull_request.base.sha`
- reuse that frozen lock across dry-run and boot
- keep head-image overrides only for components rebuilt by the PR
- let existing and future SHA/tag-based compat rules reason from the resolved base SHA

## Change
- add a `resolve-baseline` job to `.github/workflows/test-suite-orchestrate-e2e-tests.yml`
- resolve the baseline with `./fhevm-cli resolve --target sha --sha "$BASE_SHA" --reset`
- upload the produced lock as an artifact and thread `lock-artifact-name` into the reusable e2e workflow
- require non-manual orchestrated runs to use that frozen lock instead of falling back to `latest-main`
- keep `workflow_dispatch` orchestrated runs able to resolve `latest-main` manually when no artifact is supplied
- simplify `sha` resolution so it constructs the repo-owned image tags from the requested SHA without branch/ref/release logic

## Testing
- `cd test-suite/fhevm && bun test src/resolve.test.ts src/cli.test.ts`
- `ruby -e 'require "yaml"; YAML.load_file(ARGV.fetch(0)); puts "ok"' .github/workflows/test-suite-orchestrate-e2e-tests.yml`
- `ruby -e 'require "yaml"; YAML.load_file(ARGV.fetch(0)); puts "ok"' .github/workflows/test-suite-e2e-tests.yml`
- `cd test-suite/fhevm && bun run src/cli.ts resolve --target sha --sha $(git rev-parse --short=7 origin/main) --reset`
- PR orchestrated validation run passed before this simplification: https://github.com/zama-ai/fhevm/actions/runs/24889356352


Eikix:
@claude /pr-review

.github/workflows/test-suite-orchestrate-e2e-tests.yml
```diff
@@ -91,99 +127,50 @@ jobs:
...
-          relayer-docker-build|relayer_migrate_build_result|relayer-migrate-version
-          relayer-docker-build|relayer_build_result|relayer-version
-          test-suite-docker-build|build_result|test-suite-version
-          EOF
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v0.8.0
+        with:
+          script: |
+            const headTag = process.env.NEW_COMMIT_HASH.slice(0, 7);
+            const buildResults = JSON.parse(process.env.DOCKER_BUILD_RESULTS);
+            core.setOutput("kms-core-version", "");
```
PanGan21, line 141: Is the empty string correct here? If we want to validate the base branch `CORE_VERSION` should probably be used

Eikix: Yes, this is intentional.

`kms-core-version` is empty here because KMS core is not produced by any repo Docker build in this workflow. The base value is already part of the lock produced by `resolve-baseline`; passing a non-empty workflow input here would override that lock value.

So the flow is:

1. `resolve-baseline` creates the base stack lock, including kms core version (the latest one compatible, it is hardcoded)
2. Repo-owned Docker builds override the lock only when they actually built a PR image.
3. Non-repo companions such as `CORE_VERSION` stay owned by the resolver lock.

One problem that I see is that for backports I'm not sure how it will work.

PanGan21: Do you think we should move forward and resolve backports later?

PanGan21: Let's move forward like this for now and if we see it doesn't work on backports we can pick it up again

PanGan21 (approved):

Eikix:
@mergifyio queue

PR #2293, opened 2026-04-13
fix(relayer): make all config env-var configurable

manoranjith:
## Summary

- Add a generic `deserialize_vec_from_map_or_seq<T>` so every `Vec` config field supports K8s-style indexed env vars (`APP_HOST_CHAINS__0__*`, `APP_METRICS__QUERY_DURATION_HISTOGRAM_BUCKET__0`, …)
- Add `.try_parsing(true)` to the `Environment` source so env-var strings coerce into numeric types (needed for `host_chains[].chain_id: u64`)
- Add a sweep test that overlays an env-only file onto the YAML base, mirroring the real K8s deployment path

## Context

`host_chains` — a new required config field — fails to deserialize when set via indexed env vars with:

```
invalid type: map, expected sequence
```

The `config` crate surfaces indexed env vars as a map, but `Vec<HostChainConfig>` expects a sequence. Since `host_chains[].url` comes from a K8s secret, it cannot be hardcoded in YAML — the deserializer is the only fix.

The existing `listeners` field had a bespoke map-or-seq deserializer; nothing else did. Generalizing it fixes `host_chains` and uncovers (and fixes) the same latent issue in `copro_kms_backoff_intervals` and the four histogram-bucket `Vec<f64>` fields.

Refs zama-ai/fhevm-internal#1191

## Changes

- Rename `deserialize_listeners_from_map_or_seq` → generic `deserialize_vec_from_map_or_seq<T>`, apply it to `host_chains`, `listeners`, `copro_kms_backoff_intervals`, and the four histogram-bucket fields
- Add `.try_parsing(true)` to `Settings::new`'s `Environment` source
- Add three unit tests:
  - `test_deserialize_host_chains_from_indexed_env_vars` — indexed env var map
  - `test_deserialize_host_chains_standard_yaml_still_works` — YAML array regression guard
  - `test_settings_from_yaml_with_env_overrides` — full env-override sweep via `tests/relayer-test-env-only.env`
- `#[serial]` on env-mutating tests to prevent cross-test env var leakage

## Test plan

- [x] `make check` — fmt + clippy + OpenAPI drift (green)
- [x] `make test-unit` — 173 passed, 0 failed (includes the three new tests)
- [x] Deploy to devnet with `APP_HOST_CHAINS__0__*` env vars and verify startup


nboisde (approved):
LGTM !

manoranjith:
@mergify queue

PR #2300, opened 2026-04-14
chore(gateway-contracts): bump ACL,GatewayConfig and UPGRADE_FROM_TAG to `v0.12.1`

PanGan21:
### Summary
- Bump ACL v0.3.0 → v0.4.0 and GatewayConfig v0.5.0 → v0.6.0 to acknowledge bytecode changes that landed after the v0.12.1 tag without version bumps
- Advance UPGRADE_FROM_TAG from v0.11.1 to v0.12.1 in all upgrade CI workflows


gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -226,18 +226,24 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
-    function reinitializeV5(uint256 initialKmsContextId) public virtual reinitializer(REINITIALIZER_VERSION) {
+    function reinitializeV6(uint256 initialKmsContextId) public virtual reinitializer(REINITIALIZER_VERSION) {
         GatewayConfigStorage storage $ = _getGatewayConfigStorage();
 
         if (initialKmsContextId == 0) {
             revert InvalidNullKmsContextId();
         }
 
+        // If the per-context KMS list is already populated, the migration been performed
+        if ($.kmsTxSenderAddressesForContext[initialKmsContextId].length > 0) {
```
obatirou: I think this should be removed.
It means reinitializeV5 was already executed
And reinitializeV6 should do nothing. There is nothing to migrate

.github/workflows/contracts-upgrade-version-check.yml
```diff
@@ -9,7 +9,7 @@ on:
 # This avoids unnecessary reinitializer bumps when multiple PRs modify
 # the same contract between deployments. Keep in sync with *-upgrade-tests.yml.
 env:
-  UPGRADE_FROM_TAG: v0.11.1
+  UPGRADE_FROM_TAG: v0.12.1
```
obatirou, line 12: I am confused: it is not deployed yet. But it will be. And we compare main to this one then right?

gateway-contracts/contracts/GatewayConfig.sol
PanGan21: So the whole body of reinitializeV6 should be removed right?

.github/workflows/contracts-upgrade-version-check.yml
obatirou: Answering my question after reflexion: yes

gateway-contracts/contracts/GatewayConfig.sol
obatirou: `function reinitializeV6() public virtual reinitializer(REINITIALIZER_VERSION) {}`

cc @isaacdecoded for confirmation

.github/workflows/contracts-upgrade-version-check.yml
PanGan21: I used the latest release here. v0.12.0 is also not deployed to prod yet.
What I understand is that we want to replicate the production upgrade so maybe v0.12.0 is better? What do you think?

gateway-contracts/contracts/GatewayConfig.sol
obatirou: > So the whole body of reinitializeV6 should be removed right?

Unless I miss something, yes

.github/workflows/contracts-upgrade-version-check.yml
obatirou: We should use the latest release.
So v0.12.X as, when deploying, the upgrade will be done against the last v0.12.X

obatirou: But we need to way to not forget to update UPGRADE_FROM_TAG as soon as there is a release freeze or a new v0.12.X

gateway-contracts/contracts/GatewayConfig.sol
PanGan21: Fixed

PanGan21:
Closing in favor of https://github.com/zama-ai/fhevm/pull/2338

PR #2305, opened 2026-04-15
feat(kms-connector): keygen on ethereum

eudelins-zama:
## Description of changes

This PR migrates key generation event listening and transaction sending from the Gateway chain to Ethereum (L1), implementing dual-chain routing in the KMS Connector.

**Event Listening (`gw-listener`)**
- `EthereumListener` now polls `KMSGeneration` events (`PrepKeygenRequest`, `KeygenRequest`, `CrsgenRequest`) directly from Ethereum using finalized blocks, with its own cancellation token and error handling.
- `PrssInit` and `KeyReshareSameSet` handling have been removed (see RFC-005 and RFC-013) for more context
- `GatewayListener` is simplified to only handle `Decryption` events — the `MonitoredContract` enum and `JoinSet`-based multi-contract polling are removed.
- Healthcheck now monitors both Gateway and Ethereum RPC providers.

**Transaction Sending (`tx-sender`)**
- The monolithic `TransactionSenderInner` is split into `GatewayTransactionSender` (decryption responses) and `EthereumTransactionSender` (keygen responses), each with its own provider and config.
- `EthereumTransactionSender` waits for finalized confirmations (configurable, default 64) with a receipt timeout before considering a tx confirmed.
- `TransactionSender::forward_response` routes responses to the correct chain based on the response kind.
- New config fields: `ethereum_url`, `ethereum_chain_id`, `ethereum_tx_required_confirmations`, `ethereum_tx_get_receipt_timeout`.

**Event Model Cleanup**
- Removed `PrssInit` and `KeyReshareSameSet` event types, DB queries, gRPC handlers, and test cases throughout the stack.
- Added `extra_data` column to `prep_keygen_requests`, `keygen_requests`, and `crsgen_requests` tables via a new SQL migration.
- Keygen event bindings now come from `fhevm_host_bindings` (Ethereum contracts) instead of `fhevm_gateway_bindings`.

**Testing**
- Added `KMSGenerationMock.sol` and `KMSVerifierMock.sol` test contracts for Ethereum-side integration tests.

Closes https://github.com/zama-ai/fhevm-internal/issues/1229
Closes https://github.com/zama-ai/fhevm-internal/issues/1119
Closes https://github.com/zama-ai/fhevm-internal/issues/1167


kms-connector/crates/utils/src/tests/db/requests.rs
```diff
@@ -362,4 +353,18 @@ impl InsertRequestOptions {
...
+        self.epoch_id = Some(epoch_id);
+        self
+    }
+
+    pub fn build_extra_data(&self) -> Vec<u8> {
+        let context_id = self.context_id.unwrap_or(TESTING_KMS_CONTEXT);
+        let epoch_id = self.epoch_id.unwrap_or(TESTING_KMS_EPOCH);
+        let mut extra_data = vec![EXTRA_DATA_V2_VERSION];
+        extra_data.extend(context_id.to_be_bytes_vec());
+        extra_data.extend(epoch_id.to_be_bytes_vec());
```
dvdplm, line 367: This is test code and I bet the compiler optimizes this, but do you really need to convert the ids to `Vec` before extending? Isn't there a way to get a slice view into the U256's bytes and `extend` from the slice (which copies/clones anyway I think)?

kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -1,102 +1,204 @@
...
-            EXPECTED_EXTRA_DATA_LENGTH
-        ));
-    }
-
-    let version = extra_data[0];
-    if version != EXTRA_DATA_VERSION {
-        return Err(anyhow!(
-            "Unsupported extra_data version: 0x{:02x}, expected 0x{:02x}",
+/// Format (v2, RFC 005):
+/// - Byte 0: version (`0x02`)
```
dvdplm, line 33: What is `0x02`? The default?

kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -1,102 +1,204 @@
...
+                return Err(anyhow!(
+                    "extra_data too short for v1: {} bytes, expected at least {} bytes",
+                    extra_data.len(),
+                    EXTRA_DATA_V1_LENGTH
+                ));
+            }
+
+            let context_id_bytes: [u8; 32] = extra_data[1..33]
+                .try_into()
+                .map_err(|e| anyhow!("Failed to extract context_id from extra_data: {e}"))?;
```
dvdplm, line 57: What is the failure mode here? We just checked that the `len()` is >= EXTRADATA_V1_LENGTH so can this actually happen?

kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -1,102 +1,204 @@
...
+
+            let epoch_id_bytes: [u8; 32] = extra_data[33..65]
+                .try_into()
+                .map_err(|e| anyhow!("Failed to extract epoch_id from extra_data: {e}"))?;
+
```
dvdplm, line 80: Same: can this actually happen after the length check?

dvdplm (approved):
Still LGTM. Left some nitpicks&comments.

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: It's the versioning of the `extraData` field format

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: Nope, should be unreachable, but can't think of a clever way to let the compiler know about it...

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: Same

kms-connector/crates/utils/src/tests/db/requests.rs
eudelins-zama: Probably not indeed.
Will merge it as it though to avoid the long living branch :innocent:

PR #2309, opened 2026-04-15
feat(host-contracts): migrate KMSVerifier to ProtocolConfig

obatirou:
Closes https://github.com/zama-ai/fhevm-internal/issues/1227


obatirou:
@claude /pr-review. Skip rust_bindings

obatirou:
@claude /pr-review. Avoid rust_bindings.

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity ^0.8.24;
 import {IProtocolConfig} from "./interfaces/IProtocolConfig.sol";
 import {IKMSGeneration} from "./interfaces/IKMSGeneration.sol";
 import {KmsNode} from "./shared/Structs.sol";
+import {KMS_CONTEXT_COUNTER_BASE} from "./shared/Constants.sol";
 import {kmsGenerationAdd} from "../addresses/FHEVMHostAddresses.sol";
 import {UUPSUpgradeableEmptyProxy} from "./shared/UUPSUpgradeableEmptyProxy.sol";
 import {ACLOwnable} from "./shared/ACLOwnable.sol";
```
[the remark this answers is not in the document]
obatirou: Not relevant for same context

.github/workflows/host-contracts-upgrade-tests.yml
```diff
@@ -120,6 +120,37 @@ jobs:
           cp -r ../../previous-fhevm/host-contracts/contracts ./previous-contracts
           docker cp fhevm-sc-deploy:/app/addresses ./
 
+      - name: Bootstrap host contracts missing from UPGRADE_FROM_TAG
```
isaacdecoded, line 123: Sorry for circling it, but I still don't understand why all this testing logic around the `UPGRADE_FROM_TAG` is needed.

host-contracts/contracts/shared/Constants.sol
```diff
@@ -2,3 +2,7 @@
 pragma solidity ^0.8.24;
 
 uint8 constant HANDLE_VERSION = 0;
+
+/// @dev Base value for KMS context IDs. Format: [0x07 type tag | 31 counter bytes].
+/// See KMSRequestCounters on Gateway for the shared counter scheme.
```
isaacdecoded: nit: just wondering if this Gateway reference is meaningful.

.github/workflows/host-contracts-upgrade-tests.yml
obatirou: this CI step is testing the upgrade of KMSVerifier. But any version before the deployement of ProtocolConfig and KMSGeneration on ethreum are not part of previous-contracts/ so they are not compiled.
Though there is a KMSVerifier dependancy to ProtocolConfig due to the on-chain check introduced by `_requireProtocolConfigReady`

So `ProtocolConfig` must be compiled and deployed "manually".

obatirou: We could argue to keep `_requireProtocolConfigReady` or not

host-contracts/contracts/shared/Constants.sol
obatirou: indeed will remove it

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -284,17 +186,7 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
...
-    }
-
-    /**
-     * @notice              Get the threshold for signature.
-     * @return threshold    Threshold for signature verification.
-     */
-    function getThreshold() public view virtual returns (uint256) {
-        KMSVerifierStorage storage $ = _getKMSVerifierStorage();
-        return $.contextThreshold[$.currentKmsContextId];
+        return PROTOCOL_CONFIG.getKmsSigners();
```
isaacdecoded, line 183: Also wondering if it's worth introducing the `ProtocolConfig.getKmsSigners()` when we already have `ProtocolConfig.getKmsSignersForContext()`. Meaning that if we don't, here we must do an extra call to `ProtocolConfig.getCurrentKmsContextId()`.

But more precisely, my question is: is getKmsSigners() needed by any other component, or does it have real usage value? Maybe we'd like to enforce the contextId specification instead of providing methods that shadow the current context ID. Same for `ProtocolConfig.isKmsSigner()`

Hope I'm explaining myself.

obatirou: As my understanding is the protocol should always use the current context id, it makes sense to use special methods to avoid increase gas consumption (Maybe a rename getKmsSigners -> getKmsSignersForCurrentContext and related function would be in order)
Query on old context would be rare. Let me know if I have a wrong assumption here.

Do we want to enforce the contextId specification and avoid having methods for currentContext?

host-contracts/tasks/upgradeContracts.ts
isaacdecoded, line 1: nit: these refactors seem not to be in the scope of the intended RFC-013 changes. But mostly, it would be good to continue keeping these tasks aligned with the ones at the gateway-contracts, since both usually follow the same patterns/logics. Meaning that we would like to replicate the changes made here also in the gateway-contracts/tasks/upgradeContracts.ts

host-contracts/test/migration-only-previous-contracts/KMSVerifier.sol
isaacdecoded, line 1: Do we really need this previous KMSVerifier contract? What is the purpose of its deployment in the tests below?

obatirou: It is needed as fixture to test the pre-flight `assertLegacyVerifierMatchesEnv` which verifies ENV variable match legacy KMSVerifier on chain values before migration. Which is very important as we do not want any miss configuration during the migration as `initializeFromMigration` accept blindly user inputs.

Maybe there is a way to simplify / refactor. Open to ideas

host-contracts/contracts/shared/Constants.sol
obatirou: [92a0c8b](https://github.com/zama-ai/fhevm/pull/2309/commits/92a0c8bac709ba576db419acf87d5bff5f154cc8)

host-contracts/tasks/upgradeContracts.ts
obatirou: Your're right
Removed with https://github.com/zama-ai/fhevm/commit/a86acf6120531403eb871f527688f969891b33f2

.github/workflows/host-contracts-upgrade-tests.yml
isaacdecoded: It's clear to me now, thanks. Resolving it.

host-contracts/contracts/KMSVerifier.sol
isaacdecoded: > As my understanding is the protocol should always use the current context id, it makes sense to use special methods to avoid increase gas consumption

This answers what I was wondering about, thanks.

host-contracts/test/migration-only-previous-contracts/KMSVerifier.sol
isaacdecoded: I see 🤔

> Maybe there is a way to simplify / refactor. Open to ideas

Maybe we can minimize it to the needed [LEGACY_VERIFIER_ABI](https://github.com/zama-ai/fhevm/pull/2309/changes/BASE..1a17d59edb49114d20195ab3b03028d1a3d6dd6d#diff-fadb1da363b80f10b637906ffd02a97b03e3b9898cbc0f911fb90e4f70e3afcbR537-R541) methods only?

isaacdecoded (approved):
LGTM! 👍

host-contracts/test/migration-only-previous-contracts/KMSVerifier.sol
obatirou: simplified
Also reworked tests to simplify 
Will improve again when doing the issue on scripts anyway

ci/abi-compat/exceptions.ts
```diff
@@ -3,7 +3,24 @@ import type { PackageName } from "./config";
...
   "host-contracts": {
     ACL: ["error ExpirationDateBeforeOneHour()"],
-    KMSVerifier: ["event NewContextSet(address[],uint256)"],
+    // Note: initializeFromEmptyProxy signature change (4-param → 2-param) is excluded
+    // by the /^initialize/ regex in config.ts and does not need an explicit exception.
+    KMSVerifier: [
+      "event NewContextSet(address[],uint256)",
+      "function defineNewContext(address[],uint256)",
+      "function destroyKmsContext(uint256)",
+      "function getThreshold() returns (uint256)",
```
obatirou: @isaacdecoded is it ok to remove the getThreshold?
Won't this break anything regarding SDK or other part of the stack?
I did not find any call to it from what I looked but want to be sure

obatirou: because @Eikix spotted in the rollout, contract are first deployed and the stack need to continue working

isaacdecoded: Took another look [at the relayer-sdk](https://github.com/zama-ai/relayer-sdk/blob/main/src/sdk/kms/KmsSignersVerifier.ts#L89), and we need to keep this method indeed.

isaacdecoded (changes requested):
We need to keep the getThreshold method as pointed out in [this discussion](https://github.com/zama-ai/fhevm/pull/2309#discussion_r3108796847).

ci/abi-compat/exceptions.ts
obatirou: Thanks!
Will make the change and ping you back

obatirou: Done with [9ab429a](https://github.com/zama-ai/fhevm/pull/2309/commits/9ab429adca5a7ebca5be33e6313dc2ed93eba8f7)

isaacdecoded (approved):

PR #2310, opened 2026-04-15
fix(test-suite): gate host-listener consumer compat

Eikix:
## What
Gate the new host-listener consumer topology behind the host-listener bundle version so legacy supported bundles stay on the old path.

## Why
`latest-supported` is still pinned to `v0.11.0`, so the new `listener-core` step and `coprocessor-host-listener-consumer` service must not be required there. Without that gate, `compat-smoke` and any legacy boot path fail because the old image does not provide the new runtime contract.

## How
- add a single compat helper for the `v0.13.0` cutover, including prereleases on that line
- skip `listener-core` and the consumer service on legacy bundles in readiness, repair, service selection, and compat smoke
- point the local consumer build back at the existing `host-listener` Docker stage so `--build` remains valid
- add compat and repair coverage for the legacy path


coprocessor/fhevm-engine/host-listener/src/bin/consumer.rs
```diff
@@ -0,0 +1,132 @@
...
+    health_port: u16,
+
+    #[arg(
+        long,
+        value_parser = clap::value_parser!(Level),
+        default_value_t = Level::INFO
+    )]
+    log_level: Level,
+
+    #[arg(long, default_value = "host-listener-poller")]
```
obatirou, line 58: need to be changed

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,262 @@
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use std::time::Duration;
+
+use alloy::primitives::Address;
+use alloy::rpc::types::Log;
+use alloy_primitives::LogData;
```
obatirou, line 7: use alloy_primitives::LogData; and use alloy::primitives::Address; can be factorized

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+        Err(err) => error!(error = %err, "Consumer task panicked"),
+    }
+    client.cancel();
+    client.unregister_contracts(&contracts).await.unwrap();
+    health_check_cancel_token.cancel();
+    Ok(())
+}
+
+#[allow(clippy::too_many_arguments)]
+async fn ingest_with_retry(
```
obatirou, line 229: Isn't exact copy of poller method?

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+                config.database_retry_interval,
+                ingest_options,
+            )
+            .await
+            {
+                Ok(retries) => {
+                    db_errors.fetch_add(retries, Ordering::Relaxed);
+                    processed_blocks.fetch_add(1, Ordering::Relaxed);
+                    db.tick.update();
+                    inc_blocks_processed(&chain_id_str, 1);
```
obatirou, line 185: processed_blocks is misleading as it increased in success and failure but the metric only on success 
Should be `attempted_blocks` or `blocks_attempted`?

rudy-6-4:
Integrated in https://github.com/zama-ai/fhevm/pull/2302.
Comment will be solved there.

coprocessor/fhevm-engine/host-listener/src/bin/consumer.rs
rudy-6-4: done

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
rudy-6-4: removed, replaced by metrics later.

PR #2311, opened 2026-04-16
feat(host-contracts): rename abortKeygen parameter to prepKeygenId

obatirou:
Closes https://github.com/zama-ai/fhevm-internal/issues/1278


isaacdecoded (approved):
LGTM as long as the CI workflow passes 👍

.github/workflows/host-contracts-upgrade-tests.yml
```diff
@@ -120,6 +120,37 @@ jobs:
           cp -r ../../previous-fhevm/host-contracts/contracts ./previous-contracts
           docker cp fhevm-sc-deploy:/app/addresses ./
 
+      - name: Bootstrap host contracts missing from UPGRADE_FROM_TAG
```
isaacdecoded, line 123: I got a bit lost here: why do we need this bootstrap in this PR? I mean, [the KMSVerifier is not implementing the ProtocolConfig](https://github.com/zama-ai/fhevm/blob/rename-abortKeygen-parameter/host-contracts/contracts/KMSVerifier.sol) on this branch, no?

obatirou: Yes but in this case `KMSGeneration` needs `ProtocolConfig`. So it needs to be compiled.
Note : This new upgrade test with UPGRADE_FROM_TAG was added in the rebase from main. That is why this problems surface now.

cc @Eikix

isaacdecoded: To me it looks really weird since in the `feat/RFC013`branch there's no ProtocolConfig and KMSGeneration upgrading logic/tasks, so I don't understand why they are involved in this CI workflow. Also, it didn't happen before that we needed to execute specific bootstrapping on our contracts deployment, even when at the gateway-contracts we have several intentional circular dependencies, e.g., the GatewayConfig contract points to the Decryption contract, and the Decryption contract also points to the GatewayConfig 🤔

I'll be missing something for sure, but I'd like to know that we fully understand the impact of all this logic that is being introduced around the `host/gateway-contracts-upgrade-tests.yml` workflows, and that it's the best approach for the sake of maintainability. Is this documented somewhere? So I can read and try to find my answers there.

obatirou: You’re right that the upgrade path is still only `ACL`, `FHEVMExecutor`, `KMSVerifier`, `InputVerifier`, and `HCULimit`, and no `ProtocolConfig` or  `KMSGeneration` on `feat/RFC013`.

The bootstrap is needed earlier, at compile time. The workflow reuses the `addresses/` directory from `v0.11.1`, but the current tree already contains `ProtocolConfig.sol` and `KMSGeneration.sol`, which import `kmsGenerationAdd` / `protocolConfigAdd` from the generated `addresses/FHEVMHostAddresses.sol`. Since `v0.11.1` predates both contracts, those symbols are missing there, so the later `compile:specific --contract contracts` used by the upgrade tasks would fail before we even reach the actual upgrade path. 

That’s also why this differs from the gateway circular-deps case: there, both sides already existed at `UPGRADE_FROM_TAG`; here, the current code references address constants that the baseline tag never emitted. The bootstrap is meant to stay minimal and idempotent: `task:ensureMigrationProxyAddresses` only seeds the missing empty proxies / address constants, and `task:deployProtocolConfig` / `task:deployKMSGeneration` only run when the baseline predates those contracts. Once `UPGRADE_FROM_TAG` includes the feature branch , this should become a no-op and can be removed.

This is not obvious from the workflow today. Will try to find a simplification

obatirou: cc @Eikix

obatirou: Made 2 commits with simplification and explanation.
It is necessary to keep those for https://github.com/zama-ai/fhevm/pull/2309.
Let me know if not clear we can huddle to discuss

Eikix: I think it's needed and light. it's a by-product of having inlined addresses in the bytecode..
it becomes a no-op once v0.13 is the "UPGRADE_FROM_TAG"

i'd recommend to add a comment to delete this compat shim once we reach the no-op stage
even 'echo' in the CI -> compat shim became a no-op, recommend to delete it or something

isaacdecoded: Thanks both for the explanations. Resolving it.

PR #2317, opened 2026-04-16
feat(relayer): context aware user decrypt threshold

manoranjith:
## Summary

  Draft implementation of context aware user decrypt threshold feature.

  Refs zama-ai/fhevm-internal#1260

  ## Done

  - When a share is received from the gateway, the `context_id` is extracted from `extra_data` and the corresponding user decrypt shares threshold is
  fetched from the `ProtocolConfig` contract on the Ethereum chain.
  - The fetched value is cached in-memory (moka). On restarts, values are re-fetched from the `ProtocolConfig` contract on the next share arrival.
  - For requests with `extra_data = 0x00` (or empty), the threshold value from config is used, ensuring backward compatibility for older requests during
  the migration period.
  - The resolved threshold is stored in the DB when marking a request as completed, so that the HTTP handler knows how many shares to fetch for the final
  result.
  - To maintain backward compatibility on the read side: `COALESCE(resolved_threshold, static_config)` is used for the share LIMIT, so pre-migration rows
  (where `resolved_threshold` is NULL) fall back to the static config value.
  - DB migration adds nullable `resolved_threshold BIGINT` column (backward-compatible with old software).
  - New config: `protocol_config`.

  ## Tests

  **Unit tests** — extra_data parsing (`extra_data.rs`, `threshold_resolver.rs`):
  1. Empty or `0x00` extra_data returns default context (zero)
  2. Valid v1 and v2 formatted extra_data correctly extracts context_id
  3. Truncated or unknown version extra_data returns an error
  4. Large context_id values (e.g., KMS context counter base) parse correctly

  **Mock-based tests** — threshold resolution (`threshold_resolver_test.rs`, using `ethereum_rpc_mock`):
  1. Default context (ID 0) returns the static config threshold without making any RPC call
  2. Non-zero context ID fetches the threshold from the ProtocolConfig contract mock
  3. A previously fetched threshold is served from cache without a second RPC call
  4. When the first RPC attempt fails, the retry logic recovers on the next attempt
  5. When all retry attempts are exhausted, an error is returned to the caller
  6. A failed fetch is not cached, so the next request for the same context retries from scratch

  ## Pending (should not block testing)

  - [ ] Eliminate silent type conversions: use U256-compatible type for threshold throughout, no range assumptions
  - [ ] Validate context_id on incoming HTTP request to reject invalid ones early
  - [ ] Pin `fhevm_host_bindings` to git tag before merge
  - [ ] ProtocolConfig addresses in mainnet/testnet config examples


relayer/relayer-migrate/migrations/20260416120000_add_resolved_threshold.sql
```diff
@@ -0,0 +1,5 @@
+-- Add resolved_threshold column to user_decrypt_req.
+-- Stores the dynamic threshold used when the write side marked the request as completed.
+-- Nullable for backward compatibility: old software ignores this column, new rows get the value.
+-- NULL means "use static config default" (for rows created before this migration).
+ALTER TABLE user_decrypt_req ADD COLUMN resolved_threshold BIGINT;
```
isaacdecoded, line 5: nit: maybe to enforce the NOT NULL constraint rule, we can set the current "static config default" instead of leaving it as NULL

relayer/src/config/settings.rs
```diff
@@ -538,7 +541,7 @@ pub struct ContractConfig {
     pub decryption_address: String,
     pub input_verification_address: String,
     /// Number of shares required for user decryption threshold consensus
-    pub user_decrypt_shares_threshold: u16,
+    pub user_decrypt_shares_threshold: u32,
```
isaacdecoded, line 544: nit: just wondering why not leave it as `u16`, or even lower like `u8`?

relayer/src/host/threshold_resolver.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// Cached thresholds keyed by context ID. The on-chain uint256 is narrowed
+    /// to u32 at the contract fetch boundary with an explicit range check.
+    /// The repository layer handles u32 ↔ i64 conversion for DB storage (BIGINT).
+    context_thresholds: Cache<U256, u32>,
+    retry_config: RetrySettings,
+}
+
+impl ThresholdResolver {
+    pub async fn new(
+        config: &ProtocolConfigSettings,
```
isaacdecoded: nit: naming this argument as `protocol_config_settings` could help to reduce a bit of the developer's mental load

relayer/src/host/threshold_resolver.rs
```diff
@@ -0,0 +1,232 @@
...
+///
+/// Uses moka over DashMap/HashMap+Mutex because:
+/// 1. `entry().or_try_insert_with()` coalesces concurrent fetches for the same
+///    key — only one RPC call is made, others wait for its result.
+/// 2. Built-in max capacity bound with eviction.
+///
+/// Thresholds are cached permanently (no TTL) since they don't change after
+/// context creation. Context ID 0 is pre-seeded with the static config default.
+pub struct ThresholdResolver {
+    contract: HostProtocolConfig,
```
isaacdecoded:
```suggestion
    protocol_config_contract: HostProtocolConfig,
```

relayer/src/host/threshold_resolver.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// Cached thresholds keyed by context ID. The on-chain uint256 is narrowed
+    /// to u32 at the contract fetch boundary with an explicit range check.
+    /// The repository layer handles u32 ↔ i64 conversion for DB storage (BIGINT).
+    context_thresholds: Cache<U256, u32>,
```
isaacdecoded:
```suggestion
    context_thresholds_cache: Cache<U256, u32>,
```

relayer/src/host/threshold_resolver.rs
```diff
@@ -0,0 +1,232 @@
...
+            ProviderBuilder::new()
+                .network::<alloy::network::AnyNetwork>()
+                .connect_http(url),
+        );
+
+        let contract = IProtocolConfig::new(address, provider);
+
+        let context_thresholds = Cache::builder().max_capacity(max_capacity).build();
+
+        // Pre-seed default: context_id 0 is invalid and maps to the static config value
```
isaacdecoded: If context_id 0 is an invariant violation, why intentionally introduce it in the cache? Also, this could be error-prone somehow since we shouldn't have such a default threshold but only what our source of truth dictates, no?

test-suite/fhevm/config/relayer/local.yaml
```diff
@@ -3,6 +3,13 @@ host_chains:
     url: "http://host-node:8545"
     acl_address: "0x05fD9B5EFE0a996095f42Ed7e77c390810CF660c"
 
+protocol_config:
+  ethereum_http_rpc_url: "http://host-node:8545"
+  address: "0x1234567890123456789012345678901234567890"
```
isaacdecoded:
```suggestion
  address: "0x52054F36036811ca418be59e41Fc6DD1b9e4F4c8"
```

test-suite/fhevm/templates/config/relayer.yaml
```diff
@@ -3,6 +3,13 @@ host_chains:
     url: "http://host-node:8545"
     acl_address: "0x05fD9B5EFE0a996095f42Ed7e77c390810CF660c"
 
+protocol_config:
+  ethereum_http_rpc_url: "http://host-node:8545"
+  address: "0x1234567890123456789012345678901234567890"
```
isaacdecoded:
```suggestion
  address: "0x52054F36036811ca418be59e41Fc6DD1b9e4F4c8"
```

isaacdecoded (commented):
A few nit changes/suggestions here and there.

relayer/src/host/threshold_resolver.rs
manoranjith: Context ID being 0, maps to our existing default context ID, no ? Which will have extra_data `0x00` ?

relayer/src/config/settings.rs
manoranjith: Previously, the threshold was a static config constant. u16 or u8 would have been enough.

Now it varies per request — resolved from ProtocolConfig based on extra_data, stored in the DB so the
GET path knows how many shares to fetch.

u32 is a convenient choice here. It maps to PostgreSQL's smallest numeric type (INTEGER), covers the
range, and keeps application code unsigned.

The on-chain uint256 narrows to u32 once at the contract boundary with a range check. After that it
flows through without conversion.

At the DB boundary, the repository handles u32 ↔ i64 — safe widening on write, checked on read.

relayer/src/host/threshold_resolver.rs
manoranjith: Fixed in 4599dd213

relayer/src/host/threshold_resolver.rs
manoranjith: Fixed in 4599dd213

relayer/src/host/threshold_resolver.rs
manoranjith: Fixed in 4599dd213

test-suite/fhevm/config/relayer/local.yaml
manoranjith: Fixed in 4599dd213

test-suite/fhevm/templates/config/relayer.yaml
manoranjith: Fixed in 4599dd213

relayer/relayer-migrate/migrations/20260416120000_add_resolved_threshold.sql
manoranjith: Yes, we could do it like that as well. Since its a nit, for now, I leave this one like this.

relayer/src/host/threshold_resolver.rs
isaacdecoded: From v0.12, our current context ID is `700000000000000000000000000000000000000000000000000000000000001`, which is the very first ID for KMS contexts.

For example, in the testnet you can check the [getCurrentKmsContextId method](https://sepolia.etherscan.io/address/0xbE0E383937d564D7FF0BC3b46c51f0bF8d5C311A#readProxyContract#F6) that returns `3166189940082864718613269121331309980362851143201109172953918312716374638593` (this is the decimal representation of the value above `70...01`)

manoranjith: From `0.12`, indeed we would have atleast `V1` Context IDs. But because of backward compatibility (which is still being discussed), we need to support a few (atleast 1) previous version of `relayer-sdk`.

Since `v0.5.x` will be the first version of relayer-sdk supporting context IDs and `v0.4.x` will still send a `0x00` as `extra_data`, I think we should still have support for version 0.

Is KMS Connector and Gateway as well aligned with this ? 

I guess, we should even have an e2e test case, where `0x00` is sent as `extra_data`. wdyt ?

isaacdecoded: Ah yes, you're right! So, good to me, thanks!

> Is KMS Connector and Gateway as well aligned with this ?

Yes, they're also backward compatible regarding the `0x00` extra data

> I guess, we should even have an e2e test case, where 0x00 is sent as extra_data. wdyt ?

Makes sense, like installing in the test-suite older versions of the relayer-sdk and running at least the decryption use cases.

isaacdecoded (approved):
LGTM! :clap: Although, there's a CI still failing.

relayer/src/host/threshold_resolver.rs
manoranjith: Thanks for the clarification !

manoranjith:
Updated version of `rustls-webpki` is already merged to main in [PR 2374](https://github.com/zama-ai/fhevm/pull/2374/changes#diff-6b020ac22bdbd912c00b2945675957218692e8995f5f9d378fea076a18572443). So, the CI should fix if we rebase feature branch to main 

We could either

1. Rebase feature branch to main, rebase this branch on updated feature branch and then CI for this PR should pass.
2. Merge this into feature branch as-is and then rebase feature branch to main (the same CI check should pass then). Then merge feature branch to main

PR #1268, opened 2025-11-05
chore: update `compute_handle` to include the domain separators

ppoliani:
Add domain separators when computing the handles during encryption

PR #2318, opened 2026-04-17
feat(host-contracts): split canonical-host deploy task

Eikix:
## Summary

Updates host-contract deployment for the multi-chain host model:

- `task:deployAllHostContracts` now requires an explicit `--with-kms-generation true|false` flag.
- Use `--with-kms-generation true` on the canonical host chain, where `KMSGeneration` is deployed.
- Use `--with-kms-generation false` on non-canonical host chains, where `KMSGeneration` must not be deployed.
- `task:deployLegacyAllHostContracts` is kept as a temporary compatibility alias for canonical-host deployments during the rollout.

Also removes the on-chain `hasPendingKeyManagementRequest` guard from `ProtocolConfig.defineNewKmsContext`: it was unreliable against DAO-proposal execution and made `defineNewKmsContext` unusable on non-canonical host chains where no `KMSGeneration` exists. Coordination becomes a runbook concern, documented in `host-contracts/README.md`.

Closes zama-ai/fhevm-internal#1268.

## Note on issue scope vs. implementation

Issue #1268's "What" table references `test-suite/fhevm/scripts/deploy-fhevm-stack.sh` and `test-suite/fhevm/env/staging/.env.host-sc`. Those paths predate the fhevm-cli (PR #2134, `ef080b9dc`) and no longer exist. The orchestration edits land in the CLI equivalents: `src/generate/compose.ts`, `src/flow/up-flow.ts`, `src/flow/artifacts.ts`, and `src/render-compose.test.ts`.

## Rollout notes

- Existing direct callers must choose the host deployment role explicitly:
  - `npx hardhat task:deployAllHostContracts --with-kms-generation true`
  - `npx hardhat task:deployAllHostContracts --with-kms-generation false`
- The test-suite compose template defaults to canonical-host deployment, and generated extra host-chain services set `HOST_SC_WITH_KMS_GENERATION=false`.
- `library-solidity` mirrors the flag-based task behavior so downstream task callers get the same interface.
- `charts/contracts/values-kmsgen.yaml` is kept as reference config for host-side canonical KMSGeneration trigger jobs; the contracts chart does not render those top-level keys directly.

## Binding regen commit includes tooling drift

The bindings-regen commit (`ef43076d4`) touches ~40 rust binding files beyond the 4 KMS-related ones — extra `#[allow(dead_code)]` attributes from an alloy-sol-macro version difference. Pre-existing drift, not caused by this PR. Clippy clean.

## Test plan

- [x] `forge test` on `host-contracts` — 445/445 passing.
- [x] `npx hardhat test` on `host-contracts` — 2518 passing.
- [x] `make check-bindings` / `make check-selectors` — clean locally.
- [x] `bun test src/render-compose.test.ts` on `test-suite/fhevm` — 11/11 passing.
- [x] `bun test src/flow.test.ts src/render-compose.test.ts` on `test-suite/fhevm` — 28/28 passing.
- [x] Single-host `fhevm-cli up` deploys host contracts including `KMSGeneration`.
- [x] Multi-chain `fhevm-cli up --scenario multi-chain` deploys `KMSGeneration` only on the default host; chain-b's addresses file lacks `KMS_GENERATION_CONTRACT_ADDRESS`.

## Coordination

- Watch PR #2309 (`migrate-KMSVerifier-to-ProtocolConfig`) — it also edits `taskDeploy.ts`. Whichever merges second resolves the conflict.

## Follow-ups (out of scope)

- gitops PR: update per-chain `deployCommands` values for canonical and non-canonical host chains.
- Decouple `host-contracts` address references from compile-time `constant`s (Isaac's smell on PR #2311).
- Phase-1 e2e regression exercising manual mirror of `defineNewKmsContext` on non-canonical — meaningful only after PR #2309 merges.


host-contracts/test/protocolConfig/protocolConfig.t.sol
```diff
@@ -338,6 +329,20 @@ contract ProtocolConfigTest is HostContractsDeployerTestUtils {
         assertEq(protocolConfig.getKmsNodesForContext(newContextId).length, 1);
     }
 
+    function test_defineNewKmsContextSucceedsWithoutKMSGeneration() public {
```
obatirou: No need for this one imo

host-contracts/README.md
```diff
@@ -148,3 +148,36 @@ MIGRATION_KEY_COUNTER="…" \
 # … (all variables above) …
 npx hardhat task:deployKMSGenerationFromMigration --network <network>
 ```
+
+## KMS Context Rotation Runbook (Canonical Host)
+
+The on-chain `hasPendingKeyManagementRequest` guard in `ProtocolConfig.defineNewKmsContext`
```
obatirou: no need to mention hasPendingKeyManagementRequest. Only that there is no onchain guard

host-contracts/tasks/taskDeploy.ts
```diff
@@ -40,28 +40,46 @@ function readExistingHostEnv(): Record<string, string> {
...
   await hre.run('task:deployKMSVerifier');
   await hre.run('task:deployInputVerifier');
   await hre.run('task:deployHCULimit');
 
+  console.log('Common host contracts deployment done!');
+});
+
+task('task:deployCanonicalHostContracts').setAction(async function (_, hre) {
+  await hre.run('task:deployCanonicalEmptyUUPSProxies');
+  // KMSGeneration reads its own address from the generated FHEVMHostAddresses.sol, so it must be
```
obatirou: `KMSGeneration reads its own address from the generated FHEVMHostAddresses.sol` not sure about that
It imports only protocolConfigAdd?

host-contracts/tasks/taskDeploy.ts
```diff
@@ -40,28 +40,46 @@ function readExistingHostEnv(): Record<string, string> {
...
+  // recompiled after deployCanonicalEmptyUUPSProxies wrote kmsGenerationAdd.
+  await hre.run('compile:specific', { contract: 'contracts/KMSGeneration.sol' });
+
+  await hre.run('task:deployKMSGeneration');
+
+  console.log('Canonical host contracts deployment done!');
+});
+
+// Alias: keeps existing call sites working. Runs both as if the one chain were canonical.
+task('task:deployAllHostContracts').setAction(async function (_, hre) {
```
obatirou, line 99: should this be kept as the default?
The canonical is the exception no?

Eikix: i dont think we should change the behaviour no? for backward compat
we just need to make sure infra doesnt use it anymore?

host-contracts/tasks/taskDeploy.ts
```diff
@@ -378,6 +416,77 @@ task('task:assertProtocolConfigReady').setAction(async function (_, hre) {
   }
 });
 
+// Off-chain pre-flight replacing the removed on-chain hasPendingKeyManagementRequest guard.
+// A key/CRS request is pending iff more ?*Request events have fired than paired Activate/Abort events.
```
obatirou: `?*Request` what's this?

host-contracts/tasks/taskDeploy.ts
obatirou: ok make sense

host-contracts/tasks/taskDeploy.ts
```diff
@@ -378,6 +416,77 @@ task('task:assertProtocolConfigReady').setAction(async function (_, hre) {
...
+    const kmsGenAddress =
+      taskArguments.address ?? process.env.KMS_GENERATION_CONTRACT_ADDRESS ?? readExistingHostEnv().KMS_GENERATION_CONTRACT_ADDRESS;
+    if (!kmsGenAddress) {
+      throw new Error(
+        'KMSGeneration address not resolved. Pass --address 0x…, or set KMS_GENERATION_CONTRACT_ADDRESS, or generate addresses/.env.host via a deploy task first.',
+      );
+    }
+
+    const code = await hre.ethers.provider.getCode(kmsGenAddress);
+    if (code === '0x' || code.length === 2) {
```
obatirou: not sure about this condition?
Why have code.length === 2?

host-contracts/tasks/taskDeploy.ts
```diff
@@ -378,6 +416,77 @@ task('task:assertProtocolConfigReady').setAction(async function (_, hre) {
...
+    if (!version.startsWith('KMSGeneration')) {
+      throw new Error(
+        `Contract at ${kmsGenAddress} reports version "${version}"; expected "KMSGeneration v…". Re-check --address.`,
+      );
+    }
+
+    const [prepKeygen, activateKey, abortKeygen] = await Promise.all([
+      kmsGen.queryFilter(kmsGen.filters.PrepKeygenRequest()),
+      kmsGen.queryFilter(kmsGen.filters.ActivateKey()),
+      kmsGen.queryFilter(kmsGen.filters.AbortKeygen()),
```
obatirou: not sure this possible and not sure events scanning is the way to go: this do not have a fromBlock and RPC limits to 10K blocks eth_getLogs if not mistaken. So reliability is not perfect
Storage read would be more convenient but would require view functions to be added:
* getKeyCounter() / getCrsCounter() to get the lastest requests
* isRequestDone(uint256) to see if they are done or not.

cc @isaacdecoded

obatirou: Maybe isRequestDone is not even mandatory and it is possible to use getKeyMaterials(keyCounter) / getCrsMaterials(crsCounter) with reverts but would reduce lisibility and comprehension for a discutable reduction in getters

isaacdecoded: > Storage read would be more convenient but would require view functions to be added

Adding view functions to facilitate debugging and contract state checks has been done for the Gateway contracts in the past, so I think it's also valid to consider it here.

Eikix:
Requesting changes.

The canonical-host split is the right direction, but the PR should make the `KMSGeneration` deployment choice explicit at the call site. Keeping no-argument deployment paths that silently deploy `KMSGeneration` leaves the unsafe behavior in place.

1. `task:deployAllHostContracts` should require an explicit `--with-kms-generation` flag.

Deploying `KMSGeneration` is now a topology decision: it must happen only for the canonical host chain. Please make the main deploy task require one of:

- `npx hardhat task:deployAllHostContracts --with-kms-generation true`
- `npx hardhat task:deployAllHostContracts --with-kms-generation false`

There should be no default. If the flag is omitted, fail with actionable guidance, for example:

```text
Missing required --with-kms-generation flag.

KMSGeneration is deployed only on the canonical host chain. Use:
  --with-kms-generation true   when deploying the canonical host chain
  --with-kms-generation false  when deploying a non-canonical host chain

The no-flag behavior was removed in v0.13 to avoid accidentally deploying KMSGeneration on every host chain.
```

For one-release compatibility, add a separate temporary alias such as `task:deployLegacyAllHostContracts` that keeps the previous behavior and prints a deprecation warning:

```text
task:deployLegacyAllHostContracts is deprecated and will be removed after the v0.13 rollout.
It deploys KMSGeneration and is valid only for canonical-host deployments.
Use task:deployAllHostContracts --with-kms-generation true instead.
```

Helper tasks may still exist internally, but new deployment call sites should go through the explicit flag.

2. Update local deployment call sites to pass the flag.

Canonical host deployments should use:

```bash
npx hardhat task:deployAllHostContracts --with-kms-generation true
```

Non-canonical host deployments should use:

```bash
npx hardhat task:deployAllHostContracts --with-kms-generation false
```

This includes at least:

- `host-contracts/docker-compose.yml`
- `test-suite/fhevm/docker-compose/host-sc-docker-compose.yml`
- `test-suite/fhevm/src/generate/compose.ts` for extra host chains
- `host-contracts/hardhat.config.ts` test override
- `library-solidity/hardhat.config.ts` test override
- `charts/contracts/values.yaml`

3. `library-solidity` should support the same explicit public deploy interface.

`library-solidity` is single-chain local/dev scaffolding, so it does not need to model production topology through separate public split tasks. But it should not keep a no-argument `deployAllHostContracts` path that silently deploys `KMSGeneration`.

Please make its public deploy entrypoint follow the same contract as `host-contracts`:

- `task:deployAllHostContracts --with-kms-generation true` deploys the full local stack, including `KMSGeneration`
- `task:deployAllHostContracts --with-kms-generation false` deploys only the non-canonical/common host contracts
- omitting `--with-kms-generation` fails with the same guidance as in `host-contracts`

The `library-solidity` test override can call `--with-kms-generation true`, preserving current local test behavior while keeping the task interface consistent across packages.

4. Update GitOps values or link the required GitOps rollout PR.

The local GitOps checkout still has `task:deployAllHostContracts` in:

- `/Users/work/code/zama/gitops/values/eth-blockchain/eth-sc-deploy/values-zama-mainnet.yaml`
- `/Users/work/code/zama/gitops/values/eth-blockchain/eth-sc-deploy/values-zws-testnet.yaml`
- `/Users/work/code/zama/gitops/values/kms-dev/host-sc-deploy/values-tkms-staging.yaml`

These should be changed to the explicit flag before 0.13 deployment. If GitOps is intentionally out of scope for this PR, please link the follow-up PR/issue because this is part of the safe rollout.

5. Canonical `KMS_GENERATION_CONTRACT_ADDRESS` should be validated during discovery.

`test-suite/fhevm/src/flow/discovery.ts` requires `PROTOCOL_CONFIG_CONTRACT_ADDRESS` for every modern host, but does not require `KMS_GENERATION_CONTRACT_ADDRESS` on the canonical/default host. `host-deploy` catches the normal path, but resume/regenerate paths can render an empty runtime env and fail later during bootstrap or container startup.

Discovery validation should require `KMS_GENERATION_CONTRACT_ADDRESS` on the canonical host when modern host artifacts are expected, and continue to reject it on non-canonical hosts.

6. Legacy gateway trigger services should remain in local gateway overrides.

`test-suite/fhevm/src/layout.ts` removed `gateway-sc-trigger-keygen` and `gateway-sc-trigger-crsgen` from `GROUP_BUILD_SERVICES['gateway-contracts']`, but bootstrap still uses those services when `usesHostKmsGeneration(state)` is false. A legacy bundle with `--override gateway-contracts` would deploy local gateway contracts but trigger keygen/CRS from the resolved image.

Please add those services back and keep matching build specs in `COMPONENT_BUILD_SPECS['gateway-sc']`.

7. Please add a regression test for the removed guard’s operational consequence.

The README documents that coordination is now runbook-enforced, which is fine, but the contract behavior should be explicit in tests:

- start keygen under context A
- rotate `ProtocolConfig` to context B while keygen is pending
- assert the old context A response can no longer complete because `KMSGeneration` validates signer/tx-sender against the current context
- abort the stuck request
- assert a new request can start

This test documents that `defineNewKmsContext` is intentionally allowed during pending key-management requests, and that abort is the recovery path if operators violate the runbook.

library-solidity/tasks/taskDeploy.ts
```diff
@@ -9,37 +9,72 @@ import path from 'path';
...
 ////////////////////////////////////////////////////////////////////////////////
 // All Host Contracts
 ////////////////////////////////////////////////////////////////////////////////
 
-task('task:deployAllHostContracts').setAction(async function (_, hre) {
-  if (process.env.SOLIDITY_COVERAGE !== 'true') {
-    await hre.run('clean');
-  }
+task('task:deployAllHostContracts')
+  .addOptionalParam(
```
obatirou: why optional and then have parseWithKmsGeneration force it?
addParam should exist?

Would allow to drop parseWithKmsGeneration also

host-contracts/test/tasks/taskHelpers.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export function readHostAddress(key: string): string {
+  const value = dotenv.parse(fs.readFileSync(HOST_ENV_FILE))[key];
+  if (!value) {
+    throw new Error(`Missing ${key} in ${HOST_ENV_FILE}`);
+  }
+  return value;
+}
+
+export async function deployFreshEmptyProxy(deployer: Wallet): Promise<string> {
```
obatirou: there is deployEmptyProxy in `host-contracts/test/utils/deploymentHelpers.ts`
Only one should be kept

host-contracts/tasks/taskDeploy.ts
```diff
@@ -36,33 +58,126 @@ function readExistingHostEnv(): Record<string, string> {
...
+    const wrapped = new Error(`${errorMessage} (${formatError(err)})`) as Error & { cause?: unknown };
+    wrapped.cause = err;
+    throw wrapped;
+  }
+}
+
+// OZ upgrades' upgradeProxy sometimes returns before the upgradeToAndCall tx has been mined on
+// interval-mining networks. Poll a view that only the upgraded implementation answers to so the
+// task only returns once the new impl is observable on-chain.
+async function waitForUpgradeLanded(
```
obatirou, line 82: seems not present in library solidity?

obatirou:
@claude /pr-review

obatirou:
@claude /pr-review

obatirou (dismissed):

obatirou (dismissed):

obatirou (dismissed):

host-contracts/tasks/utils/kmsGenerationConstants.ts
```diff
@@ -0,0 +1,4 @@
+export const KMS_CONTEXT_COUNTER_BASE = 0x07n << 248n;
```
isaacdecoded: Is this intentional that this counter uses hexadecimal format instead of BigInt like the others? Can I know why?

host-contracts/docker-compose.yml
```diff
@@ -97,6 +97,8 @@ services:
     depends_on:
       fhevm-sc-deploy:
         condition: service_completed_successfully
+      trigger-keygen:
```
isaacdecoded: Why is the crsgen dependant of the keygen? 

Moreover, I'd say keygen and crsgen also depend on:
```
anvil-node:
        condition: service_started
```

test-suite/fhevm/docker-compose/gateway-sc-docker-compose.yml
```diff
@@ -38,3 +38,31 @@ services:
     # contract is compiled, and it imports these addresses.
     volumes:
       - ${FHEVM_STATE_DIR:-../../../.fhevm}/runtime/addresses/gateway:/app/addresses  # workdir in gateway's Dockerfile is /app
+
+  gateway-sc-trigger-keygen:
```
isaacdecoded, line 42: Why do we need to roll back these services? Have you taken into account that the KMSGeneration on main (or v0.13.x) is read-only?

obatirou: cc @Eikix 
My understanding is to be use it for rollout tests

host-contracts/docker-compose.yml
obatirou: Good catch will fix

host-contracts/tasks/utils/kmsGenerationConstants.ts
obatirou: no reason, will sync

test-suite/fhevm/docker-compose/gateway-sc-docker-compose.yml
Eikix: Yes, v0.13 gateway KMSGeneration is read-only. These services are kept only for legacy pre-v0.13 gateway-KMS compatibility and should not be invoked for modern bundles. We can improve the comment to make that explicit.

host-contracts/docker-compose.yml
```diff
@@ -97,8 +100,6 @@ services:
     depends_on:
       fhevm-sc-deploy:
         condition: service_completed_successfully
-      trigger-keygen:
-        condition: service_completed_successfully
```
obatirou, line 101: this needs to be dependant on anvil too @Eikix

Eikix: trigger-keygen and trigger-crsgen both depend on fhevm-sc-deploy, and fhevm-sc-deploy now depends on anvil-node

obatirou (approved):

obatirou:
@mergify queue

PR #2325, opened 2026-04-17
feat(coprocessor): new host-listener-consumer

rudy-6-4:


coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+        chain_id = %config.chain_id,
+        "Host listener consumer graceful stop"
+    );
+    match consumer_result {
+        Ok(Ok(())) => info!("Consumer task completed successfully"),
+        Ok(Err(err)) => error!(error = %err, "Consumer task failed with error"),
+        Err(err) => error!(error = %err, "Consumer task panicked"),
+    }
+    client.cancel();
+    client.unregister_contracts(&contracts).await.unwrap();
```
obatirou: why unwrap instead of proper error?
Also the cancel below will not run if panic

obatirou: Also why unregister_contracts at the end of run_consumer?
We might want to do it only when necessary manually?

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                &mut db,
+                &block_logs,
+                config.acl_address,
+                config.tfhe_address,
+                config.database_retry_interval,
+                ingest_options,
+            )
+            .await
+            {
+                Ok(retries) => {
```
obatirou: retries is not used here

test-suite/fhevm/docker-compose/coprocessor-docker-compose.yml
```diff
@@ -62,6 +62,26 @@ services:
...
+    env_file:
+      - ${FHEVM_STATE_DIR:-../../../.fhevm}/runtime/env/coprocessor.env
+    command:
+      - host_listener_consumer
+      - --database-url=${DATABASE_URL}
+      - --acl-contract-address=${ACL_CONTRACT_ADDRESS}
+      - --tfhe-contract-address=${FHEVM_EXECUTOR_CONTRACT_ADDRESS}
+      - --url=redis://listener-redis:6379
+      #- --url=amqp://user:pass@listener-rabbitmq:5672
+      - --chain-id=12345
```
obatirou: hardcode chain id?

rudy-6-4: should use CHAIN_ID which is avail in coprocessor env

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
rudy-6-4: Good question. It's should be removed I think.
@nboisde , on normal stop & start and regular upgrade, do you confirm we should not unregister to not loss 1 block (assuming the queue is consummed 30 minutes later).

nboisde: I confirm the fact that you should never unregister unless if you want to stop watching a specific contract with a specific consumer. 
What unregister does is basically remove the filter on the core, so the core stops to publish for this contract on the destination queue.
For the coprocessor, unless there is a contract upgrade (hence a move of the address), we should never call this function, it won't lead to block loss, but could lead to a block missing your events.
Nonetheless, you should always try to create one, since the contract registering is silent if one already exists (helpful for contract address updates).

rudy-6-4:
Missing: update charts

.github/workflows/listener-core-docker-build.yml
```diff
@@ -0,0 +1,165 @@
...
+            // - workflow_dispatch: build if input is true, otherwise skip
+            const event = process.env.EVENT_NAME;
+            const needs = JSON.parse(process.env.NEEDS);
+            const inputs = JSON.parse(process.env.INPUTS);
+            const isLatestCommit = needs['is-latest-commit'].outputs?.is_latest === 'true';
+            const isWorkflowCall = inputs.is_workflow_call ?? false;
+
+            const decideAction = (changes, manualInput) => {
+              if (event === 'release') return 'build';
+              if (event === 'push') return isLatestCommit ? (changes ? 'build' : 'build') : 'skip';
```
obatirou: changes -> build always

coprocessor/fhevm-engine/host-listener/src/bin/consumer.rs
```diff
@@ -0,0 +1,132 @@
...
+        dependence_cross_block: args.dependence_cross_block,
+        dependent_ops_max_per_chain: args.dependent_ops_max_per_chain,
+        chain_id: args.chain_id,
+    };
+
+    run_consumer(config).await
+}
+
+// cargo run --bin  host_listener_consumer -- --url "amqp://user:pass@localhost:5672" --acl-contract-address 0xca2e8f1f656cd25c01f05d0b243ab1ecd4a8ffb6 --tfhe-contract-address 0xd82385dada1ae3e969447f20a3164f6213100e75 --database-url postgresql://postgres:postgres@localhost:5432/coprocessor --chain-id 1
+// cargo run --bin  host_listener_consumer -- --url "amqp://user:pass@localhost:5672" --acl-contract-address 0xf0ffdc93b7e186bc2f8cb3daa75d86d1930a433d --tfhe-contract-address 0x92c920834ec8941d2c77d188936e1f7a6f49c127 --database-url postgresql://postgres:postgres@localhost:5432/coprocessor --chain-id 11155111
```
obatirou: should be removed?

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+    });
+
+    let ingest_options = IngestOptions {
+        dependence_by_connexity: config.dependence_by_connexity,
+        dependence_cross_block: config.dependence_cross_block,
+        dependent_ops_max_per_chain: config.dependent_ops_max_per_chain,
+    };
+
+    let chain_id_str = config.chain_id.to_string();
+    let consumer_task = client.consume(move |payload, _cancel| {
```
obatirou, line 252: I realize there is prefetch, does it guarantee order? 
Is this a problem for dependence_chains?

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+        "Host listener consumer graceful stop"
+    );
+    match consumer_result {
+        Ok(Ok(())) => info!("Consumer task completed successfully"),
+        Ok(Err(err)) => error!(error = %err, "Consumer task failed with error"),
+        Err(err) => error!(error = %err, "Consumer task panicked"),
+    }
+    client.cancel();
+    health_check_cancel_token.cancel();
+    Ok(())
```
obatirou: We can return ok even if there was a panic in consumer task. Is this ok?

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+#[allow(clippy::too_many_arguments)]
+async fn ingest_with_retry(
```
obatirou, line 363: Could be factorized with poller

listener/Dockerfile
```diff
@@ -25,9 +25,11 @@ COPY . .
...
 # ── Runtime ──────────────────────────────────────────────────────────
-FROM gcr.io/distroless/cc-debian12:nonroot
+FROM gcr.io/distroless/cc-debian12:nonroot AS prod
 COPY --from=builder /app/target/release/listener_core /app/listener_core
 COPY --from=builder /app/crates/listener_core/migrations/ /app/migrations/
 WORKDIR /app
 ENTRYPOINT ["/app/listener_core"]
 CMD ["--config", "/config.yaml"]
+
+FROM prod AS dev
```
obatirou, line 35: why?

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+    let contracts = vec![config.acl_address, config.tfhe_address];
+    let chain_id: u64 = config.chain_id.parse()?;
+    let chain_id = ChainId::try_from(chain_id)?;
+
+    let blockchain_tick = HeartBeat::new();
+    let blockchain_timeout_tick = HeartBeat::new();
+    let blockchain_provider = Arc::new(RwLock::new(None));
+
+    let broker_url = config.url; // e.g."amqp://user:pass@localhost:5672";
+    let broker = Broker::from_url(&broker_url).await?;
```
nboisde, line 207: I understand now the ensure publish question.
Lets keep it like this for now, since we only publish for 2 contracts and mostly consume events.

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
nboisde: I agree with this, if prefetch count is not set to 1, the consumer is able to handle many messages at the same time, for ethereum it won't cause an issue i'd say since there is a block every 13 seconds.

Nonetheless, on faster chains, it could break order (40ms chains), but from my understanding its not mendatory to follow the exact order.
We should look, and keep in mind the tradeoff here for the fastest chains indeed, between reducing the prefetech count or pass it to 1, and if we are able to keep up with fastests chains.

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+                Err((err, retries)) => {
+                    inc_db_errors(&chain_id_str, 1);
+                    error!(
+                        block_number = block_summary.number,
+                        block_hash = ?block_logs.summary.hash,
+                        error = %err,
+                        retries = retries,
+                        "Failed to ingest block"
+                    );
+                    Err(HandlerError::Transient(err.into()))
```
nboisde, line 329: Tremendous catch ! 
The question above is valid here, and I assume that the block ordering is not mendatory here.

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
rudy-6-4: The payloads are pushed in order in the queue, I expect the client to get them in order !?
Do you think it's not guaranteed ?
It will less efficient if inserted in bad order but should still work.

listener/crates/shared/broker/src/lib.rs
```diff
@@ -242,6 +242,17 @@ impl Broker {
         topology.as_ref().clone()
     }
```
nboisde, line 244: Note: Could be helpful in the future to add an optionnal parameter to ensure publish

listener/Dockerfile
rudy-6-4: docker build is called with these label (prod, dev). it's a kind of convention  to use the generic actions (same as in other docker file for copro)

.github/workflows/listener-core-docker-build.yml
rudy-6-4: It's needed until a docker image is published in main I think. The change detection does if no previous image exists. (Eg the existing shared code is not working ith a new docker image , never published)

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
obatirou: honestly, I do not have the answer that is why I ask. Maybe @nboisde or @maxnovawind?
Also, if less efficient we should be careful by how much and the impact it would have

nboisde (commented):
Thank you @rudy-6-4, Looks good to me.

Just a overall remark here: 
Depending on our strategy, for the first deployment, maybe we should have the feature of the host listener launches the core by publishing a message to the right queue of the core.
We could try to implement this in a follow up PR, I can even implement a function on the consumer library, in order to handle this on my next PR.
Indeed we maybe don't want that core receiving blocks before the host listener has the chance to create his queues and called the watch function.

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
nboisde: I think, the events will be consumed in order yes, but you just introduce the way of consuming multiple events at the same time. So if a block processing is longer than another one, the db insertion could happen after yes (e.g  processing block1 is slower than processing block2, you can have "unordered reception in the db" i'd say, but events are still delivered in order). I am sure this is the right behaviour for rmq. I need this to be confirmed by @maxnovawind regarding redis.

nboisde (dismissed):
As said before the latest change, looks good to me ! 
Some quick improvements regarding the release process that could be discussed and addressed in a follow up PR.

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
obatirou: So maybe we should have the option to configure the prefetch from the consumer?
In the case of host-listner, I think we should consume one by one for now and evaluate the impact of prefetch later when we have fast chains?

nboisde: I would say its possible here since there is only a block each 13~ seconds, but we should be careful in the future, we should also be careful if the coprocessor host-listener takes more than 13 seconds to process a block. 

Also, in the catch-up feature (WIP) or reorg process, we should be careful as well.
Indeed the limitation I see, is that if you ask for 1k block during catchup lets say, we can have a problem by processing then once at the time. 
Same if at some point the core stalls for an unexpected reason.

What I would like to avoid is: too many pending message in a queue. 

But definitely, have an option to setup the prefetch count for the host-listener is really something useful.

.github/workflows/listener-core-docker-build.yml
rudy-6-4: replaced by Ghislain's workflow

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
rudy-6-4: Fixed

nboisde (dismissed):
Just to be checked: prefetch count to 1 or not, just in case you guys want to be sure for order, but I don't think its necessary, normally the copro should not take more that 13 seconds to process a block. Maybe for reorg backtracking, could be useful.
No ensure publish needed for now (e.g no catchup feautre yet), and only 2 publish at startup, which is normally good.
We also figured out that for the first startup we are ok to miss some blocks since we still have the others listeners running in parallel, so no need to start the core from the copro.

Thank you for adding the changes to the namings here @rudy-6-4.
LGTM, for me we can proceed if we are good with yohan changes about the panic.

obatirou (dismissed):
lgtm
Only prefetch to set to 1 for now. Factorization with poller but could be done later

rudy-6-4:
Resolving everything to merge. I'll add ticket later for what is still to be done

nboisde (dismissed):

obatirou (dismissed):

rudy-6-4:
@mergifyio queue

rudy-6-4:
orchestrated e2e: https://github.com/zama-ai/fhevm/actions/runs/25319513290/job/74227905563?pr=2325

rudy-6-4:
@claude /pr-review

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+
+pub async fn run_consumer(config: ConsumerConfig) -> Result<()> {
+    info!("Starting consumer with config: {:?}", config);
+    let contracts = vec![config.acl_address, config.tfhe_address];
+    let chain_id: u64 = config.chain_id.parse()?;
+    let chain_id = ChainId::try_from(chain_id)?;
+
+    let blockchain_tick = HeartBeat::new();
+    let blockchain_timeout_tick = HeartBeat::new();
+    let blockchain_provider = Arc::new(RwLock::new(None));
```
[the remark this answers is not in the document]
rudy-6-4: The provider is only checked directly in case of no signal (blockchain_tick).

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -347,6 +352,14 @@ jobs:
         run: |
           ./fhevm-cli test "$TEST_PROFILE"
 
+      - name: Host listener consumer only selected e2e suite
+        working-directory: test-suite/fhevm
+        run: |
+          trap 'docker start coprocessor-host-listener coprocessor-host-listener-poller>/dev/null 2>&1 || true' EXIT
+          docker stop coprocessor-host-listener coprocessor-host-listener-poller
+          ./fhevm-cli test "$TEST_PROFILE"
+          trap - EXIT
```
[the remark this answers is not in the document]
rudy-6-4: Waiting CI to finish orchestrated test before fixing.

Eikix:
Did some back and forth with AI. Here is the report.

I focused mostly on the new listener-core path, the FHEVM CLI, and the E2E/compat coverage.

The latest commit improves the workflow cleanup a bit, but the main concerns below still stand.

A few things look blocking to me:

1. **The new consumer should run drift-revert init before using the coprocessor DB.**

`host_listener` and `host_listener_poller` both initialize drift-revert before reading or changing DB state.

`host_listener_consumer` opens the DB and calls `promote_all_dep_chains_to_fast_priority` before doing that.

If a revert signal is pending, the consumer can ingest listener-core blocks against pre-revert state.

2. **The consumer drops listener-core block flow semantics.**

Every listener-core `BlockPayload` becomes `BlockLogs { catchup: false, finalized: false }`.

So `payload.flow` is ignored. Reorged and catchup blocks are treated like normal pending live blocks.

The old listener paths either insert catchup blocks as finalized or call `update_finalized_blocks` for live blocks. That is what maintains finalized/orphaned block status in the coprocessor DB.

The new path should preserve the same semantics.

3. **Multi-coprocessor consumers need unique listener-core identities.**

`two-of-two.yaml` injects unique `--service-name` values for `host-listener-consumer`.

But `two-of-three.yaml` and `two-of-three-multi-chain.yaml` do not.

The consumer builds its listener-core identity from `service_name + chain_id`. If several coprocessors share the default service name, they can share the same listener-core topic/group and split messages instead of each coprocessor receiving the same blocks.

This is important because direct E2E defaults to `two-of-three`, and orchestrated E2E uses `two-of-three-multi-chain`.

The safer fix is to generate `--service-name=${serviceName}` automatically for every cloned `host-listener-consumer`.

4. **The “consumer only” E2E step still does not fully isolate the new path.**

The latest commit restarts the primary old listeners after the test, which is good.

But the step still stops only `coprocessor-host-listener` and `coprocessor-host-listener-poller`.

In multi-coprocessor or multi-chain scenarios, other old listener containers can still be running, for example `coprocessor1-host-listener` or `coprocessor-host-listener-chain-b`.

So the test can still pass through the old listener path.

This step should stop all old host-listener / host-listener-poller containers for the rendered topology. It should also skip clearly when the resolved bundle does not include `host_listener_consumer`.

5. **Direct PR E2E does not build listener-core.**

Direct PR E2E sets `BUILD=true` and runs `fhevm-cli up --build`, but `--build` expands only `OVERRIDE_GROUPS`.

`listener-core` is not an override group, and the listener-core compose file has no generated local build override.

So regular PR E2E can build the new coprocessor consumer while still running a published listener-core image.

That weakens coverage for this PR, because it changes both sides of the new path.

Either listener-core should become a real build/override target, or direct PR E2E should receive the image built by `listener-docker-build`.

Less blocking, but still worth fixing:

6. **FHEVM CLI repair/resume does not track listener-core itself.**

Repair now tracks `coprocessor-host-listener-consumer`, but not `listener-redis` or `listener-publisher-for-anvil`.

So a resumed stack can be considered healthy while the listener-core side is down.

If `listener-core` is a real lifecycle step, repair should include its steady-state containers when the consumer is supported.

7. **Helm values mention a consumer that is not deployed.**

The chart adds `hostListenerConsumerShared` and per-chain `hostListenerConsumer` values, but I do not see a Deployment, Service, or ServiceMonitor template for `host-listener-consumer`.

So enabling the new values does not start the new ingestion path.

Either the templates should be added here, or the values should not be presented as deployable yet.

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
rudy-6-4, line 222: could be after drift_revert::init t minimize eventual catchup phase

rudy-6-4:
@mergifyio queue

Eikix (dismissed):
https://github.com/zama-ai/fhevm-internal/issues/1348

Eikix:
Interesting AI findings:

1. `kms_generation_address` is passed to ingestion, but only ACL and TFHE contracts are registered with listener-core. In consumer-only mode, KMSGeneration logs will not be delivered by the broker, so key/CRS activation events can be missed. `config.kms_generation_address` should be included in the registered contracts.

2. Drift-revert startup coordination is not run before normal DB access. The consumer creates its regular `Database` handle and may promote dependence chains before any shared `drift_revert::init(..., None)` check is performed. The per-message drift gate only runs after a broker payload is delivered, so startup during `Pending`/`Reverting` can still touch DB state before revert completion, and unresolved `Failed` signals are not rejected. The shared drift-revert init should run before DB state reads/writes, while keeping the Done/catchup gate for replay ordering.

rudy-6-4:
> Interesting AI findings:
> 
> 1. `kms_generation_address` is passed to ingestion, but only ACL and TFHE contracts are registered with listener-core. In consumer-only mode, KMSGeneration logs will not be delivered by the broker, so key/CRS activation events can be missed. `config.kms_generation_address` should be included in the registered contracts.
> 2. Drift-revert startup coordination is not run before normal DB access. The consumer creates its regular `Database` handle and may promote dependence chains before any shared `drift_revert::init(..., None)` check is performed. The per-message drift gate only runs after a broker payload is delivered, so startup during `Pending`/`Reverting` can still touch DB state before revert completion, and unresolved `Failed` signals are not rejected. The shared drift-revert init should run before DB state reads/writes, while keeping the Done/catchup gate for replay ordering.

Yes 1 has been forgotten during the rebase.
For 2 , I can postpone it to the first write in db.

rudy-6-4:
> Interesting AI findings:
> 
> 1. `kms_generation_address` is passed to ingestion, but only ACL and TFHE contracts are registered with listener-core. In consumer-only mode, KMSGeneration logs will not be delivered by the broker, so key/CRS activation events can be missed. `config.kms_generation_address` should be included in the registered contracts.
> 2. Drift-revert startup coordination is not run before normal DB access. The consumer creates its regular `Database` handle and may promote dependence chains before any shared `drift_revert::init(..., None)` check is performed. The per-message drift gate only runs after a broker payload is delivered, so startup during `Pending`/`Reverting` can still touch DB state before revert completion, and unresolved `Failed` signals are not rejected. The shared drift-revert init should run before DB state reads/writes, while keeping the Done/catchup gate for replay ordering.

For 2. Drift init will not be called as it requires auto-catchup mecanism. Creating the database object should not do any extra writing in database or promotion. I ll add a bare database object constructor if needed and delay promotion to first write.

rudy-6-4:
@claude /pr-review Can you check the drift-pausing mecanism. The consumer is not able to catchup after the db drift revert. So it pause during drift-revert and wait until the main host listener catchup to where the consumer was before.

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+            )
+            .await;
+            match drift_revert_is_over {
+                Ok(false) => {
+                    return Err(HandlerError::Transient(Box::from(
+                        "Drift in progress",
+                    )))
+                }
+                Ok(true) => (), // all good
+                Err(err) => error!(%err, "Can't check drift-revert status"),
```
Eikix: In case the drift revert table is not reachable, do we want to log an error and continue? Or do we want a transient error?

i dont know

rudy-6-4: I have changed to have a transient error == a retry later

Eikix (approved):

rudy-6-4:
@mergifyio queue

nboisde:
@mergifyio queue

rudy-6-4:
@mergifyio refresh

nboisde (approved):

eudelins-zama:
@mergifyio refresh

PR #2328, opened 2026-04-20
chore(kms-connector): ethereum chain id for keygen eip712

eudelins-zama:
Needed to validate the EIP712 of the keygen flow now that `KMSGeneration` contract lives on Ethereum.

### DIFF
--- a/kms-connector/config/kms-worker.toml
+++ b/kms-connector/config/kms-worker.toml
@@ -12,7 +12,7 @@ gateway_url = "http://localhost:8545"
 # ENV: KMS_CONNECTOR_GATEWAY_CHAIN_ID
 gateway_chain_id = 54321
 
-# Chain ID for the Ethereum network (required)
+# Chain ID for the Ethereum-like network; defaults to Sepolia (required)
 # ENV: KMS_CONNECTOR_ETHEREUM_CHAIN_ID
 ethereum_chain_id = 11155111
 
--- a/kms-connector/config/kms-worker.toml
+++ b/kms-connector/config/kms-worker.toml
@@ -12,6 +12,10 @@ gateway_url = "http://localhost:8545"
 # ENV: KMS_CONNECTOR_GATEWAY_CHAIN_ID
 gateway_chain_id = 54321
 
+# Chain ID for the Ethereum network (required)
+# ENV: KMS_CONNECTOR_ETHEREUM_CHAIN_ID
+ethereum_chain_id = 11155111
+
 # List of KMS-core shards gRPC endpoints (required)
 # Format: http://host:port
 # ENV: KMS_CONNECTOR_KMS_CORE_ENDPOINTS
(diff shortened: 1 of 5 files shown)


kms-connector/config/kms-worker.toml
```diff
@@ -12,6 +12,10 @@ gateway_url = "http://localhost:8545"
 # ENV: KMS_CONNECTOR_GATEWAY_CHAIN_ID
 gateway_chain_id = 54321
 
+# Chain ID for the Ethereum network (required)
```
dvdplm:
```suggestion
# Chain ID for the Ethereum-like network; defaults to Sepolia (required)
```

kms-connector/crates/kms-worker/src/core/event_processor/kms.rs
```diff
@@ -28,7 +28,7 @@ where
         let domain = Eip712DomainMsg {
             name: config.kms_generation_contract.domain_name.clone(),
             version: config.kms_generation_contract.domain_version.clone(),
-            chain_id: U256::from(config.gateway_chain_id).to_be_bytes_vec(),
+            chain_id: U256::from(config.ethereum_chain_id).to_be_bytes_vec(),
```
dvdplm, line 31: Not new code, but does `to_be_bytes_vec` allocate? If yes, that is probably not necessary. The `U256` type is stack allocated (right?) and the bytes are `Copy`.

kms-connector/crates/kms-worker/src/core/config.rs
```diff
@@ -245,6 +250,7 @@ mod tests {
             env::set_var("KMS_CONNECTOR_EVENTS_BATCH_SIZE", "15");
             env::set_var("KMS_CONNECTOR_GATEWAY_URL", "http://localhost:9545");
             env::set_var("KMS_CONNECTOR_GATEWAY_CHAIN_ID", "31888");
+            env::set_var("KMS_CONNECTOR_ETHEREUM_CHAIN_ID", "31444");
```
dvdplm, line 253: Is 31444 just a random number or an actual testnet ID? Either way, a comment would be good.

dvdplm (approved):
LGTM, just nitpicks.

kms-connector/crates/kms-worker/src/core/config.rs
eudelins-zama: (It's the test section here).

Its' a random value just for testing the override of the config by env var.

We don't have comments for other values, so not sure why we would add one just for this one

kms-connector/config/kms-worker.toml
eudelins-zama: Fixed in d2d668018810611a8a728cf703cc661c09559c1c

kms-connector/crates/kms-worker/src/core/event_processor/kms.rs
eudelins-zama: According to Claude:

> Good catch from the reviewer in principle — to_be_bytes_vec does heap-allocate a 32-byte Vec<u8>.
> But here the allocation is unavoidable: Eip712DomainMsg.chain_id is a protobuf-generated Vec<u8>, so any alternative (to_be_bytes::<32>().to_vec(), manual buffer + .to_vec()) ends up doing the same allocation.
> Plus this runs once at KMSGenerationProcessor::new() — not a hot path.

PR #2329, opened 2026-04-20
feat(gateway-contracts): unify user decryption

manoranjith:
## Summary

- Introduces a unified payload for user decryption supporting direct,
  delegated, and mixed batches via a new
  `userDecryptionRequest(HandleEntry[], ...)` overload.
- Gateway performs no on-chain signature verification on this path —
  emits the full payload verbatim in a second overloaded
  `UserDecryptionRequest` event for off-chain verification by the KMS
  Connector.
- Refs zama-ai/fhevm-internal#1292 for full context, deviations, and
  rollout notes.

## Compatibility

- Legacy `userDecryptionRequest(CtHandleContractPair[])`,
  `delegatedUserDecryptionRequest`, associated event, and on-chain
  EIP-712 helpers retained and marked `@custom:deprecated`.
- Kept for backward compatibility with older relayer-sdk versions
  during the deprecation window.
- Kept for forward compatibility with pre-upgrade off-chain
  components (kms-connector, relayer) during the rollout window.
- Event overloading gives each shape a distinct `topic0` so legacy
  and new subscribers don't interfere.

## Notes

- `hardhat.config.ts` enables `viaIR: true` because the 9-field event
  emit exceeds Solidity's 16-slot stack without it.
- Existing legacy tests pin the canonical overload signature to
  resolve ethers v6 ambiguity.
- `rust_bindings/` regeneration blocked on `forge >= 1.3.1` (0.2.0
  installed locally); run `foundryup` then `make update-bindings`
  before merge.

## Tests

Added `Unified User Decryption (EIP-712)` describe block in
`test/Decryption.ts` (17 tests). Coverage:

- **Happy paths** — direct-only, mixed direct/delegated, permissive
  (empty `allowedContracts`), specific `allowedContracts`, opaque
  signature accepted, legacy event not emitted on the unified path.
- **Request-format reverts** — empty `handles`, `allowedContracts`
  over max length, `durationSeconds` zero or over max,
  `startTimestamp` in the future, expired validity window.
- **Handle conformance reverts** — mismatched chain IDs across a
  batch, unregistered host chain.
- **View-function overload** —
  `isUserDecryptionReady(HandleEntry[], bytes)` returns true / false
  on registered / missing / empty inputs.

Full suite: 350 passing, 0 failing (89 preserved legacy-path + 17
new unified-path tests).

---

**EDIT (2026-04-21):** `rust_bindings/` are now regenerated in the
follow-up commit `chore(gateway-contracts): regenerate rust bindings`
(after a local `foundryup` to forge 1.5.1). `foundry.toml` mirrors
`hardhat.config.ts` with `via_ir = true` so `forge bind` compiles
cleanly. `make lint-bindings` (clippy) passes clean.


gateway-contracts/contracts/interfaces/IDecryption.sol
```diff
@@ -33,8 +36,53 @@ interface IDecryption {
...
+    }
+
+    /**
+     * @notice The signed payload carried by the unified `UserDecryptionRequest` event.
+     * @dev Packs every field the EIP-712 signature covers plus the signature itself. Signature
+     * validation has moved off-chain, so the gateway forwards the full signed payload verbatim.
+     * `snsCtMaterials` and `handles` are not part of the signed struct and stay as separate
+     * event parameters.
+     */
+    struct UserDecryptionRequestPayload {
```
isaacdecoded, line 66: nit: maybe worth it also including the `HandleEntry[] handles` in this payload.

gateway-contracts/contracts/interfaces/IDecryption.sol
```diff
@@ -340,7 +475,7 @@ interface IDecryption {
     ) external view returns (bool);
 
     /**
-     * @notice Indicates if handles are ready to be decrypted by a user.
+     * @notice Indicates if handles are ready to be decrypted by a user (legacy path input shape).
      * @param ctHandleContractPairs The ciphertext handles with associated contract addresses.
      * @param extraData Generic bytes metadata for versioned payloads. First byte is for the version.
      */
```
isaacdecoded, line 489: Should we also add the `@custom:deprecated` comment? [Similar to isDelegatedUserDecryptionReady](https://github.com/zama-ai/fhevm/pull/2329/changes#diff-14c62001db03729747b5ad7bc744639d1f05199712fca3b9e99187c3c47b4994R514-R517) below.

isaacdecoded (approved):
LGTM! 👍

gateway-contracts/contracts/interfaces/IDecryption.sol
manoranjith: Accepted. Fixed in https://github.com/zama-ai/fhevm/pull/2329/commits/1cc7a59a336275935c92fa4be8ced068f8677dc6

gateway-contracts/contracts/interfaces/IDecryption.sol
manoranjith: Was thinking about it. But per [RFC 016](https://github.com/zama-ai/tech-spec/blob/fce9bcdf73605a408f548a1d74cfea8fb02dd2f1/rfcs/016-unified-eip712-decryption-request.md?plain=1#L82-L87), this field is deliberately excluded from the attestation payload and I keep the struct to include only those fields that are attested.


So, two possibilities here:

1. Make it part of the field (as per your suggestion). Because need not be necessarily aligned.

2. Make the documentation adequate, so its obvious why its excluded.



Which one would you prefer @isaacdecoded ?

isaacdecoded: Ah, I see. I think it's ok then since it's aligned with what was already agreed on the RFC, as you mentioned. Thanks for the clarification.

PR #2331, opened 2026-04-20
feat(test-suite): add rollout compatibility matrix capability

Eikix:
## What
This PR adds a rollout compatibility matrix capability to the test suite.

The feature allows a compat target to define:
- a `from` bundle
- a `to` bundle
- an explicit ordered rollout plan
- grouped steps or ordered substeps where needed

From that, the CLI and CI can:
- materialize every emitted rollout state as a lock file
- generate a GitHub Actions matrix for those states
- boot each state and run the standard test suite on that state

This PR also includes one concrete compat target that demonstrates the feature:
- `v0.12 -> main`

That target is the first real example of how to use the capability, not the whole purpose of the PR.

## Why
The operational need is to validate a proposed rollout order before rollout day.

Concretely:
- rollout should be described as an explicit ordered sequence of emitted states
- every emitted state should be testable in CI
- failures should expose either an invalid rollout order or a real compatibility break
- compat logic should help mixed-version states boot, but should not patch semantic failures into passes

This PR introduces the machinery for that.

### Descriptor and planner
Compat tests now support an ordered rollout descriptor with named steps.
A step can be:
- one grouped unit
- or a set of ordered substeps

The CLI expands that descriptor into:
- one lock file per emitted rollout state
- one matrix entry per emitted rollout state

### CI execution
The compatibility workflow is available through `workflow_dispatch` for now.
Each emitted state is booted and tested in its own job with `standard`.

### Harness behavior
Compat runs use the current test-suite harness code.
For compat runs, the local `test-suite` image is built and used automatically.

For the current contract, the harness derives `@zama-fhe/relayer-sdk` from the `from` bundle.
That keeps the rollout client contract on the `from` SDK line while still using the current harness code and compat shims.

### Example target
The PR includes a first checked-in compat target:
- `v0.12 -> main`

It demonstrates:
- explicit rollout ordering
- contracts grouped together
- kms grouped together
- coprocessor split into ordered substeps
- matrix generation for each emitted state

## Current contract
This PR currently proves:
- every emitted rollout state passes `standard` with the `from` SDK line

This PR does not yet claim:
- emitted rollout states work with the `to` SDK line

## Validation
- `cd test-suite/fhevm && bun test src/rollout.test.ts src/cli.test.ts src/render-compose.test.ts`
- `cd test-suite/fhevm && ./fhevm-cli rollout --compat-test ./compat-tests/v0.12-to-main.json --matrix | jq '.include[:2], .include[-1:]'`
- `docker build -f test-suite/e2e/Dockerfile --build-arg RELAYER_SDK_VERSION=0.5.0-alpha.1 --target builder -t fhevm-test-suite-sdk-smoke .`

## Out of scope
- proving emitted rollout states with the `to` SDK line
- automatically searching for alternate rollout orders
- moving contract migration semantics out of deploy scripts and into rollout

closes https://github.com/zama-ai/fhevm-internal/issues/1252


.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -9,7 +9,7 @@ on:
       - labeled
       - unlabeled
   workflow_dispatch:
-    inputs: &workflow_inputs
```
eudelins-zama, line 12: Why removing the YAML anchor here?

We now have 200 lines just to define the inputs of the workflow... 🫠

.github/workflows/test-suite-compatibility-matrix.yml
```diff
@@ -0,0 +1,129 @@
+name: test-suite-compatibility-matrix
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+      - unlabeled
```
eudelins-zama, line 12: I think we want to run this on Mergify pull request only, not every PR, so we need to add filtering (similar to what was done in `test-suite-e2e-tests.yml`.

We could even start with a `workflow_dispatch` only first, to be sure it's working properly

Eikix: Agreed

.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -102,26 +114,26 @@ on:
...
         required: true
-    inputs: *workflow_inputs
+    inputs: *shared_inputs
 
 permissions: {}
 
 # Cancel stale PR runs on new pushes while keeping manual dispatches independent.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && format('pr-{0}', github.event.pull_request.number) || github.ref }}-${{ github.event_name == 'workflow_dispatch' && github.run_id || 'auto' }}
+  group: ${{ github.workflow }}-${{ inputs.rollout-step != '' && format('compat-{0}-{1}', inputs.compat-test || 'default', inputs.rollout-step) || github.event_name == 'pull_request' && format('pr-{0}', github.event.pull_request.number) || github.ref }}-${{ github.event_name == 'workflow_dispatch' && github.run_id || 'auto' }}
```
PanGan21: Can it be that: `compat-{compat-test}-{step}` key format conflict across different prs? If that is the case, consider adding `github.ref` or the pr number so different pr runs do not cancel each other.

PanGan21 (approved):

Eikix:
@mergifyio queue

eudelins-zama (approved):

Eikix:
@mergifyio queue

PR #2332, opened 2026-04-20
fix(host-contracts): use V1 extra data in KMSGeneration and drop V2

obatirou:


host-contracts/contracts/KMSGeneration.sol
```diff
@@ -1053,7 +1050,7 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
     }
 
     function _encodeRequestExtraData(uint256 contextId) internal pure virtual returns (bytes memory) {
-        // epochId is hardcoded to 0 until resharing (key rotation across epochs) is implemented.
-        return abi.encodePacked(EXTRA_DATA_V2, contextId, uint256(0));
+        // Emit V1 (contextId only) until resharing (RFC 005) is implemented.
```
isaacdecoded, line 1053: nit: not sure if it's meaningful for this contract version to explain that the V1 will be used until RFC 005, but not a big thing if we leave it anyway.

isaacdecoded (approved):
Awesome 👍

host-contracts/contracts/KMSGeneration.sol
obatirou: Let's merge to not block E2E more, we will see on the last pass with the PR on main if we remove it

PR #2335, opened 2026-04-20
chore(common): add infra team responsibilities for codeowner of Dockerfiles

maxnovawind:


eudelins-zama (commented):
Your PR title was not passing the CI (invalid format).
I updated it, but you need to force push so the CI workflow is properly re-triggered.
If I just re-run the workflow, it uses the old title

PR #2338, opened 2026-04-21
fix(contracts): load upgrade baseline from latest release

Eikix:
## Summary
- resolve the contract upgrade baseline from the latest stable GitHub release instead of a hardcoded tag in the three upgrade-related workflows
- update the upgrade-test warnings so they no longer refer to `UPGRADE_FROM_TAG`
- retroactively bump `GatewayConfig` for the current post-`v0.12.1` release cycle by moving its semantic patch version to `0.5.1`, moving `REINITIALIZER_VERSION` to `7`, and replacing the current upgrade entrypoint with `reinitializeV6(uint256 initialKmsContextId)`

## Validation
- `python3` YAML parse of the three updated workflows
- `bun ci/upgrade-check/list.ts --from v0.12.1 --package gateway-contracts`

closes zama-ai/fhevm-internal#1277


gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -226,39 +226,16 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
+     * @notice Re-initializes the contract from V5.
+     * @dev No storage migration is required for the post-v0.12.1 behavior change, but the
+     * upgrade flow keeps the KMS context ID argument so existing upgrade tasks keep working.
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    function reinitializeV5(uint256 initialKmsContextId) public virtual reinitializer(REINITIALIZER_VERSION) {
-        GatewayConfigStorage storage $ = _getGatewayConfigStorage();
-
+    function reinitializeV6(uint256 initialKmsContextId) public virtual reinitializer(REINITIALIZER_VERSION) {
```
PanGan21: Do we still need to pass `initialKmsContextId` if it is not used?

Eikix: Good catch, will update upgrade tasks

PanGan21 (approved):

Eikix:
If you have a minute @zmalatrax or @isaacdecoded

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -226,40 +226,12 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
     }
 
     /**
-     * @notice Re-initializes the contract from V4.
-     * @dev Define a `reinitializeVX` function once the contract needs to be upgraded.
+     * @notice Re-initializes the contract from V5.
+     * @dev No storage migration is required for the post-v0.12.1 behavior change.
```
isaacdecoded, line 230: nit: I'm not sure we can confirm that we will not need any storage migration post-v0.12.1 beforehand. Probably for v0.13 we don't, but not sure in v0.14, v0.15, etc.

isaacdecoded (approved):
LGTM 👍

gateway-contracts/contracts/GatewayConfig.sol
Eikix: We will change the comment since at every release we might that reinitializer

Eikix:
@mergifyio queue

PR #2341, opened 2026-04-21
feat(coprocessor): automatic drift recovery via revert

dartdart26:
Try to resolve a state drift when a minotiry of coprocessors are drifting, meaning that the revert only triggers if a conensus on a ciphertext has been reached.

The coprocessor state revert script is run when a drift is detected. It works by separating services into two categories:
 * runner - gw-listener
 * non-runners - all other services

When a drift is detected for a ciphertext, we determine the host chain and the block ID to revert to and insert a signal into the DB with this data. When observing this signal, all services re-exec and replace themselves with a new instance and then:
 * the gw-listener waits for a grace period before running the state revert script
 * all services (including gw-listener) wait until revert completes and continue executing normally

In this commit, we also add support for 2 out of 3 coprocessors for E2E tests.

We also recognize that, as of now, we can only recover from a state drift due to computations being bad. For input proofs, we can't recover as revert only works based on host chain and block ID and if an input ciphertext is not used on a host chain, we won't observe a drift. For that, we change the drift injector such that it only inserts drifts for computation handles.

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
@@ -1,0 +1,479 @@
+//! Auto drift recovery: shared coordination logic.
+//!
+//! When the gw-listener detects ciphertext drift, it writes a row to
+//! `drift_revert_signal`. All coprocessor services poll that table and
+//! re-exec themselves when a signal appears. On startup, all fresh
+//! processes check for a pending signal:
+//!  * gw-listener runs the revert SQL,
+//!  * other services wait until it's done, then all proceed normally
+
+use std::time::Duration;
+
+use sqlx::postgres::PgPoolOptions;
+use sqlx::{Pool, Postgres, Row};
+use tokio_util::sync::CancellationToken;
+use tracing::{error, info, warn};
+
+const REVERT_SQL_TEMPLATE: &str =
+    include_str!("../../db-migration/db-scripts/revert_coprocessor_db_state.sql");
+
+/// How often services poll `drift_revert_signal` for state changes.
+pub const POLL_INTERVAL: Duration = Duration::from_secs(2);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SignalStatus {
+    Pending,
+    Reverting,
+    Done,
+    Failed(String),
+}
+
+impl SignalStatus {
+    fn as_db_str(&self) -> String {
+        match self {
+            Self::Pending => "pending".to_owned(),
+            Self::Reverting => "reverting".to_owned(),
+            Self::Done => "done".to_owned(),
+            Self::Failed(reason) => format!("failed: {reason}"),
+        }
+    }
+
+    fn from_db_str(s: &str) -> Self {
+        match s {
+            "pending" => Self::Pending,
+            "reverting" => Self::Reverting,
+            "done" => Self::Done,
+            other => Self::Failed(other.strip_prefix("failed: ").unwrap_or(other).to_owned()),
+        }
+    }
+
+    pub fn is_in_flight(&self) -> bool {
+        matches!(self, Self::Pending | Self::Reverting)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct DriftRevertSignal {
+    pub id: i64,
+    pub host_chain_id: i64,
+    pub offending_host_block_number: i64,
+    pub status: SignalStatus,
+}
+
+/// Create a new drift revert signal.
+/// Returns `Some(id)` if created, `None` if there's already an in-flight
+/// revert for this host chain.
+pub async fn create_revert_signal(
+    pool: &Pool<Postgres>,
+    host_chain_id: i64,
+    offending_host_block_number: i64,
+) -> anyhow::Result<Option<i64>> {
+    let row = sqlx::query(
+        "INSERT INTO drift_revert_signal (host_chain_id, offending_host_block_number, status) \
+         SELECT $1, $2, $3 \
+         WHERE NOT EXISTS ( \
+             SELECT 1 FROM drift_revert_signal \
+             WHERE host_chain_id = $1 AND (status = $3 OR status = $4) \
+         ) \
+         RETURNING id",
+    )
+    .bind(host_chain_id)
+    .bind(offending_host_block_number)
+    .bind(SignalStatus::Pending.as_db_str())
+    .bind(SignalStatus::Reverting.as_db_str())
+    .fetch_optional(pool)
+    .await?;
+
+    match row {
+        Some(r) => {
+            let id: i64 = r.get("id");
+            Ok(Some(id))
+        }
+        None => Ok(None),
+    }
+}
+
+/// Fetch the latest signal row, if any.
+pub async fn latest_signal(pool: &Pool<Postgres>) -> anyhow::Result<Option<DriftRevertSignal>> {
+    let row = sqlx::query(
+        "SELECT id, host_chain_id, offending_host_block_number, status \
+         FROM drift_revert_signal ORDER BY id DESC LIMIT 1",
+    )
+    .fetch_optional(pool)
+    .await?;
+
+    Ok(row.map(|r| {
+        let status_str: String = r.get("status");
+        DriftRevertSignal {
+            id: r.get("id"),
+            host_chain_id: r.get("host_chain_id"),
+            offending_host_block_number: r.get("offending_host_block_number"),
+            status: SignalStatus::from_db_str(&status_str),
+        }
+    }))
+}
+
+/// Update the status of a signal row.
+pub async fn update_signal_status(
+    pool: &Pool<Postgres>,
+    signal_id: i64,
+    status: &SignalStatus,
+) -> anyhow::Result<()> {
+    sqlx::query("UPDATE drift_revert_signal SET status = $1, updated_at = NOW() WHERE id = $2")
+        .bind(status.as_db_str())
+        .bind(signal_id)
+        .execute(pool)
+        .await?;
+    Ok(())
+}
+
+/// Called by the drift detector when it detects a ciphertext drift for a
+/// given handle. Looks up the host chain block where the drifted computation
+/// originated and creates a revert signal so all services can coordinate.
+///
+/// Transient failures (DB errors) are logged and swallowed so they don't
+/// cascade into the drift detector — the next detection cycle will retry.
+pub async fn on_drift_detected(pool: &Pool<Postgres>, handle: &[u8], host_chain_id: i64) {
+    // Byte 21 is 0xff for compute outputs; inputs encode the ciphertext index
+    // in the proof (0x00..=0xfe). Input drift is out of scope for auto-recovery.
+    if handle.len() != 32 || handle[21] != 0xff {
+        warn!(
+            host_chain_id,
+            "Drifted handle is a ZK input; auto-recovery is not supported for input handles"
+        );
+        return;
+    }
+
+    let host_block: Option<i64> = match sqlx::query_scalar(
+        "SELECT block_number FROM computations \
+         WHERE output_handle = $1 AND host_chain_id = $2 \
+         LIMIT 1",
+    )
+    .bind(handle)
+    .bind(host_chain_id)
+    .fetch_optional(pool)
+    .await
+    {
+        Ok(row) => row,
+        Err(e) => {
+            error!(error = %e, host_chain_id, "Failed to look up computation for drifted handle");
+            return;
+        }
+    };
+
+    let Some(block) = host_block else {
+        error!(
+            host_chain_id,
+            "Cannot create revert signal: no computation found for drifted handle"
+        );
+        return;
+    };
+
+    match create_revert_signal(pool, host_chain_id, block).await {
+        Ok(Some(id)) => {
+            info!(
+                host_chain_id,
+                block, id, "Drift revert signal created successfully"
+            );
+        }
+        Ok(None) => {
+            warn!(
+                host_chain_id,
+                block, "Drift revert signal already in flight, skipping creation"
+            );
+        }
+        Err(e) => {
+            error!(error = %e, "Failed to create drift revert signal");
+        }
+    }
+}
+
+/// Poll `drift_revert_signal` until a signal is in flight (status Pending or
+/// Reverting). Used by `run_signal_watcher` in non-revert-runner services to
+/// detect that a drift revert is happening so they can re-exec.
+pub async fn wait_for_in_flight_signal(pool: &Pool<Postgres>) -> anyhow::Result<DriftRevertSignal> {
+    loop {
+        if let Some(signal) = latest_signal(pool).await? {
+            if signal.status.is_in_flight() {
+                return Ok(signal);
+            }
+        }
+        tokio::time::sleep(POLL_INTERVAL).await;
+    }
+}
+
+/// Poll until the latest signal reaches a terminal status.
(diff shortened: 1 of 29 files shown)


antoniupop:
Some issues reported by an agent review that look accurate to me so far:
  1. Multi-chain coordination gap: latest_signal() and handle_pending_signal_on_startup() only look at the single most-recent row (ORDER BY id DESC LIMIT 1). If chain A drifts first (signal S1 Pending), then chain B drifts before S1 is reverted (signal S2 Pending for a different chain_id — the dedupe guard is per-host_chain_id, so S2 is inserted), the runner will revert only S2 on its next startup. S1 stays Pending forever. Worse, create_revert_signal's guard then blocks any new signal for chain A, so the chain-A drift is uncorrectable by auto-recovery. This PR explicitly adds two-of-three-multi-chain.yaml, so multi-chain is in scope. Fix: iterate over all in-flight signals rather than only the latest, or loop until no pending/reverting signal remains.
  2. Silent watcher failure. In init_with_reexec, the background watcher is tokio::spawned, and if wait_for_in_flight_signal returns Err (e.g. transient DB loss), the task just logs and exits. The service keeps running but will miss every future drift signal. Watcher should retry on DB errors, or the service should die and let the orchestrator restart it.
  3. Failed-revert proceeds as if healthy. wait_for_revert_done maps Failed to Ok(()) and services continue startup. The comment says "operator can investigate via logs," but this means a coprocessor will happily resume serving over a DB that's now partially reverted. Consider at minimum marking the healthcheck unhealthy while a Failed row is the latest signal, so it's impossible to miss operationally.

dartdart26:
> Some issues reported by an agent review that look accurate to me so far:
> 
> 1. Multi-chain coordination gap: latest_signal() and handle_pending_signal_on_startup() only look at the single most-recent row (ORDER BY id DESC LIMIT 1). If chain A drifts first (signal S1 Pending), then chain B drifts before S1 is reverted (signal S2 Pending for a different chain_id — the dedupe guard is per-host_chain_id, so S2 is inserted), the runner will revert only S2 on its next startup. S1 stays Pending forever. Worse, create_revert_signal's guard then blocks any new signal for chain A, so the chain-A drift is uncorrectable by auto-recovery. This PR explicitly adds two-of-three-multi-chain.yaml, so multi-chain is in scope. Fix: iterate over all in-flight signals rather than only the latest, or loop until no pending/reverting signal remains.
> 2. Silent watcher failure. In init_with_reexec, the background watcher is tokio::spawned, and if wait_for_in_flight_signal returns Err (e.g. transient DB loss), the task just logs and exits. The service keeps running but will miss every future drift signal. Watcher should retry on DB errors, or the service should die and let the orchestrator restart it.
> 3. Failed-revert proceeds as if healthy. wait_for_revert_done maps Failed to Ok(()) and services continue startup. The comment says "operator can investigate via logs," but this means a coprocessor will happily resume serving over a DB that's now partially reverted. Consider at minimum marking the healthcheck unhealthy while a Failed row is the latest signal, so it's impossible to miss operationally.

1. I am fixing it.
2. I am fixing it.
3. Revert can't be partial, it is run in a DB txn. But if we make it unhealthy on a health-check with a row present that is failed - not ideal. Let me think about that.

dartdart26:
> > Some issues reported by an agent review that look accurate to me so far:
> > 
> > 1. Multi-chain coordination gap: latest_signal() and handle_pending_signal_on_startup() only look at the single most-recent row (ORDER BY id DESC LIMIT 1). If chain A drifts first (signal S1 Pending), then chain B drifts before S1 is reverted (signal S2 Pending for a different chain_id — the dedupe guard is per-host_chain_id, so S2 is inserted), the runner will revert only S2 on its next startup. S1 stays Pending forever. Worse, create_revert_signal's guard then blocks any new signal for chain A, so the chain-A drift is uncorrectable by auto-recovery. This PR explicitly adds two-of-three-multi-chain.yaml, so multi-chain is in scope. Fix: iterate over all in-flight signals rather than only the latest, or loop until no pending/reverting signal remains.
> > 2. Silent watcher failure. In init_with_reexec, the background watcher is tokio::spawned, and if wait_for_in_flight_signal returns Err (e.g. transient DB loss), the task just logs and exits. The service keeps running but will miss every future drift signal. Watcher should retry on DB errors, or the service should die and let the orchestrator restart it.
> > 3. Failed-revert proceeds as if healthy. wait_for_revert_done maps Failed to Ok(()) and services continue startup. The comment says "operator can investigate via logs," but this means a coprocessor will happily resume serving over a DB that's now partially reverted. Consider at minimum marking the healthcheck unhealthy while a Failed row is the latest signal, so it's impossible to miss operationally.
> 
> 1. I am fixing it.
> 2. I am fixing it.
> 3. Revert can't be partial, it is run in a DB txn. But if we make it unhealthy on a health-check with a row present that is failed - not ideal. Let me think about that.

For 3 - maybe we should stop altogether as if we don't, we're gonna miss the drift. Wdyt @antoniupop?

dartdart26:
Made drift auto revert signal creation off by default for prod, but on on E2E.

coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
```diff
@@ -0,0 +1,577 @@
...
+             SELECT 1 FROM drift_revert_signal \
+             WHERE host_chain_id = $1 AND (status = $3 OR status = $4) \
+         ) \
+         RETURNING id",
+    )
+    .bind(host_chain_id)
+    .bind(offending_host_block_number)
+    .bind(SignalStatus::Pending.as_db_str())
+    .bind(SignalStatus::Reverting.as_db_str())
+    .fetch_optional(pool)
```
antoniupop, line 158: Unless I missed something it looks like if the drift is not fixed by revert, this will loop forever. Once a revert is in Done state, a new revert to the same block is possible.

Not sure we want that even if indeed in such case manual intervention is unavoidable for now as this will impact other chains, so probably need to check on Done state as well and either say we never allow hititng twice same block or maybe have a backoff scheme.

coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
```diff
@@ -0,0 +1,577 @@
...
+                error!(
+                    signal_id = signal.id,
+                    host_chain_id = signal.host_chain_id,
+                    reason,
+                    "Drift revert transitioned to Failed while waiting"
+                );
+                anyhow::bail!("drift revert signal {} is Failed: {reason}", signal.id);
+            }
+        }
+        if oldest_in_flight_signal(pool).await?.is_none() {
```
antoniupop: I think this *could* be a race condition here, if "failed" is reached after previous check on latest_signal, but before this check here, then I think it breaks.
Safer would be to check latest_signal in this if block.

antoniupop (changes requested):

coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
dartdart26: I think not hitting the exact same block won't fix it as even if hit a slightly different block, we might still end up not fixing the revert.

I guess a backoff scheme might be good, but I don't want to overcomplicate it too much. Maybe we can have a global backoff, irrespective of which block we go back to? It is simpler, wdyt?

dartdart26: I've added a global check - if we have multiple reversals for a small period of time, we give up and require operator intervention.

My reasoning is that it is simpler. It does potentially waste some time doing the same reversal twice, but I am willing to accept that. Wdyt?

coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
dartdart26: Yeah, good catch, it is a race. I fixed it.

dartdart26:
E2E running here: https://github.com/zama-ai/fhevm/actions/runs/25056980713/job/73399946574

coprocessor/fhevm-engine/tfhe-worker/src/lib.rs
```diff
@@ -65,20 +64,11 @@ pub async fn async_main(
     let cancel_token = CancellationToken::new();
     info!(target: "async_main", args = ?args, "Starting runtime with args");
 
-    let database_url = resolve_database_url_from_option(args.database_url.clone())?;
-    let health_check = health_check::HealthCheck::new(database_url);
+    let database_url = args.database_url.clone().unwrap_or_default();
```
antoniupop: I think this is a regression post the IAM auth being merged: when DATABASE_IAM_AUTH_ENABLED is set but no URL is provided, this silently reads the DATABASE_URL env var so an IAM deploymetn with a misconf URL would now start up and try to connect with a default/empty URL instead of failing fast. 

I'd restore line 68 instead 
```
let database_url = resolve_database_url_from_option(args.database_url.clone())?; 
```

coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
```diff
@@ -0,0 +1,712 @@
...
+/// Primarily used by tests to swap in a mock.
+pub async fn init_with_reexec<R: ReExec + 'static>(
+    database_url: &str,
+    cancel_token: CancellationToken,
+    runner_cfg: Option<RevertRunnerConfig>,
+    re_exec: R,
+) -> anyhow::Result<()> {
+    let pool = PgPoolOptions::new()
+        .max_connections(1)
+        .connect(database_url)
```
antoniupop: Again this won't work with IAM - IAM-aware DB helpers need to be used. This bypasses the IAM token generation/refresh and the DB resolution rules, so auto-revert will fail in deployment using DATABASE_IAM_AUTH_ENABLED .

antoniupop:
Aside from comments above, I've re-run an agentic review and got two issues that look accurate to me:

 - High: coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs:258 derives the offending block with SELECT block_number ... LIMIT 1 for a handle, but the codebase already
    treats “same output handle appears in multiple computations” as a real case in coprocessor/fhevm-engine/tfhe-worker/src/tests/revert_coprocessor_db_state.rs:546. Without an ORDER
    BY, the chosen block is arbitrary, so the signal can point at the wrong host block and revert too far or not far enough.
  - High: same-chain drift signals do not converge to the earliest offending block. coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs:119 rejects any second in-flight
    signal for the same chain, and coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs:293 just drops later detections. But coprocessor/fhevm-engine/fhevm-engine-common/
    src/drift_revert.rs:394 assumes earlier out-of-order drifts will be caught later, while coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs:415 only replays still-open gateway
    handles after restart. If consensus for host block 120 is seen before the true earliest bad block 110, the branch can lock in a revert to 119 and never revisit the already-closed
    110 handle. That means it still does not reliably solve the stated minority-drift recovery problem

Eikix:
Trying to understand the approach. In the meantime, here are some (AI-generated) edge-case findings to look into apart from what Antoniu said:
- Concurrent signal creation can enqueue duplicate reverts -> e.g.  two gw-listener drift detectors observe the same chain drift at the same time (recommended solution -> Add a partial unique index so only one in-flight signal can exist per chain)
- Multiple runner instances can execute the same signal, e.g. during a Kubernetes rolling update, old and new gw-listener pods both read drift_revert_signal (recommended solution Claim the signal atomically before running revert SQL)

coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
dartdart26: Good catch! Let me fix it by accepting a pool from the services.

dartdart26:
> * coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs:258 derives the offending block with SELECT block_number ... LIMIT 1 for a handle, but the codebase already
>   treats “same output handle appears in multiple computations” as a real case in coprocessor/fhevm-engine/tfhe-worker/src/tests/revert_coprocessor_db_state.rs:546. Without an ORDER
>   BY, the chosen block is arbitrary, so the signal can point at the wrong host block and revert too far or not far enough.

Good catch, let me fix them.

dartdart26:
> Trying to understand the approach. In the meantime, here are some (AI-generated) edge-case findings to look into apart from what Antoniu said:
> 
> * Concurrent signal creation can enqueue duplicate reverts -> e.g.  two gw-listener drift detectors observe the same chain drift at the same time (recommended solution -> Add a partial unique index so only one in-flight signal can exist per chain)
> * Multiple runner instances can execute the same signal, e.g. during a Kubernetes rolling update, old and new gw-listener pods both read drift_revert_signal (recommended solution Claim the signal atomically before running revert SQL)

Do we run multiple gw-listeners? Should we cater for this case?

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -1064,13 +1057,40 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
...
-    tokio::spawn(async move { health_check_server.start().await });
+    tokio::spawn(async move {
+        if let Err(err) = health_check_server.start().await {
+            error!(error = %err, "Health check server failed");
+        }
+    });
+
+    // Drift-revert: must run before any DB state reads so we don't read
+    // pre-revert state.
+    drift_revert::init(
```
antoniupop: Here as well as in the similar place in the poller, you need to use a separate pool rather than the main one which can get closed on reconnect, which happens on insertion failure.

something like this should do:
```
  let (drift_revert_pool, _pool_refresh_handle) = connect_pool_with_options(
      &args.database_url,                                                                                                                                                               
      PgPoolOptions::new().max_connections(1),
      Some(&cancel_token),                                                                                                                                                              
  ).await?;                                                                                                                                                                             
  drift_revert::init(drift_revert_pool, cancel_token.clone(), None).await?;
```

coprocessor/fhevm-engine/host-listener/src/poller/mod.rs
```diff
@@ -133,6 +133,33 @@ pub async fn run_poller(config: PollerConfig) -> Result<()> {
...
+        config.health_port,
+        cancel_token.clone(),
+    );
+    tokio::spawn(async move {
+        if let Err(err) = health_check_server.start().await {
+            error!(error = %err, "Health check server failed");
+        }
+    });
+
+    fhevm_engine_common::drift_revert::init(
```
antoniupop, line 166: Here's the other

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
dartdart26: I would argue that we don't need the reconnect() function at all, the one where we close and replace the pool. AFAIU, the pool will try to reconnect itself. So the pattern looks weird to me. Any idea why we need it?

dartdart26: Anyway, I will leave potential removal of `reconnect()`  to another PR to not overcomplicate this one.

antoniupop (dismissed):

antoniupop (dismissed):

dartdart26:
@mergifyio queue

antoniupop (approved):

dartdart26:
@mergifyio queue

PR #2347, opened 2026-04-22
refactor(coprocessor): remove key generation listening from gw-listener

obatirou:
Part of https://github.com/zama-ai/fhevm-internal/issues/1230


obatirou:
@claude /pr-review

coprocessor/fhevm-engine/gw-listener/src/metrics.rs
```diff
@@ -49,54 +49,6 @@ pub(crate) static GET_LOGS_FAIL_COUNTER: LazyLock<IntCounter> = LazyLock::new(||
     .unwrap()
 });
 
-pub(crate) static ACTIVATE_CRS_SUCCESS_COUNTER: LazyLock<IntCounter> = LazyLock::new(|| {
```
rudy-6-4, line 52: Is there something to do on infra for this removal ?

coprocessor/fhevm-engine/Cargo.toml
```diff
@@ -40,7 +40,12 @@ anyhow = "1.0.98"
 aws-config = "1.8.5"
 aws-credential-types = "1.2.6"
 aws-sdk-kms = { version = "1.68.0", default-features = false }
-aws-sdk-s3 = { version = "1.103.0", features = ["test-util"] }
+aws-sdk-s3 = { version = "1.103.0", default-features = false, features = [
+    "default-https-client",
```
rudy-6-4, line 44: is it a subset of feature test-util ?

rudy-6-4 (approved):

coprocessor/fhevm-engine/gw-listener/src/metrics.rs
obatirou: Good question, I would say, once merge we will need to update the dashboards

coprocessor/fhevm-engine/Cargo.toml
obatirou: `test-utils` is removed now that gw-listener is removed.
But my understanding is that default of aws-sdk-s3 is the problem and pulls faulty rustls-webpki:
aws-sdk-s3 default pulls a rustls-webpki 0.101.7 so reducing features to a minimum to avoid this. Will check again. Not an expert at it, if you have any advice please share

obatirou: My understanding is that rustls from default feature is the legacy path and default-https-client use the modern rustls-aws-lc which removes the audit flagging

PR #2353, opened 2026-04-23
chore(coprocessor): upgrade tfhe-rs to 1.6.1 with disabled OPRF key

PanGan21:
## Summary

- Bumps `tfhe` `1.5.4→1.6.1`, `tfhe-versionable` `0.6.2→0.7.0`, and `tfhe-zk-pok` `0.8.0→0.8.1` in the coprocessor workspace.
- The dedicated OPRF key is not enabled yet

## Note

`kms-connector` and `sdk/rust-sdk` are **not upgraded** and remain on `tfhe = "=1.5.1"`. Upgrading those two workspaces requires a KMS release that supports tfhe 1.6.0.

Closes: https://github.com/zama-ai/fhevm-internal/issues/1289

### DIFF
--- a/coprocessor/fhevm-engine/tfhe-worker/src/tests/random.rs
+++ b/coprocessor/fhevm-engine/tfhe-worker/src/tests/random.rs
@@ -155,10 +155,12 @@ async fn test_fhe_random_bounded() -> Result<(), Box<dyn std::error::Error>> {
     let mut rand_types = Vec::new();
     let mut bounds = Vec::new();
 
-    // Per-type bounds matching the old gRPC test to avoid GPU edge cases.
+    // Type 1 uses bound=16 (full FheUint4 range) so it generates 4 random bits,
+    // matching the unbounded path and avoiding a small-range collision introduced
+    // by the tfhe-rs 1.6.x OPRF algorithm update.
     let type_bounds: &[(i32, &str)] = &[
         (0, "2"),
-        (1, "4"),
+        (1, "16"),
         (2, "128"),
         (3, "16384"),
         (4, "1073741824"),
@@ -180,7 +182,7 @@ async fn test_fhe_random_bounded() -> Result<(), Box<dyn std::error::Error>> {
         let tx_id = next_handle();
         let mut tx = harness.listener_db.new_transaction().await?;
 
-        // First sample with seed [1,0,...,0]
+        // First sample
         let output1 = next_handle();
         insert_event(
             &harness.listener_db,
@@ -190,7 +192,7 @@ async fn test_fhe_random_bounded() -> Result<(), Box<dyn std::error::Error>> {
                 caller: zero_address(),
                 upperBound: as_scalar_uint(&bound),
                 randType: to_ty(rand_type),
-                seed: FixedBytes::from([1_u8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
+                seed: FixedBytes::from([0_u8; 16]),
                 result: output1,
             }),
             true,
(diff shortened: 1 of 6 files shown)


coprocessor/fhevm-engine/tfhe-worker/src/tests/random.rs
```diff
@@ -138,15 +138,9 @@ async fn test_fhe_random_basic() -> Result<(), Box<dyn std::error::Error>> {
...
-///
-/// Uses per-type bounds that match the old gRPC test to avoid edge cases
-/// (e.g. upper_bound=1 produces 0 random bits, which behaves differently
-/// on GPU).
+/// and that different seeds yield different results for non-bool types.
+/// Bool is excluded because FheBool::generate_oblivious_pseudo_random
+/// produces the same plaintext for all seeds with a given key.
 #[tokio::test]
 #[serial(db)]
 async fn test_fhe_random_bounded() -> Result<(), Box<dyn std::error::Error>> {
```
PanGan21, line 152: In my understanding the `OprfServerKey` introduced in 1.6.0 is a different key from the PBS bootstrapping key, so it produces different outputs for the same seeds. With upper_bound=4 for `FheUint4` there are only 4 possible values, and seeds 1 and 7 both happened to map to 0 with the new key.

The fix here use seeds 0 and 42 (proven to differ in `test_fhe_random_basic`) and set the type 1 bound to 16 (full FheUint4 range) so the bounded path generates the same 4 random bits as the unbounded path.

Hopefully this is correct rationale for the reason that this test was failing after the upgrade.

PanGan21 (commented):
[AUTOMATED]

## Code Review

No blocking issues found. Checked for bugs, guideline compliance, and error handling.

**What was reviewed:**
- Correctness of `oprf_key` threading through all three `ServerKey::into_raw_parts()` / `from_raw_parts()` call sites (`keys.rs`, `gw-listener/sks_key.rs`, `tfhe-worker/utils.rs`) — all consistent ✓
- `use_dedicated_oprf_key(true)` placement in the key generation config — correct ✓
- Test changes in `random.rs`: seed `[0; 16]` is a valid PRF input with no degenerate behavior; bound=16 for `FheUint4` avoids the small-range collision documented in the inline comment ✓
- The removed misleading comment `// rerandomisation keyswitching key excluded` was factually wrong (the key IS included) — removing it is a fix ✓
- Version mismatch between coprocessor (1.6.0) and kms-connector/sdk (1.5.1) is acknowledged in the PR; tfhe-rs handles it gracefully via backward-compatibility version upgrading (`oprf_key: None` for old keys, falling back to PBS key for fheRand) ✓

**One informational note (non-blocking):** The dedicated OPRF key security isolation only takes effect for keys freshly generated with the new 1.6.0 config. Keys deserialized from pre-1.6.0 format will transparently fall back to using the bootstrapping key for fheRand — correct behavior but worth confirming in the rollout plan that key regeneration is part of the deployment.

coprocessor/fhevm-engine/tfhe-worker/src/tests/random.rs
```diff
@@ -155,10 +155,12 @@ async fn test_fhe_random_bounded() -> Result<(), Box<dyn std::error::Error>> {
     let mut rand_types = Vec::new();
     let mut bounds = Vec::new();
 
-    // Per-type bounds matching the old gRPC test to avoid GPU edge cases.
+    // Type 1 uses bound=16 (full FheUint4 range) so it generates 4 random bits,
+    // matching the unbounded path and avoiding a small-range collision introduced
+    // by the tfhe-rs 1.6.x OPRF algorithm update.
```
antoniupop: Out of the context of this PR, this comment would be strange in the code IMO.

coprocessor/fhevm-engine/tfhe-worker/src/tests/random.rs
antoniupop: I think htat makes sense, yes

antoniupop (dismissed):
LGTM

coprocessor/fhevm-engine/tfhe-worker/src/tests/random.rs
PanGan21: reverted

antoniupop (approved):

PanGan21:
@mergify queue

PR #2365, opened 2026-04-24
fix(coprocessor): reduce 0.12 migration lock time

Eikix:
## What

Backport a two-phase 0.12 coprocessor DB migration path for large testnet-sized databases.

Changes:
- add an explicit online phase to `/initialize_db.sh` behind `RUN_MIGRATIONS_UNTIL_REMOVE_TENANTS=true`
- in that online phase, run SQLx migrations up to the migration immediately before `remove_tenants`, then create the seven tenant-free indexes one by one with `CREATE UNIQUE INDEX CONCURRENTLY`
- keep `20260128095635_remove_tenants.sql` as the blocking/downtime migration boundary
- keep the same index definitions in `remove_tenants` as `CREATE UNIQUE INDEX IF NOT EXISTS`, so the downtime migration is self-contained and skips the pre-created indexes
- keep the metadata-only constant-default rewrite for `host_chain_id` / `key_id_gw`
- avoid one SQLx migration per index and avoid a separate pre-index script
- include the same behavior in the workspace db-migration image used by local/fhevm-cli builds
- keep dependency fixes needed by CI on this backport branch

## Why

The vanilla 0.12.1 migration is too slow on the testnet-sized snapshot. On the debug DB it took about 4h13m total, with `20260128095635_remove_tenants` taking about 4h10m.

The first optimization, without online pre-indexing, reduced the same migration to about 1h10m, but that is still above the target downtime window.

The final approach splits the rollout operationally:

1. Online phase, while 0.11 services are still running: build only the indexes that can be built concurrently.
2. Downtime phase: stop services and run the blocking `remove_tenants` migration plus the remaining 0.12 migrations.

This is intentionally a one-time operational split for the 0.11 -> 0.12 testnet upgrade. After `remove_tenants` has run, future releases will start from the post-tenant schema and the `RUN_MIGRATIONS_UNTIL_REMOVE_TENANTS` flag should not be used.

## Operational Flow

Before either phase, check for stale transactions:

```sql
SELECT pid, application_name, client_addr, state,
       now() - xact_start AS xact_age,
       left(query, 200) AS query
FROM pg_stat_activity
WHERE datname = 'coprocessor'
  AND xact_start IS NOT NULL
ORDER BY xact_start;
```

Phase 1, online on 0.11:

```bash
RUN_MIGRATIONS_UNTIL_REMOVE_TENANTS=true /initialize_db.sh
```

This does two things:

1. ensures SQLx migrations are applied through `20260120102002_unused_index_cleaning.sql`
2. creates these indexes with `CREATE UNIQUE INDEX CONCURRENTLY`, one by one:
   - `idx_allowed_handles_no_tenant`
   - `idx_input_blobs_no_tenant`
   - `idx_ciphertext_digest_no_tenant`
   - `idx_ciphertexts_no_tenant`
   - `idx_ciphertexts128_no_tenant`
   - `idx_computations_no_tenant`
   - `idx_pbs_computations_no_tenant`

It deliberately does not mark `remove_tenants` as applied.

Phase 2, during maintenance:

```bash
/initialize_db.sh
```

This resumes at `20260128095635_remove_tenants.sql`. The pre-created indexes are skipped by `CREATE UNIQUE INDEX IF NOT EXISTS`, and the blocking schema changes happen while the services are stopped.

## Assumptions

- Testnet has one tenant and no multichain data today.
- The seven tenant-free indexes are backward-compatible with 0.11 writes; they enforce uniqueness that should already be true for single-tenant data.
- `CREATE UNIQUE INDEX CONCURRENTLY` is acceptable while 0.11 services are running. It avoids long write-blocking table locks, but it can still wait behind old transactions.
- Operators clear stale transactions before each phase. We observed an idle IntelliJ/DataGrip transaction block `host_chain_blocks_valid`, which invalidated one timing run.
- If a concurrent index build is interrupted and leaves an invalid index relation, operators should drop that invalid index and rerun Phase 1.

## Current Testnet Debug Data

Known timings:

| Run | Result |
| --- | --- |
| Vanilla v0.12.1 | succeeded in ~4h13m total; `remove_tenants` ~4h10m |
| Lean monolithic migration | succeeded in ~1h10m total; `remove_tenants` ~1h09m |
| Online concurrent index phase, one index at a time | succeeded in 27m17s |
| Downtime phase after online indexing | succeeded in 1m30.83s total; `remove_tenants` 29.46s |

The earlier post-preindex timing that reported ~34m was invalid because an external IntelliJ/DataGrip transaction blocked `host_chain_blocks_valid` for part of the run.

## Caveats

- This still edits SQLx migration `20260128095635`. Any environment that already applied the previous 0.12.x version will hit a SQLx checksum mismatch unless it is reset/rebased or handled explicitly.
- If Phase 1 is skipped on a large DB, `remove_tenants` remains correct but will build the large indexes during downtime.
- This flag is not a general migration framework. It exists for the 0.11 -> 0.12 tenant-removal upgrade and should become unused after affected environments have crossed that boundary.
- The `NOT VALID` checks are intentionally not validated in this backport to avoid extra table scans during the upgrade path.

## Validation

- Built db-migration image locally.
- Verified `RUN_MIGRATIONS_UNTIL_REMOVE_TENANTS=true /initialize_db.sh` applies migrations up to `20260120102002`, creates all seven indexes concurrently, and does not apply `remove_tenants` on local Postgres.
- Verified a second `/initialize_db.sh` resumes at `remove_tenants` and completes on local Postgres.
- Verified a full `/initialize_db.sh` from scratch completes on local Postgres.
- `SQLX_OFFLINE=true cargo test -p tfhe-worker migrations -- --nocapture` passed: 3 tests.
- Debug DB timings listed above.


goshawk-3 (approved):
LGTM

coprocessor/fhevm-engine/db-migration/initialize_db.sh
```diff
@@ -17,12 +17,86 @@ else
...
+    END")
+
+  case "$index_state" in
+    valid)
+      echo "Skipping valid pre-created index $index_name"
+      ;;
+    missing)
+      echo "Creating pre-upgrade index $index_name"
+      run_sql "$create_sql"
+      ;;
```
goshawk-3, line 59: nitpick: maybe we can print the elapsed time here too.

Eikix: true, i assumed logs in infra stack have timestamp

PR #2366, opened 2026-04-24
fix(coprocessor): address coprocessor chart issues

deasydoesit:
This PR addresses 5 issues:

1) `DATABASE_URL` incompatibility
- Problem: `commonConfig.databaseUrl` was rendered into `DATABASE_URL` as the first env var listed, before `DATABASE_USER`/`PASSWORD`/`ENDPOINT` envs it references, so variable expansion failed and the `commonConfig.databaseUrl` value was useless. Component-specific escape hatch `env` could have been used to "solve" the issue, but not cleanly for a new Chart release.
- Solution: Add `DATABASE_USER`/`PASSWORD`/`ENDPOINT`  envs applied before  `DATABASE_URL` so variable expansion succeeds
2) `S3_BUCKET_NAME` incompatibility 
- Problem: `snsWorker.config.s3BucketName` previously only accepted a raw string value, which was incompatible with valueFrom format used in live environments. Again, component-specific escape hatch `env` could have been used to "solve" the issue.
- Solution: update `snsWorker.config.s3BucketName` to accept both value and valueFrom (e.g., configmap, secret).
3) `arg` brittleness 
- Problem: previous chart version had args defined in the manifest templates and relied on Helm interpolation to inject from `values.yaml` into the `arg` defintion in k8s deployments (e.g., `--log-level={{ $.Values.hostListenerShared.args.logLevel }}`). This tightly coupled the chart to binary-specific args, which made coprocessor upgrades extremely fragile and created immense maintenance overhead. Basically, the args were hardcoded into the chart, but if any of the args were no longer accepted by the underlying coprocessor binary, it would cause CrashLoopBackOff with no way to fix it.
- Solution: kept only chart-owned inline args, meaning flags derived from env-backed values or other structured `values.yaml` fields. We made those inline args conditional so they render only when their backing config exists, moved all other runtime flags out to `extraArgs`, and added `disabledArgs` as an escape hatch so consumers can suppress any template-injected arg and replace it themselves.
4) `TENANT_API_KEY` incompatibility similar to (1) and (2)
5) Adds more env vars into `values.yaml` schema


charts/coprocessor/templates/_helpers.tpl
deasydoesit, line 1: Didn't like the `coprocessor.renderChainEnvValue` helper. Added confusion both here and where it was used in the deployment manifests

piizama (dismissed):

eudelins-zama (dismissed):

eudelins-zama (approved):

piizama (approved):

deasydoesit:
@mergify queue

deasydoesit:
@mergify queue

PR #2368, opened 2026-04-27
fix(library-solidity): fix codegen to regenerate host contracts idempotently

PanGan21:
### Summary:

Running `npm run codegen` in host-contracts was producing a diff against the committed contracts because templates were not updated. This pr updates the template to regenerate the same host contracts idempotently.


obatirou (approved):

PanGan21:
@mergify queue

PR #2370, opened 2026-04-27
feat(common): add docker build image for listener

maxnovawind:


.github/workflows/listener-docker-build.yml
```diff
@@ -0,0 +1,167 @@
...
+          - listener/Dockerfile.ci
+          - listener/Cargo.toml
+          - listener/Cargo.lock
+          - listener/crates/**
+          - listener/rust-toolchain.toml
+
+  ########################################################################
+  #                        BUILD DECISIONS                               #
+  # Centralizes all build/re-tag logic in one place for maintainability  #
+  ########################################################################
```
eudelins-zama: AFAIU, the listener is a single service, so no need for the build decision pattern, this is for multi-services build.

You can take `host-contracts-docker-build.yml` as an example. It should be documented in the `.github/CONTRIBUTING.md` as well.

maxnovawind: Done

nboisde (dismissed):
LGTM

nboisde (dismissed):

Eikix:
I recommend addding the label `e2e orchestrated` which will launch the orchestrated workflow from this PR without waiting for the merge queue. Will save you a lot of time if something is off.

Eikix (changes requested):
The `listener-core-version` output is currently not consumed by the e2e stack.

This PR adds `listener-docker-build` to `test-suite-orchestrate-e2e-tests.yml` and computes `listener-core-version`, but the value stops in that workflow: `run-e2e-tests` does not pass it to `test-suite-e2e-tests.yml`, the reusable e2e workflow has no matching input/env var, and `test-suite/fhevm` has no `LISTENER_CORE_VERSION` image override.

So the listener image may be built, but orchestrated e2e will not boot that built image.

If the goal is to validate this listener image in e2e, the missing path is:

1. Add a `listener-core-version` input to `.github/workflows/test-suite-e2e-tests.yml`.
2. Export it as an env var, for example `LISTENER_CORE_VERSION`.
3. Teach `test-suite/fhevm` to use that env var in the generated compose/service for the listener image.
4. Pass `listener-core-version` from `test-suite-orchestrate-e2e-tests.yml` into the reusable e2e workflow.

If the goal is only to build/publish the listener image for now, then the e2e-orchestrator output should probably be removed to avoid implying e2e coverage that is not actually happening.

Eikix:
EDIT: It seems the new listener is not yet integrated in the coprocessor flow, so if runtime integration is intentionally coming in a later PR, I think a short comment should make clear that the e2e-specific `listener-core-version` is build-only coverage. When the listener replaces/backs the coprocessor host listener, the follow-up should add the full path: reusable e2e input, CLI lock/default env, generated compose service, and readiness/log handling

once you confirm this I can approve the PR too

Eikix (dismissed):

Eikix (approved):

maxnovawind:
@mergify queue

rudy-6-4 (approved):

rudy-6-4:
@mergifyio queue

PR #2372, opened 2026-04-27
feat(common): add workflow to verify host-contracts and library-solidity codegen is up-to-date

PanGan21:
Closes: https://github.com/zama-ai/fhevm-internal/issues/1320


obatirou (dismissed):

eudelins-zama (commented):
I don't have the full context here, but could this be integrated in an already existing host-contracts workflow, instead of creating a new one?

Just to avoid expanding the number of workflows if possible.

If what I say does not makes sense, ignore this 😇

PanGan21:
> I don't have the full context here, but could this be integrated in an already existing host-contracts workflow, instead of creating a new one?
> 
> Just to avoid expanding the number of workflows if possible.
> 
> If what I say does not makes sense, ignore this 😇

It does make sense however the only place this could be added is the `host-contracts-integrity-checks` workflow which is only relevant to host contracts and bypasses the library-solidity. Then we will need an additional workflow for library-solidity checks as well..

eudelins-zama:
> It does make sense however the only place this could be added is the host-contracts-integrity-checks workflow which is only relevant to host contracts and bypasses the library-solidity. Then we will need an additional workflow for library-solidity checks as well...

I don't think it would be required to have an additional workflow for library-solidity check :slightly_smiling_face: 

Because you could have everything in the `host-contracts-integrity-checks`, just by expanding the `check-changes` and having multiple "targets", like `codegen-changes` and `host-contracts-only-changes`.
Then you can choose what to run depending on which target is `true`.

Anyway, it's not a blocker to create a dedicated workflow, just sharing ideas and tips :wink:

PanGan21:
> > It does make sense however the only place this could be added is the host-contracts-integrity-checks workflow which is only relevant to host contracts and bypasses the library-solidity. Then we will need an additional workflow for library-solidity checks as well...
> 
> I don't think it would be required to have an additional workflow for library-solidity check 🙂
> 
> Because you could have everything in the `host-contracts-integrity-checks`, just by expanding the `check-changes` and having multiple "targets", like `codegen-changes` and `host-contracts-only-changes`. Then you can choose what to run depending on which target is `true`.
> 
> Anyway, it's not a blocker to create a dedicated workflow, just sharing ideas and tips 😉

No problem, added in `host-contracts-integrity-checks` as suggested

obatirou (approved):

PanGan21:
@mergify queue

PR #2320, opened 2026-04-17
feat(test-suite): point keygen bootstrap to host

isaacdecoded:
E2E passed https://github.com/zama-ai/fhevm/actions/runs/25172003163/job/73793714571

Closes https://github.com/zama-ai/fhevm-internal/issues/1261

PR #2381, opened 2026-04-27
feat(coprocessor): host-listener, migrate KMS activation handling

rudy-6-4:


coprocessor/fhevm-engine/db-migration/migrations/20260421160000_keys_crs_status.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE keys DROP CONSTRAINT unique_key_id_gw;
+ALTER TABLE keys DROP CONSTRAINT unique_key_id;
+
+-- add extra-info
+ALTER TABLE keys
```
rudy-6-4, line 5: the chain_id need to be propagated for old rows

rudy-6-4, line 5: There are missing IF. TO be able to replay the file.

obatirou: Should be done before merge?

coprocessor/fhevm-engine/host-listener/src/cmd/mod.rs
```diff
@@ -1117,6 +1144,10 @@ pub async fn main(args: Args) -> anyhow::Result<()> {
                 )
                 .await;
             }
+            background_tasks.spawn(process_kms_generation_activations(
```
obatirou, line 1147: This is not done in Poller

coprocessor/fhevm-engine/host-listener/src/poller/mod.rs
obatirou, line 118: kms address is not passed

charts/coprocessor/templates/coprocessor-host-listener-deployment.yaml
```diff
@@ -55,6 +55,9 @@ spec:
             - --url=$(ETHEREUM_RPC_URL)
             - --acl-contract-address=$(ACL_CONTRACT_ADDRESS)
             - --tfhe-contract-address=$(FHEVM_EXECUTOR_CONTRACT_ADDRESS)
+            {{- if .Values.commonConfig.kmsGenerationContractAddress }}
+            - --kms-generation-address=$(KMS_GENERATION_ADDRESS)
```
obatirou, line 64: should be added to poller also

test-suite/fhevm/src/flow/up-flow.ts
obatirou: Should be host now?

obatirou: You might need to rebase on feat/RFC013, host-contract docker should have those

obatirou: In fact no, test-suite/fhevm/docker-compose/host-sc-docker-compose.yml should be updated
@isaacdecoded is already on this with https://github.com/zama-ai/fhevm/pull/2320

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+            );
+            let server_key_sns = server_key
+                .as_ref()
+                .map(|bytes| extract_server_key_without_ns(bytes))
+                .transpose()?;
+            Ok(set_ready_key_activation(
+                tx,
+                activation,
+                server_key,
+                server_key_sns,
```
obatirou: I am not sure about this: `server_key_sns` is `without ns`

https://github.com/zama-ai/fhevm/blob/4bfa7e8d867e490289523d3fa16faa7eefea477f/coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs#L652-L653

rudy-6-4: Yep it's the reverse I think, I'll recheck.

test-suite/fhevm/src/flow/up-flow.ts
rudy-6-4: Yes it's a tmp just to see E2E progress that will be removed.

coprocessor/fhevm-engine/db-migration/migrations/20260421160000_keys_crs_status.sql
rudy-6-4: Yes

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -0,0 +1,591 @@
...
+        FROM kms_key_activation_events
+        WHERE status = 'pending'
+        "#,
+    )
+    .fetch_one(db_pool)
+    .await?;
+    Ok(row.unwrap_or(0) as u64)
+}
+
+pub(crate) async fn count_crs_activation_remaining_pending(
```
obatirou, line 139: nit: could be factorized with `count_key_activation_remaining_pending`?
Not a fan of macro though so maybe should stay as is.
Same remark for `cancel_orphaned_*_activations`

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+    // rows are locked so there's no double work
+    let mut tx = db_pool.begin().await?;
+    let key_activations =
+        all_pending_key_activations_to_download(&mut tx).await?;
+    let crs_activations =
+        all_pending_crs_activations_to_download(&mut tx).await?;
+    if key_activations.is_empty() && crs_activations.is_empty() {
+        info!("No pending KMSGeneration activation to download");
+        return Ok(0);
+    }
```
obatirou, line 167: in case of panic, we have no observability on this. 
Tasks are aborted in the main loop
```rust
background_tasks.abort_all();
while background_tasks.join_next().await.is_some() {}
```

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+    download_and_store_key_activations(&mut tx, &s3_client, key_activations)
+        .await?;
+    download_and_store_crs_activations(&mut tx, &s3_client, crs_activations)
+        .await?;
+    info!("Downloading succeeded for KMSGeneration and CRS");
+    tx.commit().await?;
+    let remain_pending = count_key_activation_remaining_pending(&db_pool)
+        .await?
+        + count_crs_activation_remaining_pending(&db_pool).await?;
+    Ok(remain_pending)
```
obatirou, line 186: Same here no observability

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
rudy-6-4: Ok I have reuse sks_key and sns_pk naming to be sure to stick to what was done before.

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
rudy-6-4: replaced by logging error

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
rudy-6-4: To be rechecked as I had the reverse intuition on keys about which was doing what.

coprocessor/fhevm-engine/db-migration/migrations/20260421160000_keys_crs_status.sql
rudy-6-4: done

coprocessor/fhevm-engine/db-migration/migrations/20260421160000_keys_crs_status.sql
```diff
@@ -0,0 +1,57 @@
...
+        UPDATE keys
+        SET chain_id = hcid
+        WHERE chain_id IS NULL;
+    END IF;
+END $$;
+
+-- Guarantees idempotency of live ingest / catchup replays: the same
+-- (chain_id, block_hash, key_id) cannot be inserted twice.
+CREATE UNIQUE INDEX IF NOT EXISTS uniq_keys_chain_block_id_gw
+    ON keys (chain_id, block_hash, key_id);
```
obatirou: name says id_gw but indexed column is `key_id` and not `key_id_gw`.
Uniqueness of `key_id_gw` is dropped above which can be a problem in `activate_ready_key_activations` as the filter is only `key_id_gw` for UPDATE keys AS k.
Even if we should not have duplicated rows by `key_id_gw` in keys something feels off here.

obatirou: also chain_id is filled up but block_hash is not. Chain id column should be changed to NOT_NULL after the fill and the index uniq_keys_chain_block_id_gw should be NULLS NOT DISTINCT ?

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -0,0 +1,601 @@
...
+                sns_pk    = COALESCE(EXCLUDED.sns_pk, keys.sns_pk),
+                key_id_gw = EXCLUDED.key_id_gw
+            "#,
+            chain_id,
+            &block_hash,
+            &key_id
+        )
+        .execute(tx.deref_mut())
+        .await?;
+        if update_result.rows_affected() == 0 {
```
obatirou, line 265: is this possible ?
`update_result` is not an update anymore but an insert so name should be changed.

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -0,0 +1,601 @@
...
+pub(crate) async fn cancel_orphaned_key_activations(
+    tx: &mut Transaction<'_, Postgres>,
+) -> anyhow::Result<u64> {
+    let query = sqlx::query!(
+        "
+        UPDATE kms_key_activation_events AS e
+        SET status = 'cancelled'
+        FROM host_chain_blocks_valid AS b
+        WHERE
+            e.status = 'pending'
```
obatirou: Why not update ready also?
and delete corresponding already downloaded materials

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -0,0 +1,601 @@
...
+    }
+    Ok(result)
+}
+
+pub async fn set_ready_key_activation(
+    tx: &mut Transaction<'_, Postgres>,
+    activation: &PendingKeyActivation,
+    sns_pk: Option<Vec<u8>>,
+    sks_key: Option<Vec<u8>>,
+    public_key: Option<Vec<u8>>,
```
obatirou, line 493: Why have option here?
We should have the 3 available or do nothing 
Should be checked in `download_and_store_key_activation`?

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -0,0 +1,601 @@
...
+
+pub async fn mark_key_activation_error(
+    tx: &mut Transaction<'_, Postgres>,
+    error_msg: &str,
+    activation: PendingKeyActivation,
+) {
+    if let Err(err) = sqlx::query!(
+        r#"
+        UPDATE kms_key_activation_events
+        SET last_error = $1, last_updated_at = NOW()
```
obatirou: no update of the status?
I do not see any update of retry_count also, should this be kept in the schema?

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -0,0 +1,601 @@
...
+pub(crate) async fn cancel_orphaned_crs_activations(
+    tx: &mut Transaction<'_, Postgres>,
+) -> anyhow::Result<u64> {
+    let query = sqlx::query!(
+        "
+        UPDATE kms_crs_activation_events AS e
+        SET status = 'cancelled'
+        FROM host_chain_blocks_valid AS b
+        WHERE
+            e.status = 'pending'
```
obatirou: same here

coprocessor/fhevm-engine/db-migration/migrations/20260421160000_keys_crs_status.sql
rudy-6-4: yes but in the next release to facilitate downgrade ?

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
rudy-6-4: true

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
rudy-6-4: yes but the 2 first one are sync but the event has an array so could be only public or server.

rudy-6-4: asking for confirmation on guild

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
rudy-6-4: if the select fails, the rows where locked but who knows.
i will remove the continue so it's only logged.

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -0,0 +1,601 @@
...
+    )
+    .execute(tx.deref_mut())
+    .await?;
+    if query.rows_affected() > 0 {
+        info!("Marked {} pending CRS activations as cancelled due to orphaned blocks", query.rows_affected());
+    }
+    Ok(query.rows_affected())
+}
+
+pub(crate) async fn activate_ready_key_activations(
```
obatirou, line 200: could be simplified into only one query: 
```
WITH ready AS (
  <to_activate_statement>
), 
upserted AS (
   <INSERT INTO keys>
),
UPDATE kms_key_activation_events AS e
...
```

obatirou: same for activate_ready_crs_activations

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+        "Received ActivateKey event"
+    );
+    let key_path =
+        format!("{}/{}", to_key_prefix(key_type), key_id_to_aws_key(key_id));
+    let s3_client = s3_client.clone();
+    let Ok(bytes) =
+        download_key_from_s3(&s3_client, &activation.storage_urls, key_path, 0)
+            .await
+    else {
+        ACTIVATE_KEY_FAIL_COUNTER.inc();
```
obatirou: it is already done in download_and_store_key_activations?

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+            let sks_key = server_key
+                .as_ref()
+                .map(|bytes| extract_server_key_without_ns(bytes))
+                .transpose()?;
+            Ok(set_ready_key_activation(
+                tx, activation, server_key, sks_key, public_key,
+            )
+            .await?)
+        }
+        (Err(err), _) | (_, Err(err)) => anyhow::bail!(err),
```
obatirou, line 265: if both error, only the first one is returned?

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+    tx: &mut Transaction<'_, Postgres>,
+    s3_client: &A,
+    activation: &PendingKeyActivation,
+) -> anyhow::Result<()> {
+    let server_key =
+        download_key_activation(activation, KeyType::ServerKey, s3_client)
+            .await;
+    let public_key =
+        download_key_activation(activation, KeyType::PublicKey, s3_client)
+            .await;
```
obatirou, line 247: Could be downloaded in parrallel?

rudy-6-4: yes  it could but I have kept it simple.  it's not a frequent event. Are public key big ?
let make it work and improve perf after

coprocessor/fhevm-engine/db-migration/migrations/20260421160000_keys_crs_status.sql
obatirou: not sure to understand your answer regarding the 2 comments

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
obatirou: I do not know about the size

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
rudy-6-4: half done, delete is not done

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
rudy-6-4: I have let the continue so that if this happen we don't set as activated the staging table.

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -584,7 +584,7 @@ pub async fn mark_crs_activation_error(
     if let Err(err) = sqlx::query!(
         r#"
         UPDATE kms_crs_activation_events
-        SET last_error = $1, last_updated_at = NOW()
+        SET last_error = $1, last_updated_at = NOW(), retry_count = COALESCE(retry_count, 0) + 1
```
obatirou, line 587: when is status set to error?

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
rudy-6-4: The first one if two exists.

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
rudy-6-4: never in current state of code. error is only for final  error, so it should be done after some retry and some time (in case we want a window of manual fix ( eg the key was uploaded in the wrong format ...).
Any suggestion ?

coprocessor/fhevm-engine/db-migration/migrations/20260421160000_keys_crs_status.sql
rudy-6-4: Sorry. I have changed the index to be on key_id_gw.
If we have NOT NULL now, it can break on downgrade ?

obatirou: For downgrade I do not know. Having an alter table after the fill is problematic?
I think it possible, in remove tenant there was some alter table to specify not null https://github.com/zama-ai/fhevm/blob/84709a2befd361db750a07f766ffa062a15a0a27/coprocessor/fhevm-engine/db-migration/migrations/20260128095635_remove_tenants.sql#L117

rudy-6-4: If the field is new and you remove the default value (here NULL), then you cannot downgrade without adding the default value again.

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
obatirou: if understanding correctly you want a suggestion for manual fix window: 48H seems good?

coprocessor/fhevm-engine/db-migration/migrations/20260421160000_keys_crs_status.sql
```diff
@@ -0,0 +1,57 @@
...
+        UPDATE keys
+        SET chain_id = hcid
+        WHERE chain_id IS NULL;
+    END IF;
+END $$;
+
+-- Guarantees idempotency of live ingest / catchup replays: the same
+-- (chain_id, block_hash, key_id) cannot be inserted twice.
+CREATE UNIQUE INDEX IF NOT EXISTS uniq_keys_chain_block_id_gw
+    ON keys (chain_id, block_hash, key_id_gw);
```
obatirou, line 31:
```suggestion
-- (chain_id, block_hash, key_id_gw) cannot be inserted twice.
CREATE UNIQUE INDEX IF NOT EXISTS uniq_keys_chain_block_id_gw
    ON keys (chain_id, block_hash, key_id_gw);
```
key_id_gw does not makes sense anymore now we are on ethereum.
Also what is the difference between key_id and key_id_gw?

obatirou: According to the sql migration file 
* key_id: ID from server-key metadata, used in ciphertext metadata.
* key_id_gw: ID from the Gateway event, which could differ.

This is not a Gateway event anymore

obatirou (approved):

PR #2383, opened 2026-04-28
feat(coprocessor): re-randomise input ciphertexts before expand

antoniupop:
BREAKING CHANGE: this PR moves re-randomisation of input ciphertexts from before compression to before expand. This means that the ciphertexts before and after will not be bitwise equal for the same input, which would break consensus if upgrade rollout is not synchronised.

### DIFF
--- a/coprocessor/fhevm-engine/zkproof-worker/src/tests/mod.rs
+++ b/coprocessor/fhevm-engine/zkproof-worker/src/tests/mod.rs
@@ -129,7 +129,7 @@ async fn test_max_input_index() {
 
 #[tokio::test]
 #[serial(db)]
-async fn test_verify_proof_rerandomises_ciphertexts_before_storage() {
+async fn test_verify_proof_rerandomises_compact_list_before_expansion() {
     let (pool_mngr, _instance) = utils::setup().await.expect("valid setup");
     let pool = pool_mngr.pool();
 
@@ -192,6 +192,17 @@ async fn test_verify_proof_rerandomises_ciphertexts_before_storage() {
             .zip(&baseline)
             .all(|(stored_ct, baseline_ct)| stored_ct.ciphertext != *baseline_ct),
         "stored ciphertexts should differ from the pre-rerandomization compression"
+    );
+
+    let expected = utils::compress_inputs_with_compact_list_rerandomization(&pool, &zk_pok, &aux.0)
+        .await
+        .unwrap();
+    assert_eq!(
+        stored
+            .iter()
+            .map(|ct| ct.ciphertext.clone())
+            .collect::<Vec<_>>(),
+        expected
     );
 
     let decrypted = utils::decrypt_ciphertexts(&pool, &handles).await.unwrap();
--- a/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
+++ b/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
@@ -16,6 +16,7 @@ use sqlx::{postgres::PgListener, PgPool};
 use sqlx::{Postgres, Transaction};
 use std::str::FromStr;
 use tfhe::integer::ciphertext::IntegerProvenCompactCiphertextListConformanceParams;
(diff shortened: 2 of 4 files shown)


coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
```diff
@@ -501,16 +498,29 @@ fn verify_proof_only(
...
     request_id: i64,
     the_list: &tfhe::ProvenCompactCiphertextList,
+    blob_hash: &[u8],
+    cpk: &tfhe::CompactPublicKey,
 ) -> Result<Vec<SupportedFheCiphertexts>, ExecutionError> {
     if the_list.is_empty() {
         return Ok(vec![]);
     }
 
+    let mut re_rand_context = ReRandomizationContext::new(
```
PanGan21, line 520: Is it correct that we don't use the aux data anymore for rerandomization?

antoniupop: Yes, we don't need the metadata anymore.

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
```diff
@@ -414,27 +414,24 @@ pub(crate) fn verify_proof(
...
-
-    // Step 3: Compute blob hash and set re-randomization metadata on all ciphertexts
+    // Step 2: Compute blob hash used to seed the compact-list re-randomization
     let mut h = Keccak256::new();
     h.update(RAW_CT_HASH_DOMAIN_SEPARATOR);
     h.update(raw_ct);
     let blob_hash = h.finalize().to_vec();
 
+    // Step 3: Re-randomize the verified compact list, then expand
+    let mut cts = rerand_and_expand_verified_list(request_id, &verified_list, &blob_hash, &key.pks)
```
goshawk-3, line 433: Will that change produce different CTs compared to the prev approach of re-randomizing all ciphertexts before compression?

Are we maintaining consensus over the input ciphertexts?

antoniupop: Not yet - but if the different copros have different inputs that will reflect on all computations that derive from them.

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
```diff
@@ -501,16 +498,29 @@ fn verify_proof_only(
...
+    );
+    re_rand_context.add_ciphertext(the_list);
+    let mut seed_gen = re_rand_context.finalize();
+    let seed = seed_gen
+        .next_seed()
+        .map_err(FhevmError::ReRandomisationError)?;
+
     let expanded: tfhe::CompactCiphertextListExpander = the_list
-        .expand_without_verification()
+        .re_randomize_and_expand_without_verification(cpk, seed)
```
goshawk-3: Maybe we should add some log.info or otel traces around the new re_randomize_and_expand

goshawk-3 (dismissed):
For my benefit, is there an issue re rationale of the change? Is the new approach more secure, more performant or just API update?

nsarlin-zama (approved):
use of compact list rerand seems correct!

coprocessor/fhevm-engine/zkproof-worker/src/tests/mod.rs
```diff
@@ -129,7 +129,7 @@ async fn test_max_input_index() {
 
 #[tokio::test]
 #[serial(db)]
-async fn test_verify_proof_rerandomises_ciphertexts_before_storage() {
+async fn test_verify_proof_rerandomises_compact_list_before_expansion() {
```
nsarlin-zama, line 132: as a nit, we usually use randomize with a 'z'

goshawk-3:
@mergifyio refresh

goshawk-3 (approved):
LGTM

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

PR #2386, opened 2026-04-29
fix(host-contracts): remove redundant `resultType` from `FheSum` event

PanGan21:
### Summary

Remove the redundant `resultType` parameter from the `FheSum` event in `FHEEvents.sol`. The result type is already encoded in the handle itself and was never used from the event by the coprocessor.


PanGan21:
@claude /pr-review

Eikix (approved):
lgtm

PanGan21:
@mergify queue

PR #2391, opened 2026-04-29
feat(common): add FHE.isIn native opcode (op 29) end-to-end

PanGan21:
### Summary

- Adds FHE.isIn as a native coprocessor opcode (op 29). Solidity emits a single FheIsIn event carrying the full input array; the coprocessor executes the isIn using tfhe-rs.
- Supported types: euint8, euint16, euint32, euint64, euint128, euint160, euint256.
- Array size limits: 100 elements for Uint8–Uint32, 60 for Uint64–Uint256.

Closes: https://github.com/zama-ai/fhevm-internal/issues/1238


host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -681,6 +682,41 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
+        FheType valueType
+    ) public virtual returns (bytes32 result) {
+        uint256 supportedTypes = (1 << uint8(FheType.Uint8)) +
+            (1 << uint8(FheType.Uint16)) +
+            (1 << uint8(FheType.Uint32)) +
+            (1 << uint8(FheType.Uint64)) +
+            (1 << uint8(FheType.Uint128)) +
+            (1 << uint8(FheType.Uint160)) +
+            (1 << uint8(FheType.Uint256));
+        if ((1 << uint8(valueType)) & supportedTypes == 0) revert UnsupportedType();
```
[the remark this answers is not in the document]
PanGan21: The operator precedence is not a bug.

Solidity lists `&` before `==` in its precedence table, meaning `&` is evaluated first, so `(1 << uint8(valueType)) & supportedTypes == 0` correctly parses as `((1 << uint8(valueType)) & supportedTypes) == 0`. 

Reference: https://docs.soliditylang.org/en/latest/cheatsheet.html

The existing Foundry fuzz test `test_FheIsInNonSupportedTypesRevertAsExpected` passes 2,560 runs with random unsupported types and the UnsupportedType revert fires correctly every time. If the check were broken, this test would fail.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -681,6 +682,41 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
+
+        uint256 maxSize = (valueType == FheType.Uint64 ||
+            valueType == FheType.Uint128 ||
+            valueType == FheType.Uint160 ||
+            valueType == FheType.Uint256)
+            ? 60
```
antoniupop: Could we maybe define these as constants rather than just numerical entries inline? Just for readability and ease of update. Maybe a comment about the type width to count limit would also help.

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
```diff
@@ -101,6 +101,30 @@ fn get_fhe_sum_size_on_gpu(
...
+    input_operands: &[SupportedFheCiphertexts],
+) -> Result<u64, FhevmError> {
+    if input_operands.is_empty() {
+        return Ok(0);
+    }
+    let n = input_operands.len() as u64;
+    // No dedicated API exists; using N * ciphertext_size as an approximation.
+    match &input_operands[0] {
+        SupportedFheCiphertexts::FheUint8(v) => Ok(v.get_size_on_gpu() * n),
+        SupportedFheCiphertexts::FheUint16(v) => Ok(v.get_size_on_gpu() * n),
```
antoniupop: Just a nit since n is known to be small (for now) but maybe use a checked/sat mul ... maybe overkill.

host-contracts/contracts/FHEVMExecutor.sol
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2391/changes/626642486f8a2b737fc6101a5784db33093e1dfe

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2391/commits/fac5c5a57a0eb24824c97796d1bc80ac5b96771e

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -993,6 +1036,35 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
         ACL.allowTransient(result, msg.sender);
     }
 
+    function _naryOp(
```
PanGan21, line 1039: Note this is an overload for existing `_naryOp`. The existing one cannot be reused since it only accepts an array of values while here we have two operands. The lfs being a single value and the rfs being an array of values.

host-contracts/contracts/FHEVMExecutor.sol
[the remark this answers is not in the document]
obatirou: I concur with you: for [0.8.24](https://docs.soliditylang.org/en/v0.8.24/cheatsheet.html), `()`, `<<` and `&` takes precedence over `==` . Not a bug

Eikix:
Small (AI generated) comment found using the new zama context jumpstart:
> getTxHCUFromTxReceipt builds its parsing interface from this ABI list, but FheIsIn is missing here and there is no switch branch for it below. Any e2e HCU test or downstream helper that runs a transaction containing FheIsIn will silently drop that event in relevantLogs, underreport globalTxHCU, and fail to propagate the result handle depth. The host-contracts and library-solidity copies were updated with FheIsIn, so this test-suite copy should get the same event ABI and HCU branch.

Eikix:
Another question / comment -> 
> the core worker implementation does not match RFC 022’s stated design. RFC 022 says isIn should map to tfhe-rs contains_parallelized. The PR implements value.eq(ct) for every set element, collects Vec<FheBool>, then OR-reduces them. tfhe 1.6.1 does expose contains_parallelized, so this looks like an implementation miss, not an API blocker.

PanGan21:
> Another question / comment ->
> 
> > the core worker implementation does not match RFC 022’s stated design. RFC 022 says isIn should map to tfhe-rs contains_parallelized. The PR implements value.eq(ct) for every set element, collects Vec, then OR-reduces them. tfhe 1.6.1 does expose contains_parallelized, so this looks like an implementation miss, not an API blocker.

Fixed

PanGan21:
> Small (AI generated) comment found using the new zama context jumpstart:
> 
> > getTxHCUFromTxReceipt builds its parsing interface from this ABI list, but FheIsIn is missing here and there is no switch branch for it below. Any e2e HCU test or downstream helper that runs a transaction containing FheIsIn will silently drop that event in relevantLogs, underreport globalTxHCU, and fail to propagate the result handle depth. The host-contracts and library-solidity copies were updated with FheIsIn, so this test-suite copy should get the same event ABI and HCU branch.

Fixed

library-solidity/lib/FHE.sol
```diff
@@ -8855,6 +8855,118 @@ library FHE {
...
+            if (!isInitialized(elem)) elem = asEaddress(address(0));
+            handles[i] = eaddress.unwrap(elem);
+        }
+        return ebool.wrap(Impl.isIn(handle, handles, FheType.Uint160));
+    }
+
+    /**
+     * @dev Evaluates isIn(euint256 value, euint256[] memory values) and returns the result.
+     */
+    function isIn(euint256 value, euint256[] memory values) internal returns (ebool) {
```
Eikix, line 8957: Do we use the type euint256, ever? Do we even test it?

Eikix: Same question for eaddress

PanGan21: Do you think we should drop support for euint256? It cannot be used for FHE.sum for example because it doesn't support FHE.add but I couldn't find a reason to not support it for FHE.isIn.

Regarding the test, you are right I can add a test for it.

PanGan21: Added tests for both types.

PanGan21: I skipped initially those tests since I saw large types are not tested everywhere most likely in order to improve test performance. We can remove later if we see it takes long

Eikix:
Are the e2e tests exercised? The ones in contracts/operations? I have a memory hole

PanGan21:
> Are the e2e tests exercised? The ones in contracts/operations? I have a memory hole

Yep , they are used from this ts file which runs the actual tests through bun: https://github.com/zama-ai/fhevm/blob/57cb1396b18e2943d91d4e6f17f9d8e29d693757/test-suite/e2e/test/fhevmOperations/manual.ts

However a change might be required in order to run in the pr as well

PanGan21:
> Are the e2e tests exercised? The ones in contracts/operations? I have a memory hole

Launched the heavy operator e2e tests: https://github.com/zama-ai/fhevm/actions/runs/25367451166

PanGan21:
> > Are the e2e tests exercised? The ones in contracts/operations? I have a memory hole
> 
> Launched the heavy operator e2e tests: https://github.com/zama-ai/fhevm/actions/runs/25367451166

@Eikix e2e tests seem to pass: https://github.com/zama-ai/fhevm/actions/runs/25367451166/job/74382117416

Eikix (approved):

PanGan21:
@mergify queue

PanGan21:
@mergify queue

PR #2392, opened 2026-04-29
fix(gateway-contracts): introduce context-aware consensus

isaacdecoded:
## Summary

  In v0.12, the KMS context (signer set + threshold) used to validate a decryption response was extracted from each response's `extraData`, not pinned to the request.
  Across a KMS context rotation, this meant a single `decryptionId` could accumulate responses from different contexts in the same bucket, with the threshold check applied
  against whichever context the latest response declared. Two consequences:

  - **User decryption (low):** a rotated-out signer for an old context with `userDecryptionThreshold == 1` could solo-trigger `UserDecryptionResponseThresholdReached` after
   current-context responders had already pushed entries, prematurely latching `decryptionDone[id] = true`.
  - **Public decryption (high):** an old-context signer quorum could produce a valid certificate by submitting `extraData = 0x01 || bytes32(oldCtxId)`, locking
  `decryptionConsensusDigest[id]` to an attacker-chosen result and preventing legitimate certificates for the same `decryptionId`. This breaks the revocation guarantee of
  context rotation.

  ## Fix

  Pin the KMS contextId at **request time** and reject any response that declares a different one.

  - New storage: `decryptionContextId[decryptionId]`, written in `publicDecryptionRequest`, `userDecryptionRequest`, and `delegatedUserDecryptionRequest`.
  - In both response handlers, compute `responseContextId = _extractContextId(extraData)` and revert with `DecryptionContextMismatch` if it differs from the pinned
  `requestContextId`.
  - Signer validation (`_validateDecryptionResponseEIP712Signature`) and threshold evaluation (`_isConsensusReachedPublic` / `_isThresholdReachedUser`) now use
  `requestContextId` rather than the response-supplied contextId.

  ## What changed

  - `contracts/Decryption.sol` — new `decryptionContextId` mapping, mismatch revert in both response paths, `requestContextId` plumbed through signer/threshold checks.
  Storage layout bumped: `MINOR_VERSION 4 → 5`, `REINITIALIZER_VERSION 5 → 6`, `reinitializeV4 → reinitializeV5`.
  - `contracts/interfaces/IDecryption.sol` — new `DecryptionContextMismatch(decryptionId, requestContextId, responseContextId)` error.
  - `test/Decryption.ts` — new tests covering: pinned-context happy path, mismatch revert on response, behavior across an in-flight context rotation.

  ## Review focus

  - `Decryption.sol:343` — request-time pin (public).
  - `Decryption.sol:505`, `Decryption.sol:597` — request-time pin (user / delegated).
  - `Decryption.sol:386–397` and `Decryption.sol:648–658` — mismatch check and use of `requestContextId` in both response paths.
  - Storage layout: confirm `decryptionContextId` is appended at the end of `DecryptionStorage` and that `reinitializeV5` is the right migration step from v0.12.


isaacdecoded:
### CR-01 follow-up (OpenZeppelin v0.12.1 diff audit)

This PR addresses CR-01's primary attack (cross-context vote pooling)
by pinning `decryptionContextId` at request time and rejecting
responses whose `extraData` contextId doesn't match.

The audit also flagged a residual concern: since the pin comes from
`_extractContextId(extraData)` rather than `getCurrentKmsContextId()`,
an attacker can submit a `publicDecryptionRequest` with v1 `extraData`
naming an arbitrary prior contextId, and leaked old-context keys
could then satisfy that context's threshold.

We're keeping the pin source unchanged. The reason the client supplies
`contextId` in the first place is the user-decryption EIP-712 flow:
the user's signature covers `extraData`, so the relayer-sdk must build
it before the user signs (see
[RFC 003](../tech-spec/rfcs/003-kms-context-switch.md), line 46).
`publicDecryptionRequest` carries no user signature, so it *could*
pin server-side — but the RFC chose symmetry with the user path.

The residual leaked-keys scenario is real, but its proper fix is
**L-04 (zama-ai/fhevm-internal#1366)**: add `destroyKmsContext` /
`isValidKmsContext` on the gateway side and gate
`_checkKmsContextSignerMatchesTxSender`
(`gateway-contracts/contracts/shared/GatewayConfigChecks.sol:156`) on
context validity. Once that lands, old-context signers stop satisfying
verification regardless of pin source.

In the meantime, the KMS Connector validates `contextId` off-chain
against destroyed/invalid contexts before signing:

- `kms-connector/crates/kms-worker/src/core/event_processor/context.rs:18-43`
  (`validate_context`)
- `kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs:270-271`
  (called on every decryption event)
- `kms-connector/crates/gw-listener/src/core/publish.rs:256-273`
  (`kms_context` table populated from on-chain events)

That's defense-in-depth covering Connector-mediated paths; it doesn't
replace the on-chain control L-04 will provide against an adversary
with raw leaked keys.

jot2re (approved):
Without having  proper knowledge of the contracts' code, the changes look right to me, and with ample testing :) 
Thank you

gateway-contracts/contracts/Decryption.sol
```diff
@@ -290,11 +297,11 @@ contract Decryption is
     }
 
     /**
-     * @notice Re-initializes the contract from V3.
+     * @notice Re-initializes the contract from V4.
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    function reinitializeV4() public virtual reinitializer(REINITIALIZER_VERSION) {}
+    function reinitializeV5() public virtual reinitializer(REINITIALIZER_VERSION) {}
```
obatirou, line 306: in flight request will not be accepted if nothing is done?
any request made before the upgrade has decryptionContextId[id] == 0 if not mistaken

gateway-contracts/contracts/Decryption.sol
```diff
@@ -257,6 +257,13 @@ contract Decryption is
         /// @notice The number of user decryption requests, used to generate request IDs (`decryptionId`)
         /// @notice (including delegated user decryption requests).
         uint256 userDecryptionCounter;
+        // prettier-ignore
```
obatirou: why?

gateway-contracts/contracts/Decryption.sol
```diff
@@ -577,6 +598,9 @@ contract Decryption is
         // The publicKey and ctHandles are used during response calls for the EIP712 signature validation.
         $.userDecryptionPayloads[userDecryptionId] = UserDecryptionPayload(publicKey, ctHandles);
 
+        // Pin the KMS context at request time. See `decryptionContextId` storage docs.
+        $.decryptionContextId[userDecryptionId] = _extractContextId(extraData);
```
obatirou, line 604: in publicDecryptionRequest the comment do not send back to storage docs.

gateway-contracts/contracts/Decryption.sol
```diff
@@ -622,12 +646,21 @@ contract Decryption is
         // Compute the digest of the UserDecryptResponseVerification structure.
         bytes32 digest = _hashUserDecryptResponseVerification(userDecryptResponseVerification);
 
-        // Extract the context ID from the extraData for context-aware verification.
-        uint256 contextId = _extractContextId(extraData);
+        // Use the context ID pinned at request time, and reject responses whose extraData
```
obatirou: comment differs from above

obatirou (approved):
nit comments

obatirou:
@claude /pr-review

gateway-contracts/contracts/Decryption.sol
isaacdecoded: Yes, good catch. Added [this fallback logic](https://github.com/zama-ai/fhevm/pull/2392/changes/913027821683988a58ed1ee1901d1bad929cc970#diff-11acb2d556c4a634fdde4f84cc97005bbfe910dfb55f1e811ab08856ba711a2aR388-R397) in both decryption responses for such a situation.

gateway-contracts/contracts/Decryption.sol
isaacdecoded: Unnecessary indeed. Removed [here](https://github.com/zama-ai/fhevm/pull/2392/changes/913027821683988a58ed1ee1901d1bad929cc970#diff-11acb2d556c4a634fdde4f84cc97005bbfe910dfb55f1e811ab08856ba711a2aL260).

gateway-contracts/contracts/Decryption.sol
isaacdecoded: Aligned the comment [here](https://github.com/zama-ai/fhevm/pull/2392/changes/913027821683988a58ed1ee1901d1bad929cc970#diff-11acb2d556c4a634fdde4f84cc97005bbfe910dfb55f1e811ab08856ba711a2aR349).

gateway-contracts/contracts/Decryption.sol
isaacdecoded: Aligned both of them in the [last commit](https://github.com/zama-ai/fhevm/pull/2392/changes/913027821683988a58ed1ee1901d1bad929cc970#diff-11acb2d556c4a634fdde4f84cc97005bbfe910dfb55f1e811ab08856ba711a2aR388-R390).

gateway-contracts/contracts/Decryption.sol
```diff
@@ -290,11 +297,11 @@ contract Decryption is
     }
 
     /**
-     * @notice Re-initializes the contract from V3.
+     * @notice Re-initializes the contract from V4.
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
```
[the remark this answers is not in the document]
isaacdecoded: Solved by implementing a fallback logic during decryption responses. See [here](https://github.com/zama-ai/fhevm/pull/2392/changes/913027821683988a58ed1ee1901d1bad929cc970#diff-11acb2d556c4a634fdde4f84cc97005bbfe910dfb55f1e811ab08856ba711a2aR651-R661).

gateway-contracts/contracts/Decryption.sol
```diff
@@ -373,12 +385,20 @@ contract Decryption is
...
 
-        // Extract the context ID from the extraData for context-aware verification.
-        uint256 contextId = _extractContextId(extraData);
+        // Validate the response's context against the one pinned at request time, falling back
+        // to the response's declared context for requests recorded before context pinning.
+        // See `decryptionContextId` storage docs.
+        uint256 requestContextId = $.decryptionContextId[decryptionId];
+        uint256 responseContextId = _extractContextId(extraData);
+        if (requestContextId == 0) {
+            requestContextId = responseContextId;
```
obatirou, line 394: I am not sure we can do that.
_extractContextId could send back 0 with V1 no?

obatirou: Maybe a simpler solution is to drain in flight before migration?

gateway-contracts/contracts/Decryption.sol
```diff
@@ -210,7 +210,7 @@ contract Decryption is
      * This constant does not represent the number of time a specific contract have been upgraded,
      * as a contract deployed from version VX will have a REINITIALIZER_VERSION > 2.
      */
-    uint64 private constant REINITIALIZER_VERSION = 5;
+    uint64 private constant REINITIALIZER_VERSION = 6;
```
zmalatrax, line 213: this version has been already bumped by #2512 no?

gateway-contracts/contracts/Decryption.sol
isaacdecoded: > _extractContextId could send back 0 with V1 no?

I don't think so, extra data V1 contains the context ID. If you meant V0, it should return the current context ID.

> Maybe a simpler solution is to drain in flight before migration?

It's also a way, like pausing the contract, draining, and then upgrade, yes.

gateway-contracts/contracts/Decryption.sol
isaacdecoded: Yes, it will probably result in a conflict I'll need to solve before merging this PR.

gateway-contracts/contracts/Decryption.sol
obatirou: > I don't think so, extra data V1 contains the context ID. If you meant V0, it should return the current context ID.

What is preventing a KMS node to use extra data V1 and encoding 0?

isaacdecoded: KMS node uses whatever extra data it receives from the decryption request event - so, if the decryption event emits extra data V0, the KMS connector will verbosely propagate it to the KMS core, and the same extra data V0 will be sent back to the decryption response. Meaning that neither the KMS connector nor the KMS core alters the extra data, they only use what they get from the event arguments.

obatirou: Indeed, even in the case of malicious KMS this will not work. Thanks !

zmalatrax (approved):
lgtm, might need rebase

zmalatrax (approved):

PR #2393, opened 2026-04-29
feat(kms-connector): erc-1271 signature verification

eudelins-zama:
### Description

Implements RFC-012: moves user-decryption signature verification from `Decryption.sol` to the off-chain KMS Connector and adds an ERC-1271 fallback so smart accounts (Safe, Argent, ERC-4337) can sign without on-chain delegation.

**Flow**: EIP-712 digest → `ecrecover` fast path for EOAs → gas-bounded STATICCALL to `isValidSignature` on the host chain, accepting only the canonical `0x1626ba7e` magic value. Empty signatures bypass `ecrecover` to support Safe's `approveHash`. Configurable `erc1271_gas_limit` (default 100k).

### Skipping the `eth_getCode` pre-check

RFC-012's algorithm prescribes two RPCs in the smart-account branch — `eth_getCode` to detect EOAs, then `isValidSignature`. This PR takes the option proposed in the RFC's Open Question 1 ("Skip or cache `eth_getCode`?") and collapses it to a single RPC: STATICCALL to a no-code address succeeds with empty returndata, which is itself the no-code signal. We classify the result from the returndata alone (`EmptySigOnEoa` / `EoaMismatchNoCode` / `WrongMagic` / `Rejected`).

This way, we can have one RPC instead of two on the smart-account path; zero extra RPCs for EOAs.

Soundness: post-Dencun (EIP-6780) code can't be removed by later `SELFDESTRUCT`, so the only false-positive is a deployed contract with a fallback that returns empty bytes — same rejection outcome, only a slightly less precise error variant. Trade-off acknowledged inline.

Closes https://github.com/zama-ai/fhevm-internal/issues/1329

### DIFF
--- a/kms-connector/crates/kms-worker/src/core/config.rs
+++ b/kms-connector/crates/kms-worker/src/core/config.rs
@@ -69,6 +69,11 @@ pub struct Config {
     #[serde(with = "humantime_serde", default = "default_s3_connect_timeout")]
     pub s3_connect_timeout: Duration,
 
+    /// Gas cap for the host-chain `IERC1271.isValidSignature` static call (RFC-012).
+    /// Bounded to prevent resource exhaustion from malicious smart-account contracts.
+    #[serde(default = "default_erc1271_gas_limit")]
+    pub erc1271_gas_limit: u64,
+
     /// The service name used for tracing.
     #[serde(default = "default_service_name")]
     pub service_name: String,
@@ -162,6 +167,10 @@ fn default_s3_connect_timeout() -> Duration {
     Duration::from_secs(3)
 }
 
+fn default_erc1271_gas_limit() -> u64 {
+    100_000
+}
+
 impl DeserializeConfig for Config {}
 
 // Default implementation for testing purpose
@@ -188,6 +197,7 @@ impl Default for Config {
             max_decryption_attempts: default_max_decryption_attempts(),
             s3_ciphertext_retrieval_retries: default_s3_ciphertext_retrieval_retries(),
             s3_connect_timeout: default_s3_connect_timeout(),
+            erc1271_gas_limit: default_erc1271_gas_limit(),
             service_name: default_service_name(),
             task_limit: default_task_limit(),
             monitoring_endpoint: default_monitoring_endpoint(),
--- a/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
@@ -1,13 +1,18 @@
 use crate::core::{
     config::Config,
-    event_processor::{ProcessingError, context::ContextManager, s3::S3Service},
+    event_processor::{
+        ProcessingError,
+        context::ContextManager,
+        s3::S3Service,
+        signature::{compute_user_decrypt_digest, verify_signature},
+    },
 };
 use alloy::{
     consensus::Transaction,
     hex,
     primitives::{Address, Bytes, FixedBytes, U256, map::DefaultHashBuilder},
     providers::Provider,
-    sol_types::SolCall,
+    sol_types::{Eip712Domain, SolCall},
 };
 use anyhow::anyhow;
 use connector_utils::types::{
@@ -45,6 +50,9 @@ pub struct DecryptionProcessor<GP: Provider, HP: Provider, C> {
 
     /// The entity used to collect ciphertexts from S3 buckets.
     s3_service: S3Service<GP>,
+
+    /// Gas cap for the `IERC1271.isValidSignature` static call (RFC-012).
+    erc1271_gas_limit: u64,
 }
 
 impl<GP, HP, C> DecryptionProcessor<GP, HP, C>
@@ -76,6 +84,7 @@ where
             decryption_contract,
             acl_contracts,
             s3_service,
+            erc1271_gas_limit: config.erc1271_gas_limit,
         }
     }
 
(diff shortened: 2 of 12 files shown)


maksymsur:
@claude /pr-review

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
```diff
@@ -1084,4 +1163,81 @@ mod tests {
...
     }
+
+    // -------------------------------------------------------------------------
+    // RFC-012: signature verification wired into check_user_decryption_request_v2
+    // -------------------------------------------------------------------------
+
+    /// A flipped byte in `payload.signature` makes ecrecover return some other address; with
+    /// no contract code at `userAddress`, the ERC-1271 fallback rejects with Irrecoverable.
+    /// No invalidation/ACL RPC is reached.
+    #[tokio::test]
```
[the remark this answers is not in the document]
eudelins-zama: Had a chat with Claude about this. Agreed this is a bit of a fragile setup, no other choice than to go for the [`biased`](https://docs.rs/tokio/latest/tokio/macro.try_join.html#fairness) approach for test to be a bit more reliable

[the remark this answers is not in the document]
eudelins-zama: Fixed 9e841e2aae8ffc070353e5ab667a53cbd38de544

kms-connector/crates/kms-worker/src/core/event_processor/signature.rs
```diff
@@ -0,0 +1,454 @@
+//! RFC-012 EIP-712 signature verification for unified user-decryption requests.
+//!
+//! Implements the `ecrecover` → ERC-1271 fallback specified in RFC12.
```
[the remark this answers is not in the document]
eudelins-zama: Fixed 85adc971218c9c781dd39ad19988d14a227158ed

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
```diff
@@ -101,6 +102,23 @@ pub enum ProcessingError {
...
+/// and validity-window limits.
+impl From<Erc1271Error> for ProcessingError {
+    fn from(err: Erc1271Error) -> Self {
+        match err {
+            Erc1271Error::EoaMismatchNoCode(_) | Erc1271Error::EmptySigOnEoa(_) => {
+                Self::Irrecoverable(anyhow::Error::new(err))
+            }
+            Erc1271Error::Transport(_)
+            | Erc1271Error::WrongMagic(..)
+            | Erc1271Error::Rejected(..) => Self::Recoverable(anyhow::Error::new(err)),
```
maksymsur, line 117: It would be good to impose a monitoring and gather these errors as metrics to better evaluate user behavior. May be a part of future PR.

eudelins-zama: Good idea!

Agreed that it would be better in a dedicated PR. Here's the tracking [issue](https://github.com/zama-ai/fhevm-internal/issues/1341) :slightly_smiling_face:

kms-connector/crates/kms-worker/src/core/event_processor/signature.rs
```diff
@@ -0,0 +1,454 @@
...
+/// `chain_id` is the host `contractsChainId` (extracted from the handle), **not** the Gateway
+/// chain id used for KMS gRPC requests.
+#[allow(dead_code)] // see `DEFAULT_DOMAIN_NAME` for rationale
+pub fn default_user_decrypt_domain(chain_id: u64, verifying_contract: Address) -> Eip712Domain {
+    Eip712Domain {
+        name: Some(DEFAULT_DOMAIN_NAME.into()),
+        version: Some(DEFAULT_DOMAIN_VERSION.into()),
+        chain_id: Some(U256::from(chain_id)),
+        verifying_contract: Some(verifying_contract),
+        salt: None,
```
maksymsur, line 87: Don't we still use salt anywhere?

kms-connector/crates/kms-worker/src/core/config.rs
```diff
@@ -162,6 +167,10 @@ fn default_s3_connect_timeout() -> Duration {
     Duration::from_secs(3)
 }
 
+fn default_erc1271_gas_limit() -> u64 {
+    100_000
```
maksymsur, line 171: shall this be adjustable? Also would be good to put a comment from the RFC explaining why we have this

maksymsur (approved):
Some miner comments but LGTM!

kms-connector/crates/kms-worker/src/core/event_processor/signature.rs
eudelins-zama: No, I don't think so (and I'm not aware of any on-going discussion on that topic)

kms-connector/crates/kms-worker/src/core/config.rs
eudelins-zama: This is adjustable via the `KMS_CONNECTOR_ERC1271_GAS_LIMIT`, but we still need a default value for when it's not set.
Here is the default that was suggested in the RFC :slightly_smiling_face:

And we can still update the default in a future PR if we're not satisfied with it!

maksymsur: Ok... somehow missed KMS_CONNECTOR_ERC1271_GAS_LIMIT ))

manoranjith:
Signature.rs part LGTM (which can be used in the relayer part to do the same validations, later).

PR #2396, opened 2026-04-29
test(e2e): add js-sdk smoke profile

Eikix:
Adds a focused `js-sdk` e2e profile that encrypts a `uint64` input with the packed `@fhevm/sdk`, submits it to the existing `TestInput` contract, then verifies private and public decrypt return `49`.

The profile is included in `standard` so CI exercises SDK client compatibility while still allowing targeted runs with `fhevm-cli test js-sdk`.

The e2e Docker image now uses Node 22, builds the local SDK package, packs `sdk/js-sdk/src`, and extracts that package into `node_modules` so the test covers package exports rather than Vitest source aliases.

### DIFF
--- a/test-suite/e2e/Dockerfile
+++ b/test-suite/e2e/Dockerfile
@@ -1,4 +1,4 @@
-ARG NODE_VERSION=20.19.1-alpine3.21
+ARG NODE_VERSION=22.15.1-alpine3.21
 ARG RELAYER_SDK_VERSION=""
 
 # --- Builder Stage ---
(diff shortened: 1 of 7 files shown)


test-suite/e2e/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG NODE_VERSION=20.19.1-alpine3.21
+ARG NODE_VERSION=22.15.1-alpine3.21
```
Eikix, line 1: Not sure this is the correct image

PR #2400, opened 2026-04-30
fix(coprocessor): add upsert logic for keys and crs activation

isaacdecoded:
## What

  Fix the host-listener so that KMSGeneration.ActivateKey and ActivateCrs events actually populate the keys and crs tables.

 ## Why

  Once the new on-chain KMSGeneration flow started emitting ActivateKey / ActivateCrs events, the host-listener kept logging this every block, forever:

  ERROR Failed to update keys table with activated key content for activation
  ERROR Failed to update crs table with activated CRS content for activation

  Root cause: activate_ready_key_activations / activate_ready_crs_activations in fhevm-engine/host-listener/src/kms_generation/database.rs issued a plain UPDATE against
  keys / crs. For a brand-new key/CRS id no row exists yet, so rows_affected = 0, the loop hit continue, and the activation event stayed pinned at status = 'ready' — which
  is why the same error replayed on every iteration. No other component (engine, charts, test-suite, docker) inserts a keys / crs row for a new gen, so the activation step
  has to own that INSERT.

## What changed

  - activate_ready_key_activations is now an INSERT … ON CONFLICT (chain_id, block_hash, key_id) DO UPDATE against the existing uniq_keys_chain_block_id_gw index. It writes
   pks_key, sks_key, sns_pk, plus key_id_gw = key_id to keep the legacy lookup column in sync.
  - activate_ready_crs_activations is the analogous upsert against (chain_id, block_hash, crs_id).
  - Both selects now require their content columns to be NOT NULL before activating, so partial events stay in 'ready' rather than blowing up on keys.pks_key / keys.sks_key
   / crs.crs NOT NULL constraints. This keeps the change contained to one file with no schema migration.
  - Integration test cleanup: dropped the placeholder INSERT INTO keys/crs from seed_test_rows (the upsert now creates those rows itself), and removed two tests that pinned
   the old broken behavior — keygen_missing_target_row_stays_ready_after_materialization (asserted activations get stuck when the target row is missing — exactly the bug)
  and keygen_only_public_or_server_key_updates_partial_row (relied on partial updates against a pre-seeded row, which no longer apply under the deferred-insert design).
  - Regenerated .sqlx/ to match the new queries.

## Out of scope (worth a follow-up)

  - Downstream consumers — sns-worker/src/keyset.rs:103 and fhevm-engine-common/src/db_keys.rs:177 — query keys by key_id_gw alone and assume one row per key. The
  (chain_id, block_hash, key_id) upsert key permits multiple rows per key_id_gw if the same key id is ever re-activated on a different block. The previous UPDATE-only code
  had the same fragility; just calling it out.
  - Truly partial activations (server-only or public-only events) now stay in 'ready' indefinitely. If the contract is ever expected to emit them, we'll need to relax the
  NOT NULL on keys.pks_key / keys.sks_key via a migration.


coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -234,17 +236,25 @@ pub(crate) async fn activate_ready_key_activations(
...
                 AND e.block_hash = $2
-                AND k.key_id_gw = $3
-                AND e.key_id = k.key_id_gw
+                AND e.key_id = $3
+                AND e.key_content_public IS NOT NULL
+                AND e.key_content_sks_key IS NOT NULL
+            ON CONFLICT (chain_id, block_hash, key_id) DO UPDATE
+            SET pks_key   = EXCLUDED.pks_key,
+                sks_key   = EXCLUDED.sks_key,
+                sns_pk    = COALESCE(EXCLUDED.sns_pk, keys.sns_pk),
```
rudy-6-4, line 256: why not coalisce everywhere ?

rudy-6-4 (approved):

PR #2404, opened 2026-04-30
feat(test-suite): execute stateful rollout runbooks

Eikix:
## What

Replaces the fresh-start rollout matrix with one stateful rollout runbook.

This PR has two functional pieces:

1. In-place runtime upgrades
   - extends `fhevm-cli upgrade` with `--lock-file`
   - regenerates env/compose from the selected lock
   - runs DB migration services when present
   - restarts only the selected runtime group
   - supports relayer, KMS core + connector, listener-core, coprocessor, and test-suite transitions

2. Stateful rollout runbook execution
   - removes `test-suite-compatibility-matrix`
   - removes generated compat-test rollout locks
   - adds `rollout run <runbook.ts>` as the rollout command
   - adds `test-suite-stateful-rollout` CI
   - adds `test-suite/fhevm/rollouts/v0.12-to-v0.13/`
   - writes a rollout receipt with applied lock deltas, contract tasks, test gates, and Docker-inspected image IDs

## Current runbook pins

- `from`: v0.12.4 stack, relayer `v0.11.0`, KMS core `v0.13.10`
- `to`: repo-owned services at `v0.13.0-5`, KMS core `v0.13.20-0`
- harness: test-suite is built from this branch in CI, with `@zama-fhe/relayer-sdk@0.4.2`; runtime service images use the pins above

`v0.13.0-5` is the current latest v0.13 prerelease and is a strict superset of `v0.13.0-2`, so the #2451 KMS connector compatibility fix and the #2469 KMS migration verification script are still in the target tag. On top of `v0.13.0-2`, the new tag adds the `v0.13.0-3` host-contract/gateway-contract fixes (HCULimit zero-handle guard, KMS context pin at response time, context-aware consensus, KMS context threshold cap, etc.), the `v0.13.0-4` per-context KMS threshold setters and ACL delegation expiration boundary fix, and the `v0.13.0-5` SDK merge (#2610).

## Current order

1. boot v0.12.4 baseline once
2. run rollout-standard e2e
3. preserve old contract sources, switch gateway/host deploy images to v0.13.0-5, export gateway migration state, upgrade gateway contracts first, then deploy/migrate/upgrade host contracts, then assert the migration succeeded (task:assertKmsMigrationSucceeded from #2469)
4. refresh discovery and run rollout-standard e2e
5. upgrade relayer in place
6. run rollout-standard e2e
7. upgrade KMS core + connector in place
8. run rollout-standard e2e
9. upgrade listener-core in place
10. upgrade coprocessor services in place
11. run rollout-standard e2e

## Rollout e2e scope

`rollout-standard` covers encrypted input, FHE compute/write paths, user decrypt, delegated user decrypt, public decrypt, and ERC20 transfer coverage.

`ROLLOUT_TEST_PROFILE=rollout-heavy` adds sequential checkpoint coverage for selected heavier profiles, including operators, random subset, HCU block cap, public decrypt, DB state revert, and ciphertext drift recovery.

## Rollout receipt

The runner writes:

- `.fhevm/rollout/receipt.jsonl` for machine-readable evidence
- `.fhevm/rollout/receipt.md` for reviewer-readable CI logs

Each applied step records the action, lock file, version delta from the previous applied lock, contract task command when relevant, test gate result, and Docker-inspected container image refs/image IDs after mutating actions. CI prints the Markdown receipt with `if: always()`.

## Validation

- `cd test-suite/fhevm && bun test src/rollout-v013.test.ts src/rollout-run.test.ts src/contracts.test.ts src/cli.test.ts`
- `npx prettier --check .github/workflows/test-suite-e2e-tests.yml .github/workflows/test-suite-stateful-rollout.yml test-suite/fhevm/src/commands/rollout-receipt.ts test-suite/fhevm/src/commands/rollout-run.ts test-suite/fhevm/src/rollout-run.test.ts test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts test-suite/fhevm/rollouts/v0.12-to-v0.13/versions.ts test-suite/fhevm/rollouts/utils.ts`
- `zizmor .github/workflows/test-suite-e2e-tests.yml .github/workflows/test-suite-stateful-rollout.yml`
- `git diff --check`

## Decision log

- 2026-05-06: moved rollout from fresh-start compatibility matrices to a stateful executable runbook.
- 2026-05-06: added in-place runtime upgrade support via version locks.
- 2026-05-06: modeled the v0.13 contract migration explicitly in the runbook before runtime consumers move.
- 2026-05-07: added rollout receipts with lock deltas, commands, test gates, and Docker-inspected image evidence.
- 2026-05-07: kept v0.12.4 baseline single-chain; new v0.13 host-chain activation is not modeled as a legacy chain migration in this runbook.
- 2026-05-08: persisted the kms-core private vault across the KMS core container replacement.
- 2026-05-08: moved relayer before KMS core/connector after the rollout exposed the legacy `extraData=0x00` compatibility issue.
- 2026-05-08: validated #2451 as the KMS connector compatibility fix for relayer-sdk v0.4.2.
- 2026-05-11: rebased on current `main`, removed upstreamed #2441 cherry-pick glue, and retargeted repo-owned services to prerelease `v0.13.0-1`.
- 2026-05-11: recreated the test-suite container before each rollout test gate so regenerated phase env is loaded, and documented why listener-core has no separate gate.
- 2026-05-11: routed v0.13 prerelease relayer images to the fhevm registry while keeping legacy console relayer tags on console.
- 2026-05-11: retried Docker compose startup only for registry timeout pull failures after repeated GHCR timeouts in rollout CI.
- 2026-05-12: aligned contract migration order with the updated runbook: export gateway state, upgrade gateway contracts first, then migrate host contracts.
- 2026-05-12: treated v0.13 prerelease images as v0.13 release-family for runtime feature shims while keeping semver prerelease behavior for incompatibility checks.
- 2026-05-12: validated PR #2469 on the merge SHA before the v0.13.0-2 RC; bumped repo-owned services to v0.13.0-2 once published.
- 2026-05-12: ran task:assertKmsMigrationSucceeded inline in the rollout, tolerating only the ProtocolConfig mpc threshold mismatch that the 1-KMS-node fixture produces via normalizeLocalOneNodeMpcThreshold; any other assertion failure still fails the rollout.
- 2026-05-22: bumped repo-owned services from `v0.13.0-2` to `v0.13.0-5`, the latest published v0.13 RC. Strict superset of `v0.13.0-2`; picks up the `-3` host/gateway-contract fixes, `-4` per-context KMS threshold setters and ACL delegation expiration boundary fix, and the `-5` SDK merge (#2610).

Closes zama-ai/fhevm-internal#1355


test-suite/fhevm/compat-tests/v0.12-to-main.json
```diff
@@ -20,26 +20,26 @@
...
-    "COPROCESSOR_DB_MIGRATION_VERSION": "186c343",
-    "COPROCESSOR_HOST_LISTENER_VERSION": "186c343",
-    "COPROCESSOR_GW_LISTENER_VERSION": "186c343",
-    "COPROCESSOR_TX_SENDER_VERSION": "186c343",
-    "COPROCESSOR_TFHE_WORKER_VERSION": "186c343",
-    "COPROCESSOR_ZKPROOF_WORKER_VERSION": "186c343",
-    "COPROCESSOR_SNS_WORKER_VERSION": "186c343"
+    "RELAYER_VERSION": "413f461",
+    "RELAYER_MIGRATE_VERSION": "413f461",
+    "GATEWAY_VERSION": "413f461",
```
isaacdecoded: What changes does the version ' 413f461 ' contain? Is this built from main? If yes, it means we need to update this file every time we update the main branch then?

Eikix: Yes basically, until the code freeze! And then it's done
And for next release we'll just create a new file v0.13 to v0.14

host-contracts/tasks/taskMigrate.ts
```diff
@@ -219,6 +291,32 @@ task(
   return preparedUpgrade;
 });
 
+task(
+  'task:deployProtocolConfigFromMigration',
```
Eikix, line 239: Why create this task?

test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts
```diff
@@ -0,0 +1,112 @@
...
+import { lockSources, locks, scenario } from "./versions";
+
+type Env = Record<string, string>;
+
+const lock = (ctx: RolloutRunContext, name: string, env: Env) =>
+  ctx.lock(name, { env, sources: lockSources });
+
+const implementation = (root: string, name: string) => `${root}/${name}.sol:${name}`;
+
+const upgradeTask = (task: string, currentImplementation: string, newImplementation: string) =>
```
Eikix: should be a util

test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts
```diff
@@ -0,0 +1,112 @@
...
+  [
+    "npx hardhat",
+    task,
+    `--current-implementation ${currentImplementation}`,
+    `--new-implementation ${newImplementation}`,
+    "--verify-contract false",
+    "--use-internal-proxy-address true",
+  ].join(" ");
+
+const upgradeContract = async (runTask: (command: string) => Promise<void>, task: string, name: string) => {
```
Eikix, line 12: should be a shared util

test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts
```diff
@@ -0,0 +1,112 @@
...
+  const current = implementation("previous-contracts", name);
+  const next = implementation("contracts", name);
+  console.log(`[contracts] ${name}: ${current} -> ${next}`);
+  await runTask(upgradeTask(task, current, next));
+};
+
+const pickMigrationEnv = (value: unknown) => {
```
Eikix: what is that

test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts
```diff
@@ -0,0 +1,112 @@
...
+    Object.entries(value as Record<string, unknown>).map(([key, item]) => {
+      if (item === undefined || item === null) {
+        throw new Error(`migration-state.json export.${key} is empty`);
+      }
+      return [key, String(item)];
+    }),
+  );
+};
+
+const exportKmsMigrationEnv = async (ctx: RolloutRunContext) => {
```
Eikix: what is this? why is it needed

test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts
```diff
@@ -0,0 +1,112 @@
...
+      `--output /app/addresses/${output}`,
+    ].join(" "),
+  );
+
+  const stateFile = path.join(ctx.stateDir(), "runtime", "addresses", "gateway", output);
+  const migrationState = JSON.parse(await Bun.file(stateFile).text()) as { export?: unknown };
+  return pickMigrationEnv(migrationState.export);
+};
+
+export default async function run(ctx: RolloutRunContext) {
```
Eikix, line 194: we need more comments and space between phases to increase readability
while remaining lean

test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts
```diff
@@ -0,0 +1,112 @@
...
+  await ctx.test();
+
+  await ctx.snapshotContracts("host");
+  await ctx.snapshotContracts("gateway");
+  await ctx.transition("contracts", {
+    lockFile: contracts,
+    allowedKeys: ["GATEWAY_VERSION", "HOST_VERSION"],
+    overrides: [{ group: "host-contracts" }],
+  });
+  await ctx.hostTask("npx hardhat task:deployEmptyProxiesProtocolConfigKMSGeneration");
```
Eikix: we need to convince ourselves that this follows the runbook from @obatirou

obatirou: Should be the case https://github.com/zama-ai/fhevm/pull/2441

test-suite/fhevm/src/commands/rollout-run.ts
```diff
@@ -0,0 +1,206 @@
...
+  applyVersionLock(label: string, options: RolloutVersionLockOptions): Promise<void>;
+  readState(): Promise<State>;
+  refreshDiscovery(): Promise<void>;
+  runGatewayContractTask(command: string, options?: RolloutContractTaskOptions): Promise<void>;
+  runHostContractTask(command: string, options?: RolloutContractTaskOptions): Promise<void>;
+  snapshotContracts(surface: "host" | "gateway"): Promise<void>;
+  stateDir(): string;
+  test(profile?: string, options?: RolloutTestOptions): Promise<void>;
+  up(options: RolloutUpOptions): Promise<void>;
+  upgradeRuntime(group: string, options?: RolloutRuntimeUpgradeOptions): Promise<void>;
```
Eikix: [AUTOMATED] **Naming: `upgradeRuntime` drops the word `Group`, making the method ambiguous.**

Rule (global CLAUDE.md): "Prefer explicit verbose names over short clever ones. Avoid abbreviations or ambiguous names."

The public `RolloutRunContext` interface exposes `upgradeRuntime(group: string, ...)` while the underlying export from `up-flow.ts` is `upgradeRuntimeGroup`. The interface name doesn't communicate that the argument is a *group*, not an individual service or runtime process.

Suggested fix: rename the interface method (and its implementation at line 158) to `upgradeRuntimeGroup` to match the exported function name.

Confidence: 85/100

test-suite/fhevm/src/commands/rollout-run.ts
```diff
@@ -0,0 +1,206 @@
...
+  async upgradeRuntime(group, options = {}) {
+    await upgradeRuntimeGroup(group, options);
+    await receipt.record("upgrade-runtime", group, { docker: true, lockFile: options.lockFile });
+  },
+  async writeVersionLock(name, options) {
+    const filename = name.endsWith(".json") ? name : `${name}.lock.json`;
+    const file = path.join(STATE_DIR, "rollout", filename);
+    await ensureDir(path.dirname(file));
+    const bundle = {
+      target: options.target ?? "sha",
```
Eikix: [AUTOMATED] **Lock-file semantics: `writeVersionLock` defaults `target` to `"sha"` for phase locks that are not SHA-pinned bundles.**

Rule (README.md lock-file section): The `target` field is a required discriminator. `"sha"` means an exact repo-SHA bundle was supplied by the caller. But the rollout runbook never passes `target`, so `options.target ?? "sha"` silently stamps every phase lock as a SHA bundle — even though `ctx.up(...)` boots with `target: "latest-main"` (line 70).

Suggested fix: pass `target: "latest-main"` explicitly in `writePhaseVersionLock`, or change the default in `writeVersionLock` to `"latest-main"`.

Confidence: 82/100

test-suite/fhevm/src/commands/rollout-run.ts
```diff
@@ -0,0 +1,206 @@
...
+  test(profile?: string, options?: RolloutTestOptions): Promise<void>;
+  up(options: RolloutUpOptions): Promise<void>;
+  upgradeRuntime(group: string, options?: RolloutRuntimeUpgradeOptions): Promise<void>;
+  writeVersionLock(name: string, options: RolloutLockOptions): Promise<string>;
+};
+
+export type RolloutRunbook = (ctx: RolloutRunContext) => Promise<void> | void;
+
+const upOptions = (options: RolloutUpOptions): UpOptions => ({
+  target: "latest-main",
```
Eikix, line 70: [AUTOMATED] **Resolution-order violation: forcing `target: "latest-main"` alongside a `--lock-file` contradicts the documented invariant.**

Rule (README.md, Resolution Order section): "The lock file replaces *only* the version resolution step." When `--lock-file` is present, `--target` must either be absent or match the lock file's own `target`; passing both means the CLI double-resolves versions.

Here `upOptions` always sets `target: "latest-main"` regardless of what the supplied lock file declares. This breaks the resolution-order contract for phase locks whose `target` differs (e.g. `"sha"` from Issue above).

Suggested fix: omit `target` from `upOptions` (or set it to `undefined`) when a `lockFile` is present, letting the lock file's own `target` field govern resolution.

Confidence: 81/100

test-suite/fhevm/src/cli.ts
```diff
@@ -259,32 +262,12 @@ const root = defineCommand({
...
     }),
     rollout: defineCommand({
-      meta: { name: "rollout", description: "Generate cumulative mixed-version rollout lock files from a compat-test definition." },
+      meta: { name: "rollout", description: "Execute a TypeScript rollout runbook against one stateful stack." },
       args: {
-        "compat-test": { type: "string", description: "Compat-test JSON path." },
-        out: { type: "string", description: "Output directory for generated lock files, or file path when --step is set." },
-        step: { type: "string", description: "Render only one rollout step index instead of the full directory." },
-        matrix: { type: "boolean", description: "Print the rollout matrix JSON instead of writing lock files." },
+        run: { type: "string", description: "Execute a TypeScript rollout runbook." },
```
Eikix: [AUTOMATED] **Command structure: `rollout --run <path>` is a flag-based command, not an imperative subcommand.**

Rule (prompt guideline, command structure): Commands should follow the imperative subcommand form `rollout run` / `rollout receipt` consistent with every other multi-action command in this CLI (e.g. `test standard`, `scenario list`).

Currently `rollout` is a leaf command that accepts `--run` as a named option, not a command group with `run` and `receipt` subcommands. Additionally, `rollout-receipt.ts` is imported nowhere in `cli.ts` — there is no `rollout receipt` subcommand exposed to users.

Suggested fix: restructure `rollout` as a command group with `run` (positional runbook path) and `receipt` subcommands, matching the pattern of `test` and `scenario`.

Confidence: 88/100

test-suite/fhevm/rollouts/utils.ts
```diff
@@ -0,0 +1,26 @@
+import type { RolloutRunContext } from "../src/commands/rollout-run";
+
+export type RolloutEnv = Record<string, string>;
+
+export const logPhase = (label: string) => {
+  console.log(`\n[rollout] ${label}`);
+};
+
+const implementation = (root: string, name: string) => `${root}/${name}.sol:${name}`;
```
Eikix: [AUTOMATED] **Naming: `implementation` is a generic noun that does not describe what it constructs.**

Rule (global CLAUDE.md): "explicit verbose names over short clever ones — use descriptive names that say what the thing does."

`const implementation = (root, name) => \`${root}/${name}.sol:${name}\`` constructs a Solidity contract path string. The name `implementation` is ambiguous: it could refer to a class implementation, an interface implementation, or something else entirely.

Suggested fix: rename to `contractSolidityPath` or `solidityImplementationPath`.

Confidence: 80/100

.github/workflows/test-suite-stateful-rollout.yml
```diff
@@ -0,0 +1,159 @@
...
+
+      - name: Install foundry
+        uses: foundry-rs/foundry-toolchain@82dee4ba654bd2146511f85f0d013af94670c4de # v1.4.0
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GHCR_READ_TOKEN }} # zizmor: ignore[secrets-outside-env] Same-repository/manual rollout needs this read token for private images.
```
Eikix: [AUTOMATED] **Workflow security: `secrets.GHCR_READ_TOKEN` is used unconditionally in a PR-triggered workflow without a fallback.**

Rule (CONTRIBUTING.md, zizmor guidance): secrets usage in PR workflows must be carefully scoped. The `zizmor: ignore` justification claims "same-repository/manual rollout needs this read token," but the workflow *also* triggers on `pull_request` events guarded only by a label check. Any maintainer who can apply the `rollout` label can trigger this path and cause the secret to be used.

The deleted `test-suite-compatibility-matrix.yml` used the safer pattern `${{ secrets.GHCR_READ_TOKEN || github.token }}`, which falls back to the built-in GITHUB_TOKEN when the org secret is absent — a defense-in-depth approach.

Suggested fix: change line 67 to `password: ${{ secrets.GHCR_READ_TOKEN || github.token }}` consistent with the previous workflow's pattern.

Confidence: 83/100

Eikix (commented):
[AUTOMATED]

## Guideline-Compliance Review

6 high-confidence violations found (all ≥80). Details in inline comments.

test-suite/fhevm/src/compat/compat.ts
```diff
@@ -174,58 +211,83 @@ const usesModernWorkspaceProtocol = (state: CompatState) =>
   );
 
 export const requiresMultichainAclAddress = (state: CompatState) =>
-  !usesModernWorkspaceProtocol(state) &&
-  versionLt(state.versions.env.COPROCESSOR_TX_SENDER_VERSION ?? "", [0, 12, 0]);
+  !usesModernWorkspaceProtocol(state) && versionLt(state.versions.env.COPROCESSOR_TX_SENDER_VERSION ?? "", [0, 12, 0]);
 
 /** Detects when relayer readiness config must stay on the legacy shape. */
 export const requiresLegacyRelayerReadinessConfig = (state: Pick<CompatState, "versions">) =>
   versionLt(state.versions.env.RELAYER_VERSION ?? "", [0, 10, 0]);
```
Eikix: [AUTOMATED] **Feature-shim selector uses prerelease-strict `versionLt` — inconsistent with the rest of the file.**

This selector triggers a *legacy feature shim*. Neighboring shim selectors in the same file (`requiresLegacyKmsCoreConfig` L222, `requiresLegacyHostPauserTaskFlag`, `requiresLegacyGatewayKmsGenerationAddress`, and the `legacyShims` loop in `compatPolicyForState` around L328) all use `versionBeforeReleaseFamily`. Per the PR's stated policy (decision-log: "treated v0.13 prerelease images as v0.13 release-family for runtime feature shims"), prerelease tags should NOT activate legacy shims.

For `RELAYER_VERSION="v0.10.0-rc1"`, `versionLt(...,[0,10,0])` returns `true` because `parsed.prerelease` triggers the strict-less branch (compat.ts L190). Result: a v0.10 prerelease falsely gets the legacy readiness shape.

Fix:
```ts
export const requiresLegacyRelayerReadinessConfig = (state: Pick<CompatState, "versions">) =>
  versionBeforeReleaseFamily(state.versions.env.RELAYER_VERSION ?? "", [0, 10, 0]);
```

Confidence: 90/100

test-suite/fhevm/src/compat/compat.ts
```diff
@@ -174,58 +211,83 @@ const usesModernWorkspaceProtocol = (state: CompatState) =>
...
 
 /** Detects when kms-core still expects the legacy config schema. */
 export const requiresLegacyKmsCoreConfig = (state: Pick<CompatState, "versions">) =>
-  versionLt(state.versions.env.CORE_VERSION ?? "", [0, 13, 10]) &&
-  !sameCompatBase(state.versions.env.CORE_VERSION ?? "", [0, 13, 10]);
+  versionBeforeReleaseFamily(state.versions.env.CORE_VERSION ?? "", [0, 13, 10]);
 
 /** Detects when test-suite should use the legacy relayer base URL. */
 export const requiresLegacyRelayerUrl = (state: Pick<CompatState, "versions">) =>
   versionLt(state.versions.env.TEST_SUITE_VERSION ?? "", [0, 11, 0]);
```
Eikix: [AUTOMATED] **Same prerelease-strict bug as L218 — `requiresLegacyRelayerUrl` should use `versionBeforeReleaseFamily`.**

`TEST_SUITE_VERSION="v0.11.0-rc1"` would be told to use the legacy `/v1` relayer base URL because `versionLt` treats prerelease < release. Inconsistent with the same file's `requiresLegacyKmsCoreConfig` and with `compatPolicyForState`'s shim loop, both of which use `versionBeforeReleaseFamily`.

Fix: switch to `versionBeforeReleaseFamily(state.versions.env.TEST_SUITE_VERSION ?? "", [0, 11, 0])`.

Confidence: 90/100

test-suite/fhevm/src/compat/compat.ts
```diff
@@ -174,58 +211,83 @@ const usesModernWorkspaceProtocol = (state: CompatState) =>
   );
 
 export const requiresMultichainAclAddress = (state: CompatState) =>
-  !usesModernWorkspaceProtocol(state) &&
-  versionLt(state.versions.env.COPROCESSOR_TX_SENDER_VERSION ?? "", [0, 12, 0]);
+  !usesModernWorkspaceProtocol(state) && versionLt(state.versions.env.COPROCESSOR_TX_SENDER_VERSION ?? "", [0, 12, 0]);
```
Eikix: [AUTOMATED] **`requiresMultichainAclAddress` contradicts `compatPolicyForState`'s shim loop for the same input.**

This helper uses `versionLt(..., [0, 12, 0])` while the matrix-driven loop in `compatPolicyForState` (around L328) evaluates the same `COPROCESSOR_TX_SENDER_VERSION` against the same `[0, 12, 0]` floor via the `legacy-tx-sender-gateway-flags` SHIM_PROFILES entry — using `versionBeforeReleaseFamily`.

Result for `COPROCESSOR_TX_SENDER_VERSION="v0.12.0-rc1"`:
- `requiresMultichainAclAddress` → `true` → drives env + discovery to require/set `MULTICHAIN_ACL_ADDRESS` (generate/env.ts L186, flow/discovery.ts L122)
- `compatPolicyForState` → does NOT inject `--multichain-acl-address` flag for tx-sender

Two contradictory views of the same release family. Latent today (current runbook uses non-prerelease `v0.12.4`) but trips on any future v0.12.x prerelease.

Fix: use `versionBeforeReleaseFamily` here too.

Confidence: 88/100

test-suite/fhevm/src/compat/compat.ts
```diff
@@ -128,7 +158,10 @@ const parseCompatVersion = (version: string) => {
   };
 };
 
-const usesModernRelayerRepository = (version: string) => !parseCompatVersion(version);
+const usesModernRelayerRepository = (version: string) => {
+  const parsed = parseCompatVersion(version);
+  return !parsed || parsed.parts[0] > 0 || parsed.parts[1] >= 13;
+};
```
Eikix, line 166: [AUTOMATED] **Unparseable / empty RELAYER_VERSION silently selects the MODERN registry.**

`parseCompatVersion("")` returns `undefined` (regex fails to match). `usesModernRelayerRepository(version)` then evaluates `!parsed || …` and short-circuits to `true`, so `relayerImageRepository("")` and `relayerMigrateImageRepository("")` both resolve to `MODERN_…`. In `compatPolicyForState` L356-358 the lookups are `state.versions.env.RELAYER_VERSION ?? ""`, so a missing env value coerces silently to the modern `ghcr.io/zama-ai/fhevm/relayer{,-migrate}` repo.

This is the wrong default: an unset/empty version should fail loudly (preflight), not pick a registry route silently — especially since this PR's decision log explicitly distinguishes the legacy `console` org from the new `fhevm` registry.

Fix: in `usesModernRelayerRepository`, treat an unparseable version as an error instead of `true`:
```ts
const usesModernRelayerRepository = (version: string) => {
  const parsed = parseCompatVersion(version);
  if (!parsed) throw new PreflightError(`relayer version unparseable: ${version!r}`);
  return parsed.parts[0] > 0 || parsed.parts[1] >= 13;
};
```
Or add a non-empty assertion at the call sites in `compatPolicyForState`.

Confidence: 85/100

test-suite/fhevm/src/commands/rollout-receipt.ts
```diff
@@ -0,0 +1,189 @@
...
+          health: value.State?.Health?.Status,
+        }))
+        .sort((a, b) => (a.service ?? a.name).localeCompare(b.service ?? b.name)),
+    };
+  } catch (error) {
+    return {
+      containers: [],
+      error: error instanceof Error ? error.message : String(error),
+    };
+  }
```
Eikix, line 98: [AUTOMATED] **`inspectContainers` swallows all errors into `{ containers: [], error }` with no stdout signal.**

Three error paths (non-zero `docker inspect` exit, JSON parse failure, generic exception) all return `{ containers: [], error: … }`. The error string is persisted in `entry.dockerInspectError` and rendered as an italic footnote in markdown, but **never printed to stdout**. The only console output per record is `[receipt] N. kind: title` (L183).

Impact: if `docker inspect` fails mid-rollout (daemon glitch, mass container recreation race), the receipt for every subsequent action records `containers: []`, the CI markdown shows an empty container table, and an operator scanning the live CI log sees no signal that container-state evidence is missing. Post-migration crashes become invisible.

Fix: log a one-line `console.warn(…)` from inside the catch / non-zero branches so the failure shows up in the live CI log, even though the JSONL+markdown also record it.

Confidence: 88/100

.github/workflows/test-suite-stateful-rollout.yml
```diff
@@ -0,0 +1,159 @@
...
+            local container="$2"
+            local filter="${3:-}"
+            echo "::group::${group}"
+            if ! docker ps -a --format '{{.Names}}' | grep -Fxq "$container"; then
+              echo "container not present: ${container}"
+              echo "::endgroup::"
+              return
+            fi
+            if [ -n "$filter" ]; then
+              docker logs --tail 120 "${container}" 2>&1 | grep -v "$filter" || true
```
Eikix: [AUTOMATED] **`docker logs … || true` masks log-capture failures in the post-failure diagnostic step.**

This step runs under `if: always()` to capture diagnostics after a failed rollout — exactly the moment where missing logs hurt the most. If `docker logs` itself fails (container removed before this step runs, daemon error, container name typo'd in a future edit), `|| true` swallows the non-zero exit and `2>&1` keeps the error from showing through the `grep -v` filter on L128. The CI group simply prints nothing and exits 0, with no marker that log capture was attempted and failed.

Fix: replace `|| true` with a visible fallback so the operator sees that the snapshot attempt failed:
```bash
docker logs --tail 120 "${container}" 2>&1 | grep -v "$filter" || echo "[warn] docker logs failed for ${container}"
```
Same on L130.

Confidence: 92/100

test-suite/fhevm/src/commands/rollout-run.ts
```diff
@@ -0,0 +1,206 @@
...
+    await (
+      await loadRolloutRunbook(script)
+    )(context);
+    await receipt?.record("complete", "runbook completed", { docker: true });
+  } catch (error) {
+    await receipt?.record("failed", "runbook failed", {
+      details: { error: error instanceof Error ? error.message : String(error) },
+      docker: true,
+    });
+    throw error;
```
Eikix, line 203: [AUTOMATED] **Catch block can lose the original runbook error if `receipt.record` itself throws.**

```ts
} catch (error) {
  await receipt?.record("failed", "runbook failed", { … });   // <-- may throw
  throw error;
}
```

If `record()` throws (most likely failure mode: `fs.appendFile` to a state dir that wasn't created because `start()` was never called — see the `start("(unknown)")` fallback in rollout-receipt.ts L160-162 which truncates if `started` is false), then the new error propagates and the original runbook `error` is dropped. The operator sees a misleading filesystem/receipt error mid-rollout, not the actual root cause.

Fix: nest the record call in its own try/catch so the original error always wins:
```ts
} catch (error) {
  try {
    await receipt?.record("failed", "runbook failed", { details: { error: error instanceof Error ? error.message : String(error) }, docker: true });
  } catch (recordError) {
    console.error("[receipt] failed to record runbook failure", recordError);
  }
  throw error;
}
```

Confidence: 85/100

Eikix (commented):
[AUTOMATED]

## Bug-Hunter / Error-Handling Review (second pass)

Follow-up pass after the earlier guideline review (4270984980). 7 validated high-confidence findings — 4 in `compat.ts` (the `versionLt` -> `versionBeforeReleaseFamily` sweep is incomplete, and an unparseable version coerces to MODERN repo), 3 silent-failure / lost-error paths in receipt + runbook + CI logs.

Good news from validation: the `v0.13.0-1` patch-id claim for #2451 is verifiable (the v0.13-branch commit touches the same 4 files as #2458 on main); the in-place upgrade ordering is internally consistent; PC-2 (extraData v1/v2 incompatibility window) is **not** a real risk — `RELAYER_SDK_VERSION` is pinned to 0.4.2 and that build only emits `0x00` extraData, which v0.12.4 KMS connector accepts as the legacy default-context marker.

test-suite/fhevm/src/commands/rollout-run.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. Renamed interface method to `upgradeRuntimeGroup` (aliased the up-flow import to `upgradeStackRuntimeGroup` to avoid shadowing); runbook + tests + README updated.

test-suite/fhevm/src/commands/rollout-run.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. `writeVersionLock` default `target` changed from `"sha"` to `"latest-main"` — matches what these runbook-generated locks actually describe.

test-suite/fhevm/src/commands/rollout-run.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. Resolved as a side effect of the writeVersionLock default change above: phase locks now declare `target: "latest-main"`, matching the `target` `upOptions` passes, so there is no longer a contradiction. The resolver still treats the lock-file `target` as authoritative.

test-suite/fhevm/src/cli.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. `rollout` is now a command group: `rollout run <runbook>` (imperative positional) and `rollout receipt` (prints the markdown receipt). Workflow + README updated.

test-suite/fhevm/rollouts/utils.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. Renamed to `fullyQualifiedContractName(sourceRoot, contractName)`.

.github/workflows/test-suite-stateful-rollout.yml
Eikix: [AUTOMATED] Fixed in f836ce7cb. Restored `${{ secrets.GHCR_READ_TOKEN || github.token }}` fallback on the GHCR docker login.

test-suite/fhevm/src/compat/compat.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. Switched to `versionBeforeReleaseFamily` (compat.ts requiresLegacyRelayerReadinessConfig).

test-suite/fhevm/src/compat/compat.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. Switched to `versionBeforeReleaseFamily` (compat.ts requiresLegacyRelayerUrl).

test-suite/fhevm/src/compat/compat.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. Switched to `versionBeforeReleaseFamily` (compat.ts requiresMultichainAclAddress).

test-suite/fhevm/src/compat/compat.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. `usesModernRelayerRepository` now throws `PreflightError` on empty input (call sites also guard with truthiness checks before invoking). SHA-style unparseable inputs keep returning MODERN as before.

test-suite/fhevm/src/commands/rollout-receipt.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. `inspectContainers` now routes both failure paths through an `inspectFailed` helper that `console.warn`s, so docker-inspect failures surface in live CI logs.

.github/workflows/test-suite-stateful-rollout.yml
Eikix: [AUTOMATED] Fixed in f836ce7cb. Replaced `docker logs ... || true` with a captured-output pattern: on failure it prints `[warn] docker logs failed for <container>` and best-effort dumps any partial output, instead of a silent empty group.

test-suite/fhevm/src/commands/rollout-run.ts
Eikix: [AUTOMATED] Fixed in f836ce7cb. Wrapped `receipt?.record("failed", ...)` in an inner try/catch with `console.error`, so the original runbook error always propagates through the outer `throw error`.

test-suite/fhevm/src/flow/up-flow.ts
```diff
@@ -1204,25 +1199,131 @@ export const listScenarios = async () => {
...
+      `${label} lock changes unrelated version keys: ${disallowed.join(", ")}. Allowed: ${allowedVersionKeys.join(", ")}`,
     );
   }
-  await ensureRuntimeArtifacts(state, "upgrade");
-  const { component, group, runtimeServices, step } = resolveUpgradePlan(state, groupValue);
-  if (!state.completedSteps.includes(step)) {
-    throw new PreflightError(`upgrade requires a stack that has completed the ${step} step`);
+};
+
+const applyRuntimeUpgradeLock = async (
```
PanGan21: I think `--lock-file` upgrades can still keep using local workspace images here. This functions updates `state.versions`, but keeps the existing override for the same group, so the lock-file version may never actually take effect. Should we clear that override or error out?

.github/workflows/test-suite-stateful-rollout.yml
```diff
@@ -0,0 +1,163 @@
...
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+      - unlabeled
+  workflow_dispatch:
+    inputs:
+      runbook:
+        description: "TypeScript rollout runbook path"
+        default: "test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts"
```
PanGan21: I guess the default is subject to change in later releases? Also I am not sure if we will upgrade from v0.12-to-v0.13 and not v0.12-to-v0.13.2 for example.

test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts
```diff
@@ -0,0 +1,212 @@
...
+  const testMode = resolveRolloutTestMode(process.env.ROLLOUT_TEST_PROFILE);
+  const baselineLock = await writePhaseVersionLock(ctx, "00-baseline", phaseVersions.baseline);
+  const contractsLock = await writePhaseVersionLock(ctx, "01-contracts", phaseVersions.contracts);
+  const relayerLock = await writePhaseVersionLock(ctx, "02-relayer", phaseVersions.relayer);
+  const kmsLock = await writePhaseVersionLock(ctx, "03-kms", phaseVersions.kms);
+  const listenerCoreLock = await writePhaseVersionLock(ctx, "04-listener-core", phaseVersions.listenerCore);
+  const coprocessorLock = await writePhaseVersionLock(ctx, "05-coprocessor", phaseVersions.coprocessor);
+
+  logPhase("00 baseline: boot v0.12.4 with the target test-suite harness");
+  await ctx.up({ lockFile: baselineLock, scenario, overrides: [{ group: "test-suite" }] });
```
PanGan21, line 204: We only override test-suite, so local host-contracts and gateway-contracts changes are not exercised by this rollout. Unless I am mistaken that means only fixes on the tests will be picked up for this job

.github/workflows/test-suite-stateful-rollout.yml
Eikix: good point

test-suite/fhevm/rollouts/v0.12-to-v0.13/run.ts
Eikix: yes exactly

PanGan21: Ok thanks I wanted to validate that this is on purpose

test-suite/fhevm/src/flow/up-flow.ts
Eikix: Good point!

PanGan21 (dismissed):

Eikix:
Waiting on https://github.com/zama-ai/fhevm/pull/2479 before merging

PanGan21 (dismissed):

PanGan21 (dismissed):

Eikix:
@mergify queue

PanGan21 (dismissed):

PanGan21 (dismissed):

Eikix:
It seems I need more approvals than expected, cc @eudelins-zama

fegmorte (commented):
your actions version are all outdated in `.github/workflows/test-suite-stateful-rollout.yml`. Please check the repo of different actions to use the commit sha of the latest release tag

eudelins-zama (approved):
LGTM on kms-connector side

PanGan21 (approved):

nboisde (approved):

Eikix:
@mergifyio queue

PR #2407, opened 2026-04-30
feat(host-contracts): add DAO KMS migration tooling

obatirou:
closes https://github.com/zama-ai/fhevm-internal/issues/1231


host-contracts/docs/dao_kms_migration_dao_operator_runbook.md
```diff
@@ -0,0 +1,435 @@
...
+- Do not execute any proposal until the full artifact package has been reviewed and signed off.
+- Do not reopen new key-management traffic until post-cutover validation is complete.
+
+Execute proposals in this order:
+
+| Label | Proposal                                  | Must execute after                  |
+| ----- | ----------------------------------------- | ----------------------------------- |
+| A     | Host `ProtocolConfig` migration           | snapshot export and artifact review |
+| B     | Host `KMSGeneration` migration            | proposal A                          |
+| C     | Gateway `KMSGeneration` view-only upgrade | proposal B                          |
```
isaacdecoded: I'd place "D- Host `KMSVerifier` upgrade" **before** "C - Gateway `KMSGeneration` view-only upgrade"

I mean, finishing with the Host changes before touching the Gateway, so if something goes wrong on the Host, we avoid worrying (or needing a rollback) because we have already changed the Gateway.

But probably just a matter of personal opinion.

Eikix:
(AI + Human in the loop review) Remaining findings from the KMS migration review:

1. [P2] Require hydrated ACL address before migration proxy deployment

`task:deployEmptyProxiesProtocolConfigKMSGeneration` compiles and deploys `EmptyUUPSProxy`, whose upgrade authorization embeds `aclAdd` from `addresses/FHEVMHostAddresses.sol`. The normal deploy flow writes `aclAdd` first, then compiles `EmptyUUPSProxy`; later recompilation only fixes implementation contracts and cannot fix an already deployed empty proxy compiled against the wrong ACL. The migration task should preflight that `addresses/.env.host` and `FHEVMHostAddresses.sol` contain the target ACL address.

2. [P2] Recompile migration implementations after address hydration

The prepare-only migration tasks deploy implementations using current artifacts/factories. That is risky because `ProtocolConfig` depends on `aclAdd`, and `KMSGeneration` depends on `protocolConfigAdd`, both generated from `addresses/FHEVMHostAddresses.sol`. If artifacts were compiled before target addresses were written, `prepareUpgrade` can deploy bytecode with stale constants while still emitting coherent DAO calldata. The task should compile the relevant contracts after address hydration and before `readArtifact` / `getContractFactory`.

3. [P3] State the GatewayConfig context precondition before exporting migration state

`task:exportKmsMigrationState` uses the v0.12+ `GatewayConfig` context API. That is fine for the intended path, but the runbook should state that the v0.12 `GatewayConfig` context upgrade must already have executed before this exporter is run. For mainnet, where v0.12 and v0.13 work may be grouped through DAO multicalls, operators should not run this exporter against the pre-v0.12 `GatewayConfig` implementation.

4. [P2] Make proposal artifacts uniform for all four DAO proposals

The infra runbook says prepare logs include every proxy address, implementation address, calldata value, and transaction hash needed for proposal assembly. The new host `ProtocolConfig` and host `KMSGeneration` migration tasks print a full proposal artifact shape, including `outerCalldata`, but existing `prepareUpgradeKMSVerifier` and gateway `prepareUpgradeKMSGeneration` only print the implementation address and inner reinitializer calldata. For DAO review, all four proposals should have the same deterministic artifact format, or the runbook must clearly say how those two artifacts are derived.

5. [P3] Link the KMS core delivery constraint into the migration runbooks

zama-ai/planning-blockchain#1199 is part of this migration: KMS core `v0.13.20-0` makes context/`extra_data` part of keygen, CRS, preprocessing, and signatures. The runbooks should reference it and state that new key-management traffic must stay frozen until the KMS core compatibility phase and the on-chain context/KMSGeneration migration are both complete.

Eikix (approved):

isaacdecoded (approved):

obatirou:
@mergify queue

PR #2409, opened 2026-04-30
feat(coprocessor): add contract config maps

deasydoesit:
This PR:

- Adds `hostContractConfigMap` and `gatewayContractConfigMap` creation


charts/coprocessor/configs/gateway-mainnet.yaml
```diff
@@ -0,0 +1,9 @@
+contracts.version: "v0.10.0"
+ciphertext_commits.address: "0xd82cF70FC102028cd01acB87D0E107780ae4F41F"
```
piizama: We should remove the contract version. It will be really hard to keep this value up to date.

piizama (changes requested):

piizama (approved):

eudelins-zama (approved):

deasydoesit:
@mergify queue

PR #2410, opened 2026-04-30
feat(coprocessor): add upgrade orchestrator service for blue-green upgrades

goshawk-3:
Introduces `coproc-mngr`, a new long-running service - one per coprocessor
  party - that owns the upgrade lifecycle FSM described in RFC-021. It
  listens on the Postgres `event_upgrade` channel, drives the per-stack
  state machine on `proposedUpgrade` / `CoprocUpgraded` events, and
  coordinates the other services through `pg_notify` and a small set of
  new tables.

### DIFF
--- a/coprocessor/fhevm-engine/Cargo.lock
+++ b/coprocessor/fhevm-engine/Cargo.lock
@@ -2409,6 +2409,33 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


goshawk-3:
This PR is deprecated in favor of  https://github.com/zama-ai/fhevm/pull/2604, which supersedes it.

PR #2413, opened 2026-05-01
fix(common): add codeql triggers for push and pull_request (#1337)

paulo-zama:
Add codeql triggers for push and pull_request

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -13,6 +13,10 @@ name: codeql
 permissions: {} # No permissions needed at workflow level
 
 on:
+  push:
+    branches:
+      - main
+  pull_request:
   schedule:
     - cron: '30 5 * * 1-5'
 


chilcano (commented):
## Review: `fix(common): add codeql triggers for push and pull_request`

Thanks for adding event-driven CodeQL scanning — the scheduled-only setup was the main gap before this PR. The intent is right, but the trigger configuration as written will create significant CI overhead. See suggestions below.

---

### Problem with the current diff

The bare `pull_request` trigger (no `branches` or `paths` filter) fires on **every push to every open PR branch** across the whole repo. CodeQL scans 4 languages in parallel (`actions`, `javascript-typescript`, `python`, `rust`), so each push to any PR spawns 4 jobs. On a repo with many concurrent PRs and frequent pushes, this will saturate the CI queue.

```yaml
# current PR — too broad
on:
  push:
    branches:
      - main
  pull_request:          # ← no filter: fires on every push to every PR branch
  schedule:
    - cron: '30 5 * * 1-5'
```

---

### Suggested approach: split into two workflows

Rather than patching filters onto a single file, splitting into two purpose-built workflows allows each to be tuned independently and avoids future accidental coupling.

#### `codeql-main.yml` — full scan, no rush

Covers the authoritative branch and the existing schedule. No cancellation logic needed here.

```yaml
name: codeql-main
permissions: {}

on:
  push:
    branches:
      - main
  schedule:
    - cron: '30 5 * * 1-5'

jobs:
  analyze:
    # identical to current jobs.analyze block
```

#### `codeql-pr.yml` — fast, targeted, cancellable

Covers PRs targeting `main` only. Cancels stale runs when a new push arrives and skips non-code changes.

```yaml
name: codeql-pr
permissions: {}

concurrency:
  group: codeql-pr-${{ github.ref }}
  cancel-in-progress: true

on:
  pull_request:
    branches:
      - main
    paths:
      - '**.ts'
      - '**.js'
      - '**.py'
      - '**.rs'
      - '**.sol'
      - '.github/workflows/**'

jobs:
  analyze:
    # identical to current jobs.analyze block
```

Key differences from the current PR:
- `branches: [main]` on `pull_request` — ignores feature-branch-to-feature-branch PRs
- `paths` filter — skips scans when only docs/markdown/configs change (CodeQL can't analyze those file types anyway, so no coverage loss)
- `concurrency: cancel-in-progress: true` — drops the stale scan immediately when a new commit is pushed to the same PR, instead of letting both run in parallel

---

### Is there a security gap?

No. The `codeql-main` workflow has **no `paths` filter**, so every push to `main` triggers a full scan regardless of what changed. Even if `codeql-pr` was skipped on a PR (e.g., only docs changed), the merged result is always scanned on `main`. The nightly schedule is a second safety net.

One dependency to verify: make sure the `codeql-pr / codeql/analyze-*` checks are listed as **required status checks** in branch protection for `main`. Without that, a reviewer can merge before the scan completes regardless of workflow design.

---

### Summary of changes requested

| File | Action |
|---|---|
| `.github/workflows/codeql.yml` | Rename to `codeql-main.yml`; keep only `push: main` + `schedule` triggers |
| `.github/workflows/codeql-pr.yml` | Create new file with `pull_request: branches: [main]` + `paths` + `concurrency` |

The `jobs.analyze` block is identical in both files — no changes to the scan logic itself.

PR #2417, opened 2026-05-04
fix(coprocessor): reduce 0.12 migration lock time

Eikix:
## What

Ports the 0.12 migration lock-time reduction from #2365 onto `main`.

This keeps the optimized `remove_tenants` migration shape on main: large tenant-free indexes can be built online with `RUN_MIGRATIONS_UNTIL_REMOVE_TENANTS=true`, and the downtime migration then skips those pre-created indexes and uses metadata-only constant defaults for `host_chain_id` / `key_id_gw`.

The port preserves the current main db-migration `prepare_database_url.sh` flow and adds a main-specific guard so the normal path works under `set -u` when `RUN_MIGRATIONS_UNTIL_REMOVE_TENANTS` is unset.

## Validation

- `bash -n coprocessor/fhevm-engine/db-migration/initialize_db.sh`
- `git diff --check origin/main...HEAD`
- Stubbed `sqlx`/`psql` smoke test for normal and `RUN_MIGRATIONS_UNTIL_REMOVE_TENANTS=true` script branches

`cargo test -p tfhe-worker migrations --no-run` and the pre-commit `cargo check` could not complete locally because the machine ran out of disk space while writing Rust build artifacts.


coprocessor/fhevm-engine/Dockerfile.workspace
```diff
@@ -109,7 +109,7 @@ WORKDIR /app
 # Install sqlx-cli
 RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked \
     cargo install sqlx-cli --version 0.7.2 \
-    --no-default-features --features postgres,rustls --locked
+    --no-default-features --features postgres --locked
```
antoniupop: This looks unrelated to this branch - and this will break IAM auth on RDS.

antoniupop (commented):
Looks good aside from removing TLS

PanGan21 (dismissed):

coprocessor/fhevm-engine/Dockerfile.workspace
Eikix: Indeed it was done because of a dep CI failure

PanGan21 (approved):

antoniupop (approved):

Eikix:
@mergifyio queue

Eikix:
@mergifyio queue

Eikix:
@mergifyio queue

Eikix:
@mergifyio refresh

Eikix:
@mergifyio queue

Eikix:
@mergify dequeue

Eikix:
@mergify queue

PR #2420, opened 2026-05-04
fix(test-suite): use receipt block number in reorg isolation test

PanGan21:
### Summary

Fixes a flaky multichain test where `evm_revert` on Chain A does not affect Chain B state would intermittently fail because `providerB.getBlockNumber()` can return a cached pre-transaction block number before ethers.js's internal polling interval fires. Replace it with the block number from the transaction receipt, which is always accurate.


rudy-6-4 (approved):

PanGan21:
@mergify queue

PR #2432, opened 2026-05-05
fix(coprocessor): suppress drift alerts during gw-listener replay

Eikix:
## Summary

- skip end-of-batch drift finalization while replay mode is active

## Why this differs from `release/0.12.x`

The `release/0.12.x` PR contains extra code because that branch still has the operator-triggered gw-listener replay path:

- `--replay-from-block`
- alias `--catchup-kms-generation-from-block`
- `--replay-skip-verify-proof`
- `replay_from_block` state passed through `run_get_logs`

That path rewinds `last_processed_block_num` and reprocesses old gateway logs through the normal polling loop. Without that branch-specific handling, historical `CiphertextCommits` logs can reach `DriftDetector::end_of_batch` as if they were live logs.

On `main`, the explicit replay CLI path was removed by #2347 (`refactor(coprocessor): remove key generation listening from gw-listener`). The remaining replay on `main` is startup drift-detector rebuild from `earliest_open_ct_commits_block` to `last_processed_block_num`, with a separate `end_of_rebuild` finalization point. Therefore this PR only hardens the shared `end_of_batch` invariant.

## Behavioral equivalence with `release/0.12.x`

The intended common effect is: historical gateway log replay must not emit live drift alerts during normal batch finalization.

- On `release/0.12.x`, #2431 applies that behavior to both startup rebuild replay and explicit `--replay-from-block` replay.
- On `main`, explicit `--replay-from-block` replay no longer exists, so this PR applies the shared `end_of_batch` replay guard only.

## Testing

- `cargo test -p gw-listener end_of_batch_is_noop_while_replaying --lib`
- `cargo check -p gw-listener`
- pre-commit: `cargo fmt`, cargo check, clippy

### DIFF
--- a/coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
+++ b/coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
@@ -564,6 +564,9 @@ impl DriftDetector {
     /// Finalize a normal log-polling batch: check deferred consensus results,
     /// alert on completed-without-consensus handles, and evict stale handles.
     pub(crate) async fn end_of_batch(&mut self, db_pool: &Pool<Postgres>) -> anyhow::Result<()> {
+        if self.replaying {
+            return Ok(());
+        }
         self.refresh_pending_consensus_checks(db_pool).await?;
         self.finalize_completed_without_consensus();
         self.evict_stale(Instant::now());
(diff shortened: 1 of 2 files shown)


Eikix:
Converting to draft as it's not ready yet anymore.

Eikix:
Closing this main PR for now.

The 0.12 backport addressed a concrete operator replay issue: `--replay-from-block` can reprocess historical gateway logs through the normal polling loop and re-emit drift alerts. That path no longer exists on main.

For main, the remaining replay-like path is startup drift-detector rebuild. Treating this as the same patch is not precise enough: the important main-branch question is broader startup-rebuild soundness, especially what happens if rebuild fails halfway. Continuing live polling with partial reconstructed drift state is not a strong enough model for high-severity drift alerts.

We should replace this with a dedicated main hardening PR/design:
- make startup rebuild all-or-nothing, or explicitly define safe partial-state behavior;
- avoid refiring historical drift alerts after restart/rebuild;
- eventually persist drift alert identity if report-once semantics must hold across process restarts and replays.

The backport remains useful and scoped. This main PR should not be merged as-is.

PR #2364, opened 2026-04-24
feat(host-contracts): add decryption signature invalidation to ACL

obatirou:
Closes https://github.com/zama-ai/fhevm-internal/issues/1304

PR #2438, opened 2026-05-06
fix(contracts): scope upgrade check matrix and add diagnostics

Eikix:
## Summary

- scopes `contracts-upgrade-version-check` matrix rows to the packages selected by `check-changes`
- keeps host-only Forge setup behind a host package condition instead of using matrix `include`
- adds failure diagnostics that print candidate commits touching the failed contract source and package build/config files between the baseline tag and target SHA

## Validation

- `npx prettier --check ci/upgrade-check/check.ts .github/workflows/contracts-upgrade-version-check.yml`
- `git diff --check`
- YAML parse check for `.github/workflows/contracts-upgrade-version-check.yml`
- locally reproduced the PR #2364 gateway failure against `v0.12.3`; diagnostics pointed at `ed5c787ca feat(gateway-contracts): unify user decryption (#2329)`

### DIFF
--- a/ci/upgrade-check/check.ts
+++ b/ci/upgrade-check/check.ts
@@ -1,6 +1,8 @@
 #!/usr/bin/env bun
 // Checks that upgradeable contracts have proper version bumps when bytecode changes.
 // Usage: bun ci/upgrade-check/check.ts <baseline-pkg-dir> <pr-pkg-dir>
+import { execFileSync } from "child_process";
+import { basename } from "path";
 
 import { collectUpgradeVersionResults } from "./lib";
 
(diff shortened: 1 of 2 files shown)


ci/upgrade-check/check.ts
```diff
@@ -13,6 +15,63 @@ if (!baselineDir || !prDir) {
...
+      `${pkg}/contracts/${contract}.sol`,
+    ]);
+    printCommits("Non-exhaustive candidate commits touching package build/config files:", baseRef, targetRef, [
+      `${pkg}/foundry.toml`,
+      `${pkg}/hardhat.config.ts`,
+      `${pkg}/package.json`,
+      `${pkg}/package-lock.json`,
+      `${pkg}/remappings.txt`,
+    ]);
+  } finally {
```
obatirou: why try/finally?
printCommits and console.log cannot fail

obatirou (dismissed):
only one nit else lgtm

ci/upgrade-check/check.ts
Eikix: thanks will fix

Eikix:
simplified the PR

obatirou (approved):

Eikix:
@mergifyio queue

PR #2440, opened 2026-05-06
fix(common): handle release branch docker retags

Eikix:
## Summary

Fix Docker change detection on release branch creation pushes, where GitHub sets `github.event.before` to the all-zero SHA.

The reusable check workflow now falls back to `github.sha` before calling `git rev-list`, and the reusable retag workflow skips empty or same-source retags.

## Validation

- Parsed all workflow YAML with PyYAML.
- Ran `git diff --check`.
- Locally verified the all-zero SHA fallback resolves to the current commit.


eudelins-zama (approved):

PanGan21 (approved):

Eikix:
@mergifyio queue

PR #2443, opened 2026-05-07
fix(coprocessor): fail gw-listener on drift rebuild errors

Eikix:
## Summary

- Make gw-listener startup drift-detector rebuild errors bubble out of `run_get_logs` instead of continuing with partial in-memory state.
- This lets the existing outer retry/backoff loop retry recovery and prevents live polling after incomplete rebuild.

## Validation

- `cargo fmt -p gw-listener --check`
- `cargo check -p gw-listener --manifest-path coprocessor/fhevm-engine/Cargo.toml`
- pre-commit hook: cargo check + CI-aligned clippy


antoniupop (approved):

Eikix:
The only thing that 'scares' me with this PR is that if there is an unpredictable failure mode for the drift-detector that triggers under the happy path (no drift, just a restart), then it might bring down the gw-listener and page someone as it'll halt the network

Eikix:
Closing for now

PR #2451, opened 2026-05-08
fix(kms-connector): accept legacy extra data

Eikix:
## What
Accept legacy empty/`0x00` KMS connector `extra_data` values as the default context.

## Why
The v0.13 KMS worker currently rejects `extraData=0x00` with `Unsupported extra_data version`, but old relayer / SDK paths can still emit `0x00` during rollout. That blocks user decryption in a mixed-version stack where KMS connector has moved before all request producers emit versioned extra data.

## Behavior
- `[]` maps to no explicit context id and no epoch
- `[0x00]` maps to no explicit context id and no epoch
- `0x01` and `0x02` parsing remains unchanged
- context validation is skipped only when no explicit context id is present
- other version bytes still error

The original `extra_data` bytes are still forwarded to KMS request payloads; this PR does not rewrite legacy `0x00` to a synthetic v1 value.

## Validation
- `cd kms-connector && cargo test -p connector-utils extra_data --lib`
- `cd kms-connector && cargo check -p kms-worker`
- `cd kms-connector && cargo fmt --check`
- `git diff --check`

### DIFF
--- a/.github/workflows/contracts-upgrade-version-check.yml
+++ b/.github/workflows/contracts-upgrade-version-check.yml
@@ -60,7 +60,9 @@ jobs:
     steps:
       - name: Skip unchanged package
         if: env.PACKAGE_CHANGED != 'true'
-        run: echo "${{ matrix.package }} did not change."
+        env:
+          PACKAGE: ${{ matrix.package }}
+        run: echo "$PACKAGE did not change."
 
       - name: Checkout PR branch
         if: env.PACKAGE_CHANGED == 'true'
--- a/.github/workflows/contracts-upgrade-version-check.yml
+++ b/.github/workflows/contracts-upgrade-version-check.yml
@@ -44,26 +44,32 @@ jobs:
   check:
     name: contracts-upgrade-version-check/${{ matrix.package }} (bpr)
     needs: check-changes
-    if: ${{ needs.check-changes.outputs.packages != '[]' }}
     permissions:
       contents: 'read' # Required to checkout repository code
     runs-on: ubuntu-latest
+    env:
+      PACKAGE_CHANGED: ${{ contains(fromJSON(needs.check-changes.outputs.packages), matrix.package) }}
     strategy:
       fail-fast: false
       matrix:
-        package: ${{ fromJSON(needs.check-changes.outputs.packages) }}
         include:
           - package: host-contracts
             extra-deps: forge soldeer install
           - package: gateway-contracts
             extra-deps: ''
     steps:
+      - name: Skip unchanged package
+        if: env.PACKAGE_CHANGED != 'true'
+        run: echo "${{ matrix.package }} did not change."
+
       - name: Checkout PR branch
+        if: env.PACKAGE_CHANGED == 'true'
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           persist-credentials: 'false'
 
       - name: Resolve baseline release tag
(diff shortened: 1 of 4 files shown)


kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -36,13 +37,16 @@ pub struct ExtraData {
 /// - Bytes 65..: optional additional data (ignored)
 ///
 /// Version `0x01` → epoch_id is `None` (caller should fall back to DEFAULT_EPOCH_ID).
-/// Empty, `[0x00]`, or any other version byte → error.
+/// Empty or `[0x00]` → no explicit context and no epoch.
 pub fn parse_extra_data(extra_data: &[u8]) -> anyhow::Result<ExtraData> {
-    let Some(version) = extra_data.first() else {
-        return Err(anyhow!("extra_data is empty"));
-    };
+    if extra_data.is_empty() || extra_data == [0x00] {
```
dd23, line 42: Nit: should we only match on exactly 1-Byte of `0x00`, or do we want to match on everything that starts with `0x00` but ignore what might come after?

dd23: I guess it's fine, but wanted to raise the question.

Eikix: I would imagine: 
only accept exactly 0x00 as this is 'deprecation' / backward compat support
and any other malformed payload should be refused

tawadaa (approved):

.github/workflows/contracts-upgrade-version-check.yml
```diff
@@ -44,26 +44,34 @@ jobs:
...
       fail-fast: false
       matrix:
-        package: ${{ fromJSON(needs.check-changes.outputs.packages) }}
         include:
           - package: host-contracts
             extra-deps: forge soldeer install
           - package: gateway-contracts
             extra-deps: ''
     steps:
+      - name: Skip unchanged package
```
jot2re, line 61: Just curious, why are these yml changes part of this PR?

dd23 (approved):
LGTM!

kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -16,13 +16,14 @@ pub const EXTRA_DATA_V2_LENGTH: usize = 65;
...
-    pub context_id: U256,
+    pub context_id: Option<U256>,
     pub epoch_id: Option<U256>,
 }
 
 /// Parses the `extra_data` bytes to extract a context ID and an optional epoch ID.
 ///
-/// Only versions `0x01` and `0x02` are accepted (rolling compatibility window).
+/// Versions `0x01` and `0x02` are accepted (rolling compatibility window).
+/// Empty extra_data and `0x00` are legacy default-context markers.
```
jot2re, line 26: Nitpick: should the comment not mention that empty and `0x00` is also accepted?

jot2re (approved):
Looks good! Just added a couple of clarifying comments

.github/workflows/contracts-upgrade-version-check.yml
Eikix: somehow this job was blocking the pr 
i had to include that as a fix

kms-connector/crates/utils/src/types/extra_data.rs
Eikix: True! But since we want to deprecate it we don't need to advertise it's accepted

PR #2458, opened 2026-05-11
fix(kms-connector): accept legacy extra data

Eikix:
Ports the full #2451 PR from `release/0.13.x` to `main`.

Included commits:
- `fix(kms-connector): accept legacy extra data`
- `fix(ci): keep required contract checks stable`
- `fix(ci): avoid workflow template expansion in shell`

What changed:
- Accept empty `extra_data` and legacy `0x00` as default context markers.
- Keep context validation only for versioned extra data carrying an explicit context id.
- Normalize `0x00` to empty bytes before user-decryption KMS signing, preserving relayer-sdk <=0.4.2 signature verification behavior.
- Keep `contracts-upgrade-version-check` required checks stable for both contract packages, while skipping unchanged package work inside the job.
- Avoid workflow template expansion in the skip step shell command.

Equivalence checks:
- `git range-diff 14054f9^..origin/pr-2451-head 302834c^..HEAD`
  - commit 1: exact patch match
  - commit 2: same CI change adapted to current `main` workflow drift
  - commit 3: exact patch match
- `git patch-id --stable` matches for the KMS connector fix and the zizmor/template-expansion fix.

Validation:
- `cargo fmt --check`
- `cargo test -p connector-utils extra_data`
- `cargo test -p kms-worker kms_decryption_extra_data`
- `zizmor .github/workflows/contracts-upgrade-version-check.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/contracts-upgrade-version-check.yml
+++ b/.github/workflows/contracts-upgrade-version-check.yml
@@ -60,7 +60,9 @@ jobs:
     steps:
       - name: Skip unchanged package
         if: env.PACKAGE_CHANGED != 'true'
-        run: echo "${{ matrix.package }} did not change."
+        env:
+          PACKAGE: ${{ matrix.package }}
+        run: echo "$PACKAGE did not change."
 
       - name: Checkout PR branch
         if: env.PACKAGE_CHANGED == 'true'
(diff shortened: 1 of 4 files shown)


PanGan21 (approved):

Eikix:
@mergifyio queue

leventdem (approved):
OK!

jot2re (approved):
Also looks good

PR #2460, opened 2026-05-11
test(test-suite): strengthen multichain isolation coverage

Eikix:
## Summary

- replace the multichain handle isolation negative path with a valid Chain A encrypted input and proof before replaying it on Chain B
- assert the Chain A ACL grant is present on Chain A, absent on Chain B, and does not authorize Chain B user decryption
- keep the test scoped to the existing multichain isolation fixture

## Why

The previous checks could pass without proving the intended isolation properties: the handle case used malformed proof data, and the ACL case only checked ordinary Chain B balance state.

Fixes zama-ai/fhevm-internal#1234.

## Validation

- `npx prettier --check test-suite/e2e/test/multiChain/multiChainIsolation.ts`
- `git diff --check`

Not completed locally:

- `./node_modules/.bin/tsc --noEmit -p test-suite/e2e/tsconfig.json` currently fails on pre-existing checkout issues, including missing generated `test-suite/e2e/types` typechain modules and existing config/test typing errors.
- Runtime `multi-chain-isolation` e2e was not run because no multichain stack is running in this checkout.

### DIFF
--- a/test-suite/e2e/test/multiChain/multiChainIsolation.ts
+++ b/test-suite/e2e/test/multiChain/multiChainIsolation.ts
@@ -175,40 +175,56 @@ describe('Multi-Chain State Isolation', function () {
       const erc20A = this.chainA.erc20 as unknown as EncryptedERC20;
       const erc20B = this.chainB.erc20 as unknown as EncryptedERC20;
 
-      const handleA = await erc20A.balanceOf(this.deployerA.address);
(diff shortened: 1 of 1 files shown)


PanGan21 (approved):
Makes sense, my bad for overlooking it

Eikix:
@mergifyio queue

Eikix:
@mergifyio dequeue

Eikix:
@mergify queue

PR #2464, opened 2026-05-11
fix(coprocessor): improve drift recovery resilience

dartdart26:
If a coprocessor service loses its DB connection, it might miss a drift revert signal. When DB connection recovers, it might continue operating with stale data in memory, without having re-execed.

In order to minimize the chance of this happening, make sure services exit if the DB has been down for 60 seconds. Also, require that the drift revert grace period is at least twice that. Exiting in such cases means that the service will observe the signal after restart, if the DB connection is back up again. And if it doesn't observe it, it has restarted, so it should start from a good DB state.

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
@@ -21,6 +21,35 @@ const REVERT_SQL_TEMPLATE: &str =
 /// How often services poll `drift_revert_signal` for state changes.
 pub const POLL_INTERVAL: Duration = Duration::from_secs(2);
 
+/// Per-iteration bound on each signal-poll query.
+pub const POLL_QUERY_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Cumulative limit on DB unreachability. If the watcher cannot reach the
+/// DB (no successful poll) for longer than this, the process exits so the
+/// supervisor restarts it fresh. Prevents a service from running with stale
+/// in-memory state through a long DB outage during which the drift runner
+/// may have completed a revert without it noticing.
+pub const DRIFT_REVERT_DB_DOWN_LIMIT: Duration = Duration::from_secs(60);
+
+/// The revert runner's grace period must be at least this many times
+/// `DRIFT_REVERT_DB_DOWN_LIMIT`.
+pub const MIN_GRACE_PERIOD_MULTIPLIER: u32 = 2;
+
+#[derive(Clone, Copy, Debug)]
+pub struct WatcherTimeouts {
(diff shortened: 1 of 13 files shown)


coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
```diff
@@ -362,16 +391,46 @@ pub async fn on_drift_detected(pool: &Pool<Postgres>, handle: &[u8], host_chain_
...
-        match oldest_in_flight_signal(pool).await {
-            Ok(Some(signal)) => return Ok(signal),
-            Ok(None) => {}
-            Err(e) => {
-                error!(error = %e, "Drift-revert watcher poll failed, retrying");
+        match tokio::time::timeout(timeouts.poll_query_timeout, oldest_in_flight_signal(pool)).await
+        {
+            Ok(Ok(Some(signal))) => return signal,
+            Ok(Ok(None)) => {
+                last_success = std::time::Instant::now();
```
rudy-6-4, line 410: could be a continue here as no time check is required

rudy-6-4 (approved):

coprocessor/fhevm-engine/gw-listener/src/lib.rs
```diff
@@ -60,7 +61,7 @@ impl Default for ConfigSettings {
             gateway_config_address: None,
             drift_no_consensus_timeout: Duration::from_secs(5),
             drift_post_consensus_grace: Duration::from_secs(2),
-            drift_auto_revert_grace_period: Duration::from_secs(60),
+            drift_auto_revert_grace_period: Duration::from_secs(120),
```
rudy-6-4, line 64: in what cases the end user see the copro stale for 2 minutes ? (eg user cannot decrypt new ciphertext)
Ie how many copro must have drifted for that to happen ?

dartdart26: The whole auto reversal feature assumes a minority of coprocessors are drifting, i.e. a drift is considered a drift for auto reversal only if consensus has been reached. So to your question - users should not be observing it.

dartdart26:
E2E: https://github.com/zama-ai/fhevm/actions/runs/25738899839

dartdart26:
@mergifyio queue

coprocessor/fhevm-engine/fhevm-engine-common/src/drift_revert.rs
dartdart26: I think in theory it is better we keep the check, should this loop doesn't see the signal. This is in theory... continue is good as well in practice, AFAICT. I will leave as is for now, unless you feel strongly against it.

PR #2465, opened 2026-05-11
feat(test-suite): add native runner and binary replace

Eikix:
## Summary

This PR adds two related test-suite developer workflows:

1. `fhevm-cli test --runner native` runs the checked-out `test-suite/e2e` workspace directly against an existing Docker stack, while keeping Docker as the default runner.
2. `fhevm-cli replace` hot-swaps selected Rust service binaries into running stack containers for a narrower debug loop.

## Native e2e runner

- Adds `--runner docker|native` to `fhevm-cli test`; `docker` remains the default.
- Reuses `test-suite/e2e/run-tests.sh` for the native path instead of duplicating e2e execution logic.
- Rewrites generated Docker DNS URLs to host-reachable localhost endpoints for native execution.
- Pins `DOTENV_CONFIG_PATH` to the generated CLI env so stale local e2e dotfiles do not leak into native runs.
- Adds a fetch-time MinIO URL rewrite for the relayer SDK path that reads object URLs such as `http://minio:9000/...` off-chain.

## Binary replace command

- Adds `fhevm-cli replace` for copying local service binaries into matching running containers and restarting only those containers.
- Supports `replace --local-build`, which runs a cached Linux Cargo build in Docker's Linux VM.
- Stores Cargo registry, git, and target caches under `.fhevm/runtime/replace-build`.
- Supports `--build-profile dev|release`; `dev` is the default for fast iteration.
- Supports `--binary <path>` as the explicit prebuilt-binary escape hatch.
- Does not rerender Compose, run migrations, change versions, or alter stack state beyond the targeted container restart.

## CI smoke

When the e2e workflow already runs in the normal non-compat/non-orchestrated path, it now also runs a small native `input-proof` smoke.

The PR-trigger gate is unchanged: direct same-repo PRs still need the `e2e` label for this workflow to run.

## Validation

- `bun run check`
- `bun test src/cli.test.ts`
- `bun test src`
- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/test-suite-e2e-tests.yml"); puts "yaml ok"'`
- `./fhevm-cli replace --help && ./fhevm-cli test --help`

Refs zama-ai/fhevm-internal#1346

### DIFF
--- a/test-suite/fhevm/src/commands/test.ts
+++ b/test-suite/fhevm/src/commands/test.ts
@@ -1,12 +1,13 @@
 /**
  * Runs named e2e test profiles, standard/heavy CI suites, and topology-specific test flows.
  */
+import path from "node:path";
 import { compatPolicyForState, supportsCoprocessorDbStateRevert } from "../compat/compat";
 import { DRIFT_CLEANUP_SQL, DRIFT_INSTALL_SQL, driftDatabaseName, parseDriftInstanceIndex, parsePositiveInteger } from "../drift";
 import { PreflightError, formatCliError } from "../errors";
 import { dockerInspect } from "../flow/readiness";
-import { pause, shellEscape, unpause } from "../flow/up-flow";
-import { hostReachableRpcUrl } from "../utils/fs";
+import { pause, unpause } from "../flow/up-flow";
+import { hostReachableMaterialUrl, hostReachableRpcUrl, readEnvFile, writeEnvFile } from "../utils/fs";
 import { run, runWithHeartbeat } from "../utils/process";
 import { loadState } from "../state/state";
 import { topologyForState } from "../stack-spec/stack-spec";
@@ -16,9 +17,11 @@ import {
   DEFAULT_POSTGRES_DB,
   DEFAULT_POSTGRES_PASSWORD,
   DEFAULT_POSTGRES_USER,
+  envPath,
   HEAVY_TEST_PROFILES,
   LIGHT_TEST_PROFILES,
   POSTGRES_HOST,
+  REPO_ROOT,
   STANDARD_TEST_PROFILES,
   TEST_GREP,
   TEST_PARALLEL,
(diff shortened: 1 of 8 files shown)


Eikix (commented):
[AUTOMATED]

## Code Review

Reviewed the native e2e runner, the new `replace` command, and the workflow gating. Solid PR — `replace.ts` is well-structured, the runner switching in `test.ts` is clean, and the `NATIVE_SMOKE_ONLY` gating logic in the workflow handles labelled/unlabelled PRs and mergify branches correctly.

**Items checked and cleared during review:**
- `docker cp` against a running container's executing binary (works — `docker cp` uses tar streaming, avoiding `ETXTBSY`).
- Build-profile injection surface in `replace.ts` (regex-gated, then passed via argv to `Bun.spawn` — no shell interpolation).
- `npm ci` running at repo root in the workflow (correct — root is the npm workspaces root and `test-suite/e2e` is a workspace; `node_modules` hoists upward).
- `waitForRestart` health-vs-running selection after `docker restart` (correct — `State.Health` is populated as soon as the image declares a healthcheck, and `postBootHealthGate` runs after).
- Workspace/package dedup key in cargo build planning (`host_listener_*` binaries correctly collapse into one `host-listener` cargo build).
- `NATIVE_SMOKE_ONLY` ternary in YAML evaluates correctly across all event/label combinations.

One minor nit inline.

test-suite/fhevm/src/commands/test.ts
```diff
@@ -1,12 +1,13 @@
 /**
  * Runs named e2e test profiles, standard/heavy CI suites, and topology-specific test flows.
  */
+import path from "node:path";
 import { compatPolicyForState, supportsCoprocessorDbStateRevert } from "../compat/compat";
```
Eikix, line 5: [AUTOMATED] Nit: `.prettierrc.yml` sets `importOrderSeparation: true` with groups `["<THIRD_PARTY_MODULES>", "^[./]"]`, so there should be a blank line between `import path from "node:path";` and the first local import. `replace.ts` and `rollout.ts` (also added in this PR) follow the convention; this file is the only outlier.

Confidence: 92/100

Eikix (commented):
[AUTOMATED]

## Re-review after fix commits

I pushed two follow-up commits (`cd2e75560`, `25a4aebd4`) to fix the failing `test-suite-e2e-tests/fhevm-e2e-test` job. Root causes and fixes:

**1. `compile:specific` permanently mutated `paths.sources`** — `test-suite/e2e/hardhat.config.ts:70` calls `compile:specific('examples')` on every `hardhat test` invocation, which left `hre.config.paths.sources` pointing at a nonexistent directory. The subsequent `runSuper()` auto-compile then saw zero `.sol` files (`Nothing to compile`), so no artifacts were produced and the test failed at `ethers.getContractFactory('TestInput')`. Docker mode masked this because artifacts were pre-baked into the test-suite image. Fix: wrap the mutation in try/finally so the sources path is always restored.

**2. `RPC_URL` (and friends) weren't reaching hardhat** — the native runner spread rewritten URLs into the spawn env and pointed `DOTENV_CONFIG_PATH` at the container-shaped `test-suite.env` (which has `RPC_URL=http://host-node:8546`). Hardhat's `dotenv.config()` uses default no-override behavior, so if spawn-env propagation hiccupped at all, hardhat fell back to the container hostname (or to its own default). Fix: `prepareNativeTestEnv` now writes a host-reachable `test-suite.native.env` to disk and points `DOTENV_CONFIG_PATH` at it; the rewritten map is also spread into the spawn env so any developer-side `RPC_URL` in the shell can't shadow it. Single source of truth on disk + explicit spawn-env override = deterministic across CI and local dev.

**Verification**: typecheck clean, 151/151 unit tests pass, local repro of the failing path now compiles 26 contracts and reaches the expected stack-connection point with no `RPC_URL not set` warning. CI is rerunning; will keep an eye on it.

Eikix:
Waiting on https://github.com/zama-ai/fhevm/pull/2479 to be able to properly rebase

Eikix:
@mergifyio refresh

Eikix:
@mergifyio rebase

Eikix:
Deprioritizing for now

PR #2469, opened 2026-05-11
feat: add KMS migration verification script

obatirou:
Assert migrated ProtocolConfig, KMSGeneration, and KMSVerifier state against the gateway state and emit DAO upgrade artifacts for prepared upgrades.

### DIFF
--- a/host-contracts/tasks/taskMigrate.ts
+++ b/host-contracts/tasks/taskMigrate.ts
@@ -1,6 +1,9 @@
-import { FunctionFragment, Interface, type InterfaceAbi, getAddress } from 'ethers';
+import { calculateERC7201StorageLocation } from '@openzeppelin/upgrades-core/dist/utils/erc7201';
+import { AbiCoder, FunctionFragment, Interface, type InterfaceAbi, getAddress, keccak256, toBeHex } from 'ethers';
+import fs from 'fs';
 import { task, types } from 'hardhat/config';
 import type { HardhatRuntimeEnvironment } from 'hardhat/types';
+import path from 'path';
 
 import {
   assertContractMatchesVersionPrefix,
@@ -126,15 +129,9 @@ function printPreparedDaoUpgrade(data: PreparedDaoUpgrade): void {
   );
 }
 
-function assertEqual(label: string, actual: string | number | boolean, expected: string | number | boolean): void {
+function assertEqual<T>(label: string, actual: T, expected: NoInfer<T>): void {
   if (actual !== expected) {
     throw new Error(`${label} mismatch: expected ${expected}, got ${actual}.`);
-  }
-}
-
-function assertBigIntEqual(label: string, actual: bigint, expected: bigint): void {
-  if (actual !== expected) {
-    throw new Error(`${label} mismatch: expected ${expected.toString()}, got ${actual.toString()}.`);
   }
 }
 
@@ -176,23 +173,6 @@ function normalizeKeyDigest(digest: KeyDigestLike): { keyType: number; digest: s
     keyType: Number(digest.keyType),
     digest: digest.digest.toLowerCase(),
   };
-}
-
-function storageUrlsForSenders(
-  nodes: readonly ProtocolConfigMigrationKmsNode[],
-  txSenders: readonly string[],
-): string[] {
-  const storageUrlsByTxSender = new Map(
-    nodes.map((node) => [getAddress(node.txSenderAddress), node.storageUrl] as const),
-  );
-
-  return txSenders.map((txSender) => {
-    const storageUrl = storageUrlsByTxSender.get(getAddress(txSender));
-    if (storageUrl === undefined) {
-      throw new Error(`Migration snapshot has consensus tx sender ${txSender} without a matching KMS node.`);
-    }
-    return storageUrl;
-  });
 }
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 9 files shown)


host-contracts/tasks/taskMigrate.ts
```diff
@@ -243,3 +350,201 @@ task(
...
   console.log('KMSGeneration migration code set successfully at address:', proxyAddress);
 });
+
+////////////////////////////////////////////////////////////////////////////////
+// Post-execution verification
+////////////////////////////////////////////////////////////////////////////////
+
+task(
+  'task:assertKmsMigrationSucceeded',
+  'Asserts the live host migration state matches the MIGRATION_* snapshot env',
```
isaacdecoded: nit: I'm wondering if this task should check the migrated state against the Gateway contracts state instead of the MIGRATION_* envs? I mean, the env variables could contain errors while the Gateway contracts could be more reliable. Just commenting, in any case.

isaacdecoded (approved):
LGTM!

host-contracts/tasks/taskMigrate.ts
obatirou: mmmh indeed, I was looking at it in the lens: "are the migration values from the export were correctly set" but having the check on gateway directly is a way better invariant.

isaacdecoded: I see. OTOH, you already defined a section "To go further: verify by hand the exported state" in the runbooks. So, it's probably enough to check against the migration env variables here, and leave the double-check to the steps you described in that section.

obatirou: yes that was the plan but I am afraid not all reviewers will do it as it is a real pain, manual and long.
We could have both without letting other reviewers knows there is a second automatic post deployment check

obatirou: Done 
If you think this is too much or not necessary please share
cc @Eikix

host-contracts/tasks/taskMigrate.ts
```diff
@@ -108,6 +129,58 @@ function printPreparedDaoUpgrade(data: PreparedDaoUpgrade): void {
   );
 }
 
+function assertEqual(label: string, actual: string | number | boolean, expected: string | number | boolean): void {
+  if (actual !== expected) {
+    throw new Error(`${label} mismatch: expected ${expected}, got ${actual}.`);
+  }
+}
+
+function assertBigIntEqual(label: string, actual: bigint, expected: bigint): void {
```
Eikix: 😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂😂

host-contracts/tasks/taskMigrate.ts
```diff
@@ -243,3 +330,352 @@ task(
...
+      const snapshotPath = path.resolve(hre.config.paths.root, '..', 'gateway-contracts', 'migration-state.json');
+      const snapshotBlock = fs.existsSync(snapshotPath)
+        ? (JSON.parse(fs.readFileSync(snapshotPath, 'utf8')) as { metadata?: { exportBlockNumber?: number } }).metadata
+            ?.exportBlockNumber
+        : undefined;
+      // Reject snapshots from a different chain (dev runs against an in-process hardhat node at block 0).
+      if (snapshotBlock !== undefined && snapshotBlock <= (await gatewayProvider.getBlockNumber())) {
+        gatewayBlockTag = snapshotBlock;
+      } else {
+        console.warn(
```
Eikix, line 448: [AUTOMATED] **Silent fallback to `'latest'` undermines the determinism of the verification (medium-high)**

When `migration-state.json` is absent or its `exportBlockNumber` exceeds the current Gateway chain head, the task falls through with a `console.warn` and continues using `'latest'` as the block tag. For a verification script, reading the Gateway against a moving `'latest'` can produce a false-positive "KMS migration verification succeeded." if Gateway state has evolved beyond the migration snapshot between export and verification.

The `--gateway-block-tag` param description ("Defaults to metadata.exportBlockNumber … else latest with a warning.") signals this is intentional, but for non-`hardhat` networks it would be safer to require a deterministic anchor:

```ts
if (hre.network.name !== 'hardhat') {
  throw new Error(
    'Cannot determine Gateway block tag: migration-state.json is missing or its exportBlockNumber is ahead of the current chain. ' +
    'Pass --gateway-block-tag explicitly.',
  );
}
console.warn('Gateway block tag defaulting to "latest" — pass --gateway-block-tag for deterministic verification.');
```

Confidence: 85/100

host-contracts/tasks/taskMigrate.ts
```diff
@@ -243,3 +330,352 @@ task(
...
+
+    // On hardhat the mock Gateway shares the in-process network; everywhere else require the URL.
+    const rpcUrl = hre.network.name === 'hardhat' ? process.env.GATEWAY_RPC_URL : getRequiredEnvVar('GATEWAY_RPC_URL');
+    const gatewayProvider = rpcUrl ? new hre.ethers.JsonRpcProvider(rpcUrl) : hre.ethers.provider;
+
+    let gatewayBlockTag: number | 'latest' = taskArgs.gatewayBlockTag ?? 'latest';
+    if (gatewayBlockTag === 'latest') {
+      const snapshotPath = path.resolve(hre.config.paths.root, '..', 'gateway-contracts', 'migration-state.json');
+      const snapshotBlock = fs.existsSync(snapshotPath)
+        ? (JSON.parse(fs.readFileSync(snapshotPath, 'utf8')) as { metadata?: { exportBlockNumber?: number } }).metadata
```
Eikix, line 441: [AUTOMATED] **Nit: `JSON.parse` on `migration-state.json` has no `try/catch`**

`fs.readFileSync` + `JSON.parse` here is only guarded by `fs.existsSync`. If the file exists but is malformed (partial write, corruption), `JSON.parse` throws a bare `SyntaxError` with no indication the snapshot file is the source. Wrapping the parse in a `try/catch` that prints the snapshot path before falling through to the `'latest'` warn path would make runbook debugging easier:

```ts
let snapshotBlock: number | undefined;
try {
  const raw = JSON.parse(fs.readFileSync(snapshotPath, 'utf8')) as { metadata?: { exportBlockNumber?: number } };
  snapshotBlock = raw.metadata?.exportBlockNumber;
} catch (err) {
  console.warn(`Failed to parse Gateway snapshot at ${snapshotPath}: ${(err as Error).message}`);
}
```

Confidence: 82/100

host-contracts/tasks/taskMigrate.ts
```diff
@@ -243,3 +330,352 @@ task(
...
+      }
+    }
+    const gatewayCallOpts = { blockTag: gatewayBlockTag };
+
+    const gatewayConfig = new hre.ethers.Contract(gatewayConfigProxy, GATEWAY_CONFIG_VIEW_ABI, gatewayProvider);
+    const gatewayKmsGeneration = new hre.ethers.Contract(
+      gatewayKmsGenerationProxy,
+      GATEWAY_KMS_GENERATION_VIEW_ABI,
+      gatewayProvider,
+    );
```
Eikix, line 460: [AUTOMATED] **Consider: no identity/version sanity check on the Gateway proxies (defensive gap)**

The host-side `protocolConfig` / `kmsGeneration` / `kmsVerifier` are guarded by `assertContractMatchesVersionPrefix` before any state is compared. The two Gateway proxies are bound directly via `new ethers.Contract(… GATEWAY_*_VIEW_ABI …)` with no equivalent guard. If an operator pastes the wrong address (e.g., `GatewayConfig` where `KMSGeneration` is expected), most view calls return ABI-decoded zeros / empty arrays — and the host-side equality could trivially pass for an empty or partial migration if the host is itself in an early state.

A cheap sanity check would be to assert each Gateway proxy has code at `gatewayBlockTag` (`gatewayProvider.getCode(addr, gatewayBlockTag)` non-empty) and ideally a probe call specific to each contract (e.g., `getCurrentKmsContextId` on `GatewayConfig`, a counter read on `KMSGeneration`) before declaring success.

Confidence: 80/100

Eikix (commented):
[AUTOMATED]

## Code Review — PR #2469: KMS migration verification script

Reviewed by parallel agents (bug hunters, guideline compliance, error handling, protocol-compatibility).

**No protocol-boundary risk.** All edits stay within Hardhat tooling for migration/upgrade verification. Storage offsets, mapping-slot encoding, host-vs-Gateway ABI mirroring, and the `upgradeToAndCall(address,bytes)` outer-calldata construction all check out against the actual `KMSGeneration` / `ProtocolConfig` layouts and the exporter on this branch.

**Three non-blocking findings**, all on `host-contracts/tasks/taskMigrate.ts`. The script's role as a verification tool means "silently succeeds when state is unknowable" is the failure mode worth tightening. Details inline.

Additional notes (informational, not flagged inline):
- `gateway-contracts/tasks/exportMigrationState.ts` switched `MIGRATION_KMS_THRESHOLDS.*` serialization from `string` to `number`, and dropped `metadata.legacyHostKmsVerifierProxy` from the export envelope — worth a heads-up for any downstream consumer of `migration-state.json`.
- `verifyPreparedImplementation` unconditionally sleeps 2 min before calling `verify:verify` (taskMigrate.ts:100). Matches the prior production path, but on local devnets that don't pass `verifyContract: false` it burns 2 min before a guaranteed Etherscan failure.
- The prepared DAO upgrade artifact (proxy/newImpl/innerCalldata/outerCalldata) is printed via `console.log` only — never persisted to disk like `exportMigrationState.ts` does. Running two prepare tasks back-to-back leaves the operator dependent on terminal buffer. Consider writing a JSON artifact to `host-contracts/dao-upgrades/<contract>-<timestamp>.json` for auditability.

Eikix (approved):
claude review is mostly nits, up to you to fix or not

isaacdecoded (approved):

PR #2470, opened 2026-05-11
feat(common): implement mulDiv operator

PanGan21:
### Summary

- Adds `FHE.mulDiv` as a native coprocessor opcode (op 30). 
- Solidity emits a single `FheMulDiv` event carrying lhs, rhs, plaintext divisor, and scalarByte; the coprocessor executes the fused mul + scalar div using tfhe-rs.
- Supported types: euint8, euint16, euint32, euint64 (both enc×enc and enc×scalar).
- Divisor is always plaintext
- HCU prices: fheMulDiv[T] = fheMul[2T] + fheDiv[2T] (tfhe-rs widens to 2T internally).

Closes: https://github.com/zama-ai/fhevm-internal/issues/1352


PanGan21:
Successful e2e test run for `mulDiv` operations: https://github.com/zama-ai/fhevm/actions/runs/25732228026

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
```diff
@@ -1849,6 +1849,86 @@ pub fn get_op_size_on_gpu(
...
+            };
+            assert_eq!(input_operands.len(), 3);
+            match (&input_operands[0], &input_operands[1], &input_operands[2]) {
+                (
+                    SupportedFheCiphertexts::FheUint8(a),
+                    SupportedFheCiphertexts::FheUint8(b),
+                    SupportedFheCiphertexts::Scalar(d),
+                ) => Ok(widened_workspace(
+                    a.get_mul_size_on_gpu(b),
+                    a.get_div_size_on_gpu(to_be_u8_bit(d)),
```
rudy-6-4, line 1865: why the div size is not done on the product size ? it's what is accounted by the widened_workspace ?

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -804,6 +804,78 @@ pub fn check_fhe_operand_types(
...
+                            fhe_operation: format!(
+                                "{:?}: divisor operand (index 2) must be a scalar",
+                                fhe_op
+                            ),
+                            input_types: vec![],
+                        });
+                    }
+                    let lhs_type = get_ct_type(&input_handles[0])?;
+                    let lhs_width_bytes: usize = match lhs_type {
+                        2 => 1,
```
rudy-6-4: Need either naming or comments (indicating named type). Don't we have an enum for that type of match ? If not we should have one. BTW width_bytes_from_type could  be a function.
Also it does 2 different things:
E.g.
if !is_valid_lhs_type_for_div(lhs_type) {
   return Err...
}
let lhs_width_bytes = width_bytes_from_type(lhs_type);

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -804,6 +804,78 @@ pub fn check_fhe_operand_types(
...
+                            got_operands: input_handles.len(),
+                        });
+                    }
+                    if is_input_handle_scalar[0] {
+                        return Err(FhevmError::FheOperationOnlySecondOperandCanBeScalar {
+                            scalar_input_index: 0,
+                            only_allowed_scalar_input_index: 1,
+                        });
+                    }
+                    if !is_input_handle_scalar[2] {
```
rudy-6-4: could be in order by checking input 0, 1 and 2
Here there is a check on 0 then 2 then 0 (lhs) then 1

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -804,6 +804,78 @@ pub fn check_fhe_operand_types(
...
+                        if rhs_type != lhs_type {
+                            return Err(FhevmError::FheOperationDoesntHaveUniformTypesAsInput {
+                                fhe_operation,
+                                fhe_operation_name: format!("{:?}", fhe_op),
+                                operand_types: vec![lhs_type, rhs_type],
+                            });
+                        }
+                    }
+                    let start = input_handles[2]
+                        .len()
```
rudy-6-4: A comparison would be more straitforward and avoid the ok_or_else lambda nesting.
Eg
let divisor_size = ...
if divisor_size < lhs_width_bytes  {
   return Err...
}
let start = divisor_size -  lhs_width_bytes;

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -804,6 +804,78 @@ pub fn check_fhe_operand_types(
...
+                        .checked_sub(lhs_width_bytes)
+                        .ok_or_else(|| FhevmError::UnsupportedFheTypes {
+                            fhe_operation: format!(
+                                "{:?}: divisor operand is shorter than operand width",
```
rudy-6-4: why is it a condition ? the operand is a constant it does not have a maximum width.

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -804,6 +804,78 @@ pub fn check_fhe_operand_types(
...
+                                fhe_op
+                            ),
+                            input_types: vec![],
+                        })?;
+                    let divisor_low = &input_handles[2][start..];
+                    if divisor_low.iter().all(|b| *b == 0) {
```
rudy-6-4, line 866: Ok, undersood now, so start could be renamed ignored_hsb or someting more explicit.
And there should be no error if no ignored_hsb exists !?

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -3302,6 +3376,72 @@ pub fn perform_fhe_operation_impl(
                 }),
             }
         }
+        SupportedFheOperations::FheMulDiv => {
+            assert_eq!(input_operands.len(), 3);
```
rudy-6-4: ?? return an error here. Better than a panic.

coprocessor/fhevm-engine/fhevm-engine-common/src/gpu_memory.rs
PanGan21: You're right that the scalar div actually runs on the 2T wide product, not on `a`.
The rationale why use `a` is that this is the only handle we have here and don't construct the wide product just to measure it.

The ×4 still gives the right approximation because FHE scales as O(blocks^2), and widening T→2T doubles the block count: (2 × blocks)^2 = 4 × (blocks^2). Same factor for mul and div.

An alternative is to trivially encrypt a wide ciphertext if you think the approach is not good enough. It drops the ×4 but costs a real allocation and duplicates tfhe-rs's widening table. 
Which I thought is not worth it for a size estimate.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -900,9 +901,35 @@ impl SupportedFheOperations {
...
+    /// an encrypted ciphertext handle.
+    ///
+    /// Plaintext operands always come at the end of the operand list. Two things
+    /// can make an operand plaintext:
+    /// 1. The operation has a fixed plaintext tail. For example, `FheMulDiv` has
+    ///    operands `[lhs, rhs, divisor]` and `divisor` is always plaintext.
+    /// 2. The work item is marked scalar (`is_work_scalar == true`), which makes
+    ///    the rhs plaintext too — e.g. `fheAdd(encrypted, plaintext)`.
+    pub fn is_operand_scalar(&self, is_work_scalar: bool, idx: usize, n_deps: usize) -> bool {
+        let trailing = self.num_trailing_always_scalar_operands(n_deps) + is_work_scalar as usize;
```
rudy-6-4: This is cryptic. Why having an intermediate integer ?
Eg. Is it equivalent to
```
SupportedFheOperations::FheRand
            | SupportedFheOperations::FheRandBounded
            | SupportedFheOperations::FheTrivialEncrypt => True
SupportedFheOperations::FheMulDiv => idx == 2 || (idx == 1 && is_work_scalar)
_ => idx == 1 && is_work_scalar
```

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2470/commits/25dab96e3f0e4b1bb212ffe08078f060c3a72c3c

rudy-6-4 (commented):
Small improvements asked but nothing blocking to merge.

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2470/changes/32dccc10273673b56898009a52fa18ff15b5ec97

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2470/commits/0d4c9f7fb5e847f38786f4fa674aa817bf05b2fd

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Fixed in: https://github.com/zama-ai/fhevm/pull/2470/commits/fb65d4da95a36c12e88b1a6ebfd6a9e440ef16ac

It was there as defensive validation. If the divisor would ever be different than 32 bytes this will panic. however this is unreachable assuming that L1 only emits with bytes32 divisor

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2470/commits/2ae1b5732fd43d418598fb016b0a3a0b126640dc

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2470/commits/bf25971fe636042eb727f0f1a437777b075557d5

However there are multiple cases where we use `assert_eq` the same way: 
- [FheSub](https://github.com/zama-ai/fhevm/blob/c2d1a202aa5fea04c34ba62780b1a1e33a307f00/coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs#L960)
- [FheMul](https://github.com/zama-ai/fhevm/blob/c2d1a202aa5fea04c34ba62780b1a1e33a307f00/coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs#L1021)
- etc

So this is exception in the established patterns of this function

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
PanGan21: Good catch. Refactored in: https://github.com/zama-ai/fhevm/pull/2470/changes/84e932442d31fbec4233901c500b2deda7c8a56f

PanGan21:
@claude /pr-review

rudy-6-4 (dismissed):

rudy-6-4 (approved):

PanGan21:
@mergify queue

PR #2485, opened 2026-05-13
ci(security): resolve workflow zizmor findings

Eikix:
## Summary

- bump the CI `zizmor` version from `1.17.0` to `1.23.1`
- remove the inline `zizmor: ignore` from `unverified_prs` by fixing its `concurrency-limits` finding
- fix `secrets-outside-env` findings by assigning secret-consuming jobs to the existing `main` GitHub environment
- fix `superfluous-actions` findings by replacing `dtolnay/rust-toolchain` steps with direct `rustup` commands that preserve the requested toolchain and components
- fix remaining pedantic `concurrency-limits` and `anonymous-definition` findings

## Scope rule

Every workflow behavior change in this PR corresponds to a `zizmor --persona pedantic` finding on `origin/main`, except for the `zizmor` version bump itself.

## Validation

- `zizmor --persona pedantic .github/workflows` with local `zizmor 1.23.1`
- `zizmor .github/workflows` with local `zizmor 1.23.1`
- `ruby -e 'require "yaml"; Dir[".github/workflows/*.{yml,yaml}"].sort.each { |f| YAML.load_file(f) }; puts "yaml ok"'`
- `git diff --check`
- staged pre-commit workflow hook

`actionlint` was not available in the local environment.

## Security follow-up

The workflow code now satisfies `zizmor` 1.23.1 with the pedantic persona, but GitHub settings still matter. The `main` environment already exists in `zama-ai/fhevm`; for secret scoping to be effective beyond scanner compliance, the relevant secrets should be configured as environment-scoped secrets and the required protection policy should be decided.

Tracks broader settings cleanup in zama-ai/fhevm-internal#1381.

### DIFF
--- a/.github/workflows/charts-helm-checks.yml
+++ b/.github/workflows/charts-helm-checks.yml
@@ -5,6 +5,10 @@ on:
 
 permissions: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
 env:
   HELM_VERSION: v3.16.4
 
--- a/.github/workflows/charts-helm-release.yml
+++ b/.github/workflows/charts-helm-release.yml
@@ -8,6 +8,10 @@ on:
 
 permissions: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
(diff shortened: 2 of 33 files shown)


Eikix:
This draft is being closed because the  findings need to be handled as a security-owned cleanup after the full report and environment model are reviewed. The findings have been documented in zama-ai/fhevm-internal#1381.

PR #2495, opened 2026-05-13
feat(coprocessor): host-listener-consumer, metrics for latency, gaps,…

rudy-6-4:
… duplicate


coprocessor/fhevm-engine/host-listener/src/consumer/metrics.rs
```diff
@@ -21,6 +24,38 @@ pub(crate) static DB_ERRORS: LazyLock<IntCounterVec> = LazyLock::new(|| {
     .expect("host_consumer_db_errors metric must register")
 });
 
+pub(crate) static BLOCKS_MISSING: LazyLock<IntCounterVec> =
+    LazyLock::new(|| {
+        register_int_counter_vec!(
+            "host_consumer_blocks_processed",
```
obatirou: host_consumer_blocks_missing ?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -594,6 +601,117 @@ impl Database {
...
+            self.chain_id.as_i64(),
+            finalization_margin,
+        )
+        .fetch_one(&pool)
+        .await?;
+
+        Ok(StatsForConsumer {
+            number_of_new_gaps: row.number_of_new_gaps,
+            total_new_gap_size: row.total_new_gap_size,
+            number_of_duplicated_inserts: row.number_of_duplicated_inserts,
```
nboisde, line 694: Ok so if i understand correctly here, those numbers will reflect if the new strategy inserted blocks the first one.
For example if the number of inserted blocks is 1000 and the number_of_duplicated_inserts is 200, it means that 20 percent of it was inserted by another strategy than the new listener ?

nboisde (approved):

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: The number in StatsForConsumer are only about duplicated and missing, do you want to add the total number off inserted block to do a percentage more easily ? I expect all these number to be 0 (particularly for missing blocks). The numbers are just to prove they stay at 0 or detect we have something to fix and verify the fix.

rudy-6-4: There is already BLOCKS_PROCESSED so no need to add it here.

nboisde (dismissed):
LGTM !

rudy-6-4:
@mergifyio queue

nboisde (approved):
Overall lgtm for capturing the metrics. Thanks for this work @rudy-6-4 !

Eikix (approved):

PR #2496, opened 2026-05-13
fix(coprocessor): missing index for host_chain_blocks_valid

rudy-6-4:
Index :
    "blocks_valid_pkey" PRIMARY KEY, btree (chain_id, block_hash)

UPDATE host_chain_blocks_valid
SET block_status = CASE
WHEN block_hash = $2
   THEN 'finalized'
   ELSE 'orphaned'
   END
WHERE block_number = $3 AND chain_id = $1


Eikix:
Ahh I frontran you with https://github.com/zama-ai/fhevm/pull/2499 🙏

PR #2509, opened 2026-05-14
feat(coprocessor): support compressed xof key sets

antoniupop:
Add host-listener support for ingesting CompressedXofKeySet server-key blobs from S3, while preserving the existing DB/runtime contract by normalizing them into ServerKey / CompressedServerKey rows.

Existing/legacy uncompressed keys as well as the LFS test keys remain compatible through the legacy-column fallback.

  - We now expect by default a CompressedXofKeySet in S3
    - On 404 from S3 on compressed, try reading uncompressed.
  - Download uses /CompressedXofKeySet/{key_id}
  - Digest verification remains on the downloaded S3 bytes before deserialization/decompression.
  - CompressedXofKeySet stored in full in DB and decompression happens on full keyset first before we split it
  - Readers prefer compressed DB columns when present, with legacy fallback.
  - Migration for new compressed key columns and tests for XOF format handling.


coprocessor/fhevm-engine/fhevm-engine-common/src/db_keys.rs
```diff
@@ -63,8 +63,15 @@ impl DbKeyCache {
         }
 
         // Only fetch the heavy key blobs when the latest key is not already cached.
+        // We pull just one of sks_key / sks_key_compressed per row: COALESCE
+        // picks the compressed blob when populated. The is_compressed flag tells the
+        // deserializer which encoding came back.
         let row = sqlx::query(
```
rudy-6-4: Should we move to compiled query ? We have recent code that doesn't use them.
That a bit safer and more efficient (not important for that query).

coprocessor/fhevm-engine/fhevm-engine-common/src/db_keys.rs
```diff
@@ -188,7 +214,29 @@ impl DbKeyCache {
         #[cfg(feature = "gpu")]
         {
             let num_gpus = get_number_of_gpus() as u64;
-            let csks: tfhe::CompressedServerKey = safe_deserialize_key(&sks_key)?;
+            // The GPU path needs a CompressedServerKey. The XOF ingest
+            // path writes one to sks_key_compressed (is_compressed =
+            // true); the legacy-ServerKey fallback path writes a plain
+            // ServerKey to sks_key (is_compressed = false) which the
+            // deserialize below cannot consume. The LFS test fixture
```
rudy-6-4: should we support the old LFS test and not have it write in sks_key_compressed ? it's only for test backward compatibility (vs real prod backward compatibility)?

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
```diff
@@ -238,23 +239,33 @@ pub(crate) async fn activate_ready_key_activations(
...
                 e.chain_id = $1
                 AND e.block_hash = $2
                 AND e.key_id = $3
+                -- For now test on legacy uncompressed as always available
                 AND e.key_content_public IS NOT NULL
                 AND e.key_content_sks_key IS NOT NULL
             ON CONFLICT (chain_id, block_hash, key_id_gw) DO UPDATE
-            SET pks_key   = EXCLUDED.pks_key,
-                sks_key   = EXCLUDED.sks_key,
-                sns_pk    = COALESCE(EXCLUDED.sns_pk, keys.sns_pk),
```
rudy-6-4, line 256: I am not sure why this one was a COALESCE.

coprocessor/fhevm-engine/host-listener/src/kms_generation/sks_key.rs
```diff
@@ -42,10 +150,201 @@ pub fn extract_server_key_without_ns(
...
+        decompression_key,
+        noise_squashing_key,
+        noise_squashing_compression_key,
+        cpk_re_randomization_key,
+        oprf_key,
+        tag,
+    ) = compressed_server_key.into_raw_parts();
+
+    if noise_squashing_key.is_none() {
+        anyhow::bail!("Compressed server key does not have noise squashing");
```
rudy-6-4: Not clear for me. (but I am french :) )
"The compressed server key is missing its noise_squashing_key"

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/host-listener/src/kms_generation/database.rs
antoniupop: It was not changed in the PR, it's a guar dagainst a NULL being set ... not sure it's essential, that would happen if we don't have the sns_pk on update event.

coprocessor/fhevm-engine/fhevm-engine-common/src/db_keys.rs
antoniupop: Fair point, yes. Not related to this PR, but makes sense to clean that up now that I'm here. Will patch up.

coprocessor/fhevm-engine/fhevm-engine-common/src/db_keys.rs
antoniupop: Will do, it is better.

coprocessor/fhevm-engine/host-listener/src/kms_generation/sks_key.rs
antoniupop: I'll reword, sure.

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/host-listener/src/kms_generation/mod.rs
```diff
@@ -277,29 +294,115 @@ async fn download_and_store_crs_activation<
...
+            // tell whether the XOF prefix is genuinely absent or just
+            // unreachable. Log the real error so an operator can
+            // distinguish, and fall through to the legacy prefix; if
+            // the cause was transient the legacy probe will fail too
+            // and the activation will be retried.
+            info!(
+                key_id = key_id.to_string(),
+                xof_error = %err,
+                "CompressedXofKeySet probe failed across all buckets (could be missing, transient, or permissions); trying legacy ServerKey prefix"
+            );
```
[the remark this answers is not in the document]
antoniupop: S3 error handling should be a separate PR, this already surfaces the error, so this is out of scope.

coprocessor/fhevm-engine/host-listener/src/kms_generation/sks_key.rs
```diff
@@ -1,20 +1,128 @@
...
+            sns_pk_compressed: Some(b"key_bytes".to_vec()),
+            sks_key_compressed: Some(b"key_bytes".to_vec()),
+        });
     }
 
-    let server_key: ServerKey = safe_deserialize_sns_key(sns_key)?;
+    let compressed_key_set: CompressedXofKeySet =
+        safe_deserialize_sns_key(key_bytes)?;
+    let (_seed, _compressed_public_key, compressed_server_key) =
+        compressed_key_set.into_raw_parts();
```
kc1212: not sure if it's needed, but if copro ever need the full `CompressedXofKeySet`, then a re-download from S3 would be needed as it's not stored

antoniupop: Not sure it makes sense to add this to DB as a duplicate (for most parts) if we don't have a known need. Storing as a single and then adding helpers to get the needed part would have download overhead, so not reasonable... It's not like the re-download is a big issue so I'd lean towards leaving it as is.

rudy-6-4 (approved):

antoniupop:
@mergifyio queue

PR #2511, opened 2026-05-15
fix(host-contracts): task:assertKmsMigrationSucceeded runnable standalone

obatirou:
- Compile contracts on entry so the task works without a prior deploy/migrate step.
- Add useInternalProxyAddress to load addresses from the addresses/ env file; default reads them from process.env.


isaacdecoded (approved):
Great! 👍

PanGan21 (approved):

tawadaa (approved):
LGTM, thanks!

tawadaa (approved):
LGTM, thanks!

PR #2512, opened 2026-05-15
fix(gateway-contracts): improve operator toolkit granularity

isaacdecoded:
## Summary

  Addresses OpenZeppelin L-04 (zama-ai/fhevm-internal#1366) plus the long-open
  zama-ai/fhevm-internal#161. Adds operator incident-response primitives to
  `GatewayConfig` and propagates them through the gating modifiers.

  ## What the audit asked vs. what we did

  ### 1 — Pause coupling on admin setters
  **Audit:** require `Decryption` and `InputVerification` paused before
  `updateKmsContext`, `updateCoprocessors`, and the threshold setters.

  **Chose:** pause coupling only on `updateCoprocessors` and
  `updateCoprocessorThreshold` (their rewrites are destructive). `updateKmsContext`
  is forward-only with per-context preservation, so in-flight requests aren't
  disrupted by rotation. Threshold setters were rewritten as per-context variants
  (see 2); current-context updates are allowed and the stale-request risk is
  documented as an operational concern rather than enforced via pause.

  ### 2 — Asymmetric administration API
  **Audit:** add per-context threshold setters so older contexts can be adjusted
  after rotation.

  **Chose:** replaced the four in-place setters with `update*ThresholdForContext`
  variants taking an explicit `contextId`. Skipped per-context getters for
  `mpcThreshold` and `kmsGenThreshold` — neither has an on-chain consumer on the
  gateway (KMSGeneration is view-only post-migration, MPC threshold is not consumed).

  ### 3 — Cannot invalidate an old KMS context
  **Audit:** add `destroyKmsContext` and `isValidKmsContext` on the gateway.

  **Chose:** added `destroyKmsContext` (irreversible, refuses the current context).
  All per-context views on `GatewayConfig` revert with `InvalidKmsContext` for destroyed/unknown contexts
  (mirrors `ProtocolConfig`). The verification path inherits the rejection through
  `_checkKmsContextSignerMatchesTxSender` reading those views — closing the
  bypass-Connector + leaked-old-key scenario CR-01 left open.

  ### 4 — Cannot stop traffic from a single host chain
  **Audit:** add reversible enable/disable bits for host chains; explicitly flagged
  `publicDecryptionRequest` as needing CiphertextCommits-side gating since it has
  no top-level chain modifier.

  **Chose:** all three operations — `disableHostChain` / `enableHostChain` for
  incidents (reversible), plus `removeHostChain` for permanent cleanup (requires
  disabled first; clears the registry entry so the same chainId can be
  re-registered later). `onlyRegisteredHostChain` and
  `onlyHandleFromRegisteredHostChain` now also reject disabled chains with a
  distinct `HostChainDisabled` error. `CiphertextCommits.getSnsCiphertextMaterials`
  runs the check per handle in its loop, so `publicDecryptionRequest` inherits the
  rejection.

  ## Versioning

  `REINITIALIZER_VERSION` and `MINOR_VERSION` bumped on the 4 contracts whose
  bytecode changed (`GatewayConfig`, `CiphertextCommits`, `Decryption`,
  `InputVerification`). `Decryption` and `InputVerification` are bumped despite
  unchanged source because the inherited modifier bytecode changed — and
  `tasks/upgradeContracts.ts` requires a fresh reinitializer per upgrade.

  ABI compat exceptions added for the 8 removed signatures (4 in-place setters +
  4 events) on `GatewayConfig`.

Closes https://github.com/zama-ai/fhevm-internal/issues/1366
Closes https://github.com/zama-ai/fhevm-internal/issues/161


gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -335,45 +368,53 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
+        emit UpdateKmsGenThresholdForContext(contextId, newKmsGenThreshold);
     }
 
     /**
      * @notice See {IGatewayConfig-updateCoprocessorThreshold}.
      */
-    function updateCoprocessorThreshold(uint256 newCoprocessorThreshold) external virtual onlyOwner {
+    function updateCoprocessorThreshold(
+        uint256 newCoprocessorThreshold
+    ) external virtual onlyOwner whenInputVerificationPaused {
```
obatirou, line 414: according to the issue Decryption should also be paused
So a whenDecryptionPaused might be necessary. Can be applied to several functions

gateway-contracts/contracts/shared/GatewayConfigChecks.sol
```diff
@@ -57,6 +57,14 @@ abstract contract GatewayConfigChecks {
      */
     error HostChainNotRegistered(uint256 chainId);
```
obatirou: duplicated from gateway-contracts/contracts/interfaces/IGatewayConfig.sol

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -285,13 +299,32 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
+     */
+    function destroyKmsContext(uint256 kmsContextId) external virtual onlyOwner {
+        GatewayConfigStorage storage $ = _getGatewayConfigStorage();
+
+        if (kmsContextId == $.currentKmsContextId) {
+            revert CurrentKmsContextCannotBeDestroyed(kmsContextId);
+        }
+        // Context must exist (signer set populated by updateKmsContext or initializer)
+        // and not be already destroyed.
+        if ($.kmsSignerAddressesForContext[kmsContextId].length == 0 || $.destroyedKmsContexts[kmsContextId]) {
```
obatirou: use _isValidKmsContext?

obatirou: or even _requireValidContext

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -285,13 +299,32 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
+    /**
+     * @notice See {IGatewayConfig-destroyKmsContext}.
+     */
+    function destroyKmsContext(uint256 kmsContextId) external virtual onlyOwner {
+        GatewayConfigStorage storage $ = _getGatewayConfigStorage();
+
+        if (kmsContextId == $.currentKmsContextId) {
+            revert CurrentKmsContextCannotBeDestroyed(kmsContextId);
+        }
+        // Context must exist (signer set populated by updateKmsContext or initializer)
```
obatirou: updateKmsContext is not updated with modifier?

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -285,13 +299,32 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         );
     }
 
+    /**
+     * @notice See {IGatewayConfig-destroyKmsContext}.
+     */
+    function destroyKmsContext(uint256 kmsContextId) external virtual onlyOwner {
```
obatirou, line 305: on the issue it is ask for isValidKmsContext view

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -624,45 +736,52 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
     }
 
     /**
      * @notice See {IGatewayConfig-getKmsSignersForContext}.
+     * @dev Reverts with `InvalidKmsContext` if the context is unknown or destroyed.
      */
     function getKmsSignersForContext(uint256 contextId) external view virtual returns (address[] memory) {
-        GatewayConfigStorage storage $ = _getGatewayConfigStorage();
-        return $.kmsSignerAddressesForContext[contextId];
+        _requireValidContext(contextId);
```
obatirou, line 780: I am still wondering about the reverts. I know we discuss it already...but still remember that the lib should not revert or something like that no?
Still not clear to me

obatirou: If we keep this require, are all views have it?

obatirou (changes requested):
a few things missing, let's run claude once rebased + fixes done

obatirou:
Also regarding Pausing:
>  so in-flight requests aren't disrupted by rotation.

Without request are not pinned to a coproc context like KMS context. This should be made aware to auditors that it is a know issue and feature is already planned

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: I didn't add the `whenDecryptionPaused` because I'm not sure we need to pause the Decryption contract for coprocessor rotation/threshold updates. I mean, decryption responses have nothing to do with coprocessor signers and their configured threshold. What do you think?

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Initially, I think this is not a requirement to pause the Decryption contract for calling `updateKmsContext`, since the decryptions are already context-aware. Meaning that a KMS context rotation shouldn't break the decryption flows - like Coprocessors that are not context-aware, and rotating them means getting bad behavior during input verifications. What do you think here?

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Yes, this discussion is a thing indeed. But basically I added the revert for two reasons:
1. the verification flow propagates the rejection automatically through _checkKmsContextSignerMatchesTxSender reading these views, so we don't need an extra check in GatewayConfigChecks
2. it mirrors the [ProtocolConfig implementation](https://github.com/zama-ai/fhevm/blob/main/host-contracts/contracts/ProtocolConfig.sol#L182-L206)

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Changed to a [view function here](https://github.com/zama-ai/fhevm/pull/2512/changes/7c84bafb869adce67c6594773681478469079208#diff-4f4932d9ac552e4be3ed692009cbfc9d84ae05bf12e6bb137a4a5e2e7e2c32baR693).

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Good catch, [reused _requireValidContext](https://github.com/zama-ai/fhevm/pull/2512/changes/7c84bafb869adce67c6594773681478469079208#diff-4139d8055b558914fe9fc072c5a6a27496114f8e6bc3e081b09d321f2528dd59R311-R312).

gateway-contracts/contracts/shared/GatewayConfigChecks.sol
isaacdecoded: Yes, good catch. Reused the [error from the GatewayConfig](https://github.com/zama-ai/fhevm/pull/2512/changes/7c84bafb869adce67c6594773681478469079208#diff-74bf15292a1eb4a898dbcda53af25516fa0be5bd2a52b62801f8e2c7992bc904L54-L59).

gateway-contracts/contracts/GatewayConfig.sol
obatirou: indeed good point 👍

gateway-contracts/contracts/GatewayConfig.sol
obatirou: you are totally right
if this is correct for all setter then all good

obatirou (approved):

obatirou:
@claude /pr-review

gateway-contracts/contracts/interfaces/IGatewayConfig.sol
```diff
@@ -549,6 +681,17 @@ interface IGatewayConfig {
...
+    /**
+     * @notice Returns whether a KMS context ID has been created and not destroyed.
+     * @dev Non-reverting predicate intended for off-chain consumers. The per-context state
+     *      views (e.g. `isKmsSignerForContext`, `getKmsNodeForContext`) revert with
+     *      `InvalidKmsContext` for unknown/destroyed contexts; use this function to test
+     *      validity without exception handling.
+     * @param contextId The KMS context ID to check.
+     * @return True if the context exists and has not been destroyed.
+     */
+    function isValidKmsContext(uint256 contextId) external view returns (bool);
```
zmalatrax, line 693: This function is declared in the interface and implemented*, but the PR body states that it won't be implemented. It is used by many getters/setters, is the PR body outdated, or should it be refactored?

`destroyKmsContext` uses `isValidKmsContext`

> ## 3 — Cannot invalidate an old KMS context
> Did not add a public isValidKmsContext

isaacdecoded: Good catch! Yes, the PR body was outdated. Just updated it already.

zmalatrax (approved):

PR #2513, opened 2026-05-15
fix(host-contracts): add protocolConfigSource switch to deployAllHostContracts

obatirou:


obatirou:
@claude /pr-review

host-contracts/tasks/taskDeploy.ts
```diff
@@ -115,7 +136,11 @@ task('task:deployAllHostContracts')
 
     await hre.run('task:deployACL');
     await hre.run('task:deployFHEVMExecutor');
-    await hre.run('task:deployProtocolConfig');
+    if (protocolConfigSource === 'migration') {
+      await hre.run('task:deployProtocolConfigFromMigration');
+    } else {
+      await hre.run('task:deployProtocolConfig');
```
tawadaa, line 142: Does it make sense to have `deployProtocolConfig` task? as any new host chain should have same state as ethereum

obatirou: that's a good point. 
for next release, the migration would have happen and the source of truth will be ProtocolConfig from Ethereum. So a new task/some modification will be necessary to take the state of ethereum instead.
Maybe deployProtocolConfig is not mandatory indeed

isaacdecoded: I think it's useful for deployments from scratch, e.g., the test-suite which uses these deploy tasks during the gateway/host contracts deployment on Anvil. Also, I don't think it harms keeping the task for deploying the ProtocolConfig contract, as all the other contracts have.

obatirou: I am wondering if we should keep those deployment from scratch as they would not reflect the current state / what is actually used.
deployProtocolConfig will never be used in production.
* for 0.13 we have deployProtocolConfigFromMigration (from gateway state)
* post 0.13, for new chains, we should have deployProtocolConfigFromEthereumState.

There will always be a need for a migration from a state it seems 
So `initializeFromEmptyProxy`  might not be relevant after all

isaacdecoded: Good point 🤔 but still I think it doesn't harm keeping it since it facilitates the test-suite setup, or external developers usage - we can deploy the ProtocolConfig manually using `forge create` of course, but it seems more cumbersome, no?

Anyway, if we decide to remove it, we must be sure not to break the test-suite at least.

obatirou: after circling back with @isaacdecoded: 
I agree. We need to be able to deploy the protocol from scratch for devs, or other external productions.
So it seems this should stay as is for now

PanGan21 (approved):

haroldsphinx (approved):

PR #2514, opened 2026-05-15
feat(coprocessor): implement multi output computations support

PanGan21:
### Summary
Implements multi-output FHE op infrastructure on the coprocessor side: a single tfhe-rs op can now produce N output ciphertexts in one dispatch, stored as N rows in the existing computations tables keyed by a shared `group_id` and `output_index` 0..N-1.

- migration adds three columns: `group_id`, `output_count` and `output_index`, where group id is the handle of the first output
- host-listener writes N grouped rows atomically
- tfhe-worker reassembles N rows sharing a group_id into one logical DFGOp (singletons key on their own handle)
- scheduler carries `Vec<DFGOutput>` and routes results by output handle
- a failed multi-output op is reported to the worker once, as one error for the whole group; the worker classifies it once and stamps every output row with it
- dependence chains: boundary consumption is keyed by group id, so consumers of different outputs of one op fork instead of serializing on the producer chain
- host listener and scheduler track permissions per output
- scheduler validates all output counts and types atomically
- host listener enforces the shared output-arity limit

Related: https://github.com/zama-ai/fhevm-internal/issues/1395


coprocessor/fhevm-engine/db-migration/migrations/20260515120000_relax_op_level_columns_for_multi_output.sql
```diff
@@ -0,0 +1,11 @@
+-- Multi-output ops store op-level fields only on the primary row (output_index = 0).
+-- Sibling rows leave these columns NULL.
+--
+-- Rolling deploy order: (1) run this migration, (2) deploy tfhe-worker, (3) deploy
```
PanGan21: Comment will be removed eventually. It is there for a reminder on the deployment order

PanGan21:
@claude /pr-review

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -502,29 +611,51 @@ fn run_computation(
...
+                    (graph_node_index, Err(e.into()))
+                }
+            };
+        }
+    }
+
+    let op = FheOperation::try_from(operation);
+    match op {
+        Ok(FheOperation::FheGetCiphertext) => {
             let ct_type = inputs[0].type_num();
```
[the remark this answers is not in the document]
PanGan21: fixed

PanGan21:
Successful e2e test run for the poc multi output operators: https://github.com/zama-ai/fhevm/actions/runs/26027734404

coprocessor/fhevm-engine/db-migration/migrations/20260515120000_relax_op_level_columns_for_multi_output.sql
```diff
@@ -0,0 +1,11 @@
+-- Multi-output ops store op-level fields only on the primary row (output_index = 0).
+-- Sibling rows leave these columns NULL.
```
rudy-6-4: could it be easier to have copies in each rows ? and only output_index to differ ?

host-contracts/contracts/FHEEvents.sol
```diff
@@ -50,4 +50,10 @@ contract FHEEvents {
         bytes1 scalarByte,
         bytes32 result
     );
+    /// @dev Sample multi-output op (proof-of-concept). Emits two result handles:
+    ///      `resultValue` (index 0, same type as input) and `resultFound` (index 1, ebool).
+    event FheSampleMultiOutput(address indexed caller, bytes32 ct, bytes32 resultValue, bytes32 resultFound);
+    /// @dev High-arity sample multi-output op. Emits 100 result handles, ordered by
+    ///      tfhe-rs result tuple index. Every entry shares the input type.
+    event FheSampleMultiOutput100(address indexed caller, bytes32 ct, bytes32[] results);
```
rudy-6-4: I think a variable input arity would be more interesting for FheSampleMultiOutput100. As it's already supported. Ie output-arity is fixed by the event itself.
It could be used to emulate array returned array if we don't have them soon.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -366,11 +368,59 @@ impl Database {
...
+            .first()
+            .ok_or_else(|| {
+                SqlxError::Protocol(
+                    "insert_multi_output_computation: results is empty".into(),
+                )
+            })?
+            .to_vec();
+        let mut any_inserted = false;
+        for (idx, result) in results.iter().enumerate() {
+            let output_index = i16::try_from(idx).map_err(|_| {
```
rudy-6-4: This should be not break the insertion of all other events ?

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -336,16 +337,23 @@ fn execute_partition(
...
+                    for h in node.result_handles.iter() {
+                        res.insert(h.clone(), Err(SchedulerError::CyclicDependence.into()));
+                    }
                 }
             }
             continue 'tx;
         };
         let edges = dfg.graph.map(|_, _| (), |_, edge| *edge);
         for nidx in ts.iter() {
+            let producer_handles: Vec<Handle> = {
```
rudy-6-4: why not done after line 360 ? (doing the same test)

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -366,11 +368,59 @@ impl Database {
...
+        fhe_operation: FheOperation,
+        scalar_byte: &FixedBytes<1>,
+        log: &LogTfhe,
+    ) -> Result<bool, SqlxError> {
+        let dependencies =
+            dependencies.iter().map(|d| d.to_vec()).collect::<Vec<_>>();
+        let group_id = results
+            .first()
+            .ok_or_else(|| {
+                SqlxError::Protocol(
```
rudy-6-4: I know it's a fake op, but just to see what we should do on a real one. If the event have no result we can "safely" discard it since no computation depends on it.

coprocessor/fhevm-engine/db-migration/migrations/20260515120000_relax_op_level_columns_for_multi_output.sql
PanGan21: The problem i see with this approach is that for large outputs (not KVStore but shuffle) a lot of fields are duplicated increasing the size of the db. For example dependencies column can become quite large and it will be replicated across all output rows. Therefore I chose to make those nullable. 
Do you still think we should use another approach?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
PanGan21: Fixed in: https://github.com/zama-ai/fhevm/commit/e6252d015e5ea4442a0362b37a444e3104d5ff06

Added a warning and a comment that this is a defensive check

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
PanGan21: Good catch, reusing it now to avoid dublication: https://github.com/zama-ai/fhevm/pull/2514/commits/96c1b7e64ea9323667d398e3b7fe0acc5b31291f

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
PanGan21: Discarding with a warning now: https://github.com/zama-ai/fhevm/pull/2514/changes/57b976829f6dbc76203315bcde525632cae3cb82

host-contracts/contracts/FHEEvents.sol
PanGan21: Yes good idea. Do you think I should look into variable output-arity as well for potentially supporting `earray` in the future? Although even with `earray` I am unsure if we would need the variable multi output.

rudy-6-4: Yes, that's what I had in mind. The output arity depend on the op + inputs (input arity included).

coprocessor/fhevm-engine/db-migration/migrations/20260515120000_relax_op_level_columns_for_multi_output.sql
rudy-6-4: No you are right for N inputs to N outputs, it's quadratic in size in the db. But that could be restricted to inputs field only (can be empty array btw instead of null) to avoid extra options in rust for other fields.

host-contracts/contracts/FHEEvents.sol
PanGan21: Just a note on this. The size of the output needs to be known on Solidity side, so the user knows how to destructure the bytes32[] into separate decryptable handles

PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2514/commits/221b8397a30ade86ba0a0539a258be7237361396

I added FheSampleVariableInputOutput which does f(N) = 2N in size so we can emulate variable input and output length.

coprocessor/fhevm-engine/db-migration/migrations/20260515120000_relax_op_level_columns_for_multi_output.sql
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2514/commits/130b06b23acdedbddc614798ded1d06bb1e916dc

Removed the migration and now only dependencies is an empty array.

rudy-6-4 (dismissed):

PanGan21:
@claude /pr-review

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -250,12 +251,9 @@ pub async fn ingest_block_logs(
...
     for tfhe_log in tfhe_event_log.iter_mut() {
-        tfhe_log.is_allowed =
-            if let Some(result_handle) = tfhe_result_handle(&tfhe_log.event) {
-                is_allowed.contains(&result_handle.to_vec())
-            } else {
-                false
-            };
+        tfhe_log.is_allowed = tfhe_result_handles(&tfhe_log.event)
+            .iter()
+            .any(|h| is_allowed.contains(&h.to_vec()));
```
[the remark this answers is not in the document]
PanGan21: Added an explanatory comment

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -502,33 +604,48 @@ fn run_computation(
...
-    let op = FheOperation::try_from(operation);
-    match op {
-        Ok(FheOperation::FheGetCiphertext) => {
-            // Compression span (no FHE here)
-            let _guard = tracing::info_span!(
-                "compress_ciphertext",
+
+    // Multi-output ops dispatch through a separate impl that returns Vec.
+    if let Ok(sup_op) = SupportedFheOperations::try_from(operation as i16) {
+        if sup_op.is_multi_output() {
```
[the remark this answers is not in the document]
PanGan21: This will be used in the future from multi output ops

coprocessor/fhevm-engine/db-migration/migrations/20260805120000_add_multi_output_columns.sql
```diff
@@ -0,0 +1,23 @@
...
+--
+-- Both tables are written by the wave-1 dual-write in the host-listener, so
+-- both need the columns: `computations` is the current read path and
+-- `computations_branch` is the wave-2 read path.
+
+ALTER TABLE computations
+    ADD COLUMN IF NOT EXISTS group_id BYTEA NULL,
+    ADD COLUMN IF NOT EXISTS output_index SMALLINT NOT NULL DEFAULT 0;
+
+CREATE INDEX IF NOT EXISTS idx_computations_group_id
```
PanGan21: We might want to create this index from the script? Not sure what is best

coprocessor/fhevm-engine/db-migration/migrations/20260805120000_add_multi_output_columns.sql
```diff
@@ -0,0 +1,23 @@
...
+-- `output_index` 0..N-1. Singletons keep `group_id IS NULL` and
+-- `output_index = 0`, so the defaults leave existing writers untouched.
+--
+-- Both tables are written by the wave-1 dual-write in the host-listener, so
+-- both need the columns: `computations` is the current read path and
+-- `computations_branch` is the wave-2 read path.
+
+ALTER TABLE computations
+    ADD COLUMN IF NOT EXISTS group_id BYTEA NULL,
+    ADD COLUMN IF NOT EXISTS output_index SMALLINT NOT NULL DEFAULT 0;
```
PanGan21: Would we consider adding `output_count` as well to enforce correctness or this is redundant information?

coprocessor/fhevm-engine/db-migration/migrations/20260805120000_add_multi_output_columns.sql
PanGan21: The indexes have been removed since they were not really used

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -440,11 +444,55 @@ impl Database {
...
+        let dependencies =
+            dependencies.iter().map(|d| d.to_vec()).collect::<Vec<_>>();
+        if !is_valid_multi_output_arity(results.len()) {
+            warn!(target: "host_listener",
+                arity = results.len(),
+                max = MAX_MULTI_OUTPUT_ARITY,
+                "unsupported multi-output arity; skipping event");
+            return Ok(false);
+        }
+        let group_id = results[0].to_vec();
```
PanGan21, line 510: A loud note that group id is the handle of the first result.

PanGan21: Another option is to hash all output handles, however this can hide risks and has small performance penalty.

coprocessor/fhevm-engine/db-migration/migrations/20260805120000_add_multi_output_columns.sql
PanGan21: Added `output_count` in: https://github.com/zama-ai/fhevm/pull/2514/changes/d39e5006f6e187733cb18ef35d6149bf5528f3df

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
rudy-6-4: Is it possible that the same handle 0 could be generated by 2 multi output ops in the same block without having the same handle_1 etc ?

I think not, if the second and subsequent handle are all generated from the same basis as the first one.
Ie in contract
handle_0 = hash(same suff, number of outputs, index 0)
handle_1 = hash(same suff, number of outputs, index 1)
...
This make handle_0 the "same" as hashing everything ?
If that way of generating is enforced or documented, we don't need to rehash all handles.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -993,6 +999,14 @@ pub const COMPUTED_HANDLE_INDEX_MARKER: u8 = 0xff;
 /// constant in `host-contracts/contracts/shared/Constants.sol`.
 pub const HANDLE_VERSION: u8 = 0;
 
+/// Technical bound only; each operation sets its own smaller cap.
+pub const MAX_MULTI_OUTPUT_ARITY: usize = u8::MAX as usize;
```
rudy-6-4, line 1004: "Technical bound" intent is not clear. Is it just for a defensive check against bug or forged event to prevent very long operations ?

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -1168,7 +1163,9 @@ pub async fn synthesize_finalized_fallback_grants(
     let chains =
         dependence_chains(&mut logs, &db.dependence_chain, false, false).await;
     for log in &logs {
-        let dst_handle = tfhe_result_handle(&log.event)
+        let dst_handle = tfhe_result_handles(&log.event)
```
rudy-6-4: could be dst_handles here ? why checking ?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -588,6 +657,9 @@ impl Database {
         let insert_computation_bytes = |tx, result, dependencies_handles, dependencies_bytes, scalar_byte| {
             self.insert_computation_bytes(tx, result, dependencies_handles, dependencies_bytes, fhe_operation, scalar_byte, log)
         };
+        // TODO: when adding a multi-output op, define a closure mirroring the two above,
```
rudy-6-4, line 659: why not adding the group index and size to insert_computation ? and providing the None 0 1 here instead of inside insert_computation ? (I assume insert_computation_bytes will never be used for multi output.

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -383,96 +383,119 @@ fn execute_partition(
...
                         continue;
                     };
-                    let handle = node.result_handle.clone();
-                    let is_allowed = node.is_allowed;
+                    let outputs = node.outputs.clone();
+                    let producer_handles: Vec<Handle> =
+                        outputs.iter().map(|o| o.handle.clone()).collect();
                     let opcode = node.opcode;
                     match op_result {
-                        Ok(working) => {
```
rudy-6-4, line 404: Maybe it could be good to unest a bit the code with error handling first, like an early exit (for the 2 first evel of error).
let Ok(working) = op_result else {
    errror handling
};
normal path unnested.

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -383,96 +383,119 @@ fn execute_partition(
...
+                                    };
+                                    let Some(out_idx) =
+                                        producer_handles.iter().position(|h| h == &dep_handle)
+                                    else {
+                                        error!(target: "scheduler",
+                                        { handle = ?hex::encode(&dep_handle) },
+                                        "Consumer dependence handle not found in producer outputs - graph inconsistency");
+                                        continue;
+                                    };
+                                    if let Some(Some(value)) = forwarded.get(out_idx) {
```
rudy-6-4: Why is there a double Option level here ? No None should give rise to an error here ?

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -533,18 +557,20 @@ fn try_execute_node(
...
             return Err(SchedulerError::ReRandomisationError.into());
         }
         let elapsed = started_at.elapsed();
         RERAND_LATENCY_BATCH_HISTOGRAM.observe(elapsed.as_secs_f64());
     }
     let opcode = node.opcode;
-    let output_type = get_ct_type(&node.result_handle).map_err(|e| {
-        error!(target: "scheduler", { handle = ?hex::encode(&node.result_handle), error = ?e },
-               "Invalid result handle: cannot read type byte");
+    let output_type = get_ct_type(&node.outputs[0].handle).map_err(|e| {
```
rudy-6-4: Maybe you can pass all types ? since on kv store get they can be different ?

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
PanGan21: Yes it is purely defensive as an upper bound for all multi output ops. I updated the docs to clarify it further: https://github.com/zama-ai/fhevm/pull/2514/commits/4723f9dfc8a2d673a58d2ab685c0809cae5c9783

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
PanGan21: Correct, this is the rationale for group id. I believe we can enforce it because the executor is the sole creator of computation handles.

Worth noting one consequence: two identical multi-output ops in the same transaction derive identical handles and collapse into a single group via `ON CONFLICT (output_handle, transaction_id) DO NOTHING`. That's the same logic we already rely on for single output ops, and I think the group stays consistent

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
PanGan21: It can't be `dst_handles` (plural). These are synthetic TrivialEncrypt events, single output by construction. And the check was pointless indeed,
Fixed it and now the handle is recorded next to the log and zipped in the loop, so the expect is gone: https://github.com/zama-ai/fhevm/pull/2514/changes/a1fe0ff2c941bcd0ea095b054ce06d0e2bb62349

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
PanGan21: Done in [ce281d9dc](https://github.com/zama-ai/fhevm/pull/2514/changes/ce281d9dce1c3078095363d696f3f7d7fea2ee30) `insert_computation` now takes the group values and callers pass None, 0, 1 for single-output ops.

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2514/commits/ce281d9dce1c3078095363d696f3f7d7fea2ee30
Both levels are early exits now. One note: let-else won't exactly work, since it drops the `Err` that `fan_out_error` needs, so it's match … continue for the dispatch error and if let Err(…) for the `validate_results` one

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
PanGan21: Removed in: https://github.com/zama-ai/fhevm/pull/2514/commits/fd15b1573199cb6bdfc65af982332d5cbbffb1f3 and explained in the comment.
The outer level was unreachable. The remaining None means the compression failure was already marked, so forwarding nothing avoids reporting it twice.

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
PanGan21: Makes sense. I was thinking to defer this for when it is needed but doesn't harm if it is implemented now as well.
Done in: https://github.com/zama-ai/fhevm/pull/2514/commits/5bc01268c42fdb1759a190849b62410dec58cb73

rudy-6-4 (dismissed):

rudy-6-4 (approved):

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -4704,6 +4708,15 @@ fn stamp_text(cerr: &(dyn std::error::Error + Send + Sync), retryable: bool) ->
     }
 }
 
+/// Dependents not yet propagated to in this pass; marks them as propagated.
+fn unpropagated(propagated: &mut HashSet<Vec<u8>>, dependents: &[Vec<u8>]) -> Vec<Vec<u8>> {
```
rudy-6-4: name is not clear maybe: mark_as_propagated ?

rudy-6-4 (dismissed):

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
PanGan21: Yes: https://github.com/zama-ai/fhevm/pull/2514/commits/cf0f5d9849a8bdba18d1e6468abf3e13ab328b25

rudy-6-4 (approved):

PanGan21:
@Mergifyio queue

PR #2488, opened 2026-05-13
fix(host-contracts): pin KMS context at response time in KMSGeneration

obatirou:
Response handlers read the live KMS context, so a rotation between request and response could reject request-time signers and evaluate consensus under the new committee's threshold. Each handler now decodes the `contextId` from `requestExtraData` and gates membership, signer match, and consensus against it. Adds per-context getters for `kmsGen` / `mpc` thresholds on `IProtocolConfig`.

Fixes [`zama-ai/fhevm-internal#1361`](https://github.com/zama-ai/fhevm-internal/issues/1361).

PR #2516, opened 2026-05-15
fix(host-contracts): pin KMS context at response time in KMSGeneration

obatirou:
## Summary
- Forward-port of #2488 from `release/0.13.x` to `main` via cherry-pick of 8b81b2605c09187ef5ac5c1412922a0e6127654a.
- Response handlers in KMSGeneration now decode `contextId` from `requestExtraData` and gate membership, signer match, and consensus against it, preventing a context rotation between request and response from invalidating request-time signers or applying the new committee's threshold.
- Adds per-context getters for kmsGen/mpc thresholds on `IProtocolConfig`.

Fixes zama-ai/fhevm-internal#1361 (M-01).


PanGan21 (approved):

obatirou:
@mergify queue

PR #2518, opened 2026-05-15
chore(gateway-contracts): refresh rust bindings

obatirou:
## Summary
- Regenerated `gateway-contracts/rust_bindings/` via `make conformance` (`gateway-contracts/Makefile:81`).
- Drift is bytecode/metadata-only (solc version / metadata-hash bytes embedded in `DEPLOYED_BYTECODE` and `CREATION_BYTECODE` byte-string literals).
- Function selectors and ABI are unchanged — `gateway-contracts/selectors.txt` is untouched.

Affected files:
- `gateway-contracts/rust_bindings/src/decryption.rs`
- `gateway-contracts/rust_bindings/src/gateway_config.rs`
- `gateway-contracts/rust_bindings/src/input_verification.rs`
- `gateway-contracts/rust_bindings/src/protocol_payment.rs`


obatirou:
Main is correct. My local gateway-contracts/addresses/ was stale (predated the RFC013-related slot rotation). Deleting addresses/ and re-running make conformance clears it.

PR #2521, opened 2026-05-15
feat(listener): catchup feature

nboisde:
### Core:
- Not relying on database.
- Doesn't handle reorgs (on live processing only), 
- Not limited up to finality
- Splits long ranges of blocks into sub ranges at the core level.
- Catchup is for a given consumer.
- Publish to a new queue "{consumer-id}.catchup-event"
- Publish a block payload.
- Uses cursor data structure for parallelism and performance.
- Dashboard panel for the catchup. 

### Consumer library: 
- Optionnal configurable values. 
- New broker and consumer for the catchup queue.

### Example:
- Usage example on the zama token (for both live and catchup feature).


listener/crates/consumer/src/client.rs
```diff
@@ -76,6 +135,30 @@ impl ListenerConsumer {
...
+    ///
+    /// Publishes a [`CatchupPayload`] to the chain-namespaced
+    /// `routing::CATCHUP` control plane. The listener fans the range out to
+    /// its `range-catchup` workers; the resulting events are delivered on
+    /// `{consumer_id}.catchup-event` and consumed via
+    /// [`consume_catchup`](Self::consume_catchup).
+    pub async fn request_catchup(
+        &self,
+        block_start: u64,
+        block_end: u64,
```
rudy-6-4, line 148: Not essential and maybe for a next PR but can "block_end" be optional ? From the consumer side, we don't always have a valid block number easily (no rpc url). Use case of replaying everything up to present time from a block (eg spin up a new copro) 
I know the live flow already handle the gaps so there should be no miss wrt to present block but just for end user it could be simpler to say "start and consumer from that point in past" .

listener/crates/consumer/src/client.rs
```diff
@@ -101,23 +184,77 @@ impl ListenerConsumer {
...
+        let opts = &self.live_options;
+
+        let mut builder = self
             .broker
             .consumer(&topic)
             .group(topic.to_string())
-            .prefetch(1)
-            .max_retries(3)
-            .circuit_breaker(3, Duration::from_secs(30))
+            .prefetch(opts.prefetch())
```
rudy-6-4, line 201: thx

listener/crates/consumer/src/client.rs
rudy-6-4: ok I have seen your example to sync live and catchup, great.

listener/config/listener-default.yaml
```diff
@@ -51,6 +51,11 @@ blockchain:
     compute_block: false
     compute_block_allow_skipping: true
     max_exponential_backoff_ms: 20000
+  catchup:
+    prefetch: 5
+    claim_min_idle_secs: 3600
+    catchup_max_range: 1000
```
rudy-6-4: It's the number of concurrent request during the catchup ? (since each block is 1 request).
If yes could be max_concurrent_get_block.

listener/crates/listener_core/src/core/evm_listener.rs
```diff
@@ -930,6 +932,194 @@ impl EvmListener {
...
+        );
+
+        // Step 3 — shared producer/consumer state.
+        let range_start_time = Instant::now();
+        let buffer = AsyncSlotBuffer::<FetchedBlock>::new(range_length);
+        let cancel_token = CancellationToken::new();
+
+        // Step 4 — spawn the same producer used by the live cursor: pure fetch,
+        // no DB or hash work — perfectly reusable.
+        let fetcher_handle = tokio::spawn(fetch_blocks_in_parallel(
```
rudy-6-4, line 1050: ok so each block is 1 request, is it ok if we need to replay lots of block (eg several days of blocks) ?

listener/crates/listener_core/src/core/evm_listener.rs
rudy-6-4, line 1421: (Not this PR but) I think all errors should be displayed. The user need to know if1 block wasnot replayed of if many fails.

rudy-6-4, line 1410: (Not this PR but) I think all errors should be displayed (maybe they are by the tasks them selves). The user need to know if 1 block was not replayed of if many fails. Could be 1 or 1000 blocks, which is not the same.

listener/crates/listener_core/src/core/evm_listener.rs
```diff
@@ -1066,6 +1256,79 @@ async fn cursor_processing(
...
+    range_length: usize,
+    consumer_id: String,
+) -> Result<(), EvmListenerError> {
+    let chain_id_u64 = listener.repositories.chain_id() as u64;
+
+    for i in 0..range_length {
+        let block_number = range_start + i as u64;
+
+        let fetched_block = tokio::select! {
+            biased;
```
rudy-6-4, line 1291: (Not this PR but) I think all errors should be displayed. The user need to know if1 block wasnot replayed of if many fails.

listener/config/listener-default.yaml
rudy-6-4: It's also the maximum number of blocks in memory at once.
Could be smaller just in case. We use 50 or 100 in prod on a similar options mostly due to node limitation that we can encounter even with 50 (can be too big if blocks are full).

listener/crates/listener_core/src/core/publisher.rs
```diff
@@ -415,109 +413,196 @@ pub async fn publish_block_events(
...
+        return Ok(());
+    }
+
+    // 2. Build a narrowly-scoped index + payload (same helpers as the live path).
+    let filter_index = FilterIndex::from_filters(filters);
+    let payloads =
+        filter_index.build_block_payloads(fetched_block, chain_id, BlockFlow::Catchup)?;
+
+    // 3. Publish the target consumer's payload (if any) to catchup-event.
+    //    With single-consumer filters indexed, payloads contains 0 or 1 entry.
```
rudy-6-4: I would early return if payloads is empty. And the second check cid == consumer_id cannot be false, no ?
And use the first cell with an error log if more than 1 cell.

listener/crates/listener_core/src/core/evm_listener.rs
```diff
@@ -1066,6 +1256,79 @@ async fn cursor_processing(
...
+            &consumer_id,
+            &listener.broker,
+            &listener.event_publisher,
+            &listener.publish_config,
+        )
+        .await
+        .map_err(|source| {
+            cancel_token.cancel();
+            EvmListenerError::PayloadBuildError { source }
+        })?;
```
rudy-6-4, line 1320: in case of error, the retry is done for the full range ?

listener/crates/listener_core/config.yaml
```diff
@@ -24,10 +24,10 @@ database:
...
     # max_lifetime_secs: 1800
 broker:
-  broker_type: redis
-  broker_url: redis://localhost:6379
-  # broker_type: amqp
-  # broker_url: amqp://user:pass@localhost:5672/%2f
+  # broker_type: redis
+  # broker_url: redis://localhost:6379
+  broker_type: amqp
+  broker_url: amqp://user:pass@localhost:5672/%2f
```
rudy-6-4: ? intentional ?
is it better to test locally with amqp ?

listener/monitoring/grafana/dashboards/listener.json
```diff
@@ -825,6 +825,241 @@
       ],
       "title": "Block Hash Failures Rate",
       "type": "timeseries"
+    },
```
rudy-6-4, line 828: is it auto published to grafana after ?

rudy-6-4 (dismissed):

listener/config/listener-default.yaml
nboisde: I tested it on eth mainnet, but indeed, I can reduce the default range and rename it !

listener/crates/listener_core/src/core/evm_listener.rs
nboisde: Everything is executed in parallel calls, similar to what was achieved in the cursor algorithm, hence the range of 1000 (tested on a public node).
So we are basically fetching 1k block in parallel here, not blocks 1 by 1.
On eth-mainnet, results are pretty descent, we are getting 1 day of events in 3-4 minutes. For fastest chain, it will be obviously a little bit more, but those params could be tweaked.

listener/crates/listener_core/src/core/evm_listener.rs
nboisde: Yes true, I will add both logs for this error or the one above, good catch !

listener/crates/listener_core/src/core/evm_listener.rs
nboisde: Same than above.

listener/crates/listener_core/config.yaml
nboisde: Indeed just forgot to change this. It was easier to test by hand publishing on rabbitmq UI, but both brokers was tested.

listener/crates/listener_core/src/core/evm_listener.rs
nboisde: Only the sub range, unless the error comes from the above dispatcher.
But indeed, here we will have a retry of the message.

listener/crates/listener_core/src/core/publisher.rs
nboisde: What do you mean ? I don't get it, you mean you don't want the block to be published as we do for the new-event flow (e.g getting all the blocks when catching up) ?

listener/monitoring/grafana/dashboards/listener.json
nboisde: Not sure if we have the pipeline to auto publish this to grafana, but I followed what ghislain did regarding erpc dashboards here.
To start with we can also import them manually, at least we have a source of truth here if needed.

rudy-6-4 (approved):

rudy-6-4:
@mergifyio queue

PR #2530, opened 2026-05-18
docs(coprocessor): document per-block HCU cap and HCUBlockLimitExceeded

antoniupop:


PanGan21 (approved):

leventdem (approved):
Good explanation on how to avoid or solve the situation when the block level limit is reached!

PR #2544, opened 2026-05-18
feat(relayer): unified EIP-712 user-decryption

manoranjith:
## Summary

Relayer side of the unified user decryption flow (paired with
gateway-side #2329). Adds `POST /v3/user-decrypt` accepting the
unified EIP-712 payload, dispatching to
`userDecryptionRequest(HandleEntry[], …)`. v2 endpoints stay
during the deprecation window.

Refs zama-ai/fhevm-internal#1293.

## Wire format vs issue

- camelCase JSON keys (matches v2). EIP-712 typehash is decoupled
  from the HTTP transport.
- Typed-attestation envelope `{ attestationType, attestedPayload,
  signature }` (zama-ai/fhevm-internal#1293 comment 4278777024)
  rather than flat fields — leaves room for non-EIP-712
  attestations later without a v4 bump.

## Notes

- `UserDecryptRequest` is one tagged enum so the downstream
  pipeline has one path. Legacy variants are deprecated and
  collapse out when removed.
- One readiness processor for all three variants. Host ACL
  dispatches per variant; the gateway readiness call is the same
  `isUserDecryptionReady_1` for all.
- Unified host ACL is now enforced per-`HandleEntry`: direct
  (`isAllowed(handle, user)`) when `owner == user`, otherwise
  `isHandleDelegatedForUserDecryption(owner, user, contract,
  handle)`. Mis-grants fail with 4xx instead of being submitted
  and rejected on-chain.
- SQL: drops the obsolete `legacy` enum value; `user_decrypt` /
  `delegated_user_decrypt` are marked deprecated, `unified` is
  current.

## Tests

`make check` + `make test-run` green. New
`tests/user_decrypt_v3_test.rs` covers the issue checklist plus
an end-to-end happy path; v2 suites unchanged.

## Pending

- [ ] Pin `fhevm_gateway_bindings` to the final
  `feat/RFC-012-016` merge commit before merge.


manoranjith:
@obatirou @eudelins-zama @isaacdecoded @maksymsur 

The contract version is bumped in this commit: https://github.com/zama-ai/fhevm/pull/2544/changes/0b825d49cbfb302ea91de3ca95b5571baf7bc94b

relayer/relayer-migrate/migrations/20260519152353_drop_legacy_user_decrypt_req_type.sql
```diff
@@ -0,0 +1,66 @@
...
+-- Postgres has no `ALTER TYPE ... DROP VALUE`; the recipe is to
+-- recreate the enum and swap the column type, casting through text
+-- so existing values land unchanged. sqlx runs each migration in a
+-- transaction by default, so a mid-step failure rolls back cleanly.
+
+-- Safety: refuse to run if any rows still carry the `legacy` value.
+-- Cast through text so the comparison works regardless of which
+-- variants the current enum has (e.g. if a previous attempt got
+-- partway through and the enum was already partially rewritten).
+DO $$
```
nboisde, line 19: Normally good due to cron, but lets have a look into the different db environment to guarantee there is no leftovers.

relayer/relayer-migrate/migrations/20260519152353_drop_legacy_user_decrypt_req_type.sql
```diff
@@ -0,0 +1,66 @@
...
+-- type while a default of the old type is set), then cast through
+-- text so the surviving values map straight across.
+ALTER TABLE user_decrypt_req
+    ALTER COLUMN req_type DROP DEFAULT,
+    ALTER COLUMN req_type TYPE user_decrypt_req_type
+        USING req_type::text::user_decrypt_req_type;
+
+DROP TYPE user_decrypt_req_type_old;
+
+COMMENT ON TYPE user_decrypt_req_type IS
```
nboisde, line 58: 👍

relayer/src/core/event.rs
```diff
@@ -4,6 +4,7 @@ use crate::http::endpoints::v2::types::DelegatedUserDecryptRequestJson;
 use crate::http::endpoints::v2::types::{
```
nboisde, line 4: If I understand correctly, UserDecrypt events are now unified under only one set of events which are branched by attestation kind ? 
LGTM.

relayer/src/core/user_decrypt_request.rs
```diff
@@ -1,89 +1,154 @@
...
+                allowed_contracts,
+                request_validity: _,
+                signature: _,
+                public_key,
+                extra_data,
+            } => {
+                // Variant tag prevents v3 unified hashes from colliding
+                // with v2 legacy hashes that share the same user_address +
+                // handles.
+                hasher.update(b"variant:eip712_unified_v1:");
```
nboisde, line 113: Nice check for requests caching.

relayer/src/gateway/arbitrum/transaction_calldata.rs
```diff
@@ -1,11 +1,14 @@
 use crate::core::{
     errors::EventProcessingError,
-    event::{DelegatedUserDecryptRequest, UserDecryptRequest},
+    event::{HandleContractPair, UserDecryptRequest},
 };
 use crate::gateway::arbitrum::bindings::{
     Decryption,
-    Decryption::CtHandleContractPair,
-    IDecryption::{ContractsInfo, DelegationAccounts, RequestValidity},
+    Decryption::{CtHandleContractPair, HandleEntry as SolHandleEntry},
```
nboisde: Just find the name SolHandleEntry confusing, what do you mean here by Sol ?
same for the function: sol_ct_handle_contract_pairs and SolRequestValidity
If i understand correctly could it be better to prefix as Host ?

relayer/src/host/acl_checker.rs
```diff
@@ -362,6 +365,114 @@ impl HostAclChecker {
...
+    /// - When `owner_address == user`, the entry is a direct access and
+    ///   only `isAllowed(handle, user)` is asserted.
+    /// - Otherwise the entry is delegated, and
+    ///   `isHandleDelegatedForUserDecryption(owner, user, contract, handle)`
+    ///   is asserted (the contract itself rolls in `isAllowed(handle,
+    ///   owner)`, `isAllowed(handle, contract)`, and the delegation lookup).
+    ///
+    /// Calls are grouped by host chain id (extracted from the handle bytes)
+    /// and dispatched through one multicall per chain.
+    pub async fn check_unified_user_decrypt(
```
nboisde, line 380: Delegation is now verified as a branching arm, great, LGTM.

relayer/src/http/endpoints/common/types.rs
```diff
@@ -46,3 +46,45 @@ pub struct RequestValidityJson {
...
+#[derive(Debug, Deserialize, Clone, Serialize, Hash, ToSchema, Validate)]
+#[serde(rename_all = "camelCase")]
+pub struct HandleEntryJson {
+    /// Ciphertext handle from an on-chain FHE operation. `0x` + 64 hex chars.
+    #[schema(example = "0xabcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890")]
+    pub ct_handle: String,
+    /// Address of the contract that produced this ciphertext handle. `0x` + 40 hex chars.
+    #[schema(example = "0x1234567890123456789012345678901234567890")]
+    pub contract_address: String,
+    /// Owner address for this handle. For direct-access handles this equals
```
nboisde, line 62: Good to know

nboisde (dismissed):
Not that much to correct on this, flows looks good to me ! Thanks @manoranjith for this huge work. 
Using the same internals for Both legacy userDecrypt and delegation is much more clear, nonetheless we should not forget during deployment on tweaking the configuration variables accordingly taking into account that we now have 3 flows into one throttler instead of 2 different throttlers for 2 flows.

relayer/relayer-migrate/migrations/20260519152353_drop_legacy_user_decrypt_req_type.sql
manoranjith: I agree, could be checked manually since we have read access to all DBs before applying. I will add it in the release notes for this release.

relayer/src/core/event.rs
manoranjith: Yes, this is correct :+1:

manoranjith:
> Not that much to correct on this, flows looks good to me ! Thanks @manoranjith for this huge work. Using the same internals for Both legacy userDecrypt and delegation is much more clear, nonetheless we should not forget during deployment on tweaking the configuration variables accordingly taking into account that we now have 3 flows into one throttler instead of 2 different throttlers for 2 flows.

I agree, this looks like a change. But I guess at KMS level all the three would be just "user decrypt" operations, so maybe using just one limit should be fine? Wdyt?

relayer/src/gateway/arbitrum/transaction_calldata.rs
manoranjith: Agreed, better approach would be to import only `Decryption` / `IDecryption` contracts into scope and use the types for solidity bindings as `Decryption::ContractsInfo` and so on.

Updated in this commit: https://github.com/zama-ai/fhevm/pull/2544/commits/3f3c6e42d34411432f20c5f7c849edb9cfcd7f0e

nboisde:
I agree @manoranjith , its even better to have everything in one single user-decrypt operation flow, just mentioning that we could refine the setting, now since the flows are using the same throttler. We can if the value is not set at "max capacity", increase it.

relayer/src/gateway/arbitrum/transaction_calldata.rs
nboisde: LGTM ! Thanks, its cleaner this way ! Thanks Mano !

nboisde (dismissed):
LGTM ! 🙏

manoranjith:
@mergify queue

nboisde (approved):

manoranjith:
@mergify queue

PR #2552, opened 2026-05-19
chore: N-05 single deadline for user decryption

manoranjith:
Implements [N-05][issue].

[issue]: https://github.com/zama-ai/fhevm-internal/issues/1373

Reinitializer not bumped — the prior `Decryption` reinitializer bump
already on this branch covers this version, per the [issue
thread][issue-comment].

Deviations from the issue's sketch:

- Field named `requestDeadline` (not `deadline`) so it reads
  unambiguously in the signed EIP-712 payload and in event data,
  where the parent function name doesn't disambiguate.
- Error kept as `UserDecryptionRequestExpiredSeconds` (now carries
  `uint256 deadline`). Solidity disallows overloading the legacy
  `UserDecryptionRequestExpired`, so the suffix is parked until the
  legacy paths are removed.
- The "deadline beyond max" test pins `block.timestamp` via
  `evm_setNextBlockTimestamp` — without it Hardhat's timestamp
  drift made `now + MAX + 1` resolve inside the cap by the time the
  contract evaluated it.

Compatibility: non-breaking in-place update. v0.13 shipped the
unified format only in `kms-connector` and the gateway, not in the
relayer or any SDK, so no external client is signing the old shape.
Older relayers stay on the legacy
`userDecryptionRequest(CtHandleContractPair[], …)` /
`delegatedUserDecryptionRequest` overloads, which are untouched.

[issue-comment]: https://github.com/zama-ai/fhevm-internal/issues/1373#issuecomment-4475601865

## Test plan

- [ ] CI: `gateway-contracts` Hardhat suite (locally: 104/104 pass).
- [ ] CI: `gateway-contracts` selectors / bindings conformance
      (regenerated via `make update-conformance`).
- [ ] CI: `relayer` `cargo test --lib` (locally: 185/185 pass).
- [ ] CI: `relayer` integration-tests feature builds.

## Follow-ups (out of scope)

- KMS-connector / SDK adaptation to the `requestDeadline` wire
  field. Deferred because the unified path isn't reaching them via
  real traffic yet.
- Rename `UserDecryptionRequestExpiredSeconds` →
  `UserDecryptionRequestExpired` once the legacy paths are removed.
- `relayer/Cargo.lock` will pick up a `source = "git+…"` line for
  `fhevm_gateway_bindings` on first `cargo build` after this branch
  lands (currently a path-resolution residue; cargo refuses to
  refresh against the unpushed rev).


manoranjith:
We decided not to proceed with this change and drop it. Because combining `startTimestamp` and `duration` into `deadline` implies 

- `deadline` need not be computed, but can be used as-is.
- But, we loose access to information on "from when" the attestation is valid, which is needed for validating a sub set of signature types as described in RFC-016 (refer to [this section](https://github.com/zama-ai/tech-spec/blob/main/rfcs/016-unified-eip712-decryption-request.md#acl-contract-changes-host-chain)).

@eudelins-zama @maksymsur @isaacdecoded @nboisde

PR #2565, opened 2026-05-20
feat(common): implement ciphertext-attestation crate

eudelins-zama:
## What it does

Adds a new shared Rust crate (`shared/ciphertext-attestation`) that holds the **single source of truth** for the off-chain ciphertext attestation format from RFC-023. Both the Coprocessor (signing) and the KMS Connector (verifying) will depend on this crate so the two sides cannot diverge on encoding.

## Why

RFC-023 removes the on-chain `CiphertextCommits` consensus and replaces it with signed attestations published by Coprocessors alongside ciphertext uploads. Producer and consumer must encode, sign, and verify those attestations **byte-identically** — a shared crate is the only safe way to guarantee that.

## What's in the crate

- **`CiphertextAttestationPayload`** — the full signed view (handle, key id, context id, both digests, version).
- **`canonical_bytes()` / `canonical_digest()`** — the deterministic `abi.encodePacked(...)`-equivalent encoding with a `bytes8("FHEVMCTA")` domain tag, hashed with keccak256.
- **`sign()` / `verify()`** — raw-prehash secp256k1 over the canonical digest; verifier reconstructs `handle` and `coprocessor_context_id` from the S3 lookup path.
- **`CiphertextAttestation`** — the JSON wire form persisted as the `x-amz-meta-ct-attestation` S3 metadata header. `handle` and `context_id` are intentionally stripped (bound by the signature, supplied from the path).
- **`Version` enum** with `try_from = "u8"` so unknown versions are rejected at parse time.

Closes https://github.com/zama-ai/fhevm-internal/issues/1409


.github/workflows/shared-crates-tests.yml
```diff
@@ -0,0 +1,84 @@
...
+        persist-credentials: 'false'
+    - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36  # v3.0.2
+      id: filter
+      with:
+        # Define paths that trigger specific component workflows
+        filters: |
+          ciphertext-attestation:
+            - shared/ciphertext-attestation/**
+          user-decryption-signature:
+            - shared/user-decryption-signature/**
```
eudelins-zama, line 40: To avoid conflict with `feat/RFC-012-016` branch.

The crate does not exist yet, but the CI is just ignoring the tests of this crate so all good.

shared/ciphertext-attestation/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+/// Not serializable: `handle` and `coprocessor_context_id` are intentionally
+/// stripped from the wire form (the verifier reconstructs them from the S3
+/// lookup path).
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CiphertextAttestationPayload {
+    pub version: Version,
+    pub handle: B256,
+    pub key_id: U256,
+    pub coprocessor_context_id: U256,
+    pub ciphertext_digest: B256,
```
rudy-6-4, line 79: Why do we have both digest here ? Only one will be checked against the ciphertext.

eudelins-zama: This is what we agreed on in the [RFC](https://github.com/zama-ai/tech-spec/blob/main/rfcs/023-ciphertext-commits-removal.md#attestation-format).
This was to stay as close as the `CiphertextCommits` [addCiphertextMaterial](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/CiphertextCommits.sol#L116-L121) implementation (which this guild aim to replace)

That being said, I am not familiar enough with the topic to understand the purpose of each digest.
Why were we "uploading" both on `CiphertextCommits` in the first place?
Can you start an internal thread to discuss about this if important?

shared/ciphertext-attestation/src/sign.rs
```diff
@@ -0,0 +1,232 @@
...
+        let recovered = sig
+            .recover_address_from_prehash(&digest)
+            .map_err(|e| AttestationError::Recovery(e.to_string()))?;
+        if recovered != self.signer {
+            return Err(AttestationError::SignerMismatch {
+                recovered,
+                expected: self.signer,
+            });
+        }
+        Ok(recovered)
```
rudy-6-4: why is it returned ? since we have checked it's self.signer ?

rudy-6-4 (dismissed):

shared/ciphertext-attestation/src/sign.rs
eudelins-zama: Good catch.

Fixed in f5347d13eef566a83403acf0b543b2f4f02a017b

rudy-6-4 (approved):

eudelins-zama:
@mergifyio queue

eudelins-zama:
@mergifyio queue

PR #2567, opened 2026-05-20
test(test-suite): make delegated expiry portable

Eikix:
## Summary

- Make delegated user decryption expiry tests portable across deployed ACL behavior.
- Probe whether the deployed smart-wallet/ACL path accepts a 75-second expiry before sending a transaction.
- If accepted, live networks wait for the short expiry and local networks can fast-forward.
- If the probe fails with the v0.11 `ExpirationDateBeforeOneHour()` custom error, skip only this expiry case on live networks instead of reverting or waiting an hour.
- If a local network returns the same v0.11 error, use the 3660-second compatibility expiry and fast-forward.
- Re-throw any other probe failure so unrelated live regressions are not hidden as skips.

## Validation

- `git diff --check -- test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts`
- `npm run tsc -w test-suite/e2e` could not run because `tsc` is not installed in this worktree.
- Previous `test-suite-e2e-tests` workflow dispatch passed for `negative-acl` before the selector-tightening amendment: https://github.com/zama-ai/fhevm/actions/runs/26155256612

### DIFF
--- a/test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
+++ b/test-suite/e2e/test/delegatedUserDecryption/delegatedUserDecryption.ts
@@ -4,17 +4,17 @@ import { ethers } from 'hardhat';
 
 import { EncryptedERC20, SmartWalletWithDelegation, WildcardDelegationTarget } from '../../types';
 import { aclAddress, createInstances } from '../instance';
+import { isLiveNetwork } from '../network';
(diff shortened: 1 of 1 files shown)


PanGan21 (approved):

Eikix:
@mergifyio queue

Eikix:
@mergifyio queue

tawadaa (approved):

PR #2591, opened 2026-05-20
fix(host-contracts): acl delegation expiration boundary

obatirou:
Make read user decryption delegations inactive at the exact expiration timestamp to match write path


isaacdecoded (approved):
Great! 👍

rudy-6-4 (approved):

PR #2592, opened 2026-05-20
fix(gateway-contracts): add context validation and pausing modifier

isaacdecoded:
Refs https://github.com/zama-ai/fhevm-internal/issues/1426


gateway-contracts/contracts/interfaces/IGatewayConfig.sol
```diff
@@ -378,8 +386,9 @@ interface IGatewayConfig {
 
     /**
      * @notice Update the KMS context: nodes and thresholds for a given context ID.
-     * @dev ⚠️ This function should be used with caution as it can lead to unexpected behavior in
-     * some requests and the contracts should first be paused. It will be deprecated in the future.
+     * @dev Requires `Decryption` to be paused first — shifting `currentKmsContextId` changes the
+     *      context that new requests with empty extraData get pinned to, and the operator must
+     *      first drain any in-flight requests pinned to the outgoing context.
```
zmalatrax, line 391: No more deprecation notice on `updateKmsContext`?

gateway-contracts/test/Decryption.ts
```diff
@@ -550,6 +550,16 @@ describe("Decryption", function () {
       ).to.be.revertedWithCustomError(decryption, "DecryptionContextMismatch");
     });
 
+    it("Should revert when extraData pins a null contextId", async function () {
```
zmalatrax: worth mentioning in the test that it is a null contextId for a non-zero version, no? (i.e. v1 extraData)

gateway-contracts/contracts/interfaces/IGatewayConfig.sol
isaacdecoded: The `updateKmsContext` method is not deprecated yet since we still need to keep the KMS contexts on GatewayConfig synced with ProtocolConfig for the decryption use cases. In other words, when we update the KMS contexts, we need to call the `updateKmsContext` in both contracts.

zmalatrax: I meant, a "future deprecation" ; Previous version had this:
>  @dev ⚠️ This function should be used with caution as it can lead to unexpected behavior in
     * some requests and the contracts should first be paused. It will be deprecated in the future.

zmalatrax (approved):

gateway-contracts/contracts/interfaces/IGatewayConfig.sol
isaacdecoded: Ah I see. It's true. I think it's a bit vague anyway.

gateway-contracts/test/Decryption.ts
isaacdecoded: Good catch, find [here the clarifications](https://github.com/zama-ai/fhevm/pull/2592/commits/1a21f459bcde0369ac68f54187ed328f281ccd7c).

zmalatrax (approved):

PR #2595, opened 2026-05-20
fix(host-contracts): re-encode fheMulDiv scalarByte as rhs-indexed bitmask

PanGan21:
### Summary
- `fheMulDiv` encodes scalarByte as a rhs-indexed bitmask: 0x01 (rhs encrypted) and 0x03 (rhs scalar)
- NatSpec added on `fheMulDiv` documenting the two legal values.

Closes: https://github.com/zama-ai/fhevm-internal/issues/1425


PanGan21:
@claude /pr-review

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -722,9 +722,9 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
     /**
      * @notice              Computes FHEMulDiv operation: (lhs * rhs) / divisor with intermediate widening.
      * @param lhs           LHS (always encrypted).
-     * @param rhs           RHS (encrypted when scalarByte=0x00, plaintext scalar when scalarByte=0x01).
+     * @param rhs           RHS (encrypted when scalarByte=0x01, plaintext scalar when scalarByte=0x03).
```
jatZama: I think the arguments names are a bit misleading here, lhs and rhs are good for binary operators, but since mulDiv is ternary, the rhs name for the middle argument is not great... Not sure what is the best name tbh, but maybe something like `fheMulDiv(factor1, factor2, divisor, scalarByte)` is already less confusing, wdyt?

PanGan21: I think renaming to factor1, factor2 make sense, will work on it.
Only issue is that it will be different from existing `fheMul` which is using `lhs` and `rhs` (hence the naming in muldiv).

PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2595/commits/b7be2725023f719c14fa3f6deb1b7d738abe7f89

jatZama (dismissed):

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -520,11 +520,11 @@ impl Database {
                 insert_computation(tx, result, &deps, &NO_SCALAR).await
             }
 
-            E::FheMulDiv(C::FheMulDiv { lhs, rhs, divisor, scalarByte, result, .. }) => {
-                if scalarByte.const_is_zero() {
-                    insert_computation_bytes(tx, result, &[lhs, rhs], &[divisor.to_vec()], &NO_SCALAR).await
+            E::FheMulDiv(C::FheMulDiv { factor1, factor2, divisor, scalarByte, result, .. }) => {
+                if fhe_mul_div_factor2_is_scalar(scalarByte) {
+                    insert_computation_bytes(tx, result, &[factor1], &[factor2.to_vec(), divisor.to_vec()], &HAS_SCALAR).await
```
Eikix, line 525: why did we go from &[lhs, rhs] to &[factor1], and then the factor2 in the second slice?

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -998,26 +998,26 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
-        bytes32 lhs,
-        bytes32 rhs,
+        bytes32 factor1,
+        bytes32 factor2,
         bytes32 divisor,
         bytes1 scalarByte,
         FheType resultType
     ) internal virtual returns (bytes32 result) {
-        _checkBoolean(scalarByte);
+        if (scalarByte != 0x01 && scalarByte != 0x03) revert ScalarByteIsNotBoolean();
```
Eikix: where can we find out why scalar byte must be 0x01 or 0x03? why not 0x00

Eikix: ah just saw the comment below

obatirou: `ScalarByteIsNotBoolean` is weird with this condition

obatirou: `InvalidMulDivScalarByte` would be more appropriate?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
PanGan21: The condition was flipped from `const_is_zero ` to `factor2_is_scalar ` so the branch bodies also got swapped.

host-contracts/contracts/FHEVMExecutor.sol
PanGan21: I added NatSpec for the scalarByte for now since the logic is not documented elsewhere. Where do you think is the best place to have this documented since it is just a formula but it is repeated for many ops?

PanGan21: Would you agree if I add a new error `InvalidScalarByte`? 
The reason I reused this error is this specific comment: https://github.com/zama-ai/fhevm/blob/b7be2725023f719c14fa3f6deb1b7d738abe7f89/host-contracts/contracts/FHEVMExecutor.sol#L61

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -998,26 +998,26 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
 
-        if (!ACL.isAllowed(lhs, msg.sender)) revert ACLNotAllowed(lhs, msg.sender);
-        if (scalarByte == 0x00) {
-            if (!ACL.isAllowed(rhs, msg.sender)) revert ACLNotAllowed(rhs, msg.sender);
-            // resultType == _typeOf(lhs) (set by the caller's _verifyAndReturnType).
-            if (resultType != _typeOf(rhs)) revert IncompatibleTypes();
+        if (!ACL.isAllowed(factor1, msg.sender)) revert ACLNotAllowed(factor1, msg.sender);
+        if (scalarByte == 0x01) {
```
obatirou: could we use constants instead of magic numbers?

host-contracts/contracts/FHEVMExecutor.sol
PanGan21: I refactored to use `InvalidMulDivScalarByte` in: https://github.com/zama-ai/fhevm/pull/2595/changes/bdf99a502d66f3b2875d8f5db56fe9b4b22b9a07

obatirou: `InvalidScalarByte` / `InvalidMulDivScalarByte` is fine to me.
I think natspec conflicting with name is confusing and explicitness is better

host-contracts/contracts/FHEVMExecutor.sol
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2595/commits/758c8da25c8bcee0140a06357a09af833cd6d433

host-contracts/contracts/FHEVMExecutor.sol
Eikix: tbh no idea 🤷
maybe tech spec

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -151,6 +154,10 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
     bytes8 private constant COMPUTATION_DOMAIN_SEPARATOR = "FHE_comp";
     bytes8 private constant SEED_DOMAIN_SEPARATOR = "FHE_seed";
 
+    /// `fheMulDiv` `scalarByte` bitmask: bit 0 (divisor) is always set; bit 1 marks `factor2` as scalar.
+    bytes1 private constant FHE_MUL_DIV_SCALAR_BYTE_ENC = 0x01;
+    bytes1 private constant FHE_MUL_DIV_SCALAR_BYTE_SCALAR = 0x03;
```
obatirou: nit:
* `FHE_MUL_DIV_FACTOR2_ENCRYPTED`
* `FHE_MUL_DIV_FACTOR2_SCALAR`

PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2595/commits/3d400f2f2afa16758c9a95bcd7219757d1a32ac5

obatirou (approved):

Eikix (approved):

PanGan21:
@mergify queue

PR #2598, opened 2026-05-20
feat(host-contracts): add per-context KMS threshold setters

obatirou:
Adds 4 ACL-owner-only setters on `ProtocolConfig` to update each KMS threshold (`publicDecryption`, `userDecryption`, `kmsGen`, `mpc`) on any valid, non-destroyed context. Emits a per-field event; reuses existing threshold validation.


obatirou:
@claude /pr-review

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -158,6 +158,48 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
         emit KmsContextDestroyed(kmsContextId);
     }
 
+    /// @inheritdoc IProtocolConfig
+    function updatePublicDecryptionThresholdForContext(
```
PanGan21, line 162: I understand that KMSGeneration pins `contextId` but it still reads the threshold at the moment of the request. Therefore this function is able to update threshold for on going keygen requests. Is this intended behavior?

obatirou: For keygen consensus, `updatePublicDecryptionThresholdForContext` does not matter imo. In any cases changing threshold for user / public decryption should not disrupt anything in a correct setting as all nodes should answer genuinely. So passing from 3/5 to 4/5 is ok 
On the other hand keygen uses `kmsGenThreshold`, so maybe `updateKmsGenThresholdForContext` should be restricted. But calling this function is a DAO operation. So maybe only a comment is in order?

fyi on old gatewayConfig there was no restriction https://github.com/zama-ai/fhevm/blob/c83929ea8128a100c0801031165a706960e9e606/gateway-contracts/contracts/GatewayConfig.sol#L394

And use a comment in interface https://github.com/zama-ai/fhevm/blob/c83929ea8128a100c0801031165a706960e9e606/gateway-contracts/contracts/interfaces/IGatewayConfig.sol#L276

Let me update the comment to match it. Good catch!

obatirou: Oups wrong link 
https://github.com/zama-ai/fhevm/blob/c83929ea8128a100c0801031165a706960e9e606/gateway-contracts/contracts/interfaces/IGatewayConfig.sol#L336

No comment on `updateKmsGenThresholdForContext`

obatirou: Wondering if necessary as CI is 1h at least and audit begins now

PanGan21: Makes sense, thanks for the explanation

PanGan21 (approved):

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: I agree with Yohan that since these are admin methods, it's the responsibility of the operators to call them when necessary and to control the impact of changing these configurations. This was like that in the GatewayConfig in the past at least.

PR #2599, opened 2026-05-20
fix(coprocessor): fix a SNS worker race on ciphertext128_format

antoniupop:


coprocessor/fhevm-engine/sns-worker/src/lib.rs
```diff
@@ -247,15 +247,36 @@ impl HandleItem {
...
+            .await?;
+        } else {
+            let ct128_format: i16 = ct128_format.into();
+            sqlx::query(
+                "INSERT INTO ciphertext_digest (
+                    tenant_id, handle, transaction_id, ciphertext128_format
+                )
+                VALUES ($1, $2, $3, $4)
+                ON CONFLICT (tenant_id, handle) DO UPDATE
+                SET ciphertext128_format = EXCLUDED.ciphertext128_format
```
goshawk-3, line 276: Since it's a race, isn't it possible to end up with different ordering where the format:10 overwrites the format 11

```
15:20:39.411Z	Mark ct128 uploaded — digest 0x9becf5e0..., format: 11
15:20:39.415Z	Mark ct128 uploaded — digest 0x9becf5e0..., format: 10
```

Maybe we can just set format:11 as default in DB for now.

antoniupop: In current S3 no overwriting, first wins.
Changing DB is not justifiable IMO (and not a small fix) - I would be more inclined to set it in the query with a COALESCE, but I don't think that's needed if we can just set the right format to begin with.

antoniupop: The initial ordering was 10 then 11, so 10 stuck and that's why we had the problem. If the 11 then 10 ordering happens, then no issue would occur.

goshawk-3: I meant this small fix (just a migration): 
```
ALTER TABLE ciphertext_digest
ALTER COLUMN ciphertext128_format
SET DEFAULT 11;
```

> In current S3 no overwriting, first wins.

Since the `check_object_exists` and `S3 put_object` calls are not atomic, I don't think that is always true.

antoniupop: Aside from the issue of pushing a DB migration to prod to fix a SW bug (which I do not think is reasonable), using a DB default as a way of getting out of a race issue doesn't feel right. We'd just hide the problem, then later if we have a different format (e.g. switch to GPU), we'd get the same bug pop up and want to set default to a new value.

antoniupop: Regarding the race fix in this PR, we would just not have the possibility of having a race with different formats, so we can still have a race, but between two writes that are both doing format = 11 and upload the same CT for the same digest. I can add a fix to avoid the race on write to S3, but that increases the risk of introducing bugs along with it. I'd do that in main or possibly v0.13, but not further back.

goshawk-3: > We'd just hide the problem 

You know that by having ` SET ciphertext128_format = EXCLUDED.ciphertext128_format` you also hide the problem. Here both of us are talking about a hotfix/patch. 

Apart from the hotfix, we need to introduce a real fix to address the race condition properly before we switch to GPU.

 If you feel more comfortable with the DB override then go for it.

goshawk-3 (approved):

rudy-6-4 (approved):

goshawk-3 (approved):

PR #2604, opened 2026-05-21
feat(common): coprocessor blue-green upgrade protocol

goshawk-3:
RFC-021-blue-green-upgrade: rfcs/021-blue-green-upgrade.md


coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// Conditional UPDATE: only flips if the GCS row is still in `UpgradeActivated`.
+/// Always followed by an `event_unpause` notify with `{chain_id, start_block}`.
+async fn transition_to_dry_run_started(pool: &Pool<Postgres>) -> Result<(), Error> {
```
PanGan21, line 565: Per RFC, shouldn't we emit pg_notify(event_unpause) here?. `UNPAUSE_CHANNEL` is defined but never used and I think without this, dry-run will not start

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+    let rows: Vec<(String, Option<i64>)> =
+        sqlx::query_as("SELECT state, end_block FROM upgrade_state WHERE status = 'in_progress'")
+            .fetch_all(pool)
+            .await?;
+
+    match rows.len() {
+        0 => Ok(None),
+        1 => {
+            let (state, end_block) = &rows[0];
+            if state != "UpgradeActivated" {
```
PanGan21: Shouldn't you check `if state != "DryRunStarted"` here?
I understand that consensus check should run after the dry run starts which is when the state is `DryRunStarted`

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+        None => {
+            return Err(Error::Payload(
+                "no GCS row in upgrade_state — cannot run cutover".to_string(),
+            ));
+        }
+    };
+
+    let mut tx = pool.begin().await?;
+
+    sqlx::query("SELECT pg_advisory_xact_lock($1)")
```
PanGan21, line 828: Is it correct that we use advisory locks instead of `LOCK TABLE` as the RFC mentions? (just trying to understand why)

goshawk-3: Correct. The latest idea  is to change this approach and use a table-level lock instead.

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
goshawk-3: Indeed. We cannot reach unanimity until all parties have enabled the dry-run mode.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
goshawk-3: This is up to consideration yet.

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -41,6 +41,25 @@ interface IProtocolConfig {
      */
     event KmsContextDestroyed(uint256 indexed kmsContextId);
 
+    /**
+     * @notice Emitted when a coprocessor upgrade is proposed. The proposal is identified
+     *         by a caller-supplied `proposalId`; the contract does not track adoption state.
+     *         Off-chain machinery (host-listener, upgrade-controller, consensus-detector)
+     *         observes this event and drives the dry-run / consensus / cutover lifecycle.
```
obatirou: it is not clear what `adoption state` is?

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -41,6 +41,25 @@ interface IProtocolConfig {
...
+     * @param proposalId Caller-supplied identifier for this upgrade attempt.
+     * @param softwareVersion The coprocessor software version for the proposal.
+     * @param chainUpgradeWindows The per-host-chain replay windows for the upgrade.
+     * @param gwStartBlock The Gateway block at which GCS's gateway-listener resumes from.
+     * @param ciphertextVersion The ciphertext version the new software writes; promoted into the `versioning` singleton at cutover.
```
obatirou: nit: "promoted into the `versioning` singleton at cutover." seems not necessary to keep?

host-contracts/contracts/interfaces/IProtocolConfig.sol
PanGan21: It means that the function doesn't store anything related yo the proposal id. It just emits an event. I will make the comment more clear

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -20,11 +20,11 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
 
     string private constant CONTRACT_NAME = "ProtocolConfig";
     uint256 private constant MAJOR_VERSION = 0;
-    uint256 private constant MINOR_VERSION = 1;
+    uint256 private constant MINOR_VERSION = 2;
     uint256 private constant PATCH_VERSION = 0;
 
     /// @dev Shared between `initializeFromEmptyProxy` and `initializeFromMigration`.
```
obatirou:
```suggestion
    /// @dev Shared between `initializeFromEmptyProxy`, `initializeFromMigration` and `reinitializeV2`
```

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -131,6 +131,16 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
         _defineKmsContext(existingKmsNodes, existingThresholds);
     }
 
+    /**
+     * @notice Reinitializer for proxies previously initialized at version 2. Reserved for
+     *         the version-3 storage layout; no state changes are required because coprocessor
+     *         context state is no longer tracked on chain (proposals are off-chain-driven and
+     *         identified by caller-supplied `proposalId`s emitted in `CoprocessorUpgradeProposed`).
```
obatirou: nit: seems not necessary to add those details here?
```suggestion
     * @notice Re-initializes the contract from V1.
```

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -159,6 +169,54 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        if (bytes(softwareVersion).length == 0) {
+            revert EmptySoftwareVersion();
+        }
+        if (chainUpgradeWindows.length == 0) {
+            revert EmptyChainUpgradeWindows();
+        }
+        if (gwStartBlock == 0) {
+            revert ZeroGwStartBlock();
+        }
+        if (ciphertextVersion > uint16(type(int16).max)) {
```
obatirou: maybe this check should be documented with a one line comment?
If not mistaken this is due to the fact the ciphertextVersion is stored as SMALLINT/int16 in db right?

host-contracts/contracts/interfaces/IProtocolConfig.sol
PanGan21: Fixed in: https://github.com/zama-ai/fhevm/pull/2604/commits/c54743ce8a390eaa31c354d4bee90fda9564e53b

host-contracts/contracts/interfaces/IProtocolConfig.sol
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2604/commits/c54743ce8a390eaa31c354d4bee90fda9564e53b

host-contracts/contracts/ProtocolConfig.sol
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2604/commits/c54743ce8a390eaa31c354d4bee90fda9564e53b

host-contracts/contracts/ProtocolConfig.sol
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/2604/commits/c54743ce8a390eaa31c354d4bee90fda9564e53b

host-contracts/contracts/ProtocolConfig.sol
PanGan21: Correct doing this in order to avoid overflows. Added the comment here: https://github.com/zama-ai/fhevm/pull/2604/commits/f47c6e04e0a4523432b021361f0c82c295a60c2f

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -0,0 +1,1102 @@
...
+    //    `ON CONFLICT DO UPDATE` lets the GCS rows win on PK collisions: GCS is
+    //    the canonical writer for its window.
+    let merge_sql = format!(
+        "INSERT INTO public.ciphertexts
+             (tenant_id, handle, ciphertext, ciphertext_version, ciphertext_type,
+              input_blob_hash, input_blob_index, created_at, ciphertext128, is_input)
+         SELECT tenant_id, handle, ciphertext, $1::smallint, ciphertext_type,
+                input_blob_hash, input_blob_index, created_at, ciphertext128, is_input
+         FROM {GCS_SCHEMA_QUOTED}.ciphertexts
+         ON CONFLICT (tenant_id, handle, ciphertext_version) DO UPDATE
```
PanGan21: `tenant_id` has been dropped from the composite key. I think we should remove it from the query otherwise `ON CONFLICT` won't match the unique constraint.

coprocessor/fhevm-engine/consensus-detector/src/state_hash.rs
```diff
@@ -0,0 +1,295 @@
...
+/// `public`: hashing the live (BCS) `computations`/`ciphertexts` would silently
+/// produce the wrong digest, so a missing GCS schema must error rather than
+/// read public. Runtime `sqlx::query` because the schema name is only known at
+/// runtime.
+async fn compute_gcs_hash(
+    pool: &Pool<Postgres>,
+    chain_id: i64,
+    block_number: i64,
+) -> anyhow::Result<Option<String>> {
+    let sql = format!(
```
PanGan21, line 112: The state hash here only includes ciphertexts produced by FHE operations. It skips zkproof input ciphertexts, which are located in `input_handles`. 
Maybe we should consider adding them here as well since that means divergent inputs between operators wouldn't be caught by consensus.

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -353,6 +420,102 @@ pub async fn ingest_block_logs(
...
+            }
+        },
+        Err(_) => {
+            PROTOCOL_CONFIG_EVENT_DECODE_FAIL_COUNTER.inc();
+        }
+    }
+    Ok(())
+}
+
+async fn notify_coprocessor_upgrade_proposed(
```
PanGan21, line 607: It should a known issue for now that if upgrade controller is down for some reason the upgrade pg_notify will be lost since this function doesn't persist the upgrade in the database.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -0,0 +1,1102 @@
...
+        proposal_id = %payload.proposal_id,
+        chain_id = payload.chain_id,
+        start_block = payload.start_block,
+        end_block = payload.end_block,
+        gw_start_block = payload.gw_start_block,
+        ciphertext_version = payload.ciphertext_version,
+        "event_upgrade_activated received — inserting upgrade_state row"
+    );
+
+    let result = sqlx::query(
```
PanGan21, line 194: If `execute_cutover` fails here, we're stuck. First step already committed the FSM to `UpgradeAuthorized`, but second step (the actual merge) errored. 
Do you think it is worth adding a retry?

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
PanGan21: tracked for next release

coprocessor/fhevm-engine/consensus-detector/src/state_hash.rs
PanGan21: tracked for next release

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
PanGan21: tracked for next release

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
PanGan21: tracked for next release

charts/coprocessor/values.yaml
```diff
@@ -79,6 +79,11 @@ commonConfig:
   aclContractAddress: {}            # required — e.g. 0x05fD9B5EFE0a996095f42Ed7e77c390810CF660c
   fhevmExecutorContractAddress: {}  # required — e.g. 0x...
   kmsGenerationContractAddress: {}  # optional — e.g. 0x...
+  protocolConfigContractAddress: {}  # required — e.g. 0x...
```
PanGan21: Flagging this. I believe it should be optional and added only for the host listener of the canonical chain (ethereum) following the pattern of other similar contracts

PanGan21: Fixed in: https://github.com/zama-ai/fhevm/pull/2604/commits/b19b824eb07219bef4a9b249baa2ae79f8af118d

goshawk-3:
Duplicated https://github.com/zama-ai/fhevm/pull/3044

PR #2607, opened 2026-05-21
fix(test-suite): run input proof on shared networks

Eikix:
## Summary

Remove the live/shared-network skip from the plain input proof e2e test.

## Root cause

The guard was added to protect intentional ciphertext drift runs from shared/live networks, but it was placed inside `test user input uint64 (non-trivial)`. That test is the normal `input-proof` smoke path and does not install the drift SQL trigger. The drift profile already has a CLI preflight guard where drift injection is actually enabled.

## Impact

`input-proof` now runs on `sepolia`, `mainnet`, and `zwsDev` instead of being reported as skipped. The dedicated `ciphertext-drift` profile remains blocked on those networks.

## Validation

- Inspected the profile mappings and confirmed `input-proof` maps to `test user input uint64` while drift profiles default to `test add 42 to uint64 input and decrypt`.
- `npm run tsc` in `test-suite/e2e` could not run because `tsc` is not installed in this worktree.

### DIFF
--- a/test-suite/e2e/test/userInput/inputFlow.ts
+++ b/test-suite/e2e/test/userInput/inputFlow.ts
@@ -4,9 +4,6 @@ import { ethers } from 'hardhat';
 import { createInstances } from '../instance';
 import { getSigners, initSigners } from '../signers';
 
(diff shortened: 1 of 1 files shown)


0xalexbel (approved):

Eikix:
@mergifyio queue

tawadaa (approved):

PR #2615, opened 2026-05-21
feat(host-contracts): implement epoch lifecycle for KMS context rotation

obatirou:
Closes https://github.com/zama-ai/fhevm-internal/issues/1130 
Closes https://github.com/zama-ai/fhevm-internal/issues/1267


host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -222,17 +222,23 @@ interface IKMSGeneration {
...
      */
     error AbortKeygenInvalidId(uint256 prepKeygenId);
 
     /**
-     * @notice Error thrown when the keygen was already completed and cannot be aborted.
+     * @notice Error thrown when the keygen was already finalized and cannot be aborted.
      * @param prepKeygenId The preprocessing keygen ID.
      */
-    error AbortKeygenAlreadyDone(uint256 prepKeygenId);
+    error AbortKeygenAlreadyFinalized(uint256 prepKeygenId);
```
isaacdecoded: nit: I'm not an English-native writer, but I think "finished" is a better verb here, i.e., `AbortKeygenAlreadyFinished`. Still open to know if I'm wrong.

host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -352,6 +358,33 @@ interface IKMSGeneration {
...
+     * @param keyId The key ID.
+     * @return The preprocessing keygen ID.
+     */
+    function getPrepKeygenId(uint256 keyId) external view returns (uint256);
+
+    /**
+     * @notice Get the context and epoch encoded in a request's extra data.
+     * @param requestId The request ID.
+     * @return contextId The KMS context ID pinned to the request.
+     * @return epochId The epoch ID pinned to the request, or 0 for legacy v0/v1 requests.
```
isaacdecoded: For v0/v1, it should probably return the [default epoch ID that the KMS uses](https://github.com/zama-ai/kms/blob/main/core/service/src/consts.rs#L168-L173) for no specified epoch, i.e., `0x080....01`.

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -23,23 +24,117 @@ interface IProtocolConfig {
...
+    event NewEpochForCurrentKmsContext(
+        uint256 indexed kmsContextId,
+        uint256 indexed epochId,
+        uint256 previousEpochId,
+        PreviousKeyInfo[] keys,
+        PreviousCrsInfo[] crsList
+    );
+
+    /**
+     * @notice Emitted when enough old and new signers confirm a pending KMS context was created.
```
isaacdecoded:
```suggestion
     * @notice Emitted when enough old and new signers confirm that a pending KMS context was created.
```

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -23,23 +24,117 @@ interface IProtocolConfig {
...
+     * @param keyIds Key IDs included in the activation.
+     * @param crsIds CRS IDs included in the activation.
+     * @param kmsNodeStorageUrls Storage URLs for nodes in the activated context.
      */
-    event NewKmsContext(uint256 indexed kmsContextId, KmsNode[] kmsNodes, KmsThresholds thresholds);
+    event ActivateEpoch(
+        uint256 indexed kmsContextId,
+        uint256 indexed epochId,
+        uint256[] keyIds,
+        uint256[] crsIds,
```
isaacdecoded: Interesting, this means that we actually don't need to propagate the `EpochKeyResult` and the `EpochCrsResult` structs in this event, right? 🤔 I probably misunderstood it when I thought the Coprocessor needed to validate the EIP-712 signatures, which was wrong.

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -23,23 +24,117 @@ interface IProtocolConfig {
...
+        string[] kmsNodeStorageUrls
+    );
 
     /**
      * @notice Emitted when a KMS context is destroyed.
      * @param kmsContextId The destroyed context ID.
+     * @param cancelledEpochId The pending epoch cancelled with the context, or 0 if none.
      */
-    event KmsContextDestroyed(uint256 indexed kmsContextId);
+    event KmsContextDestroyed(uint256 indexed kmsContextId, uint256 indexed cancelledEpochId);
```
isaacdecoded: Do we really need this `cancelledEpochId` here? I think that if the context is destroyed, it implicitly means all its linked epochs are also destroyed.

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -121,16 +216,144 @@ interface IProtocolConfig {
...
+     * @notice Confirm activation of a pending epoch.
+     * @dev All confirming KMS tx senders must submit the same `(keyIds, crsIds)` arrays, ordered
+     * identically. The first confirmation establishes the canonical set; later confirmations must
+     * match it exactly. Each referenced ID must already be a completed (non-aborted) entry in
+     * `KMSGeneration`.
+     * @param epochId The pending epoch ID.
+     * @param keyIds The key IDs to associate with the epoch.
+     * @param crsIds The CRS IDs to associate with the epoch.
+     */
+    function confirmEpochActivation(uint256 epochId, uint256[] calldata keyIds, uint256[] calldata crsIds) external;
```
isaacdecoded: Is this intended to have this `confirmEpochActivation` method overloaded? Also, I think the correct one is the one below since this method needs to validate all the EIP-712 signatures, right?

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -167,11 +390,49 @@ interface IProtocolConfig {
...
+     */
+    function getActiveKmsContextAndEpoch() external view returns (uint256 contextId, uint256 epochId);
+
+    /**
+     * @notice Returns the active KMS context and epoch IDs.
+     * @dev Backward-compatible alias for `getActiveKmsContextAndEpoch`.
+     * @return contextId The active context ID.
+     * @return epochId The active epoch ID.
+     */
+    function getCurrentKmsContextAndEpoch() external view returns (uint256 contextId, uint256 epochId);
```
isaacdecoded, line 452: I think we don't need this backward compatibility since there's no other component calling this method at this time, or am I missing something? Meaning that we can just implement `getActiveKmsContextAndEpoch` - and probably update the RFC005.

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -167,11 +390,49 @@ interface IProtocolConfig {
...
+
+    /**
+     * @notice Returns the context and epoch new KMS generation requests should target.
+     * @dev Pending lifecycle transitions prepare material for the pending epoch. Otherwise,
+     *      requests target the active context and epoch.
+     * @return contextId The target KMS context ID.
+     * @return epochId The target epoch ID.
+     */
+    function getKmsGenerationTargetContextAndEpoch() external view returns (uint256 contextId, uint256 epochId);
```
isaacdecoded: Why is this needed? I think keygen/crsgen must use `getActiveKmsContextAndEpoch()`. Note that KMS context rotation and key resharing do not involve a keygen/crsgen, and the same in the opposite way.

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -230,6 +500,18 @@ interface IProtocolConfig {
...
     function getKmsNodeForContext(uint256 kmsContextId, address txSender) external view returns (KmsNode memory);
 
+    /**
+     * @notice Returns KmsNode metadata for generation in an existing context.
+     * @dev Unlike active serving checks, generation can target Pending or Created contexts.
+     * @param kmsContextId The context ID.
+     * @param txSender The tx sender address.
+     * @return The KmsNode struct.
+     */
+    function getKmsGenerationNodeForContext(
```
isaacdecoded: Same question as [above](https://github.com/zama-ai/fhevm/pull/2615/changes#r3288843150).

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -269,6 +551,14 @@ interface IProtocolConfig {
      */
     function getKmsGenThresholdForContext(uint256 kmsContextId) external view returns (uint256);
 
+    /**
+     * @notice Returns the kmsGen threshold for generation in an existing context.
+     * @dev Unlike active serving checks, generation can target Pending or Created contexts.
+     * @param kmsContextId The context ID.
+     * @return The kmsGen threshold for generation in the context.
+     */
+    function getKmsGenerationThresholdForContext(uint256 kmsContextId) external view returns (uint256);
```
isaacdecoded: Same question as [above](https://github.com/zama-ai/fhevm/pull/2615/changes#r3288843150).

host-contracts/contracts/shared/Constants.sol
```diff
@@ -5,3 +5,21 @@ uint8 constant HANDLE_VERSION = 0;
...
+uint256 constant EPOCH_COUNTER_BASE = uint256(0x08) << 248;
+
+/// @dev Base value for preprocessing keygen request IDs. Format: [0x03 type tag | 31 counter bytes].
+uint256 constant PREP_KEYGEN_COUNTER_BASE = uint256(0x03) << 248;
+
+/// @dev Base value for keygen request IDs. Format: [0x04 type tag | 31 counter bytes].
+uint256 constant KEY_COUNTER_BASE = uint256(0x04) << 248;
+
+/// @dev Base value for CRS generation request IDs. Format: [0x05 type tag | 31 counter bytes].
+uint256 constant CRS_COUNTER_BASE = uint256(0x05) << 248;
```
isaacdecoded: nit: maybe worth ordering these constants in ascendency order, i.e., 0x03, 0x04, ..., 0x08

host-contracts/contracts/shared/Structs.sol
```diff
@@ -14,3 +14,34 @@ struct KmsNode {
     /// @notice URL address of the KMS node's storage where ciphertexts are stored
     string storageUrl;
 }
+
+/**
+ * @notice Enclave attestation values for a KMS software build.
+ */
+struct PcrValues {
```
isaacdecoded, line 21: Not sure if we should keep them as `bytes` as they are [defined in the KMS core](https://github.com/zama-ai/kms/blob/main/core/grpc/proto/kms.v1.proto#L590-L594). I mean, the bytes32 conversion could result in a silent truncation (?)

host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -352,6 +358,33 @@ interface IKMSGeneration {
...
+     */
+    function getRequestContextAndEpoch(uint256 requestId) external view returns (uint256 contextId, uint256 epochId);
```
isaacdecoded: Can I ask why we need this view function?

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -212,11 +213,11 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
...
 
     /**
-     * @notice              Returns the current KMS context ID.
-     * @return contextId    The current KMS context ID.
+     * @notice              Returns the active KMS context ID.
+     * @return contextId    The active KMS context ID.
      */
-    function getCurrentKmsContextId() public view virtual returns (uint256) {
-        return PROTOCOL_CONFIG.getCurrentKmsContextId();
+    function getActiveKmsContextId() public view virtual returns (uint256) {
```
isaacdecoded: I think we shouldn't change this method name since it will introduce breaking changes on the client side running on older ABIs.

Also, I don't think we need this `getActiveKmsContextId()` method, we can reuse the `getActiveKmsContextAndEpoch()` method and return only the contextId.

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -167,11 +390,49 @@ interface IProtocolConfig {
     function updateMpcThresholdForContext(uint256 kmsContextId, uint256 threshold) external;
 
     /**
-     * @notice Returns the current active KMS context ID.
-     * @return The current context ID.
+     * @notice Returns the active KMS context ID.
+     * @return The active context ID.
+     */
+    function getActiveKmsContextId() external view returns (uint256);
```
isaacdecoded: I don't think this method is needed if we already have the `getActiveKmsContextAndEpoch()` one. Also see: https://github.com/zama-ai/fhevm/pull/2615/changes#r3288960661

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -34,14 +37,37 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
     // -----------------------------------------------------------------------------------------
     // ERC-7201 namespaced storage
     // -----------------------------------------------------------------------------------------
 
     /// @custom:storage-location erc7201:fhevm.storage.ProtocolConfig
     struct ProtocolConfigStorage {
-        /// @notice Current KMS context ID counter.
-        uint256 currentKmsContextId;
+        /// @notice Highest allocated KMS context ID (monotonic counter).
+        uint256 lastKmsContextId;
```
isaacdecoded: As far as I can remember, we can't rename storage variables or the upgrade will fail due to mismatching storage layout.

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -62,6 +88,50 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        /// @notice Current active KMS context ID.
+        uint256 activeKmsContextId;
+        /// @notice Epoch ID counter.
+        uint256 epochCounter;
+        /// @notice Current active epoch ID.
+        uint256 activeEpochId;
+        /// @notice Pending context switch, if any.
+        uint256 pendingKmsContextId;
+        /// @notice Pending epoch, if any.
+        uint256 pendingEpochId;
```
isaacdecoded: Do we really need to explicitly track these pending KMS context and epoch IDs? Need to read more carefully the ProtocolConfig changes, but I'm wondering if this "pending" state could be inferred from other mappings/state vars.

isaacdecoded (commented):
Contract events LGTM overall. I have some implementation questions here and there.

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: Made the changes. this is not correct

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: deleted

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: After IRL sync, indeed mixed up, I deleted.

host-contracts/contracts/shared/Structs.sol
obatirou: Done

host-contracts/contracts/ProtocolConfig.sol
obatirou: isn't renaming is ok?
You do not change the storage layout. But adding in between or removing is not authorized.

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: Removed

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: The difference in the guard used: have a valid context, exist + active but not really need. We can move this check to the generic one. Will remove those

host-contracts/contracts/shared/Constants.sol
obatirou: done

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: removed as explained above

host-contracts/contracts/interfaces/IKMSGeneration.sol
obatirou: removed

host-contracts/contracts/ProtocolConfig.sol
obatirou: reverted

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: Done

host-contracts/contracts/KMSVerifier.sol
obatirou: Done

host-contracts/contracts/ProtocolConfig.sol
obatirou: I am afraid it will complexify and obscure things. Will look into it

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -89,75 +141,220 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+
         $.currentKmsContextId = KMS_CONTEXT_COUNTER_BASE;
-        uint256 newContextId = _defineKmsContext(initialKmsNodes, initialThresholds);
-        emit NewKmsContext(newContextId, initialKmsNodes, initialThresholds);
+        uint256 newContextId = _storeAndActivateKmsContext(initialKmsNodeParams, initialThresholds);
+
+        $.contextAnchors[newContextId] = KmsContextAnchor({
+            emissionBlockNumber: block.number,
+            contextInfoHash: keccak256(abi.encode(initialKmsNodeParams, initialThresholds, softwareVersion, pcrValues))
+        });
```
isaacdecoded, line 199: I'm missing the reinitializeV2() method, which I think must store the "genesis" KMS context anchor, and emit the `NewKmsContext` with the "genesis sentinel" of `previousContextId = KMS_CONTEXT_COUNTER_BASE` instead of `0`. Also, same for the `initializeFromEmptyProxy` regarding the `previousContextId`. Wdyt?

obatirou: Should be better now, let me know what you think https://github.com/zama-ai/fhevm/compare/38caa50a9052d61aaaf954c7eb53e35fcd8e2aea..84605927c1704231cd22d79f70431ed29e53578c

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -23,24 +24,132 @@ interface IProtocolConfig {
...
+     * @param kmsContextId The activated context ID.
+     * @param epochId The activated epoch ID.
+     * @param keys Key results included in the activation.
+     * @param crsList CRS results included in the activation.
+     * @param kmsNodeStorageUrls Storage URLs for nodes in the activated context.
+     */
+    event ActivateEpoch(
+        uint256 indexed kmsContextId,
+        uint256 indexed epochId,
+        EpochKeyResult[] keys,
```
isaacdecoded, line 112: super-nit: wondering if we should call it `keyList` to match `crsList` 😅

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -857,29 +860,44 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
...
-    function _extractContextIdFromExtraData(bytes memory extraData) internal view virtual returns (uint256 contextId) {
+    function _extractContextIdAndEpochIDFromExtraData(
+        bytes memory extraData
+    ) internal view virtual returns (uint256 contextId, uint256 epochId) {
         // v0 (0x00 prefix or empty): uses the current context. Trailing bytes are
         // ignored for forward-compatibility with potential v0 extensions.
         if (extraData.length == 0 || uint8(extraData[0]) == 0x00) {
-            return PROTOCOL_CONFIG.getCurrentKmsContextId();
+            (uint256 activeContextId, ) = PROTOCOL_CONFIG.getCurrentKmsContextAndEpoch();
+            return (activeContextId, DEFAULT_EPOCH_ID);
```
isaacdecoded: I'm not sure about using the `DEFAULT_EPOCH_ID` here, shouldn't it return the `activeEpochId` returned by the `getCurrentKmsContextAndEpoch` method instead?

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -857,29 +860,44 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
...
         // mload at offset 33 reads 32 bytes starting after the 1-byte version prefix.
         assembly {
             contextId := mload(add(extraData, 33))
         }
+        if (version == EXTRA_DATA_V2) {
+            assembly {
+                epochId := mload(add(extraData, 65))
+            }
+        } else {
+            epochId = DEFAULT_EPOCH_ID;
```
isaacdecoded: Or should it get the `activeEpochId` for the `contextId` above from the ProtocolConfig contract? I mean, it seems error prone using `DEFAULT_EPOCH_ID` for all possible `contextIds`

host-contracts/tasks/utils/protocolConfigMigrationEnv.ts
```diff
@@ -27,13 +35,28 @@ export type ProtocolConfigMigrationThresholds = {
 
 export function buildProtocolConfigInitializeFromMigrationArgs(): [
   bigint,
-  ProtocolConfigMigrationKmsNode[],
+  ProtocolConfigMigrationKmsNodeParams[],
```
isaacdecoded: Or should we remove this migration-specific code already? I mean, are we really executing a migration again with these new KmsNodeParams?

host-contracts/contracts/KMSGeneration.sol
obatirou: I think all in all this should be:  
* V0 : activeContextId + activeEpochId from ProtocolConfig contract.
* V1: contextId from extra data + activeEpochId from the contextId found in extra data
* V2: contexid + epochId from extra data 

What do you think?

host-contracts/contracts/KMSGeneration.sol
obatirou: see https://github.com/zama-ai/fhevm/pull/2615/changes/BASE..b06c36657272d12d3a8189edd289a0ec8878298f#r3336317146

host-contracts/tasks/utils/protocolConfigMigrationEnv.ts
obatirou: This will be reused anyway for the `migrationFromEthereum` / multi chain deployment imo 
For the KMSGeneration we could remove it though.

host-contracts/contracts/KMSGeneration.sol
isaacdecoded: Yes, I think you're right.

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -206,18 +528,43 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
 
     /// @inheritdoc IProtocolConfig
     function getCurrentKmsContextId() external view virtual returns (uint256) {
-        return _getProtocolConfigStorage().currentKmsContextId;
+        ProtocolConfigStorage storage $ = _getProtocolConfigStorage();
+        return $.activeKmsContextId;
+    }
+
+    /// @inheritdoc IProtocolConfig
+    function getCurrentKmsContextAndEpoch() external view virtual returns (uint256 contextId, uint256 epochId) {
```
isaacdecoded, line 617: Seems like we should rename this to `getActiveKmsContextAndEpoch` since "active" is becoming more commonly used than "current" in this PR?

edit: but at the same time, it seems like we can have multiple active contexts/epochs, so not sure if we should keep the "current" naming and instead rename the states `activeKmsContextId/activeEpochId` to `currentKmsContextId/currentEpochId`

edit2 (sorry for the noise): I just realized that we already have a `currentKmsContextId` used for tracking the context counter ID 😅 so not sure how to make this active vs. current naming clearer.

host-contracts/tasks/taskDeploy.ts
isaacdecoded, line 1: Should we also implement the tasks for defining a new context and define new epoch for current KmsContext?

isaacdecoded: In the last sync meeting, we said to create [this follow-up issue](https://github.com/zama-ai/fhevm-internal/issues/1521) for scripts and runbooks regarding the KMS context switch and key resharing triggering. So, we can tackle this hardhat task during this issue if you prefer.

host-contracts/contracts/shared/Structs.sol
```diff
@@ -14,3 +14,38 @@ struct KmsNode {
...
+struct KmsNodeParams {
+    address txSenderAddress;
+    address signerAddress;
+    string ipAddress;
+    string storageUrl;
+    int32 partyId;
+    string mpcIdentity;
+    bytes caCert;
+    /// @notice Full ECDSA public key, matches proto MpcNode.verification_key.
+    bytes verificationKey;
```
isaacdecoded: Based on internal discussions, this field is a duplication of the `signerAddress` above - so we should remove this `verificationKey`. cc @eudelins-zama

host-contracts/tasks/taskDeploy.ts
obatirou: I would prefer to it once this is merge to avoid pushing back this merge if you agree

host-contracts/contracts/ProtocolConfig.sol
obatirou: I think renaming the counter to `kmsContextCounter` would make things better but if I remember correctly you said it was not possible to rename a storage variable or the upgrade will fail due to mismatching storage layout? 
I do not know if this is correct or not.
Regarding the getters, I am in favor of renaming to active but this is ABI break + (if not mistaken) getCurrentKmsContext needs to stay for retro compatibility?

host-contracts/contracts/shared/Structs.sol
obatirou: removed with [aa49644](https://github.com/zama-ai/fhevm/pull/2615/commits/aa496446eaac7be3db127b69599c3428c9d75ad5)

host-contracts/tasks/utils/protocolConfigMigrationEnv.ts
obatirou: KMGengeration part removed in [9599ed5](https://github.com/zama-ai/fhevm/pull/2615/commits/9599ed5b6641f1bfecc20f73ae9ec237879f6608)

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: > I remember correctly you said it was not possible to rename a storage variable or the upgrade will fail due to mismatching storage layout?

Yes, unfortunately, we can't rename state variables.

> Regarding the getters, I am in favor of renaming to active but this is ABI break + (if not mistaken) getCurrentKmsContext needs to stay for retro compatibility?

In this case, there's no backward compatibility to grant since this `getCurrentKmsContextAndEpoch` is a brand-new method, so any component/user is calling it yet. However, let's not block this PR anymore, I'm ok with keeping the `getCurrentKmsContextAndEpoch` naming (or `getCurrentKmsContextAndEpochIds` if we want to be very accurate with the returned data). Again, sorry for the noise.

host-contracts/contracts/KMSGeneration.sol
obatirou: Done with [aa49644](https://github.com/zama-ai/fhevm/pull/2615/commits/aa496446eaac7be3db127b69599c3428c9d75ad5)

-> default for V1 is best effort and not really a problem as not consumed anyway

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: I am ok with it but need RFC update + will affect KMS connector no?

isaacdecoded: Yes, this is why I said "super-nit". So let's not waste more effort here.

host-contracts/contracts/ProtocolConfig.sol
obatirou: yes for `getCurrentKmsContextAndEpoch` there is no retro compatiblity but if we rename it we should also rename `getCurrentKmsContextId` which we cannot do right? 

Will rename to `getCurrentKmsContextAndEpochIds` at least

https://docs.openzeppelin.com/upgrades-plugins/writing-upgradeable#modifying-your-contracts
It seems only type change / order is not possible, rename could 🤔 ?
Capture d’écran 2026-06-16 à 16 43 15

obatirou: Ok my bad https://docs.openzeppelin.com/upgrades-plugins/faq#how-can-i-rename-a-variable-or-change-its-type

isaacdecoded: > https://docs.openzeppelin.com/upgrades-plugins/faq#how-can-i-rename-a-variable-or-change-its-type

Yes, also note that in our case, it's the ERC 7201 (storage layout) that doesn't allow storage variable renaming.

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: Ok closing thread then

host-contracts/contracts/ProtocolConfig.sol
obatirou: Ok let's close this then, seems we are blocked for the current / active

obatirou:
@claude /pr-review

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -304,11 +305,13 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
...
+        if (version == EXTRA_DATA_V1 || version == EXTRA_DATA_V2) {
+            // v1: [version(1)] [contextId(32)]
+            // v2: [version(1)] [contextId(32)] [epochId(32)]
+            if (
+                (version == EXTRA_DATA_V1 && extraData.length < 33) ||
+                (version == EXTRA_DATA_V2 && extraData.length != 65)
+            ) {
                 revert DeserializingExtraDataFail();
             }
             uint256 contextId;
```
geoxel, line 317: Would it be possible to add somewhere a function that returns the ProtocolConfig address, like : 
```
    /**
     * @notice Getter function for the ProtocolConfig contract address.
     * @return protocolConfigAddress Address of the ProtocolConfig contract.
     */
    function getProtocolConfigAddress() public view virtual returns (address) {
        return address(PROTOCOL_CONFIG);
    }
```

isaacdecoded: @geoxel note that if you're asking it for the relayer-sdk to get the ProtocolConfig address from the KMSVerifier contract, it's not the right approach since the KMSVerifier contract can be deployed across multiple chains, and the relayer-sdk needs to fetch any information from the ProtocolConfig **on Ethereum** specifically.

obatirou: What should be done then?

geoxel: I let you decide :-) We can include the ProtocolConfig address in the SDK config if it is logically necessary.

isaacdecoded: > We can include the ProtocolConfig address in the SDK config if it is logically necessary

Yes, I believe the relayer-sdk should point to the ProtocolConfig on Ethereum since it's our only source of truth regarding the protocol configurations at any point, as it's stated in tech specs and RFCs at this time.

But let us know if you have any thoughts, please @obatirou

obatirou: good for me yes I agree

host-contracts/contracts/interfaces/IProtocolConfigMultichain.sol
```diff
@@ -0,0 +1,163 @@
...
+     * @param contextId The mirrored canonical context ID.
+     * @param kmsNodeParams The KMS nodes mirrored from the canonical context, including MPC metadata.
+     * @param thresholds The thresholds mirrored from the canonical context.
+     * @param softwareVersion The KMS software version of the canonical context.
+     * @param pcrValues Accepted enclave PCR values of the canonical context.
+     * @param sourceChainId The chain ID of the canonical `ProtocolConfig`.
+     * @param sourceBlockNumber The block at which the canonical context was read.
+     * @param sourceProtocolConfig The canonical `ProtocolConfig` address.
+     */
+    event MirroredKmsContext(
```
isaacdecoded: nit: it has been a while since we started having some unaligned event names, so considering this nit comment will still leave some other events unaligned, but: maybe consider renaming the events to `MirrorKmsContext`, `MirrorPublicDecryptionThreshold`, etc., to be aligned with other ones like `ActivateEpoch`, `NewKmsContext`, etc.

To be honest, I'm more inclined to use past-tense for events, e.g., `KmsContextMirrored`, but it was already discussed long ago to follow the current naming above.

obatirou: ok will do

host-contracts/contracts/KMSVerifier.sol
```diff
@@ -304,11 +305,13 @@ contract KMSVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain,
...
-            // v1 (0x01 prefix): reads a 32-byte context ID starting at byte 1.
-            // Trailing bytes after byte 33 are ignored for forward-compatibility
-            // with potential v1 extensions.
-            if (extraData.length < 33) {
+        if (version == EXTRA_DATA_V1 || version == EXTRA_DATA_V2) {
+            // v1: [version(1)] [contextId(32)]
+            // v2: [version(1)] [contextId(32)] [epochId(32)]
+            if (
+                (version == EXTRA_DATA_V1 && extraData.length < 33) ||
+                (version == EXTRA_DATA_V2 && extraData.length != 65)
```
isaacdecoded, line 313: Since this method if for extracting the KMS context ID, do we really care if the `extraData.length != 65`? I mean, checking that `extraData.length < 33` seems enough since that's where the context ID is.

obatirou: if we do not validate that it is at least 65, this potentially means this is not a valid EXTRA_DATA_V2?
But this raise the point of `!=` vs `<`: we might want to ignore trailing bytes after byte 65 for forward-compatibility?
So we should use instead `version == EXTRA_DATA_V2 && extraData.length < 65`

host-contracts/contracts/interfaces/IProtocolConfigMultichain.sol
obatirou: Done

host-contracts/contracts/ProtocolConfigMultichain.sol
```diff
@@ -0,0 +1,447 @@
...
+    /// @notice Upper bound on the KMS committee size and on every per-context threshold.
+    /// @dev Driven by the proof format consumed in
+    ///      `KMSVerifier.verifyDecryptionEIP712KMSSignatures`, which encodes the signature count
+    ///      in a single byte (`uint8(decryptionProof[0])`). A context mirrored above this bound
+    ///      cannot ever satisfy verification, so the limit is enforced at mirror time to reject the
+    ///      misconfiguration loudly rather than silently bricking the context.
+    uint256 private constant MAX_KMS_SIGNERS = type(uint8).max;
+
+    /// @custom:storage-location erc7201:fhevm.storage.ProtocolConfig
+    struct ProtocolConfigMultichainStorage {
```
obatirou: With this we will begin divergence of storage between ethereum and other chains. This is ok as this is now a new contract but we are embarking also the fact that polygon is already deployed with ProtocolConfig V1.
Juste a note, not sure we can really something better unfortunately

host-contracts/contracts/KMSVerifier.sol
isaacdecoded: > But this raise the point of != vs <: we might want to ignore trailing bytes after byte 65 for forward-compatibility?

If we want to strictly validate the extra data, I think we should use `v1.length != 33` and `v2.length != 65`. If we want to just validate that the extra data contains the bytes where the context ID is expected, I think `v1/v2 < 33` is enough.

Also, we can consider having a separate `_validateExtraData` method, so the `_extractKmsContextId` remains clearest.

host-contracts/contracts/ProtocolConfigMultichain.sol
isaacdecoded: Good point indeed. It means we'll need to redeploy for Polygon, right?

host-contracts/contracts/KMSVerifier.sol
obatirou: if we do v1.length != 33 and v2.length != 65, forward-compatibility is not retained, that's ok?
good point on _validateExtraData

obatirou: Also do we need to keep v0 case?

host-contracts/contracts/ProtocolConfigMultichain.sol
obatirou: yes we will have to. Also we might have other chains before 0.14 reaching mainnet so maybe additionnal upgrades to.

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -864,33 +884,52 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
...
     /**
-     * @notice Extracts the KMS context ID from request extraData.
+     * @notice Extracts the context ID and epoch ID from the request extraData.
      * @param extraData The stored extra data.
      * @return contextId The extracted context ID, or the current context if extraData is empty.
+     * @return epochId The extracted epoch ID for v2 payloads, or the active epoch for v0/v1 payloads.
      */
-    function _extractContextIdFromExtraData(bytes memory extraData) internal view virtual returns (uint256 contextId) {
-        // v0 (0x00 prefix or empty): uses the current context. Trailing bytes are
+    function _extractContextIdAndEpochIDFromExtraData(
```
obatirou: epochId is not used in any calling site, I would argue this is not necessary to return it and we can simplify the function

host-contracts/contracts/KMSVerifier.sol
isaacdecoded: > if we do v1.length != 33 and v2.length != 65, forward-compatibility is not retained, that's ok?

Not sure if forward compatibility is a thing here: if for any reason, we decide to put more bytes in the extra data v1 or v2, I think the correct way is to define a new v3, meaning that all on-chain and off-chain components must upgrade with this new information. But maybe I'm being too conservative.

> Also do we need to keep v0 case?

Good question. It seems we don't need to since the KMS side is upgrading to execute work based on a given context and epoch. It's not like the decryptions that some outdated clients could continue sending extra data v0.

obatirou: > Not sure if forward compatibility is a thing here: if for any reason, we decide to put more bytes in the extra data v1 or v2, I think the correct way is to define a new v3, meaning that all on-chain and off-chain components must upgrade with this new information. But maybe I'm being too conservative.

Ok I was referring to the comment in the code which mention extension on v0 / v1 but if not necessary let's be conservative and use v1.length != 33 and v2.length != 65

> Good question. It seems we don't need to since the KMS side is upgrading to execute work based on a given context and epoch. It's not like the decryptions that some outdated clients could continue sending extra data v0.

Who should we ping to have confirmation on this?

host-contracts/contracts/KMSGeneration.sol
isaacdecoded: Good catch. Let's keep it simple then. 👍

host-contracts/contracts/KMSVerifier.sol
eudelins-zama: Should we keep v0 for compatibility with `relayer-sdk` v0.4.2?

I have this issue opened on my side: https://github.com/zama-ai/fhevm-internal/issues/1506, let me know if I'm missing something

isaacdecoded: > > Good question. It seems we don't need to since the KMS side is upgrading to execute work based on a given context and epoch. It's not like the decryptions that some outdated clients could continue sending extra data v0.

> Who should we ping to have confirmation on this?

Ah no, I think I'm very wrong, sorry. This is KMSVerifier, so it's public decryption, of course. Yes, we need to continue supporting extra data v0 because of the reason above: outdated clients/relayer sdk sending/validating `extraData = 0x`

EDIT: yes, what Simon said is correct.

obatirou: OK thanks then:
keeping v0 + doing `v1.length != 33 and v2.length != 65`

host-contracts/contracts/KMSGeneration.sol
obatirou: done

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -418,6 +847,131 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        if ($.epochState[latestEpochId] == EpochState.Pending && $.contextForEpoch[latestEpochId] == kmsContextId) {
+            $.epochState[latestEpochId] = EpochState.None;
+            delete $.contextForEpoch[latestEpochId];
+        }
+        delete $.contextCreationPreviousSignerThreshold[kmsContextId];
+        delete $.contextCreationNewSignerConfirmationCount[kmsContextId];
+        delete $.contextCreationPreviousSignerConfirmationCount[kmsContextId];
+    }
+
+    function _getPreviousMaterials()
```
obatirou: I really think we should consider having the kms connector do the retrieval: `NewKmsEpoch` would emit a block number instead and kms connector would fetch previous material info from ethereum pinning that block.
This would allow several things:
* simplify current protocol config by having less code
* remove  ProtocolConfig coupling with KMSGeneration 
* remove the need for `ProtocolConfigMultichain` which could become a mainetance nightmare. We could deploy the same ProtocolConfig implementation on all chains (maybe add a setter for mirror only but that's it). Dao would be responsible to call the correct functions. KMS connector would only listen ethereum anyway.
* KMS connector would only need to do one thing: fetch previous material from KMSGeneration from the block number supply. I am not sure if it needs to do more than that regarding previous material

Unless there is a hard constraint on MPC protocol side?

host-contracts/contracts/KMSVerifier.sol
obatirou: done

host-contracts/contracts/ProtocolConfigMultichain.sol
obatirou: I said yes, but no this is not possible, polygon is deployed with 0.13.0. We will need to upgrade it.
So this adds weights to my comment below https://github.com/zama-ai/fhevm/pull/2615#discussion_r3462478972

host-contracts/contracts/ProtocolConfig.sol
obatirou: Pushed a POC 77b1fdab573cbd574f2a27ddd343d086c1f72285
This greatly simplify things and I am more incline to favor this solution.
The coupling protocolConfig <-> KMSGeneration seems more a flaw in the RFC design unless I miss something at the MPC protocol level

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -90,75 +167,339 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        KmsNodeParams[] calldata kmsNodeParams,
+        KmsThresholds calldata thresholds,
+        string calldata softwareVersion,
+        PcrValues[] calldata pcrValues
     ) external virtual onlyACLOwner {
-        uint256 newContextId = _defineKmsContext(kmsNodes, thresholds);
-        emit NewKmsContext(newContextId, kmsNodes, thresholds);
+        ProtocolConfigStorage storage $ = _getProtocolConfigStorage();
+
+        _requireNoInFlightSwitch();
```
obatirou: I suggest removing it. This is a DAO controlled function and it is its responsability to ensure it is called correctly

host-contracts/contracts/ProtocolConfig.sol
obatirou: @isaacdecoded It seems this is ok for @eudelins-zama.
Will try to clean the POC and advance on this waiting on your input

host-contracts/contracts/ProtocolConfig.sol
obatirou: done with [394981a](https://github.com/zama-ai/fhevm/pull/2615/commits/394981a52de36d8759ae64a29ad01c1da8b23356)

host-contracts/contracts/ProtocolConfig.sol
obatirou: I really prefer the new design with KMS connector doing the retrieval and KMSGeneration decoupling.
Will think about it again today

host-contracts/contracts/ProtocolConfigMultichain.sol
obatirou: resolved with the new design retrieval

host-contracts/contracts/ProtocolConfig.sol
obatirou: As discussed and validated by concerned persons we will keep this new design, closing thread

obatirou:
@claude /pr-review

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -90,75 +168,327 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        KmsNodeParams[] calldata existingKmsNodeParams,
         KmsThresholds calldata existingThresholds
     ) public virtual onlyFromEmptyProxy reinitializer(REINITIALIZER_VERSION) {
         if (existingContextId < KMS_CONTEXT_COUNTER_BASE + 1) {
             revert InvalidKmsContext(existingContextId);
         }
-
         ProtocolConfigStorage storage $ = _getProtocolConfigStorage();
-        // Seed counter so _defineKmsContext's ++counter lands on the original context ID
+        $.epochCounter = EPOCH_COUNTER_BASE;
```
obatirou: it should not begin from scratch but take a in input

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -201,24 +539,97 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+    /// @inheritdoc IProtocolConfig
+    /// @dev Mirror path for non-canonical hosts: imports the canonical context as already active
+    ///      without replaying context-creation confirmations.
+    function mirrorKmsContext(
+        uint256 contextId,
+        KmsNodeParams[] calldata kmsNodeParams,
+        KmsThresholds calldata thresholds,
+        string calldata softwareVersion,
+        PcrValues[] calldata pcrValues
+    ) external virtual onlyACLOwner {
```
isaacdecoded, line 573: I'm wondering if we should add a validation/modifier like `onlyNonEthereum` to check that this method is not called on the canonical contract (Ethereum)? Wdyt @obatirou?

obatirou: I also considered adding guard for canonical / non canonical calls but I think relying on DAO responsability is enough if those function are correctly documented.
This is like the pause modifier we removed imo

host-contracts/contracts/ProtocolConfig.sol
obatirou: done

obatirou:
@claude /pr-review

isaacdecoded (dismissed):
LGTM! 👏

isaacdecoded (dismissed):

Eikix (dismissed):

isaacdecoded:
@mergifyio queue

isaacdecoded (approved):

antoniupop (approved):

isaacdecoded:
@mergifyio queue

PR #2621, opened 2026-05-21
fix(coprocessor): fix a SNS worker race on ciphertext128_format

antoniupop:


coprocessor/fhevm-engine/sns-worker/src/lib.rs
```diff
@@ -250,16 +250,44 @@ impl HandleItem {
...
-            &self.key_id_gw,
-            self.handle,
-            self.transaction_id,
-        )
-        .execute(db_txn.as_mut())
-        .await?;
+        let ct128_format = self.ct128.format();
+
+        if self.ct128.is_empty() {
+            sqlx::query(
```
goshawk-3, line 256: For my benefit, in which cases we can have ct128 empty?

```
/// If inserted into the `ciphertext_digest` table means that the both (ct64 and ct128)
/// ciphertexts are ready to be uploaded to S3.
```

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -111,33 +111,60 @@ async fn run_uploader_loop(
...
+                                    error = %err,
+                                    handle = to_hex(&item.handle),
+                                    "Failed to lock pending uploads",
+                                );
+                                trx.rollback().await?;
+                                continue;
+                            }
+                        };
+
+                        if row.ciphertext.is_some() {
```
goshawk-3, line 144: If I understand it correctly this here means -  if `row.ciphertext.is_some()`  that means the ciphertext was uploaded so we don't need to retry uploading it again. Correct?

Can we add some comments to describe that?

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -111,33 +111,60 @@ async fn run_uploader_loop(
...
+                            item.ct64_compressed = Arc::new(Vec::new());
+                        }
+
+                        if row.ciphertext128.is_some() {
+                            item.ct128 = Arc::new(BigCiphertext::default());
+                        } else if !item.ct128.is_empty() {
+                            item.ct128 = Arc::new(
+                                BigCiphertext::new_with_format_id(
+                                    item.ct128.bytes().to_vec(),
+                                    row.ciphertext128_format,
```
goshawk-3, line 154: Why do we use the `row.ciphertext128_format` instead of the item.ciphertext128_format?

coprocessor/fhevm-engine/sns-worker/src/lib.rs
antoniupop: ct128 can be empty as a placeholder, not as a valid ct128 upload payload
  - before computation: query_sns_tasks initialises ct128 as BigCiphertext::default() until compute_task fills it. If computation is fails it can remain empty
  - if ciphertext128 is already uploaded in ciphertext_digest, ct128 clears to BigCiphertext::default() so it won’t re-upload 
  - in recovery where only ct64 is missing, ct128 can be empty because there is no ct128 work left

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
antoniupop: Indeed - will add

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
antoniupop: Because item may be stale.
item.ct128 is built earlier in fetch_pending_uploads from whatever ciphertext_digest.ciphertext128_format was at that time but the row isn't owned yet. By the time UploadJob::DatabaseLock is processed, another worker may have inserted/updated the correct ciphertext128_format, uploaded ct128, etc.
So after we acquire with FOR UPDATE SKIP LOCKED, the locked DB row becomes the source of truth.

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
antoniupop: done

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
goshawk-3: Noted.

goshawk-3 (approved):

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

PR #2654, opened 2026-05-26
fix(coprocessor): declarative host_chains seeding + worker hardening for chain isolation

Eikix:
## Why

The Polygon Amoy integration on zws-dev failed with a cascading outage: the `host_chains` Postgres table had no row for chainId 80002, so `zkproof-worker` panicked on the first Polygon proof, `spawn_join_set_with_db_retry` restarted it, the same `verified IS NULL` row was re-fetched, and the loop repeated. **Five workers crash-looping on one poison row stopped every chain, not just Polygon — Sepolia went down with it.**

This PR enforces Amina's invariant ("polygon failing should never impact ethereum") and removes the trigger that hit it by making `.Values.chains` the source of truth for both the per-chain listener Deployments (already today) and the `host_chains` DB row (new).

### `refactor(zkproof-worker)` — pre-filter by known chains (Panos's suggestion)
- The worker's SELECT now includes `WHERE chain_id = ANY($1::bigint[])` bound to the cached known chain ids. Workers only fetch rows for chains they can handle.
- A proof that arrives before its chain row has been seeded (race during rollout) stays `verified IS NULL` and gets picked up once the cache reloads — better recovery semantics than permanently marking it failed.
- Trade-off: a row tagged with a `chain_id` that will never be registered stays NULL indefinitely. Observable via `COUNT(*) WHERE verified IS NULL`, rare in practice (proofs only enter via per-chain listeners), and a monitorable anomaly rather than a hot-path failure.

Regression test inserts an unknown-chain row at lower `zk_proof_id` followed by a known-chain row at higher `zk_proof_id`; asserts the known-chain row is processed and the unknown-chain row remains NULL.

### `feat(coprocessor)` — declarative seeding (removes the trigger)
- `initialize_db.sh` now seeds `host_chains` from numbered env vars rendered from `.Values.chains` (`HOST_CHAINS_COUNT` + `HOST_CHAIN_<i>_{ID,NAME,ACL}`). `INSERT ... ON CONFLICT (chain_id) DO NOTHING` makes it idempotent.
- `coprocessor-db-migration.yaml` renders those env vars from `.Values.chains`. Each chain's `*_ACL` uses the same `valueFrom: configMapKeyRef` the listeners use — one source of truth (the deploy-time ConfigMap) for the address.
- `checksum/host-chains` annotations on zkproof-worker / gw-listener / sns-worker / tx-sender pod templates roll those Deployments whenever `.Values.chains` actually changes. `HostChainsCache::load()` is one-shot at startup; without this, an upgrade that adds a chain leaves running pods unaware.
- `insert_test_host_chain.sh` deleted; its logic is now inside `initialize_db.sh`. Removed from both `Dockerfile` and `Dockerfile.workspace`.
- `Chart.yaml` bumped `0.12.0 → 0.13.0`.
- `chore(coprocessor)`: legacy scalar `dbMigration.config.{chainId,aclContractAddress,inputVerifierAddress}` removed — no live consumer remained (the helm Job never invoked the script that read them).

### `test(test-suite)` — production parity
- `generate/env.ts` now writes `HOST_CHAINS_COUNT` + per-chain env vars to `envs["coprocessor"]` BEFORE `buildInstanceEnvs` so multi-copro topology instances inherit them. The dbMigration container seeds the same way prod does.
- Drops the obsolete chained `&& insert_test_host_chain.sh` command from both docker-compose files. `render-compose.test.ts` snapshot updated.
- `registerExtraChainInCoprocessor` is left as a safety net for the re-run-with-new-chain edge case (`coprocessorDbsSeeded` short-circuits dbMigration on subsequent `fhevm-cli up` calls). The INSERT was already idempotent.

## How operators add a chain after this lands

One PR per environment per chain in gitops:

```yaml
# zama-zws/gitops:environments/zws-{dev,testnet}/fhevm-coproc/values.yaml
chains:
  - name: sepolia
    chainId: "11155111"
    aclContractAddressValueFrom:
      configMapKeyRef: { name: eth-sc-addresses,     key: acl.address }
    # ...existing listener fields...
  - name: polygonAmoy
    chainId: "80002"
    aclContractAddressValueFrom:
      configMapKeyRef: { name: polygon-sc-addresses, key: acl.address }
    # ...existing listener fields...
```

`helm upgrade` re-runs the dbMigration Job, idempotently inserts new rows, rolls the chain-agnostic Deployments via the checksum annotation, and the per-chain listener Deployments fan out from the same list (already does today).

## Strategy

- **This PR (forward port to main):** confirm via CI + `e2e orchestrate test` label, iterate on review.
- **Backport to `release/0.13.x`:** opened once this is green, after sanity-checking that main hasn't diverged structurally on the touched files.
- **Gitops PR (`zama-zws/gitops`):** opened last, after the chart version is published.

## Out of scope (separate work)
- Updating an ACL on an existing chain (`ON CONFLICT DO NOTHING` is intentional — rotation needs an explicit gated runbook).
- Deleting a chain (no `DELETE` path; `chain_id` rows are preserved).
- On-chain event-driven discovery via `gw-listener` watching `GatewayConfig.addHostChains` (RFC for 0.14.x).
- QuickNode rate-limit split and eRPC crashloop on gw-listener (infra, separate workstreams — both resolved by infra in parallel).
- Prometheus counter for `unknown_chain_skipped_total` — easy follow-up if alerting on NULL buildup is needed.

## Tested locally
- `cargo check -p zkproof-worker --tests`
- `helm template charts/coprocessor` with a 2-chain values file (renders `HOST_CHAINS_COUNT`, three env vars per chain, four `checksum/host-chains` annotations).
- `helm lint charts/coprocessor`.

CI label `e2e orchestrate test` will exercise the seeder + filter end-to-end.


coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
```diff
@@ -267,28 +268,58 @@ async fn execute_verify_proof_routine(
...
+        .map(|c| c.chain_id.as_i64())
+        .collect();
+
     let mut txn: sqlx::Transaction<'_, sqlx::Postgres> = pool.begin().await?;
-    if let Ok(row) = sqlx::query!(
+    if let Ok(row) = sqlx::query(
         "SELECT zk_proof_id, input, chain_id, contract_address, user_address, transaction_id
             FROM verify_proofs
             WHERE verified IS NULL
+              AND chain_id = ANY($1::bigint[])
```
PanGan21, line 301: Should we do the same here? https://github.com/zama-ai/fhevm/blob/4647474e517670cf03eb2da1e46e32bdf444cbc5/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs#L626
I think it is related

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
```diff
@@ -267,28 +268,58 @@ async fn execute_verify_proof_routine(
...
+    // rows.
+    let known_chain_ids: Vec<i64> = host_chain_cache
+        .all()
+        .iter()
+        .map(|c| c.chain_id.as_i64())
+        .collect();
+
     let mut txn: sqlx::Transaction<'_, sqlx::Postgres> = pool.begin().await?;
-    if let Ok(row) = sqlx::query!(
+    if let Ok(row) = sqlx::query(
```
PanGan21: Ideally we can keep `query!` for compile time sql support

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
Eikix: Good catch — done in a8a2b82c2.  now takes the same `known_chain_ids` slice and applies `AND chain_id = ANY($1::bigint[])` to the COUNT. The reported count now reflects what workers actually pick up; unregistered-chain rows are excluded from both the SELECT and the count, so the two stay in lockstep.

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
Eikix: Done in a8a2b82c2. Switched to `sqlx::query!` and regenerated the `.sqlx/` cache via `cargo sqlx prepare --workspace -- --all-targets` against a local Postgres seeded by `initialize_db.sh`. The cache diff is just one renamed file per query (old hash → new hash). Compile-time SQL check preserved.

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
Eikix: sorry for this slop response 😭
claude has rebelled

PanGan21 (dismissed):

PanGan21 (dismissed):

tawadaa (approved):

Eikix:
@mergifyio queue

PanGan21 (approved):

PR #2664, opened 2026-05-27
feat(gateway-contracts): add priority coprocessor finalization

Eikix:
## Summary

- Add a GatewayConfig priority coprocessor transaction sender flag.
- In priority mode, ciphertext material and input verification/rejection consensus finalize only when the configured sender responds.
- Keep emitted consensus payloads and consensus view helpers singleton-scoped to the priority sender for downstream relayer / KMS connector compatibility.
- Preserve normal threshold-based consensus when the priority sender is removed.
- Add an idempotent `priority-coprocessor` e2e profile that temporarily enables priority mode, runs the existing input + compute + decrypt flow, then restores the prior priority value.

## Non-goals

- No relayer, relayer-sdk, KMS connector, host contract, or SDK behavior changes in this PR.
- No weighted consensus implementation.

## Validation

- `DOTENV_CONFIG_PATH=.env.example npx hardhat test test/GatewayConfig.ts test/CiphertextCommits.ts test/InputVerification.ts --grep "priority|Priority"`
- `PATH=$HOME/.foundry/versions/v1.3.1:$PATH make check-bindings`
- `PATH=$HOME/.foundry/versions/v1.3.1:$PATH make check-selectors`
- `make check-mocks`
- `git diff --check`

Full e2e execution of `priority-coprocessor` still requires a running fhevm stack.


Eikix:
Also Ive decide to close to PR to main as it's too early and it's painful to maintain both.
Real PR for comments is https://github.com/zama-ai/fhevm/pull/2665

PR #2665, opened 2026-05-27
feat(gateway-contracts): add priority coprocessor finalization

Eikix:
## Summary

- Backport of #2664 onto `release/0.13.x`.
- Add a GatewayConfig priority coprocessor transaction sender flag.
- In priority mode, ciphertext material and input verification/rejection consensus finalize only when the configured sender responds.
- Keep emitted consensus payloads and consensus view helpers singleton-scoped to the priority sender for downstream relayer / KMS connector compatibility.
- Preserve normal threshold-based consensus when the priority sender is removed.
- Add an idempotent `priority-coprocessor` e2e profile that temporarily enables priority mode, runs the existing input + compute + decrypt flow, then restores the prior priority value.

## Non-goals

- No relayer, relayer-sdk, KMS connector, host contract, or SDK behavior changes in this PR.
- No weighted consensus implementation.

## Validation

- `DOTENV_CONFIG_PATH=.env.example npx hardhat test test/GatewayConfig.ts test/CiphertextCommits.ts test/InputVerification.ts --grep "priority|Priority"`
- `PATH=$HOME/.foundry/versions/v1.3.1:$PATH make check-bindings`
- `PATH=$HOME/.foundry/versions/v1.3.1:$PATH make check-selectors`
- `make check-mocks`
- `git diff --check`

Full e2e execution of `priority-coprocessor` still requires a running fhevm stack.
closes https://github.com/zama-ai/fhevm-internal/issues/1487


gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -264,10 +264,20 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
      * @notice Re-initializes the contract from V6.
+     * @param initialPriorityCoprocessorTxSender The registered priority coprocessor transaction sender to set,
+     *        or zero to leave priority mode disabled.
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    function reinitializeV7() public virtual reinitializer(REINITIALIZER_VERSION) {}
+    function reinitializeV7(
+        address initialPriorityCoprocessorTxSender
+    ) public virtual onlyOwner reinitializer(REINITIALIZER_VERSION) {
```
isaacdecoded: I think we should bump the `REINITIALIZER_VERSION` since QA is already testing the last version from `v0.13.0-3`. Also note that [I missed to bump the MINOR_VERSION](https://github.com/zama-ai/fhevm/blame/v0.13.0-3/gateway-contracts/contracts/GatewayConfig.sol#L46) during the last added hotfixes.

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -345,6 +355,13 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         uint256 newCoprocessorThreshold
     ) external virtual onlyOwner whenInputVerificationPaused {
         GatewayConfigStorage storage $ = _getGatewayConfigStorage();
+        address priorityCoprocessorTxSender = $.priorityCoprocessorTxSender;
+        if (
+            priorityCoprocessorTxSender != address(0) &&
+            !_containsCoprocessorTxSender(newCoprocessors, priorityCoprocessorTxSender)
+        ) {
+            revert PriorityCoprocessorTxSenderNotRegistered(priorityCoprocessorTxSender);
```
isaacdecoded:
```suggestion
            revert PriorityCoprocessorNotInNewCoprocessors(priorityCoprocessorTxSender);
```

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -1115,6 +1157,51 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         $.coprocessorThreshold = newCoprocessorThreshold;
     }
 
+    /**
+     * @notice Sets the priority coprocessor transaction sender.
+     * @param coprocessorTxSenderAddress The priority coprocessor transaction sender, or zero to disable priority mode.
```
isaacdecoded, line 1180: If we already have a `removePriorityCoprocessorTxSender` for disabling, shouldn't we enforce a valid coprocessor tx address for updating? If not, it seems we can then get rid of the remove method.

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -1115,6 +1157,51 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
+     */
+    function _updatePriorityCoprocessorTxSender(address coprocessorTxSenderAddress) internal virtual {
+        GatewayConfigStorage storage $ = _getGatewayConfigStorage();
+        $.priorityCoprocessorTxSender = coprocessorTxSenderAddress;
```
isaacdecoded, line 1184: Since the priority coprocessor will be a single one, maybe we should also validate that the current coprocessor threshold is 1, to avoid the protocol getting stuck by mistake.

gateway-contracts/contracts/GatewayConfig.sol
Eikix: I think we agreed that we only bump reinitializer once per release, 
And that only Testnet matters in that regard
Which would suggest devnet can always redeploy

But we can re-align with QA

gateway-contracts/contracts/GatewayConfig.sol
Eikix: priority mode ignores threshold so it shouldnt be required

gateway-contracts/contracts/GatewayConfig.sol
Eikix: setPriorityCoprocessorTxSender() only accepts an already registered coprocessor tx-sender
i'll clarify comment

isaacdecoded: Sorry, I meant a non-zero coprocessor tx address

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: Ah yes, you're right. Thanks.

gateway-contracts/contracts/GatewayConfig.sol
Eikix: but this is the internal entrypoint

removePriorityCoprocessorTxSender just calls it with address(0)

isaacdecoded: I see, it's clear to me now. Thanks for clarifying it to me 👍

isaacdecoded (approved):
LGTM!

Eikix:
Reconverting to draft as we will merge this PR as part of 0.13.1 bump afaik.

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -263,11 +263,22 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
+     * @notice Re-initializes the contract from V7.
+     * @dev If a priority coprocessor is set here, every host-chain `InputVerifier` must accept its signer
+     *      with threshold 1 before user inputs rely on priority mode.
+     * @param coprocessorTxSenderAddress The registered priority coprocessor transaction sender to set,
+     *        or zero to leave priority mode disabled.
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    function reinitializeV7() public virtual reinitializer(REINITIALIZER_VERSION) {}
+    function reinitializeV8(address coprocessorTxSenderAddress) public virtual onlyOwner reinitializer(REINITIALIZER_VERSION) {
```
obatirou:
```suggestion
    function reinitializeV8(address coprocessorTxSenderAddress) public virtual reinitializer(REINITIALIZER_VERSION) {
```

test-suite/e2e/test/userInput/priorityCoprocessorInputFlow.ts
```diff
@@ -0,0 +1,231 @@
...
+      await restorePriorityTx.wait();
+    }
+  } finally {
+    if (!snapshot.gatewayWasPaused) {
+      await unpauseGatewayInputVerification(gatewayInputVerification);
+    }
+  }
+};
+
+const runAdd42InputAndDecrypt = async function (this: Mocha.Context) {
```
obatirou: already defined in test-suite/e2e/test/userInput/inputFlow.ts?

Eikix:
thanks for review

Eikix:
Looking for a second round of review so that I can reduce the backlog of open PRs targetting feature/0.13.1 @obatirou when you have time

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -343,8 +346,26 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
     function updateCoprocessors(
         Coprocessor[] calldata newCoprocessors,
         uint256 newCoprocessorThreshold
-    ) external virtual onlyOwner whenInputVerificationPaused {
```
obatirou, line 339: can you update the natspec on methods where the whenInputVerificationPaused to have context?

gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -318,6 +324,21 @@ contract CiphertextCommits is ICiphertextCommits, UUPSUpgradeableEmptyProxy, Gat
...
+    function _canFinalizeCiphertextMaterial(
+        address coprocessorTxSender,
+        uint256 coprocessorCounter
+    ) internal view virtual returns (bool canFinalize, bool finalizedByPriority) {
+        address priorityCoprocessorTxSender = GATEWAY_CONFIG.getPriorityCoprocessorTxSender();
+        if (priorityCoprocessorTxSender != address(0)) {
+            finalizedByPriority = coprocessorTxSender == priorityCoprocessorTxSender;
+            return (finalizedByPriority, finalizedByPriority);
+        }
+        return (_isConsensusReached(coprocessorCounter), false);
```
obatirou: nit: the body of  `_canFinalizeCiphertextMaterial` and `_canFinalizeCoprocessorConsensus` are equal. 
You could make a shared helper in `GatewayConfigChecks` with a rename (but more change as `_isConsensusReached` is also duplicated) or inline `_canFinalizeCiphertextMaterial` as it is only used once in `CiphertextCommits.sol`.
As we might want to limit the changes included + depreciation of gateway at somepoint, I would lean towards the second option. Maybe gateway team has more inputs.

gateway-contracts/contracts/InputVerification.sol
```diff
@@ -370,6 +403,14 @@ contract InputVerification is
...
         InputVerificationStorage storage $ = _getInputVerificationStorage();
 
+        address priorityConsensusTxSender = $.priorityRejectProofConsensusTxSender[zkProofId];
+        if (priorityConsensusTxSender != address(0)) {
+            // Keep the getter aligned with the singleton consensus result exposed in priority mode.
+            address[] memory txSenders = new address[](1);
+            txSenders[0] = priorityConsensusTxSender;
+            return txSenders;
+        }
+
```
obatirou, line 413: Could be simplified by add an equivalent `_getAddCiphertextMaterialConsensusTxSenders` helper ?

Eikix: i think i'tsa bit overkill otherwise we have to do:

```
function _getVerifyProofConsensusTxSenders(
    InputVerificationStorage storage $,
    uint256 zkProofId
) internal view virtual returns (address[] memory) {
    address priorityConsensusTxSender = $.priorityVerifyProofConsensusTxSender[zkProofId];
    if (priorityConsensusTxSender != address(0)) {
        address[] memory txSenders = new address[](1);
        txSenders[0] = priorityConsensusTxSender;
        return txSenders;
    }

    bytes32 consensusDigest = $.verifyProofConsensusDigest[zkProofId];
    return $.verifyProofConsensusTxSenders[zkProofId][consensusDigest];
}

function _getRejectProofConsensusTxSenders(
    InputVerificationStorage storage $,
    uint256 zkProofId
) internal view virtual returns (address[] memory) {
    address priorityConsensusTxSender = $.priorityRejectProofConsensusTxSender[zkProofId];
    if (priorityConsensusTxSender != address(0)) {
        address[] memory txSenders = new address[](1);
        txSenders[0] = priorityConsensusTxSender;
        return txSenders;
    }

    return $.rejectProofConsensusTxSenders[zkProofId];
}
```

Eikix:
[AUTOMATED]

Live priority-mode toggles should be covered in the rollout runbook.

Because these endpoints are DAO-executed and cannot rely on pausing/draining, the proposal procedure should account for in-flight requests. Enabling priority can strand a proof/material item when the future priority sender already responded before the flip; removing priority can strand an item when enough responses were collected while priority was active but no later response re-evaluates threshold consensus.

The runbook should define the accepted mitigation, such as an off-chain quiet-period/pending-request check before execution, or explicitly accepted bounded in-flight loss during the rollout.

isaacdecoded (approved):
LGTM! I guess we should be sure that this PR closes [this other issue](https://github.com/zama-ai/fhevm-internal/issues/1487).

PR #2681, opened 2026-05-28
test(sdk): code refactoring for fheTest cleartext/ciphertext

geoxel:
The four fheTest directories (`viem/`, `viem-cleartext/`, `ethers/`,
`ethers-cleartext/`) each held 10 test files that were nearly identical —
differing only in the client factory (`createFhevmXxx` vs
`createFhevmCleartextXxx`) and the `runIf` gate (`isCleartext` vs
`!isCleartext`). Any change to test logic had to be applied in four places.

This commit eliminates that duplication by introducing two shared directories:

- `viem-common/` — 10 `define*Tests(runIf, options)` factory functions for the
  viem adapter
- `ethers-common/` — same for the ethers adapter, using the ethers contract API
  (`fheTest.getHandleOf!(...)`, `tx.wait()`, switch-based dispatch) instead of
  the viem equivalents

Each of the 40 original test files is replaced by a 5–9 line thin wrapper that
imports the factory and calls it with the appropriate `runIf` condition and
client factory. The `cleartext-ethers/` and `cleartext-viem/` directories are
also renamed to `ethers-cleartext/` and `viem-cleartext/` for consistency with
the naming convention of their ciphertext counterparts.

Behavioral changes: none — same test suites, same assertions, same `runIf`
gates.


0xalexbel (commented):
LGTM!

0xalexbel (approved):
LGTM!

PR #2689, opened 2026-05-29
feat(charts/contracts): scJobs list with per-job rendering

piizama:
## Summary

The contracts chart used to template a single deploy/upgrade Job whose name
was fixed per chart release. Bumping the image tag could not be done without
manually deleting the prior Job, and one-off operations (add-pauser, ownership
transfer, keygen, …) had no clean place to live.

This refactor makes `scJobs:` the only way to declare work: each entry renders
as its own uniquely-named Job (`<release>-<kind>-<name>-<tagSlug>`), so
re-installs with a new tag yield new Jobs side-by-side with completed ones
instead of collisions. The legacy `scDeploy` / `scUpgrade` blocks are removed
entirely, hence the 1.0.0 chart version bump (breaking).

## Changes

- New `scJobs:` schema with `kind: deploy | upgrade | task`, an `enabled`
  flag, and an optional `envFile` (path/glob of address files to record,
  default `addresses/.env.*`) so jobs sharing a PVC never re-stamp another
  contract family's `.version` keys. One ConfigMap holds per-job runner
  scripts.
- `configmap`, `resources` and `securityContext` are now top-level values
  shared by all Job containers and the debug pod.
- Addresses ConfigMap stores `<contract>.address` (immutable, skip if set)
  AND `<contract>.version=<image.tag>` per Job. The global `contracts.version`
  key is dropped, since contracts can legitimately run at different versions.
- `scDebug` pod reuses the image/env/oldImage/persistence of the last enabled
  `scJobs` entry (rendering fails with a clear message if none), with
  `mountMode: readOnly | readWrite` (default `readOnly`).
- All Jobs annotated with `argocd.argoproj.io/sync-options: Prune=false` so
  removing an `scJobs` entry preserves the historical Job in-cluster as a
  re-run sentinel, plus `argocd.argoproj.io/sync-wave: <index>` for
  sequential execution in `scJobs` order.
- `verifyContracts` / `hardhat verify:verify` removed; verification should run
  as a dedicated `task` job.
- Chart README added.

## Breaking changes

- `scDeploy` and `scUpgrade` values are no longer honored; migrate to
  `scJobs:` entries. `scDeploy.configmap` moves to the top-level `configmap`,
  `scDeploy.resources`/`securityContext` move to top level.
- `verifyContracts` is gone.

The `values-*.yaml` reference files are intentionally left untouched.

### DIFF
--- a/charts/contracts/templates/sc-deploy-statefulset.yaml
+++ b/charts/contracts/templates/sc-deploy-statefulset.yaml
@@ -1,4 +1,8 @@
 {{- if .Values.scDebug.enabled }}
+{{- $mountMode := .Values.scDebug.mountMode | default "readOnly" }}
+{{- if not (has $mountMode (list "readOnly" "readWrite")) }}
+{{- fail (printf "scDebug.mountMode must be 'readOnly' or 'readWrite', got %q" $mountMode) }}
+{{- end }}
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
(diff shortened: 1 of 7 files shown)


tawadaa (dismissed):

tawadaa (approved):

charts/contracts/templates/sc-deploy-statefulset.yaml
```diff
@@ -1,4 +1,8 @@
 {{- if .Values.scDebug.enabled }}
+{{- $mountMode := .Values.scDebug.mountMode | default "readOnly" }}
```
tawadaa, line 2: Need an update on scDebug env, as it still uses scDeploy.env, so all scDeploy reference must be checked

tawadaa (changes requested):

charts/contracts/templates/sc-deploy-statefulset.yaml
piizama: Fixed

PR #2694, opened 2026-06-01
fix(relayer): stop retrying ACL reverts on gas estimation

manoranjith:
Removes `PublicDecryptNotAllowed` and `AccountNotAllowedToUseCiphertext`
from the retryable gas-estimation reverts: the ACL check is now done
directly on the host chain, not during gas estimation, so only
`CiphertextMaterialNotFound` remains retriable.

See [zama-ai/fhevm-internal#1474][issue] for details.

Follow-up (out of scope): the `RevertedACLSelector` metric now fires
only for `CiphertextMaterialNotFound`; renaming is deferred.

[issue]: https://github.com/zama-ai/fhevm-internal/issues/1474

### DIFF
--- a/.github/workflows/relayer-tests.yaml
+++ b/.github/workflows/relayer-tests.yaml
@@ -173,7 +173,10 @@ jobs:
         run: |
           if ! command -v sqlx &> /dev/null; then
             echo "Installing sqlx-cli..."
-            cargo install sqlx-cli --no-default-features --features postgres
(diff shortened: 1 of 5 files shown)


zmalatrax (dismissed):
lgtm ty!

zmalatrax (approved):

manoranjith:
@mergify queue

PR #2695, opened 2026-06-01
fix(test-suite): sync drift recovery on revert signal

Eikix:
## Summary

Synchronizes `ciphertext-drift-auto-recovery` on the actual `drift_revert_signal` row in the faulty coprocessor DB before starting recovery behavior assertions.

The peer-submission drift warning still proves divergent submissions were observed, but it is no longer treated as the auto-revert start signal.

## Validation

- `bun test src/cli.test.ts`
- `bun test src`
- `bun run check`
- `git diff --check`


PanGan21 (approved):

Eikix:
@mergifyio queue

PR #2697, opened 2026-06-01
fix(coprocessor): host-listener-consumer, chains promotion only occur…

rudy-6-4:
…s at startup or after drift

chains promotion was done for each block instead of at startup only
the after drift is for consistency with other listeners


rudy-6-4:
@claude /pr-review

rudy-6-4:
see https://github.com/zama-ai/fhevm/pull/2848/changes#diff-d02f8e282c0199fdc6c0ff3047d5213671d8777e7bf221a3bea70f167d463313R37

PR #2698, opened 2026-06-01
feat(relayer): pre-check v3 user-decryption sigs

manoranjith:
## Summary

Signature pre-check on the v3 user-decryption endpoint. Recomputes
the unified EIP-712 digest and runs the shared verifier before
forwarding, so a bad signature gets a 400 at the relayer instead of
a gateway/KMS round-trip. Stacked on the unified flow (#2544); base
is `mano/relayer/unified-eip-712`, not `main`.

The KMS Connector remains the authoritative check.

Refs zama-ai/fhevm-internal#1301.

## Design decisions

- ERC-1271 gas cap is configurable; the default is set to 100,000
  (same as the KMS Connector).
- Transport errors mirror the host ACL checks: retried with the same
  `host_acl_check.retry` policy, then surfaced as 500 — not
  failed-open. Only definitive failures (ecrecover mismatch,
  wrong/empty ERC-1271 magic, revert, short returndata) yield 400.
- Domain name/version come from the shared `default_user_decrypt_domain`
  (`Decryption`/`1`), matching the gateway `Decryption` contract's
  `__EIP712_init` and the domain the KMS Connector verifies against.
- Handles spanning multiple host chains are rejected — one EIP-712
  domain carries one `contractsChainId`, so there's no single digest
  to verify against.

## Implementation notes

- Verification is not reimplemented: it reuses the shared
  `user-decryption-signature` crate (the same library the KMS
  Connector calls). The one addition,
  `compute_user_decrypt_digest_from_parts`, lets the relayer hash
  from its own request model rather than building the gateway
  binding's `UserDecryptionRequestPayload`; a parity test pins it to
  the existing payload-based digest.
- The relayer pulls `fhevm_gateway_bindings` from a git rev while the
  shared crate uses the in-tree path bindings. Only alloy primitives
  cross the crate boundary, so the two binding copies never need to
  be the same type.
- Test utils gain a signer + an envelope-signing helper, so the v3
  endpoint tests submit real EOA signatures and exercise the
  pre-check end to end.

## Tests

`cargo test --lib` green (195) incl. 10 `host::signature_prechecker`
unit tests; shared crate 11/11. `--features integration-tests`:
`user_decrypt_v3_test` (19) covers the happy paths with real
signatures plus `v3_rejects_invalid_signature` (bad sig → 400).

## Pending

- [ ] Pin `fhevm_gateway_bindings` to the final `feat/RFC-012-016`
  merge commit before merge.


shared/user-decryption-signature/src/lib.rs
```diff
@@ -94,6 +94,30 @@ pub fn compute_user_decrypt_digest(
...
+    UserDecryptRequestVerification {
+        userAddress: user_address,
+        publicKey: public_key.clone(),
+        allowedContracts: allowed_contracts.to_vec(),
+        startTimestamp: start_timestamp,
+        durationSeconds: duration_seconds,
+        extraData: extra_data.clone(),
+    }
+    .eip712_signing_hash(domain)
+}
```
eudelins-zama: Nit: not a huge fan of function with so much args (ignoring the clippy lint), but if more convenient for you than implementing `From<...> for UserDecryptRequestVerification` traits in Relayer code and/or cleaner interface, fine for me

manoranjith: @eudelins-zama I agree. I updated the relayer as well to use the type from bindings and rebased the PR to latest main, as there were some conflicts. Could you review it once more ?

eudelins-zama: Looks better to me, thanks!

eudelins-zama (dismissed):
LGTM on the signature verification side!

Letting someone else approve for the rest of the Relayer's code

manoranjith:
@eudelins-zama Thanks for your review for signature part.

Will wait for review from @zmalatrax on the relayer part.

relayer/src/config/settings.rs
```diff
@@ -542,6 +542,12 @@ pub struct ContractConfig {
     pub user_decrypt_shares_threshold: u32,
 }
 
+/// User-decryption signature check configuration.
+#[derive(Debug, Deserialize, Clone)]
+pub struct UserDecryptSignatureCheckConfig {
+    pub erc1271_gas_limit: u64,
```
zmalatrax, line 548: I'd add a default value so that it becomes optional in the config as in the kms connector config: https://github.com/zama-ai/fhevm/blob/88b11de0d67a096ef5e0659eb8106bb691acf763/kms-connector/crates/kms-worker/src/core/config.rs#L75-L78

relayer/Makefile
```diff
@@ -463,13 +463,15 @@ clippy-fix: ## Fix code issues with clippy
 	@cargo clippy --all-targets --all-features --fix --allow-dirty --allow-staged
 	@cargo clippy --all-targets --all-features --fix --allow-dirty --allow-staged --manifest-path $(MOCK_RPC_MANIFEST)
 
+# `--all` is intentionally omitted: for this single-package crate it would descend into path
+# dependencies (e.g. the in-tree generated `rust_bindings`), which are not ours to format.
 fmt: ## Check code formatting
-	@cargo fmt --all -- --check
-	@cargo fmt --all --manifest-path $(MOCK_RPC_MANIFEST) -- --check
+	@cargo fmt -- --check
```
zmalatrax:
```suggestion
	@cargo fmt --package fhevm-relayer -- --check
```

relayer/Makefile
```diff
@@ -463,13 +463,15 @@ clippy-fix: ## Fix code issues with clippy
 	@cargo clippy --all-targets --all-features --fix --allow-dirty --allow-staged
 	@cargo clippy --all-targets --all-features --fix --allow-dirty --allow-staged --manifest-path $(MOCK_RPC_MANIFEST)
 
+# `--all` is intentionally omitted: for this single-package crate it would descend into path
+# dependencies (e.g. the in-tree generated `rust_bindings`), which are not ours to format.
```
zmalatrax: It reads strangely to me

```suggestion
# Scope fmt to owned packages; generated path dependencies are excluded.
```

zmalatrax (changes requested):
The v3 user decrypt http contract currently requires a signature of length 4 minimum:
https://github.com/zama-ai/fhevm/blob/65c98c2ca75fbd6b8382557c76148123f166031c/relayer/src/http/endpoints/v3/types/user_decrypt.rs#L36-L45

It'll block the check `isValidSignature(digest, "")` described in the RFC-012 algorithm, used for Safe wallets
https://github.com/zama-ai/tech-spec/blob/main/rfcs/012-erc-1271-signature-verification.md#verification-flow

shared `verify_signature` function: https://github.com/zama-ai/fhevm/blob/65c98c2ca75fbd6b8382557c76148123f166031c/shared/user-decryption-signature/src/lib.rs#L180-L182

relayer/src/config/settings.rs
manoranjith: In the relayer we don't set config defaults by convention — every value is defined explicitly. This keeps each environment (local, devnet, mainnet, etc.)  configured with the right value instead of silently falling back to a default, which matters even more as we move to multi-chain.

  The per-environment values are tracked in dedicated config files in the repo.

relayer/Makefile
manoranjith: Practically its the same, but I like being explicit in config, so I'd take it.  

Thanks for attention to small details.

relayer/Makefile
manoranjith: I agree, I guess by oversight I left the verbose comment from Claude.

I like the more concise one. Thanks for attention to detail again :pray:

manoranjith:
> The v3 user decrypt http contract currently requires a signature of length 4 minimum:
> 
> https://github.com/zama-ai/fhevm/blob/65c98c2ca75fbd6b8382557c76148123f166031c/relayer/src/http/endpoints/v3/types/user_decrypt.rs#L36-L45
> 
> It'll block the check `isValidSignature(digest, "")` described in the RFC-012 algorithm, used for Safe wallets https://github.com/zama-ai/tech-spec/blob/main/rfcs/012-erc-1271-signature-verification.md#verification-flow
> 
> shared `verify_signature` function:
> 
> https://github.com/zama-ai/fhevm/blob/65c98c2ca75fbd6b8382557c76148123f166031c/shared/user-decryption-signature/src/lib.rs#L180-L182

Thanks for the catch. This is an error, I am fixing it in a follow up commit.

zmalatrax (dismissed):
lgtm

manoranjith:
@mergify queue

manoranjith:
@mergify queue

manoranjith:
@mergify queue

zmalatrax (dismissed):

manoranjith:
@mergify queue

zmalatrax (approved):

manoranjith:
@mergify queue

PR #2700, opened 2026-06-01
feat(kms-connector): implement epoch changes and key resharing

eudelins-zama:
## Context-switch event handling in kms-connector

Adds end-to-end handling of protocol-config / context-switch events through the kms-connector pipeline.

### Main changes
- **Data model**: new protocol-config event types and DB migrations (shared `utils` types).
- **gw-listener**: listens for context-switch events on the gateway and publishes them.
- **kms-worker**: processes protocol-config events (new `protocol_config` event processor).
- **tx-sender**: forwards the resulting context-switch responses on-chain.
- **Supporting**: regenerated sqlx query cache + bindings, tests with a new `ProtocolConfigMock`, and updated helm charts / staging env.

Commits are split by pipeline stage (bindings → types → gw-listener → kms-worker → tx-sender → tests → deploy) to read top-to-bottom.

### Notable design decisions

- **Two-chain polling**: block tracking moved from per-event-type (`last_block_polled`) to per-chain (`last_block_polled_by_chain`, `ChainName::{Ethereum,Gateway}`), as we are using a single `eth_getLogs` call for events coming from a single chain.
- **Previous-context is re-fetched, not trusted**: on `NewKmsContext`, the worker resolves the previous context via `getKmsContextAnchor` → log filter at the anchor block, then verifies it against the on-chain `contextInfoHash`.
- **Nested arrays as `BYTEA`**: nested Solidity arrays (node params, PCR values, keys, CRS list) are stored ABI-encoded and decoded losslessly on read, rather than in relational columns.

### Issues

Closes https://github.com/zama-ai/fhevm-internal/issues/1116
Closes https://github.com/zama-ai/fhevm-internal/issues/1117
Closes https://github.com/zama-ai/fhevm-internal/issues/1118
Closes https://github.com/zama-ai/fhevm-internal/issues/1341
Closes https://github.com/zama-ai/fhevm-internal/issues/571
Closes https://github.com/zama-ai/fhevm-internal/issues/1505
Closes https://github.com/zama-ai/fhevm-internal/issues/1548


kms-connector/.sqlx/query-f486e4a3be5017b2bad9e05f6ee463794f85f7378545727fd3ed590e71fb2e20.json
```diff
@@ -0,0 +1,19 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "INSERT INTO epoch_result_responses(context_id, epoch_id, keys, crs_list, created_at, otlp_context) VALUES ($1, $2, $3, $4, $5, $6) ON CONFLICT DO NOTHING",
```
dvdplm: If there's a conflict we assume some other worker or thread already inserted the data and it's all good?

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -60,13 +69,14 @@ where
...
-        let kms_verifier_contract = KMSVerifier::new(config.kms_verifier_address, provider.clone());
+        let protocol_config_contract =
+            ProtocolConfig::new(config.protocol_config_contract.address, provider.clone());
         Self {
             db_pool,
             provider,
             config: config.clone(),
             cancel_token,
-            kms_verifier_contract,
+            protocol_config_contract,
```
dvdplm, line 86: Maybe it'll become clearer when I read on, but I wonder why the protocol_config_contract gets "special treatment" like this and is its own member. Other contract metadata is kept inside the `config` member – couldn't the protocol config live there too?

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -83,34 +93,36 @@ where
...
+    /// Stores the active KMS context and epoch found on-chain in the database.
+    ///
+    /// TODO: better init/update of this table (https://github.com/zama-ai/fhevm-internal/issues/1467)
     pub async fn store_on_chain_context(&self) -> anyhow::Result<()> {
-        let current_context_id = self
-            .kms_verifier_contract
-            .getCurrentKmsContextId()
+        let active = self
+            .protocol_config_contract
+            .getCurrentKmsContextAndEpoch()
```
dvdplm, line 110: naming nitpick: I think that we sometimes overdo it a bit when trying to name things in a readable way. Somtimes it's justified but sometimes it gets a bit over the top. In our heads I think "context" is somehow distinct from "epoch" so it becomes natural to name this `ContextAndEpoch` but in reality the word "context" encompasses all kinds of "surrounding circumstances that define the situation we're in". All of that to say that if this was named just `let active = self.kmsContext().call().await?;` the rest of the code would read just fine.

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -293,52 +319,58 @@ async fn publish_crsgen_request<'e>(
...
         );
     }
 
     Ok(())
 }
 
-pub async fn publish_context_id(db_pool: &Pool<Postgres>, context_id: U256) -> anyhow::Result<()> {
-    info!("Publishing KMS context #{context_id} in DB...");
+/// Persists the active `(context_id, epoch_id)` pair fetched at startup via
+/// `KMSVerifier::getActiveKmsContextAndEpoch()` (RFC 005). Inserted into `kms_context` with the
```
dvdplm: It's `EthereumListener` now I think?

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -83,34 +93,36 @@ where
...
             .call()
             .await?;
 
-        publish_context_id(&self.db_pool, current_context_id).await
+        publish_active_context_and_epoch(&self.db_pool, active.contextId, active.epochId).await
```
dvdplm: Same theme here: if this was `publish_context(…)` it would be readable enough imo.

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -83,34 +93,36 @@ where
...
     }
 
-    /// Polling loop to listen to [`KMSGeneration`] events on Ethereum.
+    /// Polling loop to listen to [`KMSGeneration`] and [`ProtocolConfig`] events on Ethereum.
     async fn run_poll_loop(&self) -> anyhow::Result<()> {
-        let event_signatures = KMS_GENERATION_EVENT_TYPES
+        let event_signatures = ETHEREUM_EVENT_TYPES
             .iter()
             .map(|e| e.signature_hash())
             .collect::<Vec<_>>();
```
dvdplm: This allocation is annoying. AFAICT there's no good reason why `event_signature` couldn't take a slice of `FixedBytes<32>`. I don't think there's anything we can do about it.

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -130,12 +142,12 @@ where
                 }
                 Err(e) => {
                     EVENT_LISTENING_ERRORS
-                        .with_label_values(&["kms_generation"])
+                        .with_label_values(&["ethereum"])
                         .inc();
                     consecutive_errors = consecutive_errors.saturating_add(1);
-                    warn!("KMSGeneration listening error: {e} ({consecutive_errors}/{max_errors})");
+                    warn!("Ethereum listening error: {e} ({consecutive_errors}/{max_errors})");
                     if consecutive_errors >= max_errors {
```
dvdplm, line 157: What is the default `max_consecutive_polling_errors`? I think setting the default explicitly in `Config` is helpful and avoids potential bugs with this set to `0`.

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -170,20 +182,48 @@ where
...
+        let kms_generation_address = self.config.kms_generation_contract.address;
+        let protocol_config_address = self.config.protocol_config_contract.address;
+
         let mut events = Vec::with_capacity(logs.len());
         for log in logs {
-            let event_kind = KMSGenerationEvents::decode_log(&log.inner)
-                .map_err(|e| anyhow!("Failed to decode KMSGeneration event: {e}"))?
-                .data
-                .try_into()?;
+            let log_address: Address = log.inner.address;
```
dvdplm:
```suggestion
            let log_address = log.inner.address;
```

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -170,20 +182,48 @@ where
...
+            let event_kind = if log_address == kms_generation_address {
+                KMSGenerationEvents::decode_log(&log.inner)
+                    .map_err(|e| anyhow!("Failed to decode KMSGeneration event: {e}"))?
+                    .data
+                    .try_into()?
+            } else if log_address == protocol_config_address {
+                let protocol_config_event = ProtocolConfigEvents::decode_log(&log.inner)
+                    .map_err(|e| anyhow!("Failed to decode ProtocolConfig event: {e}"))?
```
dvdplm, line 217: Maybe you could use `.context("bla bla")?` here to *add* info to any errors from `decode_log` rather than replace the error with `.map_err`? Not sure it's a good idea, but `context` is pretty neat and I don't think we're using it enough.

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -200,25 +240,22 @@ where
...
     /// Determines the block to start event listening from.
-    async fn get_start_block(
-        &self,
-        from_block_config: Option<u64>,
-        event_types: &[EventType],
-    ) -> anyhow::Result<u64> {
+    async fn get_start_block(&self, from_block_config: Option<u64>) -> anyhow::Result<u64> {
         if let Some(from_block) = from_block_config {
-            info!("Found configured from_block_number ({from_block}) for KMSGeneration polling");
+            info!("Found configured from_block_number ({from_block}) for Ethereum polling");
```
dvdplm: Could be a `debug!` perhaps?

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -277,13 +320,73 @@ mod tests {
...
             .unwrap();
 
         assert!(row.try_get::<bool, _>("is_valid").unwrap());
+        assert_eq!(
+            row.try_get::<Vec<u8>, _>("epoch_id").unwrap(),
+            epoch_id.as_le_slice()
+        );
+    }
+
+    #[rstest::rstest]
```
dvdplm, line 351: What does `rstest` do for us here? There are no parameters in the test I think?

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -277,13 +320,73 @@ mod tests {
...
+    #[timeout(Duration::from_secs(90))]
+    #[tokio::test]
+    async fn test_prepare_events_skips_genesis_new_kms_context() {
```
dvdplm, line 354: Do you think we need some tests to nail down corner cases, like context ID from the future, previous context id set to zero or bigger than the `contextId`, that sort of thing?

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -181,25 +183,22 @@ where
     }
 
     /// Determines the block to start event listening from.
-    async fn get_start_block(
-        &self,
-        from_block_config: Option<u64>,
-        event_types: &[EventType],
-    ) -> anyhow::Result<u64> {
+    async fn get_start_block(&self, from_block_config: Option<u64>) -> anyhow::Result<u64> {
```
dvdplm: naming nitpick: Rust typically tries to avoid prefixing getters with `get_` (and everyone ignores that guidline!). I for one think that `listener.start_block()` reads better than `listener.get_start_block()`... ;)

dvdplm: The `from_block_config` argument comes from the `Config` in the caller I think? If that is the case, can't we remove the argument here and get the block number from the config directly?

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -181,25 +183,22 @@ where
...
+        let last_block = sqlx::query_scalar!(
+            "SELECT block_number FROM last_block_polled_by_chain WHERE chain_name = $1",
+            chain,
         )
         .fetch_one(&self.db_pool)
         .await?;
 
-        match min_block {
+        match last_block {
             Some(last_block_polled) => Ok(last_block_polled as u64 + 1),
```
dvdplm:
```suggestion
            Some(last_block_polled) => Ok(last_block_polled.checked_add(1).expect("sane block number in the DB")),
```

If we ever end up with garbage data in the DB (like u64::MAX) I think we should panic.

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -181,25 +183,22 @@ where
...
         }
 
-        info!("Fetching min last block polled from DB for {event_types:?}...");
-        let min_block = sqlx::query_scalar!(
-            "SELECT MIN(block_number) FROM last_block_polled WHERE event_type = ANY($1::event_type[])",
-            event_types as &[EventType],
+        let chain = ChainName::Gateway.as_str();
+        info!("Fetching last block polled from DB for chain {chain}...");
+        let last_block = sqlx::query_scalar!(
+            "SELECT block_number FROM last_block_polled_by_chain WHERE chain_name = $1",
```
dvdplm, line 197: Is `chain_name` a Postgres an ENUM column or just a string?

kms-connector/crates/gw-listener/src/core/gateway.rs
```diff
@@ -181,25 +183,22 @@ where
...
+            chain,
         )
         .fetch_one(&self.db_pool)
         .await?;
 
-        match min_block {
+        match last_block {
             Some(last_block_polled) => Ok(last_block_polled as u64 + 1),
             None => {
                 info!("No block polled yet. Listening from latest block number instead...");
```
dvdplm, line 209: (commenting here but really meant line 212, `mod tests`)

No new tests added in this file. For the `EthereumListener` you did add a test. Do we need one here too? I'm not sure, your call!

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -21,20 +25,36 @@ use sqlx::{
...
 
+/// Chain identifier used as the primary key in `last_block_polled_by_chain`.
+#[derive(Debug, Clone, Copy)]
+pub enum ChainName {
+    Ethereum,
+    Gateway,
+}
+
+impl ChainName {
+    pub fn as_str(self) -> &'static str {
```
dvdplm: I think it's more idiomatic to take `&self` for `as_somethign` methods.

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -347,9 +379,77 @@ pub async fn publish_context_id(db_pool: &Pool<Postgres>, context_id: U256) -> a
...
+    // issue will validate this payload against the canonical encoding.
+    let kms_node_params = event.kmsNodeParams.abi_encode();
+    let thresholds = event.thresholds.abi_encode();
+    let pcr_values = event.pcrValues.abi_encode();
+
+    sqlx::query!(
+        "INSERT INTO new_kms_context(\
+            context_id, previous_context_id, kms_node_params, thresholds, software_version, \
+            pcr_values, tx_hash, created_at, otlp_context\
+        ) \
```
dvdplm: Do you really need to escape newlines like in bash?

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -347,9 +379,77 @@ pub async fn publish_context_id(db_pool: &Pool<Postgres>, context_id: U256) -> a
...
+        );
     }
     Ok(())
 }
+
+async fn publish_new_kms_context<'e>(
+    executor: impl PgExecutor<'e>,
+    event: NewKmsContext,
+    tx_hash: Option<FixedBytes<32>>,
+    created_at: DateTime<Utc>,
```
dvdplm, line 392: It seems to be the convention in this code base to use the event's creation timestamp as the source of truth? That is slightly unconventional I think, in the sense that most DBAs want their `create_at` columns to reflect the **database** clock (and be set automatically). We could have two datetime columns, `created_at` and `evt_created_at`? I am no good at hacking, but maybe there are attacks where the attacker manipulates clocks in such a way that the event's `created_at` data is messed up; if that happens we just ingest it here without question. Would that be bad?

kms-connector/crates/gw-listener/tests/common/mod.rs
```diff
@@ -185,6 +190,49 @@ pub async fn mock_event_on_gw(
                 .await?;
             (tx, event.into())
         }
+        TestEventType::NewKmsContext => {
+            let thresholds = rand_kms_thresholds();
+            let kms_node_params = vec![rand_kms_node_params()];
+            let software_version = format!("v{}", rand_u256());
+            let pcr_values = vec![rand_pcr_values()];
+            // `defineNewKmsContextAndEpoch` emits a non-genesis switch: contextId = base + 2,
+            // previousContextId = base + 1 (a real, non-sentinel predecessor).
```
dvdplm, line 199: I don't understand this comment!

kms-connector/crates/gw-listener/tests/common/mod.rs
```diff
@@ -185,6 +190,49 @@ pub async fn mock_event_on_gw(
...
+            let event = NewKmsContext {
+                contextId: KMS_CONTEXT_COUNTER_BASE + U256::from(2),
+                previousContextId: KMS_CONTEXT_COUNTER_BASE + U256::ONE,
+                kmsNodeParams: kms_node_params.clone(),
+                thresholds: thresholds.clone(),
+                softwareVersion: software_version.clone(),
+                pcrValues: pcr_values.clone(),
+            };
+            let tx = test_instance
```
dvdplm, line 208: My `rust-analyzer` thinks `tx` is of type `!`. Maybe we can help it out with a `tx: PendingTransactionBuilder<_>` here?

kms-connector/.sqlx/query-f486e4a3be5017b2bad9e05f6ee463794f85f7378545727fd3ed590e71fb2e20.json
eudelins-zama: Yes!

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Currently used to call this view function: https://github.com/zama-ai/fhevm/blob/4d9880d078dafb6f9d1631f8a6a4ad56643c3047/kms-connector/crates/gw-listener/src/core/ethereum.rs#L100-L105

That's the main difference :slightly_smiling_face:

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: FYI I don't have control on the `getCurrentKmsContextAndEpoch` naming, it comes from the `host-contracts`, I'm just using their Rust bindings. :innocent: 

That being said, I prefer this more exhaustive name rather than `kmsContext()` tbh.

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: It should be `ProtocolConfig` IMO. Good catch, will fix

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: IMO, Context and Epoch are two distinct things, having their distinct ID (and ID counters in the contracts) etc.

I prefer being more explicit. Maybe `publish_context_and_epoch` is enough though.

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: > I don't think there's anything we can do about it.

Indeed AFAIU.

> This allocation is annoying

It should done only once in the whole lifetime of the program, so I'm not super annoyed tbh :innocent:

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: It's `20`.

https://github.com/zama-ai/fhevm/blob/4d9880d078dafb6f9d1631f8a6a4ad56643c3047/kms-connector/crates/gw-listener/src/core/config.rs#L100-L102

> I think setting the default explicitly in Config is helpful and avoids potential bugs with this set to 0.

This is what we do AFAICT

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Printed only once in the whole lifetime of the program.

If something is wrong at this step, IMO it's nice to know it just by reading the logs, instead of having to restart it with a special env var to enable debug logs etc.

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: It's needed for the `#[timeout(Duration::from_secs(90))]` annotation.

https://github.com/zama-ai/fhevm/blob/4d9880d078dafb6f9d1631f8a6a4ad56643c3047/kms-connector/crates/gw-listener/src/core/ethereum.rs#L337

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Good question. It's not the job of the kms-connector IMO.

We trust the `host-contracts` to not put garbage in the events. It's not user input.

Only problematic case I see would be a malicious RPC node. In such case, I think it would need to be used by multiple parties to be able to mess our setup. But in such case it could do many bad things, not just this :sweat_smile:

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: > Rust typically tries to avoid prefixing getters with `get_`

It's not really a getter...
If it's set in the config, it uses it, otherwise, it will look at the DB to see the latest block that was polled.

Maybe the config lookup can be done outside the function, and this function called `fetch_...` instead if you prefer :slightly_smiling_face:

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: Just a string for now

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: I would say no. This covers it already: https://github.com/zama-ai/fhevm/blob/4d9880d078dafb6f9d1631f8a6a4ad56643c3047/kms-connector/crates/gw-listener/tests/block_tracking.rs 

For `EthereumListener`, there was more to tests (the genesis edge case)

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Not sure, maybe not! :slightly_smiling_face:

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: It was a tradeoff discussed for performance metrics computation.

The `created_at` is not part of the event or any user input, it's computed from the `kms-connector`, so attack seems unrealistic to me. Only issue could be a clock drift between DB and kms-connector.

This is why we chose to take the `kms-connector` as the source of truth, so that when computing performance metrics, the DB clock is not taken into account.

Hard to fully explain it in a thread, let me know if this need a huddle :slightly_smiling_face:

kms-connector/crates/gw-listener/tests/common/mod.rs
eudelins-zama: When we'll upgrade the contract, a `NewKmsContext` event will be emitted, called "genesis" because it would be the first one.

This event should be ignored. It is created so that the following context switch can use the `getKmsContextAnchor` as expected, otherwise there would be no previous event anchor to get the caCert etc. for the genesis context.

This comment is just here to explain for our integration tests, we on-purpose don't want to mock the genesis switch scenario

kms-connector/crates/gw-listener/tests/common/mod.rs
eudelins-zama: Hmm, no issues on my side :thinking: 

Screenshot From 2026-06-03 14-19-24

It's only a test file, do you think it's still worth the annotation?

kms-connector/crates/kms-worker/src/core/event_picker/notifier.rs
```diff
@@ -98,6 +106,8 @@ impl DbEventNotifier {
                 _ = prep_keygen_ticker.tick() => prep_keygen_ticker.deliver(),
                 _ = keygen_ticker.tick() => keygen_ticker.deliver(),
                 _ = crsgen_ticker.tick() => crsgen_ticker.deliver(),
+                _ = new_kms_context_ticker.tick() => new_kms_context_ticker.deliver(),
+                _ = new_kms_epoch_ticker.tick() => new_kms_epoch_ticker.deliver(),
```
dvdplm, line 110: This method I don't quite understand why it is this way. It seems redundant to me to have 7 tickers when in reality we only have two intervals ("fast" and "long"). The purpose seems to be to multiplex together the two event sources (tickers and real PG_NOTIFY events) to avoid executing redundant DB queries for the case where a ticker fires AND a real DB event happen close in time. To my eyes it seems like we're paying a high "code complexity price" for a rather small gain.

Maybe something like this could work:

1. Have only two tickers, "fast" and "long"
1. When "fast" fires, enqueue both PublicDecryptionRequest and UserDecryptionRequest polls. Same for "long": enqueue all queries to execute one by one (or in parallel)
1. When `db_listener.recv()` fires run that one

The downside is that all polling queries execute at regular interval. The current code starts that way too, but as real PG_NOTIFY events happen, each type's ticker is offset against the others so at a steady state polling queries get spread out somewhat evenly. I'd be surprised if this was a real performance issue though.

Another downside is what I mentioned above: with my suggested setup it becomes possible that a polling query happens close to a real PG_NOTIFY event so some wasted work.

Regardless of the suggested refactoring, some other suggestions:

- The `deliver()` method isn't really delivering anything; could use a rename;
- We could put all tickers in an array and use `futures::select_all` to get 

```
let notification = select! { 
	(kind, ..) = select_all(tickers.iter_mut().map(|t| Box::pin(t.tick()))) => kind,
	result = self.db_listener.recv() => …
}
```

…where `tick()` would be something like:

```
pub async fn tick(&mut self) -> EventType {
    self.ticker.tick().await;
    self.kind
}
```

- Similarly, the reset would collapse into

```
    if let Some(ticker) = tickers.iter_mut().find(|t| t.kind == notification) {
        ticker.reset();
    }
```

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
```diff
@@ -427,6 +494,27 @@ impl KmsClient {
...
+        .map_err(|e| {
+            ProcessingError::Irrecoverable(anyhow!(
+                "Invalid context_id in NewMpcContextRequest: {e}"
+            ))
+        })
+}
+
+fn irrecoverable_error<T>(err: anyhow::Error) -> (i16, Result<T, ProcessingError>) {
+    (0, Err(ProcessingError::Irrecoverable(err)))
+}
```
dvdplm, line 516: What does the `0i16` mean? EDIT: looks like it's the number of attempts we made while trying to do something? Why is it an `i16` though? Can it ever be negative?

dvdplm: nitpick: "unrecoverable" is probably more idiomatic.

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
```diff
@@ -0,0 +1,252 @@
...
+    NewMpcContextRequest, NewMpcEpochRequest, PcrValues, PreviousEpochInfo,
+};
+
+/// Builder for the KMS Core gRPC requests triggered by `ProtocolConfig` events.
+#[derive(Clone)]
+pub struct ProtocolConfigProcessor<P: Provider> {
+    /// EIP-712 domain of the `ProtocolConfig` contract.
+    domain: Eip712DomainMsg,
+
+    /// The `ProtocolConfig` contract instance, used to fetch the context event anchors.
```
dvdplm, line 27: I kind of know what "context event anchors" mean, but I think many readers will struggle. I think we should explain it properly.

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
```diff
@@ -0,0 +1,252 @@
...
+        let previous_context_anchor = self
+            .protocol_config_contract
+            .getKmsContextAnchor(new_context_event.previousContextId)
+            .call()
+            .await?;
+        let previous_context_block: u64 =
+            previous_context_anchor.emissionBlockNumber.saturating_to();
+
+        let filter = Filter::new()
+            .address(*self.protocol_config_contract.address())
```
dvdplm, line 112: What does this do? Casts a U256 to a u64? Is that ok? Why on earth is the `emissionBlockNumber` a U256 anyway?

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
```diff
@@ -0,0 +1,252 @@
...
+pub fn compute_anchor_event_hash(event: &NewKmsContext) -> FixedBytes<32> {
+    // keccak256(abi.encode(initialKmsNodeParams, initialThresholds, softwareVersion, pcrValues))
+    let encoded_data = (
+        event.kmsNodeParams.clone(),
+        event.thresholds.clone(),
+        &event.softwareVersion,
+        event.pcrValues.clone(),
+    )
+        .abi_encode_sequence();
+
```
dvdplm, line 309:
```suggestion
   let encoded_data = (
        event.kmsNodeParams.as_slice(),
        event.thresholds.clone(),
        &event.softwareVersion,
        event.pcrValues.as_slice(),
    )
 ```

File: kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs [hunk 27]
```diff
@@ -174,6 +183,49 @@ impl DbKmsResponsePublisher {
...
+        response: NewKmsContextResponse,
+        created_at: DateTime<Utc>,
+        otlp_ctx: PropagationContext,
+    ) -> anyhow::Result<PgQueryResult> {
+        sqlx::query!(
+            "INSERT INTO new_kms_context_responses(\
+                context_id, created_at, otlp_context\
+            ) \
+            VALUES ($1, $2, $3) ON CONFLICT DO NOTHING",
+            response.context_id.as_le_slice(),
```
dvdplm, line 193: Hmm, how come the bytes are little endian? Is that expected?

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
```diff
@@ -174,6 +183,49 @@ impl DbKmsResponsePublisher {
...
+        created_at: DateTime<Utc>,
+        otlp_ctx: PropagationContext,
+    ) -> anyhow::Result<PgQueryResult> {
+        sqlx::query!(
+            "INSERT INTO epoch_result_responses(\
+                context_id, epoch_id, keys, crs_list, created_at, otlp_context\
+            ) \
+            VALUES ($1, $2, $3, $4, $5, $6) ON CONFLICT DO NOTHING",
+            response.context_id.as_le_slice(),
+            response.epoch_id.as_le_slice(),
```
dvdplm, line 214: Same question here. Maybe it's obvious why they're little-endian but I'm still surprised!

kms-connector/crates/utils/src/types/kms_response.rs
```diff
@@ -172,13 +206,70 @@ impl KmsResponseKind {
...
+        keys.push(EpochKeyResult {
+            prepKeygenId: preproc_id,
+            keyId: key_id,
+            keyDigests: key_digests,
+            signature: key.external_signature.into(),
+        });
+    }
+
+    for crs in response.crs_responses {
+        let Some(crs_id) = crs.request_id.map(request_id_to_u256).transpose()? else {
```
dvdplm, line 259: `request_id_to_u256` reads the data as big-endian bytes, but then they seemed to be stored as little-endian bytes? Scary!

kms-connector/crates/utils/src/types/kms_response.rs
```diff
@@ -75,6 +82,23 @@ pub struct CrsgenResponse {
...
+pub struct NewKmsContextResponse {
+    pub context_id: U256,
+}
+
+/// Result of an MPC epoch creation.
+///
+/// `keys` and `crs_list` are the ABI-encoded `EpochKeyResult[]` / `EpochCrsResult[]` arrays
+/// expected by `confirmEpochActivation`.
+#[derive(Clone, Debug, PartialEq)]
+pub struct EpochResultResponse {
```
dvdplm, line 95: There's a type named exactly the same in `kms::kms_v1` which threw me off for a while. Not sure what you can do about that tbh.

kms-connector/crates/utils/src/types/event.rs
```diff
@@ -304,12 +340,67 @@ pub fn from_keygen_row(row: &PgRow) -> anyhow::Result<ProtocolEvent> {
     });
     Ok(ProtocolEvent {
         kind,
+        tx_hash: row
+            .try_get::<Vec<u8>, _>("tx_hash")
+            .ok()
+            .and_then(|h| FixedBytes::try_from(h.as_slice()).ok()),
```
dvdplm, line 344: So `h` comes from the DB, so it's LE yeah? And that's what `FixedBytes` expect?

kms-connector/crates/gw-listener/src/core/ethereum.rs
dvdplm: But couldn't the above be `self.config.protocol_config_contract.getCurrentKmsContextAndEpoch().call().await?`?

kms-connector/crates/gw-listener/src/core/ethereum.rs
dvdplm: So `default_max_consecutive_polling_errors` sets the default for `serde` (good!) but not for the Rust struct itself. I wonder if there's some neat way of impl `Default` for a struct and then automagically leverage that to make sure `serde` use the right values as well?

kms-connector/crates/gw-listener/src/core/ethereum.rs
dvdplm: ah, I thought that was a separate macro, sorry!

kms-connector/crates/gw-listener/src/core/ethereum.rs
dvdplm: Fair.

kms-connector/crates/gw-listener/src/core/gateway.rs
dvdplm: I do like `fetch` better; we'd have `fetch_start_block(&self)`, which is nice. I like it.

kms-connector/crates/gw-listener/src/core/gateway.rs
dvdplm: What are your thoughts on sql ENUMs? A hassle?

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: No, in `config`, we have the `address` (and optional EIP-712 domain config stuff) of the contract, we don't have the contract instance that exposes the `getCurrentKmsContextAndEpoch`.

If we wanted to do this as you propose, it would mean mixing networking stuff (connection to the RPC node etc.) with the configuration parsing

kms-connector/crates/gw-listener/src/core/publish.rs
dvdplm: Fair enough, I did not realize we augment the event in `kms-connector`. I think this makes sense.

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: No strong opinion, we used some already.

I don't think we'll need more values that what we currently have, so fine with me to go for Postgres enum. But very low risk to leave strings IMO.

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: `\` avoids line breaks in the string, that is the only difference.
Will remove them for readability

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Fixed dbebe8d5ddf0a5db0d9da94cb46390a4b99f622e

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: Fixed dbebe8d5ddf0a5db0d9da94cb46390a4b99f622e

kms-connector/crates/gw-listener/src/core/gateway.rs
eudelins-zama: Fixed dbebe8d5ddf0a5db0d9da94cb46390a4b99f622e

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Fixed 34996492e5ebb0e5627100b601469011599f6340

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Fixed 34996492e5ebb0e5627100b601469011599f6340

kms-connector/crates/tx-sender/src/core/kms_response_picker/picker.rs
```diff
@@ -223,4 +225,50 @@ impl DbKmsResponsePicker {
         .map(kms_response::from_crsgen_row)
         .collect()
     }
+
+    async fn pick_new_kms_context_responses(&self) -> anyhow::Result<Vec<KmsResponse>> {
+        sqlx::query(
+            "
+                UPDATE new_kms_context_responses
+                SET status = 'under_process'
```
dvdplm, line 233: What does "under process" mean?

kms-connector/crates/tx-sender/src/core/kms_response_picker/picker.rs
```diff
@@ -223,4 +225,50 @@ impl DbKmsResponsePicker {
...
+                FROM (
+                    SELECT context_id
+                    FROM new_kms_context_responses
+                    WHERE status = 'pending'
+                    LIMIT $1 FOR UPDATE SKIP LOCKED
```
dvdplm, line 238: How should I think about locked rows here, does that mean that some other process is updating them with the same data as this query is making? Like: why would rows be locked? And why is it ok to skip them?

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Wdym?

`default_max_consecutive_polling_errors` is used as the default value of the `Config` struct when deserialized (and also in the `Default` trait, but that is for tests only, not production code).
Then the `EthereumListener` only have access to its `config` field, so it's also using it.

kms-connector/crates/tx-sender/src/core/kms_response_picker/notifier.rs
```diff
@@ -111,6 +122,8 @@ impl DbKmsResponseNotifier {
                             KmsResponseNotification::PrepKeygen => prep_keygen_ticker.reset(),
                             KmsResponseNotification::Keygen => keygen_ticker.reset(),
                             KmsResponseNotification::Crsgen => crsgen_ticker.reset(),
+                            KmsResponseNotification::NewKmsContext => new_kms_context_ticker.reset(),
+                            KmsResponseNotification::EpochResult => epoch_result_ticker.reset(),
```
dvdplm, line 126: This is very similar to the other one. I guess the same questions apply here.

kms-connector/crates/utils/src/types/event.rs
```diff
@@ -414,15 +501,44 @@ async fn update_crsgen_status(
...
+    already_sent: bool,
+) {
+    let query = sqlx::query!(
+        "UPDATE new_kms_epoch SET status = $1, already_sent = $2 WHERE epoch_id = $3",
+        status as OperationStatus,
+        already_sent,
+        epoch_id.as_le_slice()
+    );
+    execute_update_event_query(db, query).await;
+}
```
dvdplm, line 535: What does the `as OperationStatus` do here?

kms-connector/crates/tx-sender/tests/response_picker/polling.rs
```diff
@@ -1,50 +1,30 @@
...
 
 #[rstest]
+#[case::public_decryption(TestResponseType::PublicDecryption)]
+#[case::user_decryption(TestResponseType::UserDecryption)]
+#[case::prep_keygen(TestResponseType::PrepKeygen)]
+#[case::keygen(TestResponseType::Keygen)]
+#[case::crsgen(TestResponseType::Crsgen)]
+#[case::new_kms_context(TestResponseType::NewKmsContext)]
+#[case::epoch_result(TestResponseType::EpochResult)]
 #[timeout(Duration::from_secs(60))]
```
dvdplm, line 18: nice!

kms-connector/crates/tx-sender/tests/response_picker/parallel.rs
```diff
@@ -95,7 +75,7 @@ async fn test_parallel_response_picking(request_str: &str) -> anyhow::Result<()>
     }
 
     info!("Done! Checking there is no uncompleted response in DB...");
```
dvdplm, line 77:
```suggestion
    info!("Done! Checking there is no incomplete response in DB...");
```

kms-connector/crates/tx-sender/tests/response_picker/parallel.rs
```diff
@@ -110,25 +90,34 @@ async fn init_response_picker(db: Pool<Postgres>) -> anyhow::Result<DbKmsRespons
 
 async fn check_no_uncompleted_response_in_db(
```
dvdplm, line 91:
```suggestion
async fn check_no_incomplete_response_in_db(
```

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Maybe I tested it wrong ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024)), but I think I prefer the `.map_err` approach

kms-connector/crates/kms-worker/src/core/event_picker/notifier.rs
eudelins-zama: Thanks for the detailed suggestion! :smile: 

There is probably room for improvement indeed. I suggest we handle this in a follow up PR as this one is already quite big, and that is a bit orthogonal to the PR (we are only extending the current behavior here).
Wdyt?

Here's the [tracking issue](https://github.com/zama-ai/fhevm-internal/issues/1497).

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
eudelins-zama: AFAIR, `i16` is because Postgres does not support unsigned int, so as the error counter is stored in DB I preferred to use `i16` to avoid back and forth to `u16`

eudelins-zama: For the "unrecoverable": we already named the error `ProcessingError::Irrecoverable`, so the function name is aligned with this.
Probably worth a follow up PR if we really want it as it would touches lot of files in the codebase.

Plus it does not seem to be a clear idiomatic way: source [reddit](https://www.reddit.com/r/NoStupidQuestions/comments/uyz6dm/what_is_the_difference_between_irrecoverable/) :joy:

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: > What does this do? Casts a U256 to a u64? 

Yes. We could panic instead maybe

> Why on earth is the emissionBlockNumber a U256 anyway?

Because in Solidity `block.number` is a `uint` ([docs](https://docs.soliditylang.org/en/latest/units-and-global-variables.html)) which is an alias for `uint256` ([docs 2](https://docs.soliditylang.org/en/latest/types.html#integers)).

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: This is not a kms-connector specific concept, so I feel this would be duplicated information to re-explain this in the kms-connector codebase.

Maybe worth adding a link to the [RFC](https://github.com/zama-ai/tech-spec/blob/main/rfcs/005-key-resharing.md#cross-context-party-communication) stating that this concept is defined there?

But would not help an external reader though...

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: Nice!

kms-connector/crates/utils/src/types/kms_response.rs
eudelins-zama: LE is only used to encode/decode `U256` when we store/retrieve them from the DB.

The rest of the time, we use BE for `U256` (Solidity convention + imposed by KMS Core in their `RequestId`)

This `request_id_to_u256` is for converting types received from GRPC requests/responses to the kms-core into `U256`.

kms-connector/crates/utils/src/types/kms_response.rs
eudelins-zama: Yes sorry about this :/

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: It's bytes, not `U256` so we don't have to choose LE/BE, we let the `FixedBytes` type handle this as they want under the hood in their `try_from()` / `to_vec()` functions

kms-connector/crates/tx-sender/src/core/kms_response_picker/picker.rs
eudelins-zama: It means this row was already picked by the `tx_sender` for processing (trying to send it to Ethereum).

So the next time the DB is polled for available responses, we ignore this row and don't send it twice

kms-connector/crates/tx-sender/src/core/kms_response_picker/picker.rs
eudelins-zama: In the case of the `tx_sender`, as we support only a single one per kms-connector, it is the same process, not another one.

We do this so the next time the DB is polled for available responses, we ignore locked row

kms-connector/crates/tx-sender/src/core/kms_response_picker/notifier.rs
eudelins-zama: Indeed, same [tracking issue](https://github.com/zama-ai/fhevm-internal/issues/1497) :slightly_smiling_face:

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: Needed for compile time verification of DB queries using custom types: https://docs.rs/sqlx/latest/sqlx/trait.Type.html#compile-time-verification

kms-connector/crates/tx-sender/tests/response_picker/parallel.rs
eudelins-zama: I'm not sure for this one.

To me, incomplete would mean we would look for responses with missing data, which are "incomplete".

Uncompleted mean that the responses are not processed yet, unfinished. This is what we want IMO.

Best source of truth corroborating my feeling: [reddit](https://www.reddit.com/r/EnglishLearning/comments/q3ha34/what_is_the_difference_between_incomplete_and/) :smile:

kms-connector/crates/tx-sender/tests/response_picker/parallel.rs
eudelins-zama: Same

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: Fixed 785a6b60aa5b124a6591c7896238e4d4a23d2dc1

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Fixed 785a6b60aa5b124a6591c7896238e4d4a23d2dc1

kms-connector/crates/gw-listener/src/core/ethereum.rs
dvdplm: I see. Ok, fair enough.

kms-connector/crates/gw-listener/src/core/ethereum.rs
dvdplm: You are correct, I somehow thought the `default_something` methods were used only for serde and not used also to impl `Default. My bad!

kms-connector/crates/gw-listener/src/core/ethereum.rs
dvdplm: Your playground link is perhaps wrong?

Anyway, I meant something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b06ce7b11e67201afe587730d49da0cf

```
use anyhow::{anyhow, Context};
#[derive(Debug)]
struct Out { data: Vec<u8>}
struct In { inner: String}

struct ProtocolConfigEvents;
impl ProtocolConfigEvents {
    fn decode_log(log: &str) -> Result<Out, std::io::Error> {
        // Ok(Out {data: vec![1,2,3]})
        let err = std::io::Error::new(std::io::ErrorKind::ConnectionRefused, "YOU SHALL NOT PASS");
        Err(err)
    }
}

fn main() -> Result<(), anyhow::Error> {
    let log = In { inner: "Beef".into()};
    let protocol_config_event = ProtocolConfigEvents::decode_log(&log.inner)
        .map_err(|e| anyhow!("Failed to decode ProtocolConfig event: {e}"))
        .context("Decoding ProtocolConfig beefy event and other good context")?
        .data;
    println!("protocol_config_event: {:?}", protocol_config_event);
    Ok(())
}
```

kms-connector/crates/gw-listener/src/core/gateway.rs
dvdplm: Last famous words! ;)

kms-connector/crates/gw-listener/tests/common/mod.rs
dvdplm: No, if it works better for you then let's not touch it. Maybe it's a regression in my nightly rust-analyzer. Or something.

kms-connector/crates/kms-worker/src/core/event_picker/notifier.rs
dvdplm: That works!

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
dvdplm: Solidity is that guy who, when at a cross roads with a sign saying "insanity" to the left and "free beer" to the right, consistently goes with "insanity".

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
dvdplm: lol

Ok, fine!

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
dvdplm: I say "external reader" but I really mean "me in two days". A link to the RFC is an excellent solution.

kms-connector/crates/tx-sender/src/core/kms_response_picker/picker.rs
dvdplm: I see. In hindsight perhaps `processing` would have been a better name.

kms-connector/crates/utils/src/types/event.rs
dvdplm: I see, I had no idea. Removing the `as ` does not signal any Rust-side error in the IDE, but I guess macros are funny like that.

kms-connector/crates/tx-sender/tests/response_picker/parallel.rs
dvdplm: I see, yeah "incomplete" would be "still missing a piece" . Maybe `Checking for unprocessed responses in the DB…`?

eudelins-zama: Leaving as it is then (it's test code logs so not that important!) :slightly_smiling_face:

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Oh thanks for the code snippet, indeed it can be useful! :pray: 

Keeping the original piece of code just because it's supposed to be an unreachable error (we only listen to events we're supposed to be able to decode), so not worth the extra context IMO.

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Fixed 50e1e5701b8804a8888066592eeecc09c122d612

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: Fixed 06193b752d350924402b75cb7ed27eb647b44bd0

charts/kms-connector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: kms-connector
 description: A helm chart to distribute and deploy the Zama KMS Connector services
-version: 1.5.0
+version: 1.5.2
```
eudelins-zama: `1.5.2` because this [PR](https://github.com/zama-ai/fhevm/pull/2865) will probably be merged first

piizama:
The chart changes look good to me. Thanks @eudelins-zama !

piizama (dismissed):

kms-connector/crates/gw-listener/tests/integration_test.rs
```diff
@@ -31,3 +37,76 @@ async fn test_publish_event(#[case] event_type: TestEventType) -> anyhow::Result
...
+        .send()
+        .await?
+        .get_receipt()
+        .await?;
+
+    poll_db_for_invalid_context(test_instance.db(), destroyed_context_id).await?;
+    assert_eq!(
+        fetch_valid_flags(test_instance.db(), other_context_id).await?,
+        vec![true],
+        "the untouched context should remain valid"
```
dvdplm, line 77: The test name "test_kms_context_destroyed_invalidates_context_epochs" made me expect the test to check that the two epochs associated with `destroyed_context_id` were also gone/invalid. 
Instead the the test only checks that the other context is untouched, which is surprising.

Do we need another test that checks the epochs? Or can it be added here?

kms-connector/crates/gw-listener/tests/integration_test.rs
```diff
@@ -31,3 +37,76 @@ async fn test_publish_event(#[case] event_type: TestEventType) -> anyhow::Result
...
+    let gw_listener_task =
+        start_test_listener(&mut test_instance, cancel_token.clone(), None).await;
+
+    // Two epochs of the context to destroy, plus one epoch of an untouched context
+    let destroyed_context_id = KMS_CONTEXT_COUNTER_BASE + U256::from(7);
+    let other_context_id = KMS_CONTEXT_COUNTER_BASE + U256::from(8);
+    for (context_id, epoch_id) in [
+        (destroyed_context_id, U256::from(1)),
+        (destroyed_context_id, U256::from(2)),
+        (other_context_id, U256::from(2)),
```
dvdplm: is it ok to have two epochs with identical IDs like this? I would have expected epoch IDs to be monotonically increasing, regardless of the context id.

kms-connector/crates/gw-listener/tests/integration_test.rs
```diff
@@ -31,3 +37,76 @@ async fn test_publish_event(#[case] event_type: TestEventType) -> anyhow::Result
...
+        vec![true],
+        "the untouched context should remain valid"
+    );
+    info!("Context successfully invalidated! Stopping GatewayListener...");
+
+    cancel_token.cancel();
+    Ok(gw_listener_task?.await?)
+}
+
+async fn fetch_valid_flags(db: &Pool<Postgres>, context_id: U256) -> anyhow::Result<Vec<bool>> {
```
dvdplm, line 89: nit: Maybe `fetch_is_valid_flag` (singular)?

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
```diff
@@ -1,49 +1,169 @@
...
-impl ContextManager for DbContextManager {
-    async fn validate_context(&self, context_id: U256) -> Result<(), ProcessingError> {
-        let context_row = sqlx::query!(
-            "SELECT is_valid FROM kms_context WHERE id = $1",
+impl<P: Provider> ContextManager for DbContextManager<P> {
+    #[tracing::instrument(skip(self))]
+    async fn validate_context(&self, extra_data: &ExtraData) -> Result<(), ProcessingError> {
+        let Some(context_id) = extra_data.context_id else {
+            return Ok(());
+        };
```
dvdplm, line 46: I think we should document that a missing context is ok (for backwards compat reasons I assume). Maybe it's documented on the trait already?

kms-connector/crates/kms-worker/tests/context.rs
```diff
@@ -229,3 +249,176 @@ async fn test_decryption_context_invalid(#[case] event_type: TestEventType) -> a
...
+    let err = context_manager
+        .validate_context(&extra_data)
+        .await
+        .unwrap_err();
+    assert!(
+        matches!(err, ProcessingError::Irrecoverable(_)),
+        "unexpected error: {err}"
+    );
+    Ok(())
+}
```
dvdplm, line 366: This test strikes me as a bit odd/superfluous. What is it testing actually? If the context is invalid, what does the unknown epoch matter? Would the outcome change if the epoch was known&valid?

kms-connector/crates/kms-worker/tests/context.rs
```diff
@@ -229,3 +249,176 @@ async fn test_decryption_context_invalid(#[case] event_type: TestEventType) -> a
...
+    assert!(
+        matches!(err, ProcessingError::Recoverable(_)),
+        "unexpected error: {err}"
+    );
+
+    let cached: i64 = sqlx::query_scalar("SELECT COUNT(*) FROM kms_context WHERE id = $1")
+        .bind(context_id.as_le_slice())
+        .fetch_one(test_instance.db())
+        .await?;
+    assert_eq!(cached, 0, "an inactive pair should not be cached");
```
dvdplm, line 333: It'd be nice to have the epoch become valid, retry the validation and assert we now have the expected data. Doable here?

dvdplm (commented):
LGTM, left some comments but no blockers.

kms-connector/crates/gw-listener/tests/integration_test.rs
eudelins-zama: Why? It returns a `Vec<bool>` :slightly_smiling_face:

kms-connector/crates/kms-worker/tests/context.rs
eudelins-zama: I just that in such case (destroyed context / unknown epoch), we'd like an irrecoverable error, and not a recoverable error as it would be for (valid context / unknown epoch), so no RPC fallback, no retries etc.

But agree this test is not super critical, I can remove it if needed

EDIT: I would keep it personally

kms-connector/crates/gw-listener/tests/integration_test.rs
eudelins-zama: The check is implicit in `poll_db_for_invalid_context`.
But I'll add a explicit `assert` for this :slightly_smiling_face:

kms-connector/crates/gw-listener/tests/integration_test.rs
eudelins-zama: At the contract level, it should not be possible no.
At the DB level however, it is possible.

Will update to be closer to "real life"

kms-connector/crates/gw-listener/tests/integration_test.rs
eudelins-zama: Test code anyway, will leave it like this

kms-connector/crates/kms-worker/tests/context.rs
eudelins-zama: I would argue that it is covered by the [test_validate_context_fallback_caches_valid_pair](https://github.com/zama-ai/fhevm/blob/25efb5ea63e8406056dea363a88e7076c7aba32b/kms-connector/crates/kms-worker/tests/context.rs#L277) (once the epoch become valid, then we are in this test area).

Wdyt?

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: Fixed f9430991378e16355e9acc5de9b651cd2a596d46

kms-connector/crates/gw-listener/tests/integration_test.rs
eudelins-zama: Fixed f9430991378e16355e9acc5de9b651cd2a596d46

kms-connector/crates/gw-listener/tests/integration_test.rs
eudelins-zama: Fixed f9430991378e16355e9acc5de9b651cd2a596d46

kms-connector/crates/kms-worker/tests/context.rs
dvdplm: Is this the matrix?

`(destroyed_context, unknown epoch)` => irrecoverable (this test)
`(valid_context, unknown epoch)` => recoverable (a test for this exists already I think?)
`(destroyed_context, known epoch)` => recoverable (is there a test for this?)
`(valid_context, known epoch)` => all good

dvdplm (dismissed):

kms-connector/crates/kms-worker/tests/context.rs
dvdplm: Fair

kms-connector/crates/kms-worker/tests/context.rs
eudelins-zama: No it's:

(destroyed_context, unknown epoch) => irrecoverable (this test)
(valid_context, unknown epoch) => recoverable (covered by [test_validate_context_pending_epoch_is_recoverable](https://github.com/zama-ai/fhevm/blob/eudelins/feat/1117/context-switch-event-listening-4/kms-connector/crates/kms-worker/tests/context.rs#L307C10-L307C60))
(destroyed_context, known epoch) => irrecoverable  (covered by [test_decryption_context_invalid](https://github.com/zama-ai/fhevm/blob/eudelins/feat/1117/context-switch-event-listening-4/kms-connector/crates/kms-worker/tests/context.rs#L153C10-L153C41))
(valid_context, known epoch) => all good

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
```diff
@@ -0,0 +1,258 @@
...
+        .kmsNodeParams
+        .iter()
+        .map(|n| MpcNode {
+            mpc_identity: n.mpcIdentity.clone(),
+            party_id: n.partyId,
+            external_url: n.ipAddress.clone(),
+            ca_cert: Some(n.caCert.to_vec()),
+            public_storage_url: n.storageUrl.clone(),
+            // Public key used to verify the signature of this KMS node
+            verification_key: Some(n.signerAddress.to_vec()),
```
isaacdecoded: While testing the KMS context switch, we faced the following error between the kms-connector-worker and the kms-core components:

```
error_string="Grpc failure on requestID 0100000000000000000000000000000000000000000000000000000000000000 with metric new_mpc_context and error code Client specified an invalid argument.
Error message: deserialization error: Could not deserialize key LimitExceeded"
```

This error seems to be related to this line where we are trying to map the `signerAddress` which is a 20-byte address to the `verification_key` which is [expected to be a PublicSigKey](https://github.com/zama-ai/kms/blob/main/core/service/src/cryptography/signatures.rs#L116-L138). 

Do you have any idea here @eudelins-zama @dd23? Seems like we need to map somehow the `signerAddress` to a valid `PublicSigKey` for the KMS core to process it. Also note that this `verification_key` is marked as optional in the gRPC definition.

eudelins-zama: Fixed by upgrading kms-core with this [PR](https://github.com/zama-ai/kms/pull/670)

dvdplm (approved):

isaacdecoded (approved):

eudelins-zama:
@mergifyio queue

PR #2704, opened 2026-06-01
feat(listener): block payload to publish full block payload and redis adaptative drain

nboisde:
Redis Streams and Rabbitmq message max size are 512MiB max which is quite more than a Full block size.
- Feature for getting full block for delegate consumer to filter events and log on full block.
- Adaptative draining feature for redis to allow faster message consuming at similar pace to rabbitmq.


maxnovawind (approved):
LGTM !!

nboisde:
@mergifyio queue

PR #2716, opened 2026-06-02
test(test-suite): add multi-KMS (threshold) support to CI e2e

Eikix:
## What

Adds a **threshold (multi-party) KMS topology** to the fhevm CI e2e harness so multi-party
KMS behavior can be validated pre-merge, instead of only the centralized single-node KMS
(which errors on context/epoch ops).

KMS topology is modeled in the existing scenario abstraction via an optional `kms` block.
When absent (every existing scenario), behavior is unchanged — today's centralized single
node. When `mode: threshold`, the harness generates:

- an **N-party KMS core cluster** — peer list + per-party `compose_{i}.toml`, a one-shot
  `kms-core-gen-keys` that generates **only signing keys** (`--cmd signing-keys`, sized to N),
  and `kms-core-init` (PRSS/epoch). `src/generate/kms-core.ts`, `core-threshold` component;
- **N per-party connectors** (mirrors `buildCoprocessorOverride`) — each with its own core
  endpoint, signer key, DB, and funded tx-sender;
- **N-signer contract wiring** — `NUM_KMS_NODES`, `MPC_THRESHOLD`, reconstruction thresholds
  (2t+1), N signer + N tx-sender addresses, Test params (`KEYGEN_PARAMS_TYPE=1`).

Scenario: `scenarios/four-party-threshold-kms.yaml` — 4 parties (t=1), **secure** keygen,
Test FHE params, multi-coprocessor (2-of-2).

## Keygen: secure, not insecure

Threshold mode uses **secure keygen** (real DKG preprocessing). The on-chain `KMSGeneration`
flow drives `prepKeygen -> keygen -> crsgen`; each party's connector signs the EIP-712
`KeygenVerification` over the **real on-chain `prepKeygenId`**, so it passes host verification
and `activeKeyId`/`activeCrsId` are set. There is **zero connector diff** — the published
connector image is used as-is.

> An earlier exploration used the Core's `InsecureKeyGen` endpoint (via a connector cargo
> feature). It was dropped: insecure keygen signs a static `INSECURE_PREPROCESSING_ID`, not
> the on-chain `prepKeygenId`, so it fails the host `KeygenVerification`
> (`KmsSignerDoesNotMatchTxSender`). Secure keygen avoids the bug entirely and, with Test
> params, finishes in ~360s for 4 parties — viable in CI. (Insecure work preserved on the
> `archive/multi-kms-insecure-keygen` tag.)

## How it runs in CI

The threshold flow is a **leg of the orchestrated e2e suite**
(`.github/workflows/test-suite-orchestrate-e2e-tests.yml`). Under the **`e2e orchestrate test`**
label (and on the merge queue / `release/*`), the suite builds the PR's images once, then runs
two legs in parallel off that single build + base-sha lock:

| Leg | Scenario | Profile |
|-----|----------|---------|
| `run-e2e-tests` | `two-of-three-multi-chain` | `standard` |
| `run-e2e-tests-threshold-kms` | `four-party-threshold-kms` | `kms-generation` |

So the threshold topology is validated against the **PR's built images**, not just published
ones. The threshold leg is intentionally **not** a merge-queue gate yet (`.mergify.yml` gates on
`run-e2e-tests / fhevm-e2e-test` only) — promote it to a `merge_condition` once it has proven
stable. A standalone threshold run (no full image build) is a direct dispatch:

```
gh workflow run test-suite-e2e-tests.yml -f scenario=four-party-threshold-kms \
  -f test-profile=kms-generation -f build=false
```

(There is no longer a dedicated `e2e multi kms` workflow — it was one entry point too many; the
`e2e multi kms` label is now unused.)

### What it asserts — the `kms-generation` profile

A new harness-side `fhevm-cli test kms-generation` profile turns the implicit boot signal into
named, on-chain checks:

- **on-chain KMSGeneration/ProtocolConfig audit (via `cast`):** `activeKeyId`/`activeCrsId` set
  and matching discovery; key + CRS material published; `getKeyParamsType`/`getCrsParamsType`
  equal the scenario's params (`Test`), proving the on-chain keygen/crsgen ran with the intended
  parameters; and `ProtocolConfig.getKmsSigners()` registers **N distinct signers** — read from
  chain, not scraped from KMS-core logs;
- **2t+1 quorum proof (fault injection):** decrypts at exactly the quorum (3/4 live), **cannot**
  decrypt below it (2/4 live), and **recovers** once parties restart — proving the threshold
  property is genuinely enforced, not incidental.

## Scope / deferred

- **Test params only.** Default (prod-size) params need a multi-hour DKG (32-96 GiB/party);
  deferred to a follow-up. (`random`/`standard` profiles flake on Test params, fhevm-internal#685,
  so the threshold leg uses the deterministic `kms-generation` profile.)
- **KMS-context guarantees** (context-aware signature verification, `confirmEpochActivation()`
  quorum, cross-context replay, resharing) are **out of scope** here — gated on RFC-005 landing in
  the host `KMSGeneration` contract + connector `epoch_id` validation. This PR builds the
  multi-party substrate that makes them testable; the assertions are a follow-up (not stubbed).
- **Promote the threshold leg to a merge-queue gate** once it has a stable track record.

## Notes

- `tsc` + harness unit tests pass (186); the threshold flow has run green end-to-end in CI
  (on-chain audit + 2t+1 quorum proof).
- Threshold mode blocks `--resume`/`--from-step` and KMS-group upgrades (the single-core
  lifecycle map doesn't model a cluster); recreate with a fresh `up`.


Eikix:
## Multi-KMS threshold e2e — decisions, rationale, and data

### Scope of this PR
- Adds an optional scenario `kms` block (`centralized` default | `threshold`). Threshold mode generates an N-party KMS core cluster + N per-party connectors + N-signer contract wiring. Absent block = unchanged centralized behaviour.
- **Secure keygen + Test FHE params only.** Threshold mode uses secure DKG; it uses the **published connector** (no local build).
- Adds a **dispatch-only** workflow `test-suite-e2e-threshold-kms` that reuses the standard e2e machinery to boot the threshold + multi-coprocessor (2-of-2) stack and run a KMS-relevant test profile.

### Why secure keygen (not insecure)
We first wired insecure keygen (to skip the multi-hour DKG). It is **blocked by a KMS-core bug**: the threshold insecure keygen signs `KeygenVerification.prepKeygenId` with a hardcoded `INSECURE_PREPROCESSING_ID` placeholder instead of the on-chain `prepKeygenId`, so the host contract recovers a non-registered signer and reverts `KmsSignerDoesNotMatchTxSender` — `activeKeyId` never gets set. (Root-caused via the core's gRPC `KeyGenResult` + EIP-712 recovery; `prepKeygen` and `crsgen` verify fine because only keygen binds `prepKeygenId`.)

**Secure keygen signs the real on-chain `prepKeygenId`**, so it passes verification. With **Test params it is ~360s for 4 parties** (measured) — viable for CI. So this PR uses secure keygen and drops the insecure-keygen connector changes (preserved on tag `archive/multi-kms-insecure-keygen` for when we need them — YAGNI).

### Data (local, 4-party t=1, secure keygen, Test params, latest-main `c57f52f` core)
| | result |
|---|---|
| Boot (clean → stack ready) | ~600s |
| `getActiveKeyId()` | set (was `0` with insecure keygen) |
| `user-decryption` profile | **8 passing** (each needs 2t+1 = 3 of 4 party shares) |
| `light` (input-proof + ERC20) | passing |

A version note: the KMS core must be the one paired with the SDK/coprocessor (`c57f52f` for latest-main). Pinning an unmatched core (`v0.13.20-0`) caused a tfhe-rs key-format mismatch (`safe_deserialize` failure on encrypt). The workflow uses the resolved latest-main core, so this is automatic.

### Deferred (follow-ups)
- Insecure keygen — needs the KMS-core fix (sign the on-chain `prepKeygenId`, not the placeholder).
- Default FHE params — secure DKG with Default is hours + 32–96 GiB/party.
- KMS-context test profiles (context-aware sig verification, `confirmEpochActivation` quorum, cross-context replay) and resharing (RFC-005, fhevm 0.14) — the workflow's `test-profile` input is the extension point.

Eikix:
CI fails due to a https://github.com/zama-ai/fhevm-internal/issues/1495 flagged flake on e2e auto revert

test-suite/fhevm/src/generate/compose.ts
```diff
@@ -425,11 +427,56 @@ const buildCoprocessorOverride = async (plan: StackSpec) => {
...
 };
 
+/**
+ * Replicates the KMS connector tier once per threshold party (mirrors
+ * buildCoprocessorOverride). Party 1 keeps the base `kms-connector-*` names so
+ * it replaces the single-node template; parties 2..N get `kms-connector-{i}-*`.
+ * Each party's services read `kms-connector[.i].env` (own core endpoint, signer
+ * key, DB) and depend on that party's own db-migration.
+ */
+const buildKmsConnectorOverride = async (plan: StackSpec) => {
```
PanGan21: `--override kms-connector` is ignored here in threshold mode. We should either not support it and return error or allow for local build

test-suite/fhevm/scenarios/four-party-threshold-kms.yaml
```diff
@@ -0,0 +1,33 @@
...
+  resharing (RFC-005) is deferred to fhevm 0.14. Uses the published connector
+  (no local build); Default FHE params are deferred to a follow-up PR.
+topology:
+  count: 2
+  threshold: 2
+instances:
+  - index: 0
+    args:
+      host-listener-consumer:
+        - --service-name=coprocessor-host-listener-consumer
```
isaacdecoded, line 22: nit: to align the service name with the one below at line 25.

```suggestion
        - --service-name=coprocessor0-host-listener-consumer
```

test-suite/fhevm/src/commands/kms-generation.ts
```diff
@@ -0,0 +1,168 @@
...
+    reconstruct,
+    stopForTolerance: parties - reconstruct,
+    stopForFloor: parties - (reconstruct - 1),
+  };
+};
+
+/** Stops or starts each container, tolerating already-stopped / missing ones. */
+const setRunning = async (containers: string[], action: "start" | "stop") => {
+  for (const container of containers) {
+    await run(["docker", action, container], { allowFailure: true });
```
PanGan21, line 43: An agent finding on this: 
`docker stop runs with allowFailure: true and the result is discarded, with no dockerInspect check that the container reached exited. If a stop silently no-ops, the below-quorum probe runs with all parties alive, decryption succeeds, and the harness throws "the 2t+1 quorum is not enforced" — a red build with a wrong diagnosis. The floor probe's infra-vs-quorum guard only protects the negative verdict, not this positive one. Suggest asserting State.Status === "exited" after each stop (a waitForPartiesStopped mirroring the existing waitForPartiesRunning).`

test-suite/fhevm/src/commands/kms-generation.ts
```diff
@@ -0,0 +1,168 @@
+/**
+ * The `kms-generation` acceptance profile for `fhevm-cli test`.
+ *
+ * It turns the implicit "the boot didn't fail" signal into named, on-chain assertions about the
+ * KMSGeneration flow, then proves the threshold property is real by varying the live-party count
```
isaacdecoded: I'd suggest reviewing all "KMSGeneration flow" and "threshold" mentions since I think these terms are not fully clear:

- There's no "KMSGeneration flow," but FHE key and CRS generation flows - KMSGeneration is the smart contract implementing such flows. 

- Also, "threshold" alone could be a bit confusing since it's widely used in the protocol across multiple use cases - so maybe we can use "threshold-mode" when referring to the KMS cluster setup property.

test-suite/fhevm/scenarios/four-party-threshold-kms.yaml
Eikix: instance 0 is unprefixed everywhere else (see two-of-two.yaml and the coprocessor-* vs coprocessor1-* container names)
 but indeed we should align the repo

PanGan21 (approved):

Eikix:
@mergifyio queue

PR #2721, opened 2026-06-03
fix(gateway-contracts): remove `whenDecryptionPaused` modifer from KMS threshold (N-01)

zmalatrax:
Fixes https://github.com/zama-ai/fhevm-internal/issues/1484


gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -151,6 +151,7 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         /// @notice The user decryption threshold per context
         mapping(uint256 contextId => uint256) userDecryptionThresholdForContext;
         /// @notice The MPC threshold per context
+        /// @dev The SDK derives the MPC threshold from the MPC nodes it knows about instead of reading this value.
```
obatirou: The SDK should not query gateway anymore but ProtocolConfig.
This comment should be added to the getter on ProtocolConfig.

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -392,7 +393,9 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
     function updateMpcThresholdForContext(
         uint256 contextId,
         uint256 newMpcThreshold
-    ) external virtual onlyOwner whenDecryptionPaused {
+    ) external virtual onlyOwner {
+        // This threshold has no Gateway-side on-chain consumer that requires draining
+        // in-flight decryption requests.
```
obatirou:
```suggestion
```

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -428,7 +431,9 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
     function updateKmsGenThresholdForContext(
         uint256 contextId,
         uint256 newKmsGenThreshold
-    ) external virtual onlyOwner whenDecryptionPaused {
+    ) external virtual onlyOwner {
+        // This threshold is consumed by host-side KMS generation, not Gateway-side
+        // decryption.
```
obatirou:
```suggestion
```

gateway-contracts/contracts/GatewayConfig.sol
zmalatrax: removed from GatewayConfig.sol, added to ProtocolConfig.sol

obatirou (approved):

isaacdecoded (changes requested):
The issue says that we need to remove the `whenDecryptionPaused` modifier from the `updateMpcThresholdForContext` and `updateKmsGenThresholdForContext`, but this PR is also changing other methods like for updating decryption related thresholds and KMS context updating. I think we shouldn't change these others since the modifier was added precisely to avoid stuck decryption responses after these thresholds or KMS context changes.

immortal-tofu (approved):

isaacdecoded (approved):
It was internally discussed to rely on operational control for draining any ongoing decryption request before changing decryption-related thresholds or KMS contexts info.

leventdem (approved):
OK!

PR #2722, opened 2026-06-03
fix(gateway-contracts): add `contextId` validation on decryption requests (M-01)

zmalatrax:
Fixes https://github.com/zama-ai/fhevm-internal/issues/1478

> Separately, the request entrypoints pin the caller-supplied context without checking that it refers to a valid context, deferring all validation to response time. A request that pins a context which is not valid still pays its fee at request time and can only fail later, when its responses revert on the invalid context.


obatirou (approved):

isaacdecoded (approved):
LGTM! 💯

PR #2724, opened 2026-06-03
fix(gateway-contracts): pin migration export to a finalized block hash (L-05)

obatirou:
Fixes zama-ai/fhevm-internal#1483


zmalatrax (approved):
lgtm

PR #2728, opened 2026-06-03
test(test-suite): fail fast when a coprocessor is down before drift injection

Eikix:
## Why

`ciphertext-drift-auto-recovery` has been flaky in CI (`fhevm-e2e-test`). Investigating two failing jobs showed both failures share one upstream cause: **one coprocessor instance came up degraded and submitted nothing for the entire run.**

In each failing run, that node's own gw-listener logged `Consensus was observed but local digests never became available for comparison` ~95–97× and it never appeared as a submitter. In one job the crash is captured directly: a coprocessor's `sns-worker` hit `duplicate key value violates unique constraint "ciphertext_digest_pkey"` (23505), classified it as a non-transient error, and **shut the worker down** — after which it produced no digests.

The drift test has **zero liveness margin**: with `two-of-three` it corrupts one coprocessor and needs both remaining honest nodes to reach consensus so the faulty node auto-reverts. If any coprocessor is down:
- if it's the injection target → the injection trigger never fires (`timed out waiting for drift injection trigger`);
- if it's an honest peer → no honest majority, no consensus, no revert (`timed out waiting for drift_revert_signal to reach status=reverting`).

Either way the test burns its full timeout (~180–425s) and reports a misleading error instead of the real one: a degraded cluster.

## What this does

Adds a **pre-injection guard**: before injecting drift, check that every coprocessor is still submitting `AddCiphertextMaterial` — distinct senders across the recent on-chain event window must be `>= topology.count`. If not, fail fast with the real cause and which senders were seen.

This is intentionally small and only changes the test harness. It does **not** fix the underlying coprocessor crash (the `sns-worker` exiting on a `ciphertext_digest_pkey` 23505 — the PK is still the legacy composite `(tenant_id, handle)` while the upsert keys on `(handle)`); that needs a separate coprocessor-side fix. The guard just makes the flake legible and fast instead of a confusing timeout.

## Test

- `bun test src/cli.test.ts` — 36 pass (incl. new `recentCoprocessorSenders` unit test)
- `tsc --noEmit` — clean

### DIFF
--- a/test-suite/fhevm/src/cli.test.ts
+++ b/test-suite/fhevm/src/cli.test.ts
@@ -8,6 +8,7 @@ import {
   dbRevertDeleteExpectations,
   dbRevertTargetBlock,
   keyBootstrapLogArgs,
+  recentCoprocessorSenders,
   validateNamedProfileGrep,
   waitForKeyBootstrap,
 } from "./commands/test";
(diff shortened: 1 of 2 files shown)


antoniupop (dismissed):

PanGan21 (approved):

Eikix:
@mergify queue

PR #2731, opened 2026-06-03
fix(coprocessor): collapse overlapping unique keys on ciphertext_digest, computations, and ciphertexts

Eikix:
## What

Collapse the **overlapping unique keys** on three tables down to a single unique
key each, by promoting the existing tenant-free UNIQUE index to the primary key
and dropping the vestigial tenant-prefixed composite PK:

| table | dropped (vestigial) PK | promoted to PK (already existed) |
|---|---|---|
| `ciphertext_digest` | `(tenant_id, handle)` | `idx_ciphertext_digest_no_tenant (handle)` |
| `computations` | `(tenant_id, output_handle, transaction_id)` | `idx_computations_no_tenant (output_handle, transaction_id)` |
| `ciphertexts` | `(tenant_id, handle, ciphertext_version)` | `idx_ciphertexts_no_tenant (handle, ciphertext_version)` |

This is the **root-cause fix** for the flaky `ciphertext-drift-auto-recovery`
e2e failures and the underlying sns-worker crash — and it closes the *same class*
of failure on the two other tables that share the pattern.

## Why

Since `remove_tenants`, `tenant_id` is a constant default, yet these three tables
each still carry **two overlapping unique indexes**: a vestigial tenant-prefixed
PRIMARY KEY plus a tenant-free UNIQUE index that the writers actually target.

Each writer upserts with `INSERT … ON CONFLICT (<tenant-free cols>)`, whose
arbiter covers only the tenant-free index:

```
sns-worker     ON CONFLICT (handle) DO UPDATE              -> ciphertext_digest
host-listener  ON CONFLICT (output_handle, transaction_id) -> computations
tfhe-worker    ON CONFLICT (handle, ciphertext_version)    -> ciphertexts
```

When a state/drift revert runs `DELETE FROM <table>` concurrently with such an
upsert, the duplicate can surface on the **other** unique index (the composite
PK), which the arbiter does not cover — raising a hard `duplicate key value
violates unique constraint "..._pkey"` (SQLSTATE 23505) instead of being absorbed
by `ON CONFLICT`. The sns-worker treated that as fatal, exited (code 0, swallowed)
and was not restarted, so the coprocessor produced no further digests and the
zero-margin drift test failed. host-listener and tfhe-worker share the identical
upsert+revert shape, so the same race is reachable on `computations` and
`ciphertexts`.

Collapsing each table to a single unique key removes the blind spot: the
`ON CONFLICT` arbiter then covers the table's only unique index, so any conflict
is always absorbed by the upsert. Dropping `tenant_id` from the key is safe
because handles already embed `chain_id`, so the tenant-free columns are globally
unique on their own.

Full root-cause writeup, proofs, diagrams, and timeline: zama-ai/fhevm-internal#1495.

## Safety / cost

- **Catalog-only, no backfill.** Each tenant-free unique index already exists and
  its columns are already `NOT NULL` (they came from the old composite PKs), so
  `ADD PRIMARY KEY USING INDEX` promotes the existing index in place — no table
  rewrite, no scan. Verified on PostgreSQL 15.7: all three `ALTER`s leave
  `relfilenode` unchanged, and at 3M rows the migration runs in ~13 ms (it does
  **not** scale with row count or payload size).
- No foreign key references any of the three PKs (checked), so the swaps are safe.
- After the change each table has **exactly one** unique index (its new PK), so no
  residual blind spot remains and every writer's column-inference `ON CONFLICT`
  arbiter still resolves — no worker code change needed.
- `tenant_id` is kept as a plain column on every table.

## No-downtime rollout

The only real exposure is the brief `ACCESS EXCLUSIVE` lock each `ALTER` takes:
the operation is sub-millisecond, but while it waits to acquire the lock, traffic
queues behind it. This will be rolled out with a **bounded `lock_timeout`** (a few
seconds) so a stuck/long transaction can never turn the swap into an unbounded
stall — if it can't acquire quickly it aborts and is retried, rather than blocking
the table. The migration is first validated on a **testnet debug DB** restored
from the latest RDS snapshot (≈40M-row tables) to confirm catalog-only behavior at
scale before mainnet.

## Verification

- **PostgreSQL 15.7, all three tables:** replayed the post-`remove_tenants` schema
  (composite PK + tenant-free UNIQUE + non-unique indexes), seeded rows, applied
  the migration → each table ends with a single `..._pkey PRIMARY KEY` on the
  tenant-free columns, `relfilenode` unchanged (no rewrite), all rows preserved.
- **Race closed (before/after):** the concurrent `INSERT … ON CONFLICT … DO …`
  + `DELETE` workload produces duplicate-key `..._pkey` (23505) errors on the
  two-index schema and **0** once collapsed to a single unique key — reproduced on
  both `ciphertext_digest` (375 → 0) and `computations` (4 → 0; same `_pkey`
  signature).

Supersedes the interim retry-path approach (closed); the test-suite fast-fail
guard (fhevm#2728) is orthogonal and stays.


PanGan21 (dismissed):
LGTM!

PanGan21:
Maybe also generate sqlx cache?

Eikix:
> Maybe also generate sqlx cache?

are you sure there is any sqlx cache to regenerate? we dont do queries just a migration

PanGan21 (approved):

rudy-6-4:
git grep tenant_id return some lines in tfhe-worker/src/tests/migrations.rs
is the test still relevant ?

Eikix:
> git grep tenant_id return some lines in tfhe-worker/src/tests/migrations.rs is the test still relevant ?

seems that test (test_remove_tenants_migration_with_data) is scoped to the remove_tenants migration
still reelvant

Eikix:
@mergify queue

PR #2755, opened 2026-06-07
feat: improve Solana track

nicolasgarcia214:
The four big changes are:
  
  1. **FHE eval is now plan-based**
      - What: `zama-host::fhe_eval` now runs a whole multi-step FHE plan, including binary ops, ternary/select, random, trivial encrypt, and input binding.
      - Why: avoid creating durable ACL accounts for temporary scratch values. Similar to the execute_frame functionality on the `openzeppelin-solana-track` branch but more robust.

  2. **Operator/delegated transfers were removed**
      - What: set_operator, close_operator, confidential_transfer_from, operator callback paths, and operator state are gone.
      - Why: simpler authority model, Solana-native approach, and fewer stale/rent/approval edge cases.

  3. **Single verifier authorities became verifier sets**
      - What: zama-host now has versioned threshold VerifierSet accounts. Token mints now point to disclosure and redemption verifier sets.
      - Why: replace “one trusted signer” with scoped N-of-M quorum checks.

  4. **Disclosure/redemption now use request witness accounts**
      - What: new DisclosureRequest and BurnRedemptionRequest accounts bind request hash, material commitment, verifier set version, chain id, expiry, and target handle.
      - Why: a KMS certificate should prove exactly which on-chain request it answers.


Eikix:
[AUTOMATED]

No findings from the code review.

CI is still failing, so this should not be merged until the failing checks are resolved.

Clarifications on deviations:
- Operator/delegated-transfer removal appears intentional and documented as an ERC7984 parity gap.
- `ProtocolConfig` / `KMSGeneration` / canonical KMS context mirroring is not implemented here and is documented as product-open.
- Native-v0 response publishing / gateway-compatible wiring also remains product-open.

Clarification needed on `VerifierSet`:
`VerifierSet` appears to be token/host-local PoC verifier configuration, not a canonical protocol `ProtocolConfig` mirror. This should be stated explicitly in the PR or docs, including whether token disclosure/redemption verifier sets can authorize certificates independently from the eventual canonical protocol KMS context.

coprocessor/fhevm-engine/host-listener/src/solana_adapter.rs
```diff
@@ -56,50 +142,492 @@ pub struct SolanaIngestStats {
...
+                && log_events.iter().any(needs_ordered_db_log_index) =>
+        {
+            Err(SolanaEventDecodeError::MixedHostEventTransport {
+                cpi_events: cpi_events.len(),
+                log_events: log_events.len(),
+            })
+        }
+        (false, false) => {
+            Ok(cpi_events.into_iter().chain(log_events).collect())
+        }
```
KitHat, line 272: Two branches with the same arm.

coprocessor/fhevm-engine/host-listener/src/solana_adapter.rs
```diff
@@ -56,50 +142,492 @@ pub struct SolanaIngestStats {
...
+    events
+}
+
+pub fn merge_solana_transport_events(
+    cpi_events: Vec<SolanaHostEvent>,
+    log_events: Vec<SolanaHostEvent>,
+) -> Result<Vec<SolanaHostEvent>, SolanaEventDecodeError> {
+    match (cpi_events.is_empty(), log_events.is_empty()) {
+        (true, _) => Ok(log_events),
+        (_, true) => Ok(cpi_events),
```
KitHat, line 260: are we giving priority to log events over cpi events?

solana/programs/confidential-token/src/instructions/create_random_amount.rs
```diff
@@ -76,28 +76,74 @@ fn create_random_amount_inner(
...
+                    mint_key,
+                    ctx.bumps.compute_signer,
+                )?,
+                output_authority: fhe::OutputAuthority::transaction_signer(&ctx.accounts.owner),
+                output: amount_output,
+                system_program: &ctx.accounts.system_program,
+            },
+            zama_fhe::BoundedU64UpperBound::from_be_bytes(upper_bound)
+                .map_err(invalid_eval_plan)?,
+        )?,
```
KitHat, line 103: Is it intentionally not using zama_fhe::EvalBuilder

solana/programs/confidential-token/src/state/transfer_callback_settlement.rs
```diff
@@ -51,5 +47,5 @@ pub struct TransferCallbackSettlement {
 
 impl TransferCallbackSettlement {
     /// Serialized size of the account body, excluding Anchor discriminator.
-    pub const SPACE: usize = (32 * 19) + 2;
+    pub const SPACE: usize = (32 * 17) + 2;
 }
```
KitHat, line 51: It is a matter of taste, but I dislike the manually declared space.
It was quite easy before the agentic development and now it is even easier to update the structure and forget to update the space. At the very best integrational tests will catch it, otherwise it can slip into production. I suggest to do one of the two actions:

1. Use InitSpace derivation macro
2. Write a test per each manually declared space where space is matched against the struct.

KitHat (approved):
Looks good!
One more thing to think about.
Maybe we can write some macros to adapt fhe::eval?
It is really hard to read the logic of the program because of many lines in each.
However, it is a request for the future and production phase.

coprocessor/fhevm-engine/host-listener/src/solana_adapter.rs
nicolasgarcia214: No, they're two separate arms. They share the same pattern but the first one has a match guard.

KitHat: I am sorry, missed the match guard.

PR #2757, opened 2026-06-09
feat(coprocessor): gw listener and tx sender use an http url

rudy-6-4:


antoniupop (commented):
Some fair points raised by a review agent: 

  - Proof-row data loss on LB 5xx — the new retry policy retries connection errors (Custom) and 429/503, but an nginx/ingress 502/504 surfaces as HttpError, which is neither retried by the layer nor matched by is_retry_err()/is_backend_gone in the ops. In verify_proof it lands in the generic branch that burns retry_count (default 6, remove_after_max_retries=true, 1-2-4s sleeps) — a ~20s LB blip can permanently DELETE queued verify_proofs rows. Under WS the same outage was a reconnect/BackendGone, never per-row retry burn.
  - Compat break that CI won't catch — the shared compose now feeds http://… to whatever image tag is pinned, every released gw-listener/tx-sender image is WS-only, and compat.ts has no legacy shim mapping the flags back to GATEWAY_WS_URL (which is still rendered, now unconsumed). The compat-smoke gate kills services at 3s and treats timeout as pass, so this ships silently and bites the v0.12→v0.13 rollout baseline and latest-supported/network targets.
  - Helm breaking rename under a patch bump — existing values setting commonConfig.gatewayUrl render no GATEWAY_URL env while args still pass --gw-url=$(GATEWAY_URL) literally → CrashLoopBackOff after a "patch" upgrade of the published OCI chart. The repo's own _helpers.tpl has a fail-on-deprecated-key precedent that wasn't used.
  - Permanent retry-layer degradation via leaked counter — alloy 1.1.2's RetryBackoffService only decrements requests_enqueued on return paths (no Drop guard). Your 4s send_txn_sync/health-check timeouts ≤ the 4s retry backoff guarantee dropped futures during outages, each leaking +1 forever; with CUPS=100 (5 req/s budget) ~25 leaks already push every retry past the 4s timeouts until restart.
  - gw-listener didn't get the custom policy — it uses RetryBackoffLayer::new (default policy), so connection errors aren't retried at all there; --provider-max-retries/--provider-retry-interval are effectively dead config and the README rewritten in this PR documents behavior the code doesn't have (it also still shows the stale 1000000 default vs u32::MAX). Each blip restarts run_get_logs and can leave drift detection silently disabled.

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

rudy-6-4:
@claude /pr-review

PR #2759, opened 2026-06-11
fix(coprocessor): revert input rerand before expand (#2383)

antoniupop:
This reverts commit 847c02777d92a49718d3341bfcd38fd4c860fc7a.

The original change is consensus-breaking: re-randomising inputs before expand instead of before compression means ciphertexts are no longer bitwise-equal across coprocessors for the same input with the planned upgrade process. The upcoming release cannot ship consensus-breaking changes without production downtime, so we revert now and re-introduce this once blue-green upgrade is available.


rudy-6-4 (approved):

antoniupop:
@mergifyio queue

PR #2657, opened 2026-05-26
feat(host-contracts): multi-chain deploy/verify split + canonical ProtocolConfig export

Eikix:
The multi-chain host-contract deploy/verify task split, plus a DAO-reviewable canonical `ProtocolConfig` export. Supersedes #2626 (the original split PR, now closed); this is the version that lands on `main`.

## What's in this PR

Three related but distinct changes — reviewable separately:

1. **Deploy-role split** (originally #2626). Host deployment is now canonical vs secondary. Canonical owns `KMSGeneration` and seeds the first `ProtocolConfig` KMS context; secondary mirrors `ProtocolConfig` from canonical and never deploys `KMSGeneration`. Verify tasks split the same way.
2. **Block-explorer verify resilience** (folds in #2718). Per-contract `task:verify*` tolerate the benign "Already Verified" response so a re-run from a deploy script does not abort under `set -eo pipefail`.
3. **DAO-reviewable mirror seed.** `task:exportCanonicalProtocolConfig` reads the canonical KMS context pinned to a block and writes a JSON snapshot. DAO signers re-run it with `--block-number <blockNumber>` to reproduce the artifact at that exact block — reproducible even after a later committee rotation (`defineNewKmsContext`) — and diff it before accepting secondary-host ownership. The secondary compose flow also waits for the canonical `PROTOCOL_CONFIG_CONTRACT_ADDRESS` before deploying.

## Task surface

```text
task:deployCanonicalHost
  -> task:deployHostSkeleton
  -> task:deployProtocolConfigCanonical
  -> task:deployKMSVerifier
  -> task:deployKMSGeneration              # canonical only

task:deploySecondaryHost --canonical-rpc-url <URL> --canonical-protocol-config-address <ADDR>
  -> task:deployHostSkeleton --skip-kms-generation
  -> task:deployProtocolConfigSecondary    # mirrors canonical via initializeFromMigration
  -> task:deployKMSVerifier

task:exportCanonicalProtocolConfig --canonical-rpc-url <URL> --canonical-protocol-config-address <ADDR> [--block-number <N>]
  -> writes a canonical snapshot for DAO review (no deploy); --block-number reproduces a prior export

task:verifyCanonicalHost     # host contracts + ProtocolConfig + KMSVerifier + KMSGeneration
task:verifySecondaryHost     # same, excluding KMSGeneration
```

[...]

PR #2772, opened 2026-06-12
feat(host-contracts): export and mirror canonical ProtocolConfig state

Eikix:
## Summary

- Implements zama-ai/fhevm-internal#1501: backports the export/mirror pieces of #2657 onto the 0.13.x line, excluding the canonical/secondary deploy-task split — extended with an artifact-driven apply and a DAO path so the flow matches how infra actually operates per environment.
- The flow is **artifact-centric**: export → review → apply, with the JSON artifact as the single pivot so what is applied is exactly what reviewers diffed.

### Export
- `task:exportCanonicalProtocolConfig` writes a reviewable JSON artifact (canonical chainId, pinned blockNumber, contract address, current KMS context id, KMS nodes, all four thresholds; bigints as strings). Re-running with `--block-number` reproduces a prior artifact byte-for-byte even after a `defineNewKmsContext` rotation.
- `readCanonicalSnapshot` (ported from #2657) pins all reads to one block and guards identity (`getVersion`), no-active-context, and destroyed-context cases.

### Apply — one prepare step, two executors
- **devnet / local**: `task:deployProtocolConfigFromCanonical --snapshot <artifact.json>` — direct upgrade with `DEPLOYER_PRIVATE_KEY`, no canonical RPC needed. Without `--snapshot` it still supports the live-read mode (`--canonical-rpc-url` + `--canonical-protocol-config-address`) for quick iteration.
- **testnet / mainnet (DAO)**: `task:prepareDeployProtocolConfigFromCanonical --snapshot <artifact.json>` — deploys the implementation and prints the `upgradeToAndCall(initializeFromMigration(...))` payload without mutating the proxy, mirroring `task:prepareDeployProtocolConfigFromMigration`.
- Both run the **same prepare step** (`buildCanonicalUpgradeProposal`: deploy implementation + build the `upgradeToAndCall(initializeFromMigration(…))` payload); the devnet task then simply executes the produced payload with the deployer key (`executeUpgradeProposal`) — **devnet runs byte-identical calldata to what the DAO signs**, so every devnet deploy exercises the DAO payload end to end. The replica lands on canonical's `currentKmsContextId` (empty-proxy deployment idiom, not an upgrade of a live ProtocolConfig).
- Shared `buildSnapshotArtifact`/`parseSnapshotArtifact` keep export and both apply paths on one validated JSON shape.

### e2e: fhevm-cli seeds non-canonical chains like production

- `task:deployAllHostContracts` gains `--protocol-config-source canonical` (+ canonical RPC/address params), running the mirror in sequence with the other host contracts.
- The fhevm-cli multi-chain up flow patches each non-canonical chain's deploy args with the canonical chain's RPC + ProtocolConfig address (available on disk after the canonical deploy, which the flow already sequences first) — so the e2e stack reproduces production topology instead of seeding every chain "fresh", and every multi-chain e2e run executes the byte-identical payload the DAO would sign.
- Gated by a 0.13.1 compat floor (`supportsCanonicalProtocolConfigSeeding`): compat bundles with older host images keep the fresh seeding.

## Non-goals

- No canonical/secondary deploy-task split — explicitly out of scope per the issue.
- No deletion of the Gateway-migration tasks: `task:exportKmsMigrationState` + `FromMigration` tasks remain the mechanism for the one-time Gateway → Ethereum migration of existing 0.12 deployments; they are only superseded as the **seeding source** for new non-canonical chains (README clarified accordingly).
- No contract changes.

## Notes for reviewers

- Tests cover: the read→apply flow onto a fresh proxy, block pinning under rotation, no-active-context and identity-check failures on the shared read path, artifact round-trip through both apply paths (DAO calldata decode + devnet apply), and the deploy task's arg validation.
- The issue's destroyed-context failure mode is guarded in `readCanonicalSnapshot` but untestable on-chain: `destroyKmsContext` reverts for the current context (`CurrentKmsContextCannotBeDestroyed`), so the guard is defense-in-depth — same as on `main` (#2657).
- Both apply tasks and the export compile first (`compile:specific contracts`): `ProtocolConfig` embeds `aclAdd` at compile time, so a stale artifact could deploy bytecode authorized against the wrong ACL.
- `npm run lint:ts` fails on the base branch independently of this PR (missing `.eslintignore`).

## Validation

- `DOTENV_CONFIG_PATH=.env.example npx hardhat test test/tasks/*.ts --network hardhat` (37 passing across migration + taskDeploy + ownership/pausing suites)
- `npx prettier --check` on all changed files

### DIFF
--- a/host-contracts/tasks/protocolConfigMirror.ts
+++ b/host-contracts/tasks/protocolConfigMirror.ts
@@ -4,7 +4,7 @@ import type { HardhatRuntimeEnvironment } from 'hardhat/types';
 import type { ProtocolConfig } from '../types';
 import { assertContractMatchesVersionPrefix } from './utils/contractVersion';
 import { formatError } from './utils/formatError';
-import { type UpgradeProposal, buildUpgradeProposal, getFunctionFragment } from './utils/upgradeProposal';
+import { type UpgradeProposal, buildUpgradeProposal } from './utils/upgradeProposal';
 
 export type KmsNode = {
   txSenderAddress: string;
@@ -117,12 +117,10 @@ export async function buildCanonicalUpgradeProposal(
     `Mirroring ProtocolConfig from canonical chain ${snapshot.canonicalChainId} at block ${snapshot.blockNumber} (${snapshot.blockHash}): contextId=${snapshot.currentKmsContextId}, kmsNodes=${snapshot.kmsNodes.length}.`,
   );
 
-  const artifact = await hre.artifacts.readArtifact('ProtocolConfig');
-  const innerFunctionSignature = getFunctionFragment(artifact.abi, 'initializeFromMigration').format('sighash');
   return buildUpgradeProposal(hre, {
     proxyAddress,
     contractName: 'ProtocolConfig',
-    innerFunctionSignature,
+    innerFunctionName: 'initializeFromMigration',
     decodedArgs: [snapshot.currentKmsContextId, snapshot.kmsNodes, snapshot.thresholds],
   });
 }
--- a/host-contracts/tasks/protocolConfigMirror.ts
+++ b/host-contracts/tasks/protocolConfigMirror.ts
@@ -5,11 +5,6 @@ import type { ProtocolConfig } from '../types';
 import { formatError } from './utils/formatError';
 import { getRequiredEnvVar } from './utils/loadVariables';
 
-export type SecondaryDeployArgs = {
-  canonicalRpcUrl: string;
-  canonicalProtocolConfigAddress: string;
-};
-
 export type KmsNode = {
   txSenderAddress: string;
   signerAddress: string;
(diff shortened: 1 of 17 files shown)


PanGan21:
Should we add ProtocolConfig to the upgrade manifest so reinitializers check is done on it as well?

PanGan21 (dismissed):

host-contracts/tasks/protocolConfigMirror.ts
```diff
@@ -0,0 +1,202 @@
...
+  const { ethers } = hre;
+  const { canonicalProvider, canonicalProtocolConfigAddress } = options;
+
+  // Handshake before the identity check so a dead or mistyped RPC URL is reported as an RPC
+  // problem, not as a contract identity failure.
+  let canonicalChainId: bigint;
+  let blockNumber: number;
+  try {
+    canonicalChainId = (await canonicalProvider.getNetwork()).chainId;
+    blockNumber = options.blockNumber ?? (await canonicalProvider.getBlockNumber());
```
obatirou: could you take a finalized block for canonicalProvider.getBlockNumber()?

host-contracts/tasks/protocolConfigMirror.ts
```diff
@@ -0,0 +1,202 @@
...
+  ]);
+  const kmsNodes: KmsNode[] = rawNodes.map((node) => ({
+    txSenderAddress: node.txSenderAddress,
+    signerAddress: node.signerAddress,
+    ipAddress: node.ipAddress,
+    storageUrl: node.storageUrl,
+  }));
+  const thresholds: KmsThresholds = { publicDecryption, userDecryption, kmsGen, mpc };
+
+  return { currentKmsContextId, kmsNodes, thresholds, canonicalChainId, blockNumber };
```
obatirou: block hash should be returned also

host-contracts/tasks/protocolConfigMirror.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  let canonicalVersion: string;
+  try {
+    canonicalVersion = await canonicalProtocolConfig.getVersion();
+  } catch (err) {
+    throw new Error(
+      `Canonical ProtocolConfig identity check failed: contract at ${canonicalProtocolConfigAddress} does not expose getVersion() (${formatError(err)}).`,
+    );
+  }
+  if (!/^ProtocolConfig v\d/.test(canonicalVersion)) {
```
obatirou: already exists with `assertContractMatchesVersionPrefix`

host-contracts/tasks/protocolConfigMirror.ts
```diff
@@ -0,0 +1,202 @@
...
+    throw new Error(
+      `Snapshot artifact field "blockNumber" must be a number, got ${JSON.stringify(artifact.blockNumber)}.`,
+    );
+  }
+  if (!Array.isArray(artifact.kmsNodes) || artifact.kmsNodes.length === 0) {
+    throw new Error('Snapshot artifact field "kmsNodes" must be a non-empty array.');
+  }
+  for (const [index, node] of artifact.kmsNodes.entries()) {
+    for (const field of ['txSenderAddress', 'signerAddress', 'ipAddress', 'storageUrl'] as const) {
+      if (typeof node?.[field] !== 'string') {
```
obatirou, line 200: Should address format be validated?

obatirou (approved):

PR #2775, opened 2026-06-12
feat(kms-connector): epoch_id validation

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/1467

## What

Handles the validation of the `epoch_id` embedded in the `extra_data` of several requests before forwarding to the kms-core.

### gw-listener
- `epoch_id` column added to the `kms_context` DB table (primary key is now `(context_id, epoch_id)`)
- at `gw-listener` startup, this table is filled via on-chain view function call to retrieve the current context/epoch
- `gw-listener` now monitor `KmsContextDestroyed` events to update this table as well

### kms-worker
- check the `kms_context` DB table to check context and epoch validity
- if invalid, fallback to an on-chain view function call for a potential DB update
  - should only happen once per context switch, so performance of the hot path should be fine

## Notes
- Epoch expiration/cleanup is intentionally out of scope (RFC-005): context destruction is the only invalidation path for now
- `Option` on `ExtraData::context_id`/`epoch_id` stays until relayer-sdk 0.4.2 (v1 extra_data) is deprecated.


kms-connector/connector-db/migrations/20260526120000_add_protocol_config_events.sql
```diff
@@ -38,6 +38,12 @@ ON CONFLICT DO NOTHING;
 -- Persist the active epoch alongside the active context fetched at startup
 -- via KMSVerifier::getActiveKmsContextAndEpoch().
 ALTER TABLE kms_context ADD COLUMN IF NOT EXISTS epoch_id BYTEA;
+-- Use the default epoch ID (little endian representation) for all existing contexts
+UPDATE kms_context SET epoch_id = '\x0100000000000000000000000000000000000000000000000000000000000008'::bytea WHERE epoch_id IS NULL;
+ALTER TABLE kms_context ALTER COLUMN epoch_id SET NOT NULL;
+ALTER TABLE kms_context
+    DROP CONSTRAINT kms_context_pkey,
+    ADD PRIMARY KEY (id, epoch_id);
```
dvdplm, line 46: Shouldn't these be wrapped in a transaction so that they either all succeed or all fail? `BEGIN ……… COMMIT` or some such?

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -450,3 +450,61 @@ async fn publish_new_kms_epoch<'e>(
...
+    ///
+    /// Boxed to keep the enum small: `ProtocolEvent` is an order of magnitude bigger than
+    /// the other variant (`clippy::large_enum_variant`).
+    StoreEvent(Box<ProtocolEvent>),
+    /// Invalidate all the epochs of a destroyed KMS context in the `kms_context` table.
+    ///
+    /// Context destruction is the only invalidation that can ever happen: once active, an
+    /// epoch stays active on-chain until its context is destroyed (RFC-005 leaves epoch
+    /// expiration/cleanup out of scope).
+    InvalidateContext(U256),
```
jot2re, line 467: Just out of curiosity why is it called `InvalidateContext` and not `DestroyContext`?

kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -36,7 +37,8 @@ pub struct ExtraData {
 /// - Bytes 33..65: epoch ID (32 bytes, big-endian U256)
 /// - Bytes 65..: optional additional data (ignored)
 ///
-/// Version `0x01` → epoch_id is `None` (caller should fall back to DEFAULT_EPOCH_ID).
+/// Version `0x01` → epoch_id is `None` (downstream consumers fall back to
+/// [`crate::types::DEFAULT_EPOCH_ID`]).
 /// Empty or `[0x00]` → no explicit context and no epoch.
```
dvdplm: It looks like the case of empty/[0x0] also means epoch_id and context_id are `None`. Docs should reflect that.

kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -36,7 +37,8 @@ pub struct ExtraData {
 /// - Bytes 33..65: epoch ID (32 bytes, big-endian U256)
 /// - Bytes 65..: optional additional data (ignored)
 ///
-/// Version `0x01` → epoch_id is `None` (caller should fall back to DEFAULT_EPOCH_ID).
+/// Version `0x01` → epoch_id is `None` (downstream consumers fall back to
```
dvdplm: The context id is also `None` I think.

kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -36,7 +37,8 @@ pub struct ExtraData {
...
+/// [`crate::types::DEFAULT_EPOCH_ID`]).
 /// Empty or `[0x00]` → no explicit context and no epoch.
 pub fn parse_extra_data(extra_data: &[u8]) -> anyhow::Result<ExtraData> {
     if extra_data.is_empty() || extra_data == [0x00] {
```
dvdplm: `extra_data == [0x00]` means that `[0x00DEADBEEF]` would not match – maybe that's intentional but otherwise `extra_data[0] == [0x00]` would be better?

kms-connector/crates/utils/src/tests/setup/db.rs
```diff
@@ -75,4 +77,3 @@ impl DbInstance {
 }
 
 pub const TESTING_KMS_CONTEXT: U256 = U256::ONE;
```
dvdplm: Maybe this guy could move to `crate::types` as well (and get a line of docs?)?

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
```diff
@@ -1,49 +1,169 @@
...
-            None => Err(ProcessingError::Recoverable(anyhow!(
-                "Context #{context_id} not found in DB"
-            ))),
-            Some(false) => Err(ProcessingError::Irrecoverable(anyhow!(
-                "Context #{context_id} is invalid"
-            ))),
-            Some(true) => Ok(()),
+        // `is_valid = false` is only ever written by `KmsContextDestroyed`, which invalidates the
+        // whole context at once.
+        if rows.iter().any(|r| !r.is_valid) {
```
jot2re, line 84: I am probably misunderstanding something here, but isn't `rows` the records of epoch_id and is_valid from the db? Won' this mean that if any epoch is invalid for `context_id` then we will return that the context has been destroyed? 
What about the situation where we just destroy an epoch and not the whole context?

kms-connector/crates/utils/src/types/extra_data.rs
```diff
@@ -16,6 +16,7 @@ pub const EXTRA_DATA_V2_LENGTH: usize = 65;
 /// Parsed extra_data contents.
 #[derive(Debug, Clone, PartialEq)]
 pub struct ExtraData {
+    // TODO: Remove `Option` as soon as the relayer-sdk 0.4.2 is deprecated.
```
jot2re, line 22: Should this be linked to an existing issue to make sure it is tracked?

jot2re (commented):
I just have a shallow look since I know David is also reviewing and since I am not super familiar with the Connector code base. 
So I just left a few comments

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
```diff
@@ -1,49 +1,169 @@
...
+            .protocol_config_contract
+            .isValidKmsContext(context_id)
+            .call()
+            .await
+            .map_err(|e| {
+                ProcessingError::Recoverable(anyhow!(
+                    "isValidKmsContext(#{context_id}) call failed: {e}"
+                ))
+            })?;
+        if !context_valid {
```
dvdplm, line 118: If the contract says that the context is valid it seems like we're ok with that. Do you think it would be overkill to check the context id value against what we know locally? E.g. Is it non-zero, non-empty, not already in use, not a very old context id etc?
Maybe I'm being overly cautious but data coming from an RPC endpoint that we do not control is **not** to be trusted blindly. It's data coming from some guy's web server on wild&dangerous internet...

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
```diff
@@ -1,49 +1,169 @@
...
+
+        let epoch_valid = self
+            .protocol_config_contract
+            .isValidEpochForContext(context_id, epoch_id)
+            .call()
+            .await
+            .map_err(|e| {
+                ProcessingError::Recoverable(anyhow!(
+                    "isValidEpochForContext(#{context_id}, #{epoch_id}) call failed: {e}"
+                ))
```
dvdplm, line 137: How do you know that this is recoverable?

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
```diff
@@ -1,49 +1,169 @@
...
+            true,
+            now,
+            now,
+        )
+        .execute(&self.db_pool)
+        .await;
+
+        match query_result {
+            Ok(_) => info!("Context cached as valid in DB"),
+            Err(e) => warn!("Failed to cache context: {e}"),
```
dvdplm, line 166: Why not `error!` here?

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
```diff
@@ -1,49 +1,169 @@
...
+        }
+
+        let Some(epoch_id) = epoch_id else {
+            // v1 extra_data: no epoch to validate, and nothing to cache without one
+            return Ok(());
+        };
+
+        let epoch_valid = self
+            .protocol_config_contract
+            .isValidEpochForContext(context_id, epoch_id)
```
dvdplm, line 131: I would guess this is not reasonable to ask for in this PR but I think it would be nicer if the contract could repond to `isValidKmsContext` with a list of valid epoch ids in that context (should be a small number yeah). That way we'd have a single RPC call to make.

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -450,3 +450,61 @@ async fn publish_new_kms_epoch<'e>(
     .await
     .map_err(anyhow::Error::from)
 }
+
+/// Action derived from a decoded Ethereum log.
+#[derive(Clone, Debug, PartialEq)]
+pub enum EthereumEventAction {
+    /// Store the event in DB, for the kms-worker to forward it to the KMS Core.
```
dvdplm:
```suggestion
    /// Store the event in the DB, for the kms-worker to forward it to the KMS Core.
```

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -450,3 +450,61 @@ async fn publish_new_kms_epoch<'e>(
...
+    ///
+    /// Boxed to keep the enum small: `ProtocolEvent` is an order of magnitude bigger than
+    /// the other variant (`clippy::large_enum_variant`).
```
dvdplm:
```suggestion
    // Boxed to keep the enum small: `ProtocolEvent` is an order of magnitude bigger than
    // the other variant.
```

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: I guess in DB, we just set `is_valid` to `false`, so the action we perform at kms-connector level is invalidation, not destruction, even if the event that triggers this is `KmsContextDestroyed`.

But I could rename it if needed, no strong opinion on this

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: It's stated in the PR description that:

> Epoch expiration/cleanup is intentionally out of scope (RFC-005): context destruction is the only invalidation path for now

Maybe we would need another RFC for this?

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: Yes! Will do

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -450,3 +450,61 @@ async fn publish_new_kms_epoch<'e>(
...
+                invalidate_kms_context(&mut *tx, context_id).await?
+            }
+        }
+    }
+    update_last_block_polled(&mut *tx, ChainName::Ethereum, Some(block_number)).await?;
+    tx.commit().await?;
+    Ok(())
+}
+
+/// Marks all the epochs of a destroyed KMS context as invalid.
```
dvdplm: The docs look wrong: the code is marking a single context as invalid, but it's not touching the epoch ids. I guess it might amount to the same thing, but worth pointing out perhaps: "Mark a destroyed KMS context as invalid; this implies that all associated epoch IDs also become invalid." or something like that?

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -450,3 +450,61 @@ async fn publish_new_kms_epoch<'e>(
...
+    db_pool: &Pool<Postgres>,
+    actions: Vec<EthereumEventAction>,
+    block_number: u64,
+) -> anyhow::Result<()> {
+    let mut tx = db_pool.begin().await?;
+    for action in actions {
+        match action {
+            EthereumEventAction::StoreEvent(event) => publish_event_inner(&mut *tx, *event).await?,
+            EthereumEventAction::InvalidateContext(context_id) => {
+                invalidate_kms_context(&mut *tx, context_id).await?
```
dvdplm, line 483: Is there some sort of check or guarantee that this method only ever invalidate a single context? I would assume that it is **never** legit to invalidate more than one context at a time?

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -450,3 +450,61 @@ async fn publish_new_kms_epoch<'e>(
...
+    Ok(())
+}
+
+/// Marks all the epochs of a destroyed KMS context as invalid.
+async fn invalidate_kms_context<'e>(
+    executor: impl PgExecutor<'e>,
+    context_id: U256,
+) -> anyhow::Result<()> {
+    let query_result = sqlx::query!(
+        "UPDATE kms_context SET is_valid = FALSE, updated_at = $2 WHERE id = $1",
```
dvdplm, line 499: Do you think it's worthwhile to add a condition to the `WHERE` clause to ensure only *active* contexts can be invalidated? Like so `WHERE id = $1 AND is_valid = TRUE`?
And what do you think of a `LIMIT 1` to ensure we never blanket invalidate contexts?

dvdplm: What are the consequences of not having an active context at all? I think that would be very bad yes? Is there some neat SQL to make this query only do the update if there is a different context ID in the DB that is valid? Like "invalidate this context, but only if there exists another one that is valid"?

kms-connector/connector-db/migrations/20260526120000_add_protocol_config_events.sql
eudelins-zama: AFAIU, the default behavior of `sqlx-cli` is that each migration file is ran in a transaction

Trying to find a piece of docs confirming this

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: Not in `0x01` no, only in `0x00`

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: I think we should accept `0x00DEADBEEF` indeed

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: It's a network error, or at least an error that made our RPC call failed.
Whatever it is, I think we want to retry in such case

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: We only check the contract if we have not already something registered locally (it's a fallback), so it would not make sense IMO

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: In the kms-connector, we tend to use `error!` for things that is blocking the flow to continue.

For ex, if the kms-core is down, we can't process decryption and we would use `error!`.

But here, if we fail to cache the context at the `kms-worker` level, then we would still be able to process the decryption.
The next decryption attached to that context id would then also (re-)try to cache that context, and would not be blocked either.

Makes sense to you? :slightly_smiling_face:

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: Already asked in this [thread](https://zama-ai.slack.com/archives/C0A8BJPD5UG/p1779891384281569), but seems not that obvious at the contract level unfortunately :face_with_diagonal_mouth:

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: > Do you think it's worthwhile to add a condition to the WHERE clause to ensure only active contexts can be invalidated? Like so WHERE id = $1 AND is_valid = TRUE?

I'm not sure to understand the benefit of this compared to the actual implementation tbh.

> And what do you think of a LIMIT 1 to ensure we never blanket invalidate contexts?

This I disagree, because if we have context `1` that is valid, with epoch `1` and `2` also valid, then if context `1` is destroyed, we want to invalidate both `(1, 1)` and `(1, 2)`, so `LIMIT 1` would not make sense here.
Unless I'm missing something :sweat_smile:

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Agreed, but aside carefully reviewing the DB query (which takes a single `context_id` as input), I don't see what we can do about it...

See related comment [here](https://github.com/zama-ai/fhevm/pull/2775#discussion_r3414585729)

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
dvdplm: So this is what I have in mind:

- ethereum/polygon is working as advertised
- we have a bunch of contexts in the system already, say 25
- our RPC provider is all good too, doing honest work
- …but there's a man-in-the-middle spoofing our RPC requests and feeding us false data (note: this is just hacking or spoofing a webserver, much easier than breaking ethereum)

What can happen if our RPC calls receive false responses?

Can they fool us to invalidate the wrong context (and all epochs)? If yes, what can we do? Maybe some cheap local sanity checks are worth the effort? As it stands I think that we are considering "RPC response" == "What is on-chain" and that is a sketchy assumption, especially if we trust Infura or some other third party.

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
dvdplm: That should be up to the caller to decide imo. It'd be cleaner to just bubble up the error here and let the retry logic decide which types of errors should be retried.

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
dvdplm: Ok, fair enough. I guess I could flip the argument on you and ask "If the DB can't complete a query, it's pretty important to know about it asap.", but you know best, perhaps there are other mechanisms in place to monitor the DB health?

kms-connector/connector-db/migrations/20260526120000_add_protocol_config_events.sql
eudelins-zama: ### Does `sqlx migrate run` (sqlx-cli) wrap each migration in a transaction?

Yes — **each migration file runs in its own transaction** (one `begin`/`commit` per file, not one big transaction for the whole run).

Code path proving it for the CLI specifically:

1. `sqlx migrate run` → [`sqlx-cli/src/migrate.rs::run()`](https://github.com/launchbadge/sqlx/blob/main/sqlx-cli/src/migrate.rs) loops over each migration and calls `conn.apply(table, migration)` per file.
2. That bottoms out in the per-DB `apply()` impl, e.g. [`sqlx-postgres/src/migrate.rs`](https://github.com/launchbadge/sqlx/blob/main/sqlx-postgres/src/migrate.rs):
   ```rust
   if migration.no_tx {
       execute_migration(self, table_name, migration).await?;
   } else {
       let mut tx = self.begin().await?;   // transaction per migration
       execute_migration(&mut tx, table_name, migration).await?;
       tx.commit().await?;
   }
   ```
  - `no_tx` default value is `false` (see [here](https://github.com/transact-rs/sqlx/blob/main/sqlx-core/src/migrate/migrator.rs#L35-L42))

So if a migration fails, only that file rolls back and `migrate run` stops there.

kms-connector/crates/gw-listener/src/core/publish.rs
dvdplm: > I'm not sure to understand the benefit of this compared to the actual implementation tbh.

It's been 15h and I don't see the benefit either. Brainfart, pls ignore.

> 1 that is valid, with epoch 1 and 2 also valid, then if context 1 is destroyed, we want to invalidate both (1, 1) and (1, 2), so LIMIT 1 would not make sense here.

I missed that epoch IDs were stored in the same table; I thought they were in their own table. In that case, limiting to 1 row is bad of course.

Fair enough.

kms-connector/connector-db/migrations/20260526120000_add_protocol_config_events.sql
dvdplm: Perfect, thank you for checking!

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: Yes we have a healthcheck that queries the DB regularly :slightly_smiling_face:

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: Fixed 5e457e178efb6af6f1df3c8dec7021b25b816f60

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: Fixed 5e457e178efb6af6f1df3c8dec7021b25b816f60

kms-connector/crates/utils/src/types/extra_data.rs
eudelins-zama: Fixed 16ce6899f967571e1723a364298fafb4f0298f3c

kms-connector/crates/utils/src/tests/setup/db.rs
eudelins-zama: Fixed 5e457e178efb6af6f1df3c8dec7021b25b816f60

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: Let's have a discussion on this outside this PR :innocent:

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: Let's discuss the RPC topic internally, but keep it out of the scope of the PR :slightly_smiling_face:

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Fixed 5e457e178efb6af6f1df3c8dec7021b25b816f60

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Fixed 5e457e178efb6af6f1df3c8dec7021b25b816f60

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Fixed 5e457e178efb6af6f1df3c8dec7021b25b816f60

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
jot2re: Yes, you are right, destroying an epoch is currently not part of the RFC. I do believe we should indeed make another RFC to handle this. Because the KMS will keep accepting old epochs if they are not destroyed. 
While we likely won't really make a lot of new epochs within the same context I don't think it is a big issue that we cannot destroy them individually. 
But I think that one thing we should do at the KMS is to destroy epochs associated with a context once it gets destroyed. I have [made an issue for that here.](https://github.com/zama-ai/kms-internal/issues/3048)

If you agree, I will make a new tech spec issue and post in slack about making a new RFC for handling individual destruction of epochs

kms-connector/crates/gw-listener/src/core/publish.rs
jot2re: No, no, it is fine. I just wanted to make sure I understood correctly.
FYI be aware that destruction on the KMS core side means actual deletion (for security reasons)

jot2re (approved):
Still looks good to me! 
Just one open comment about a potential follow up

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: I was not aware that the KMS need to destroy context on its side. (not well defined in RFC-005)
Currently, the kms-connector does not forward the context destruction event to the kms-core, it only updates its local DB.
So maybe let's keep it as it is right now, and let's do a proper RFC for context/epoch destruction?

dvdplm (approved):
LGTM!

PR #2776, opened 2026-06-12
fix(common): scope upgrade version check to PR-changed contracts and enforce strict release run

PanGan21:
### Summary

- Fixes PRs being blocked by upgrade-check failures they didn't cause: pre-existing drift on main after a release moves the baseline, or forward ports of fixes already in the tag.
- `check.ts` gains a `--changed-contracts` flag: on PRs, only contracts whose source the PR touched fail the check; violations on other contracts are non-blocking warnings.
- New `release: published` trigger runs the full strict check on the released tag against the previous release in its version line (e.g. `v0.13.1` → `v0.13.0` even if `v0.14.0` exists), so a missed bump surfaces at release time. Forward-ported fixes exist on both sides of that comparison, so they don't trigger it.

| Event             | Check runs? | Enforcement                                                       |
| ----------------- | ----------- | ----------------------------------------------------------------- |
| Pull request      | Yes         | Errors only for contracts the PR changed; warnings for the rest    |
| Release published | Yes         | Strict — all contracts, against the previous release in the version line |

### DIFF
--- a/.github/workflows/contracts-upgrade-version-check.yml
+++ b/.github/workflows/contracts-upgrade-version-check.yml
@@ -94,13 +94,15 @@ jobs:
           RELEASED_TAG: ${{ github.event.release.tag_name }}
         run: |
           if [ "$GITHUB_EVENT_NAME" = "release" ]; then
-            # Previous release in the published tag's version line, so a patch
-            # release compares against its own line (v0.13.1 -> v0.13.0 even
-            # when v0.14.0 exists).
+            # Highest stable release strictly below the published version, in
+            # its own version line: v0.13.1 -> v0.13.0 even when v0.14.0
+            # exists.  Prerelease suffixes are stripped first, so a prerelease
+            # compares like the version it precedes: v0.13.0-9 -> v0.12.5.
+            base="${RELEASED_TAG%%-*}"
             tag="$({ gh api "repos/${GH_REPOSITORY}/releases" --paginate \
               --jq '.[] | select((.draft or .prerelease) | not) | .tag_name'; \
-              echo "$RELEASED_TAG"; } | sort -u -V | grep -x -B1 -- "$RELEASED_TAG" | head -n1)"
-            if [ -z "$tag" ] || [ "$tag" = "$RELEASED_TAG" ]; then
+              echo "$base"; } | sort -u -V | grep -x -B1 -- "$base" | head -n1)"
+            if [ -z "$tag" ] || [ "$tag" = "$base" ]; then
               echo "::error::No previous release found to compare $RELEASED_TAG against"
               exit 1
             fi
--- a/.github/workflows/contracts-upgrade-version-check.yml
+++ b/.github/workflows/contracts-upgrade-version-check.yml
@@ -4,14 +4,18 @@ permissions: {}
 
 on:
   pull_request:
+  release:
+    types: [published]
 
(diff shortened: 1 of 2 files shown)


PanGan21:
@claude  /pr-review

obatirou (dismissed):

PanGan21:
@mergify queue

.github/workflows/contracts-upgrade-version-check.yml
```diff
@@ -44,11 +53,12 @@ jobs:
   check:
     name: contracts-upgrade-version-check/${{ matrix.package }} (bpr)
     needs: check-changes
+    if: ${{ !cancelled() && needs.check-changes.result != 'failure' }}
     permissions:
       contents: "read" # Required to checkout repository code
     runs-on: runs-on=${{ github.run_id }}/runner=4cpu-linux-x64
     env:
-      PACKAGE_CHANGED: ${{ contains(fromJSON(needs.check-changes.outputs.packages), matrix.package) }}
+      PACKAGE_CHANGED: ${{ github.event_name != 'pull_request' || contains(fromJSON(needs.check-changes.outputs.packages || '[]'), matrix.package) }}
```
obatirou: Not sure to understand after a second look

PanGan21: This makes the check still run on non pr events (e.g. release). `check-changes` only runs on prs, and  skips dependent jobs when a dependency is skipped. Do you disagree with this logic?

PanGan21: However I made it more robust on latest commit and updated the description to be more clear on the intention

.github/workflows/contracts-upgrade-version-check.yml
```diff
@@ -77,8 +91,24 @@ jobs:
...
+            tag="$({ gh api "repos/${GH_REPOSITORY}/releases" --paginate \
+              --jq '.[] | select((.draft or .prerelease) | not) | .tag_name'; \
+              echo "$base"; } | sort -u -V | grep -x -B1 -- "$base" | head -n1)"
+            if [ -z "$tag" ] || [ "$tag" = "$base" ]; then
+              echo "::error::No previous release found to compare $RELEASED_TAG against"
+              exit 1
+            fi
+          else
+            tag="$(gh api "repos/${GH_REPOSITORY}/releases/latest" --jq .tag_name)"
+          fi
```
PanGan21, line 111: This requires some attention on the baseline tag because it can potentially block releases. So giving some examples on the logic.

When creating:
- v0.13.0 -> will be checked against v0.12.5
- v0.13.1 -> will be checked against v0.13.0
- v0.14.0 -> will be checked against v0.13.x (highest)
- v0.13.0-9 -> will be checked against  v0.12.5
- v0.13.1-1 -> will be checked against v0.13.0

That means patch releases which change the contracts also require reinitializers bump. Let's review this carefully since it can become a blocker in case it is not correct.

obatirou (approved):

.github/workflows/contracts-upgrade-version-check.yml
PanGan21: cc @Eikix @obatirou for an opinion on this

PanGan21:
@mergify queue

PanGan21:
@mergify queue

PR #2782, opened 2026-06-15
test(sdk): dod script and localstack tests fixes

geoxel:
### `005483d` — fix(test-suite): fix localstack v13

Root cause: `v0.13.0-2` Docker images for `db-migration` were built before the commit that added `seed_host_chains()` to `initialize_db.sh`. The `host_chains` table was created by migrations but never populated, causing the zkproof worker to reject every proof with _"Unknown chain ID: 12345"_.

**Fixes:**
- `compat.ts`: raised the self-seeding threshold from `[0, 13, 0]` to `[0, 13, 1]` so all `v0.13.0-N` builds trigger the manual shim
- `compat.test.ts`: updated `v0.13.0-6` test to expect `true` (needs shim), added a `v0.13.1` test expecting `false` (self-seeds)
- Running stack: seeded `host_chains` directly in Postgres and restarted `coprocessor-zkproof-worker`

For future stack restarts, the `requiresLegacyHostChainSeedShim` fix ensures `fhevm-cli up` will seed the table automatically for any `v0.13.0-N` profile.

---

### `fb9d6cd` — test(sdk): fix fheTests for localstack v12

`initAliceFheTestHandlesIfNeeded` calls `initFheTest(false)` synchronously and returns. The v12 coprocessor then asynchronously registers Alice's FHE handles in the gateway's `CiphertextCommits` contract. The `clientBase.decryptPublicValue` test started immediately, submitted a public decrypt request to the relayer, and the relayer's `isPublicDecryptionReady` check (calling `CiphertextCommits.isCiphertextMaterialAdded`) failed because the coprocessor hadn't finished yet — causing the test to hit its 120-second timeout.

**Fix:** Added a 30-second `Atomics.wait` blocking delay in `setupCommon.ts:328` after fresh handle initialization, giving the coprocessor time to register the handles. The delay only fires when handles are actually initialized (not already present), so repeated test runs don't pay the penalty.

---

### `9cb3a53` — test(sdk): fix NonceManager for localstack v12

Fixes _"Error: nonce has already been used"_ when running `ethers/clientEncrypt.encryptDecrypt.slow.test.ts`.

The ethers.js `NonceManager.sendTransaction` fires an async nonce query (A), then `populateTransaction` independently queries `eth_getTransactionCount` (B) and passes that nonce to `eth_estimateGas`. If the v12 coprocessor increments Alice's nonce between A and B, `eth_estimateGas` receives a stale nonce and fails.

**Fix (`FixedNonceManager`):** Await the NonceManager's nonce first, then spread it into the tx before calling `populateTransaction`. Since `tx.nonce` is pre-set, `populateTransaction` skips its own `eth_getTransactionCount` call entirely — `eth_estimateGas` runs with the correct, authoritative nonce.

---

### `086601e` — test(sdk): dod script

Added a DoD (Definition of Done) test script for the SDK.

### DIFF
--- a/sdk/js-sdk/test/scripts/dod.mjs
+++ b/sdk/js-sdk/test/scripts/dod.mjs
@@ -1,0 +1,87 @@
+import { spawnSync } from 'node:child_process';
+import { fileURLToPath } from 'node:url';
+import { resolve, dirname } from 'node:path';
+
+const sdkRoot = resolve(dirname(fileURLToPath(import.meta.url)), '../..');
+
+function formatElapsed(ms) {
+  const s = Math.floor(ms / 1000);
+  const m = Math.floor(s / 60);
+  return m > 0 ? `${m}m${String(s % 60).padStart(2, '0')}s` : `${s}s`;
+}
+
+function banner(cmd, index, total, startTime) {
+  const counter = `${index + 1}/${total}`;
+  const elapsed = formatElapsed(Date.now() - startTime);
+  const pad = 2;
+  const minGap = 2;
+  const innerWidth = Math.max(80 - pad * 2, cmd.length + counter.length + minGap, elapsed.length);
+  const width = innerWidth + pad * 2;
+  const line = '═'.repeat(width);
+  const inner = cmd.padEnd(innerWidth);
+  const cmdLine = cmd + counter.padStart(innerWidth - cmd.length);
+  const elapsedLine = elapsed.padStart(innerWidth);
+  console.log(`╔${line}╗`);
+  console.log(`║${''.padEnd(pad)}${cmdLine}${''.padEnd(pad)}║`);
+  console.log(`║${''.padEnd(pad)}${elapsedLine}${''.padEnd(pad)}║`);
+  console.log(`╚${line}╝`);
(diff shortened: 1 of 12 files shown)


sdk/js-sdk/test/scripts/dod.mjs
```diff
@@ -0,0 +1,87 @@
...
+  './test/scripts/localcleartext-run-tests.sh --use-pack',
+  //'node test/multi-wasm/run.mjs',
+];
+
+const longCommands = [
+  'npm run test:full:testnet',
+  'npm run test:full:devnet',
+  'npm run test:localstack:v11',
+  'npm run test:localstack:v12',
+  'npm run test:localstack:v13',
```
poppyseedDev, line 53: for future when we have multiple different versions maybe passing v12 v13 as an argument could make more sense, but for now it's fine

poppyseedDev (approved):
lgtm

sdk/js-sdk/test/scripts/dod.mjs
geoxel: Yes!

PR #2783, opened 2026-06-15
chore(kms-connector): improve tests compilation

eudelins-zama:
### What

Reduces memory consumption and build time when compiling the kms-connector tests.

`cargo test` builds ~20 separate test binaries, each statically linking the full dependency graph (tfhe, alloy, tonic, sqlx…).
With the default profile this generates full debug info on every binary, making link steps memory-hungry enough to freeze machines.

We can use a smaller machine in CI to run tests thanks to this.

### How

- `[profile.dev]` / `[profile.test]` `debug = "line-tables-only"`: keeps usable backtraces while cutting binary size and link-time memory
- `tfhe` `default-features = false`: minor cleanup; tfhe's feature set is driven by `kms-grpc` via feature unification anyway

### Notes

Spotted a small bug in `rust-sdk` CI after this [commit](https://github.com/zama-ai/fhevm/commit/88b11de0d67a096ef5e0659eb8106bb691acf763#diff-e77ddbc63a771942279b53f47597c04f14abc4c8b816af1cdbcb68d81d5b2f1f): tests were not skipped even if no changes to the sdk were made. So fixed it as well.

### DIFF
--- a/kms-connector/Cargo.toml
+++ b/kms-connector/Cargo.toml
@@ -22,7 +22,7 @@ fhevm_gateway_bindings = { path = "../gateway-contracts/rust_bindings", default-
 fhevm_host_bindings = { path = "../host-contracts/rust_bindings", default-features = false }
 kms-grpc = { git = "https://github.com/zama-ai/kms.git", tag = "v0.13.20-0", default-features = true }
 bc2wrap = { git = "https://github.com/zama-ai/kms.git", tag = "v0.13.20-0", default-features = true }
-tfhe = "=1.6.1"
+tfhe = { version = "=1.6.1", default-features = false }
 user-decryption-signature = { path = "../shared/user-decryption-signature", default-features = false }
 
 #####################################################################
@@ -113,3 +113,10 @@ inherits = "release"
 opt-level = 1
 lto = false
 codegen-units = 16
+
+[profile.dev]
+debug = "line-tables-only"
+
+[profile.test]
+debug = "line-tables-only"
+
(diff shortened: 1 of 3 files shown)


fegmorte (approved):

kms-connector/Cargo.toml
```diff
@@ -22,7 +22,7 @@ fhevm_gateway_bindings = { path = "../gateway-contracts/rust_bindings", default-
 fhevm_host_bindings = { path = "../host-contracts/rust_bindings", default-features = false }
 kms-grpc = { git = "https://github.com/zama-ai/kms.git", tag = "v0.13.20-0", default-features = true }
 bc2wrap = { git = "https://github.com/zama-ai/kms.git", tag = "v0.13.20-0", default-features = true }
-tfhe = "=1.6.1"
+tfhe = { version = "=1.6.1", default-features = false }
```
dd23, line 25: we could bump to 1.6.2 while we're at it.

dd23 (approved):
LGTM

kms-connector/Cargo.toml
dd23: but not critical. should not do much for the connector, so we could also do it later.

eudelins-zama: I can't without bumping `kms-grpc` version (because of `kms` pinned deps), which we don't want to do in this PR IMO :innocent:

eudelins-zama:
@mergifyio queue

dvdplm (approved):
lgtm!

PR #2787, opened 2026-06-15
fix(host-contracts): per-contract verify tasks tolerate "Already Verified"

### LINKED ISSUE #2717: host-contracts: per-contract task:verify* should tolerate "Already Verified" on proxies
### Problem

When `task:verify*` per-contract tasks (e.g. `task:verifyACL`, `task:verifyFHEVMExecutor`) are called directly from a deploy script, they throw on a Polygonscan / Etherscan response that is not actually an error:

```
Verifying proxy: 0x1dD3c231eDcDF539C30Fb20B2f3016b92487A1D5
Failed to verify ERC1967Proxy contract at 0x1dD3c231…: Already Verified
Linking proxy 0x1dD3c231… with implementation
Successfully linked proxy to implementation.
An unexpected error occurred:
Error: Verification completed with the following errors.
Error 1: Failed to verify ERC1967Proxy contract at 0x1dD3c231…: Already Verified
```

The implementation IS verified, the proxy IS verified (auto-matched against the standard OZ `ERC1967Proxy` bytecode), and the proxy↔impl link IS recorded. But `@nomicfoundation/hardhat-verify` treats the "Already Verified" response from the proxy-verify endpoint as a hard error and rethrows.

Combined with `set -eo pipefail` in the gitops `contracts` chart `run.sh`, this aborts the entire deploy chain — every step after the first `task:verify*` is skipped.

Hit while rolling out `polygonAmoy` on `zws-dev` (gitops `values/polygon-blockchain/polygon-sc-deploy/values-zws-dev.yaml`).

### Existing workaround (not enough)

`task:verifyAllHostContracts` (`host-contracts/tasks/blockExplorerVerify.ts:184-257`) already wraps every per-contract verify in try/catch with the comment `// to not panic if Etherscan throws an error due to already verified implementation`. The per-contract tasks themselves don't. Deploy scripts that call the per-contract tasks individually (which is the normal pattern in the gitops sc-deploy job, so each verify runs right after its corresponding deploy) get the unswallowed error.

### Request

Push the same try/catch pattern down into each per-contract task. For example, in `task:verifyACL`:

```ts
await run('verify:verify', {
  address: implementationACLAddress,
  constructorArguments: [],
});
try {
  await run('verify:verify', {
    address: proxyAddress,
    constructorArguments: [],
  });
} catch (err) {
  if (String(err).includes('Already Verified')) {
    console.log(`Proxy ${proxyAddress} already verified — skipping.`);
  } else {
    throw err;
  }
}
```

Apply to: `task:verifyACL`, `task:verifyFHEVMExecutor`, `task:verifyKMSVerifier`, `task:verifyInputVerifier`, `task:verifyHCULimit`, `task:verifyPauserSet`, `task:verifyProtocolConfig`, `task:verifyKMSGeneration`. Once landed, `task:verifyAllHostContracts` can stop double-wrapping (or keep it as belt-and-braces).

### Acceptance criteria

- [ ] Each per-contract `task:verify*` tolerates `"Already Verified"` from the proxy-verify step without throwing, logs a clear skip message, and still verifies the implementation normally.
- [ ] Genuine verification failures (wrong API key, explorer down, bytecode mismatch) still throw as today.

### Motivation

Without this fix, every operator needs to remember to either:
- pipe `|| true` after every per-contract verify in their gitops script (swallows real failures too), or
- call `task:verifyAllHostContracts` at the end of the deploy (loses per-step verification ordering).

Eikix:
## What

Standalone version of #2718 targeting `main`, so the fix lives in the mainline and is inherited by future releases rather than being trapped in the unmerged `#2657` stack (`elias/backport-2626-release-0.13.x`), which is currently its only home.

## Why

The per-contract `task:verify*` tasks called `verify:verify` raw. When a deploy script invokes them directly (the gitops `sc-deploy` pattern), an "Already Verified" response is rethrown and, under `set -eo pipefail`, aborts the whole deploy. The bulk `task:verifyAllHostContracts` already tolerates this; the per-contract tasks did not.

## How

A small `verifyContract(run, address)` helper skips the benign "already verified" case (case-insensitive — covers both the auto-matched ERC1967 proxy and the deterministic implementation) and rethrows genuine failures. Every per-contract verify call routes through it. Unit tests cover forward-through, proxy/impl already-verified skip, and genuine-failure rethrow.

> Note: unlike #2718 (which imported a `formatError` util from the #2657 stack), this version inlines the error-message check so it is self-contained — `main` keeps `formatError` private inside `taskDeploy.ts`.

Closes #2717.

### DIFF
--- a/host-contracts/tasks/blockExplorerVerify.ts
+++ b/host-contracts/tasks/blockExplorerVerify.ts
@@ -1,6 +1,27 @@
 import { task, types } from 'hardhat/config';
+import type { RunTaskFunction } from 'hardhat/types';
 
 import { getRequiredEnvVar, loadHostAddresses } from './utils/loadVariables';
+
+// Verifies a single contract on the block explorer, skipping the benign "already verified" response.
+//
+// `@nomicfoundation/hardhat-verify` rethrows "Already Verified" as a hard error — for the auto-matched
+// ERC1967 proxy, and for the deterministic implementation when a prior deploy already verified it.
+// When a per-contract `task:verify*` is called straight from a deploy script (the gitops sc-deploy
+// pattern), that error combines with `set -eo pipefail` to abort the whole deploy. Genuine failures
+// (bad API key, explorer down, bytecode mismatch) are rethrown unchanged.
+export async function verifyContract(run: RunTaskFunction, address: string): Promise<void> {
+  try {
+    await run('verify:verify', { address, constructorArguments: [] });
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    if (/already verified/i.test(message)) {
+      console.log(`${address} is already verified — skipping.`);
+    } else {
+      throw error;
+    }
+  }
+}
 
 task('task:verifyACL')
   .addOptionalParam(
@@ -15,14 +36,8 @@ task('task:verifyACL')
     }
     const proxyAddress = getRequiredEnvVar('ACL_CONTRACT_ADDRESS');
     const implementationACLAddress = await upgrades.erc1967.getImplementationAddress(proxyAddress);
-    await run('verify:verify', {
-      address: implementationACLAddress,
-      constructorArguments: [],
-    });
-    await run('verify:verify', {
-      address: proxyAddress,
-      constructorArguments: [],
-    });
+    await verifyContract(run, implementationACLAddress);
+    await verifyContract(run, proxyAddress);
   });
 
 task('task:verifyFHEVMExecutor')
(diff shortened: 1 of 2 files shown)


PanGan21 (approved):

Eikix:
Superseded by #2807.

The same fix is included there as commit dbc3b4fef63fa3791f310b951aa29cb658c25011, with the same stable patch-id as this PR’s commit 72853e0358f9ffaadad9f9ed43a833096787124a:

568b55c3e05a39dedeae36dc5094136b0a8be30c

PR #2793, opened 2026-06-15
fix(coprocessor): don't put fake block metadata in host-listener

dartdart26:
In case the host-listener can't get block metadata (timestamp, parent hash, etc.), fail and try again instead of putting fake data as that could lead to a state drift as metadata is (or will) be used in rerand and bridging.


PanGan21 (approved):

rudy-6-4 (approved):

dartdart26:
@mergifyio queue

PR #2796, opened 2026-06-15
feat(charts): kms-connector multichain common config

piizama:
1. Adds Polygon support and per-network presets to the kms-connector chart.

- Templates Polygon (URL, chain id, ACL) alongside Ethereum across the gw-listener, kms-worker, and tx-sender deployments, and wires both host ACLs into the worker `hostChains`.
- Adds `configs/contracts-<network>.yaml` presets (devnet/testnet/mainnet) for gateway + host contract addresses and chain ids, resolved via a `kmsConnector.contracts` helper.
- `commonConfig.network` selects the preset; `commonConfig.*ContractAddresses` and `*ChainId` override individual entries when set.

Note: Polygon ACL addresses are placeholders for testnet/mainnet pending real values.

2. Remove invalid default values in `kmsConnectorDbMigration`, forced us to set `kmsConnectorDbMigration.env=[]` in all deployment.


maxnovawind (dismissed):
LGTM!

charts/kms-connector/values.yaml
```diff
@@ -17,27 +17,39 @@ commonConfig:
...
+    # KMS Verifier is only present on Ethereum and not other host chains
+    kmsVerifier: ""
+
+  # Polygon chain RPC node endpoint (HTTP)
+  polygonUrl: "http://polygon-node:8545"
+  # Polygon chain identifier (override the network preset when set)
+  polygonChainId: ""
+  # Polygon smart contract addresses (override the network preset when set)
+  polygonContractAddresses:
+    acl: ""
```
eudelins-zama, line 52: I don't think this is needed

charts/kms-connector/values.yaml
```diff
@@ -202,9 +214,12 @@ kmsConnectorKmsWorker:
...
     hostChains:
-      - url: "http://host-node:8545"
-        chainId: 12345
-        aclAddress: "0x05fD9B5EFE0a996095f42Ed7e77c390810CF660c"
+      - url: "$(KMS_CONNECTOR_ETHEREUM_URL)"
+        chainId: "$(KMS_CONNECTOR_ETHEREUM_CHAIN_ID)"
+        aclAddress: "$(KMS_CONNECTOR_ETHEREUM_ACL_ADDRESS)"
+      - url: "$(KMS_CONNECTOR_POLYGON_URL)"
+        chainId: "$(KMS_CONNECTOR_POLYGON_CHAIN_ID)"
+        aclAddress: "$(KMS_CONNECTOR_POLYGON_ACL_ADDRESS)"
```
eudelins-zama, line 216: This is good

charts/kms-connector/templates/kms-connector-gw-listener-deployment.yaml
```diff
@@ -61,19 +62,23 @@ spec:
...
             - name: KMS_CONNECTOR_ETHEREUM_CHAIN_ID
-              value: {{ .Values.commonConfig.ethereumChainId | quote }}
+              value: {{ $contracts.ethereumChainId | quote }}
             - name: KMS_CONNECTOR_KMS_VERIFIER_ADDRESS
-              value: {{ .Values.commonConfig.ethereumContractAddresses.kmsVerifier | quote }}
+              value: {{ $contracts.ethereumKmsVerifier | quote }}
+            - name: KMS_CONNECTOR_POLYGON_URL
+              value: {{ .Values.commonConfig.polygonUrl | quote }}
+            - name: KMS_CONNECTOR_POLYGON_CHAIN_ID
+              value: {{ $contracts.polygonChainId | quote }}
```
eudelins-zama: I don't think this is needed

charts/kms-connector/templates/kms-connector-kms-worker-deployment.yaml
```diff
@@ -61,15 +62,25 @@ spec:
...
-              value: {{ .Values.commonConfig.ethereumChainId | quote }}
+              value: {{ $contracts.ethereumChainId | quote }}
+            - name: KMS_CONNECTOR_ETHEREUM_ACL_ADDRESS
+              value: {{ $contracts.ethereumAcl | quote }}
+            - name: KMS_CONNECTOR_POLYGON_URL
+              value: {{ .Values.commonConfig.polygonUrl | quote }}
+            - name: KMS_CONNECTOR_POLYGON_CHAIN_ID
+              value: {{ $contracts.polygonChainId | quote }}
+            - name: KMS_CONNECTOR_POLYGON_ACL_ADDRESS
+              value: {{ $contracts.polygonAcl | quote }}
```
eudelins-zama, line 81: I don't think this is needed

charts/kms-connector/templates/kms-connector-tx-sender-deployment.yaml
```diff
@@ -61,19 +62,23 @@ spec:
...
             - name: KMS_CONNECTOR_ETHEREUM_CHAIN_ID
-              value: {{ .Values.commonConfig.ethereumChainId | quote }}
+              value: {{ $contracts.ethereumChainId | quote }}
             - name: KMS_CONNECTOR_KMS_VERIFIER_ADDRESS
-              value: {{ .Values.commonConfig.ethereumContractAddresses.kmsVerifier | quote }}
+              value: {{ $contracts.ethereumKmsVerifier | quote }}
+            - name: KMS_CONNECTOR_POLYGON_URL
+              value: {{ .Values.commonConfig.polygonUrl | quote }}
+            - name: KMS_CONNECTOR_POLYGON_CHAIN_ID
+              value: {{ $contracts.polygonChainId | quote }}
```
eudelins-zama: I don't think this is needed

charts/kms-connector/values.yaml
eudelins-zama: Don't we need to add something like:

```
    - name: ETHEREUM_RPC_URL
      valueFrom:
        secretKeyRef:
          name: rpc
          key: ethereum-rpc-url
```

If we use env var in the `values.yaml` though?

charts/kms-connector/templates/_helpers.tpl
```diff
@@ -12,3 +12,45 @@
...
+{{- $pol := .Values.commonConfig.polygonContractAddresses | default dict -}}
+gatewayChainId: {{ default (index $preset "gateway.chain_id") .Values.commonConfig.gatewayChainId | quote }}
+ethereumChainId: {{ default (index $preset "ethereum.chain_id") .Values.commonConfig.ethereumChainId | quote }}
+polygonChainId: {{ default (index $preset "polygon.chain_id") .Values.commonConfig.polygonChainId | quote }}
+decryption: {{ default (index $preset "gateway.decryption.address") $gw.decryption | quote }}
+gatewayConfig: {{ default (index $preset "gateway.gateway_config.address") $gw.gatewayConfig | quote }}
+kmsGeneration: {{ default (index $preset "gateway.kms_generation.address") $gw.kmsGeneration | quote }}
+ethereumAcl: {{ default (index $preset "ethereum.acl.address") $eth.acl | quote }}
+ethereumKmsVerifier: {{ default (index $preset "ethereum.kms_verifier.address") $eth.kmsVerifier | quote }}
+polygonAcl: {{ default (index $preset "polygon.acl.address") $pol.acl | quote }}
```
eudelins-zama, line 43: I don't think this is needed

charts/kms-connector/templates/_helpers.tpl
```diff
@@ -12,3 +12,45 @@
...
+{{- $preset := dict -}}
+{{- if $network -}}
+{{- $preset = .Files.Get (printf "configs/contracts-%s.yaml" $network) | fromYaml -}}
+{{- end -}}
+{{- $gw := .Values.commonConfig.gatewayContractAddresses | default dict -}}
+{{- $eth := .Values.commonConfig.ethereumContractAddresses | default dict -}}
+{{- $pol := .Values.commonConfig.polygonContractAddresses | default dict -}}
+gatewayChainId: {{ default (index $preset "gateway.chain_id") .Values.commonConfig.gatewayChainId | quote }}
+ethereumChainId: {{ default (index $preset "ethereum.chain_id") .Values.commonConfig.ethereumChainId | quote }}
+polygonChainId: {{ default (index $preset "polygon.chain_id") .Values.commonConfig.polygonChainId | quote }}
```
eudelins-zama, line 37: I don't think this is needed

charts/kms-connector/values.yaml
piizama: Yes but this is specific to the operator environment We shouldn't provide this in the chart default values.

charts/kms-connector/values.yaml
piizama: Those value need to be set to the empty string when setting a network preset.

charts/kms-connector/templates/kms-connector-gw-listener-deployment.yaml
piizama: It's true that as of 0.13 the listener doesn't connect to Polygon (it's done by the worker) but it wouldn't surprise me that in the future it will need to connect to all hosts chains. I think it's worth it to keep the config here but also okay if you insist on removing it @eudelins-zama.

charts/kms-connector/templates/kms-connector-kms-worker-deployment.yaml
piizama: This is actually used in order to populate the hostChains JSON automatically reusing the env vars from `commonConfig`. As we will support more and more chains, this helps abstract away the host chain configs for the node operator.

```
kmsConnectorKmsWorker:
  config
    hostChains:
      - url: "$(KMS_CONNECTOR_ETHEREUM_URL)"
        chainId: "$(KMS_CONNECTOR_ETHEREUM_CHAIN_ID)"
        aclAddress: "$(KMS_CONNECTOR_ETHEREUM_ACL_ADDRESS)"
      - url: "$(KMS_CONNECTOR_POLYGON_URL)"
        chainId: "$(KMS_CONNECTOR_POLYGON_CHAIN_ID)"
        aclAddress: "$(KMS_CONNECTOR_POLYGON_ACL_ADDRESS)"
```

charts/kms-connector/templates/kms-connector-tx-sender-deployment.yaml
piizama: I think this should be removed as well, even the above KMS_CONNECTOR_KMS_VERIFIER_ADDRESS config looks wrong.

@eudelins-zama  Do you confirm this whole config block can be removed from the txsender ?
```
            - name: KMS_CONNECTOR_KMS_VERIFIER_ADDRESS
              value: {{ $contracts.ethereumKmsVerifier | quote }}
            - name: KMS_CONNECTOR_POLYGON_URL
              value: {{ .Values.commonConfig.polygonUrl | quote }}
            - name: KMS_CONNECTOR_POLYGON_CHAIN_ID
              value: {{ $contracts.polygonChainId | quote }}
```

charts/kms-connector/templates/_helpers.tpl
piizama: This is used in the templating from the config preset. We have to keep it.

charts/kms-connector/templates/_helpers.tpl
piizama: This is used in the templating from the config preset. We have to keep it.

tawadaa (dismissed):

charts/kms-connector/templates/kms-connector-tx-sender-deployment.yaml
eudelins-zama: I'm handling the `KMS_VERIFIER_ADDRESS` in another PR, please leave it for now :innocent: 

But you can remove the `POLYGON_*`

charts/kms-connector/templates/kms-connector-kms-worker-deployment.yaml
eudelins-zama: This is a bit confusing to me.

I though values in `values.yaml` (like `commonConfig`) were used in `*-deployment.yaml` to set env var values like [here](https://github.com/zama-ai/fhevm/blob/3c9f43fb75ec8b0437f1fec51402565a11f1d502/charts/kms-connector/templates/kms-connector-kms-worker-deployment.yaml#L70-L71).

But now you're saying that env var values in `*-deployment.yaml` can be used to set the values in the `values.yaml` file? :sweat_smile: 

Probably a lack of helm knowledge from my side, but would appreciate a brief explanation

charts/kms-connector/templates/kms-connector-gw-listener-deployment.yaml
eudelins-zama: I would prefer to avoid anticipating too much.

For now, we have no plan for the `gw-listener` to listen to other chains than Ethereum and the Gateway, so it can be a bit confusing to add these env vars already

charts/kms-connector/values.yaml
eudelins-zama: It's a bit related to this [comment](https://github.com/zama-ai/fhevm/pull/2796#discussion_r3420570703).

I though env vars were supposed to be defined that way, not in `*-deployment.yaml`

charts/kms-connector/templates/kms-connector-kms-worker-deployment.yaml
piizama: It's a long story, env vars are only resolved on application startup. In the helm chart we set env vars that resolve to other env vars to have the possibility to set them either as literal values overrides (eg. `commonConfig.ethereumUrl: "http://eth-node:8545"`) or env vars that are loaded from kubernetes secret (eg. commonConfig.ethereumUrl: "$(ETHEREUM_URL)". The templating allow us to set the correct `MS_CONNECTOR_*` for the app.
I hope this clarifies things.

eudelins-zama: Thanks for the clarification! :pray: 

It's still blurry to me, but I think I get the overall idea. :slightly_smiling_face:

charts/kms-connector/templates/kms-connector-gw-listener-deployment.yaml
piizama: I have removed the unused POLYGON env vars.

charts/kms-connector/templates/kms-connector-tx-sender-deployment.yaml
piizama: I have removed the unused POLYGON env vars.

eudelins-zama: Thanks!

charts/kms-connector/templates/kms-connector-gw-listener-deployment.yaml
eudelins-zama: Thanks!

eudelins-zama (approved):
LGTM, thanks!

piizama:
@Mergifyio queue

PR #2800, opened 2026-06-16
feat(coprocessor): harden zk-verifier to shrink proof attack surface

antoniupop:
Restrict the conformance parameters used to deserialize an incoming ProvenCompactCiphertextList in the zkproof-worker so that only the proof shapes the current stack actually produces are accepted. Anything else is rejected during the conformance check, before any ZK verification runs.

### DIFF
--- a/coprocessor/fhevm-engine/stress-test-generator/src/zk_gen.rs
+++ b/coprocessor/fhevm-engine/stress-test-generator/src/zk_gen.rs
@@ -187,7 +187,7 @@ pub async fn generate_random_handle_vec(
     info!(target: "tool", "ZK Transaction: tx_id: {:?}, inputs = {:?}", to_hex(transaction_id.as_ref()), count);
 
     let the_list = builder
-        .build_with_proof_packed(&public_params, &aux_data, tfhe::zk::ZkComputeLoad::Proof)
+        .build_with_proof_packed(&public_params, &aux_data, tfhe::zk::ZkComputeLoad::Verify)
         .unwrap();
     let zk_pok = fhevm_engine_common::utils::safe_serialize(&the_list);
     let zk_id = ZK_PROOF_ID.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
@@ -249,7 +249,7 @@ pub async fn generate_and_insert_inputs_batch(
         let zk_id = ZK_PROOF_ID.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
         info!(target: "tool", "zkpok id: {}, count = {:?}, seq_num = {} of {} , txn: {:?}, ", zk_id, inputs_count, idx, batch_size, transaction_id );
         let the_list = builder
-            .build_with_proof_packed(&public_params, &aux_data, tfhe::zk::ZkComputeLoad::Proof)
+            .build_with_proof_packed(&public_params, &aux_data, tfhe::zk::ZkComputeLoad::Verify)
             .unwrap();
 
         let zk_pok = fhevm_engine_common::utils::safe_serialize(&the_list);
--- a/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
+++ b/coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
@@ -16,6 +16,7 @@ use sqlx::{postgres::PgListener, PgPool};
 use sqlx::{Postgres, Transaction};
 use std::str::FromStr;
 use tfhe::integer::ciphertext::IntegerProvenCompactCiphertextListConformanceParams;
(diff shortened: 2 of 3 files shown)


coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
```diff
@@ -488,6 +489,35 @@ pub(crate) fn verify_proof(
...
+    pks: &tfhe::CompactPublicKey,
+    crs: &Crs,
+) -> IntegerProvenCompactCiphertextListConformanceParams {
+    IntegerProvenCompactCiphertextListConformanceParams::from_public_key_encryption_parameters_and_crs_parameters(
+        pks.parameters(),
+        &crs.crs,
+    )
+    .forbid_hash_config(ZkPkeV2SupportedHashConfig::V0_4_0)
+    .forbid_hash_config(ZkPkeV2SupportedHashConfig::V0_7_0)
+    .forbid_compute_load(ZkComputeLoad::Proof)
```
PanGan21, line 518: Similar to this maybe this code path needs to change as well?
https://github.com/zama-ai/fhevm/blob/6cc4e7d08fe75958fe669a732c7907871edefbef/coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs#L270-L290

PanGan21: Also marking the pr with e2e label to make sure nothing was broken. I hope you don't mind.

coprocessor/fhevm-engine/stress-test-generator/src/zk_gen.rs
```diff
@@ -187,7 +187,7 @@ pub async fn generate_random_handle_vec(
     info!(target: "tool", "ZK Transaction: tx_id: {:?}, inputs = {:?}", to_hex(transaction_id.as_ref()), count);
 
     let the_list = builder
-        .build_with_proof_packed(&public_params, &aux_data, tfhe::zk::ZkComputeLoad::Proof)
+        .build_with_proof_packed(&public_params, &aux_data, tfhe::zk::ZkComputeLoad::Verify)
```
nsarlin-zama, line 190: just to double check, on the example I used `.forbid_compute_load(ZkComputeLoad::Proof)` because I assume you are using ComputeLoad Verify, that's correct, right?

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
antoniupop: Looks more like something to remove not sure why it's still here.

PanGan21: Oh it is dead code?

coprocessor/fhevm-engine/stress-test-generator/src/zk_gen.rs
antoniupop: Hmm - the SDKs produce Verify, but I can't say it's impossible to get that from outside although that wouldn't make much sense from a computation load perspective. We only had Proof in tests. Would you recommend that we don't exclude it?
Does it make a big difference if we allow both?

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
antoniupop: Yes afaik it is. Cleanup should ideally be in separate PR...

coprocessor/fhevm-engine/stress-test-generator/src/zk_gen.rs
nsarlin-zama: That's a protocol decision, so I don't know what are the concrete real life implications here. 
But if the sdk only generates Verify, and we receive Proof, high are the chance that someone is doing something fishy and that we should refuse it.

But I don't know if that's something that should be standardized beyond what the sdk does. As you said I don't see any legit reason why a client would use Proof if we accept Verify

nsarlin-zama (approved):
usage looks correct from a TFHE-rs perspective!

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
PanGan21: Ok lemme do that against main. Good that we identified it though!

PanGan21 (approved):

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
PanGan21: Here is the pr to main: https://github.com/zama-ai/fhevm/pull/2802

antoniupop:
@mergifyio queue

PR #2801, opened 2026-06-16
feat(coprocessor): support confidential handle bridging

dartdart26:
This commit implements confidential handle bridging, i.e. allowing the use of a handle (and its corresponding ciphertext) from a source to a destination host chain.

It also supports a failback mechanism to provide a plaintext value for any missing handle on the destination chain. This mechanism reuses the existing trivial encryption flow from the host-listener to the tfhe-worker.

The general mental model is that a handle/ciphertext can now materialize from:
 * computations (as before)
 * bridging (new)
 * fallback (new)

These cases matter when we deal with automatic drift reversals, because we want to know which ciphertexs/ciphertext_digests to remove, based on where they came from.

To handle potential multiple drift reversals due to interaction of that feature with bridging, we increase the drift revert max attempts from 2 to 10. Generally speaking, we let each chain (source or destination) handle its own drifts.

Implemented according to RFC-008.


coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
```diff
@@ -264,7 +264,7 @@ pub fn trivial_encrypt_be_bytes(
 }
 
 pub fn current_ciphertext_version() -> i16 {
-    0
+    crate::types::HANDLE_VERSION as i16
```
goshawk-3: Why does the `current_ciphertext_version` return handle_version? Can we have `current_handle_version` new function instead?

We’ll have a conflict because this has already been changed in the B/G upgrade PR.

https://github.com/zama-ai/fhevm/blob/7f93a7fc2bf127c04964e27a00c28333e108eb4b/coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs#L267

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
```diff
@@ -0,0 +1,205 @@
...
+        "Starting confidential bridge worker"
+    );
+
+    loop {
+        if cancel_token.is_cancelled() {
+            break;
+        }
+        match drain_associations(&pool, args.bridge_associate_batch_size, &cancel_token).await {
+            Ok(associated) if associated > 0 => {
+                info!(target: "bridge", associated, "Associated bridged handles")
```
PanGan21, line 77: I think it is high value if we add a prometheus metric here so an association is visible from dashboards as well.

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
```diff
@@ -0,0 +1,205 @@
...
+async fn associate_batch(pool: &PgPool, batch_size: i64) -> Result<u64, sqlx::Error> {
+    let mut txn = pool.begin().await?;
+
+    // A pair is ready to associate when:
+    // - both validated events are present: the destination `HandleBridged`
+    //   and the matching source `BridgeHandle` one
+    // - the source ciphertext is fully materialized: its ct64 blob exists and
+    //   both digests (ct64 and ct128) are computed
+    // - it has not been associated yet
+    let ready = sqlx::query!(
```
PanGan21, line 124: It would be great to be able to identify from a log or a metric that a ciphertext is old enough and not associated with any other ciphertext. Then it is easier to know when to call `grantFallbackPlaintext`.

coprocessor/fhevm-engine/fhevm-engine-common/src/bridge.rs
```diff
@@ -0,0 +1,139 @@
...
+    out[24..].copy_from_slice(&v.to_be_bytes());
+    out
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn metadata_bytes_are_correct() {
```
PanGan21, line 67: Since `derive_dst_handle` should be always the same between coprocessor and host contracts it could be worth adding the same test on both sides. i.e. same inputs should result to a standard same output. Not sure if this is already done from this test as a shared handle between the two.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -959,6 +961,123 @@ impl Database {
...
+                    > 0
+            }
+            BridgeContractEvents::HandleBridged(e) => {
+                // Verify the destination handle was correctly derived and ignore the
+                // event otherwise.
+                let Some(acl) = acl_contract_address else {
+                    warn!("Cannot verify HandleBridged derivation: ACL address not configured");
+                    return Ok(false);
+                };
+                let expected = derive_dst_handle(
```
PanGan21, line 1032: This validates against the destination block's parent_hash + timestamp. Since the on-chain handle includes in blockhash(n-1) + block.timestamp, a reorg most likely emit a different dstHandle.
What happens to the old `bridge_handle_events` and `handle_bridged_events` in that case?

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
```diff
@@ -0,0 +1,205 @@
...
+    let ciphertext_copied = sqlx::query!(
+        r#"
+        INSERT INTO ciphertexts (handle, ciphertext, ciphertext_version, ciphertext_type)
+        SELECT $1, ciphertext, ciphertext_version, ciphertext_type
+        FROM ciphertexts
+        WHERE handle = $2
+        ON CONFLICT (handle, ciphertext_version) DO NOTHING
+        "#,
+        dst_handle,
+        src_handle,
```
PanGan21, line 201: I am wondering if we could avoid duplication of the ciphertext by some foreign key in the database. However I understand that this can overcomplicate things, in that case we can skip it.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
```diff
@@ -101,6 +101,24 @@ BEGIN
...
+-- Placed after the key-rotation check on purpose: re-association re-copies the
+-- source bytes (no re-encryption), so the new-key hazard that check guards for
+-- does not apply to bridged copies.
+-- ===========================================================================
+
+INSERT INTO _affected_output_handles (handle)
+SELECT dst_handle
+  FROM handle_bridged_events
+ WHERE dst_chain_id = :'chain_id'
+   AND block_number > :'to_block_number';
```
PanGan21, line 120: Do we have a test that exercises the bridge cleanup logic in case of drift?

PanGan21 (commented):
Very well structured work. Some suggestions and questions for my own understanding.

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
dartdart26: Hmmm, good observation. What is the difference between the two? Ciphertext version maybe is a bad name as ciphertext versions should come from the KMS, right?

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
dartdart26: Good idea, let me do that!

coprocessor/fhevm-engine/fhevm-engine-common/src/bridge.rs
dartdart26: You mean compare the output from the rust_bindings to the native rust function used by coprocessor? I think that's a good test, yeah.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
dartdart26: We keep them forever and they might remain unassociated. It doesn't look like a problem to me, but wdyt?

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
dartdart26: We discussed that and decided it adds complexity and edge cases we might want to avoid, especially around reversals. Furthermore, the amount of duplicated ciphertexts might not be big anyway.

coprocessor/fhevm-engine/db-migration/db-scripts/revert_coprocessor_db_state.sql
dartdart26: Yeah, I added tests in the reversal integration test for that, basically asserting these tables are cleaned too. Also other tests that make sure we detect bridged ciphertexts (not coming from computation) to revert.

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
dartdart26: The reasoning from governance's perspective is that the handle on the destination is on the ACL onchain and not in the coprocessor's DB.

Maybe we could count the number of unassociated ones on every bridge poll and if governance observes count > 0, then they go in and investigate. We can leave the mechanism of that investigation for a later commit/PR, wdyt?

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
PanGan21: Thanks makes sense, just wanted to understand the rationale

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
PanGan21: I would say it is not a problem, however we don't want the db to be growing so we could consider some garbage collection strategy for that.

coprocessor/fhevm-engine/fhevm-engine-common/src/bridge.rs
PanGan21: Yes something like this. I was thinking that we really want to guard the implementation of `derive_dst_handle` from any change. I see comments emphasize this, just want to prevent an accidental change.

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
PanGan21: I created a ticket for that here: https://github.com/zama-ai/fhevm-internal/issues/1527 which I can also pick up

coprocessor/fhevm-engine/fhevm-engine-common/src/bridge.rs
dartdart26: I added a Solidity contract test that computes a handle and then use the same to verify Rust produces the same.

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
dartdart26: I did that - added a gauge with unassociated ones.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
dartdart26: Maybe we leave that for future work? The entries are very small anyways in this table. Furthermore, the association happens on the handles and chain IDs, so generally speaking, it will actually succeed even on reorg.

PanGan21: cc @antoniupop maybe we should add this into our future garbage collection strategy

coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs
dartdart26: I will make it 0 again, I've conflated the two.

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
dartdart26: Actually, I will remove that gauge for now - let's handle it separately via your issue.

PanGan21 (approved):

PR #2810, opened 2026-06-17
feat(relayer): context-aware /v2/keyurl

manoranjith:
Relayer side of RFC 005 context-aware `/v2/keyurl`. Design, routing
table, and acceptance criteria: [zama-ai/fhevm-internal#1466][issue].

[issue]: https://github.com/zama-ai/fhevm-internal/issues/1466

Notable beyond the diff:

- A single host-chain poller fetches `getActiveKeyId` /
  `getActiveCrsId` / `getCurrentKmsContextAndEpoch` on an interval,
  reading at the **`Finalized`** block tag in parity with the
  kms-connector host-chain listener, and feeds the latest value into a
  `tokio::watch` that backs the endpoint.

- Startup is gated on the first poll (fail-fast → exit/restart), so the
  served value is always chain-sourced — the watch holds a plain
  `KeyUrlResponseJson` and `contextId` / `epochId` are non-optional.

- Additive response: `contextId` / `epochId` are new fields older SDKs
  ignore; `dataId` changes value (placeholder → real on-chain id) but
  not type.

Targets `feat/1130-epoch-lifecycle` (#2615) for the new getters;
`fhevm_host_bindings` is pinned to its tip — re-pin to the merge commit
once it lands.

Tests: `cargo test --lib`, `keyurl_poller_test`, `keyurl_test`,
`threshold_resolver_test` pass locally; `clippy --all-targets` clean.


relayer/config/.env.example
```diff
@@ -5,11 +5,9 @@ APP_NETWORKS__ROLLUP__CHAIN_ID=654321
 APP_NETWORKS__ROLLUP__RETRY_DELAY=1000
 APP_NETWORKS__ROLLUP__MAX_RECONNECTION_ATTEMPTS=3
 
-# Key URL configuration
-APP_KEYURL__FHE_PUBLIC_KEY__DATA_ID=fhe-public-key-data-id
-APP_KEYURL__FHE_PUBLIC_KEY__URL=http://0.0.0.0:3001/publicKey.bin
-APP_KEYURL__CRS__DATA_ID=crs-data-id
-APP_KEYURL__CRS__URL=http://0.0.0.0:3001/crs2048.bin
+# /v2/keyurl host-chain poller: KMSGeneration address + poll interval (ms).
+APP_PROTOCOL_CONFIG__KMS_GENERATION_ADDRESS=0x1234567890123456789012345678901234567890
```
isaacdecoded: It seems we have two contract names in this env variable. Should it be `APP_KEYURL__KMS_GENERATION_ADDRESS` instead?

Same for the variable below: `APP__KEYURL_POLL_INTERVAL_MS`

relayer/config/local.mainnet.yaml.example
```diff
@@ -15,10 +15,14 @@ host_chains:
 
 protocol_config:
   ethereum_http_rpc_url: "https://ethereum-rpc.publicnode.com"
-  address: "TODO"
+  address: "0x0000000000000000000000000000000000000000"
   retry:
     max_attempts: 3
     retry_interval_ms: 200
+  # KMSGeneration contract address on the host chain (source for /v2/keyurl).
```
isaacdecoded, line 25: nit: It looks a bit confusing having the `kms_generation_address` under the `protocol_config` definitions since both of them are smart contracts. Not sure if it could be clearer to keep the `keyurl` definitions removed below at line 138.

relayer/src/config/settings.rs
```diff
@@ -695,6 +685,10 @@ impl Settings {
                 &self.gateway.contracts.input_verification_address,
             ),
             ("protocol_config", &self.protocol_config.address),
+            (
+                "protocol_config.kms_generation_address",
```
isaacdecoded: Also, it is very confusing here, at least from the contract perspective, to have both names mixed: ProtocolConfig and KMSGeneration are different smart contracts.

relayer/src/host/keyurl_poller.rs
```diff
@@ -0,0 +1,258 @@
+//! Host-chain poller that keeps the `/v2/keyurl` response in sync with on-chain KMS state.
```
isaacdecoded:
```suggestion
//! Host-chain poller that keeps the `/v2/keyurl` response in sync with on-chain ProtocolConfig and KMSGeneration contracts' states.
```

relayer/src/host/keyurl_poller.rs
```diff
@@ -0,0 +1,258 @@
...
+//!
+//! Polls the active key/CRS/context on an interval rather than subscribing to events: the
```
isaacdecoded:
```suggestion
//! Polls the active key/CRS/KMS context on an interval rather than subscribing to events: the
```

relayer/tests/common/utils.rs
```diff
@@ -16,6 +16,8 @@ use fhevm_gateway_bindings::decryption::IDecryption::{
     RequestValiditySeconds, UserDecryptionRequestPayload,
 };
 use fhevm_host_bindings::acl::ACL;
+use fhevm_host_bindings::i_protocol_config::IProtocolConfig;
+use fhevm_host_bindings::kms_generation::KMSGeneration;
```
isaacdecoded: nit: we can probably also import the interface of the KMSGeneration instead of the contract - same as the IProtocolConfig

manoranjith: Fixed in [15c8c8e](https://github.com/zama-ai/fhevm/pull/2810/commits/15c8c8ea60ee1435d1a5a5222d3aef30530a9f15)

relayer/src/config/settings.rs
manoranjith: Created a separate config for key url contracts in this commit: [15c8c8e](https://github.com/zama-ai/fhevm/pull/2810/commits/15c8c8ea60ee1435d1a5a5222d3aef30530a9f15)

relayer/config/local.mainnet.yaml.example
manoranjith: Created a separate config for key url contracts in this commit: [15c8c8e](https://github.com/zama-ai/fhevm/pull/2810/commits/15c8c8ea60ee1435d1a5a5222d3aef30530a9f15)

relayer/config/.env.example
manoranjith: Created a separate config for key url contracts in this commit: [15c8c8e](https://github.com/zama-ai/fhevm/pull/2810/commits/15c8c8ea60ee1435d1a5a5222d3aef30530a9f15)

PR #2811, opened 2026-06-17
feat(library-solidity): add FHE.toExternal methods for all encrypted types

poppyseedDev:
## Summary

Adds `FHE.toExternal`, the inverse of `FHE.fromExternal`, requested by OpenZeppelin. It takes an on-chain encrypted value (`euintXX`/`ebool`/`eaddress`) the contract already holds and re-wraps it into the matching external type (`externalEuintXX`/`externalEbool`/`externalEaddress`).

The method is **overloaded** (target type inferred from the argument), mirroring the existing `fromExternal` API for symmetry:

```solidity
euint64 a = FHE.fromExternal(handle, proof);   // external -> internal
externalEuint64 ext = FHE.toExternal(a);        // internal -> external
```

### Why it's useful
Many functions — including standardized interfaces such as confidential tokens — type their entry points as `(externalEuintXX, bytes inputProof)` because the usual caller is an EOA submitting an encrypted input + ZK proof. A contract that already holds an `euintXX` can't produce such a proof on-chain. `toExternal`, paired with the empty-proof path of `fromExternal`, bridges that gap for contract-to-contract composition and smart-contract accounts. It only changes the static type — no verification, no ACL grant — so callers must set up ACL (`FHE.allow`) first.

## Changes
- **`codegen/src/templateFHEDotSol.ts`** — generator emits an overloaded `toExternal` per type, right after `fromExternal`.
- **`lib/FHE.sol`** — regenerated (8 methods: `ebool`, `euint8/16/32/64/128/256`, `eaddress`); purely additive.
- **`test/FHEToExternal.t.sol`** (new) — Foundry unit tests: bit-exact handle preservation for all 8 types + a 256-run fuzz.
- **`test/fhevmOperations/manual.ts`** + **`examples/tests/FHEVMManualTestSuite.sol`** — Hardhat round-trip tests (uint/bool/address).
- **`test-suite/e2e/...`** — mirrored round-trip tests for the real-stack integration suite (public-decrypt style).
- **`docs/solidity-guides/inputs.md`** — documents `toExternal`, the empty-proof pairing, and the ACL caveat.

## Testing
- **Foundry**: 9/9 pass (8 types + fuzz).
- **Mocked Hardhat**: 3/3 round-trips pass.
- **e2e (live stack, real KMS public decryption)**: 3/3 round-trips pass — `toExternalEuint64`/`Ebool`/`Eaddress` (3 ciphertexts, 3 proofs, 0 divergences).

### DIFF
--- a/docs/solidity-guides/inputs.md
+++ b/docs/solidity-guides/inputs.md
@@ -127,6 +127,22 @@ function transfer(
 2. **Type conversion**:\
    The function transforms `externalEbool`, `externalEaddress`, `externalEuintXX` into the appropriate encrypted type (`ebool`, `eaddress`, `euintXX`) for further operations within the contract.
 
+### Re-exporting a handle with `FHE.toExternal`
+
+`FHE.toExternal` is the inverse of `FHE.fromExternal`: it takes an on-chain encrypted value (`euintXX`, `ebool`, `eaddress`) that the contract already holds and re-wraps it into the matching external type (`externalEuintXX`, `externalEbool`, `externalEaddress`). The function is overloaded, so the target type is inferred from the argument — exactly like `fromExternal`.
+
+This is useful for **contract-to-contract composition**. Many functions — including standardized interfaces such as confidential tokens — type their entry points as `(externalEuintXX handle, bytes inputProof)` because the usual caller is an EOA submitting an encrypted input together with a zero-knowledge proof. A contract that already holds an `euintXX` cannot produce such a proof on-chain. `toExternal`, paired with the empty-proof path of `fromExternal`, bridges that gap:
+
+```solidity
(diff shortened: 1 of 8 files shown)


geoxel:
@poppyseedDev Could I ask you to target `main` instead? (because these changes do not touch the SDK and our future merge of `devex/js-sdk` to `main` should not trigger reviewers from the `fhevm-coprocessor` team when possible)

jatZama:
looks good, but I think you also forgot to regenerate the FHE.sol in host-contracts here. @poppyseedDev

poppyseedDev:
Closing in favor of #2821, which targets `main` instead of `devex/js-sdk` (@geoxel) and also regenerates the host-contracts `FHE.sol` (@jatZama). Thanks both!

PR #2818, opened 2026-06-18
feat(kms-connector): off-chain ct attestation verifier in shadow mode

eudelins-zama:
### Summary

Off-chain ciphertext-attestation verifier implementation in the KMS connector's `kms-worker`.
Before retrieving ciphertexts from the Coprocessors' S3 buckets, the worker now fans HEAD requests out to every registered Coprocessor, computes an off-chain consensus verdict for each handle, and compares it against the authoritative on-chain `SnsCiphertextMaterial`.

This ships in **shadow mode**: a failed verification is only logged (`warn!`), so the on-chain consensus verdict stays authoritative and ciphertext retrieval proceeds regardless of the off-chain verdict.

### What changed

- **Consensus evaluation** (`shared/ciphertext-attestation/src/consensus.rs`): validates fetched attestations and evaluates the majority threshold per handle, returning the agreed-upon material and the count of distinct in-registry signers.
- **Coprocessor registry mirror** : a TTL'd snapshot of the on-chain registry (signer set, per-Coprocessor S3 bucket URLs, majority threshold), avoiding RPC calls on every decryption request.
- **Ciphertext manager**: orchestrates the attestation HEAD fan-out, consensus check, and S3 retrieval behind a single `retrieve_verified_ciphertexts` entry point. The previous `event_processor/s3.rs` was moved under `ciphertext/` as part of the ciphertext-management refactor.
- **Configuration**: new optional `[ct_attestation]` section with an emergency kill switch (`enabled`), per-bucket HEAD timeout, and registry refresh interval.

### Testing

- Backward compatible e2e tests (with non-updated copro): https://github.com/zama-ai/fhevm/actions/runs/27759121492/job/82131155748?pr=2818
- Manually e2e tested against updated copro
  - Will also be e2e tested in CI when put in the merge queue

Closes https://github.com/zama-ai/fhevm-internal/issues/1411


maksymsur:
@claude /pr-review

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/registry.rs
```diff
@@ -0,0 +1,319 @@
...
+        let get_copro_tx_senders = contract.getCoprocessorTxSenders();
+        let get_copro_threshold = contract.getCoprocessorMajorityThreshold();
+
+        let (signers, tx_senders, threshold_u256) = tokio::try_join!(
+            biased;
+            get_copro_signers.call(),
+            get_copro_tx_senders.call(),
+            get_copro_threshold.call()
+        )
+        .map_err(|e| RegistryError::Transient(e.into()))?;
```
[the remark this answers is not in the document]
eudelins-zama: Hallucination: https://docs.rs/tokio/latest/tokio/macro.try_join.html

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -0,0 +1,422 @@
...
+        // Fallback to old format header if attestation is not available
+        Err(e) => {
+            warn!("attestation fetch error: {e}. Falling back to {OLD_CT_FORMAT_HEADER} header");
+            match headers.get(OLD_CT_FORMAT_HEADER).map(AsRef::as_ref) {
+                Some(b"compressed_on_cpu") | Some(b"compressed_on_gpu") => {
+                    GrpcCiphertextFormat::BigCompressed
+                }
+                _ => GrpcCiphertextFormat::BigExpanded,
+            }
+        }
```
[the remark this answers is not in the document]
eudelins-zama: Yeah, maybe `debug!` would be more suitable here indeed

kms-connector/crates/kms-worker/tests/s3.rs
```diff
@@ -71,11 +92,7 @@ async fn test_get_unstored_s3_ciphertext() -> anyhow::Result<()> {
             .into(),
         ..Default::default()
     };
-    let s3_service = S3Service::new(&config, mock_provider, s3_client);
-    if let Ok(ct) = s3_service
-        .retrieve_s3_ciphertext(&bucket_url, &sns_ct, S3_CT_UNSTORED)
-        .await
-    {
+    if let Ok(ct) = retrieve_s3_ciphertext(&s3_client, &bucket_url, &sns_ct, S3_CT_UNSTORED).await {
         panic!("Unexpected ciphertext retrievd {ct:?}");
```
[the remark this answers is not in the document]
eudelins-zama: Test code, but yes indeed

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /// Retrieves the ciphertexts of `sns_materials` from the Coprocessors' S3 buckets, after
+    /// running the off-chain attestation check.
+    ///
+    /// Shadow mode (RFC 023): an attestation failure is only logged for now — the on-chain
+    /// snapshot stays authoritative and retrieval proceeds regardless of the verdict.
+    pub async fn retrieve_verified_ciphertexts(
+        &self,
+        sns_materials: &[SnsCiphertextMaterial],
```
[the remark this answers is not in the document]
eudelins-zama: Interesting, I'll double check this one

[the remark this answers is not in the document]
eudelins-zama: Indeed, this is true: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations

Will fix! :slightly_smiling_face:

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -0,0 +1,185 @@
...
+    ) -> anyhow::Result<Vec<TypedCiphertext>> {
+        if let Err(e) = self.verify_attestations(sns_materials).await {
+            warn!("{e}");
```
maksymsur: shall we add metrics here as that's a serious cases to track on?

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -0,0 +1,185 @@
...
+        for material in materials {
+            self.verify_material_attestations(&registry, material)
+                .await
+                .with_context(|| {
+                    format!(
+                        "Attestation verification failed for handle {}",
+                        hex::encode(material.ctHandle)
+                    )
+                })?;
+        }
```
maksymsur: So per-handle verification is sequential --> `~N×head_timeout` inline latency... are we Ok with that? Maybe incorporate smth with `join_all` ?

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/registry.rs
```diff
@@ -0,0 +1,319 @@
...
+                "Not a single Coprocessor with a non-empty S3 bucket URL in the registry"
+            )));
+        }
+
+        Ok(Self::new(
+            signers.into_iter().collect(),
+            tx_sender_to_bucket,
+            threshold,
+        ))
+    }
```
maksymsur: So empty Coprocessor signer set is never rejected... shall we have a guard about it as well?

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -0,0 +1,422 @@
...
+            Ok(attestation) => valid_attestations.push((tx_sender, attestation)),
+            Err(e) => warn!(
+                %tx_sender,
+                handle = hex::encode(handle),
+                "Failed to fetch attestation: {e}"
+            ),
+        }
+    }
+    valid_attestations
+}
```
maksymsur: So this spawns one unbounded task per bucket and blocks on `tasks.join_all().await`, so consensus can't be evaluated until every bucket answers (or times out). If I understood correctly - RFC step 1 explicitly wants early-exit once the threshold is satisfiable and bounded concurrency.. --> maybe drive the `JoinSet` with `join_next`, tally distinct in-registry signers as results arrive, and return once threshold is reached? You also may cap in-flight requests with a semaphore.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -0,0 +1,185 @@
...
+        );
+        Ok(())
+    }
+}
+
+/// Compares the consensus material against the on-chain `SnsCiphertextMaterial`.
+///
+/// Only `keyId` and `snsCiphertextDigest` exist on-chain; `ciphertext_digest` and `format` are
+/// off-chain-only (bound by the signature).
+fn compare_onchain(
```
maksymsur, line 203: Shall we design a test for this? Also this comment is relevant for other heavy refactored modules.

maksymsur (commented):
Great work! Pls see some comments / concerns

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: Good point!
I had this in mind and wanted to handle it in a dedicated PR tackling this [issue](https://github.com/zama-ai/fhevm-internal/issues/1341), because it will be shared with ACL check, consensus check, context switch check etc.
OK for you?

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: Hmm, good catch, will try to use `try_join_all!` (same as `join_all`, but abort all tasks as soon as there is one error, which is what we want here IMO)

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: We have the [spawn_with_limit](https://github.com/zama-ai/fhevm/blob/e0b3f912f3d5e0f0ceb57544554bfee8137398a1/kms-connector/crates/utils/src/tasks.rs#L26) already used to limit the number tasks spawn for each decryption handling (via a semaphore), which is what we want to cap IMO.

And this `spawn_with_limit` would be hard to use for case where we need `join_all` kind of logic AFAIK.
Also, in this case, as the number of copro is limited, I think it's fine to keep it this way.

Returning the attestations as soon as consensus can be evaluated (i.e. as soon as we have at lease `threshold` attestations) seems a bit cumbersome for a 1st implem, because the error handling, i.e. resuming attestation fetching if consensus fails would be a bit messy to do.
Agreed that is what was suggested in the RFC, but would prefer to keep it this way for now (and update the RFC if needed).

Wdty?

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: Not sure what you meant by:

> Also this comment is relevant for other heavy refactored modules

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/registry.rs
eudelins-zama: You're right, let's add a guard :slightly_smiling_face:

eudelins-zama: I thought signers and tx_senders was coming from a shared struct at the contract level, and that in theory they should necessarily have the same length.
But not a big cost to add this guard

eudelins-zama: So indeed it should not be possible because of [this](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/GatewayConfig.sol#L987-L989).
But I'll add the guard anyway :slightly_smiling_face:

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: I keep my point on the tasks limitation, not super important IMO as bounded by small number of copro, and we already cap the number of tasks per decryption.

But maybe trying to evaluate consensus sooner is better in the end. Otherwise, one dead copro would slow down every single decryption indeed...

Will try to see if I can fix this properly

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
maksymsur: Yes, totally

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
maksymsur: I meant to see if there are other important but not covered by tests functions/functionality within this PR

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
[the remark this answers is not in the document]
eudelins-zama: Fixed bbfb78bc5bf849682f3f7bb54fea064620b3eda5

kms-connector/crates/kms-worker/tests/s3.rs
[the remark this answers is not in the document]
eudelins-zama: Fixed bbfb78bc5bf849682f3f7bb54fea064620b3eda5

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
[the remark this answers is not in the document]
eudelins-zama: Fixed bbfb78bc5bf849682f3f7bb54fea064620b3eda5

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: Fixed 4820c9592dc01466cbf78b05adaf97cb03f50493

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/registry.rs
eudelins-zama: Fixed 196b36af8d3d9c37d3858f33238661042b50923e

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: Fixed fc646456088d78e0e96102c5ecded859779b4162

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: Fixed d077dbb866f66f1640768562ade0cf3394b32509

I think we cover the main changes now.

- consensus evaluation is tested in `shared/ciphertext-attestation/src/consensus.rs`
- s3 refactor tested in both `kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs` and `kms-connector/crates/kms-worker/tests/s3.rs`
- on-chain comparison was added in commit mentioned above

maksymsur:
@claude /pr-review

maksymsur (approved):
LGTM! Unless agents find smth unexpected :)

shared/ciphertext-attestation/src/consensus.rs
```diff
@@ -0,0 +1,337 @@
...
+        }),
+        Some((_, signers)) => Err(ConsensusError {
+            valid_signers: signers.len(),
+            threshold,
+        }),
+        None => Err(ConsensusError {
+            valid_signers: 0,
+            threshold,
+        }),
+    }
```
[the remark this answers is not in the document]
eudelins-zama: On-going internal discussion on that topic, but currently we expect `threshold > (numCopros/2)` so not relevant.
Will be handled in follow up PR depending on internal discussion outcome

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -0,0 +1,388 @@
...
+
+    let old_url = format!("{s3_bucket_url}/{digest_hex}");
+    let new_url = rfc023_ciphertext_url(s3_bucket_url, sns_material.ctHandle);
+    let (ciphertext, ct_format) = match retrieve_ciphertext_via_http(client, &new_url).await {
+        Ok((ciphertext, ct_format)) => (ciphertext, ct_format),
+        Err(e) => {
+            warn!("Fetching via RFC-023 URL format failed: {e}. Falling back to old URL format");
+            retrieve_ciphertext_via_http(client, &old_url).await?
+        }
+    };
```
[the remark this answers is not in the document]
eudelins-zama: No, keeping the warning, as this should happen only if kms-connector is upgraded before copro, and that for a "short" time window

eudelins-zama:
@mergifyio queue

antoniupop (approved):

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -0,0 +1,319 @@
...
+    registry: CoprocessorRegistry<P>,
+
+    /// HTTP client for the attestation HEAD fan-out and the ciphertext retrieval.
+    client: Client,
+
+    /// Off-chain ciphertext-attestation config.
+    config: CtAttestationConfig,
+
+    /// Number of retries for S3 ciphertext retrieval.
+    s3_ciphertext_retrieval_retries: u8,
```
rudy-6-4, line 35: if it's a maximum number of retry, could be explicit in name.
e.g. s3_max_retrieval_retries
just reading I was wondering if it would be a kind of counter.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -0,0 +1,319 @@
...
+#[derive(Clone)]
+pub struct CiphertextManager<P: Provider> {
+    /// Periodically-synced mirror of the on-chain Coprocessor registry.
+    registry: CoprocessorRegistry<P>,
+
+    /// HTTP client for the attestation HEAD fan-out and the ciphertext retrieval.
+    client: Client,
+
+    /// Off-chain ciphertext-attestation config.
+    config: CtAttestationConfig,
```
rudy-6-4, line 32: name could be more explicit, e.g. verifier_config

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -0,0 +1,319 @@
...
+            self.verify_material_attestations(&registry, material)
+                .await
+                .with_context(|| {
+                    format!(
+                        "Attestation verification failed for handle {}",
+                        hex::encode(material.ctHandle)
+                    )
+                })
+        });
+        try_join_all(verification_futures).await?;
```
rudy-6-4, line 112: here could display (log or metric) the number of total failed material or it could be returned as result

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -0,0 +1,319 @@
...
+        )
+        .await
+    }
+
+    /// Verifies the off-chain attestations of `materials` against the on-chain snapshot.
+    ///
+    /// Fans HEAD requests out to every registered Coprocessor bucket and computes an off-chain
+    /// consensus verdict for each handle. Stops at the first failing material: a single failed
+    /// verification invalidates the whole decryption request.
+    pub async fn verify_attestations(
```
rudy-6-4, line 89: naming is bit vague regarding the level (several ct, several attestation per ct etc)
eg could be:
verify_attestations_batch_ciphertext
verify_attestations_ciphertext

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -0,0 +1,319 @@
...
+                    continue;
+                }
+                Ok((tx_sender, Err(e))) => {
+                    warn!(%tx_sender, %handle, "Failed to fetch attestation: {e}");
+                    continue;
+                }
+                Ok((tx_sender, Ok(attestation))) => attestations.push((tx_sender, attestation)),
+            };
+
+            if attestations.len() < registry.threshold.get() {
```
rudy-6-4, line 174: all failed attestation should be displayed ?
if one coproc is not pusing anything correct, and is the slowerto respond it could be unoticied ?

rudy-6-4: The loop could be independant and the functions returns early while the loop continue to display and meter error attestation error

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/registry.rs
```diff
@@ -0,0 +1,322 @@
...
+        )
+        .await
+        .map_err(RegistryError::Transient)?
+        .into_iter()
+        .flatten()
+        .collect();
+
+        if tx_sender_to_bucket.is_empty() {
+            return Err(RegistryError::Transient(anyhow::anyhow!(
+                "Not a single Coprocessor with a non-empty S3 bucket URL in the registry"
```
rudy-6-4, line 121: it won't work on a new chain ?

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -0,0 +1,388 @@
...
+use kms_grpc::kms::v1::{CiphertextFormat as GrpcCiphertextFormat, TypedCiphertext};
+use sha3::{Digest, Keccak256};
+use std::time::Duration;
+use tracing::{debug, info, trace, warn};
+
+/// The header used to retrieve the ciphertext format from the S3 HTTP response (pre RFC-023).
+const OLD_CT_FORMAT_HEADER: &str = "x-amz-meta-Ct-Format";
+
+/// URL of a ciphertext object in a Coprocessor bucket (RFC 023 layout).
+fn rfc023_ciphertext_url(bucket_url: &str, handle: B256) -> String {
```
rudy-6-4, line 32: Could be in shared, to share with copro.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -0,0 +1,388 @@
...
+    client: &Client,
+    copro_registry: &CoprocessorRegistrySnapshot,
+    sns_materials: &[SnsCiphertextMaterial],
+    retries: u8,
+) -> anyhow::Result<Vec<TypedCiphertext>> {
+    let fetch_futures = sns_materials.iter().map(|sns_material| {
+        retrieve_s3_ciphertext_with_retry(client, copro_registry, sns_material, retries)
+    });
+
+    try_join_all(fetch_futures).await
```
rudy-6-4, line 95: one failure makes everything fail ?

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -0,0 +1,388 @@
...
+/// Retrieves a ciphertext from S3 using the bucket URL and ciphertext digest.
+pub async fn retrieve_s3_ciphertext(
+    client: &Client,
+    s3_bucket_url: &str,
+    sns_material: &SnsCiphertextMaterial,
+    digest_hex: &str,
+) -> anyhow::Result<TypedCiphertext> {
+    let fhe_type = extract_fhe_type_from_handle(sns_material.ctHandle.as_slice())?;
+
+    let old_url = format!("{s3_bucket_url}/{digest_hex}");
```
rudy-6-4, line 160: to put with rfc023_ciphertext_url

kms-connector/crates/kms-worker/src/core/config.rs
```diff
@@ -89,6 +89,44 @@ pub struct Config {
...
+/// Configuration of the off-chain ciphertext-attestation verifier.
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[cfg_attr(debug_assertions, derive(Serialize))]
+#[serde(default)]
+pub struct CtAttestationConfig {
+    /// Whether the verifier runs at all. Defaults to `true`.
+    #[serde(default = "default_ct_attestation_verifier_enabled")]
+    pub enabled: bool,
+
+    /// Per-bucket S3 attestation HEAD request timeout, in seconds. Defaults to 5.
```
rudy-6-4, line 107: Is the default a bit short ?

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: It would make us having to wait for every material verification to complete, we want to fail fast IMO (current behavior)

shared/ciphertext-attestation/src/consensus.rs
```diff
@@ -0,0 +1,337 @@
...
+    handle: B256,
+    coprocessor_context_id: U256,
+    attestations: &[(Address, CiphertextAttestation)],
+    allowed_signers: &HashSet<Address>,
+    threshold: NonZeroUsize,
+) -> Result<Consensus, ConsensusError> {
+    let mut signer_groups: HashMap<ConsensusMaterial, HashSet<Address>> = HashMap::new();
+
+    for (tx_sender, attestation) in attestations {
+        if let Err(e) = attestation.verify(handle, coprocessor_context_id) {
```
rudy-6-4, line 68: this should still be part of the result ? and log at upper level ?
depend on how we want to collect statistics

kms-connector/crates/kms-worker/src/core/config.rs
eudelins-zama: Let's start with that and see, but don't think so

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: I think copro is using an S3 client, not raw http?

Let's do it in a follow up PR if needed, but I don't want to update copro now, and keep this PR connector focused

rudy-6-4 (approved):
LGTM. Except I am not sure the current version will log all the bad attestations due to suspected early exit (on attestations and on ciphertext batch).

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/registry.rs
eudelins-zama: I'm not sure I get the question.
We currently query the `GatewayConfig` contract (on Gateway) to update the registry because copro config currently lives there AFAIK.
We would probably use `ProtocolConfig` (on Ethereum) in the future.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: same

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: Yes, if we are not able to fetch every ciphertext of the decryption request, the whole decryption is considered as failed (on purpose). There is retry for each ciphertext download fyi

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: Keeping it aligned with main `Config`

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: > all failed attestation should be displayed ?

yes it's the case

> The loop could be independant and the functions returns early while the loop continue to display and meter error attestation error

Not sure what you mean

shared/ciphertext-attestation/src/consensus.rs
eudelins-zama: Will handle this as part of https://github.com/zama-ai/fhevm-internal/issues/1341

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: IMO, your suggestions duplicates the type of the input into the function name, but agree that it's not super clear as it is right now...

Keeping it as it is for now

eudelins-zama:
@mergifyio queue

PR #2825, opened 2026-06-18
feat(coprocessor): add unassociated bridged-handles metric

PanGan21:
### Summary

Expose a Prometheus gauge `coprocessor_bridge_unassociated_handles` from the confidential-bridge worker: the number of `handle_bridged_events` rows that are not yet associated and older than a short in-flight grace period (5 min), so normal handles still materializing in the regular flow are excluded.

The gauge is recomputed and set on every bridge poll (after the association pass), giving operators visibility into bridged handles that are not getting associated. A query failure is logged and swallowed so it never breaks the
association loop.

Related: https://github.com/zama-ai/fhevm-internal/issues/1527


coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
```diff
@@ -48,6 +48,18 @@ static BRIDGE_ERROR_COUNTER: LazyLock<IntCounter> = LazyLock::new(|| {
     .unwrap()
 });
 
+// Grace period before an unassociated handle is counted, so normal in-flight
+// handles (briefly unassociated while their ciphertext materializes) are excluded.
+const IN_FLIGHT_GRACE_SECS: f64 = 300.0;
```
dartdart26: Do we need a floating point number here? Can it be an int?

dartdart26 (approved):
Looks good! Just a nit on the floating point number.

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
PanGan21: Done, now I am just converting it during the sql query

PR #2830, opened 2026-06-18
chore(coprocessor): mount RDS CA bundle for db SSL

maxnovawind:
## Summary

Wires the bundled AWS RDS global CA (`configs/global-rds-ca-root.pem`) into the coprocessor chart so `databaseSslRootCertPath` actually resolves to a file present in the pod. Previously the chart injected the `DATABASE_SSL_ROOT_CERT_PATH` env var (IAM mode) but never mounted the certificate, so the path pointed at a non-existent file.

When `commonConfig.databaseSslRootCert.enabled` and `databaseAuthMode=iam`, the chart now:
- renders the PEM into a `<release>-rds-ca-cert` ConfigMap,
- mounts it read-only into all 10 workloads (9 deployments + db-migration Job),
- auto-derives `DATABASE_SSL_ROOT_CERT_PATH` from the mount path so it can never drift.

`databaseSslRootCertPath` remains as a bring-your-own-mount fallback. Password mode renders no cert artifacts.

## Validation

- `helm lint` clean, `helm template` renders with no errors.
- All components enabled: 10/10 workloads get the env + read-only mount.
- Fallback cases verified: password mode (no artifacts), IAM + mounting disabled with manual `value`/`valueFrom`.

Chart version bumped `0.13.0` → `0.14.0`.


piizama (dismissed):

piizama (dismissed):

Eikix (dismissed):

maxnovawind:
@mergify queue

maxnovawind:
@mergify queue

Eikix (dismissed):

Eikix (approved):

maxnovawind:
@mergify queue

PR #2848, opened 2026-06-19
feat(coprocessor): block context ciphertext materialisation part 1

antoniupop:


coprocessor/fhevm-engine/tfhe-worker/benches/utils.rs
```diff
@@ -151,12 +151,11 @@ pub async fn wait_until_all_allowed_handles_computed(
 
     loop {
         tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
-        let count = sqlx::query!(
-            "SELECT count(1) FROM computations WHERE is_allowed = TRUE AND is_completed = FALSE"
+        let current_count: i64 = sqlx::query_scalar(
```
rudy-6-4, line 156: non typed query 
query_scalar!

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -80,6 +95,147 @@ const MAX_RETRY_ON_UNKNOWN_ERROR: usize = 5;
...
+    is_allowed: bool,
+    schedule_order: PrimitiveDateTime,
+    producer_block: ProducerBlock,
+}
+
+async fn insert_computation_branch_row(
+    tx: &mut Transaction<'_>,
+    row: ComputationBranchRow<'_>,
+) -> Result<bool, SqlxError> {
+    let rows_affected = sqlx::query(
```
rudy-6-4: make all query static ? (can be done once the code is "definitive")

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -547,23 +729,401 @@ impl Database {
...
+            .into_iter()
+            .map(|row| row.block_hash)
+            .collect::<Vec<_>>();
+        if direct_orphaned_hashes.is_empty() {
+            return Ok(vec![]);
+        }
+
+        let orphaned_branch_rows = sqlx::query!(
+            r#"
+            WITH RECURSIVE orphaned_branch AS (
```
rudy-6-4, line 850: Maybe add a comment saying why it does not need to be bounded. "Marking all parents of new orphaned block as orphaned as well, this is limited to the size of the fork"

rudy-6-4: Also the motive because all these blocks are already marked as orphaned after a few time with existing code.
Here we want them to be marked the fastest because it make handle resolution faster ?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -547,23 +729,401 @@ impl Database {
...
+            return Ok(());
+        }
+
+        // Capture producer tuples whose ciphertext bytes were actually produced
+        // on an orphaned branch. ACL/PBS rows can now be orphaned by their event
+        // block while still referencing a canonical producer; those should not
+        // delete the canonical ciphertext bytes. Rows that reference an
+        // orphaned producer still drive byte cleanup, even when their own event
+        // block is not in the orphaned set.
+        let orphaned_ciphertext_pairs = sqlx::query(
```
rudy-6-4: I am not sure this cleanup should be done by the listener and in real time. (if the listener is shut down)
Is it just a best effort to make scheduling faster or does it break something if not done ?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -547,23 +729,401 @@ impl Database {
...
+        if orphaned_block_hashes.is_empty() {
+            return Ok(());
+        }
+
+        // Capture producer tuples whose ciphertext bytes were actually produced
+        // on an orphaned branch. ACL/PBS rows can now be orphaned by their event
+        // block while still referencing a canonical producer; those should not
+        // delete the canonical ciphertext bytes. Rows that reference an
+        // orphaned producer still drive byte cleanup, even when their own event
+        // block is not in the orphaned set.
```
rudy-6-4: what about S3 published elements ? there are published by handle now ?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -574,29 +1134,35 @@ impl Database {
...
         )
+        .bind(self.chain_id.as_i64())
+        .bind(block_summary.hash.to_vec())
+        .bind(block_summary.parent_hash.to_vec())
+        .bind(block_summary.number as i64)
+        .bind(status)
         .execute(tx.deref_mut())
         .await?;
 
         // 2. Update to finalized or orphan if needed
```
rudy-6-4: comment could be updated to say 
says it's also marking full branch as orphaned doing full cleanup
// 2. Update to finalized the block or orphan the full branch if needed
//      Clean all related content (computation, acl, ciphertext)

coprocessor/fhevm-engine/tfhe-worker/benches/utils.rs
rudy-6-4: ok it's bench, sorry

rudy-6-4 (commented):
is it planned to move to compiled sqlx query when it's stabilised ?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -80,6 +95,147 @@ const MAX_RETRY_ON_UNKNOWN_ERROR: usize = 5;
...
+            is_scalar,
+            dependence_chain_id,
+            transaction_id,
+            is_allowed,
+            created_at,
+            schedule_order,
+            is_completed,
+            host_chain_id,
+            block_number,
+            producer_block_hash
```
rudy-6-4, line 133: why it's not block_hash alone ?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -808,24 +1476,38 @@ impl Database {
         match data {
             AclContractEvents::Allowed(allowed) => {
                 let handle = allowed.handle.to_vec();
+                let producer_block = self
```
rudy-6-4, line 1674: why is it needed here while it's not for computations ?
what if it cannot be resolved ? what is the effect on downstream ?

coprocessor/fhevm-engine/db-migration/migrations/20260610130100_branch_context_tables.sql
```diff
@@ -0,0 +1,238 @@
...
+
+-- ============================ computations ============================
+
+CREATE TABLE IF NOT EXISTS computations_branch
+(
+    LIKE computations INCLUDING DEFAULTS INCLUDING GENERATED INCLUDING IDENTITY INCLUDING STORAGE INCLUDING COMMENTS
+);
+
+ALTER TABLE computations_branch
+ADD COLUMN IF NOT EXISTS producer_block_hash BYTEA NOT NULL DEFAULT ''::BYTEA;
```
rudy-6-4, line 28: not consistent with acl where it's named block_hash

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: That is a good point and maybe we need to have a separate service do this. The main reason this is here is the host listener is already the component that discovers block ancestry, marks blocks finalized/orphaned, and knows the canonical hash from RPC. 
Right now after finalizing a block we immediately call cleanup in the same DB transaction so we have an invariant in DB state that once host_chain_blocks_valid says a branch is orphaned, the branch-scoped state has been cleaned.

Do you think this is not acceptable in the listener?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: block_hash and producer_bloc_hash mean different things. block_hash is the host block where the event was observed, for example an allow event. producer_block_hash is the block that produced the ciphertext for the handle. For ACL/PBS rows those can differ and RFC 011 depends on these.

coprocessor/fhevm-engine/db-migration/migrations/20260610130100_branch_context_tables.sql
antoniupop: see above

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: For normal computations, the worker/listener is inserting metadata for an output produced by the current block execution, so the producer block is known directly. it's the current computation block.

For allowed, the allowed event block is not necessarily the block that produced that handle’s ciphertext. If we stored the ACL event block as the producer, downstream PBS/worker lookup would try to fetch (handle, acl_block_hash), which may not exist or give wrong ciphertext provenance.

If producer resolution cannot find a branch producer, we key the row branchless, emit UNRESOLVED_PRODUCER_BLOCK_TOTAL, and log an error. Downstream then treats it as the compatibility/branchless path. This is conservative, we may lose precise fork disambiguation or hit availability/missing-ciphertext behavior, but we don't attribute ciphertext bytes to the ACL block.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: Yes absolutely - this is detailed in RFC 23, we allow for early consensus even if not canonical as cleartext is guaranteed identical, with a remediation in wave 2 that ensures that eventually we converge to consensus on canonical CT+attestation

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: Done asnyc in wave 2 now.

coprocessor/fhevm-engine/db-migration/migrations/20260610130200_branch_component_acquisition_index.sql
```diff
@@ -0,0 +1,6 @@
+-- Branch workers acquire one ingestion-time same-block component per
+-- dependence_chain_id and fetch all non-error rows in that component, including
+-- already completed same-block producers needed for in-memory forwarding.
+CREATE INDEX IF NOT EXISTS idx_computations_branch_component_rows
+ON computations_branch (dependence_chain_id, schedule_order)
+WHERE is_error = false;
```
dartdart26, line 6: Do we need a separate migration file for this?

coprocessor/fhevm-engine/fhevm-engine-common/src/pg_pool.rs
```diff
@@ -69,6 +69,18 @@ impl PostgresPoolManager {
                 .after_connect(move |conn, _meta| {
                     info!(auto_explain = ?auto_explain_with_min_duration, "New DB connection established");
                     Box::pin(async move {
+                        // Bound every statement on this pooled connection at 10s,
+                        // matching the host-listener. This caps the wave1
+                        // ciphertext_digest mirror-trigger fan-out (COUNT + DISTINCT
+                        // loop + N upserts) fired from sns-worker, so a pathological
+                        // reorg fan-out cannot hold row locks unbounded. Non-fatal:
+                        // a failed SET just leaves the connection without the bound.
```
dartdart26: This comment will be hard to maintain as it is referencing host-listener, in general, it is specific to this PR. I'd simplify it.

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -180,9 +182,16 @@ pub async fn promote_once_all_chains_to_fast(
     dependent_ops_max_per_chain: u32,
 ) {
     if dependent_ops_max_per_chain == 0 {
+        if SLOW_LANE_PROMOTION_ATTEMPTED
+            .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
+            .is_err()
+        {
+            return;
+        }
```
dartdart26, line 191: I am not sure I understand this change. Could you please explain it?

dartdart26:
Are SAVEPOINTs actually needed or an overkill? Do we expect any queries/inserts to new tables to fail?

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
```diff
@@ -313,10 +316,13 @@ async fn main() -> anyhow::Result<()> {
...
 
     let database_url = resolve_database_url_from_option(conf.database_url)?;
-    let (db_pool, _pool_refresh_handle) = connect_pool_with_options(
+    let (db_pool, _pool_refresh_handle) = connect_pool_with_options_and_connect_options(
         &database_url,
         sqlx::postgres::PgPoolOptions::new().max_connections(conf.database_pool_size),
         Some(&cancel_token),
+        // Bound the wave1 ciphertext_digest mirror-trigger fan-out fired from
+        // this writer; without it the trigger has no DB-level cancellation.
+        with_statement_timeout_10s,
```
dartdart26: Hard to maintain that comment. Maybe timeout should be made more generic.

dartdart26 (commented):
I've gone through the PR and I've added some minor comments. I need more time for a deeper review.

That said, I am concerned it would be easy to miss small things, e.g. data type conversions, some Rust intricacy.

antoniupop:
> Are SAVEPOINTs actually needed or an overkill? Do we expect any queries/inserts to new tables to fail?

Yes, I meant to clean this up, but I missed it. Removing.

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
antoniupop: This is a guard for avoiding dependence chains staying stuck in slow lane when we restart with the feature disabled. If we do, we need to ensure we successfully promoted all chains to fast or we'd potentially deadlock the copro.

coprocessor/fhevm-engine/db-migration/migrations/20260610130200_branch_component_acquisition_index.sql
antoniupop: Would you prefer if I consolidate all migrations in a single file?

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
```diff
@@ -180,9 +182,16 @@ pub async fn promote_once_all_chains_to_fast(
     dependent_ops_max_per_chain: u32,
 ) {
     if dependent_ops_max_per_chain == 0 {
+        if SLOW_LANE_PROMOTION_ATTEMPTED
```
rudy-6-4, line 186: should be attempted after any drift ? to be consistent with other listener ?
https://github.com/zama-ai/fhevm/pull/2697
(the consumer is not restarted on drift)

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -287,6 +262,34 @@ pub async fn ingest_block_logs(
         }
     }
 
+    for (event, transaction_hash) in acl_event_log {
```
rudy-6-4, line 429: was there a reason to move it after tfhe insertions ? (not against, just checking if a comment is missing)

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -164,14 +392,72 @@ impl Database {
...
+    /// Block until the wave-1 branch-context schema is present (the `*_branch`
+    /// tables, `coprocessor_settlement`, and `host_chain_blocks_valid.parent_hash`).
+    /// This degrades a pre-migration start to a bounded wait instead of a
+    /// crash-loop. Bounded so a genuinely-absent migration surfaces as an error
+    /// (→ restart) rather than hanging forever.
+    pub async fn wait_for_branch_schema(&self) -> Result<()> {
+        const MAX_ATTEMPTS: u32 = 60; // ~2 minutes at one poll / 2s
+        const POLL_INTERVAL: Duration = Duration::from_secs(2);
+        let mut attempt: u32 = 0;
+        loop {
```
rudy-6-4: for loop ?

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
antoniupop: You're right, missed this. This guard only fires once per process, so it won't re-run after a drift (the consumer isn't restarted on drift as you note). It's a no-op in practice while slow-lane is disabled (revert doesn't touch dependence_chain, new chains default to Fast), but it diverges from the other listeners. Should I switch it to the  KnownDrift flag strategy to match?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -80,6 +96,217 @@ const MAX_RETRY_ON_UNKNOWN_ERROR: usize = 5;
...
+    handle: &[u8],
+    transaction_id: Option<Vec<u8>>,
+    block_number: i64,
+    block_hash: &[u8],
+    producer_block_hash: &[u8],
+) -> Result<bool, SqlxError> {
+    // Wave-1: isolate in a savepoint (see insert_computation_branch_row) so a
+    // failure (incl. the digest-mirror trigger this insert fires) never aborts
+    // the caller's authoritative legacy pbs_computations write.
+    sqlx::query("SAVEPOINT branch_dual_write")
```
rudy-6-4: (not asking any change) SAVEPOINT are done at the event level. Is it to reduce code change ? This will have an impact in term of performance ? (3 times more statement). Maybe it's time to plan bulk insert.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -547,23 +857,449 @@ impl Database {
         tx: &mut Transaction<'_>,
         block_number: i64,
         block_hash: &BlockHash,
-    ) -> Result<(), SqlxError> {
-        sqlx::query!(
+    ) -> Result<Vec<Vec<u8>>, SqlxError> {
+        let finalized_result = sqlx::query(
```
rudy-6-4: query! when it's stabilized.

rudy-6-4 (commented):
New data table fill can fail partially. eg some operation or allow of a block can be inserted while other in the same block fail without failing the full block thx to savepoint. Why are the new insertion considered not part of the transaction ?

antoniupop:
> New data table fill can fail partially. eg some operation or allow of a block can be inserted while other in the same block fail without failing the full block thx to savepoint. Why are the new insertion considered not part of the transaction ?

This is already fixed I think. The savepoints were removed.

coprocessor/fhevm-engine/db-migration/migrations/20260610130200_branch_component_acquisition_index.sql
dartdart26: I guess the question is if we want to deploy these separately. I guess not, but wdyt?

coprocessor/fhevm-engine/host-listener/src/consumer/mod.rs
rudy-6-4: No, let's keep it simple for this PR. (we can kill the consumer if we really need a slow lane at some point). When do we need slow-lane ?

antoniupop: We need slow lane if there's a dos or other excessive activity on some contract...

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
antoniupop: you're right, this isn't obvious - I'll add a comment. this is needed because if both producer compute event and allow are in the same block, then the producer won't be found if we process acl events immediately.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: fixed/removed

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: I'll re-check if any others are leftover, thanks!

coprocessor/fhevm-engine/db-migration/migrations/20260610130200_branch_component_acquisition_index.sql
antoniupop: If you're worried about the length of index creation migrations, there's an out-of-band index concurrent creation planned for the one that would take long otherwise, the others (like this one) are noops as the table is new. I'm happy to batch them or keep split, no preference.

dartdart26: Let's keep as is.

rudy-6-4 (dismissed):

rudy-6-4 (commented):
Shouldn't bridge_handle_events have a block hash field ?
They are inserted in handle_bridge_event.

antoniupop:
> Shouldn't bridge_handle_events have a block hash field ? They are inserted in handle_bridge_event.

Good point! I'll add that.

dartdart26 (dismissed):
Since it is a relatively big PR, I focused on the fact that readers (services other than host-listener) are unchanged. Therefore, the biggest risk is in the migration(s) and I didn't see an issue there myself. At the same time, there are a number of SQL statements and I tried to verify line by line, but I could have missed something.

I am approving it, with the above caveat.

Good work!

rudy-6-4 (dismissed):

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio refresh

antoniupop:
@mergifyio update

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

PanGan21 (dismissed):

dartdart26 (commented):
Looks good! However, I have a question - will bridging now require to wait for finalized blocks on both host and destination? If yes, I am concerned if that was baked in into the RFC and if such latency is expected.

dartdart26 (dismissed):

dartdart26 (dismissed):

antoniupop:
@mergifyio queue

obatirou (dismissed):

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

obatirou (dismissed):

PanGan21 (approved):
The latest change looks correct to me

rudy-6-4 (approved):

dartdart26 (approved):

dartdart26:
Good catch on the S3 format @antoniupop, thanks!

PR #2849, opened 2026-06-19
feat(coprocessor): block context ciphertext materialisation part 2

antoniupop:
## Summary

  This PR implements wave 2 of block-scoped ciphertext materialisation and coprocessor execution, see details in RFCs 11, 19 and 20 for context. Computations are keyed by their host-chain block context, executed deterministically across coprocessors, settled against the finalized canonical chain, and only then published externally.

  The main goal is to make ciphertext state consensus-safe across forks and reorgs while improving scheduler parallelism, liveness, recovery, and blue/green rollout safety.

  ## Execution flow

```mermaid
 flowchart TD
      HC[Host-chain events] --> HL[Host listener]
      HL -->|Materialize block and fork provenance| BW[(Branch work tables)]

      BW --> TW[TFHE worker]
      TW -->|Build dependency components| SC[Parallel scheduler]
      SC -->|Execute, re-randomize and compress| CT[(Branch ciphertexts)]

      CT --> SH[Block state hash]
      SH --> CD[Consensus detector]
      CD --> BG[Blue-green cutover authorization]

      CT --> ST[Canonical settlement]
      ST --> SNS[SNS worker]

      SNS -->|Settled ciphertext and attestation| S3[Canonical S3 storage]
      SNS --> TS[Gateway transaction sender]

      S3 --> DD[Drift and publication verification]
      TS --> DD
      DD -->|Repair or revert| ST
```
  During a blue/green dry run, BCS remains live while GCS executes the new block-scoped semantics. GCS state hashes must agree across operators, but its ciphertext bytes may intentionally differ from BCS. GCS external publication remains fenced until the actual stack cutover.

  ## Changes by component

  ### Database and common lifecycle

  - Adds branch-context settlement state and canonical block progression.
  - Tracks computation provenance by chain, block number, and block hash.
  - Adds migrations and indexes for context acquisition, settlement, cleanup, terminal-error roots, S3 repair, and publication verification.
  - Supports migration/backfill from existing branch data and quarantines unrecoverable legacy rows.

  ### Host listener

  - Materializes computation, PBS, ACL, and dependence-chain rows with their producing block context.
  - Dual-writes legacy and branch state before the semantic cutover, then stops feeding legacy work tables.
  - Tracks finality and reorgs so only the canonical contiguous block sequence is settled.
  - Uses batched ingest queries and explicit provenance validation to reduce ingest overhead and prevent ambiguous branch state.

  ### Scheduler and TFHE worker

  - Acquires work by block context and executes dependency components as atomic batches.
  - Canonicalizes cross-transaction and cross-block dependencies before scheduling.
  - Uses shared ciphertext ownership for intra- and inter-partition fan-out instead of deep-cloning per consumer.
  - Adds lease heartbeats, fairer acquisition, bounded batching, and scheduler liveness safeguards.
  - Applies deterministic block-context re-randomization and compression.
  - Terminalizes deterministic invalid computations and propagates their failure through dependent work.
  - Keeps missing inputs and unclassified infrastructure panics retryable, preventing node-local failures from becoming divergent consensus state.

  ### SNS, S3, and drift recovery

  - Publishes only ciphertexts belonging to settled canonical contexts.
  - Produces signed, context-aware canonical S3 objects and verifies publication before marking work complete.
  - Adds idempotent recovery and repair queues for uploads interrupted by crashes, cutover, or reorgs.
  - Extends drift detection and auto-revert handling for block-scoped digests and expected blue/green window differences.
  - Fences GCS publication during the dry run and drains the accumulated publication backlog after cutover.

  ### Configuration and rollout

  - Registers the new worker, batching, cutover, settlement, and cleanup options in the coprocessor Helm values.
  - Documents the required blue/green cutover coordinate, BCS drain barrier, GCS side-effect fencing, and post-cutover repair process.
  - Keeps SQLx metadata centralized in the workspace .sqlx directory.

  ## Commit guide

  ### feat(coprocessor): define wave2 branch lifecycle state

  Introduces the shared lifecycle and persistence foundation: settlement APIs, branch-context state, migration/backfill support, acquisition indexes, terminal-error roots, cleanup jobs, and S3 publication/repair queues.

  ### feat(coprocessor): materialize and settle host block contexts

  Updates the host listener to attach canonical block and fork provenance to all relevant work. It implements branch-aware ingest, cutover-controlled legacy mirroring, batched inserts, reorg handling, finality tracking, and contiguous settlement advancement.

  ### feat(coprocessor): execute block contexts as atomic batches

  Implements the core TFHE execution path. It groups work by block context, constructs dependency components, schedules them in parallel, performs deterministic re-randomization and compression, and commits their results consistently. It also contains the scheduler liveness, lease renewal, ciphertext-sharing, transient-panic, and terminal-error propagation fixes.

  ### feat(coprocessor): publish settled ciphertexts canonically

  Implements settled-only publication across the SNS worker, S3 uploader, transaction sender, attestations, and drift detector. It adds canonical context-aware storage, publication verification, recovery/quarantine, repair after cutover, and reorg-safe external side effects.

  ### test(coprocessor): cover the wave2 branch lifecycle

  Adds Rust integration coverage for block-scoped execution, branch selection, settlement, reorgs, scheduler liveness, terminal errors, carry cleanup, bridge handling, publication recovery, and drift reversion.

  ### test(e2e): exercise multi-coprocessor branch consensus

  Adds two-of-three and three-of-three coprocessor scenarios, real-fork execution, state-hash consensus checks, watchdog coverage, drift handling, readiness diagnostics, and multi-chain isolation coverage.

  ### docs(coprocessor): document wave2 rollout and configuration

  Adds the operator runbook, legacy-table retirement plan, Helm visibility for all new arguments, and the required blue/green rollout and validation steps.

  ## Testing

  Coverage added across:

  - Host-listener and TFHE-worker Rust unit and integration tests.
  - Scheduler liveness and deterministic/transient error handling.
  - SNS publication, recovery, and S3 repair.
  - Fork, reorg, settlement, and drift-revert paths.
  - Multi-coprocessor e2e consensus and blue/green scenarios.
  - SQLx offline query validation and Helm configuration rendering.


coprocessor/fhevm-engine/db-migration/migrations/20260610145100_branch_event_block_hash.sql
dartdart26, line 148: This is hard to reason about for me personally. Namely, I miss the why and also what is racing with what?

Or to put it differently, can it be avoided, at least for the sake of simplicity?

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
```diff
@@ -561,6 +561,31 @@ impl SupportedFheCiphertexts {
...
+            SupportedFheCiphertexts::FheUint64(c) => SupportedFheCiphertexts::FheUint64(c & c),
+            SupportedFheCiphertexts::FheUint128(c) => SupportedFheCiphertexts::FheUint128(c & c),
+            SupportedFheCiphertexts::FheUint160(c) => SupportedFheCiphertexts::FheUint160(c & c),
+            SupportedFheCiphertexts::FheUint256(c) => SupportedFheCiphertexts::FheUint256(c & c),
+            SupportedFheCiphertexts::FheBytes64(c) => SupportedFheCiphertexts::FheBytes64(c & c),
+            SupportedFheCiphertexts::FheBytes128(c) => SupportedFheCiphertexts::FheBytes128(c & c),
+            SupportedFheCiphertexts::FheBytes256(c) => SupportedFheCiphertexts::FheBytes256(c & c),
+        })
+    }
+
```
dartdart26: I think I've missed that part and, if yes, sorry about it. But could you please share the justification for it?

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
```diff
@@ -277,34 +277,112 @@ impl DriftDetector {
             return Ok(());
         };
 
-        let row = sqlx::query(
-            "SELECT ciphertext, ciphertext128 FROM ciphertext_digest WHERE handle = $1",
+        // Prefer local rows for finalized event blocks. In a fork scenario the
+        // same handle can exist under different producer/event block hashes; an
+        // off-branch row that happens to match consensus must not mask drift on
+        // the finalized branch. Fully branchless rows remain a compatibility
+        // fallback only when no event-scoped row exists for this handle.
```
dartdart26: I can't easily parse this comment and the justification. Maybe could be useful to rewrite it in simpler language as I think it is important and will be easier to maintain.

Some suggestions and questions:
 * Maybe we can use producer block hashes only and not talk about producer/event to reduce confusion.
 * Not sure what off-branch row means to be honest, now that I think of it.
 * Fully branchless row - not sure what it means.

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
antoniupop: This is due to tfhe-rs making compression error out with CiphertextCompressionRequiresEmptyCarries - this used to be initially more relaxed, but then when this tightening was introduced (and currently in main for copro) we didn't have the issue because we (de/)compress all ciphertexts in-between every operation, so the case can't arise where an op without final PBS leaves a raised degree that errors on compress.
However now that we forward ciphertexts raw from one op to the next again (as we do block scope materialisation), this can happen. This only does the clearing conservatively on compress() erroring, not all the time, so we don't pay unless it's needed. I'm adding a test to cover this just to make sure we don't lose track of the issue over time.

cc @IceTDrinker in case I misunderstood something! Cheers!

coprocessor/fhevm-engine/gw-listener/src/drift_detector.rs
antoniupop: Ok - I'll try to simplify wording... Branchless and off-branch refer to either legacy DB tables where we're not branch-aware, or otherwise when there's no block/branch possibly associated (e.g. ZK / inputs).
Would this be clear enough as:

```
  // For handles with event scoped digest rows, compare only rows whose event
  // block is finalized on the canonical branch. The same handle may also have
  // digest rows from sibling/orphaned blocks; those must not satisfy the local
  // comparison, because they could hide drift in the finalized branch's row.
  //
  // A fully branchless digest row (producer_block_hash = '' and block_hash = '')
  // has no branch/event context, so use it only as a compatibility fallback when
  // this handle has no event-scoped digest rows at all.
```

coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
IceTDrinker: @antoniupop ciphertexts should have clean carries after operations, did you find some that don't keep them clean ? If so we have a bug and we'll fix it ASAP in a patch release for 1.6

IceTDrinker: Also, if you compressed every time before and did not encounter the bug, you should not encounter it in the future.

We can likely make sure the compression routine takes care of that cleaning on your behalf if you had actual problems in the past

IceTDrinker: Though having the signal you had to do something can help pinpoint ops that don't respect the invariant, so pondering a bit about this

antoniupop: Sharing a repro and moving convo on Slack.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -1055,6 +1422,245 @@ impl Database {
...
+            };
+
+            match self
+                .process_orphaned_branch_cleanup_job(
+                    &finalized_block_hash,
+                    &orphaned_hashes,
+                )
+                .await
+            {
+                Ok(()) => {
```
rudy-6-4: even if it immediately fail with the guarded call

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -1055,6 +1422,245 @@ impl Database {
...
+
+    async fn process_orphaned_branch_cleanup_job(
+        &self,
+        finalized_block_hash: &[u8],
+        orphaned_hashes: &[Vec<u8>],
+    ) -> Result<(), SqlxError> {
+        let Some(mut tx) = self.new_transaction().await? else {
+            // Write-guarded (GCS drain): leave the job pending; the poller
+            // retries it once writes are allowed again.
+            return Ok(());
```
rudy-6-4: could return Ok(false) so the caller knows it cannot call it again immediately and that nothing was done ?.

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -1408,6 +2547,128 @@ async fn check_ct128_objects_exist(
...
+        expected_signer,
+        expected_checksum_sha256,
+        AttestationBinding::ContentBound,
+    )
+    .await?;
+
+    match (key_exists, digest_key_exists) {
+        (true, true) => Ok(true),
+        (false, true) => {
+            // The digest copy's attestation may be bound to another handle
```
rudy-6-4: This copy was proposed to be deleted on 0.15. 
https://github.com/zama-ai/fhevm/pull/3206, we can postpone that PR if needed as it's just cleaning. But I think we should keep the copy with a more clear access (eg by handle), by having a unique key handle/block_hash or handle/digest so both branch can live on S3 and we can always recover. And for KMS  we have a fixed key handle/finalized.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -349,512 +701,4463 @@ async fn tfhe_worker_cycle(
...
     }
-    if let Err(e) = tx_graph.build(txs) {
-        // If we had an error while building the graph, we don't
-        // execute anything and return to allow any set results
-        // (essentially errors) to be set in DB.
-        warn!(target: "tfhe_worker", { error = %e }, "error while building transaction graph");
-        return Ok(tx_graph);
+
+    let mut computation_rows = Vec::new();
+    let rows = sqlx::query!(
```
rudy-6-4, line 1051: could pick a better name here for the poor reviewer ?

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -349,512 +701,4463 @@ async fn tfhe_worker_cycle(
...
+        r#"
+        SELECT output_handle AS "handle!", producer_block_hash AS "producer_block_hash!", block_number
+        FROM computations_branch
+        WHERE host_chain_id = $2
+        AND output_handle = ANY($1::BYTEA[])
```
rudy-6-4, line 1056: We can pick "futur" handle, ?
AND (
            $3::BIGINT IS NULL
            OR block_number IS NULL
            OR block_number <= $3
        )
         "#,
         cts_to_query as _,
         batch_context.host_chain_id,
        batch_context.block_number,

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -349,512 +701,4463 @@ async fn tfhe_worker_cycle(
...
+            CASE
+                WHEN a.producer_block_hash = ''::BYTEA THEN a.block_number
+                ELSE producer.block_number
+            END AS block_number
+        FROM allowed_handles_branch a
+        LEFT JOIN host_chain_blocks_valid producer
+          ON producer.chain_id = a.host_chain_id
+         AND producer.block_hash = a.producer_block_hash
+        WHERE (a.host_chain_id = $2 OR a.host_chain_id IS NULL)
+        AND a.handle = ANY($1::BYTEA[])
```
rudy-6-4: here filter to avoid futur handle as well ?
+        AND (
+            $3::BIGINT IS NULL
+            OR a.block_number IS NULL
+            OR a.block_number <= $3
+        )
+        AND (
+            $3::BIGINT IS NULL
+            OR producer.block_number IS NULL
+            OR producer.block_number <= $3
+        )

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -349,512 +701,4463 @@ async fn tfhe_worker_cycle(
...
+        UNION ALL
+        SELECT
+            a.handle AS "handle!",
+            ''::BYTEA AS "producer_block_hash!",
+            a.block_number
+        FROM allowed_handles a
+        WHERE (a.host_chain_id = $2 OR a.host_chain_id IS NULL)
+        AND a.handle = ANY($1::BYTEA[])
+        AND NOT EXISTS (
```
rudy-6-4, line 1130: Filter here also ?

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -349,512 +701,4463 @@ async fn tfhe_worker_cycle(
...
+            SELECT 1
+            FROM allowed_handles_branch b
+            WHERE b.host_chain_id = a.host_chain_id
+              AND b.handle = a.handle
+              AND b.account_address = a.account_address
+              AND b.event_type = a.event_type
```
rudy-6-4, line 1136: and here ?

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
rudy-6-4: Ok it's a little far fetched given how we generate handle, same handle can only exits at same height except maybe for trivial input and input.

coprocessor/fhevm-engine/db-migration/migrations/20260710120000_sns_recovery_quarantine.sql
```diff
@@ -0,0 +1,60 @@
...
+    (
+        is_error = FALSE
+        AND error_message IS NULL
+        AND error_at IS NULL
+    )
+) NOT VALID;
+
+DROP INDEX IF EXISTS idx_pbs_computations_branch_pending_created_at;
+
+CREATE INDEX idx_pbs_computations_branch_pending_created_at
```
rudy-6-4: Is it ok to not do concurrenlty ? (small table for now)

rudy-6-4 (commented):
The real-fork tests could be required  by CI /home/rudy/zama/fhevm/test-suite/e2e/test/consensus/realForkConsensus.ts:588 into required CI, to avoid any manual error later.

What could be added to these tests:
(C3b) Add decryption of the pre-reorg canonical ciphertext after recovery.
(C3) Assert explicitly that the winning block is finalized

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: fixed

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: fixed

coprocessor/fhevm-engine/db-migration/migrations/20260710120000_sns_recovery_quarantine.sql
antoniupop: Normally all indexes are pre-created concurrently off-migration in initialize_db. If this was missed will add it. Once it's pre-created, this migration is a nop, just needed for consistency on creating new env without running the off-migration script.

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
antoniupop: I'd prefer to avoid the extra complexity of changing the overwrite model - instead I'm adding a different repair scheme that re-computes ct128 in case repair is needed and it's been deleted locally - should not happen often enough to be an issue. If the digest doesn't match it's flagged error, but this shouldn't happen unless something else has broken (e.g. upgrade process).

antoniupop: I'll leave this unresolved until you confirm if this is OK - feel free to resolve otherwise

coprocessor/fhevm-engine/db-migration/migrations/20260710120000_sns_recovery_quarantine.sql
antoniupop: Well-spotted! It was missing indeed. Fixing along with a second that was missing.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
antoniupop: No, it's a very good point - fixing, and also the following ones.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
antoniupop: Yes indeed

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
antoniupop: done - and following one too

antoniupop:
> The real-fork tests could be required by CI /home/rudy/zama/fhevm/test-suite/e2e/test/consensus/realForkConsensus.ts:588 into required CI, to avoid any manual error later.
> 
> What could be added to these tests: (C3b) Add decryption of the pre-reorg canonical ciphertext after recovery. (C3) Assert explicitly that the winning block is finalized

Very good suggestions, thanks! I'm adding the new tests.

For the CI requirement I need to add a new flow as it requires a different topology than existing. We can enforce this in Mergify if you think it's appropriate.

Eikix:
Some finding (I drove personally)

AI summary:

## RPCs do NOT broadcast competing forks in parallel

**Alchemy docs** (verbatim):
> "When a chain reorganization occurs, this subscription will emit an event containing all new headers (blocks) for the new chain. This means that you may see **multiple headers emitted with the same height**, and when this happens the **later** block number should be taken as the correct one."

**Infura docs** (verbatim):
> "In a chain reorganization, the subscription emits all new headers for the new chain. Therefore the subscription can emit **multiple headers at the same height**."

**go-ethereum docs** (verbatim):
> "In case of a chain reorganization the subscription will emit **the last header in the new chain**."

The key word is **"the new chain"** — singular. The node has already run its fork choice rule (LMD-GHOST) and decided which fork is canonical. It then notifies you of the winner. You never see both forks "live."

---

## What you actually observe during a reorg

From the [StackExchange example](https://ethereum.stackexchange.com/questions/99514) (real logs):

```
Added block #7665191 | hash 0xe550...  ← fork A
Added block #7665191 | hash 0x068f...  ← fork B (reorg! same height, different hash)
Added block #7665192 | hash 0x36c2...  ← fork B continues (this is canonical)
```

You see fork A's block, then fork B's block at the same height, **sequentially**. The node already decided fork B wins. You discover the reorg by checking: "wait, this block's parent doesn't match my history."

---

## How the PR determines canonicality

It does **not** check attestations. It trusts the RPC implicitly:

```rust
// update_finalized_blocks_aux — the "canonicality check"
get_block_hash_by_number(block_number as u64).await
```

That's it. A by-number RPC lookup. Whatever hash the node returns for that height IS the canonical block. No attestation verification, no consensus-layer query, no fork-choice inspection.

---

## The implication for RFC 011

The "track all active forks simultaneously" model **cannot work via standard RPC** because the RPC never gives you both forks at the same time. The sequence is:

1. You see fork A's blocks (the node thought A was canonical)
2. The node's fork choice switches to B
3. You see fork B's blocks (the node now says B is canonical)
4. You discover the reorg by parent-hash mismatch

At no point do you have "both forks live." You have fork A's state in your DB (from step 1), and then fork B's blocks arrive (step 3). The RFC 011 model of "keep both, settle later" is just... keeping fork A's stale state in the DB until you're sure fork B won. Which is exactly what the revert model does too — except the revert model *deletes* fork A's state immediately instead of waiting for settlement.

Eikix:
Another question I had for the same height fork problem:

## If re-randomization uses prev_block_hash

Two competing blocks at height N, both children of block N-1 (hash `0xPP`):

```
Block N-1 (hash 0xPP)
   ├── Block N fork A (hash 0xAA, prev = 0xPP)
   └── Block N fork B (hash 0xBB, prev = 0xPP)
```

- Handle derivation: both use `blockhash(N-1)` = `0xPP` → **same handle** ✓ (already the case)
- Re-randomization with prev_block_hash: both seed with `0xPP` → **same CT bytes**
- **No conflict. No fork problem. One ciphertext per handle works.**

At height N+1:
- Fork A: prev = `0xAA` → handle H_A, seed = `0xAA`
- Fork B: prev = `0xBB` → handle H_B, seed = `0xBB`
- Different handles → no collision regardless of seed choice

**The fork problem vanishes entirely.** You don't need `(handle, producer_block_hash)` keying. You don't need branch tables. You don't need settlement. You don't need the +26K PR.

---

## Does it preserve the required properties?

| Property | current_block_hash | prev_block_hash |
|---|---|---|
| Cross-operator determinism | ✓ (all operators see same block hash) | ✓ (all operators see same parent hash) |
| Per-block freshness (different heights → different bytes) | ✓ | ✓ (different heights have different parents) |
| External verifiability | ✓ (block hash is in the header) | ✓ (parentHash is IN the block header) |
| Unlinkability across blocks | ✓ | ✓ (different heights → different seeds) |
| Same handle, same block, different forks → same bytes | ✗ (THIS IS THE PROBLEM) | **✓** |

prev_block_hash satisfies every property that current_block_hash does, **and additionally eliminates the fork divergence.**

---

## Why does this work?

Because of a structural property of forks: **competing blocks at the same height share the same parent.** The fork point is the parent. So `prev_block_hash` is identical across all forks at the fork height. It's the one piece of block context that is **fork-invariant**.

current_block_hash is fork-VARIANT (different forks have different hashes at the same height). That's what creates the problem.

prev_block_hash is fork-INVARIANT (all forks at height N agree on block N-1's hash). That's what eliminates it.

---

So why did RFC 019 use current_block_hash?

antoniupop:
> So why did RFC 019 use current_block_hash?

That (prev block hash) was discussed, but it does not satisfy the sind-cpad requirements. It was not approved as safe.

antoniupop:
> Some finding (I drove personally)
> 
> AI summary:
> 
> ## RPCs do NOT broadcast competing forks in parallel
> [...]

Correct about the RPC behavior — a node runs fork choice internally and only ever emits the winner's headers; we discover reorgs by parent-hash mismatch, exactly as the trace shows. Nothing in RFC-011 assumes otherwise.

But "track multiple forks" is a claim about storage, not the wire protocol. If fork A's block at height N is ingested — and, since we compute pre-finality, its ciphertexts produced — before fork B's sibling arrives, the DB holds both regardless of how sequentially they were observed. The only question is what to do with the loser's rows, and the equivalence "keep-until-settled = revert-but-delayed" misses four things that aren't timing:

  1. Same handle, two ciphertexts. RFC-010 leaves the first competing block after a common parent able to derive the same handle on both forks; with block-context rerand the bytes differ. A one-ciphertext-per-handle schema (= the revert model) can't represent that at all. This edge case is why RFC-011 exists.
  2. N coprocessors, N nodes, different forks at the same time. Producer-block keying is what makes cross-coprocessor divergence attributable — without it, fork churn is indistinguishable from crypto byte-drift.
  3. Post Gateway retirement, the settled region is the convergence mechanism. Attestation-quorum decryption has no arbiter, and handle-keyed S3 has one slot per handle — branch resolution + monotonic settlement is what decides which bytes that slot converges to and when they stop changing. In the meantime, speculative consensus and decryption are possible to avoid delays, knowing cleartext is the same even if bytes differ. Same handle wil always have the same cleartext even if ciphertexts differ.
  4. Revert means recompute, and recompute isn't byte-stable (e.g., tfhe-csprng 0.8.1→0.9.0 upgrade silently changing ct bytes). Settled bytes are never recomputed under the branch model; fork recovery preserves them (asserted in E3/C3b e2e tests).

Agreed the RPC is the trust root in both models — but we don't act on it blindly: finalization requires the row to have been observed by ingestion and its parent linkage not to contradict the finalized predecessor. Immediate revert is the model that acts destructively on a single unverified answer.

Eikix:
Last attempt poking (in an attempt to prove that we can't do anything except fork branch DAG)

---

## Architectural review: scope and decomposition

The direction is justified. `current_block_hash` rerand was evaluated against `prev_block_hash` and **refused by the security team**, making fork divergence a hard constraint. Fork-tracking infrastructure is necessary, not optional.

Two challenges remain: **scope** and one **correctness note**.

---

### Point 4 (byte-stability across library upgrades) is excluded by construction

Blue/green upgrades start green from a **finalized** height. Reorgs only affect the tip, which blue (old library) still handles:

```
v1 (blue):  ████████████████████████░░░░░░  ← handles tip (reorgs here)
v2 (green): ░░░░░░░░░░░░░░░░██████████████  ← starts at finality (no reorgs)
                              ↑ finalized
```

No block is ever "computed with v1, recomputed with v2." The `tfhe-csprng 0.8.1→0.9.0` scenario cannot occur within the upgrade protocol's guarantees. A reorg deeper than finality invalidates both models equally.

---

### RPCs are sequential — forks are never observed in parallel

Verified across QuickNode (our provider), Alchemy, Infura, go-ethereum:

> QuickNode Streams: *"A reorg is detected when a new block's parent hash does not match the hash of the previous block that was streamed."*

No provider broadcasts competing forks simultaneously. "Track multiple forks" is a **storage** claim, not an observation claim. You receive one chain, detect reorgs reactively, reconcile after the fact.

This doesn't invalidate the branch model — but it means the observation window is bounded. Forks arrive sequentially and resolve quickly. A bounded buffer at the tip captures the full practical scenario.

---

### Scope: orthogonal concerns bundled

| Concern | ~Lines | Fork-handling? |
|---|---|---|
| DFG rewrite | 4K | No (optimization) |
| S3 publication + repair queue | 3.7K | Separable |
| Fork-tracking + settlement | 2.5K | Yes (core) |
| CT key change (`producer_block_hash`) | 1.8K | Yes (irreducible) |
| Tests + .sqlx | 9.3K | — |

The DFG rewrite and S3 machinery are independently valuable and independently reviewable. Splitting reduces blast radius and makes the core fork-handling logic easier to audit in isolation.

---

### Settlement freeze: confirm it's monitored

Settlement correctly freezes when a computation is incomplete (`is_completed = FALSE AND is_error = FALSE`). This is the right safety behavior — you don't want to settle past incomplete state.

Since settlement is per-operator, one stuck worker doesn't block others. The risk is operational (infra failure, TFHE hang), not protocol-level. Confirm there's an alert on `settled_height` not advancing for > N minutes so operators can intervene before branch tables grow unboundedly.

---

### Lighter alternative for consideration

Given forks don't pile up (shallow, sequential, fast-resolving):

```
 S3 (immediate, speculative)
 ┌─────────────────────────────────────────┐
 │ key: (handle) → CT bytes                │
 │ Published at tip. Decryption correct    │
 │ regardless of fork (same cleartext).    │
 │ Updated on reorg. Immutable at finality.│
 └─────────────────────────────────────────┘
                  ▲
 Tip buffer (last N blocks)
 ┌─────────────────────────────────────────┐
 │ key: (handle, producer_block_hash) → CT │
 │ Attribution for operator consensus.     │
 │ Shallow reorg: switch pointer.          │
 │ GC'd at finality.                       │
 └─────────────────────────────────────────┘
                  ▲ promote
 Settled (finalized, immutable)
 ┌─────────────────────────────────────────┐
 │ key: (handle) → canonical CT            │
 └─────────────────────────────────────────┘
```

Same four properties. ~4-5K lines. No branch DAG, no settlement protocol (finality is height-based). Worth documenting why the full model was chosen over this, even if the answer is "we need the generality for future multi-chain scenarios."

---

### Summary

- Direction: justified (security constraint makes fork-tracking necessary)
- Point 4: excluded by blue/green upgrade protocol
- Scope: DFG + S3 should be separate PRs
- Settlement freeze: correct behavior, needs monitoring/alerting
- Alternative: bounded buffer satisfies the same properties at lower cost; document why the full model was preferred

*Read-only architectural review.*

antoniupop:
> Last attempt poking (in an attempt to prove that we can't do anything except fork branch DAG)
> 
> ## Architectural review: scope and decomposition
> [...]

Indeed, I agree with most of this review.

Point 4 / blue-green: I'd mostly agree. Within a followed blue/green protocol you're right — green anchors at a settled height, and blue owns the tip, no block is cross-version recomputed. I'd keep never-recompute-settled-bytes as defense rather than a primary justification: the csprng issue happened because that upgrade wasn't blue/green (rolling upgrade), so the protection is for the reality of deployments more than the theoretical behaviour. The core justifications (same-handle collision case, cross-coprocessor attribution) don't rest on point 4.

Bounded buffer: the bound is wrong in kind, not size. Agreed forks are shallow, sequential, and fast-resolving — but stale-branch accumulation is interruption-driven, not observation-driven: restarts mid-reorg, LB backend switches, downtime + catch-up, recovery flows. C3b showed exacly this issue where a stale branch was straddling the settlement frontier, left by a previous recovery — far below any "last N blocks" window. The buffer's true lower edge is the settled frontier, which trails finality whenever work, cleanup, or publication lags. A fixed N must either stall ingestion when full or drop rows settlement hasn't verified; the latter silently converts "unverified" into "looks drained," which is precisely the fail-open class this PR closes.

"No settlement protocol — finality is height-based": chain finality says nothing about the completeness of our derived state. Height-based promotion will promote a handle whose canonical ciphertext a lagging worker hasn't produced yet, whose orphan sibling rows still exist, or whose S3 object still holds the losing variant. The settled region is defined by four invariants (work drained, orphans cleaned, publication verified, writes fenced), and the write guard — which stops a slow worker inserting stale rows below the frontier — needs the scalar to compare against. "Immutable at finality" assumes exactly the asynchrony the settlement protocol exists to close. Every property of the lighter alternative inherits this flaw.

The lighter alternative is mostly the current design, minus its convergence bookkeeping. Speculative handle-keyed S3 publication at the tip is already the plan — the publication doc explicitly accepts early attestation consensus forming on a losing branch under the same-handle-same-cleartext guarantee. What the sketch drops is the repair queue and publication verification, i.e. the machinery that makes "updated on reorg" true when an uploader crashes mid-PUT or a stale process writes last. "Updated on reorg" is a liveness claim; the repair queue is its proof. So the line-count comparison isn't like-for-like.

Scope split: accepted in principle, one correction to the table. The DFG rewrite isn't an optional optimization: RFC-019/020 force block-scoped execution (same-block intermediates in memory, cross-block reuse only via canonical compressed state), and the scheduling work is how workers acquire block-scoped work at all. That actually supports your conclusion — it's a prerequisite that can land as its own earlier PR — but it's load-bearing, not separable-and-droppable. S3 publication can split too, with the caveat that settlement gates on publication verification, so the interim needs that gate stubbed or flagged. I can look at the decomposition cost if you think it's needed.

Monitoring: yes, that's correct - coprocessor_settlement.updated_at exists and quarantined cleanup jobs increment a counter, but there's no gauge for settled_height / frontier age today. Freezing is by design and per-coprocessor (quorum survives one frozen operator), but the alert is cheap. I'm adding a frontier-age gauge + alert as a follow-up.
**Edit: now done in last commit**


**The important thing to re-examine is the height-based-finality assumption** — if you accept that chain finality ≠ derived-state completeness, the settlement scalar and the frontier machinery follow, and the delta between your sketch and the implemented model reduces to the repair queue plus the settlement gate, both of which exist to make the sketch's own claims ("updated on reorg", "immutable at finality") actually hold.

PR #2852, opened 2026-06-19
feat(test-suite): confidential bridge e2e and LayerZero wiring in multi-chain

PanGan21:
## Summary

- Local LayerZero V2 infra per host chain: A new `bridge-deploy` step deploys an `EndpointV2Mock` + message lib, deploys the `ConfidentialBridge`, and wires the chains as LZ peers (`setPeer`/`setDstChainId`). Host-listeners get each chain's bridge address to ingest bridge events.
- New `confidential-bridge` e2e suite: Covers public decrypt, user decrypt, computed handle, multiple handles per send, reverse direction, deferred compose, and the `grantFallbackPlaintext` governance fallback.
- Cross RPC relayer: Mocks the behavior of the off-chain LZ executor (`PacketSent` -> `validatePacket` -> `lzReceive` -> `lzCompose`). A `BRIDGE_LZ_ENDPOINT_<CHAINKEY>` opt out targets a real endpoint instead of the mock.

Related: https://github.com/zama-ai/fhevm-internal/issues/1544


PanGan21:
Successful e2e test run: https://github.com/zama-ai/fhevm/actions/runs/27845834244

PR #2854, opened 2026-06-19
fix(gateway-contracts): add extraData v2 support in Decryption contract

geoxel:
Adds handling for `extraData` version 2 in `Decryption.sol`, function `_extractContextId()`


gateway-contracts/contracts/Decryption.sol
```diff
@@ -983,6 +983,18 @@ contract Decryption is
             return contextId;
         }
 
+        // Version 2 -> extract contextId from bytes 1..33; epochId in bytes 33..65 is for KMS use
```
isaacdecoded: Since the epochId is unused at this contract level, maybe it makes sense to change the conditional above to:

`if (version >= 1) {
...
}`

Wdyt? @geoxel

geoxel: Or maybe `if (version == 1 || version == 2) { ... }` so that v3 or v42 go to the UnsupportedExtraDataVersion line 999 ? 
I change the test on length for version 1 from `if (extraData.length < 33)` to `if (extraData.length != 33) {`. It is ok ?

gateway-contracts/contracts/Decryption.sol
```diff
@@ -969,11 +969,13 @@ contract Decryption is
             return GATEWAY_CONFIG.getCurrentKmsContextId();
         }
 
-        // Version 1 -> extract contextId from bytes 1..33
-        if (version == 1) {
-            if (extraData.length < 33) {
-                revert InvalidExtraDataLength(extraData.length, 33);
+        // Versions 1/2 -> extract contextId from bytes 1..33
+        if (version == 1 || version == 2) {
+            uint256 expectedLength = version == 1 ? 33 : 65;
```
isaacdecoded: I don't think we care that much if the extraData is 65-length for version 2, but just if it's 33-length so it contains the contextId. I mean, this `_extractContextId` is about getting the contextId from a given set of bytes, so we only need the 33 bytes to get it. To me, it seems enough:

```
if (version == 1 || version == 2) {
      if (extraData.length < 33) { ... }
}
```

Makes sense to you? @geoxel

geoxel: Fine!

isaacdecoded (approved):
LGTM! 👍

geoxel:
@mergifyio queue

PR #2855, opened 2026-06-19
chore(listener): enable iam-auth feature in listener_core images

maxnovawind:
## Problem

`listener_core` supports RDS IAM database authentication (`store/iam_auth.rs`), but the code is gated behind the `iam-auth` Cargo feature, which is **off by default** (`default = []` in `crates/listener_core/Cargo.toml`).

Both Docker images built without the feature:

```dockerfile
cargo build --release -p listener_core
```

So the shipped binary has no IAM code compiled in. When `iam_auth.enabled=true` is set in config, the binary fails fast at startup:

```
Validation error: iam_auth.enabled=true but binary was built without the iam-auth feature
```

This rejection comes from `DatabaseConfig::validate()`, which is `#[cfg(not(feature = "iam-auth"))]` — its presence proves the binary was built without the feature.

## Fix

Compile `listener_core` with `--features iam-auth` in both Dockerfiles:
- `listener/Dockerfile`
- `listener/Dockerfile.ci`

This pulls in the `aws-config` / `aws-sigv4` / `aws-credential-types` deps and the `connect_iam` path, so `iam_auth.enabled=true` works when configured.

## Notes for reviewers

- **Runtime prerequisites unchanged**: enabling IAM still requires `iam_auth.enabled: true` in config and resolvable AWS credentials (IRSA). This PR only makes IAM *possible* in the binary.
- **Cargo-chef caching** (`Dockerfile`): the `cargo chef cook` step still cooks without the feature, so the AWS crates compile in the final build step rather than the cached dep layer. Correct, just slightly less cache-efficient — can be optimized separately if desired.

## Test plan

- [ ] CI image build succeeds with the feature enabled
- [ ] Deploy with `iam_auth.enabled: true` + IRSA → no validation error, connects via IAM token

### DIFF
--- a/coprocessor/fhevm-engine/db-migration/Dockerfile
+++ b/coprocessor/fhevm-engine/db-migration/Dockerfile
@@ -10,6 +10,30 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked \
     cargo install sqlx-cli --version 0.7.2 \
     --no-default-features --features postgres,rustls --locked
 
+# Install AWS CLI v2 (self-contained bundle) for RDS IAM auth token generation.
+# prepare_database_url.sh shells out to `aws rds generate-db-auth-token`. The bundle
+# ships its own Python/libs and only depends on glibc, so it runs on the glibc-based
+# postgres runtime image below.
+ARG TARGETARCH
+RUN set -eux; \
+    apk add --no-cache unzip; \
+    arch="${TARGETARCH:-}"; \
+    if [ -z "$arch" ]; then \
+      case "$(uname -m)" in \
+        x86_64) arch=amd64 ;; \
+        aarch64) arch=arm64 ;; \
+      esac; \
+    fi; \
+    case "$arch" in \
+      amd64) aws_arch=x86_64 ;; \
+      arm64) aws_arch=aarch64 ;; \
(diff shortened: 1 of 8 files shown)


nboisde (approved):
Thanks for this catch

coprocessor/fhevm-engine/db-migration/Dockerfile
```diff
@@ -10,6 +10,30 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,sharing=locked \
     cargo install sqlx-cli --version 0.7.2 \
     --no-default-features --features postgres,rustls --locked
 
+# Install AWS CLI v2 (self-contained bundle) for RDS IAM auth token generation.
```
rudy-6-4, line 13: Could we use a cgr image to copy from it ? https://images.chainguard.dev/directory/image/aws-cli/overview

maxnovawind: After looking into it, I wouldn’t recommend this approach because we would need to copy `aws-cli` and all its dependencies, including Python. It’s not something we can simply install by copying and running the installer in one line.

rudy-6-4 (approved):

maxnovawind:
@mergify queue

tawadaa (approved):

tawadaa:
@mergify dequeue

PR #2857, opened 2026-06-19
refactor(coprocessor): decouple upgrade from context id

PanGan21:
## Summary

- Replaces the auto-incremented, on-chain-tracked coprocessor context with a caller-supplied `proposalId` emitted via a single event. 
- The contract becomes a pure announcer for upgrade proposals — the dry-run, consensus and cutover lifecycle is driven entirely off-chain by the host-listener / upgrade-controller / consensus-detector pipeline.

Related: https://github.com/zama-ai/fhevm-internal/issues/1552


host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -163,25 +159,24 @@ interface IProtocolConfig {
...
+     * @param proposalId Caller-supplied identifier for this upgrade attempt. Must be non-zero.
+     *        Uniqueness across calls is the caller's responsibility; the contract does not enforce it.
      * @param softwareVersion The coprocessor software version.
      * @param chainUpgradeWindows The per-host-chain replay windows.
      * @param gwStartBlock The Gateway block to resume from.
      * @param ciphertextVersion The ciphertext version the new software writes.
      */
-    function defineNewCoprocessorContext(
+    function proposeCoprocessorUpgrade(
+        uint256 proposalId,
```
isaacdecoded, line 173: Is this an issue if the caller supply a duplicated/already used proposal Id? Or it doesn't affect the upgrade process at all?

isaacdecoded (approved):
LGTM!

host-contracts/contracts/interfaces/IProtocolConfig.sol
PanGan21: Since an upgrade can fail, and it will be retryable, it should be okay to pass the same proposal id. @goshawk-3 wdyt?

goshawk-3 (approved):
LGTM

PR #2868, opened 2026-06-22
refactor(common): improve code quality based on OZ reported issue #20

isaacdecoded:
From the [N-04 report](https://github.com/zama-ai/fhevm-internal/issues/1492) this PR targets, it was only implemented the `getRequiredCountEnvVar` hardhat utils method and the `delete` operation in the GatewayConfig contract. The other suggested changes are irrelevant at this point.

Closes https://github.com/zama-ai/fhevm-internal/issues/1492


gateway-contracts/contracts/interfaces/IGatewayConfig.sol
```diff
@@ -729,6 +730,13 @@ interface IGatewayConfig {
      */
     function getUserDecryptionThresholdForContext(uint256 contextId) external view returns (uint256);
 
+    /**
+     * @notice Returns the maximum allowed KMS signer-set size, which is also the upper bound enforced
+     *         on every per-context threshold at registration time.
+     * @return The maximum number of KMS signers.
+     */
+    function getMaxKmsSigners() external pure returns (uint256);
```
obatirou: do you think this is really necessary? Who will use it and why?
My understanding is that related to protocol internal limitation regarding the proof format limitation + MPC limitation.
Also we are changing code in gateway which should be deprecated at some point. 
Unless critical we should avoid any changes, no?
What are your thoughts on this?

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -351,8 +351,8 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         // Remove the old coprocessors
         uint256 oldCoprocessorTxSenderAddressesLength = $.coprocessorTxSenderAddresses.length;
         for (uint256 i = 0; i < oldCoprocessorTxSenderAddressesLength; i++) {
-            $.isCoprocessorTxSender[$.coprocessorTxSenderAddresses[i]] = false;
-            $.isCoprocessorSigner[$.coprocessorSignerAddresses[i]] = false;
+            delete $.isCoprocessorTxSender[$.coprocessorTxSenderAddresses[i]];
```
obatirou, line 354: Reinit is not updated.

gateway-contracts/tasks/utils/loadVariables.ts
```diff
@@ -18,6 +18,17 @@ export function getRequiredEnvVar(name: string): string {
   return value;
 }
 
+// Get a required positive-integer environment variable. Throws if it is missing, empty, non-integer, or
+// not strictly positive, so a malformed count fails loudly instead of yielding NaN and silently no-op'ing.
+export function getRequiredCountEnvVar(name: string): number {
```
obatirou, line 23: do we need to use it in all folders?
host, library-solidity

Update: it is in host but not library

gateway-contracts/contracts/interfaces/IGatewayConfig.sol
isaacdecoded: I had the same thoughts when working on these changes, but wanted to put such an agreement "on paper" - I mean, I agree that this doesn't seem a real requirement to have this getter and better to avoid adding new code in components that are being deprecated like this one.

gateway-contracts/contracts/GatewayConfig.sol
isaacdecoded: This is because it's [already bumped on main](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/GatewayConfig.sol#L55) compared with our latest release [v0.14.1](https://github.com/zama-ai/fhevm/blob/v0.14.1/gateway-contracts/contracts/GatewayConfig.sol#L55)

gateway-contracts/tasks/utils/loadVariables.ts
isaacdecoded: I think only for gateway and host contracts, since they use env vars as counts during deployment, like `NUM_COPROCESSORS`, `NUM_KMS_NODES`, etc.

gateway-contracts/contracts/interfaces/IGatewayConfig.sol
isaacdecoded: Removed the YAGNI getMaxKmsSigners impl [here](https://github.com/zama-ai/fhevm/pull/2868/changes/dfce5b97adccfdacafe41eedfafbd4cbb6727b6a)

gateway-contracts/contracts/GatewayConfig.sol
obatirou: oh right

gateway-contracts/tasks/utils/loadVariables.ts
isaacdecoded: Checking twice, you're right that it must be implemented in the library as well. Added [here](https://github.com/zama-ai/fhevm/pull/2868/changes/21d1fe1f61d0c9bddb9a377c3ced88f97a894ead).

obatirou (approved):

nboisde (approved):
Already reviewed, passing codeowners.

isaacdecoded:
@mergifyio queue

jatZama (approved):

PR #2871, opened 2026-06-22
feat(library-solidity): add confidential bridge send and receive wrappers

poppyseedDev:
Generate FHE.bridge / FHE.quoteBridge overloads (single handle and list, for every encrypted type, plus a low-level bytes32 escape hatch) into FHE.sol and Impl.sol via codegen. The ConfidentialBridge is resolved from the ACL (getConfidentialBridgeAddress), so apps need no extra configuration.

Add the ConfidentialBridgeReceiver base contract (trusted-bridge + overridable peer authentication, raw bytes32 handle dispatch) and the LayerZero-free IConfidentialBridge types/interfaces under lib/bridge/. Include a ConfidentialOFTViaLib example, test helpers, and mock-based tests.


library-solidity/examples/bridge/ConfidentialOFTViaLib.sol
```diff
@@ -0,0 +1,133 @@
...
+        address dstApp = _dstPeers[dstEid];
+        if (dstApp == address(0)) revert PeerNotSet(dstEid);
+
+        euint64 actualAmount = _burn(msg.sender, amount);
+
+        // Payload references the same handle at index 0 (so the receiver mints it).
+        bytes memory payload = abi.encode(recipient, euint64.unwrap(actualAmount));
+
+        emit Bridged(msg.sender, dstEid, recipient);
+        return FHE.bridge(dstEid, dstApp, payload, actualAmount, mintComposeGas, msg.value);
```
PanGan21: Shouldn't we validate mintComposeGas to be > 0 here? If it is 0 then the destination call back won't run.

library-solidity/examples/bridge/ConfidentialOFTViaLib.sol
```diff
@@ -0,0 +1,133 @@
...
+pragma solidity ^0.8.24;
+
+import "@openzeppelin/contracts/access/Ownable2Step.sol";
+
+import {FHE, euint64, externalEuint64, ebool} from "../../lib/FHE.sol";
+import {ZamaConfig} from "../../config/ZamaConfig.sol";
+import {ConfidentialBridgeReceiver} from "../../lib/bridge/ConfidentialBridgeReceiver.sol";
+import {MessagingReceipt} from "../../lib/bridge/IConfidentialBridge.sol";
+
+/**
```
PanGan21, line 12: It seems this is hand written file, so we should either make it autogenerated or move it in another folder

library-solidity/codegen/src/templateImpDotSol.ts
```diff
@@ -11,12 +11,14 @@ import { removeTemplateComments } from './utils';
...
  */
-export function generateSolidityImplLib(operators: Operator[], fheTypeDotSol: string): string {
+export function generateSolidityImplLib(operators: Operator[], fheTypeDotSol: string, bridge: boolean): string {
   // Placeholders:
   // =============
   // $${FheTypeDotSol}$$
   // $${CoprocessorInterfaceOperators}$$
   // $${ImplOperators}$$
+  // $${ImplBridgeImports}$$
+  // $${ImplBridge}$$
```
PanGan21, line 22: I think you should also add `$${ImplBridgeACLGetter}$$` here since it is added below

library-solidity/codegen/src/templateImpDotSol.ts
```diff
@@ -25,9 +28,73 @@ export function generateSolidityImplLib(operators: Operator[], fheTypeDotSol: st
...
+    }
+
+    /**
+     * @notice Forwards a bridge send to the host \`ConfidentialBridge\`, paying \`nativeFee\`.
+     * @dev    Runs in the caller's context, so the bridge sees \`msg.sender == <app>\` and the
+     *         source ACL check resolves against the app.
+     */
+    function bridge(uint32 dstEid, bytes32 dstApp, bytes memory payload, bytes32[] memory handleList, uint128 lzComposeGas, bytes memory options, uint256 nativeFee) internal returns (MessagingReceipt memory receipt) {
+        receipt = getConfidentialBridge().send{value: nativeFee}(dstEid, dstApp, payload, handleList, lzComposeGas, options);
+    }
```
PanGan21: It seems that it is missing the zero compose gas guard here otherwise `FHE.bridge(..., lsComposeGas: 0)` silently succeeds.
However as discussed today the options will be removed so maybe we should wait for that to be merged first.

library-solidity/test/bridge/Bridge.ts
```diff
@@ -0,0 +1,415 @@
...
+        expect(sent.lzComposeGas).to.equal(gas);
+        expect(sent.options).to.equal('0x'); // single-handle overloads derive defaults from gas
+        expect(sent.value).to.equal(fee);
+        expect(sent.caller).to.equal(await harness.getAddress());
+      });
+    }
+
+    it('forwards a zero native fee unchanged', async function () {
+      const { bridge, harness } = await deploySendFixture();
+      await harness.bridge64(DST_EID, randAddr(), '0x', handleA, 0, { value: 0 });
```
PanGan21, line 75: Mentioned in previous comment but let's see if we still want to allow zero gas  since it can become vector.

library-solidity/codegen/src/templates/ConfidentialOAppSender.sol-template
```diff
@@ -0,0 +1,57 @@
...
+abstract contract ConfidentialOAppSender is ConfidentialOAppCore {
+    /// @notice `lzComposeGas` is 0; with the default (empty) options the destination receive
+    ///         callback would never run, so the bridged value would never be delivered.
+    error ZeroComposeGas();
+
+    /**
+     * @notice Bridge a single encrypted `euint64` handle to the peer on `dstEid`.
+     * @param dstEid        Destination LayerZero endpoint id (must have a configured peer).
+     * @param payload       Opaque app payload (the receiver decodes it).
+     * @param handle        The encrypted value to bridge; the caller must hold ACL allowance on it.
```
PanGan21: I am not sure if this is accurate. I think what is happening is that the `ConfidentialBridge` checks `isAllowed(handle, msg.sender)` where msg.sender is the OApp contract which holds `allowThis` on the handle. 
Maybe we should even mention that subclasses must authorize the external caller like `ConfidentialOFTViaLib.send` is doing.

library-solidity/codegen/src/templateFHEDotSol.ts
PanGan21, line 66: Nit: Don't forget to add the `$${FHEBridge}$$` in this list!

library-solidity/test/bridge/Bridge.ts
poppyseedDev: I think this was resolved

PR #2875, opened 2026-06-22
feat(sdk): add protocol version auto-resolution and epochId tracking in KMS extra-data (RFC-005)

geoxel:
First two commits have been written by @0xalexbel 

### feat(sdk): add protocol version auto resolver (`8ad99451`)
Introduces `ProtocolVersionResolver`, a module that detects and resolves the FHEVM protocol version at runtime by querying on-chain host contract data. Adds semver parsing utilities, a `createKmsUserDecryptEip712V2` path for the v2 EIP-712 flow, and expands `HyperWasmSolver` for multi-version dispatch. Includes `DRAFT_RFC_016.md`, full test coverage, and updated compatibility docs.

### fix(sdk): add ensureResolvedProtocolVersion, remove circular dep (`442319cf`)
Extracts version-resolution logic into a dedicated `resolveFhevmVersions-p.ts` module to break a circular dependency introduced by the previous commit. Adds `ensureResolvedProtocolVersion` as a guard invoked by client decorators before any encrypt/decrypt operation.

### feat(sdk): add ProtocolConfig contract infos (`cb8a7228`)
Adds the `ProtocolConfig` contract address to all chain definitions (`localTestnet`, `mainnet`, `sepolia`, `devnet`, localstack variants) and to the `FhevmChain` / `HostContract` types. Wires up `resolveFhevmConfig` to read the address, adds the relevant ABI fragments, and updates deploy scripts.

### fix(sdk): fix for assertRecordUintBigIntProperty (`de5bfda0`)
One-line fix in `uint.ts`: corrects a type guard in `assertRecordUintBigIntProperty` that was incorrectly narrowing the checked value.

### chore(sdk): rename getVersion into getHostContractVersion (`0d6937f5`)
Renames `getVersion` on host contract readers to the more descriptive `getHostContractVersion` across all call-sites — host action readers, `HostContractVersion-p`, KMS context helpers, `ProtocolVersionResolver`, and related tests. No behaviour change.

### chore(sdk): use specific contract name instead of just 'address' (`cfc3a616`)
Replaces generic `address` field/variable names with contract-specific names (e.g. `aclAddress`, `fhevmExecutorAddress`) throughout host-contract helpers, ACL checks, delegation, and KMS share fetching. No behaviour change — purely a readability improvement.

### feat(sdk): add epochId support to extraData (RFC-005) (`fe2915e5`)
Implements `epochId` tracking in the KMS extra-data layer per RFC-005. Adds `getCurrentKmsContextAndEpoch-p` to fetch both the current KMS context ID and its epoch in one call, updates `kmsExtraData` encode/decode for the epoch field, and propagates the value through `KmsSignersContext`, `SignedDecryptionPermit`, and the signcrypted-shares fetch path. Handles both legacy (no epoch) and epoch-aware protocol versions.

### DIFF
--- a/.github/workflows/js-sdk-tests.yml
+++ b/.github/workflows/js-sdk-tests.yml
@@ -6,15 +6,15 @@ on:
       - devex/js-sdk
       - main
     paths:
-      - '.github/workflows/js-sdk-*.yml'
-      - 'sdk/js-sdk/**'
+      - ".github/workflows/js-sdk-*.yml"
+      - "sdk/js-sdk/**"
   push:
     branches:
       - devex/js-sdk
   workflow_dispatch:
     inputs:
       release:
-        description: 'Set to true for release tagging'
+        description: "Set to true for release tagging"
         required: false
         default: false
 
@@ -29,8 +29,8 @@ jobs:
       run:
         working-directory: ./sdk/js-sdk
     permissions:
-      contents: 'read' # Required to checkout repository code
-      id-token: 'write' # Required for OIDC / npm provenance attestation
+      contents: "read" # Required to checkout repository code
+      id-token: "write" # Required for OIDC / npm provenance attestation
     steps:
       # -----------------------------------------------------------------------
       # 1. Checkout
@@ -41,7 +41,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
-          persist-credentials: 'false'
+          persist-credentials: "false"
 
       # -----------------------------------------------------------------------
       # 2. Node.js setup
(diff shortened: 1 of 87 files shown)


.github/workflows/js-sdk-tests.yml
```diff
@@ -6,15 +6,15 @@ on:
       - devex/js-sdk
       - main
     paths:
-      - '.github/workflows/js-sdk-*.yml'
-      - 'sdk/js-sdk/**'
+      - ".github/workflows/js-sdk-*.yml"
```
geoxel, line 9: 'singleQuote' rule has changed on commit 7cbde4f47a7 of May 29.

sdk/js-sdk/src/core/actions/host/resolveFhevmConfig.ts
```diff
@@ -52,6 +53,7 @@ export type ResolveFhevmConfigReturnType = {
   readonly fhevmExecutor: FhevmExecutorContractData;
   readonly inputVerifier: InputVerifierContractData;
   readonly kmsVerifier: KmsVerifierContractData;
+  readonly protocolConfig: ChecksummedAddress | undefined;
```
0xalexbel, line 56: Maybe follow the existing mechanism.
Define a ProtocolConfigContractData class that fetches view functions like version etc.

0xalexbel (approved):

PR #2880, opened 2026-06-23
feat(host-contracts): fix Bridge lzReceive gas setters

jatZama:


PanGan21 (approved):

host-contracts/contracts/bridge/HandlesSender.sol
PanGan21: This is stale comment probably needs to be updated, since the contract now always builds options internally

PanGan21: Same as the previous one I believe

host-contracts/contracts/bridge/HandlesSender.sol
```diff
@@ -120,17 +114,14 @@ abstract contract HandlesSender is OAppSenderUpgradeable, ACLOwnable, BridgeEven
...
      * @param lzComposeGas   Gas budget for the destination-side `lzCompose` (which runs
-     *                       the destination app's `onConfidentialBridgeReceived`). Must be 0 if `options`
-     *                       is non-empty.
-     * @param options        Raw LayerZero options; if empty the contract builds default
-     *                       options from `baseGas(dstEid) + handleList.length *
-     *                       perHandleGas(dstEid)` and `lzComposeGas`, where the base and
-     *                       per-handle gas are the `dstEid`'s custom overrides if set, else
-     *                       the {LZ_RECEIVE_BASE_GAS_DEFAULT}/{LZ_RECEIVE_PER_HANDLE_GAS_DEFAULT}
-     *                       constants.
+     *                       the destination app's `onConfidentialBridgeReceived`).
```
PanGan21, line 143: Consider mentioning here specifically that onConfidentialBridgeReceived only runs if lzComposeGas > 0

PanGan21 (commented):
Looks great only some docs suggestions

host-contracts/contracts/bridge/HandlesSender.sol
jatZama: I will revert if `lzComposeGas` is below some minimum value instead, in my next commit.

jatZama: Done in commit 1747883022b5c43f269d85d8d34ad38a214aa31b

host-contracts/test/bridge/Bridge.t.sol
```diff
@@ -222,35 +270,87 @@ contract BridgeTest is TestHelperOz5, HostContractsDeployerTestUtils, BridgeEven
...
+    }
+
+    function test_Send_RevertsOnEmptyHandleList() public {
+        bytes32[] memory handleList = new bytes32[](0);
+        vm.prank(srcApp);
+        vm.expectRevert(HandlesSender.EmptyHandleList.selector);
+        srcBridge.send{value: 0}(DST_EID, _addressToBytes32(address(dstApp)), "", handleList, uint64(0));
+    }
+
+    function test_Send_RevertsOnLzComposeGasBelowMin() public {
```
PanGan21, line 315: Let's add a test for `test_Send_RevertsOnLzComposeGasBelowCustomMin` as well for better coverage

host-contracts/test/bridge/Bridge.t.sol
```diff
@@ -208,12 +214,54 @@ contract BridgeTest is TestHelperOz5, HostContractsDeployerTestUtils, BridgeEven
         assertEq(srcBridge.getLzReceivePerHandleGas(DST_EID), srcBridge.LZ_RECEIVE_PER_HANDLE_GAS_DEFAULT());
     }
 
+    function test_SetLzReceivePerPayloadByteGas_OverridesEmitsAndClears() public {
```
PanGan21, line 228: Let's also add a test or an assert inside this test that payload length actually effects the gas

host-contracts/test/bridge/Bridge.t.sol
jatZama: Added in last commit 4169e3da901ced85c4e187d87cca1792a596e166

host-contracts/test/bridge/Bridge.t.sol
jatZama: This wont be visible via the `quote` method in local testing fyi because the foundry test-devtools tests actually use SimpleMessageLib  via `TestHelperOz5` which replaces the SendUln302 library with a no-op stub:
It doesn't check DVN signatures and uses a flat fee—it returns a hardcoded `MessagingFee(nativeFee, lzTokenFee)` where `nativeFee = 100` and `lzTokenFee = 99`, regardless of message size, gas in the options, or payload length. 

Anyways, I pushed in my last commit 4169e3da901ced85c4e187d87cca1792a596e166 the  new tests you wanted, for the impact on gas of payload length I am just decoding the raw options bytes array instead of calling `quote`, for the reason explained above.

PR #2885, opened 2026-06-23
chore(kms-connector): add metric for request checks

eudelins-zama:
## Summary

Add a `kms_connector_worker_request_check_errors` metric to count request pre-flight check failures in the `kms-worker`.

Closes https://github.com/zama-ai/fhevm-internal/issues/1341

### Error handling refactor

- Introduces a dedicated `event_processor/error.rs` module with:
  - `ProcessingError` (`Recoverable` / `Irrecoverable`), moved out of `processor.rs`.
  - `RequestCheckError`, a `ProcessingError` tagged with the `RequestCheckKind` that produced it. The metric increment is centralized in `RequestCheckError::record`, called at each conversion boundary, so each check failure is counted exactly once under the right family.

### Request ID tracing alignment

- Event `Display` impls now render request IDs as zero-padded big-endian hex (`0x...`, 32 bytes) via a new `u256_to_be_hex` helper, matching the request ID format used by kms-core for easier cross-log correlation.
- Closes https://github.com/zama-ai/fhevm-internal/issues/571

### DIFF
--- a/docs/metrics/metrics.md
+++ b/docs/metrics/metrics.md
@@ -280,6 +280,19 @@ Metrics for zkproof-worker are to be added in future releases, if/when needed. C
  - **Alarm**: If the counter increases over a period of time.
    - **Recommendation**: more than 60 failures in 1 minute, i.e. `sum(increase(counter[1m])) > 60`.
 
+#### Metric Name: `kms_connector_worker_request_check_errors`
+ - **Type**: Counter
+ - **Labels**:
+   - `check_type`: the family of pre-flight check that rejected the request:
+     - `acl`: ACL authorization checks and errors that prevent the ACL check from running (malformed handles, missing contract config...).
+     - `signature`: RFC-012/016 signature & request-validity checks (EIP-712/ERC-1271 signature, validity window, signature invalidation).
+     - `copro_consensus`: RFC-023 off-chain ciphertext-attestation consensus check.
+     - `kms_context`: KMS context/epoch validity check.
+     - `network`: network error (on-chain call or DB query) encountered while running any of the above checks.
+ - **Description**: Counts request pre-flight check failures. Mostly decryption checks (ACL, signature, copro consensus), but the `kms_context` family also covers key-management requests.
+ - **Alarm**: If the counter increases over a period of time.
+   - **Recommendation**: more than 60 failures in 1 minute, i.e. `sum(increase(counter[1m])) > 60`.
+
 #### Metric Name: `kms_connector_worker_decryption_latency_seconds`
  - **Type**: Histogram
  - **Labels**:
--- a/kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
@@ -4,7 +4,7 @@ use super::{
 };
 use crate::core::{
     config::{Config, CtAttestationConfig},
-    event_processor::ciphertext::COPROCESSOR_CONTEXT_ID,
+    event_processor::{RequestCheckKind, ciphertext::COPROCESSOR_CONTEXT_ID},
 };
 use alloy::{hex, primitives::B256, providers::Provider, transports::http::Client};
 use anyhow::{Context, anyhow};
@@ -28,7 +28,7 @@ pub struct CiphertextManager<P: Provider> {
     /// HTTP client for the attestation HEAD fan-out and the ciphertext retrieval.
     client: Client,
 
-    /// Off-chain ciphertext-attestation config.
+    /// Off-chain ciphertext-attestation verification config.
     config: CtAttestationConfig,
 
     /// Number of retries for S3 ciphertext retrieval.
@@ -69,6 +69,7 @@ where
         sns_materials: &[SnsCiphertextMaterial],
     ) -> anyhow::Result<Vec<TypedCiphertext>> {
         if let Err(e) = self.verify_attestations(sns_materials).await {
+            RequestCheckKind::CoproConsensus.inc_metric();
             warn!("{e:#}");
         }
 
@@ -86,10 +87,7 @@ where
     /// Fans HEAD requests out to every registered Coprocessor bucket and computes an off-chain
     /// consensus verdict for each handle. Stops at the first failing material: a single failed
     /// verification invalidates the whole decryption request.
-    pub async fn verify_attestations(
-        &self,
-        materials: &[SnsCiphertextMaterial],
-    ) -> anyhow::Result<()> {
+    async fn verify_attestations(&self, materials: &[SnsCiphertextMaterial]) -> anyhow::Result<()> {
         if !self.config.enabled {
             return Ok(());
         }
--- a/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
@@ -1,6 +1,9 @@
 use crate::core::{
     config::Config,
-    event_processor::{CiphertextManager, ProcessingError, context::ContextManager},
+    event_processor::{
+        CiphertextManager, ProcessingError, RequestCheckError, RequestCheckKind,
+        context::ContextManager,
+    },
 };
 use alloy::{
     consensus::Transaction,
@@ -88,7 +91,7 @@ where
     pub async fn check_ciphertexts_allowed_for_public_decryption(
         &self,
         sns_ciphertexts: &[SnsCiphertextMaterial],
-    ) -> Result<(), ProcessingError> {
+    ) -> Result<(), RequestCheckError> {
         info!(
             "Starting ACL check for {} handles...",
             sns_ciphertexts.len()
@@ -96,23 +99,24 @@ where
 
         for ct in sns_ciphertexts {
             let ct_chain_id = extract_chain_id_from_handle(ct.ctHandle.as_slice())
-                .map_err(ProcessingError::Irrecoverable)?;
-            let Some(acl_contract) = self.acl_contracts.get(&ct_chain_id) else {
-                return Err(ProcessingError::Recoverable(anyhow!(
-                    "No ACL contract config found for chain id {ct_chain_id}"
-                )));
-            };
+                .map_err(|e| RequestCheckError::irrecoverable(RequestCheckKind::Acl, e))?;
+            let acl_contract = self.acl_contracts.get(&ct_chain_id).ok_or_else(|| {
+                RequestCheckError::recoverable(
+                    RequestCheckKind::Acl,
+                    anyhow!("No ACL contract config found for chain id {ct_chain_id}"),
+                )
+            })?;
 
             if !acl_contract
                 .isAllowedForDecryption(ct.ctHandle)
                 .call()
                 .await
-                .map_err(|e| ProcessingError::Recoverable(anyhow::Error::from(e)))?
+                .map_err(RequestCheckError::network)?
             {
-                return Err(ProcessingError::Recoverable(anyhow!(
-                    "{} is not allowed for decrypt!",
-                    hex::encode(ct.ctHandle)
-                )));
+                return Err(RequestCheckError::recoverable(
+                    RequestCheckKind::Acl,
+                    anyhow!("{} is not allowed for decrypt!", hex::encode(ct.ctHandle)),
+                ));
             }
         }
 
@@ -126,7 +130,7 @@ where
         calldata: Vec<u8>,
         sns_ciphertexts: &[SnsCiphertextMaterial],
         user_address: Address,
-    ) -> Result<(), ProcessingError> {
+    ) -> Result<(), RequestCheckError> {
         info!(
             "Starting ACL check for {} handles...",
             sns_ciphertexts.len()
@@ -143,10 +147,13 @@ where
                         calldata.as_slice(),
                     )
                     .map_err(|e2| {
-                        ProcessingError::Irrecoverable(anyhow!(
-                            "Was not able to parse calldata for both userDecryptionRequestCall {e2} \
-                            and delegatedUserDecryptionRequestCall ({e})!"
-                        ))
+                        RequestCheckError::irrecoverable(
+                            RequestCheckKind::Acl,
+                            anyhow!(
+                                "Was not able to parse calldata for both userDecryptionRequestCall \
+                                {e2} and delegatedUserDecryptionRequestCall ({e})!"
+                            ),
+                        )
                     })?;
                     (parsed_calldata.ctHandleContractPairs, None)
                 }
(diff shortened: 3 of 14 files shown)


maksymsur:
@claude /pr-review

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -86,10 +87,7 @@ where
     /// Fans HEAD requests out to every registered Coprocessor bucket and computes an off-chain
     /// consensus verdict for each handle. Stops at the first failing material: a single failed
     /// verification invalidates the whole decryption request.
-    pub async fn verify_attestations(
-        &self,
-        materials: &[SnsCiphertextMaterial],
-    ) -> anyhow::Result<()> {
+    async fn verify_attestations(&self, materials: &[SnsCiphertextMaterial]) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
eudelins-zama: Not relevant

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Error returned by the request pre-flight checks (ACL, KMS context, ...).
+///
+/// It is just a [`ProcessingError`] tagged with the check family that produced it. The metric
+/// increment is centralized in [`RequestCheckError::record`], called at each conversion boundary.
+#[derive(Debug, Error)]
+#[error("{cause}")]
+pub struct RequestCheckError {
+    kind: RequestCheckKind,
+    cause: ProcessingError,
+}
```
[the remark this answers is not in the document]
eudelins-zama: Not needed for now, and I don't think it will make debugging harder tbh

docs/metrics/metrics.md
```diff
@@ -280,6 +280,19 @@ Metrics for zkproof-worker are to be added in future releases, if/when needed. C
...
+ - **Labels**:
+   - `check_type`: the family of pre-flight check that rejected the request:
+     - `acl`: ACL authorization checks and errors that prevent the ACL check from running (malformed handles, missing contract config...).
+     - `signature`: RFC-012/016 signature & request-validity checks (EIP-712/ERC-1271 signature, validity window, signature invalidation).
+     - `copro_consensus`: RFC-023 off-chain ciphertext-attestation consensus check.
+     - `kms_context`: KMS context/epoch validity check.
+     - `network`: network error (on-chain call or DB query) encountered while running any of the above checks.
+ - **Description**: Counts request pre-flight check failures. Mostly decryption checks (ACL, signature, copro consensus), but the `kms_context` family also covers key-management requests.
+ - **Alarm**: If the counter increases over a period of time.
+   - **Recommendation**: more than 60 failures in 1 minute, i.e. `sum(increase(counter[1m])) > 60`.
```
dvdplm, line 294: Hmm. Would this mean that if we have less than 60 requests/min and all of them fail, we would *not* trigger an alarm?

We should use some sort of proportion, e.g 5% of requests over the last minute?

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
```diff
@@ -88,31 +91,32 @@ where
...
-                .map_err(|e| ProcessingError::Recoverable(anyhow::Error::from(e)))?
+                .map_err(RequestCheckError::network)?
             {
-                return Err(ProcessingError::Recoverable(anyhow!(
-                    "{} is not allowed for decrypt!",
-                    hex::encode(ct.ctHandle)
-                )));
+                return Err(RequestCheckError::recoverable(
+                    RequestCheckKind::Acl,
+                    anyhow!("{} is not allowed for decrypt!", hex::encode(ct.ctHandle)),
```
dvdplm: Should it say "allowed TO decrypt"? Like, isn't the handle the subject of the phrase?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn from_response_status(value: tonic::Status) -> Self {
+        let anyhow_error = anyhow!("KMS GRPC error: {value}");
+        match value.code() {
+            Code::DeadlineExceeded | Code::Unavailable | Code::ResourceExhausted => {
+                Self::Recoverable(anyhow_error)
+            }
+            _ => Self::Irrecoverable(anyhow_error),
+        }
+    }
+}
```
dvdplm, line 26: Would a standard `From` impl work here? `impl From<tonic::Status> for ProcessingError`?

kms-connector/crates/kms-worker/src/monitoring/metrics.rs
```diff
@@ -76,6 +76,15 @@ pub static S3_CIPHERTEXT_RETRIEVAL_ERRORS: LazyLock<IntCounter> = LazyLock::new(
     .unwrap()
 });
 
+pub static REQUEST_CHECK_ERRORS: LazyLock<IntCounterVec> = LazyLock::new(|| {
+    register_int_counter_vec!(
+        "kms_connector_worker_request_check_errors",
+        "Counts request pre-flight check failures, by check family",
+        &["check_type"]
+    )
+    .unwrap()
```
dvdplm: I know this is a pattern in this file, but let's start using `expect` here instead of `unwrap` so we know where to look the day the prometheus config is wrong and the server refuses to start. Just `.expect("Cannot initialize REQUEST_CHECK_ERRORS")` should do it.

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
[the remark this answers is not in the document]
dvdplm: Well, `dyn Error::source()` is the modern way of doing this and `cause()` is deprecated, so perhaps the bot is not entirely wrong? https://doc.rust-lang.org/std/error/trait.Error.html#method.cause

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Error returned by the request pre-flight checks (ACL, KMS context, ...).
+///
+/// It is just a [`ProcessingError`] tagged with the check family that produced it. The metric
+/// increment is centralized in [`RequestCheckError::record`], called at each conversion boundary.
+#[derive(Debug, Error)]
+#[error("{cause}")]
+pub struct RequestCheckError {
+    kind: RequestCheckKind,
+    cause: ProcessingError,
```
dvdplm: Can't this be `dyn std::error::Error` so we can stick whatever error we want here?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -0,0 +1,132 @@
...
+            kind: RequestCheckKind::Network,
+            cause: ProcessingError::Recoverable(err.into()),
+        }
+    }
+
+    /// Records the error in [`REQUEST_CHECK_ERRORS`] and unwraps it into a [`ProcessingError`].
+    pub fn record(self) -> ProcessingError {
+        self.kind.inc_metric();
+        self.cause
+    }
```
dvdplm, line 119: Not sure about this. Why do we want `record` to do two jobs (push the metric AND convert the error)? Maybe it's good, maybe not, I'm honestly not sure. Just a bit surprising? 
If it was just `fn record(&self) { … }` and not return anything, wouldn't that work too?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+impl From<Erc1271Error> for RequestCheckError {
+    fn from(err: Erc1271Error) -> Self {
+        let kind = match &err {
+            Erc1271Error::Transport(_) => RequestCheckKind::Network,
+            _ => RequestCheckKind::Signature,
```
dvdplm: I think I'd prefer an explicit list of variants here. Noisy but safer (if anyone adds a non-signature variant we wouldn't know here).

kms-connector/crates/kms-worker/src/core/event_processor/kms.rs
```diff
@@ -50,7 +50,8 @@ where
             .map_err(ProcessingError::Irrecoverable)?;
         self.context_manager
             .validate_context(&parsed_extra_data)
-            .await?;
+            .await
+            .map_err(RequestCheckError::record)?;
```
dvdplm, line 54: I see why `record` returns a `ProcessingError` now... Maybe with the proper `From` impl(s) this could become `.await.map_err(RequestCheckError::record).into()?;`?

kms-connector/crates/utils/src/types/event.rs
```diff
@@ -560,25 +561,45 @@ impl Display for ProtocolEventKind {
...
-            ProtocolEventKind::Crsgen(e) => write!(f, "CrsgenRequest #{}", e.crsId),
-            ProtocolEventKind::NewKmsContext(e) => write!(f, "NewKmsContext #{}", e.contextId),
-            ProtocolEventKind::NewKmsEpoch(e) => write!(f, "NewKmsEpoch #{}", e.epochId),
         }
     }
 }
 
+fn u256_to_be_hex(int: U256) -> String {
+    format!("0x{}", hex::encode(int.to_be_bytes::<32>()))
+}
```
dvdplm: We need this because `U256` is LE, is that it? `U256` does impl  the standard `LowerHex` and `UpperHex`, but would that be the wrong endianess?

docs/metrics/metrics.md
eudelins-zama: This is just a recommendation for now, but yes.

Indeed a percentage of error is better, but not sure how easy it would be though, so I suggest to not over-promise and let the current reco.

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
eudelins-zama: To me, "allowed TO decrypt" feels like the handle is going to decrypt something, not that is going to be decrypted

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: I would argue no, because we want this conversion to happen only for `tonic::Status` of the response/result polling part, not of the request sending part (remember the 2 steps flow where we 1st send a request then poll ofr its result?°

eudelins-zama: (This is not new code btw)

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: I agree on this, it's not perfect, but was not able to find a better way. 

My idea is that I don't want us to be able to "forget" to increment the metric, and not to have code to handle the increment everywhere as well.

So using `From`, or having `record` decoupled from the conversion, we could miss the incrementation.

We could use a more verbose name for `record` to state that it also perform a conversion, but I personally prefer the current naming.

kms-connector/crates/kms-worker/src/core/event_processor/kms.rs
eudelins-zama: See https://github.com/zama-ai/fhevm/pull/2885#discussion_r3466900237

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: Not and endianess issue no, I was just not aware of it.

But if we want to leverage these trait, I think it will look like:

```
    format!("0x{int:064x}");
```

Not sure it's far better than the current version :sweat_smile:

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
dvdplm: I see, so "handle" is the object, not the subject. How about "Decryption is not allowed for {}"?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
dvdplm: You're saying that we get a `tonic::Status` back both when sending and receiving? But a sending error could also be recoverable or irrecoverable so it would still sort of make sense?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: Let's start simple with this strict type, and extend later if needed?

kms-connector/crates/utils/src/types/event.rs
dvdplm: I like it better! Also that probably doesn't allocate, whereas hex::encode probably does! And it's short enough that the helper is perhaps not needed? Either way, go with your gut, it's just a nitpick.

docs/metrics/metrics.md
dvdplm: I'd err on the side of caution and do "1 error per minute => alarm" and then dial it back when it gets annoying.

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: Yes, but the handling of the status is not identical for both case, so better to have this info in the name

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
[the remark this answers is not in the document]
eudelins-zama: Fixed 22624cae61cbdc35084434550735f462c4e147a6

kms-connector/crates/kms-worker/src/monitoring/metrics.rs
eudelins-zama: Fixed 3baf2de5585e404b273e167cf62d6e615ef42d7c

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: Fixed 22624cae61cbdc35084434550735f462c4e147a6

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: Fixed 65c683fe6f4c048492061b96721af63aa258e42b

docs/metrics/metrics.md
eudelins-zama: Don't want to wake up people for nothing, I prefer not alerting first and keep an eye on the metric, then adjust :wink: 
But that is a topic for later, when we'll implement the dashboard/alert for this

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
eudelins-zama: Fixed edadb8a3027e11f76a2f47f18941439b6dca3412

dvdplm (approved):

PR #2908, opened 2026-06-25
chore(kms-connector): operation recovery

eudelins-zama:
### What

Replaces the DB **garbage-collection** routine with an **operation-recovery** routine. Operations are no longer deleted from the connector's database — they are kept forever, and the background job only recovers stuck ones.

### Why

We want to stop deleting processed decryptions. More details in the [issue](https://github.com/zama-ai/fhevm-internal/issues/1548).

### Changes

- **No more deletion.** Dropped the logic that deleted `completed`/`failed` decryptions.
- **Recovery routine** (`operation_recovery.rs`, replacing `garbage_collection.rs`):
  - `unlock_stuck_operations` — resets `under_process` rows back to `pending` once they exceed `operation_under_process_timeout` (by `updated_at`). Applies to **all** operation tables.
  - `fail_expired_decryptions` — marks `pending`/`under_process` decryptions older than `decryption_expiry` (by `created_at`) as `failed`, to stop retrying them forever. Applies to **decryption** tables only.
- **Metrics.** Replaced `kms_connector_pending_events` and `kms_connector_pending_responses` with a single `kms_connector_unprocessed_operations{table,status}` gauge. It counts only `pending`/`under_process` rows so it stays bounded now that terminal rows are kept indefinitely.
- **Config / env vars** renamed accordingly

### Issues

Closes https://github.com/zama-ai/fhevm-internal/issues/1505
Closes https://github.com/zama-ai/fhevm-internal/issues/1548


kms-connector/crates/tx-sender/src/monitoring/mod.rs
```diff
@@ -1,3 +1,29 @@
...
+    "public_decryption_responses",
+    "user_decryption_responses",
+];
+
+/// DB table names of the operations handled by the KMS connector.
+const OPERATION_TABLES: &[&str] = &[
+    "public_decryption_requests",
+    "user_decryption_requests",
+    "public_decryption_responses",
+    "user_decryption_responses",
```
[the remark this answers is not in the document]
eudelins-zama: I don't see why it would be an issue tbh.
And `prss_init` and `key_reshare_same_set` are not used anymore

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+/// Marks `pending`/`under_process` decryptions older than `expiry` (by `created_at`) as `failed`.
+async fn fail_expired_decryptions(db_pool: &Pool<Postgres>, expiry: &PgInterval) {
+    for table in DECRYPTION_TABLES {
+        let query = format!(
+            "UPDATE {table} SET status = 'failed' \
+             WHERE status IN ('pending', 'under_process') AND NOW() - created_at > $1"
+        );
```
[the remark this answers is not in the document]
eudelins-zama: Interesting, will have a look to determine if it's an hallucination or not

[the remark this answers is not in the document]
eudelins-zama: Fixed a42efef4de5d9fb7c1b9e6a4e09e98ba176e550b

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
```diff
@@ -0,0 +1,91 @@
...
+/// (by `updated_at`) back to `pending` so they can be retried.
+async fn unlock_stuck_operations(
+    db_pool: &Pool<Postgres>,
+    operation_under_process_timeout: &PgInterval,
+) {
+    for table in OPERATION_TABLES {
+        let query = format!(
+            "UPDATE {table} SET status = 'pending' \
+             WHERE status = 'under_process' AND NOW() - updated_at > $1"
+        );
```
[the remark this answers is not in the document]
eudelins-zama: Fixed a42efef4de5d9fb7c1b9e6a4e09e98ba176e550b

kms-connector/crates/tx-sender/src/monitoring/metrics.rs
```diff
@@ -104,52 +92,47 @@ pub fn spawn_gauge_update_routine(
...
+/// failed.
+///
+/// `table` must be a trusted, hardcoded table name.
+async fn count_unprocessed(db_pool: &Pool<Postgres>, table: &str) -> Option<(i64, i64)> {
+    let query = format!(
+        "SELECT
+            COUNT(*) FILTER (WHERE status = 'pending') AS pending,
+            COUNT(*) FILTER (WHERE status = 'under_process') AS under_process
+        FROM {table}"
+    );
```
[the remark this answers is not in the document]
eudelins-zama: Fixed a42efef4de5d9fb7c1b9e6a4e09e98ba176e550b

kms-connector/crates/tx-sender/src/monitoring/metrics.rs
```diff
@@ -104,52 +92,48 @@ pub fn spawn_gauge_update_routine(
...
-            Err(e) => error!(error = %e, "Failed to fetch user decryption responses count"),
+/// Returns the `(pending, under_process)` row counts for the given table, or `None` if the query
+/// failed.
+///
+/// `table` must be a trusted, hardcoded table name.
+async fn count_unprocessed(db_pool: &Pool<Postgres>, table: &str) -> Option<(i64, i64)> {
+    let query = format!(
+        "SELECT
+            COUNT(*) FILTER (WHERE status = 'pending') AS pending,
+            COUNT(*) FILTER (WHERE status = 'under_process') AS under_process
```
dvdplm, line 125: I'd check the Postgres query optimizer with this query. Is `FILTER` doing the filtering inside  the query or can postgres do them in parallel? Is it faster to use sub queries, like `SELECT (SELECT COUNT(*) WHERE …) AS pending, (SELECT COUNT(*) WHERE …) as processing`? I worry that `FILTER` doesn't use the index properly.

kms-connector/crates/tx-sender/src/monitoring/mod.rs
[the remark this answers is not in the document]
dvdplm: Agreed. The bot is drunk.

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
```diff
@@ -0,0 +1,91 @@
...
+    operation_under_process_timeout: PgInterval,
+    db_pool: Pool<Postgres>,
+    cancel_token: CancellationToken,
+) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        select! {
+            _ = run_operation_recovery_routine(period, decryption_expiry, operation_under_process_timeout, db_pool) => {}
+            _ = cancel_token.cancelled() => {}
+        }
+    })
```
dvdplm, line 38: I think this  means that shutdown will be delayed by `run_operation_recovery_routine`. That is: if a shutdown signal comes in and we happen to just have begun `run_operation_recovery_routine`, then we will block until that has finished and we return here.
This is likely perfectly fine for now, but if `run_operation_recovery_routine` could ever become slow/long, then it'll be painful.

It'd be good to do some back of the envelope calculations for a moment where we have 10 years worth of data to check and are in the middle of some sort of crisis where we have a lot of ops stuck. Would it be a problem? Is there some intermediate step inside `run_operation_recovery_routine` where we can check the cancellation token and interrupt cleanly?

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
```diff
@@ -0,0 +1,91 @@
...
+    period: Duration,
+    decryption_expiry: PgInterval,
+    operation_under_process_timeout: PgInterval,
+    db_pool: Pool<Postgres>,
+) {
+    loop {
+        fail_expired_decryptions(&db_pool, &decryption_expiry).await;
+        unlock_stuck_operations(&db_pool, &operation_under_process_timeout).await;
+        tokio::time::sleep(period).await;
+    }
```
dvdplm, line 51: See above. Will we yield to the `select!` at each await here?

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/// Unlocks operations stuck in `under_process` for longer than `operation_under_process_timeout`
+/// (by `updated_at`) back to `pending` so they can be retried.
+async fn unlock_stuck_operations(
+    db_pool: &Pool<Postgres>,
+    operation_under_process_timeout: &PgInterval,
+) {
+    for table in OPERATION_TABLES {
```
dvdplm, line 77: Dumb question: will this loop lock/unlock each table for update, or will it be wrapped in one giant `BEGIN … COMMIT` across *all* tables? I think not, but worth double checking.

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
```diff
@@ -0,0 +1,91 @@
...
+        let query = format!(
+            "UPDATE {table} SET status = 'pending' \
+             WHERE status = 'under_process' AND updated_at < NOW() - $1"
+        );
+        match sqlx::query(&query)
+            .bind(*operation_under_process_timeout)
+            .execute(db_pool)
+            .await
```
dvdplm, line 85: Do we have to `await` each query here or can we fire all of them off and collect the results afterwards? Let Postgres  handle the concurrent queries itself?

kms-connector/crates/tx-sender/tests/recovery.rs
```diff
@@ -0,0 +1,318 @@
...
+        get_operation_status(test_instance.db(), operation, pending_id).await?,
+        OperationStatus::Pending
+    );
+
+    Ok(())
+}
+
+//////////////////////////////////////////////////////////////////////////////////////////////////
+//////////////////////////////////////////////////////////////////////////////////////////////////
+
```
dvdplm, line 177: Maybe remove these?

dvdplm (commented):
Code looks clean. Have a few questions/suggestions.

nit: One thing I don't love is the `under_process` naming. What's wrong with "processing". I'd also probably shorten "operation" to just "op" assuming the reader understands what the connector is doing.

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
eudelins-zama: > I think this means that shutdown will be delayed by `run_operation_recovery_routine`

I don't understand your initial assumption. AFAIU, in such case, the point of `tokio::select` is to return when the first future finish, and cancel the other future.
So in case of shutdown, we would exit the task and cancel the `run_operation_recovery_routine`, not wait for it to finish.

Am I missing something?

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
eudelins-zama: There is no giant `BEGIN ... COMMIT` no. But hard for me to prove that something does not exist :sweat_smile:

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
eudelins-zama: Bot's opinion:

> Re: blocking on shutdown — `select!` drops the routine future rather than awaiting it, and every step here is an .await on a DB round-trip or a sleep, so there's no synchronous section to delay cancellation. Where your scale concern is real is the queries themselves: each is a single unbounded UPDATE, so on a huge backlog it could hold locks / run long, and a cancel mid-query just rolls back that statement with no progress for that run. If that ever becomes a worry, batching the UPDATEs with a LIMIT would bound query cost and give clean per-batch cancellation points. Fine as-is for now.

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
eudelins-zama: As it's a background routine, I don't think we care that much about performance of the routine run.

So I don't see a huge benefits of making parallel queries. I am more concerned about saturating the DB connection pool by spawning > 10 parallel queries in a single run tbh...

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
eudelins-zama: Yes AFAIU

kms-connector/crates/tx-sender/src/monitoring/metrics.rs
eudelins-zama: Index usage isn't decided by the FILTER clause here but by the WHERE. The FILTERs run after rows are read (and already "filtered" by the WHERE), so they should neither nor hurts index usage.

Your approach with 2 nested sub-queries would make 2 reads of the entire tables so IMO it would be slower or equal, not faster.

eudelins-zama:
```
kms-connector=# EXPLAIN (ANALYZE, BUFFERS) SELECT COUNT(*) FILTER (WHERE status = 'pending') AS pending, COUNT(*) FILTER (WHERE status = 'under_process') AS under_process FROM user_decryption_requests WHERE status IN ('pending', 'under_process');
                                                                         QUERY PLAN                                                                          
-------------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=9.52..9.53 rows=1 width=16) (actual time=0.025..0.027 rows=1 loops=1)
   Buffers: shared hit=2
   ->  Bitmap Heap Scan on user_decryption_requests  (cost=4.16..9.50 rows=2 width=4) (actual time=0.020..0.021 rows=0 loops=1)
         Recheck Cond: (status = ANY ('{pending,under_process}'::operation_status[]))
         Buffers: shared hit=2
         ->  Bitmap Index Scan on idx_user_decryption_requests_status_updated_at  (cost=0.00..4.16 rows=2 width=0) (actual time=0.014..0.015 rows=0 loops=1)
               Index Cond: (status = ANY ('{pending,under_process}'::operation_status[]))
               Buffers: shared hit=2
 Planning:
   Buffers: shared hit=2
 Planning Time: 0.239 ms
 Execution Time: 0.092 ms
(12 rows)

kms-connector=# EXPLAIN (ANALYZE, BUFFERS) SELECT (SELECT COUNT(*) FROM public_decryption_requests WHERE status = 'pending') AS pending, (SELECT COUNT(*) FROM public_decryption_requests WHERE status = 'under_process') AS under_process;
                                                                              QUERY PLAN                                                                               
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Result  (cost=19.04..19.05 rows=1 width=16) (actual time=0.033..0.035 rows=1 loops=1)
   Buffers: shared hit=4
   InitPlan 1
     ->  Aggregate  (cost=9.51..9.52 rows=1 width=8) (actual time=0.022..0.023 rows=1 loops=1)
           Buffers: shared hit=2
           ->  Bitmap Heap Scan on public_decryption_requests  (cost=4.16..9.50 rows=2 width=0) (actual time=0.017..0.017 rows=0 loops=1)
                 Recheck Cond: (status = 'pending'::operation_status)
                 Buffers: shared hit=2
                 ->  Bitmap Index Scan on idx_public_decryption_requests_status_updated_at  (cost=0.00..4.16 rows=2 width=0) (actual time=0.010..0.011 rows=0 loops=1)
                       Index Cond: (status = 'pending'::operation_status)
                       Buffers: shared hit=2
   InitPlan 2
     ->  Aggregate  (cost=9.51..9.52 rows=1 width=8) (actual time=0.002..0.003 rows=1 loops=1)
           Buffers: shared hit=2
           ->  Bitmap Heap Scan on public_decryption_requests public_decryption_requests_1  (cost=4.16..9.50 rows=2 width=0) (actual time=0.002..0.002 rows=0 loops=1)
                 Recheck Cond: (status = 'under_process'::operation_status)
                 Buffers: shared hit=2
                 ->  Bitmap Index Scan on idx_public_decryption_requests_status_updated_at  (cost=0.00..4.16 rows=2 width=0) (actual time=0.001..0.001 rows=0 loops=1)
                       Index Cond: (status = 'under_process'::operation_status)
                       Buffers: shared hit=2
 Planning:
   Buffers: shared hit=4
 Planning Time: 0.264 ms
 Execution Time: 0.154 ms
(24 rows)
```

eudelins-zama: We can see the index is used right? :slightly_smiling_face:

kms-connector/crates/tx-sender/tests/recovery.rs
eudelins-zama: We can, only if I have to make other updates haha

kms-connector/crates/tx-sender/src/monitoring/metrics.rs
dvdplm: > The FILTERs run after rows are read (and already "filtered" by the WHERE), so they should neither nor hurts index usage.

That is exactly my worry: `FILTER` does its job **without** using the index, i.e. reads the data and *then* filters. This might be faster in the end, it all depends on how much data we're talking about. Using two subqueries might look slower now with small tables, but is likely faster  when the tables are full and the read returns a lot of rows (so `FILTER` must scan a big collection).

That said, it's premature optimization so whichever version you like better is fine!

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
dvdplm: > So in case of shutdown, we would exit the task and cancel the run_operation_recovery_routine, not wait for it to finish.

This is only true if we enter the `select!` and see the cancellation token. In that case we never run `run_operation_recovery_routine` and all is good. What I am talking about is this:

- enter `select!` section
- no cancellation token, proceed to `run_operation_recovery_routine`
- `run_operation_recovery_routine` starts, takes 20 minutes
- operator cancels
- operator panics and wonders WTF
- `run_operation_recovery_routine` eventually finishes
- next iteration: cancel token seen, shutdown happens

In other words: this code has a "smallest cancellation granularity" equal to "time it takes to run `run_operation_recovery_routine`". Fine now, might be nasty in a crisis.

The bot is saying the same thing, it just doesn't think it's worth worrying about. Might be true, might not.

The "fix" would be to do what the bot suggests, use bounded batches, but we could perhaps also check the cancellation token inside the `run_operation_recovery_routine` call at some strategic points to improve the shutdown latency.

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
dvdplm: I think Postgres is very good at handling and scheduling concurrent load and that use mere mortals should trust it to do the Right Thing™ as much as possible. It has a lot of battle tested code to do this.

Up to you, it likely doesn't matter here&now, but it's the better pattern imo.

kms-connector/crates/tx-sender/tests/recovery.rs
dvdplm: lol

dvdplm (approved):
My concerns are all about

1. naming
2. performance with a full DB

Each can be ignored and/or addressed separately. Approving.

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
eudelins-zama: I don't think your mental model is the right one for the `tokio::select!` (or I don't get your point).

It does not "proceed" one branch (`cancel_token.cancelled()`) or the other (`run_operation_recovery_routine`).

It will just return as soon as one of the two branches has finished, and `cancelled()` never finishes until the cancel token is cancelled.

So if `run_operation_recovery_routine` takes 20 min and token is cancelled somewhere else in the code after 5min, we won't wait for the other 15min, we would simply drop `run_operation_recovery_routine` future and return.

The only case where we could be blocked is if we have blocking sync code in the recovery routine AFAIU

dvdplm: Ok, could very well be. I suck at tokio-stuff. Thank you for the patience!

kms-connector/crates/tx-sender/src/monitoring/metrics.rs
eudelins-zama: Agree it's premature optimization and I'm happy to keep `FILTER` for now.

But thanks to the `WHERE` clause, the `FILTER` should be applied a fairly small amount of data, we should not have too many `under_process` or `pending` queries anyway.

kms-connector/crates/tx-sender/src/monitoring/operation_recovery.rs
eudelins-zama: Fully agree on trusting Postgres' scheduler,  my concern is purely client-side: the sqlx pool is bounded (~10) and shared with the tx-sender hot path, so spawning >10 UPDATEs contends for pool connections rather than stressing PG itself. For a background loop with no latency requirement I'd rather not compete with real work for connections.

eudelins-zama: Discussed offline

PR #2910, opened 2026-06-25
chore(kms-connector): speed up connector tests

eudelins-zama:
- Replaces the use of a fresh dedicated DB server container in every tests by using a shared DB server as a pre-requisite of the tests.
  - Created a `Makefile` to make this setup easier 
  - A new DB (inside the shared server) is still created for each test so there's no conflict between parallel tests.
- Spawns `anvil` as a standard process for each test instead of spawning it in a container.
- Uses `cargo nextest` in CI for better parallelization of all test binaries execution.
- Faster postgres test setup by disabling some crash-safety mechanism and tmpfs

Results:
- Before: https://github.com/zama-ai/fhevm/actions/runs/28167630308/job/83423603105?pr=2908
  - ~10m50s for test compilation+executions (not counting clippy etc.)
- After: https://github.com/zama-ai/fhevm/actions/runs/28847926966/job/85556156248?pr=2910
  - ~8m for test compilation+executions
 
Local gain are much better (without including compilation contrary to above), from ~7ms to ~40s, probably because of higher parallelization of the tests...


dvdplm (dismissed):
Looks sane to me.

I wonder if there's a way to tell Postgres to make an in-memory DB somehow? Or perhaps to use tempfiles only? Or tempfiles on a RAM-disk?

jot2re (dismissed):
Had a look at the surface and looks good, but I am of course not super familiar with the code base or the frameworks used

eudelins-zama:
> I wonder if there's a way to tell Postgres to make an in-memory DB somehow? Or perhaps to use tempfiles only? Or tempfiles on a RAM-disk?

Good idea! Tried to achieve such thing in 8db231779d9db16e56bce24115431050de0b51b3

jot2re (approved):

dvdplm (approved):
hell yeah! :)

eudelins-zama:
@mergifyio queue

isaacdecoded (approved):

PR #2919, opened 2026-06-29
feat(coprocessor): add confidential bridge round-trip coverage

PanGan21:
### Summary

Extends the bridge e2e suite with two round-trip cases the per-direction tests didn't cover:

- host->chain-b->host: re-bridge a bridge-derived handle and assert the value survives the round trip
- compute on a bridged handle on chain-b, then bridge the result back to host and assert the computed value (proves a bridged ciphertext is usable in an FHE op on the destination, not just decryptable)

Related: https://github.com/zama-ai/fhevm-internal/issues/1579


dartdart26 (approved):
Nice, thank you!

PR #2929, opened 2026-06-29
chore(kms-connector): kms context switch latest update

eudelins-zama:
### What

Adapts to latest `NewKmsEpoch` event and fetches the previous epoch material on demand.

Previously the event carried the full material inline:

```solidity
event NewKmsEpoch(..., PreviousKeyInfo[] keys, PreviousCrsInfo[] crsList);
```

Now it carries only the block at which that material is available, and the connector reads it
from the `KMSGeneration` contract when handling the event:

```solidity
event NewKmsEpoch(..., uint256 materialBlockNumber);
```

### Why

Here's the rational behind the changes made at the contract level:
- Initial "issue": https://github.com/zama-ai/fhevm/pull/2615#discussion_r3462478972
- Tech spec update: https://github.com/zama-ai/tech-spec/commit/3f858b37a0bb978d82429dd46a37cc9702a3a647

This PR simply aims to align the kms-connector with the contract and protocol updates.

### Changes

- **Event / ABI:** `NewKmsEpoch` drops `keys` + `crsList`, adds `materialBlockNumber`.
- **DB schema:** `new_kms_epoch` replaces `keys`/`crs_list` columns with `material_block_number`
  (migration + regenerated `.sqlx` descriptors + row read/write paths).
- **On-demand fetch:** new `ProtocolConfigProcessor::fetch_previous_epoch_info` queries
  `getCompletedKeyIds`/`getCompletedCrsIds` (pinned at `materialBlockNumber`) then
  `getKeyInfo`/`getCrsMaterials`; `prepare_new_kms_epoch_request` is now `async`.
- **Dependency bumps:** `kms-grpc`/`bc2wrap` → `b9087af…`, `tfhe` 1.6.1 → 1.6.2,
  plus the field renames that follow (`verification_key`→`signer_address`,
  `extra_verification_keys`→`extra_signer_addresses`, `deserialize_safe`→`deserialize_slice`).
  `fhevm_host_bindings` moves back to a path dependency.


charts/kms-connector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: kms-connector
 description: A helm chart to distribute and deploy the Zama KMS Connector services
-version: 1.5.2
+version: 1.5.3
```
eudelins-zama, line 3: To fix https://github.com/zama-ai/fhevm/actions/runs/28404342275/job/84163311556?pr=2700

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
```diff
@@ -132,6 +143,71 @@ impl<P: Provider> ProtocolConfigProcessor<P> {
...
+            let key_info = self
+                .kms_generation_contract
+                .getKeyInfo(key_id)
+                .block(material_block_number.into())
+                .call()
+                .await?;
+            let mut key_digests = vec![];
+            for d in key_info.keyDigests.iter() {
+                key_digests.push(KeyDigest {
+                    key_type: key_type_to_string(d.keyType)?,
```
eudelins-zama: Do you think we should return a `ProcessingError::Irrecoverable` here insteand of `anyhow::Error`?

If this fails:
- the RPC node is rogue/broken
  - thus making it Recoverable error gives us time for "fixing" the RPC node
- or there is a critical protocol error
  - in this case, we'll retry this context switch in a loop forever until we manually edit the DB...

This would create a bit of error handling overhead, but not too bad IMO.

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
```diff
@@ -132,6 +143,71 @@ impl<P: Provider> ProtocolConfigProcessor<P> {
...
+                    digest: d.digest.to_vec(),
+                });
+            }
+            keys_info.push(KeyInfo {
+                key_id: Some(u256_to_request_id(key_id)),
+                preproc_id: Some(u256_to_request_id(key_info.prepKeygenId)),
+                key_parameters: params_type_to_fhe_parameter(key_info.paramsType)?,
```
eudelins-zama: Same question here as [above](https://github.com/zama-ai/fhevm/pull/2929/changes#r3495005474)

kms-connector/crates/utils/src/types/event.rs
```diff
@@ -379,22 +376,19 @@ pub fn from_new_kms_context_row(row: &PgRow) -> anyhow::Result<ProtocolEvent> {
...
-
     let kind = ProtocolEventKind::NewKmsEpoch(NewKmsEpoch {
         kmsContextId: U256::from_le_bytes(row.try_get::<[u8; 32], _>("context_id")?),
         epochId: U256::from_le_bytes(row.try_get::<[u8; 32], _>("epoch_id")?),
         previousContextId: U256::from_le_bytes(row.try_get::<[u8; 32], _>("previous_context_id")?),
         previousEpochId: U256::from_le_bytes(row.try_get::<[u8; 32], _>("previous_epoch_id")?),
-        keys,
-        crsList: crs_list,
+        materialBlockNumber: U256::from_le_bytes(
+            row.try_get::<[u8; 32], _>("material_block_number")?,
```
dvdplm, line 390: It kind of feels like it should be possible to `row.try_get::<U256>` directly. No?

kms-connector/Cargo.toml
```diff
@@ -19,12 +19,10 @@ kms-worker.path = "crates/kms-worker"
...
 fhevm_gateway_bindings = { path = "../gateway-contracts/rust_bindings", default-features = false }
-# TODO: use path bindings again ASAP
-# fhevm_host_bindings = { path = "../host-contracts/rust_bindings", default-features = false }
-fhevm_host_bindings = { git = "https://github.com/zama-ai/fhevm.git", rev = "692a165b9bd99a16d9b09b47e2184090bd16797e", default-features = true }
-kms-grpc = { git = "https://github.com/zama-ai/kms.git", rev = "1edf3a089f8cc10e7bc5052e8f26d9cbaac66a57", default-features = true }
-bc2wrap = { git = "https://github.com/zama-ai/kms.git", rev = "1edf3a089f8cc10e7bc5052e8f26d9cbaac66a57", default-features = true }
-tfhe = { version = "=1.6.1", default-features = false }
+fhevm_host_bindings = { path = "../host-contracts/rust_bindings", default-features = false }
+kms-grpc = { git = "https://github.com/zama-ai/kms.git", rev = "b9087af025de26cd6ada5033e8044b05a518cf22", default-features = true }
+bc2wrap = { git = "https://github.com/zama-ai/kms.git", rev = "b9087af025de26cd6ada5033e8044b05a518cf22", default-features = true }
```
dvdplm, line 24: Maybe we should publish these two, or move them out of the `kms` repo so you don't have to do this ugliness.

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
dvdplm: The nicest thing would be to make it infallible. It seems like it can fail only if something really bad happens (DB is corrupted or someone somewhere snuck in a bad value) so one could argue that we should panic here and halt the server? (Or would a panic only crash the tokio worker perhaps?)
If that is unpalatable I think I'd prefer this to be `Irrecoverable` tbh.

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
dvdplm: I don't love that we mix test and production code like this. Best would be that a non-test build crashes loudly if the params type is anything but "default". Would that be possible with some clever `#[cfg(…)]` magic? If you don't like that approach I'd make it irrecoverable like above (or keep the anyhow result but make sure we do the right thing at the call site one level up).

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
```diff
@@ -132,6 +143,71 @@ impl<P: Provider> ProtocolConfigProcessor<P> {
...
+                key_digests,
+            });
+        }
+
+        let mut crs_info = vec![];
```
dvdplm: nit: all the lengths are known here. I bet the compiler sees this and does `with_capacity` but if you want to give it the hint explicitly...

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
```diff
@@ -132,6 +143,71 @@ impl<P: Provider> ProtocolConfigProcessor<P> {
...
+        Ok(PreviousEpochInfo {
+            context_id: Some(u256_to_request_id(context_id)),
+            epoch_id: Some(u256_to_request_id(previous_epoch_id)),
+            keys_info,
+            crs_info,
+        })
+    }
 }
 
 fn build_new_kms_context_grpc_from_event(
```
dvdplm, line 231: It seems like neither callers into this fn actually use the `NewKmsContext` after the call, so why pass it as a reference? If you move the event instead you can perhaps get rid of some clones/to_vec below?

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: Yeah unfortunately I think panicking would only crash the tokio worker.

Will go for the `Irrecoverable` I think then

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: No, Postgres have no type for such big integers, so I don't see how `row.try_get::<U256>` could work.

Anyway, we chose to represent U256 as bytes in every tables, so here as well. Thus we need to apply the usual LE encoding/decoding.

Missing something?

kms-connector/Cargo.toml
eudelins-zama: Would appreciate a pre-release, but don't want to delay this PR for this :slightly_smiling_face:

eudelins-zama: Will update in a follow up

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: Hmm, I get your point.

The problem I see would be e2e tests, where we want to use prod images, with test params AFAIU.
The moment you introduce features and/or "`#[cfg(…)]` magic", then you don't really test what goes to `main`...
Worth more discussion though, and/or a follow up issue to decide what we want to do about this.

Keeping the current behavior in this PR, and will use the Irrecoverable approach for now

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: > It seems like neither callers into this fn actually use the NewKmsContext after the call

Not sure this is true: https://github.com/zama-ai/fhevm/blob/5497856b5d5fa97461d0404b5f87bc302612144d/kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs#L65-L69
(`event` is used twice, and is a `&NewKmsContext`)

Also, we need the `to_vec` to build the GRPC requests for the `kms-core` anyway. (And `clone` as well I think).

Plus the context switch is not a super critical path, so not sure it would be a game changer :smile:

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: Fixed 969fec253caf683af7d600f84e5d878c6ebe9b7e

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: Fixed 969fec253caf683af7d600f84e5d878c6ebe9b7e

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: Fixed 969fec253caf683af7d600f84e5d878c6ebe9b7e

kms-connector/crates/utils/src/types/event.rs
dvdplm: No, you're right. If there was a `TryGet` trait we own, we could impl it but we don't (?) so... ignore.

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
dvdplm: This is what I had in mind:

```rust
    pub async fn prepare_new_kms_context_request(
        &self,
        event: NewKmsContext, // didn't check if we can take this by value
    ) -> Result<KmsGrpcRequest, ProcessingError> {
        // get the previous one first, pass event by ref
        let previous_context_event = self
            .fetch_previous_context_creation_event(&event)
            .await
            .map_err(ProcessingError::Recoverable)?;
        // build `new`, consumes `event`
        let new = build_new_kms_context_grpc_from_event(event)?;
        let old = build_new_kms_context_grpc_from_event(&previous_context_event)?;
        Ok(KmsGrpcRequest::NewMpcContext { new, old })
    }
```

Not sure if it could work though.

eudelins-zama: Ah got it.

The "top caller" `prepare_request` only holds `&mut ProtocolEvent`, so to have a raw `NewKmsContext`, we would have to clone it at some point, so I think we're stuck with the ref here

dvdplm (approved):

PR #2930, opened 2026-06-30
fix(host-contracts): replace signer by txSender on confirmKmsContextCreation

isaacdecoded:
## Summary

Adds end-to-end coverage for KMS context switching on a threshold-mode cluster —
context rotation, epoch rotation, and a genuine node swap.

## E2E test coverage

A new `kms-context-switch` acceptance profile (`./fhevm-cli test kms-context-switch`)
drives the two `ProtocolConfig` KMS-context lifecycle flows end to end and proves the
KMS reacts to the emitted events:

1. **New context** — broadcast `defineNewKmsContextAndEpoch`, wait for the new context to
   become the on-chain active one (`getCurrentKmsContextAndEpoch`), then decrypt.
2. **New epoch** — broadcast `defineNewEpochForCurrentKmsContext` (same context, new epoch),
   wait for the epoch to activate, then decrypt again.

Activation is not automatic: the cores must reshare and the connectors must submit
`confirmKmsContextCreation` / `confirmEpochActivation` before the active ids advance. If
that never happens the profile fails.

The profile runs against two scenarios:

### `four-party-threshold-kms` — same-committee switch
4-party threshold cluster (t=1). The switch keeps the committee unchanged (all members
reshare as "both"), then rotates the epoch. Proves the cluster reshares, activates, and
still decrypts under both the new context and the new epoch.

### `swap-threshold-kms` — node swap
5-core cluster: an initial 4-node committee (t=1) plus one **idle spare** (party 5, boots
with `peers=None`). The switch defines a new committee `{1,2,3,5}` — drops committee node 4
and promotes the spare into its MPC position — keeping `n = 4` so the threshold stays valid.
This exercises the core's **asymmetric two-sets resharing** with a real membership change:
node 4 leaves (outgoing set), node 5 joins (incoming set), the rest reshare as both.
Activation gates on the spare — the context id only advances once **all** new-committee
signers, including the spare, confirm.

After the switch + epoch rotation + decrypts, a final check proves the swapped-in spare
actually holds a working reshared key: it stops the top-`t` continuing members so the live
committee is exactly the `2t+1` quorum **including** the spare, then decrypts — forcing the
spare to participate.

## CI

Two jobs added to `test-suite-orchestrate-e2e-tests.yml`, mirroring
`run-e2e-tests-threshold-kms` and **non-gating** like it:
- `run-e2e-tests-kms-context-switch` — `four-party-threshold-kms` + `kms-context-switch`
- `run-e2e-tests-kms-node-swap` — `swap-threshold-kms` + `kms-context-switch`

Closes https://github.com/zama-ai/fhevm-internal/issues/1521
Closes https://github.com/zama-ai/planning-blockchain/issues/1137


host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -347,27 +347,29 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        if (isPreviousTxSender) {
             ++$.contextCreationPreviousSignerConfirmationCount[kmsContextId];
         }
-        if (isNewSigner) {
+        if (isNewTxSender) {
             ++$.contextCreationNewSignerConfirmationCount[kmsContextId];
         }
 
-        emit KmsContextCreationConfirmation(kmsContextId, signer, isPreviousSigner, isNewSigner);
+        emit KmsContextCreationConfirmation(kmsContextId, signer, isPreviousTxSender, isNewTxSender);
```
obatirou: definition of the event should be updated

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -347,27 +347,29 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
         // Record the confirmation and counts separately for the split quorum.
         $.contextCreationConfirmedBySigner[kmsContextId][signer] = true;
-        if (isPreviousSigner) {
+        if (isPreviousTxSender) {
             ++$.contextCreationPreviousSignerConfirmationCount[kmsContextId];
         }
-        if (isNewSigner) {
+        if (isNewTxSender) {
             ++$.contextCreationNewSignerConfirmationCount[kmsContextId];
         }
```
obatirou, line 367: Should state variable names be updated to reflect the changes?

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -347,27 +347,29 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        address signer = isNewTxSender
+            ? $.kmsNodeByTxSenderForContext[kmsContextId][msg.sender].signerAddress
+            : $.kmsNodeByTxSenderForContext[previousContextId][msg.sender].signerAddress;
         if ($.contextCreationConfirmedBySigner[kmsContextId][signer]) {
             revert KmsContextCreationAlreadyConfirmed(signer, kmsContextId);
         }
 
         // Record the confirmation and counts separately for the split quorum.
         $.contextCreationConfirmedBySigner[kmsContextId][signer] = true;
```
obatirou: Why involve the signer here?

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -72,6 +79,111 @@ const waitForActivation = async (
...
+const proveSpareInQuorum = async (state: State, runDecryption: DecryptionRunner) => {
+  const { kms } = state.scenario;
+  const { continuing, added } = committeeSwapPlan(kms);
+  const reconstruct = reconstructionThreshold(kms.threshold);
+  const forced = continuing.slice(continuing.length - kms.threshold);
+  console.log(
+    `[kms-context-switch] stopping ${forced.join(",")} so spare(s) ${added.join(",")} are required for the ${reconstruct}/${kms.committeeSize} (2t+1) quorum…`,
+  );
+  await setRunning(
+    forced.flatMap((party) => partyContainers(party)),
```
obatirou: you could have a `const forcedContainers = forced.flatMap((party) => partyContainers(party))` as it is redeclared below for the stop

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: Yes, updated [here](https://github.com/zama-ai/fhevm/pull/2930/changes/af35aa73a4272831e364149876037b05e9d9fc8f#diff-46eb54d4359fd5be9ee85fb292f7a594d7f52c9429506e71b1a745f3075be0a7R126-R128).

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: Indeed, updated [here](https://github.com/zama-ai/fhevm/pull/2930/changes/af35aa73a4272831e364149876037b05e9d9fc8f#diff-ba77c9fc763d9a63be867533f531523aa9b6601802eb7fa378a41c9e1f43092fR139-R144).

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: Yes, no signer involved anymore - it's [the tx sender instead](https://github.com/zama-ai/fhevm/pull/2930/changes/af35aa73a4272831e364149876037b05e9d9fc8f#diff-ba77c9fc763d9a63be867533f531523aa9b6601802eb7fa378a41c9e1f43092fR361).

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: Good catch, suggested change added [here](https://github.com/zama-ai/fhevm/pull/2930/changes/af35aa73a4272831e364149876037b05e9d9fc8f#diff-5f6f216e91265a0cf179d166e5e837d0cfd9dd3bfb84fe0d206dbd71ee9f25a1R164).

.github/workflows/test-suite-orchestrate-e2e-tests.yml
```diff
@@ -294,3 +294,77 @@ jobs:
       scenario: four-party-threshold-kms
       test-profile: kms-generation
       test-suite-version: ${{ needs.create-e2e-tests-input.outputs.test-suite-version }}
+
+  # Same orchestrated images/lock, exercised against the kms-context-switch acceptance profile on the
+  # unchanged 4-party committee: a new context (same committee, all-Both reshare) then a new epoch,
+  # proving the cluster reshares/activates and still decrypts under each. Non-gating, like the leg above.
+  # Standalone: gh workflow run test-suite-e2e-tests.yml -f scenario=four-party-threshold-kms \
+  #   -f test-profile=kms-context-switch -f build=false
+  run-e2e-tests-kms-context-switch:
```
obatirou, line 303: As this is a new job, shouldn't this be added to mergify checks?

obatirou: after a double check, all 3 e2e-tests related to KMS should be added no?

obatirou: for now we have in .mergify.yml `check-success = run-e2e-tests / fhevm-e2e-test`
I am not an expert on this but seems we would want then to be checked

obatirou: Sorry for the noise I just saw the comment above:

> It is intentionally NOT a merge-queue gate yet
  (.mergify.yml gates on `run-e2e-tests / fhevm-e2e-test` only) — promote it to a merge_condition
  once the threshold leg is proven stable.

It seems to be this is stable now no?

test-suite/fhevm/src/generate/env.ts
```diff
@@ -340,6 +342,63 @@ const applyKmsThresholdGatewayEnv = async (
...
+const kmsSwapSlots = (plan: StackSpec): { slot: number; src: number }[] => {
+  const { parties, committeeSize } = plan.kms;
+  const spares = parties - committeeSize;
+  return Array.from({ length: spares }, (_, k) => ({ slot: committeeSize - spares + k, src: committeeSize + k }));
+};
+
+/** host ProtocolConfig swap-committee env for `defineNewKmsContextAndEpoch`. The spare's connection
+ * vars (tx-sender/ip/storage-url) live only in gateway-sc (host copies just the committee), so they
+ * are pulled from there; its identity + discovered signer/CA-cert come from host-sc. Returns
+ * undefined for non-swap topologies or before the spare's signer is discovered. */
```
obatirou: Why `live only in gateway-sc`?
ProtocolConfig ethereum is the source of truth now, 0.13 is already applied.

.github/workflows/test-suite-orchestrate-e2e-tests.yml
isaacdecoded: Yes, the stability comment refers to the pre-existing kms-generation threshold leg introduced by Elias when adding support for KMS threshold-mode - this one, I guess we can promote and include as merge_condition since it has enough green history.

But for the other two new legs this PR adds (kms-context-switch, kms-node-swap), I think we can wait a bit more since we don't fully know their behavior long-term, and I'd like to avoid including in the merge_condition something we don't know if it's stable or not. I can create a follow-up issue to promote them (or the three of them at once) after having some green history as well.

Wdyt?

host-contracts/contracts/ProtocolConfig.sol
obatirou, line 181: I do not remember what was decided for this regarding chicken and egg problem. Should this be E2E tested?

.github/workflows/test-suite-orchestrate-e2e-tests.yml
obatirou: ok for me if we do not forgot and that those tests are actually run when we do modification on this feature

test-suite/fhevm/src/generate/env.ts
isaacdecoded: You're right, it was a misleading approach I missed to correct. Solved [here](https://github.com/zama-ai/fhevm/pull/2930/changes/bdfd1880db55d81ee4a858d16eeeba1be2a97e8d), thanks!

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: We rely on the current behavior by design, in which the KMS nodes create the genesis context (and epoch) during the protocol boot (see [here](https://github.com/zama-ai/tech-spec/blob/main/architecture/flows/mtls.md?plain=1#L11-L14) and [here](https://github.com/zama-ai/tech-spec/blob/main/architecture/flows/mtls.md?plain=1#L25)) - meaning that the ProtocolConfig contract will activate these genesis context and epoch IDs on initial deployment.

.github/workflows/test-suite-orchestrate-e2e-tests.yml
isaacdecoded: Created issue [here](https://github.com/zama-ai/fhevm-internal/issues/1616).

Eikix: Could the KMS e2e coverage be concentrated into the one threshold-context leg that you mention (the one with enough green history)?

We could make it so that it always boots the 5-party topology with a 4-party active committee and one idle spare. The current split pays the threshold bootstrap cost three times, while the node-swap topology should also be a valid baseline topology for the generation and same-committee context-switch checks.

That would also make the test closer to the invariant we need: after each KMS context/epoch transition, normal fhEVM application flows should still work. The profile could run the existing light smoke (`input-proof` + `erc20`) at baseline, after same-committee context switch, after epoch rotation, and after node swap. This would check that context switching remains transparent to app-level encrypted input/decryption/ERC20 flows, not only that a dedicated KMS decrypt probe succeeds.

The tradeoff is that the existing threshold baseline would no longer be the minimal 4-node topology; it would become a 5-party deployment with one idle spare. If that distinction is not important for CI, grouping looks simpler and should save runner minutes. And the total time would be ~50mn which is close to the real e2e tests anyway.

jot2re (dismissed):
I had a look, along with Claude, and it seems be to aligned with the [RFC 005 updates.](https://github.com/zama-ai/tech-spec/pull/479/changes)
But I am of course not super familiar with this code base.

The one thing I was thinking about was if there were any changes needed to `confirmEpochActivation` which still uses signer and not txSender. But at least, as I could understand from Claude, what is currently in play is correct. Just wanted to mention for completeness

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -72,6 +79,106 @@ const waitForActivation = async (
...
+ */
+const switchKmsContext = async (
+  state: State,
+  runDecryption: DecryptionRunner,
+  target: SwitchTarget,
+  baseline: ContextAndEpoch,
+): Promise<ContextAndEpoch> => {
+  const { isSwap, continuing, added, dropped } = committeeSwapPlan(state.scenario.kms);
+  // Same services as a same-set switch; a swap just points them at the swap-committee env files.
+  const hostEnv: Record<string, string> = isSwap ? { HOST_SC_CONTEXT_ENV: "host-sc-swap.env" } : {};
```
Eikix, line 122: The profile currently derives whether the context switch is a node replacement from `parties > committeeSize`. 

That makes topology select the behavior, so a 5-party scenario cannot naturally run both a same-committee context renewal and a spare-promotion context in the same profile. 

The scenario should describe the available topology, while the profile should choose the transition step explicitly, e.g. `same committee` first and `promote spare` later. The task env can then be derived from that requested transition instead of selected through `HOST_SC_CONTEXT_ENV` / `GATEWAY_SC_CONTEXT_ENV`.


Can be done in a follow-up. Shouldn't stop the code freeze

.github/workflows/test-suite-orchestrate-e2e-tests.yml
Eikix: Can be done in a follow-up. It shouldn't stop code freeze.

Eikix: One thing we could probably do now is add in the profile of the context switch small tests such as input-proof before during and after if possible to check if everything behaves as expected.

Eikix (dismissed):
overall lgtm, and interesting new CI additions

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -911,10 +910,10 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
     function _hasContextCreationQuorum(uint256 contextId) internal view virtual returns (bool) {
         ProtocolConfigStorage storage $ = _getProtocolConfigStorage();
         return
-            $.contextCreationNewSignerConfirmationCount[contextId] ==
+            $.contextCreationNewTxSenderConfirmationCount[contextId] ==
             $.kmsSignerAddressesForContext[contextId].length &&
```
obatirou: nit: 
`contextCreationNewTxSenderConfirmationCount` is compared to `kmsSignerAddressesForContext`
even if correct as we have 1-1 signer/txsender, I think it would make more sense to use `kmsNodesForContext[contextId].length`

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -300,9 +300,9 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
         $.contextState[contextId] = ContextState.Pending;
         _createPendingEpoch(contextId);
 
-        // Cache the previous-context signer quorum needed by confirmKmsContextCreation.
-        // `n - t + 1` (n = previous signers, t = previous MPC fault threshold) so no t-subset ratifies alone.
-        $.contextCreationPreviousSignerThreshold[contextId] =
+        // Cache the previous-context confirmation quorum needed by confirmKmsContextCreation.
+        // `n - t + 1` (n = previous committee size, t = previous MPC fault threshold) so no t-subset ratifies alone.
+        $.contextCreationPreviousTxSenderThreshold[contextId] =
             $.kmsSignerAddressesForContext[previousContextId].length -
```
obatirou: same here

obatirou (dismissed):

obatirou (dismissed):

obatirou (approved):
lgtm
could you run  make conformance to also ensure prettier is ran?

.github/workflows/test-suite-orchestrate-e2e-tests.yml
isaacdecoded: So, for the consolidation of the three kms-threshold legs into a single one, I updated the [follow-up issue](https://github.com/zama-ai/fhevm-internal/issues/1616) to consider it: see Track 3. At this time, I think it's better to not delay this PR for the code freeazing, and also I think it's a good approach to keep the separate/granular for a better debugging/maintenance - still we can always merge them if we discover it's a better way

Then, for the smoke tests suggested in the last comment, they were added [here](https://github.com/zama-ai/fhevm/pull/2930/changes/5fe344a856667a9f242e24a6a31433877154b5d4).

cc @Eikix

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -303,7 +303,7 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
         // Cache the previous-context confirmation quorum needed by confirmKmsContextCreation.
         // `n - t + 1` (n = previous committee size, t = previous MPC fault threshold) so no t-subset ratifies alone.
```
jot2re, line 304: Not really related to this PR but I noticed that the bound here is 13-4+1=10. Should it not be 9? I.e. 2t+1?

jot2re (approved):
Still looks good! Just noticed one thing though, although not related to this PR

isaacdecoded:
@mergifyio queue

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: It's a good catch indeed. Also updated the [follow-up issue](https://github.com/zama-ai/fhevm-internal/issues/1616) to refactor it (see Track 2)

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: So, this is a VERY important catch. Created this [follow-up issue](https://github.com/zama-ai/fhevm-internal/issues/1618) and closed the conversation to unblock the merging before code freezing.

jot2re: Thanks! Yeah, but at least it just means we are overly conservative instead of overly liberal with the security :) Of course if indeed t parties were to be corrupt, then they could block execution

obatirou: Does this means we need to put in in 0.14?
Seems we should?

PR #2933, opened 2026-06-30
fix: update glibc dynamic version to v16

fegmorte:


relayer/docker/relayer-migrate/Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN --mount=type=bind,source=relayer/relayer-migrate/Cargo.toml,target=Cargo.tom
 # runtime dependencies for the application. This often uses a different base
 # image from the build stage where the necessary files are copied from the build
 # stage.
-FROM cgr.dev/zama.ai/glibc-dynamic:15.2.0-dev AS prod
+FROM cgr.dev/zama.ai/glibc-dynamic:16-dev AS prod
```
eudelins-zama, line 38: Same comment as in the other [PR](https://github.com/zama-ai/kms/pull/676#discussion_r3499504462).

Basically, we should consider stripping the `-dev` from every images that ends with `... AS prod`

eudelins-zama (commented):
I think we need to update `golden-container-images/rust-glibc/Dockerfile` as well

relayer/docker/relayer-migrate/Dockerfile
fegmorte: we cannot because we need a package manager

fegmorte:
> I think we need to update `golden-container-images/rust-glibc/Dockerfile` as well

It's been already updated

relayer/docker/relayer-migrate/Dockerfile
eudelins-zama: We should be able to get rid of it IMO, but netter to do it in another PR

eudelins-zama (dismissed):

eudelins-zama (approved):

paulo-zama (approved):

eudelins-zama:
@mergifyio queue

obatirou (approved):

PR #2935, opened 2026-06-30
feat(host-contracts): stress test script for handlesList cOApp

jatZama:


host-contracts/lz-wiring/layerzero.config.testnet.ts
```diff
@@ -19,7 +19,7 @@ const pathways: TwoWayConfig[] = [
         // [requiredDVN[], [optionalDVN[], threshold]] — TESTNET-grade single DVN.
         [['LayerZero Labs'], []], // WARNING: This is a testnet configuration, for production use at least 2 independent operators, e.g. [['LayerZero Labs', 'Google Cloud'], []].
         // Block confirmations [src→dst, dst→src].
-        [1, 1], // WARNING: This is a testnet configuration, for production use 15 and 120 confirmations respectively.
+        [15, 120],
```
PanGan21, line 22: Is this correct change for testnet? I tested with [1,1] so I don't know what is the effect on Sepolia<>Amoy

host-contracts/scripts/stress/handlesListBridgeStress.ts
```diff
@@ -0,0 +1,560 @@
...
+        `Timed out after ${MAX_WAIT_MS / 60000} minutes with ${pending.size}/${records.length} deliveries not confirmed SUCCESS.`,
+      );
+    }
+
+    const alerted = [...pending].filter((g) => {
+      const r = guidToRecord.get(g);
+      return r && (r.receiveAlerts > 0 || r.composeAlerts > 0);
+    }).length;
+    console.log(
+      `  SUCCESS ${records.length - pending.size}/${guidToRecord.size}; ${pending.size} pending ` +
```
PanGan21: Not a blocker and optional to fix.
I think using `guidToRecord.size` instead of `records.length` might be more correct here because it keeps the numerator and denominator on the same base  whereas `records.length` includes failed sends and so counts them as successes

PanGan21 (dismissed):

host-contracts/lz-wiring/layerzero.config.testnet.ts
jatZama: Yes, we want to use the recommended number of confirmations on testnet as well, to avoid issues in case of reorg. Initially I put 1,1 here only to go faster when testing, but this is not a good value for "production testnet". 15,120 is the officially recommended values for Sepolia,Amoy according to LZ docs.

host-contracts/scripts/stress/handlesListBridgeStress.ts
jatZama: Good catch thanks. Fixed in new commit.

PanGan21 (dismissed):

host-contracts/scripts/stress/handlesListBridgeStress.ts
jatZama: Fixed in https://github.com/zama-ai/fhevm/pull/2935/changes/98aa3bc805249a26c9308a6fdc003c02c6c979fa

PanGan21 (dismissed):

PanGan21 (approved):

PR #2936, opened 2026-07-01
fix(gateway-contracts): cache priority coprocessor tx sender to avoid redundant GatewayConfig read

obatirou:


obatirou:
Acknowledged in the end

PR #2673, opened 2026-05-28
feat(host-contracts): add abortKeygen and abortCrsgen tasks

obatirou:

PR #2947, opened 2026-07-01
feat(kms-connector): add abort key crs gen

jot2re:
This PR adds the ability for the KMS Connector to handle abort requests for key and crs generation. 
It will pick up these events from the contracts and relay them to the KMS Core. 
This is handled by adding a new table in the DB containing information on requests that might have been aborted, and hence distinguish between a "normal" error and an abort. 

This should be merged after https://github.com/zama-ai/fhevm/pull/2673 

Closes https://github.com/zama-ai/fhevm-internal/issues/1271


kms-connector/connector-db/migrations/20260701000000_add_abort_key_crs_gen.sql
```diff
@@ -0,0 +1,98 @@
...
+    tx_hash BYTEA,
+    created_at TIMESTAMPTZ NOT NULL,
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    otlp_context BYTEA NOT NULL,
+    status operation_status NOT NULL DEFAULT 'pending',
+    already_sent BOOLEAN NOT NULL DEFAULT FALSE,
+    PRIMARY KEY (crs_id)
+);
+
+--- TODO it is a bit unclear to me if all these steps below are indeed needed
```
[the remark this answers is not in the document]
eudelins-zama: Removed, premature optimization. Tables will be small, no need for index yet.

kms-connector/crates/utils/src/tests/db/requests.rs
```diff
@@ -517,6 +575,14 @@ pub async fn check_no_uncompleted_request_in_db(
         EventType::CrsgenRequest => {
             "SELECT COUNT(crs_id) FROM crsgen_requests WHERE status NOT IN ('completed', 'failed')"
         }
+        EventType::AbortKeygenRequest => {
+            "SELECT COUNT(prep_keygen_id) FROM abort_keygen_requests
+            WHERE status NOT IN ('completed', 'failed')"
+        }
+        EventType::AbortCrsgenRequest => {
```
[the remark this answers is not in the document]
eudelins-zama: Only for keygen/crsgen

kms-connector/crates/utils/src/tests/db/requests.rs
```diff
@@ -517,6 +575,14 @@ pub async fn check_no_uncompleted_request_in_db(
         EventType::CrsgenRequest => {
             "SELECT COUNT(crs_id) FROM crsgen_requests WHERE status NOT IN ('completed', 'failed')"
         }
```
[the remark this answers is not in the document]
eudelins-zama: Done

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -6,21 +6,23 @@ use user_decryption_signature::Erc1271Error;
 
 #[derive(Debug, Error)]
 pub enum ProcessingError {
-    #[error("Processing failed with irrecoverable error : {0}")]
+    #[error("Processing failed with irrecoverable error: {0}")]
     Irrecoverable(anyhow::Error),
     #[error("Processing failed: {0}")]
     Recoverable(anyhow::Error),
+    #[error("Processing stopped: the operation was aborted on the KMS Core")]
```
dvdplm, line 13: "…aborted on the KMS Core" sounds like the flow is "KMS core tells connector about an abort" but it's the other way around yeah? How about just "Processing aborted"?

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
```diff
@@ -124,6 +126,12 @@ impl KmsClient {
             KmsGrpcRequest::PrepKeygen(req) => self.poll_prep_keygen_result(req).await,
             KmsGrpcRequest::Keygen(req) => self.poll_keygen_result(req).await,
             KmsGrpcRequest::Crsgen(req) => self.poll_crsgen_result(req).await,
+            // Abort has no result-polling endpoint on the Core: the send-side ack is the only
+            // signal. The original keygen/crsgen request is separately retired when its own result
+            // poll returns `Code::Aborted`.
```
dvdplm, line 131: @jot2re We should probably file a ticket to improve this on the KMS side? Provide some sort of information about the abort as it progresses on that side?

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -319,6 +327,46 @@ async fn publish_crsgen_request<'e>(
...
+async fn publish_abort_keygen_request<'e>(
+    executor: impl PgExecutor<'e>,
+    request: AbortKeygen,
+    tx_hash: Option<FixedBytes<32>>,
+    created_at: DateTime<Utc>,
+    otlp_ctx: PropagationContext,
+) -> anyhow::Result<PgQueryResult> {
+    sqlx::query!(
+        "INSERT INTO abort_keygen_requests(prep_keygen_id, tx_hash, created_at, otlp_context)
+            VALUES ($1, $2, $3, $4) ON CONFLICT DO NOTHING",
```
eudelins-zama, line 339: Just thinking: I think we should do something on conflict (override) not nothing.
Otherwise it means we could try the cancellation of a given keygen/crsgen only once.

kms-connector/crates/kms-worker/tests/integration_tests.rs
```diff
@@ -176,8 +180,16 @@ async fn test_processing_request(
...
-    check_response_data(&request, response)?;
+    if matches!(
+        event_type,
+        TestEventType::AbortKeygen | TestEventType::AbortCrsgen
+    ) {
+        // Abort events yield no response row, so just wait for it to be `completed`
+        wait_for_abort_completed(test_instance.db(), event_type).await?;
+    } else {
+        let response = wait_for_response_in_db(test_instance.db(), &request).await?;
+        check_response_data(&request, response)?;
```
dvdplm, line 191: Didn't finish reading yet, but it would perhaps have been nicer to bake this logic into `wait_for_response_in_db` and keep the code here the same as before?

kms-connector/crates/kms-worker/tests/integration_tests.rs
```diff
@@ -221,6 +235,14 @@ fn prepare_mocks(req: &ProtocolEventKind, already_sent: bool) -> MockSet {
...
     }
 
+    // Abort endpoints return `Empty` directly and have no result-polling endpoint
+    if matches!(
+        req,
+        ProtocolEventKind::AbortKeygen(_) | ProtocolEventKind::AbortCrsgen(_)
+    ) {
+        return kms_mocks;
+    }
+
```
dvdplm: Why not do the same as we do for `NewKmsContext` for these? Aren't they the same in this regard?

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Forget about it, it's not possible anyway at the contract level: https://github.com/zama-ai/fhevm/blob/main/host-contracts/contracts/KMSGeneration.sol#L510

kms-connector/crates/kms-worker/tests/integration_tests.rs
```diff
@@ -355,8 +386,36 @@ fn check_response_data(request: &ProtocolEventKind, response: KmsResponse) -> an
...
+        }
+        TestEventType::AbortCrsgen => {
+            "SELECT COUNT(crs_id) FROM abort_crsgen_requests WHERE status = 'completed'"
+        }
+        _ => unreachable!("wait_for_abort_completed only handles abort events"),
+    };
+    loop {
+        let count: i64 = sqlx::query_scalar(query).fetch_one(db).await?;
+        if count > 0 {
+            info!("OK!");
```
dvdplm: Is this log really useful?

kms-connector/crates/kms-worker/tests/integration_tests.rs
```diff
@@ -355,8 +386,36 @@ fn check_response_data(request: &ProtocolEventKind, response: KmsResponse) -> an
...
+            return Ok(());
+        }
+        warn!("Not yet...");
```
dvdplm: same

kms-connector/crates/utils/src/types/event.rs
```diff
@@ -80,6 +81,14 @@ impl ProtocolEvent {
         self.update_status(db, OperationStatus::Failed).await
     }
 
+    /// Sets the event's `status` field to `aborted` in the database.
+    pub async fn mark_as_aborted(&self, db: &Pool<Postgres>) {
```
dvdplm: nit: I think `mark_aborted` reads just as well.

kms-connector/crates/utils/src/types/event.rs
```diff
@@ -80,6 +81,14 @@ impl ProtocolEvent {
...
+        info!(
+            "Event was aborted on the KMS Core. Setting its `status` field to `aborted` in DB..."
+        );
+        self.update_status(db, OperationStatus::Aborted).await
+    }
+
     async fn update_status(&self, db: &Pool<Postgres>, status: OperationStatus) {
```
dvdplm: It's a bit odd that this method doesn't return anything. Is it infallible?

kms-connector/crates/utils/src/types/event.rs
```diff
@@ -423,6 +444,40 @@ pub fn from_crsgen_row(row: &PgRow) -> anyhow::Result<ProtocolEvent> {
...
+            .try_get::<Vec<u8>, _>("tx_hash")
+            .ok()
+            .and_then(|h| FixedBytes::try_from(h.as_slice()).ok()),
+        already_sent: row.try_get::<bool, _>("already_sent")?,
+        error_counter: 0,
+        created_at: row.try_get::<DateTime<Utc>, _>("created_at")?,
+        otlp_context: bc2wrap::deserialize_slice(&row.try_get::<Vec<u8>, _>("otlp_context")?)?,
+    })
+}
+
```
dvdplm, line 475: Maybe we can DRY these two methods a bit? They look very similar.

kms-connector/crates/utils/src/types/event.rs
```diff
@@ -504,6 +559,36 @@ async fn update_crsgen_status(
...
+    status: OperationStatus,
+    already_sent: bool,
+) {
+    let query = sqlx::query!(
+        "UPDATE abort_crsgen_requests SET status = $1, already_sent = $2 WHERE crs_id = $3",
+        status as OperationStatus,
+        already_sent,
+        id.as_le_slice(),
+    );
+    execute_update_event_query(db, query).await;
```
dvdplm: Same here: very similar, can be DRYer?

dvdplm (dismissed):
LGTM, seems idiomatic. Left some nits, all optional or can happen in follow-up PRs.

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: Technically no, because the flow is:
- kms-connectors forwards keygen/crsgen to the cores
- cores start their work together
- kms-connectors poll for keygen/crsgen result
- gets `Unavailable` GRPC code while cores are computing
- kms-connector then forwards abort keygen/crsgen to cores
- next time the kms-connectors will poll for keygen/crsgen result, they would get an `Aborted` GRPC code

So it's the core that warns the connector when something is aborted.

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
eudelins-zama: See flow [here](https://github.com/zama-ai/fhevm/pull/2947#discussion_r3512996654), I like it this way personally

kms-connector/crates/kms-worker/tests/integration_tests.rs
eudelins-zama: I thought about it too, I don't like it either. But it would have required to wrap result of `wait_for_response_in_db` into an `Option` just for this. Felt wrong to me, and simpler to handle it at call site.

kms-connector/crates/kms-worker/tests/integration_tests.rs
eudelins-zama: You're absolutely right!

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: More consistent with other `mark_as_...` fn

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: No, it will emit a `warn!` in such case.

Logic behind `mark_as_...` not returning is that it's used by kms-worker when it finishes to process a request.
Either:
- It succeeded, we got the response, and in such case it would be weird to consider the whole request as a failure just because we did not managed to  update the status of the request
- If failure/aborted, then we already wanted to mark the request as failed, so returning won't help us doing something. It will just be marked as failed/aborted at next retry

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: No, these are DB queries checked at compile time, we can't do much smaller if we want to keep the compile time check

kms-connector/crates/kms-worker/tests/integration_tests.rs
eudelins-zama: Could help understanding the code execution? Not super useful I agree, but not hurting too much as well

kms-connector/crates/kms-worker/tests/integration_tests.rs
eudelins-zama: Could help understanding the code execution? Not super useful I agree, but not hurting too much as well

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
jot2re: Me too tbh. It was a deliberate choice in the design. It is not the only endpoint that only returns a code 200 upon success. For example for context construction it is the same

kms-connector/crates/kms-worker/tests/integration_tests.rs
eudelins-zama: Fixed 75283837a79b9c174e4a11e54fd0e80a13cf71f4

Went for the previous approach of making the mock return an error instead of early return + `unreachable!` to minimize LoC

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: Fixed 75283837a79b9c174e4a11e54fd0e80a13cf71f4

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
dvdplm: I see, thank you for explaining. :)

dvdplm: Consider adding a comment with this info somewhere?

kms-connector/crates/kms-worker/tests/integration_tests.rs
dvdplm: If it stays (eew) then I think it should say something other than "Ok!". Imagine you're running tests concurrently with logging and see a bunch of "Ok!"s flash by...

kms-connector/crates/utils/src/types/event.rs
dvdplm: well... yes, but maybe they should drop the `_as` too? ;)

JK, fair enough.

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: Already [there](https://github.com/zama-ai/fhevm/blob/75283837a79b9c174e4a11e54fd0e80a13cf71f4/kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs#L130-L131) :wink:

kms-connector/crates/utils/src/types/event.rs
dvdplm: On the logic level I agree with you but it's still a DB operation that can (likely) fail for other reasons not related to our business logic. To me it should return an error and then the caller can decide if it wants to act on the error or not.

dvdplm (dismissed):

kms-connector/crates/kms-worker/tests/integration_tests.rs
eudelins-zama: Fair enough: 69188aceb01a6b40345df3a990cb3e0ad510c023 :slightly_smiling_face:

kms-connector/crates/kms-worker/tests/integration_tests.rs
eudelins-zama: Fair enough: 69188aceb01a6b40345df3a990cb3e0ad510c023 :slightly_smiling_face:

kms-connector/crates/utils/src/types/event.rs
eudelins-zama: Fix attempt 6359e4b3a80a8203f3d1bd8d8d354048a81187df

dvdplm (approved):

eudelins-zama:
@mergifyio queue

eudelins-zama:
@mergifyio refresh

PR #2948, opened 2026-07-01
fix(common): restore missing rollup an esbuild platform binaries

PanGan21:
### Summary

Lockfile has been regenerated, which triggered npm to dropp the non-darwin rollup/esbuild binaries. `npm ci` on Linux CI then couldn't install @rollup/rollup-linux-x64-gnu, so js-sdk-tests crashed at startup.

Regenerated on Linux from the last known-good lockfile, adding only the intended LayerZero deps. Restores all platform binaries with no forward version drift.


PanGan21:
@claude /pr-review

geoxel (approved):

PanGan21:
@mergifyio queue

PR #2959, opened 2026-07-02
refactor(host-contracts): route bridge verification through verifyContract wrapper

obatirou:
Routes `task:verifyBridge` through the `verifyContract` wrapper instead of calling `verify:verify` directly, so bridge verification gets the same "already verified" tolerance as the other host contracts. Adds an optional `constructorArguments` param to the wrapper (impl needs `[lzEndpoint]`, proxy none).


jatZama (approved):
LGTM

PR #2974, opened 2026-07-03
feat(coprocessor): supporting-crate rust changes for blue-green upgrade

goshawk-3:
## What part of RFC 021 this PR covers

Maps to [RFC 021 — Consensus-Preserving Blue-Green Upgrade](github.com/zama-ai/tech-spec/blob/georgi/blue-green-upgrade/rfcs/021-blue-green-upgrade.md).

**Target: prepare the stage for the first BCS role.** This PR lands the *shared foundation* every coprocessor service needs to participate in a blue/green upgrade — the versioning machinery, GCS schema routing, retired-st
ack guard, cutover serialization, the DB schema, and the producer-side plumbing the readiness/consensus checks consume. The two orchestrator binaries that *drive* the flow (`upgrade-controller`, `consensus-detector`) are
 deliberately **not** here — they land in the follow-up PR. In this state the existing (blue) stack runs unchanged, and a newer image auto-detects GCS mode.

### Versioning & GCS-mode detection — RFC §Versioning
- `STACK_VERSION` single-sourced constant (`stack_version!()` macro in `fhevm-engine-common`) + `versioning` singleton table (migration).
- `resolve_gcs_mode` at startup (newer-than-live ⇒ green), wired into every service.
- `run_stack_version_listener` + `StackMode` / `reconcile_stack_mode` — services re-evaluate blue/green mode on `event_stack_version_upgraded`.
- `--stack-version` CLI flag (`handle_stack_version_flag`) across all services, surfaced in the chart via the `app.kubernetes.io/version` pod label.

### GCS schema routing — RFC §Database: GCS schema as a Postgres namespace
- `GCS_SCHEMA` / `GCS_SCHEMA_QUOTED` / `GCS_SEARCH_PATH` constants (all derived from `stack_version!()`).
- `connect_pool_with_gcs_mode` pins `search_path = "gcs-<version>",public` so unqualified worker SQL routes to the GCS copy in green mode and `public` in blue — no per-statement branching. *(The `CREATE`/`DROP` of the sc
hema itself lives in `upgrade-controller` → follow-up.)*

### Retired-stack guard & cutover serialization — RFC §Retired-stack transaction guard, §Why an advisory lock
- `assert_not_retired` / `is_retired` and `begin_guarded_pool` / `begin_guarded_conn` — a retired (older-than-live) binary aborts its write txs.
- `cutover_gate` shared-lock cooperation, applied across `tfhe-worker`, `sns-worker`, `zkproof-worker`, `transaction-sender`, `host-listener`, `gw-listener`, so BCS writers serialize against the future exclusive cutover
transaction.

### DB migrations (schema substrate)
- `create_versioning_table`, `create_upgrade_state` (+ `add_dry_run_started_state`, `add_gw_dry_run_started_to_upgrade_state`) — versioning row and the FSM table the controller will drive.
- `create_state_hash_table` (+ `add_s3_uploaded_at_to_state_hash`), `create_input_handles_table`, `add_is_input_to_ciphertexts`, `add_block_number_to_verify_proofs`, `add_block_event_counts`.

### Readiness / consensus producer-side plumbing — RFC §Per-block tallies, §Readiness check, §Consensus mechanism
- `host-listener` records per-block `fhe_event_count` / `allow_event_count` on `host_chain_blocks_valid` at decode time (the counters the readiness predicate short-circuits on).
- `input_handles` population + `ciphertexts.is_input` (in `tfhe-worker` / `zkproof-worker`), and the `state_hash` / `input_handles` tables registered among the GCS-duplicated tables.
- These feed the readiness check and the per-block `stateCommitment` worker — whose *consumers* (`upgrade-controller` readiness loop, `consensus-detector` S3 unanimity) arrive in the follow-up.

### Required contract dependency
`host-listener` decodes the new `ProtocolConfig.CoprocessorUpgradeProposed` event (RFC §On-chain Surface), so the minimal `ProtocolConfig.sol` / `IProtocolConfig.sol` / `Structs.sol` changes + regenerated Rust bindings a
re included as the compile dependency closure.

### Explicitly deferred to the follow-up PR
`upgrade-controller` (FSM, readiness loop, `execute_cutover`, GCS schema creation) and `consenser + cross-operator S3 unanimity), plus the CI workflows and `test-suite` e2e wiring for those s
ervices.


coprocessor/fhevm-engine/host-listener/src/bin/consumer.rs
```diff
@@ -38,6 +38,9 @@ struct Args {
     )]
     kms_generation_address: String,
 
+    #[command(flatten)]
+    protocol_config: host_listener::protocol_config::ProtocolConfigArgs,
```
PanGan21, line 42: This means we need helm chart bump

coprocessor/fhevm-engine/fhevm-engine-common/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fhevm-engine-common"
-version = "0.6.1"
+version = "0.14.0"
```
goshawk-3: This is a leftover from the first approach. To be reverted

coprocessor/fhevm-engine/gw-listener/src/lib.rs
```diff
@@ -41,6 +41,12 @@ pub struct ConfigSettings {
     /// a consensus mismatch. If false, drift is still detected and logged,
     /// but no signal is created.
     pub drift_auto_revert_enabled: bool,
+    /// When true, the gw-listener runs in GCS mode: its DB connections are
+    /// pinned to `search_path = "gcs-<version>",public` so writes land in the
+    /// versioned GCS schema. It is not paused before activation — it processes
+    /// events from startup; only the cutover transition retires it. Used by the
+    /// GCS stack during the blue/green upgrade flow.
+    pub gcs_mode: bool,
```
PanGan21, line 49: Should this be added in the helm templates?

coprocessor/fhevm-engine/sns-worker/src/bin/utils/daemon_cli.rs
```diff
@@ -138,6 +138,10 @@ pub struct Args {
     #[arg(short, long)]
     pub private_key: Option<String>,
 
+    /// Print the compiled-in coprocessor stack version and exit.
+    #[arg(long)]
+    pub stack_version: bool,
```
PanGan21, line 143: Same for this for the templates

coprocessor/fhevm-engine/tfhe-worker/src/daemon_cli.rs
```diff
@@ -125,13 +125,18 @@ pub struct Args {
     #[arg(long, default_value = "0.2:5.0:0.05", value_parser = clap::value_parser!(MetricsConfig))]
     pub metric_fhe_batch_latency: MetricsConfig,
 
+    /// Print the compiled-in coprocessor stack version and exit.
+    #[arg(long)]
+    pub stack_version: bool,
```
PanGan21, line 130: To be added to the templates as well

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
```diff
@@ -144,6 +147,10 @@ struct Conf {
 
     #[arg(long, default_value_t = 10, value_parser = clap::value_parser!(u64).range(1..))]
     pub gauge_update_interval_secs: u64,
+
+    /// Print the compiled-in coprocessor stack version and exit.
+    #[arg(long)]
+    pub stack_version: bool,
```
PanGan21, line 153: To be added to the templates as well

coprocessor/fhevm-engine/zkproof-worker/src/bin/zkproof_worker.rs
```diff
@@ -72,9 +72,14 @@ pub struct Args {
     /// Prometheus metrics: "coprocessor_zkverify_op_latency_seconds",
     #[arg(long, default_value = "0.01:2.0:0.01", value_parser = clap::value_parser!(MetricsConfig))]
     pub metric_zkverify_op_latency: MetricsConfig,
+
+    /// Print the compiled-in coprocessor stack version and exit.
+    #[arg(long)]
+    pub stack_version: bool,
```
PanGan21, line 78: To be added to the templates

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
```diff
@@ -0,0 +1,381 @@
...
+/// Decide whether this binary should run in GCS (green) mode by comparing its
+/// compiled-in [`STACK_VERSION`] against the live `versioning.stack_version`
+/// row.
+///
+/// Opens a short-lived connection with the default `public` search_path, so it
+/// works before the service's main pool — whose search_path may be pinned to
+/// `gcs,public` — is built. If the `versioning` row is missing the service
+/// defaults to non-GCS (blue) mode.
+pub async fn resolve_gcs_mode(database_url: &str) -> anyhow::Result<bool> {
+    let mut conn = PgConnection::connect(database_url).await?;
```
PanGan21: This is flagged that it bypasses IAM db auth.
It probably must go through `resolve_runtime_database_url`  to be fixed

coprocessor/fhevm-engine/db-migration/migrations/20260521140000_add_dry_run_started_state.sql
```diff
@@ -0,0 +1,12 @@
+-- Extend upgrade_state.state with `DryRunStarted` — set by upgrade-controller
+-- (GCS) once the pre-snapshot completeness check passes and the unpause notify
+-- has been emitted.
+ALTER TABLE upgrade_state DROP CONSTRAINT upgrade_state_state_check;
```
PanGan21: Let's change `DROP CONSTRAINT` to `DROP CONSTRAINT IF EXISTS` so the migration is idempotent and rerunnable

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
```diff
@@ -0,0 +1,381 @@
...
+    let row: Option<(String,)> =
+        sqlx::query_as("SELECT stack_version FROM versioning WHERE singleton = TRUE")
+            .fetch_optional(&mut conn)
+            .await?;
```
PanGan21: `.await?` propagates `sqlx::Error(UndefinedTable)` when the table is missing so this might fail the workers if they start before the migration is applied.
We can catch `sqlx::Error::Database` with `SQLSTATE 42P01` and log and return `Ok(false)`

coprocessor/fhevm-engine/gw-listener/src/lib.rs
goshawk-3: This is not a CLI param

coprocessor/fhevm-engine/sns-worker/src/bin/utils/daemon_cli.rs
goshawk-3: stack-version should be added, yes

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
goshawk-3: Is there a scenario where we will run the worker before the migrations are applied?

PanGan21: According to the helm charts they start simultaneously so potentially yes. 
But if it happens it seems self healing so I am resolving this issue.

antoniupop:
Some findings from an agent:
 1. --stack-version probe CrashLoop — host-listener/src/bin/consumer.rs:129, poller.rs:160. Neither binary implements the flag their new Helm startupProbes exec. Enabling stackVersionProbe.enabled kills both pods permanently (clap exits 2, probe never passes).
  2. Bare STACK_VERSION bump silently stalls the fleet — fhevm-engine-common/src/versioning.rs:207. Deploying any binary strictly newer than the versioning row (e.g. a hotfix 0.14.1) without a proposed upgrade makes every service self-classify as green: workers and the transaction-sender park forever, and the listeners pin search_path to a nonexistent gcs-0.14.1 schema that Postgres silently skips — so their writes land in public. No error is raised anywhere.
  3. Missing versioning table = fleet-wide CrashLoop on first deploy — versioning.rs:191. A missing row defaults to blue, but a missing table is a fatal startup error in every service, and the chart ships no migration ordering (hook annotations are commented-out examples, no initContainers). Self-healing once the migration Job finishes, but a new non-retried startup dependency the previous release didn't have.
  4. Fatal uploader failure exits 0 — sns-worker/src/lib.rs:835. The uploader task now swallows its error and the join arm discards the result, so a persistent S3 failure shuts the worker down cleanly with exit code 0 — skipping the fatal log, telemetry::flush(), and exit(1) that alerting keys on.

PanGan21 (approved):

antoniupop (approved):

PR #2977, opened 2026-07-03
refactor(library-solidity): generate lib/bridge OApp abstracts via codegen

poppyseedDev:
## What

Makes the confidential-bridge **OApp abstracts under `lib/bridge/`** generated from `codegen/src`, so the whole `lib/` directory stays reproducible (`rm -rf lib/bridge && npm run codegen`) — they were hand-written before.

- New `codegen/src/templateBridge.ts` + templates for `ConfidentialOAppCore`, `ConfidentialOAppSender`, `ConfidentialOAppReceiver`, `ConfidentialOApp`, and `IDstApp`.
- The Sender's **per-encrypted-type single-handle overloads are now loop-generated** (one place) instead of hand-repeated 8×.
- Wired in `main.ts`, gated behind a new `lib.bridge` flag (`config.ts` + `codegen.config.json`) so it's **library-only** — host-contracts leaves it off and never emits the OApp abstracts.

## Notes

- **Verified reproducible & near-identical:** after wiping and regenerating, the four static abstracts (`Core`/`Receiver`/`ConfidentialOApp`/`IDstApp`) come out **byte-identical** to the current contracts; the only content deltas are two Sender doc typos the template corrects (`"32 handes"` → `"32 handles"`, and a stray double-period on the `eaddress` overload) plus the `ConfidentialMultiChainToken.quoteSend` payload natspec (`(address, bytes32)` → `address`).
- `hardhat compile` passes.
- Stacked on the method-name rename (`89d101c8b`); no naming changes here.

### DIFF
--- a/library-solidity/codegen/src/templateBridge.ts
+++ b/library-solidity/codegen/src/templateBridge.ts
@@ -1,0 +1,78 @@
+import { readFileSync } from 'fs';
+
+import { resolveTemplatePath } from './paths';
+import { removeTemplateComments } from './utils';
+
+/**
+ * Confidential-bridge OApp abstracts (`lib/bridge/`).
+ *
+ * All of these are generated from templates under `codegen/src/templates/`, so `lib/bridge/` is
+ * fully reproducible from `codegen/src` (`rm -rf lib/bridge && npm run codegen`). The static
+ * abstracts are emitted verbatim from their templates; the Sender's per-encrypted-type
+ * single-handle overloads are generated in a loop (see {generateConfidentialOAppSenderLib}) so the
+ * repetition lives in one place.
+ */
+
+/** Encrypted handle types that get a type-safe single-handle send overload, in declaration order. */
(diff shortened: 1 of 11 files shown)


library-solidity/codegen/src/templateBridge.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * The Sender abstract: a `private` `bytes32` core, one type-safe overload per encrypted type
+ * (generated below), the multi-handle list send, and the two quote helpers. The per-type
+ * overloads all delegate to the `bytes32` core via `FHE.toBytes32`.
+ */
+export function generateConfidentialOAppSenderLib(): string {
+  const typedOverloads = BRIDGE_HANDLE_TYPES.map(
+    (t) => `
```
jatZama, line 46: Can you rework the natspec here, to instead of refer to private core method, it would copy full natspec of the private function (with few needed modifications obviously)

jatZama (approved):

PR #2978, opened 2026-07-03
chore(kms-connector): increase default gauge update interval

eudelins-zama:
Increase default `gauge_update_interval` from `10s` to `30s` as it seems to be more aligned with default Prometheus scrapping time (avoid wasted work, especially now that we are running many queries in the gauge update routine).


dvdplm (approved):

eudelins-zama:
@mergifyio queue

PR #2999, opened 2026-07-04
fix(coprocessor): block context re-randomisation part 1 fixes

antoniupop:


coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -551,6 +568,67 @@ pub async fn update_finalized_blocks_aux<GetBlockHash, GetBlockHashFuture>(
...
+    for block_number in blocks_number {
+        match get_block_hash_by_number(block_number as u64).await {
+            Ok(block_hash) => canonical.push((block_number, block_hash)),
+            Err(err) => {
+                error!(
+                    block_number,
+                    ?err,
+                    "Failed to fetch block for finalization"
+                );
+                continue;
```
PanGan21: Is it correct to continue here instead of break? 
With an RPC failure we skip just this block and keep the later ones, leaving a gap. With break, correct stamping is just postponed to the next loop if I understand the rationale correctly

coprocessor/fhevm-engine/db-migration/migrations/20260704130000_blocks_valid_prune_support.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE INDEX IF NOT EXISTS idx_bridge_handle_events_block_hash
+    ON bridge_handle_events (block_hash);
+CREATE INDEX IF NOT EXISTS idx_handle_bridged_events_block_hash
+    ON handle_bridged_events (block_hash);
+CREATE INDEX IF NOT EXISTS idx_fallback_granted_events_block_hash
+    ON fallback_granted_events (block_hash);
+-- The pruner's candidate scan walks finalized rows below the retention
+-- horizon in block order.
+CREATE INDEX IF NOT EXISTS idx_host_chain_blocks_valid_chain_number
+    ON host_chain_blocks_valid (chain_id, block_number);
```
PanGan21: Is it correct to create the index from a migration? This will lock the `host_chain_blocks_valid` table, so the host-listener can only write again after the index has finished building for the whole table, which can take long since this table grows with every block.
Maybe `initialize_db.sh` is a better place where we can use `CONCURRENTLY`?

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
antoniupop: Good catch - you're right. fixing

coprocessor/fhevm-engine/db-migration/migrations/20260704130000_blocks_valid_prune_support.sql
antoniupop: Good point - but actually this is a duplicate index, the same is already created in 20260513120000_add_block_number_index_on_host_chain_blocks_valid.sql. 
Removing.

coprocessor/fhevm-engine/.sqlx/query-36b35d2fe0e18ccace96a4e293b10400719d542f6f3cf2cecdca34cf36172aef.json
```diff
@@ -0,0 +1,16 @@
+{
```
rudy-6-4, line 1: why here and not in the shared .sqlx ?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -321,13 +343,19 @@ impl Database {
                 .unwrap()
                 .into(),
             )));
+        let branch_activation_block =
+            std::env::var("FHEVM_BRANCH_ACTIVATION_BLOCK")
+                .ok()
+                .and_then(|v| v.parse().ok())
+                .unwrap_or(0);
```
rudy-6-4: bad formatted env var default to 0 is safe ?

coprocessor/fhevm-engine/.sqlx/query-36b35d2fe0e18ccace96a4e293b10400719d542f6f3cf2cecdca34cf36172aef.json
antoniupop: I can move these, but there's still lots more that have gone in this local cache. Would be best to move those in a separate PR.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: Good catch - that's not fine for consensus, would be fine in mono-coprocessor only.

0xalexbel (dismissed):
LGTM!

rudy-6-4 (dismissed):
I think delegate_user_decrypt is no longer used since it's no longer transmitted to gateway. Is there any use case that would need to have it guarded in the pruning logic like other table ? (currently it's not listed in the pruning routine)

PanGan21 (dismissed):

antoniupop:
> I think delegate_user_decrypt is no longer used since it's no longer transmitted to gateway. Is there any use case that would need to have it guarded in the pruning logic like other table ? (currently it's not listed in the pruning routine)

I can't see any reason to guard it - but better cleanup stale queue after the freeze.

PanGan21 (dismissed):

rudy-6-4 (dismissed):

0xalexbel (approved):

PanGan21 (approved):

antoniupop:
@mergifyio queue

PR #3001, opened 2026-07-04
feat(coprocessor): re-randomise input ciphertexts before expand

antoniupop:
Reintroduce the compact-list re-randomisation reverted in 537cd123c (#2759, revert of #2383): the zkproof-worker now re-randomises the verified ProvenCompactCiphertextList — seeded from the input blob hash — and expands it in a single step, instead of expanding first and re-randomising each expanded ciphertext before compression. The per-ciphertext re-randomisation metadata is no longer set at input verification, so the now-unused
SupportedFheCiphertexts::add_re_randomization_metadata is removed (the scheduler's operation-input re-randomisation helpers are unaffected).

BREAKING CHANGE: for the same input blob, stored ciphertexts are not bitwise-equal to those produced by the previous rerand-after-expand scheme, so a mixed fleet would diverge. Ship only via a synchronised (blue-green) coprocessor upgrade, never a rolling one.

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/types.rs
@@ -725,67 +725,6 @@ impl SupportedFheCiphertexts {
         }
     }
 
-    pub fn add_re_randomization_metadata(&mut self, hash_data: &[u8]) {
-        match self {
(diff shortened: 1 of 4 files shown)


goshawk-3 (approved):
LGTM

antoniupop:
@mergifyio queue

PR #3002, opened 2026-07-05
feat(coprocessor): encrypted-value + mmr acl rewrite of the solana access-control model

Eikix:
## What

This PR rewrites Solana encrypted-value ACL handling around one stable ZamaHost-owned `EncryptedValue` lineage account per logical value. Superseded-handle access is recorded as an on-account SHA-256 Merkle Mountain Range (MMR), replacing the keyed-nonce `AclRecord` model.

This is an MVP targeting `feature/solana`; breaking changes are intentional and no compatibility layer is retained.

## Authorization and finality

- Raw `create_encrypted_value` and `update_encrypted_value` lifecycle instructions are disabled. Durable lineage creation and supersession come from `fhe_eval` output provenance.
- The finalized-account fetcher retries a witness behind the queued event slot.
- Public and historical decrypt releases require an exact MMR proof verified against finalized on-chain peaks; proofless releases fail closed.
- The Solana decrypt retry cap and secp256k1 recovery validation match the intended worker/EVM behavior.

## Validation

- Repository CI is green, including Solana program tests, KMS/relayer tests, coprocessor CPU and GPU tests, and the full Solana decrypt vertical E2E.
- The vertical E2E exercises proof-backed public decrypt, user decrypt, input flow, and consumption.

## Scope boundaries

The branch intentionally does not add defensive compatibility code. Longer-term architecture and productization follow-ups remain tracked separately; this PR is limited to completing the MMR-based ACL MVP and its end-to-end authorization path.


Eikix:
This PR is too large to review effectively as a flat diff. Before doing a line-by-line review, the MVP design should be reviewed against the following architecture and behavioral contract.

This comment records the review criteria that were agreed on for the Solana ACL/MMR rewrite. The goal is not to require stacked PRs for the MVP. The goal is to make the design, intent, and high-risk boundaries explicit enough that reviewers can judge whether this is a coherent MVP and where close review is required.

## Intended architecture

The PR replaces the previous Solana `AclRecord` model with one `EncryptedValue` account per logical encrypted value.

An `EncryptedValue` represents the current encrypted handle for one app-owned value, plus the facts needed to prove access to superseded handles. When the current handle is replaced, the previous handle is sealed into the MMR for the subjects that were allowed at the time of supersession. KMS must only release decryptions after verifying proofs against finalized on-chain state.

The high-level trust boundary is:

- Solana programs enforce account identity, authority, ACL mutation, current-handle update, and MMR/public facts.
- Relayer/proof-builder/coprocessor ingestion may be useful for liveness and performance, but they are not trusted for authorization.
- KMS verifies the finalized Solana facts before releasing user decryptions or public decryptions.
- Off-chain bugs may cause retry, failure, or wasted work, but must not permit false decrypt authorization, wrong decrypt release, or token-state corruption.

## Core concepts that should be documented in the PR

The PR should include one short canonical design note, either in the PR description first or as markdown in the repo, because these concepts are not obvious from the diff alone:

1. `EncryptedValue` identity is an internal preimage computation, not an authority model.
   The irreducible components are:
   - `acl_domain_key`: app namespace, for confidential-token this is the mint.
   - `app_account`: app state object that owns the field, for balances this is the token account.
   - `encrypted_value_label`: field label, for example balance, total supply, transfer amount.
   These fields prevent collisions across domains, accounts, and fields. They should not be described as independent authorities.

2. `compute_signer` is separate from `EncryptedValue` identity.
   For confidential-token, the compute signer is a mint-scoped PDA used by the token program for FHE computation. It should be derived and checked from the mint, not treated as an arbitrary identity field.

3. The ACL model for MVP is a single allowed-subject set.
   There should not be separate durable role bits such as use/grant/public-decrypt. If a subject is allowed, it can compute/use the value, grant another subject, request user decrypt, and make an exact handle public. If a subject is not allowed, it cannot do those actions.

4. `EncryptedValue` creation must include at least one allowed subject.
   For confidential-token balances, creation should include the user/owner and the token compute authority.

5. Only the app-owned account authority can update the latest handle.
   Being an allowed subject is not sufficient to replace `current_handle`. For token balances, the token program signs through the expected token-account/app-authority path. Alice may be allowed to decrypt or compute over her balance, but that does not let Alice directly replace the balance handle unless the token app authority path authorizes it.

6. Historical authorization is handle-scoped and permanent.
   If Alice is allowed while handle `H1` is current, and the value later moves to `H2`, Alice remains authorized for `H1`. Alice may user-decrypt `H1`, allow another subject on `H1`, or make `H1` public if the proof verifies against finalized state.

7. Public decrypt is exact-handle scoped.
   If `H1` is marked public, only `H1` is public. A later `H2` is not public unless a separate public fact is created for `H2`.

8. Durable subject removal affects only current and future handles.
   Removing a subject does not revoke historical access that was already sealed into the MMR and does not remove public facts. If a subject is removed while `H2` is current, that subject loses access to `H2` immediately and must not receive a historical leaf for `H2` when `H2` is later superseded.

9. Latest-handle updates must be stale-state guarded.
   An update should fail if the current handle or the current durable subject set has changed since the update was built. Passing and checking `expected_subjects` is acceptable for the MVP and is semantically equivalent to checking a subject-set revision/hash.

10. Durable subjects should be capped for the MVP.
   The agreed MVP cap is 8 permanent subjects per `EncryptedValue`. The MMR peak count may also be capped by the chosen maximum MMR size, but the cap must be constructed so normal operation cannot realistically hit it. If the theoretical cap is reached, failure must be explicit and must not corrupt history.

11. Pause and deny-list behavior should follow EVM parity where applicable.
   Pause gates ACL mutations and update/eval output paths. The deny-list blocks a denied subject from acting as caller, but does not erase existing permissions, historical leaves, or public facts. ACL mutations such as allow, public-decrypt marking, transient allow, and remove-subject should reject a denied caller. Latest-handle update should also be deny-list gated on the acting app account authority.

12. Delegated user decrypt is out of MVP scope unless it is trivial and isolated.
   It should not be half-wired into the core ACL path.

## Confidential-token mapping

The confidential-token program should validate its app-specific account graph before calling `zama-host`, and `zama-host` should independently validate generic `EncryptedValue` invariants.

For confidential-token:

- `acl_domain_key` should be the mint.
- `app_account` should be the token account or other app state account that owns the encrypted field.
- `encrypted_value_label` should distinguish fields such as balance, total supply, wrap amount, burn amount, and transfer amount.
- `compute_signer` should be the mint-scoped FHE compute PDA.
- A balance `EncryptedValue` should be tied to the expected mint, token account, label, compute signer, and app authority.

This duplication is intentional. The token program prevents a valid host operation from being applied to the wrong token object. `zama-host` prevents callers from bypassing the generic ACL and MMR rules.

Token receive semantics should match normal token semantics: the recipient does not need to sign to receive tokens. The token program may update the recipient balance through the app authority path, and the resulting balance handle must grant the recipient access.

## Input and compute policy

External user inputs must be coprocessor-attested and bound to the user and compute subject.

Already encrypted values may be reused only when the compute subject is allowed on those values. Trivial/rand encrypted values may be valid FHE sources if there is a concrete app use case, but token tests should primarily exercise real user inputs because that is where spoofing and binding mistakes matter.

Transient authorization should be allowed for EVM parity and bridge/callback-like flows where an app receives a transient allow and then creates durable authorization or public facts. However, `remove_subject` should require durable current authorization, not merely transient authorization.

## Proof, relayer, ingestion, and KMS boundary

The proof builder and relayer are untrusted. They can construct or serve proofs, but KMS must verify them against finalized Solana state before releasing plaintext.

The MVP may keep the proof builder as a separate relayer-colocated service if the boundary is clean and replaceable. If `/user-decrypt` and `/public-decrypt` do not call the proof builder in-process in this PR, the limitation should be stated explicitly. The important requirement is that later moving the proof builder in-process must not change proof semantics.

Both event/ingestion paths should preserve the same semantic facts:

- Yellowstone gRPC reconstruction path.
- Non-Yellowstone `emit_cpi`/event path.

Both paths should expose the same current subjects, supersession leaves, public leaves, and MMR peaks. If one path is intentionally out of scope, the PR must say so explicitly with justification.

Ciphertext materiality should remain tied to gateway `CiphertextCommits`, not to Solana `zama-host`. Removing Solana materiality commitments is acceptable only if KMS/gateway still checks materiality before release.

Eager compute on a minority fork is acceptable as wasted work, provided decrypt release re-reads finalized `EncryptedValue` state and validates the relevant handle and peaks before release.

## Review focus areas

The PR should be reviewed in chunks around the risk boundaries, not by repository folder alone:

1. `EncryptedValue` state model and MMR append semantics.
   Check identity preimage, account sizing, subject cap, MMR cap, historical leaf contents, public leaf contents, and exact-handle behavior.

2. `zama-host` instruction authorization.
   Check create, allow, remove, update, public-decrypt marking, stale-state guards, pause/deny-list gates, and app-account authority checks.

3. Confidential-token CPI integration.
   Check mint/token-account/compute-signer derivations, app authority, recipient-no-sign receive flow, balance poisoning resistance, and output-handle ownership.

4. Coprocessor and ingestion.
   Check that decoded instructions and events produce the same semantic facts for Yellowstone reconstruction and emit-CPI paths.

5. Relayer/proof builder and KMS verification.
   Check finalized-state verification, proof preimages, MMR peak verification, exact-handle authorization, and materiality checks.

6. SDK/user-decrypt/public-decrypt path.
   Check subject authentication with the Solana key, private delivery to the SDK transport key, no plaintext exposure to relayer, and historical/public exact-handle behavior.

7. Tests.
   The MVP should have adversarial tests for at least:
   - stale update after current handle changed;
   - stale update after subject set changed;
   - subject removal before supersession;
   - removed subject cannot access current handle;
   - removed subject keeps historical access already sealed before removal;
   - public decrypt is exact-handle scoped;
   - denied caller cannot mutate ACL;
   - paused state blocks ACL mutation and update/eval output path;
   - cap of 8 durable subjects;
   - recipient balance update without recipient signature;
   - wrong mint/token account/compute signer/encrypted-value PDA rejected;
   - relayer/proof-builder cannot cause KMS release without finalized valid proof;
   - non-Yellowstone event path and Yellowstone reconstruction path produce equivalent facts.

## Approval bar

This PR should not be judged only on whether the happy path works. It should be judged on whether the above invariants are implemented directly, documented clearly, and covered by focused tests.

The most important properties are:

- no false decrypt authorization;
- no wrong public decrypt release;
- no balance/current-handle poisoning across token accounts or mints;
- no trust placed in relayer/proof-builder/coprocessor ingestion for authorization;
- no ambiguity between identity fields and authority fields;
- no role-bit complexity in the MVP durable ACL model;
- no hidden revocation semantics for historical handles or public facts.

Eikix (commented):
Consolidated review, scoped with the branch treated as a PoC with no backward-compatibility requirement.

Backward-compatibility-only migration concerns are intentionally not included. The items below are still considered correctness/security issues for a PoC because they can release decrypt authorization too early, release authorization without proving the MMR leaf, or accept verifier/handle states that the stated parity model rejects.

1. `coprocessor/fhevm-engine/host-listener/src/solana_finalized_account_fetcher.rs`

The finalized-account fetcher completes a job as soon as `getMultipleAccounts(finalized)` returns `Some(account)`, but the returned `context.slot` / stored `observed_slot` is not checked against `job.block_number` before the job is completed and downstream decrypt state is inserted. For an existing `EncryptedValue`, finalized RPC can return the pre-event finalized account state while the queued event's slot is not finalized yet. A `make_handle_public` event can therefore release a handle before the public MMR leaf is finalized, and an `allow_subjects` event can be completed while the finalized account still lacks the subject.

What needs to be changed: a fetched witness whose `observed_slot < job.block_number` should be retried instead of completed, or the RPC read should be issued with an equivalent `minContextSlot` requirement.

2. `coprocessor/fhevm-engine/host-listener/src/solana_finalized_account_fetcher.rs`

The finalized fetcher treats `handle_made_public` as finalized when `current_handle == handle`, and treats a superseded handle as finalized when `leaf_count > 0`. Those checks do not prove the specific authorization being released. Public decrypt authorization is represented by a public MMR leaf, and historical authorization is represented by a historical MMR leaf for the specific handle/subject. A bad decoded queue row or ingestion bug can therefore release public/historical decrypt authorization for a handle that the finalized account does not prove.

What needs to be changed: public and historical releases should require proof data and verify the corresponding `authorize_public` / `authorize_historical` condition against finalized account state. Until that is wired, releases that cannot be proven from finalized state should be refused rather than approximated from `current_handle` or nonzero `leaf_count`.

3. `solana/programs/zama-host/src/eip712.rs`

The EIP-712 recovery path accepts `v = 29/30` because `27` is subtracted and recovery ids up to `3` are allowed. The stated parity target is EVM/OpenZeppelin ECDSA, where only `v = 27/28` should be accepted. This means a Solana certificate shape can pass threshold verification while the corresponding EVM verifier would reject it.

What needs to be changed: `signature[64]` should be rejected unless it is exactly `27` or `28`, or the post-subtraction bound should reject recovery ids greater than `1`.

4. `solana/programs/zama-host/src/instructions/encrypted_value.rs`

The direct `create_encrypted_value` / `update_encrypted_value` path stores caller-provided handles without validating chain id, handle version, or FHE type. Later durable input use validates current handles, but shared ACL/KMS authorization can still prove equality or MMR membership around the stored bytes. That allows a direct ACL lineage to be created or updated around a malformed, wrong-chain, or unsupported-type handle.

What needs to be changed: `assert_handle_for_chain(handle, host_config.chain_id)` should be applied when creating a lineage and when updating to `new_handle`; the stored current handle should also be valid before it is sealed into a public leaf.

PoC-scoped notes, not blocking this review:

- The deposit app drops `ctx.remaining_accounts` when calling `confidential_transfer`; it will not work with the grant deny-list enabled.
- The deposit app increments `deposit_count` after a transfer that can complete with zero value moved; this is misleading if the event/state is intended to mean a successful value deposit.
- DD-026 still describes Solana user-decrypt `extraData` as context-only, while the PR now uses the `0x03` MMR proof tail transport. This should be corrected before the document is treated as the wire-contract reference.

Eikix:
# Orchestrated review — EncryptedValue + MMR ACL rewrite

Reviewed at `e8fa43854` against the RFC-024 functional contract and the MVP design note, using a fan-out of independent audit/journey/forward-fit agents (each finding adversarially self-verified agai
nst the code; the prior consolidated review was quarantined from discovery and re-adjudicated separately). Scope-appropriate for a PoC branch — backward-compat is not a concern here; the gates below
are about wrong decrypt authorization, permanent unavailability, and DeFi-composability walls.

## TL;DR

- **The core rewrite is sound.** MMR math, Rust↔TS preimage parity, `sol_sha256` byte-identity, on-chain supersession/authority invariants, the KMS authorization boundary, current/public decrypt scop
ing, self-transfer/same-frame semantics, and reorg-safety all held up under adversarial verification.
- **One confirmed merge-blocker:** `create_encrypted_value`/`update_encrypted_value` accept a caller-chosen raw handle with no provenance binding, which lets a forged lineage authorize decryption of
someone else's handle. Confirmed three independent ways.
- **Two token-layer composability walls** (not in the ACL core, in the confidential-token transfer surface) that would block specific DeFi patterns: single-recipient-per-sender, and no existing-handl
e spend path.
- **North star:** the current super-app product ports without a rewrite; the walls above bite future bulk/router/mass-payout flows and are fixable at the token-lineage-keying layer.

---

## Merge-blocking

### 1. Handle-provenance hole → wrong plaintext release (gate: KMS authorization boundary)

`create_encrypted_value` / `update_encrypted_value` (`solana/programs/zama-host/src/instructions/encrypted_value.rs:37`, `:172`) store a caller-supplied raw `handle: [u8;32]` as `current_handle` with
 caller-chosen `subjects`, gated only by `app_account_authority == app_account` (both caller arguments) and a PDA derived from the caller's own identity triple. There is no coprocessor attestation an
d no check that the caller already holds that handle. The KMS current-decrypt path (`zama-solana-acl` `authorize_current` at `src/lib.rs:209`; `verify_current_user_decrypt` at `kms-connector/crates/k
ms-worker/src/core/solana_encrypted_value_acl.rs:105`) authorizes on `current_handle == handle` + subject membership + a **self-asserted** `allowedAclDomainKeys` list — with no binding of the handle'
s provenance to the lineage it is decrypted under.

**Exploit:** an attacker creates a lineage at their own PDA with `current_handle = <victim handle>` (handles are public identifiers) and `subjects = [attacker]`, then sends a current-path user-decryp
t for that handle. Every check passes and the KMS releases the victim's plaintext. Both residual escape hatches were traced and fail to block it: ciphertext material is keyed purely by handle (no lin
eage binding), and neither the coprocessor ingestion nor the finalized fetcher checks that a lineage *originated* the handle it asserts.

This is the EVM invariant that is missing: on EVM you can only `allow` a handle you already hold, rooted in the `InputVerifier` attestation. **Fix** is provenance binding at create/update (require a
`CoprocessorInputAttestation` tying the handle to the creator, or require the handle be one the caller is already allowed on) — `assert_handle_for_chain` alone does not close it.

---

## Composability walls (gate: breaks desired DeFi functionality)

### 2. Single-recipient-per-sender transfer lineage

`transferred_amount_value` is keyed by `(mint, from_account, transferred_amount_label)` (`confidential_transfer.rs:33`, `common.rs:167`), but its subjects include the current recipient, and host supe
rsede requires the new subject set to equal the stored one exactly (`fhe_eval.rs:582`). So Alice→Bob seals `[Alice, Bob, compute]`; a later Alice→Carol plans `[Alice, Carol, compute]` and hits `Previ
ousStateMismatch`. **A sender can only ever transfer to one recipient.** Independently confirmed by the token audit, the confidential-swaps panel, and the ERC-7984 panel. Blocks one-to-many payouts,
`transferFrom`, routers, and swap-settlement escrow. Fix path: include recipient (or an op nonce) in the transferred-amount lineage key, or use ad-hoc decrypt-only leaves that don't mutate the shared
 sender subject set.

### 3. No existing-handle spend path

ERC-7984 has `confidentialTransfer(address, euint64)` for already-allowed handles; the Solana production transfer only accepts a fresh `CoprocessorInputAttestation` (`confidential-token/src/lib.rs:10
4`, `common.rs:194`). `fhe_eval` can *produce* durable outputs but the token transfer API can't *spend* one — so a router receiving `amountTransferred` from token A cannot forward that exact handle i
nto token B. Blocks router / vault-withdrawal / contract-computed-payout flows independently of #2.

### 4. Finalized-lag permanent material loss

The finalized-account fetcher completes a job as soon as `getMultipleAccounts(finalized)` returns any account, without checking `observed_slot >= job.block_number` (`solana_finalized_account_fetcher.
rs:582`). Against a finalized-lagging RPC, an `H1→H2` supersession can complete on a stale read showing `H1`, the claim mismatches, the job goes terminal, and **no `pbs_computations` row is ever writ
ten for `H2`** — that balance is permanently undecryptable until manual requeue (retry only covers not-found accounts, not claim-mismatched ones). The same path lets a minority-fork `handle_supersede
d` push orphan-handle material to gateway `CiphertextCommits`. It never causes a wrong *release* (KMS re-verifies against live finalized peaks), so this is availability/data-loss, not a leak — but it
's the kind of permanent-loss that warrants fixing before relying on it.

---

## Non-blocking findings (file as follow-ups)

- **SDK current-decrypt requires a caller-supplied `aclValueKey`** (`sdk/js-sdk/src/solana/actions/userDecrypt.ts:223`) — the app must pass ACL internals rather than the SDK deriving them from `(mint
, owner, label)`. Diverges from the intended "one signMessage, ACL invisible" UX; the "always send a real aclValueKey" fix only made it non-empty.
- **Single-handle user-decrypt limit** — KMS rejects any request with ≠1 handle, but the SDK will sign+post a multi-handle current request (late failure), and this divergence from EVM's handle-array
isn't recorded in `EVM_PARITY.md`.
- **Relayer `fhe_eval` decoder uses a stale account-index base (7/8 vs current 10)** (`relayer/src/solana_proof`) — wrong/no lineage attribution for real eval events; fails safe for release (peaks cr
oss-check) but breaks proof availability for eval outputs.
- **kms-worker recoverable failures skip the attempt cap** (`processor.rs:82`) — a non-subject / rotated-handle request retries forever instead of erroring out.
- **Relayer recovery tooling** — silent `getTransaction→null` cursor skip; `BacklogExceeded` points at a backfill tool that doesn't exist; hot-lineage catch-up budget exhaustion after store loss; una
uthenticated internal `/internal/solana/mmr-proof` route. All availability-only (the proof service is untrusted; proofs are re-verified against live peaks).
- **Default RPC listener transport is confirmed-commitment and compute-blind**, so RPC and gRPC don't produce identical facts — mitigated only by CI/docs pinning gRPC `--reconstruct` as the flow of r
ecord.
- **Demo flow-of-record is not fully public-SDK** — the VerifiedInput/transfer transaction is submitted by the `live-client` script, not `@fhevm/sdk/solana` or `fhevm-cli` (no transfer/deposit action
 in the public SDK surface yet).
- **EIP-712 recovery accepts v=29/30** — non-exploitable (a mutated `v` can't recover into the signer set), worth a one-line tighten.
- **Deletion remnants in the production surface** — legacy `AclPermission` fetch kind; receiver-callback error/event variants and `RandomAmountCreatedEvent` still in the production IDL/ABI golden; `H
andleMaterialCommitment` tail; dead `find_acl_records_by_handle`.
- **Doc drift** — the PR description still describes an `ACL_ROLE_USE` / `PUBLIC_DECRYPT` role-bit model that doesn't exist in the shipped account layout (which is a flat allowed-subject set, matchin
g DD-032); the "RandBounded removed" known-gap is stale (the step is implemented on this branch).
- **`extraData` `0x02` version collision** with main's RFC-016 envelope (+epochId vs +MMR tail) — real merge-to-main hazard, noted here since the branch is expected to be re-cut.

---

## Verified clean (held up under adversarial verification)

- MMR append/peak/proof math (hand-traced incl. `u64::MAX` boundary); `HistoricalAccessLeaf`/`PublicDecryptLeaf` domain separation; **Rust↔TS preimage parity** and **`sol_sha256` ↔ software hash byte
-identity** (no accumulator fork).
- On-chain invariants: supersession is authority-gated (allowed-subject ≠ authority), stale rotation guarded by exact previous-handle/subjects equality, PDA lookalike-spoofing resisted.
- KMS boundary: no authorization bypass; checks run in the correct order; **structurally mirrors the EVM connector pipeline** rather than a parallel implementation.
- Current & public decrypt: exact-handle scoped, **no roll-forward** (a proof for an old handle is rejected for a rotated one), born-public leaf ordering per DD-036, and the KMS **refuses** a public
decrypt lacking a valid `PublicDecryptLeaf` proof (so the deferred gateway `checkSignatures` leg can't cause early release).
- Self-transfer is a true no-op; same-frame supersession is recognized but **cannot be replayed cross-frame or cross-transaction** (frame state is per-invocation).
- **Reorg safety:** no path releases plaintext for an account not readable at finalized commitment with matching state; same-handle-with-different-ciphertext across a fork is unreachable (handle deri
vation binds `previous_bank_hash`/operands and FHE eval is deterministic), so the first-writer caches are safe.
- **Event-free indexing is the right call** (recommend keeping it, formalized as the hybrid it already is): Solana has no consensus event primitive, so `emit_cpi` and instruction-replay have identica
l durability — reintroducing lifecycle events would spend CPI depth and account slots (the exact headroom DeFi composition needs) for zero durability gain.

---

## North-star verdict (DeFi-readiness)

The EncryptedValue + MMR ACL core is DeFi-portable with Solana-idiomatic changes (app-driven CPI deposit, escrow PDAs, public-decrypt proof handoff for branch-on-value). The **current super-app flows
** (shield/send/unshield, vault deposit/redeem/claim, fast-redeem) map on without a structural rewrite — they're one-to-one or aggregate-then-claim. The two walls above become structural only for **f
uture** bulk multi-recipient, router, and mass-payout patterns, and both live in the token transfer surface, not the ACL. Concrete budget note: app-driven deposit sits at **CPI depth 4 exactly** when
 host event CPI is on (≈17 accounts, well under 64), so vault extensions must keep additional FHE evals sequential after the token CPI returns rather than nested.

---

*Coverage note: this consolidates the chunk audits (ACL crate, zama-host, token, listener, relayer, KMS connector), the decrypt/transfer/fork journeys, and the forward-fit panels (swaps, confidential
-defi, super-app, ERC-7984), plus prior-review adjudication, deletion sweep, gap ratings, and the event-free analysis. A consolidated SDK-codegen/docs/test-coverage pass and the historical-decrypt jo
urney were still running at time of posting and will follow up. The initial Fable-based run was cut short by a spend limit and re-run on a separate model; wrong-shape issues were surfaced organically
 by the chunk/journey agents rather than a dedicated sweep.*

Eikix:
## Follow-up: wrong-shape / architecture pass + reconciliation against current head

Addendum to the consolidated review above. This pass ran a dedicated *wrong-shape* sweep per trust domain (against the EVM-counterpart architecture) plus the SDK-codegen/docs/test-coverage and historical-decrypt legs that were still running when the first comment went up. Reviewed at `e8fa43854`, then reconciled against current head `a5260fcf4`.

### Reconciled against head

- **Already fixed:** `a5260fcf4` corrects the relayer `fhe_eval` durable-output decode base to match the host-listener — that's the decoder-index drift flagged above. Resolved. (The underlying shape point — two separate decoders that *can* drift because the logic is duplicated rather than shared — still stands as a maintainability item.)
- **Refactored, re-verify:** `9f7aaec8c` drops the `value_key` handle binding (durable outputs use the base handle, matching EVM) and updates `MMR_ACL_MVP.md`/`DESIGN_DECISIONS.md` to match. This touches the handle-derivation area, so the handle-derivation wrong-shape note below and a couple of handle-derivation doc-drift items may be partly addressed on head.
- **Unaffected:** the confirmed provenance blocker is untouched by both commits (`create_encrypted_value`/`update_encrypted_value` and the KMS current-check are not in these diffs) — **it still stands on head**, and dropping the output binding arguably reduces handle↔lineage coupling further.

### Architectural (wrong-shape) findings — the roots under the individual bugs

These are "works today, but built as a parallel path an EVM-side maintainer wouldn't recognize" — worth tracking even on a PoC because they're where future drift and re-work come from.

- **The JS SDK is a side-car, not a chain module.** `@fhevm/sdk/solana` builds `Fhevm<undefined>` with an `EthereumModule` whose methods throw, fabricates an EVM-shaped `chain`, and passes `aclProgramAddress` outside chain config (`sdk/js-sdk/src/solana/clients/createFhevmBaseClient.ts`, `internal/ethereum.ts`). Two consequences: the Solana encrypt action stops at local proof construction and leaves the `/v2/input-proof` round-trip to callers — so it **skips the returned-handle verification EVM's `encryptValue` performs**; and current-decrypt requires a caller-supplied `aclValueKey` even though `deriveValueKey(domain, appAccount, label)` already exists but isn't exported/used. Right shape: a real Solana chain module behind the core abstractions.
- **kms-connector dispatch table is forked.** EVM host chains go in one `HashMap<chain_id, ACLInstance>` with duplicate-rejection; Solana chains are skipped there and rebuilt in a second `HashMap<chain_id, SolanaHost>` via plain `.collect()`, so duplicate Solana chain-ids silently overwrite and missing-config behaviour diverges (user-decrypt fails closed, public-decrypt falls through to the EVM ACL path). The worker *pipeline* is correctly shared — it's the per-chain dispatch seam that's parallel. Right shape: one `HostAclBackend::{Evm,Solana}` table with unified validation.
- **The relayer MMR proof service is a parallel bespoke pipeline** (its own RPC poller, decoder, replay/indexer, JSON file store, public route) duplicating the host-listener's ingestion — this duplication is *why* the decoder drifted in the first place. Also: the internal `/internal/solana/mmr-proof` route is mounted on the public router without the throttle/budget the decrypt handlers have.
- **Handle derivation is scattered and the ABI golden doesn't pin it.** Host handle derivation picked keccak vs SHA-256 per op family inline rather than behind one executor-shaped primitive, and `check_solana_abi.py` hashes instruction name/discriminator/args but **not account metas** while off-chain reconstruction relies on hard-coded account indices — the systemic root that let the account-index drift happen silently. (`9f7aaec8c` may have narrowed the derivation half of this; the golden-coverage gap remains.)

### Reassuring results from this pass

- **Rust↔TS preimage parity is clean** (leaf prefixes, signing-domain v2, `aclValueKey`/`proofSlot`/proof bytes all match), and the on-chain `sol_sha256` path is byte-identical to the software hash — no accumulator fork.
- **Test counts are higher than the PR body claims, not lower** (e.g. host-listener 138 vs 100, Mollusk 102 vs 33, relayer 47 vs 27), and the two adversarial cases called out as at-risk — cap-8 enforcement and denied-caller-cannot-mutate — **both have explicit tests**.
- **Historical sealed-forever semantics are verified** in code and tests: a subject sealed into a historical leaf still decrypts that handle after `remove_subject`, and a subject removed before a later supersession correctly gets no leaf for the new handle. Historical user-decrypt itself works end-to-end today only with a hand-built MMR proof (the OQ2 productization gap) — and it fails closed if the proof is absent/stale.

Net: the ACL/MMR core holds up; the hardening work is concentrated in the SDK/relayer/kms-connector *shape* (parallel paths that should be chain modules behind the shared abstractions) and the one provenance blocker. Happy to open fhevm-internal issues for these so they don't live only in the PR thread.

Eikix:
## Must-fix before merge (fold into this PR)

Given the PoC-converges-the-spec posture, most review findings are follow-up issues (design / re-cut work) — those are being filed in fhevm-internal, not patched here. This comment is the short list that should land **in this diff before merge**, per the merge gate (wrong-decrypt, on-chain-invariant, hash-parity, data-loss, flow-breaking).

### Blocker — wrong plaintext release
- **Bind handle provenance on `create_encrypted_value` / `update_encrypted_value`.** They store a caller-chosen raw `handle` with no attestation and no check the caller already holds it (`solana/programs/zama-host/src/instructions/encrypted_value.rs:37`, `:172`), and the KMS current-path authorizes on `current_handle == handle` + subject membership + a self-asserted domain list (`solana/crates/zama-solana-acl/src/lib.rs:209`, `kms-connector/.../solana_encrypted_value_acl.rs:105`). A forged lineage at the attacker's own PDA claiming a victim's handle passes every check → the KMS releases the victim's plaintext. Untouched by `9f7aaec8c`, so it stands on head. Fix: require a `CoprocessorInputAttestation` binding the handle to the creator, or require the handle be one the caller is already allowed on (EVM `InputVerifier` parity). `assert_handle_for_chain` alone does not close it.

### Data-loss — permanent undecryptable value
- **Gate the finalized fetcher on `observed_slot >= job.block_number`.** A job completes as soon as `getMultipleAccounts(finalized)` returns any account, without comparing slots (`coprocessor/fhevm-engine/host-listener/src/solana_finalized_account_fetcher.rs:582`). Against a finalized-lagging RPC, an `H1→H2` supersession completes on a stale `H1` read, the claim mismatches, the job goes terminal, and no `pbs_computations` row is ever written for `H2` — that value is permanently undecryptable (retry only covers not-found accounts). No wrong release (KMS re-verifies), but it's data-loss. Fix: retry (don't complete) when the witness slot precedes the job's slot, or read with `minContextSlot`.

### Cheap robustness / parity — fold while you're here
- **Include `UserDecryptionSolana` in the decryption attempt-cap match** (`kms-connector/.../event_processor/processor.rs:82`) — today Solana recoverable failures (e.g. an account that never finalizes) requeue to pending forever instead of erroring out after `max_decryption_attempts`. One arm + the latency-histogram arm.
- **Reject secp256k1 `v` outside {27,28}** in the host recovery path (`solana/programs/zama-host/src/eip712.rs`) — currently accepts 29/30 (recovery id 2/3). Non-exploitable, but it's a one-line EVM/OZ parity fix.

### Doc corrections — the PR's own record is currently wrong
- The **PR description still describes a role-bit ACL model** (`PUBLIC_DECRYPT`, `subject_roles`, `ACL_ROLE_USE`) that isn't shipped — the account is a flat allowed-subject set (matches DD-032). Correct the description.
- **`RandBounded` is documented as "removed"** but is implemented end-to-end on this branch (`fhe_eval/walk.rs:258`, `create_random_amount.rs`); **DD-033 still lists four lifecycle instructions and omits `remove_subject`**; **DD-026 still says user-decrypt `extraData` is context-only** while the branch sends/parses a `0x03` MMR tail. (`9f7aaec8c` fixed some of the handle-derivation docs — worth a pass to catch the rest.)

### Explicitly NOT folded here — tracked as fhevm-internal issues
The two token composability walls (single-recipient `transferred_amount` lineage; no existing-handle spend), the SDK-side-car / kms-dispatch-fork / relayer-parallel-pipeline architecture, and the extraData↔RFC-016 collision (**#1639**) and historical-decrypt productization (**#1636**) are spec / re-cut work, not this diff. Issues incoming.

PanGan21 (approved):
Nice work

PR #3011, opened 2026-07-06
feat(kms-connector): add key material migration flow

Eikix:
Draft split from RFC-029 after the RFC-029/blue-green decision: this PR keeps the durable key-material production path and leaves worker activation/cutover to blue-green.

This version replaces the confusing `FromExisting` keygen model with an explicit migration API:

- `keyId` remains the durable key identity.
- `migrateKey(keyId)` creates a KMS migration request for the active key.
- `migrationRequestId` is the KMS consensus handle for producing migrated compressed material; it is not accepted by generic key getters.
- `migrationResponse(migrationRequestId, ...)` publishes compressed material for the durable `keyId` without activating a new key.
- `getKeyMaterials(keyId)` returns the latest material after migration.
- `getAllKeyMaterials(keyId)` returns the original and migrated consensus records, so legacy material remains available.

The standalone cutover/worker-selection path from the closed RFC-029 PR remains intentionally out of scope and should be handled by blue-green.

Validated locally:

- `npx hardhat compile` in `host-contracts`
- `forge test --offline --match-contract KMSGenerationTest` in `host-contracts` (`86` tests)
- `SQLX_OFFLINE=true cargo check --workspace --all-targets` in `kms-connector`
- `cargo fmt --check` in `kms-connector`
- `git diff --check`


host-contracts/contracts/KMSGeneration.sol
```diff
@@ -219,7 +230,7 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    function reinitializeV2() public virtual reinitializer(REINITIALIZER_VERSION) {
+    function reinitializeV2() public virtual reinitializer(3) {
```
Eikix: why leave reinitializeV2? the convention is to replace it by reinitializeV3

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -167,6 +167,17 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
         uint256[] completedKeyIds;
         /// @notice CRS IDs that reached consensus.
         uint256[] completedCrsIds;
+        // ----------------------------------------------------------------------------------------------
+        // Compressed key material production variables:
+        // ----------------------------------------------------------------------------------------------
+        /// @notice The existing key a migration keygen request re-materializes. Non-zero iff the
+        /// request ID is a migration request; this is what keeps the migration and normal keygen
+        /// response endpoints mutually exclusive.
+        mapping(uint256 migrationRequestId => uint256 keyId) migrationRequests;
```
Eikix: why do we still need migration requests?

Eikix:

No blocking bug found in the migration state machine itself; one upgrade-path question that looks like a real state-corruption risk, plus two smaller questions.

1. **`reinitializeV3` re-runs the V2 backfill with no dedup guard** — `host-contracts/contracts/KMSGeneration.sol:231-249`. `origin/main` already ships `reinitializeV2` (reinitializer version 3) with this exact backfill body; this PR renames it to `reinitializeV3` (version 4) and keeps the loop. Any proxy that has already executed `reinitializeV2` will, on this upgrade, push every completed key/CRS id into `completedKeyIds`/`completedCrsIds` a second time, so `getCompletedKeyIds()`/`getCompletedCrsIds()` return duplicates forever. The PR also deletes the only test covering this path (`test_reinitializeV2BackfillsOnlyCompletedIds` + `KMSGenerationReinitHarness` in `kmsGeneration.t.sol`) instead of adapting it. If the assumption is that v0.2.0 was never executed on any live proxy, a `if ($.completedKeyIds.length != 0) return;`-style guard (or a comment stating the assumption) would make the upgrade safe by construction.

2. **Migration prep-keygen is sent to KMS-core with the uncompressed keyset config and no migration marker** — `kms-connector/crates/kms-worker/src/core/event_processor/kms.rs:57-67` still sends `UNCOMPRESSED_KEY_SET_CONFIG` in `KeyGenPreprocRequest`, while the paired keygen sends `CompressedAll`/`UseExisting` + `KeySetAddedInfo`. The contract added `requestKind`/`keyId` to `PrepKeygenRequest` precisely so the prep phase can be distinguished, but the connector drops both: `prep_keygen_requests` has no column for them and `from_prep_keygen_row` hardcodes `requestKind: 0, keyId: 0` (`kms-connector/crates/utils/src/types/event.rs:352-357`). If KMS-core sizes preprocessing by keyset config, the prep/keygen configs are inconsistent for migrations; if it ignores it, the new event fields are currently dead weight in the connector. Worth stating which.

3. **kms-worker abort processing lost its integration coverage** — `kms-connector/crates/kms-worker/tests/integration_tests.rs` removes all four `abort_keygen`/`abort_crsgen` cases and `wait_for_abort_completed`, replacing the handlers with `unreachable!`, while the abort→KMS-core path is still live in src (`processor.rs:217`, `kms_client.rs:101`). gw-listener catchup gained abort cases, but the worker's `AbortKeyGen` gRPC path is now untested. Relatedly, `check_no_uncompleted_request_in_db` (`utils/src/tests/db/requests.rs`) no longer treats `'aborted'` as terminal even though `OperationStatus::Aborted` is still written by `mark_as_aborted` (`event.rs:89`).

Specifically verified: cross-replay between `keygenResponse` and `migrationResponse` is impossible (disjoint request-id guards at `KMSGeneration.sol:397` and `:463` over a shared counter and shared EIP-712 typehash); abort-then-retry of a migration works (`abortKeygen` sets `isRequestDone` with zero digest, which `migrateKey` accepts for retry, and late responses to the aborted request cannot reach consensus); migration ids are correctly rejected by the generic getters via `_checkGeneratedKeyId`; and the connector's `migration_key_id` round-trip (publish → picker → tx-sender's `is_migration` subquery) is consistent in LE encoding and keyed on the request id throughout.

PR #3016, opened 2026-07-06
feat(host-contracts): remove pending abort methods and add destroyKmsEpoch

obatirou:
Closes zama-ai/planning-blockchain#1254


host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -483,30 +481,18 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
-        emit PendingEpochAborted(contextId, epochId);
-    }
-
-    /// @inheritdoc IProtocolConfig
-    function abortPendingContext(uint256 kmsContextId) external virtual onlyACLOwner {
-        ProtocolConfigStorage storage $ = _getProtocolConfigStorage();
-        if ($.contextState[kmsContextId] != ContextState.Pending) {
-            revert KmsContextNotPending(kmsContextId);
+        if ($.epochState[epochId] != EpochState.Active) {
+            revert InvalidEpoch(epochId);
```
isaacdecoded: nit: maybe consider `InvalidKmsEpoch`, to be aligned with other namings like `KmsEpochDestroyed` event

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -275,15 +268,13 @@ interface IProtocolConfig {
     /// @param kmsContextId The current context ID.
     error CurrentKmsContextCannotBeDestroyed(uint256 kmsContextId);
 
-    /// @notice The epoch ID is invalid or not pending.
-    /// @param epochId The epoch ID.
-    error InvalidEpoch(uint256 epochId);
+    /// @notice Cannot destroy the current active epoch.
+    /// @param epochId The current epoch ID.
+    error CurrentKmsEpochCannotBeDestroyed(uint256 epochId);
```
isaacdecoded: nit: should we consider renaming to `LatestActiveKmsEpochCannotBeDestroyed`? Same for the `CurrentKmsContextCannotBeDestroyed` -> `LatestActiveKmsContextCannotBeDestroyed`

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -143,17 +143,10 @@ interface IProtocolConfig {
...
+     * @param epochId The destroyed epoch ID.
      */
-    event PendingEpochAborted(uint256 indexed kmsContextId, uint256 indexed epochId);
-
-    /**
-     * @notice Emitted when a pending KMS context is aborted before being created.
-     * @param kmsContextId The aborted pending context ID.
-     */
-    event PendingContextAborted(uint256 indexed kmsContextId);
+    event KmsEpochDestroyed(uint256 indexed epochId);
```
isaacdecoded, line 149: @eudelins-zama @jot2re please be aware of this new `KmsEpochDestroyed` event, for the KMS to propagate the epoch destruction.

host-contracts/tasks/kmsContext.ts
isaacdecoded, line 1: Just wondering if we should provide two hardhat tasks to generate both DAO proposals: `destroyKmsContext` and `destroyKmsEpoch` (?)

obatirou: Done

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: Done

host-contracts/contracts/ProtocolConfig.sol
obatirou: Done

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -472,41 +470,40 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
             revert InvalidKmsContext(kmsContextId);
         }
 
-        _clearContext(kmsContextId);
+        // Mark the context destroyed, clear its pending epoch (if any), and wipe context-creation bookkeeping.
+        $.destroyedContexts[kmsContextId] = true;
+        $.contextState[kmsContextId] = ContextState.None;
+        // At most one epoch is Pending at a time, so the latest-issued epoch is the only candidate.
+        uint256 latestEpochId = $.epochCounter;
+        if ($.epochState[latestEpochId] == EpochState.Pending && $.contextForEpoch[latestEpochId] == kmsContextId) {
```
obatirou: we only check for pending epoch
Should this be cleared even if in active state?
I think so.
removing `$.epochState[latestEpochId] == EpochState.Pending` would fix it?

host-contracts/contracts/ProtocolConfig.sol
obatirou, line 480: Should we update the gateway equivalent?
No updates are needed regarding 0.14 on gateway?

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: I also think we should remove this `EpochState.Pending` condition. But I'm still wondering what the behavior on the KMS side is when propagating the `KmsContextDestroyed` event? Which I guess is related to [this gRPC endpoint](https://github.com/zama-ai/kms/blob/main/core/grpc/proto/kms.v1.proto#L634-L640), where we can see it receives a list of `epoch_ids`

Maybe @jot2re you know?

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: It [seems the GatewayConfig is upgraded](https://github.com/zama-ai/fhevm/compare/v0.13.1...main#diff-4139d8055b558914fe9fc072c5a6a27496114f8e6bc3e081b09d321f2528dd59) indeed. But not sure what "equivalent update" you mean here. The event naming? If yes, I'd say we don't need to since no one is listening to this event on the Gateway side.

obatirou: I meant metadata equivalence and destroyKmsContext equivalence also (but not relevant in fact as epoch is only on ethereum).
We will call `[detroy/define]KMSContext` on gateway when we do it on ethereum. Just want to make sure we do not miss something here

host-contracts/contracts/ProtocolConfig.sol
jot2re: So the current logic is that the KMS needs to receive explicitly the epochs that are part of a context. What it then does is to physically delete the epochs and at the end the context the context destruction.

obatirou: Is this done or can this be taken into account for future implementation on KMS connector side?
cc @eudelins-zama

isaacdecoded (dismissed):
LGTM! 👏

isaacdecoded (dismissed):

host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -394,26 +385,18 @@ interface IProtocolConfig {
...
      * @notice Destroy a KMS context, preventing it from being used.
      * @param kmsContextId The context ID to destroy.
      */
     function destroyKmsContext(uint256 kmsContextId) external;
 
+    /**
+     * @notice Destroy a superseded (non-current) KMS epoch, preventing it from being used.
+     * @param epochId The epoch ID to destroy.
+     */
+    function destroyKmsEpoch(uint256 epochId) external;
```
jot2re, line 398: Shouldn't we add the same comment as for context destruction in the sense that we will not allow destroying the latest active one?

obatirou: mmh indeed. I can do that in a followup PR.
I think that is ok anyway as they are DAO functions and DAO is responsible to double check conditions

obatirou:
@mergifyio refresh

obatirou:
@mergifyio queue

PanGan21 (approved):

obatirou:
@Mergifyio refresh

eudelins-zama:
@mergifyio queue

PR #3018, opened 2026-07-06
feat(host-contracts): add ConfidentialBridge deploy and prepareUpgrade tasks for existing chains

PanGan21:
### Summary

Adds the standalone tasks needed to bring the ConfidentialBridge onto an already-deployed host chain (the fresh-chain path is already covered by task:deployAllHostContracts):

- task:deployEmptyProxyForConfidentialBridge: deploys only the bridge empty UUPS proxy and pins its address (deployEmptyUUPS + setBridgeAddress).
- task:prepareUpgradeConfidentialBridge: deploys the ConfidentialBridge implementation (with the immutable LayerZero endpoint constructor arg) and prints the governance upgradeToAndCall calldata.
- removes legacy deployEmptyProxiesProtocolConfigKMSGeneration task

### DIFF
--- a/host-contracts/tasks/upgradeContracts.ts
+++ b/host-contracts/tasks/upgradeContracts.ts
@@ -42,6 +42,19 @@ function shellQuote(arg: string): string {
   return `'${arg.replace(/'/g, `'\\''`)}'`;
 }
 
+// Parses a comma-separated integer list task arg (e.g. --dst-eids "30101,30109") into its entries.
+// Empty/undefined yields an empty array; values stay as strings so ethers coerces them to the
+// solidity integer type at encode time.
+function parseCsvIntegers(raw: string | undefined): string[] {
+  if (!raw) {
+    return [];
+  }
+  return raw
+    .split(',')
+    .map((entry) => entry.trim())
+    .filter((entry) => entry.length > 0);
+}
+
 async function upgradeCurrentToNew(
   proxyAddress: string,
   currentImplementation: string,
@@ -353,6 +366,18 @@ task('task:prepareUpgradeConfidentialBridge')
     types.boolean,
   )
   .addOptionalParam(
+    'dstEids',
+    'Comma-separated LayerZero endpoint ids to seed the dstEid → dstChainId map (paired index-by-index with --dst-chain-ids). Empty by default; pairs can also be wired later via task:setDstChainId.',
+    '',
+    types.string,
+  )
+  .addOptionalParam(
+    'dstChainIds',
+    'Comma-separated destination chain ids paired index-by-index with --dst-eids.',
+    '',
+    types.string,
+  )
+  .addOptionalParam(
     'verifyContract',
     'Verify new implementation on Etherscan (for eg if deploying on Sepolia or Mainnet)',
     true,
@@ -366,6 +391,14 @@ task('task:prepareUpgradeConfidentialBridge')
     const lzEndpoint = getRequiredEnvVar('LZ_ENDPOINT_ADDRESS');
     if (!hre.ethers.isAddress(lzEndpoint)) {
       throw new Error(`LZ_ENDPOINT_ADDRESS is not a valid address: ${lzEndpoint}`);
+    }
+
+    const dstEids = parseCsvIntegers(taskArgs.dstEids);
+    const dstChainIds = parseCsvIntegers(taskArgs.dstChainIds);
+    if (dstEids.length !== dstChainIds.length) {
+      throw new Error(
+        `--dst-eids and --dst-chain-ids must have the same length: got ${dstEids.length} eid(s) and ${dstChainIds.length} chain id(s). initializeFromEmptyProxy would revert with DstChainIdArrayLengthMismatch.`,
+      );
     }
 
     await hre.run('compile:specific', { contract: 'contracts' });
@@ -387,7 +420,7 @@ task('task:prepareUpgradeConfidentialBridge')
     console.log('New implementation deployed at:', implementationAddress);
 
     const initSignature = 'initializeFromEmptyProxy(uint32[],uint64[])';
-    const initArgs: unknown[] = [[], []];
+    const initArgs: unknown[] = [dstEids, dstChainIds];
     const initCalldata = bridgeFactory.interface.encodeFunctionData('initializeFromEmptyProxy', initArgs);
     const outerCalldata = new Interface([
       'function upgradeToAndCall(address newImplementation, bytes data) payable',
--- a/host-contracts/tasks/upgradeContracts.ts
+++ b/host-contracts/tasks/upgradeContracts.ts
@@ -4,6 +4,7 @@ import { HardhatRuntimeEnvironment, TaskArguments } from 'hardhat/types';
 
 import { buildProtocolConfigReinitializeArgs } from './taskDeploy';
 import { getRequiredEnvVar, loadHostAddresses } from './utils/loadVariables';
+import { buildUpgradeProposal, printUpgradeProposal, verifyProposalImplementation } from './utils/upgradeProposal';
 
 const REINITIALIZE_FUNCTION_PREFIX = 'reinitializeV'; // Prefix for reinitialize functions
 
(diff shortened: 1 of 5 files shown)


host-contracts/tasks/upgradeContracts.ts
```diff
@@ -345,6 +345,76 @@ task('task:prepareUpgradeACL')
...
+      await hre.upgrades.prepareUpgrade(proxyAddress, bridgeFactory, {
+        kind: 'uups',
+        constructorArgs: [lzEndpoint],
+        unsafeAllow: ['constructor', 'state-variable-immutable', 'missing-initializer-call'],
+      }),
+    );
+    console.log('New implementation deployed at:', implementationAddress);
+
+    const initSignature = 'initializeFromEmptyProxy(uint32[],uint64[])';
+    const initArgs: unknown[] = [[], []];
```
obatirou: Is this ok to leave those empty? Even for polygon?

obatirou: might need to be parametrizable

PanGan21: Empty is intentional since the contract documents this as valid and the setDstChainId and setPeer are wired post-deploy.
Do you think we should make it parametrizable so a chain that already knows its pairs can add them in the upgradeToAndCall?

obatirou: If there is a need for it yes.
Else I am not sure why those input were added in the first place?
I mean, if we know those are not supplied during the init, why add them?

PanGan21: I think they were added in the constructor for future flexibility in case we already know the dst-eids and dst-chain-ids before hand. However I think with the current plan they will not be used.
Rather than removing them i had added the parameters to the task in: https://github.com/zama-ai/fhevm/pull/3018/commits/c1d9527a97540f372f1bea8101dd234c15d2f313

host-contracts/tasks/upgradeContracts.ts
```diff
@@ -345,6 +345,76 @@ task('task:prepareUpgradeACL')
     await prepareUpgradeContract('ACL', 'ACL_CONTRACT_ADDRESS', taskArgs, hre);
   });
 
+task('task:prepareUpgradeConfidentialBridge')
```
obatirou, line 363: maybe add a comment that this is only for the first deployment from emptyProxy. On next upgrade this should be updated to use `prepareUpgradeContract` or equivalent normally.
For new chains the deployAllHostContract (or equivalent single steps) are used

host-contracts/tasks/upgradeContracts.ts
```diff
@@ -345,6 +358,96 @@ task('task:prepareUpgradeACL')
...
+    const emptyProxyFactory = await hre.ethers.getContractFactory('EmptyUUPSProxy', deployer);
+    await hre.upgrades.forceImport(proxyAddress, emptyProxyFactory);
+
+    const bridgeFactory = await hre.ethers.getContractFactory('ConfidentialBridge', deployer);
+    console.log(`Deploying "ConfidentialBridge" for prepared upgrade on proxy ${proxyAddress}...`);
+    const implementationAddress = String(
+      await hre.upgrades.prepareUpgrade(proxyAddress, bridgeFactory, {
+        kind: 'uups',
+        constructorArgs: [lzEndpoint],
+        unsafeAllow: ['constructor', 'state-variable-immutable', 'missing-initializer-call'],
```
PanGan21: @obatirou fyi this is required in order to upgrade the bridge with lz. I can pass those args all the way down with prepareUpgrade if we want to reuse prepareUpgradeContract but that would mean more changes

host-contracts/tasks/upgradeContracts.ts
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/3018/commits/28f01f6a12d8aa5d281af16f30b0b810d8b142cf

host-contracts/tasks/upgradeContracts.ts
```diff
@@ -345,6 +356,99 @@ task('task:prepareUpgradeACL')
...
+    );
+
+    if (taskArgs.verifyContract) {
+      console.log('Waiting 2 minutes before contract verification... Please wait...');
+      await new Promise((resolve) => setTimeout(resolve, 2 * 60 * 1000));
+      await hre.run('verify:verify', {
+        address: implementationAddress,
+        contract: 'contracts/bridge/ConfidentialBridge.sol:ConfidentialBridge',
+        constructorArguments: [lzEndpoint],
+      });
```
obatirou: you could use verifyProposalImplementation and extend it to accept a constructor arg?

PanGan21: Ah same comment here: https://github.com/zama-ai/fhevm/pull/3018#discussion_r3530431521

host-contracts/tasks/upgradeContracts.ts
```diff
@@ -345,6 +356,99 @@ task('task:prepareUpgradeACL')
...
+    }
+
+    await hre.run('compile:specific', { contract: 'contracts' });
+
+    const deployer = new Wallet(getRequiredEnvVar('DEPLOYER_PRIVATE_KEY')).connect(hre.ethers.provider);
+
+    const emptyProxyFactory = await hre.ethers.getContractFactory('EmptyUUPSProxy', deployer);
+    await hre.upgrades.forceImport(proxyAddress, emptyProxyFactory);
+
+    const bridgeFactory = await hre.ethers.getContractFactory('ConfidentialBridge', deployer);
```
obatirou: I think buildUpgradeProposal could be used to simplify this

host-contracts/tasks/upgradeContracts.ts
obatirou: Ah yes indeed, will let you decide in this case.
I am fine either way.
Will not block this PR anymore. Seems correct to me anyway

host-contracts/tasks/upgradeContracts.ts
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/3018/commits/8569ef50dc9935f599f6241eb54be2ef456d678c

host-contracts/tasks/upgradeContracts.ts
jatZama: Correct, they were added for flexibility, but not used today.

host-contracts/tasks/upgradeContracts.ts
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/3018/commits/8569ef50dc9935f599f6241eb54be2ef456d678c

host-contracts/tasks/upgradeContracts.ts
PanGan21: Done in: https://github.com/zama-ai/fhevm/pull/3018/commits/8569ef50dc9935f599f6241eb54be2ef456d678c

host-contracts/tasks/utils/upgradeProposal.ts
```diff
@@ -38,17 +40,22 @@ export async function buildUpgradeProposal(
...
     innerFunctionName: string;
     decodedArgs: unknown[];
+    constructorArgs?: UpgradeOptions['constructorArgs'];
+    unsafeAllow?: UpgradeOptions['unsafeAllow'];
   },
 ): Promise<UpgradeProposal> {
   const { ethers, upgrades } = hre;
   const privateKey = getRequiredEnvVar('DEPLOYER_PRIVATE_KEY');
   const deployer = new ethers.Wallet(privateKey).connect(ethers.provider);
+  const constructorArgs = params.constructorArgs ?? [];
```
obatirou: Why necessary for contructor args but not for unsafe params? Both are array though

obatirou (approved):

host-contracts/tasks/utils/upgradeProposal.ts
PanGan21: That was because constructorArgs is reused later on. But I changed and I am resolving empty array on verify.
Done in: https://github.com/zama-ai/fhevm/pull/3018/commits/e54f5b5a7f47c2624687a7951b09dcaf9eea04b0

antoniupop:
@mergifyio dequeue

PanGan21:
@Mergifyio queue

PR #3024, opened 2026-07-06
fix(sdk): fix localstack setup below v0.14 and add js-sdk tests

geoxel:
## Summary
- Pin the fhevm repo version used by the localstack test scripts so tests keep passing against fhevm versions below v0.14
- Add localstack fheTests covering `decryptValuesFromPairs` and `canDecryptValue(s)`/`canDecryptValuesFromPairs` for both ethers and viem
- Add unit tests for `core/base/errors/utils`, `core/base/object`, `core/base/trustedValue`, `core/handle/ClearValue`, and `core/handle/FhevmHandle`
- Fix `MAX_UINT_FOR_TYPE` lookup in `asClearValueType` for `euint32`/`euint256`, which was indexing by the raw FHE type name instead of the clear type name and could return `undefined` for the max bound


0xalexbel (approved):
LGTM

PR #3030, opened 2026-07-07
feat(host-contracts): add ConfidentialBridge deploy and prepareUpgrade tasks for existing chains

PanGan21:
Backport of: https://github.com/zama-ai/fhevm/pull/3018

### DIFF
--- a/.github/workflows/host-contracts-upgrade-tests.yml
+++ b/.github/workflows/host-contracts-upgrade-tests.yml
@@ -141,7 +141,7 @@ jobs:
 
       - name: Bootstrap host contracts missing from UPGRADE_FROM_TAG
         working-directory: current-fhevm/host-contracts
-        run: npx hardhat task:deployEmptyProxiesProtocolConfigKMSGeneration
+        run: npx hardhat task:deployEmptyProxyForConfidentialBridge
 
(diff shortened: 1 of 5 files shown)


obatirou (approved):
Won't deploycBridge need also destEids and dstChainId in init?

PanGan21:
> Won't deploycBridge need also destEids and dstChainId in init?

Not required now, but would you like me to add it for the future in case it is needed?

obatirou:
> > Won't deploycBridge need also destEids and dstChainId in init?
> 
> Not required now, but would you like me to add it for the future in case it is needed?

If it makes sense for 0.14 yes, else this can be done in 0.15 or whenever this is needed

PanGan21:
> > > Won't deploycBridge need also destEids and dstChainId in init?
> > 
> > 
> > Not required now, but would you like me to add it for the future in case it is needed?
> 
> If it makes sense for 0.14 yes, else this can be done in 0.15 or whenever this is needed

It is not needed for 0.14 also not for 0.15, only if we are in a chain where we know those info before hand. There are tasks to call while wiring: task:setDstChainId and task:setPeer that will set those two afterwards.

PR #3038, opened 2026-07-07
chore(common): update contracts Charts

isaacdecoded:
Update the contracts' Chart.yaml for the v0.14.x


isaacdecoded:
These changes seem outdated already since the coming code freeze is for v0.15, and changes here are for v0.14.

PR #3045, opened 2026-07-07
ci(test-suite): add Solana LiteSVM coprocessor gate

Eikix:
## What
Adds a CI job `solana-tests/litesvm-coprocessor-gate` that runs the existing but `#[ignore]`d litesvm integration tests in `tfhe-worker/src/tests/solana_poc.rs` — real on-chain programs (litesvm, in-process) → host-listener adapter → real tfhe-worker → decrypt, asserting cleartext.

## Why
These tests are the mid-tier integration signal (ingest→compute→decrypt→ACL) that today only exists at the 40–70 min docker e2e. They're `#[ignore]`d only because CI provided neither Docker (for the testcontainers Postgres the harness self-provisions) nor built SBF programs — not because they're broken. Promoting them to a gate collapses that signal from ~40–70 min to a compile-bound job whose exec is seconds.

## How
- New job in `solana-tests.yml`, path-filtered on `solana/**` + the coprocessor test deps, so a Solana-only change triggers it.
- Installs Solana CLI + Anchor (cached), builds both PoC SBF programs (`check-zama-host-idl.sh` → `zama_host.so` + `confidential_token.so`), LFS-pulls the key fixtures, verifies Docker (`docker info`), then `cargo test -p tfhe-worker solana_poc -- --ignored --test-threads=1`.
- The tests stay `#[ignore]`d in source; the `--ignored` flag runs them only in this dedicated job, so the default `cargo test` stays DB/Docker-free. **No Postgres service container / external DATABASE_URL** — the harness spins its own Postgres via `testcontainers` (`test-harness/src/instance.rs`). Cargo caching so the heavy `tfhe` build is amortized.
- Local runner: `solana/scripts/test-litesvm-coprocessor-gate.sh` + `TESTING.md` (Docker + `anchor build` + the test command).

## Scope
Gate + local runner only. **Not** coverage expansion (new MMR-ACL test cases) — that depends on the rewrite PR #3002 and is a separate follow-up under the epic.

## Issues
Closes zama-ai/fhevm-internal#1655 — litesvm CI integration gate.
Part of zama-ai/fhevm-internal#1659 — Solana testing-strategy & feedback-loop epic.

## Validation
YAML valid; `bash -n` on the runner script passes; `zizmor` clean. Full compile/runtime path (anchor build + tfhe build + testcontainers) is confirmed by this PR's own CI (the new job's path filter includes `solana-tests.yml` + the coprocessor, so it runs here) — not run locally (heavy build / disk).

### DIFF
--- a/solana/scripts/test-litesvm-coprocessor-gate.sh
+++ b/solana/scripts/test-litesvm-coprocessor-gate.sh
@@ -12,4 +12,12 @@ cd "$ROOT/solana"
 NO_DNA=1 bash scripts/check-zama-host-idl.sh
 
 cd "$ROOT/coprocessor/fhevm-engine"
-NO_DNA=1 SQLX_OFFLINE=true cargo test -p tfhe-worker solana_poc -- --ignored --test-threads=1 --nocapture
+tests=(
+  solana_fhe_eval_replays_threshold_logs_from_litesvm_metadata
+  solana_fhe_rand_creates_ciphertext_and_decrypts
+  solana_user_decrypt_acl_invariants_match_evm_semantics
+)
+
+for test in "${tests[@]}"; do
+  NO_DNA=1 SQLX_OFFLINE=true cargo test -p tfhe-worker "$test" -- --ignored --test-threads=1 --nocapture
+done
--- a/solana/scripts/test-litesvm-coprocessor-gate.sh
+++ b/solana/scripts/test-litesvm-coprocessor-gate.sh
@@ -12,4 +12,4 @@ cd "$ROOT/solana"
 NO_DNA=1 bash scripts/check-zama-host-idl.sh
 
 cd "$ROOT/coprocessor/fhevm-engine"
-NO_DNA=1 SQLX_OFFLINE=true cargo test -p tfhe-worker solana_poc -- --ignored --test-threads=1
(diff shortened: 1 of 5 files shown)


solana/scripts/test-litesvm-coprocessor-gate.sh
```diff
@@ -0,0 +1,33 @@
...
+tests=(
+  solana_confidential_transfer_with_real_ciphertexts_computes_and_decrypts
+  solana_fhe_eval_replays_threshold_logs_from_litesvm_metadata
+  solana_fhe_rand_creates_ciphertext_and_decrypts
+  solana_trivial_encrypt_then_confidential_transfer_computes_and_decrypts
+  solana_user_decrypt_acl_invariants_match_evm_semantics
+)
+
+for test in "${tests[@]}"; do
+  cargo test --profile local -p tfhe-worker "$test" -- --ignored --test-threads=1 --nocapture
```
PanGan21, line 32: Is the `--ignored` flag correct here?
I understand that this means: A test is executed only if it matches the filter and currently carries `#[ignore]`

Eikix: the main reason was because a run required a anchor build and we couldnt add it in the cargo test directly, but not sure thats a good reason

Eikix: but something feels wrong in this PR. these intermediary tests arent like what I'd want. I might close the PR and redesign a better harness for fast tests

PR #3049, opened 2026-07-07
fix(relayer): serve keyurl dataId as hex

manoranjith:
Fixes [zama-ai/fhevm-internal#1663](https://github.com/zama-ai/fhevm-internal/issues/1663).

`dataId` now uses the same `0x`-prefixed 32-byte big-endian hex
encoding already used for the `urls` path segment, instead of a
decimal string. relayer-v2 got the equivalent fix in
[zama-ai/relayer#79](https://github.com/zama-ai/relayer/pull/79).

- [x] `cargo test --features integration-tests --test keyurl_poller_test --test keyurl_test` (locally: 6/6 pass)

### DIFF
--- a/relayer/src/host/keyurl_poller.rs
+++ b/relayer/src/host/keyurl_poller.rs
@@ -42,6 +42,12 @@ struct KmsStorageNode {
     storage_prefix: String,
 }
 
+/// The canonical 32-byte big-endian hex encoding of an on-chain key/CRS id, lowercase, no `0x`
+/// prefix. Used both as the URL path segment and, `0x`-prefixed, as the served `dataId`.
+fn id_hex(id: U256) -> String {
+    hex::encode(id.to_be_bytes::<32>())
+}
+
 /// Build the full object URL the KMS Core writes to:
 /// `{storage_url}/{storage_prefix}/{segment}/{id_hex}` (`segment` = `PublicKey`|`CRS`, `id_hex` =
 /// 32-byte big-endian id, lowercase, no `0x`). The getters return only `storage_url`.
@@ -50,13 +56,12 @@ struct KmsStorageNode {
 /// equivalent; the served response carries a single URL (the SDK requires exactly one), built
 /// from the first context node.
 fn build_object_url(node: &KmsStorageNode, segment: &str, id: U256) -> String {
-    let id_hex = hex::encode(id.to_be_bytes::<32>());
     format!(
         "{}/{}/{}/{}",
         node.storage_url.trim_end_matches('/'),
         node.storage_prefix,
         segment,
-        id_hex
+        id_hex(id)
     )
 }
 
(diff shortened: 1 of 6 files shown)


relayer/tests/common/utils.rs
```diff
@@ -1018,6 +1018,12 @@ pub fn test_keyurl_expected_url(segment: &str, id: u64) -> String {
     format!("{TEST_KEYURL_STORAGE_URL}/{TEST_KEYURL_STORAGE_PREFIX}/{segment}/{id_hex}")
 }
 
+/// The served `dataId` for a given on-chain id: `0x`-prefixed 32-byte big-endian, lowercase hex.
+#[allow(dead_code)]
+pub fn test_keyurl_expected_data_id(id: u64) -> String {
+    format!("0x{}", hex::encode(U256::from(id).to_be_bytes::<32>()))
```
zmalatrax: nit: the signature could be identical to the `id_hex(U256)` function, moving up the u64 conversion to u256 to the argument on function call

relayer/src/host/keyurl_poller.rs
```diff
@@ -42,6 +42,12 @@ struct KmsStorageNode {
     storage_prefix: String,
 }
 
+/// The canonical 32-byte big-endian hex encoding of an on-chain key/CRS id, lowercase, no `0x`
+/// prefix. Used both as the URL path segment and, `0x`-prefixed, as the served `dataId`.
+fn id_hex(id: U256) -> String {
+    hex::encode(id.to_be_bytes::<32>())
+}
```
zmalatrax, line 49: I'd have it inlined for 3 calls in 2 functions but I'm fine with having it refactored into an helper

zmalatrax (dismissed):
lgtm
Note that before dynamic key url, the data_id field was returning ascii string (and not decimal string): https://github.com/zama-zws/gitops/blob/eeb4f1c935595d461549049f8d740f42500fcfdc/values/relayer/relayer/values-relayer-testnet.yaml#L232-L237

manoranjith:
> lgtm
> Note that before dynamic key url, the data_id field was returning ascii string (and not decimal string): https://github.com/zama-zws/gitops/blob/eeb4f1c935595d461549049f8d740f42500fcfdc/values/relayer/relayer/values-relayer-testnet.yaml#L232-L237

Yes, this part i am aware. The string was a placeholder and unused. So i think it should be fine.

relayer/src/host/keyurl_poller.rs
manoranjith: Since, its more than 1 places, I would keep the helper, to avoid duplication.

relayer/tests/common/utils.rs
manoranjith: I agree. Fixed in https://github.com/zama-ai/fhevm/pull/3049/commits/35aa2486ba3c188a3f4ba6381e337327b7125300

zmalatrax (approved):

manoranjith:
@mergify queue

manoranjith:
@mergify dequeue

manoranjith:
@mergify refresh

manoranjith:
@mergify queue

PR #3050, opened 2026-07-07
feat(sdk): add support for cleartext legacy

0xalexbel:
- add support for cleartext v1 (forge-fhevm)
- ran successful tests on Hoodi + ethers 
- ran successful tests on T-Rex + ethers

### DIFF
--- a/sdk/js-sdk/contracts/src/FHETest.sol
+++ b/sdk/js-sdk/contracts/src/FHETest.sol
@@ -211,7 +211,7 @@ contract FHETest {
     // optimizing away the computation (e.g. detecting a ^ a = 0).
     function _zeroEuint8() internal returns (euint8) {
         euint8 r = FHE.randEuint8();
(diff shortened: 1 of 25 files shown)


poppyseedDev (approved):

sdk/js-sdk/contracts/src/FHETest.sol
```diff
@@ -211,7 +211,7 @@ contract FHETest {
     // optimizing away the computation (e.g. detecting a ^ a = 0).
     function _zeroEuint8() internal returns (euint8) {
         euint8 r = FHE.randEuint8();
-        euint8 zeroOrOne = FHE.shr(r, 1);
+        euint8 zeroOrOne = FHE.shr(r, 7);
```
geoxel, line 214: Next time, could we create a separate commit for this kind of fix? That way, the fix would not be lost in a large implementation commit :-)

PR #3027, opened 2026-07-06
test(test-suite): cover the contract-delegate trust model in unified user decryption

maksymsur:
## Summary

Adds two e2e tests (and one mock method) pinning what happens when a user delegates decryption rights to a **contract** rather than an EOA: the delegate contract's `isValidSignature` policy — not the on-chain delegation — decides who can wield the delegated access. Answers a question raised in review of the ERC-1271 / unified user-decryption suite.

## What's added

In `test/unifiedUserDecryption/` (`delegated legs`), reusing the existing SmartWallet→ERC-1271-wallet delegation setup:

- **`contract delegate re-points its delegated access by rotating its signer`** — the ERC-1271 delegate wallet rotates its owner (bob → eve): the old owner is rejected by the signature pre-check (`400`), the new owner decrypts the delegator's handle — with no new on-chain delegation.
- **`revoking a contract delegate closes access for its current signer`** — the delegator revokes the delegation: even the wallet's current valid signer is rejected (`failed` / `not_allowed_on_host_acl`). The signature still passes ERC-1271; the ACL check is what closes it.

Fixture: `ERC1271OwnerWallet` gains an owner-gated `transferOwnership`. Both tests use fresh re-encryption keys (the relayer dedups on `publicKey`, among other fields) so each is a real authorization round, not a cache hit — documented inline.

Test files only — no protocol, SDK, relayer, or connector changes.

PR #3054, opened 2026-07-08
test(test-suite): e2e for the contract-delegate trust model in unified user decryption

maksymsur:
## Summary

Adds two e2e tests (and one mock method) pinning what happens when a user delegates decryption rights to a **contract** rather than an EOA: the delegate contract's `isValidSignature` policy — not the on-chain delegation — decides who can wield the delegated access. Answers a question raised in review of the ERC-1271 / unified user-decryption suite.

## What's added

In `test/unifiedUserDecryption/` (`delegated legs`), reusing the existing SmartWallet→ERC-1271-wallet delegation setup:

- **`contract delegate re-points its delegated access by rotating its signer`** — the ERC-1271 delegate wallet rotates its owner (bob → eve): the old owner is rejected by the signature pre-check (`400`), the new owner decrypts the delegator's handle — with no new on-chain delegation.
- **`revoking a contract delegate closes access for its current signer`** — the delegator revokes the delegation: even the wallet's current valid signer is rejected (`failed` / `not_allowed_on_host_acl`). The signature still passes ERC-1271; the ACL check is what closes it.

Fixture: `ERC1271OwnerWallet` gains an owner-gated `transferOwnership`. Both tests use fresh re-encryption keys (the relayer dedups on `publicKey`, among other fields) so each is a real authorization round, not a cache hit — documented inline.

Test files only — no protocol, SDK, relayer, or connector changes.


manoranjith (approved):
Lgtm.

( Re-creation of #3027 due to stuck CI Job)

PR #3056, opened 2026-07-08
chore(common): cut merge-queue CI latency via e2e sharding, bigger amd64 runners and queue batching

immortal-tofu:
Measured on the merge queue (July 7 runs): 98-130 min per PR, gated by a 61-66 min serial e2e job, 20-min amd64 Rust builds on 2-vCPU runners, and up to 26 min of large_ubuntu_32 starvation caused by the non-gating threshold-KMS legs. batch_size 1 caps throughput at ~14 merges/day.

- Shard the standard e2e suite into three parallel legs (stateful/decryption/compute) that exactly partition STANDARD_TEST_PROFILES (enforced by a unit test); all three gate the queue, and `run-e2e-tests` keeps its check name on the compute shard.
- Run all Rust image builds on 16cpu amd64 runners (runs-on-amd input already existed; arm64 was already 16 vCPU).
- Move the three threshold-KMS burn-in legs off merge-queue refs to a new post-merge workflow on main pushes - same per-merge cadence, no runner contention with the gating shards.
- Run github-script-only jobs (build-decisions, create-e2e-tests-input) on ubuntu-latest to drop provisioned-runner spin-up from the critical path.
- Mergify: batch_size 3 (batch failures auto-bisect, so gating protection is unchanged).

Expected: ~55-65 min worst-case queue cycle and ~3x peak throughput with identical coverage. Validate by labeling a PR `e2e orchestrate test` and rebalancing shards from observed per-profile timings.


tawadaa (approved):
LGTM, thanks 🙏🏼

immortal-tofu:
@mergifyio queue

PR #3062, opened 2026-07-08
perf(coprocessor): compile docker images through sccache backed by the CI S3 cache

immortal-tofu:
## What

Compile the 6 heavy coprocessor docker images (host-listener, gw-listener, tfhe-worker, sns-worker, zkproof-worker, transaction-sender) through **sccache** backed by the CI S3 cache bucket, so compiled dependency artifacts (tfhe & friends) are shared **across images, platforms, and runs**.

## Why

Today any source change recompiles the entire tfhe-heavy workspace from scratch in each image, per platform, per prod/dev variant (measured: host-listener 10.7 min amd64 / 16.3 min arm64 per merge-queue run with coprocessor changes):

- The Dockerfiles' buildkit cache mounts (`target/`, cargo registry) never survive CI — each run gets a fresh buildkit instance.
- The S3 layer cache only stores whole layers, and the single cargo-build layer is invalidated by any source edit.

A cargo-chef split was attempted first (see earlier commits' CI runs) but fought this workspace twice — out-of-workspace path deps, then dev-dependencies requiring a newer rustc at `cook` time. sccache caches at the rustc-invocation level and is indifferent to workspace shape.

## How

- Pinned `sccache 0.8.2` installed in the builder stage (layer rebuilds only on base-image change).
- The build routes through sccache **only when CI provides S3 credentials** as buildkit secrets (`SCCACHE_AWS_ACCESS_KEY_ID` / `SCCACHE_AWS_SECRET_ACCESS_KEY`); without them it is a plain cargo build — local builds and current CI behave exactly as on main.
- Shared cache prefix `sccache/fhevm-coprocessor` in the existing `gh-actions-cache-eu-west-3` bucket; sccache keys include compiler/target/flags so amd64/arm64 coexist safely.
- `sccache --show-stats` prints hit rates at the end of each build for verification.

## Activation requires a 2-line ci-templates change

`common-docker.yml`'s build-push-action step must pass the two secrets (reusing the same AWS secrets it already uses for the buildx layer cache), then the pinned ci-templates SHA gets bumped in the fhevm caller workflows:

```yaml
          secrets: |
            BLOCKCHAIN_ACTIONS_TOKEN=${{ secrets.BLOCKCHAIN_ACTIONS_TOKEN }}
            SCCACHE_AWS_ACCESS_KEY_ID=${{ secrets.AWS_ACCESS_KEY_S3_USER }}
            SCCACHE_AWS_SECRET_ACCESS_KEY=${{ secrets.AWS_SECRET_KEY_S3_USER }}
```

Until that lands this PR is a no-op in CI (safe to merge in either order).

## Expected effect (once activated)

- First run populates the cache (the six images miss in parallel — limited benefit).
- From the second run on, dependency compiles are cache hits; only workspace crates + linking remain. Rough expectation: host-listener amd64 toward ~4–6 min.
- Known ceiling: sccache cannot cache **linking**, and `lto = "fat"` in the release profile makes the link the dominant leftover — switching to thin LTO (needs a runtime benchmark, separate decision) is the next lever.

## Validation

Label the PR `e2e orchestrate test` to trigger the image builds. Before ci-templates passes the secrets, expect timings identical to main and no `sccache` output in build logs; after activation, check the `sccache --show-stats` block (second run should show a high hit rate).

### DIFF
--- a/coprocessor/fhevm-engine/gw-listener/Dockerfile
+++ b/coprocessor/fhevm-engine/gw-listener/Dockerfile
@@ -19,6 +19,10 @@ ARG CARGO_PROFILE=release
 
 USER root
 
+# sccache, pinned for reproducibility; this layer only rebuilds when the base
+# image changes.
+RUN cargo install sccache --locked --version 0.8.2
+
 WORKDIR /app
 
 COPY listener ./listener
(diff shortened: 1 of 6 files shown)


immortal-tofu:
Closed in favor of https://github.com/zama-ai/fhevm/pull/3071

PR #3073, opened 2026-07-09
feat(kms-connector): forward context and epoch destruction to core

eudelins-zama:
### Summary

This PR makes the KMS Connector listen to KMS context and epoch destruction on the Gateway and forward them to the KMS Core.

####  gw-listener

- Subscribes to `KmsContextDestroyed` and `KmsEpochDestroyed` from ProtocolConfig, and stores them in the two new `kms_context_destroyed` / `kms_epoch_destroyed` tables (with the usual NOTIFY triggers to wake up the kms-worker).

####  kms-worker

- Picks up the destruction events and forwards them to the Core. As for key-management operations, no response is expected from the Core (`NoResponseExpected`), so there is no result-polling step.
- `epoch_ids` is left empty in `DestroyMpcContextRequest` for now, as the Core's gRPC interface for context/epoch destruction is about to change (see TODO in `processor.rs`).

####  Validation cache rework

The former `kms_context(id, epoch_id)` cache has been split into:
- `kms_context(id, is_valid)`
- `kms_epoch(id, context_id, is_valid)`

This was needed to distinguish a destroyed context from a context with a not-cached-yet epochs while scanning the old `kms_context` table (with `(id, epoch_id)` as `PRIMARY KEY`).

Request validation (`DbContextManager`) checks both tables in a single query on the hot path, falls back to `ProtocolConfig (isValidKmsContext / isValidEpochForContext)` for uncached pairs, and caches valid pairs.

Destruction events mark the corresponding rows as is_valid = FALSE, so subsequent requests referencing a destroyed context/epoch are rejected as irrecoverable without an RPC call.
A stale epoch→context association found in the cache is not trusted: it triggers an on-chain re-check and is repaired on success.

#### Misc

- Added a sqlx offline-cache regeneration command.

Closes https://github.com/zama-ai/fhevm-internal/issues/1660


kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
```diff
@@ -230,6 +231,21 @@ impl<GP: Provider + Clone + 'static, HP: Provider, C: ContextManager> DbEventPro
                     .prepare_new_kms_epoch_request(req)
                     .await
             }
+            ProtocolEventKind::KmsContextDestroyed(req) => {
+                // TODO: the `epoch_ids` field is left empty for now, as the gRPC interface of
+                // the KMS Core for context/epoch destruction is about to change.
```
eudelins-zama, line 236: Needs https://github.com/zama-ai/kms-internal/issues/3079 to be handled first

eudelins-zama: Wrote in the comment itself to be addressed in a follow up

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -253,11 +241,13 @@ where
...
             let span = info_span!("handle_ethereum_event", event = %event_kind);
             let otlp_ctx = PropagationContext::inject(&span.context());
-            actions.push(EthereumEventAction::StoreEvent(Box::new(
-                ProtocolEvent::new(event_kind, log.transaction_hash, otlp_ctx),
-            )));
+            events.push(ProtocolEvent::new(
+                event_kind,
+                log.transaction_hash,
+                otlp_ctx,
+            ));
```
dvdplm, line 248: nice cleanup

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -413,23 +432,27 @@ pub async fn publish_context_and_epoch(
...
+    .execute(&mut *tx)
     .await?;
+    tx.commit().await?;
 
-    if query_result.rows_affected() == 1 {
-        info!("KMS context #{context_id} (epoch #{epoch_id}) was successfully published!");
-    } else {
-        debug!("KMS context #{context_id} (epoch #{epoch_id}) was not published: {query_result:?}");
-    }
+    info!("KMS context #{context_id} (epoch #{epoch_id}) was successfully published!");
```
dvdplm: to me this is a `debug` but I know y'all like your  info logs so... ;)

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
```diff
@@ -467,6 +475,48 @@ impl KmsClient {
...
+        send_request_with_retries(
+            self.grpc_request_retries,
+            || {
+                let mut client = inner_client.clone();
+                let request = request.clone();
+                async move { client.destroy_mpc_context(request).await }
+            },
+            EventType::KmsContextDestroyed,
+        )
+        .await
```
dvdplm, line 496: If the KMS fails here, won't the connector already have removed the context? Or does the connector wait  until this call has succeeded  before invalidating?

My worry is the possibility that blockchain, connector and KMS get out of sync. If they ever do, the blockchain's data is what holds so I guess worst case we can re-send the destruction event manually and force the connector/KMS to behave? Or?

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
dvdplm: Wow, that's a big one.

dvdplm (dismissed):
Looks sane to me.

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: KMS Context will happen very rarely. Definitely want an `info!` on this! :smile:

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
eudelins-zama: > If the KMS fails here, won't the connector already have removed the context? Or does the connector wait until this call has succeeded before invalidating?

Indeed, but I don't think it's an issue.
Ethereum is the source of truth for protocol management, so it's fine to have context invalidated at connector level before it's the case at KMS level IMO.
But if KMS fails, depending on error code returned, the kms-connector should retry to forward the destruction request to the KMS!

eudelins-zama: > My worry is the possibility that blockchain, connector and KMS get out of sync. If they ever do, the blockchain's data is what holds so I guess worst case we can re-send the destruction event manually and force the connector/KMS to behave? Or?

Yes indeed, something like this.

kms-connector/connector-db/migrations/20260708000000_add_context_epoch_destruction.sql
```diff
@@ -0,0 +1,102 @@
+-- Split the `kms_context` validation cache into context-level and epoch-level tables.
+--
+-- Otherwise, it was not possible to distinguish a destroyed context from a context with
+-- not-cached-yet (context, epoch) pair.
+CREATE TABLE IF NOT EXISTS kms_epoch (
+    id BYTEA NOT NULL,
+    -- NULL for invalidation written from `KmsEpochDestroyed`, which only carries the epoch ID.
+    context_id BYTEA,
+    is_valid BOOLEAN NOT NULL DEFAULT TRUE,
```
jot2re: Just want to check that the default value should indeed be true. I not completely sure in case the context has been deleted. (But not completely sure about the entire flow yet, so just mentioning for completeness)

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -54,6 +56,17 @@ pub async fn publish_batch(
 ) -> anyhow::Result<()> {
     let mut tx = db_pool.begin().await?;
     for event in events {
+        // Destruction events also invalidate the local validation cache
+        match &event.kind {
+            ProtocolEventKind::KmsContextDestroyed(e) => {
+                invalidate_kms_context(&mut *tx, e.kmsContextId).await?
```
jot2re, line 62: Should there be an invalidation of associated epochs? 
If needed, I can have the KMS core return a list of deleted epoch ids once a context gets destroyed.

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -54,6 +56,17 @@ pub async fn publish_batch(
...
+            }
+            ProtocolEventKind::KmsEpochDestroyed(e) => {
+                invalidate_kms_epoch(&mut *tx, e.epochId).await?
+            }
+            _ => {}
```
jot2re, line 67: Nitpick: would it make sense to make the other types explicit here s.t. the compiler will complain and force you to consider the cause of action when you add a new type of event?

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
```diff
@@ -67,37 +72,52 @@ impl<P: Provider> DbContextManager<P> {
...
-            // v1 extra_data carries no epoch: any valid row proves the context exists
-            None => !rows.is_empty(),
-            Some(epoch_id) => rows.iter().any(|r| r.epoch_id == epoch_id.as_le_slice()),
-        };
-        if known {
-            Ok(LocalCheck::Valid)
-        } else {
-            Ok(LocalCheck::Unknown)
+        if epoch_id.is_none() {
+            // v1 extra_data carries no epoch: the context row alone concludes
```
jot2re, line 106: What about the situation where we destroy a context. Won't we want to ensure the associated epochs are considered invalid? (I am not completely sure what is the best way to handle this, if we are fine just doing it on the KMS core alone)

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
```diff
@@ -141,6 +144,11 @@ impl KmsClient {
                 Err(e) => (0, Err(e)),
             },
             KmsGrpcRequest::NewMpcEpoch(req) => self.poll_epoch_result_response(req).await,
+            // Like aborts, destructions have no result-polling endpoint on the Core: the
```
jot2re, line 147: Nitpick: maybe clarify that there is an empty 200 OK HTTP response.

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
```diff
@@ -230,6 +231,22 @@ impl<GP: Provider + Clone + 'static, HP: Provider, C: ContextManager> DbEventPro
                     .prepare_new_kms_epoch_request(req)
                     .await
             }
+            ProtocolEventKind::KmsContextDestroyed(req) => {
+                // TODO: the `epoch_ids` field is left empty for now, as the gRPC interface of
```
jot2re, line 235: Nitpick: should there be a link to a follow up issue to ensure it gets remembered?

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
```diff
@@ -67,37 +72,52 @@ impl<P: Provider> DbContextManager<P> {
...
         &self,
         context_id: U256,
         epoch_id: Option<U256>,
     ) -> Result<LocalCheck, RequestCheckError> {
-        let rows = sqlx::query!(
-            "SELECT epoch_id, is_valid FROM kms_context WHERE id = $1",
-            context_id.as_le_slice()
+        // No row is returned when the context is not cached, and the `epoch_*` columns are
+        // `NULL` when the epoch is not cached.
+        let row = sqlx::query!(
```
jot2re, line 84: Claude pointed out that it looks like a destroyed epoch is ignored in the situation where the context is not cached.

kms-connector/crates/gw-listener/src/core/publish.rs
jot2re: Note that an info log is always made regardless of any rows being affected

jot2re (changes requested):
Great work! Mainly some small nitpick comments, and some suggestions on interaction with the KMS Core.

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Could be nice indeed!

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: I like the idea in general, but in that very specific case I would suggest not too.

In such place, doing something is the exception, not the norm. So it would hurt the readability of the `match` for small benefits.

kms-connector/connector-db/migrations/20260708000000_add_context_epoch_destruction.sql
eudelins-zama: Like it. Will fix :slightly_smiling_face:

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: Not sure I get it.

These [lines](https://github.com/zama-ai/fhevm/blob/59fac33db069ab30d1487b8988e9ff7154fefea0/kms-connector/crates/kms-worker/src/core/event_processor/context.rs#L102-L104) would check the destroyed context case.

Then, I don't see the point of also checking the associated epochs are considered invalid tbh. (aside from double checking correctness of DB state?)

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
eudelins-zama: Would be confusing IMO.

200 OK HTTP response is for the "request" step.
This comment concern the "result-polling" step.

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
eudelins-zama: There's already a link to an issue: https://github.com/zama-ai/fhevm/blob/59fac33db069ab30d1487b8988e9ff7154fefea0/kms-connector/crates/kms-worker/src/core/event_processor/processor.rs#L237 (I subscribed to it)

Not sure another issue would help more?

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
eudelins-zama: When the context is not cached, the connector will perform an on-chain [check](https://github.com/zama-ai/fhevm/blob/59fac33db069ab30d1487b8988e9ff7154fefea0/kms-connector/crates/kms-worker/src/core/event_processor/context.rs#L125)

But yes, from the scope of `check_local_db`, a destroyed context not cached is ignored, it can't do much more in its scope :sweat_smile:

kms-connector/connector-db/migrations/20260708000000_add_context_epoch_destruction.sql
eudelins-zama: Fixed 4d2f0b5812a53de14870ed1f1adf8c168153a024

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: Fixed 6c14caf12d2e15f2a84f916a28b79d47a4c4e800

kms-connector/crates/gw-listener/src/core/publish.rs
jot2re: Great! I will add it to this [PR](https://github.com/zama-ai/kms/pull/696)

kms-connector/crates/kms-worker/src/core/event_processor/context.rs
jot2re: OK, I think I misunderstood the full goal of this method

kms-connector/crates/kms-worker/src/core/event_processor/kms_client.rs
jot2re: Fair enough

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -452,10 +452,20 @@ pub async fn publish_context_and_epoch(
...
+    log_cache_insert_outcome(&context_result, &format!("KMS context #{context_id}"));
+    log_cache_insert_outcome(&epoch_result, &format!("KMS epoch #{epoch_id}"));
     Ok(())
 }
 
+/// Logs the outcome of a single-row insert in the context/epoch validation cache.
+fn log_cache_insert_outcome(query_result: &PgQueryResult, entity: &str) {
+    match query_result.rows_affected() {
+        1 => info!("{entity} was successfully published!"),
+        0 => info!("{entity} was already in the validation cache"),
```
jot2re, line 464: Nitpick: should this be debug  as there is not really any state change?

jot2re (approved):
LGTM! One nitpick comment; feel free to ignore

kms-connector/crates/gw-listener/src/core/publish.rs
eudelins-zama: This call is only made once at connector startup. I guess it's OK to have an `info!` log here, even if `debug!` would be fine as well :slightly_smiling_face:

PR #3075, opened 2026-07-09
feat(sdk): accept ebool/euintN type aliases in encrypt values

geoxel:
Solidity developers think in terms of ebool/euint8/.../eaddress, not the
SDK's internal clear-value type names (bool/uint8/.../address). Add
resolveRawValueTypeName() to FheType.ts, which maps FHE type aliases to
their clear ValueTypeName counterpart (passing other type names through
unchanged), and call it from encryptValue(), encryptValues(), and
generateZkProof() before building each TypedValue — so callers can pass
either naming convention in `type`.


0xalexbel (approved):
LGTM!

PR #3082, opened 2026-07-10
feat(coprocessor): eager per-block anchor consensus across host + Gateway-inputs tracks

goshawk-3:
fixes https://github.com/zama-ai/fhevm-internal/issues/1619
fixes https://github.com/zama-ai/fhevm-internal/issues/1600

### DIFF
--- a/coprocessor/fhevm-engine/consensus-detector/src/lib.rs
+++ b/coprocessor/fhevm-engine/consensus-detector/src/lib.rs
@@ -496,9 +496,12 @@ where
 /// Sealed Gateway blocks for which this operator has already produced a local
 /// `state_hash` row — the candidate set for the Gateway consensus poll. Bounded
 /// to `[gw_start, gw_tip)` so only sealed blocks are polled, and capped to the
-/// earliest [`MAX_ANCHOR_CANDIDATES`]. A block without a local hash can never
-/// reach unanimity (our own slot would stay empty), so restricting to
-/// locally-produced rows avoids polling peers for nothing.
+/// earliest [`MAX_ANCHOR_CANDIDATES`]. The poll reads every operator's S3 blob,
+/// including our own; our own object is uploaded only for blocks we've hashed
+/// locally (see `upload_pending_gw_state_hashes`). So a block with no local
+/// `state_hash` row is one we'll never upload — our own slot stays empty and
+/// unanimity is unreachable — hence restricting candidates to locally-produced
+/// rows avoids polling peers for a block we ourselves can't contribute to.
 async fn pending_gw_consensus_blocks(
     pool: &Pool<Postgres>,
     gw_chain_id: i64,
--- a/coprocessor/fhevm-engine/consensus-detector/src/lib.rs
+++ b/coprocessor/fhevm-engine/consensus-detector/src/lib.rs
@@ -49,6 +49,9 @@ use crate::state_hash::state_hash_key;
 /// pg_notify channels this service listens on.
 pub const NEW_BLOCK_CHANNEL: &str = "event_new_block";
 pub const NEW_OPERATOR_ADDED_CHANNEL: &str = "event_new_operator_added";
+/// Fired by gw-listener on each ingested Gateway block (payload = the new
+/// Gateway tip). Drives the per-Gateway-block consensus track.
+pub const GW_NEW_BLOCK_CHANNEL: &str = fhevm_engine_common::gcs_activation::EVENT_GW_NEW_BLOCK;
 
 /// pg_notify channels this service emits.
 ///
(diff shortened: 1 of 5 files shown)


coprocessor/fhevm-engine/consensus-detector/src/lib.rs
```diff
@@ -49,14 +54,29 @@ use crate::state_hash::state_hash_key;
...
+/// Fired by gw-listener on each ingested Gateway block (payload = the new
+/// Gateway tip). Drives the per-Gateway-block consensus track.
+pub const GW_NEW_BLOCK_CHANNEL: &str = fhevm_engine_common::gcs_activation::EVENT_GW_NEW_BLOCK;
 
 /// pg_notify channels this service emits.
 ///
 /// `unanimity_consensus` is consumed only by the upgrade procedure — no other
 /// service should treat it as a generic "agreement" signal.
 pub const UNANIMITY_CONSENSUS_CHANNEL: &str = "event_unanimity_consensus";
+/// Reserved channel for a future timeout/abort path — not emitted today.
```
PanGan21, line 66: This is big change to stop emitting `event_unanimity_consensus_timeout`. How come this was removed? Will it be done from another ticket?

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
```diff
@@ -227,134 +247,178 @@ pub(crate) async fn active_upgrade_window(
...
+            .or_insert_with(|| vec![None; urls.len()]);
+        // Operator-set size changed (new_operator_added) — restart this block's
+        // slots so indices line up with the current URL list.
+        if slots.len() != urls.len() {
+            *slots = vec![None; urls.len()];
         }
-
-        if confirmed.len() == window_size {
+        fetch_state_commitments(http, urls, track.chain_id, block, slots).await;
+        if all_some_and_identical(slots) {
```
PanGan21, line 318: This only logs when operators agree. When they respond but disagree  (the thing the dry run is meant to catch) nothing is logged, so a stuck upgrade looks identical to still waiting. Maybe adding a warning here is useful

coprocessor/fhevm-engine/consensus-detector/src/state_hash.rs
```diff
@@ -148,6 +149,138 @@ async fn compute_gcs_hash(
     Ok(state_hash)
 }
 
+/// GCS Gateway-inputs path: stamp the GCS `state_hash` (keyed by `gw_chain_id`)
+/// for **sealed, complete** Gateway blocks in `[gw_start_block, gw_tip)` that
+/// carry input ciphertexts and don't already have an entry.
+///
+/// A Gateway block is a candidate when:
+///   * it has ≥1 `input_handles` row (blocks with no inputs are skipped entirely
+///     — proofs are sparse, so unlike the L1 path we emit no sentinel);
```
PanGan21, line 158: Is it possible that we have FHE ops but no input? I suggest that we handle both the same way by emitting a timeout! Do you see any race condition on that?

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
```diff
@@ -227,134 +247,178 @@ pub(crate) async fn active_upgrade_window(
...
+    pool: &Pool<Postgres>,
+    host_chain_id: i64,
+    start: i64,
+    end: i64,
+) -> Result<Vec<i64>, Error> {
+    let sql = format!(
+        "SELECT sh.block_number
+           FROM {GCS_SCHEMA_QUOTED}.state_hash sh
+          WHERE sh.chain_id = $1
+            AND sh.block_number > $2 AND sh.block_number <= $3
```
PanGan21: This excludes `start_block (> $2)`, but the rest of the code treats the window as inclusive [start, end]

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
goshawk-3: It's out of scope for this PR. We can have another PR that fully implements  the `event_unanimity_consensus_timeout` emitting and handling.

coprocessor/fhevm-engine/consensus-detector/src/state_hash.rs
goshawk-3: Yes, that is possible. We should account for this when implementing the `timeout` handling.

Context: We need to ensure that the Gateway inputs are exercised during the evaluation window.

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
goshawk-3: Good catch!

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
PanGan21: But this pr removes it, no?

PanGan21: We can add it back in another pr if there is a ticket about it, or we should include it in this one: https://github.com/zama-ai/fhevm-internal/issues/1528

goshawk-3: It's removed as it's unused currently.  Let's add it as note in the https://github.com/zama-ai/fhevm-internal/issues/1528, if needed

PanGan21: Done, added a ref in the ticket

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
goshawk-3: Fixed - changed the host candidate predicate to `sh.block_number >= $2` so the window is inclusive `[start, end]`, matching how `start_block` is used elsewhere (BCS settles up to `start_block`, GCS prunes strictly below it, so `start_block` is the first GCS-computed block and a valid anchor). Updated the doc comments too.

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
goshawk-3: Good point - added a `warn!` in `poll_track` for the all-responded-but-divergent case (via a new `all_slots_filled` helper that separates it from still-waiting-on-a-slow-operator). It logs once per block (state hashes are deterministic, so a divergence is permanent) to avoid spamming every poll tick, and keeps polling the other candidates.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -980,64 +990,112 @@ pub async fn handle_unanimity_consensus(
...
+                    block_height = payload.block_height,
+                    "event_unanimity_consensus: gateway-track unanimity — setting gw_consensus_reached"
+                );
+                sqlx::query(
+                    "UPDATE upgrade_state SET gw_consensus_reached = TRUE, updated_at = NOW()
+                      WHERE stack_role = 'GCS' AND state = 'DryRunStarted'",
+                )
+                .execute(pool)
+                .await?;
+            }
```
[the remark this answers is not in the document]
goshawk-3: Fixed.

PanGan21 (approved):

PR #3084, opened 2026-07-10
ci(test-suite): report Solana native unit coverage

Eikix:
## What

Publish component-level native Rust line coverage for the Solana workspace in the Actions job summary and document the equivalent local command.

## Why

A single floor would be misleading: Mollusk executes prebuilt SBF artifacts that LLVM cannot instrument, while current native coverage ranges from 98.38% for `zama-solana-acl` to 6.05% for `confidential-token`.

## Validation

`cargo llvm-cov --workspace --exclude zama-solana-runtime-tests --json --summary-only --output-path /tmp/solana-native-coverage-final.json` passed in 6.00s warm and measured 5,031/11,381 lines (44.21%); the component-table `jq` parser and workflow YAML were validated locally.

## Boundary

This first signal is informational, excludes the runtime-test harness to avoid inflated attribution, and leaves host-listener and relayer for separate workspace-specific measurement lanes.

## Tracking

Tracks [zama-ai/fhevm-internal#1657](https://github.com/zama-ai/fhevm-internal/issues/1657).

### DIFF
--- a/.github/workflows/solana-tests.yml
+++ b/.github/workflows/solana-tests.yml
@@ -102,6 +102,78 @@ jobs:
       - name: Run Clippy
         run: cargo clippy --workspace --all-targets -- -D warnings
 
+  native-coverage:
+    name: solana-tests/native-coverage
+    needs: check-changes
+    if: ${{ github.event_name == 'workflow_dispatch' || needs.check-changes.outputs.changes-solana == 'true' }}
+    permissions:
+      contents: 'read'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: solana
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: 'false'
+
+      - name: Setup Rust toolchain file
+        run: cp rust-toolchain.toml ..
+
+      - uses: dsherret/rust-toolchain-file@3551321aa44dd44a0393eb3b6bdfbc5d25ecf621 # v1
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@a37010ded18ff788be4440302bd6830b1ae50d8b # v2.68.25
+        with:
+          tool: cargo-llvm-cov
+
+      - name: Cache cargo
+        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            solana/target/llvm-cov-target
+          key: ${{ runner.os }}-cargo-solana-coverage-${{ hashFiles('solana/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-solana-coverage-
+
+      - name: Measure native unit coverage
+        run: |
+          cargo llvm-cov \
+            --workspace \
+            --exclude zama-solana-runtime-tests \
+            --json \
+            --summary-only \
+            --output-path /tmp/solana-native-coverage.json
+
+          {
+            echo '### Solana native unit coverage'
+            echo
+            echo 'Mollusk executes prebuilt SBF programs, so it validates runtime behavior but cannot attribute LLVM coverage to program instruction bodies. The runtime-test harness is excluded here to avoid inflating this native signal.'
+            echo
+            echo '| Component | Covered lines | Coverage |'
+            echo '| --- | ---: | ---: |'
+            jq -r '
+              def row($label; $path):
+                [.data[0].files[] | select(.filename | contains($path)) | .summary.lines] as $parts
+                | ($parts | map(.count) | add) as $total
+                | ($parts | map(.covered) | add) as $covered
+                | if ($total // 0) == 0 then
+                    error("no coverage entries found for \($label)")
+                  else
+                    "| \($label) | \($covered)/\($total) | \((($covered * 10000 / $total) | round) / 100)% |"
+                  end;
+              [
+                row("zama-solana-acl"; "/crates/zama-solana-acl/"),
+                row("solana-ed25519-instruction"; "/crates/solana-ed25519-instruction/"),
+                row("zama-fhe"; "/crates/zama-fhe/"),
+                row("zama-host"; "/programs/zama-host/"),
+                row("confidential-token"; "/programs/confidential-token/"),
+                row("confidential-deposit-app"; "/programs/confidential-deposit-app/")
+              ][]
+            ' /tmp/solana-native-coverage.json
+          } >> "$GITHUB_STEP_SUMMARY"
+
   # The off-chain reconstruction path (host-listener, solana-reconstruct feature) is otherwise
   # only built/exercised by the full solana-e2e job. Compile it here and run its unit tests so a
   # break (e.g. the fhe_eval ACL-record account-offset) is caught in minutes, not 40+.
--- a/.github/workflows/solana-tests.yml
+++ b/.github/workflows/solana-tests.yml
@@ -154,23 +154,35 @@ jobs:
             echo '| Component | Covered lines | Coverage |'
             echo '| --- | ---: | ---: |'
             jq -r '
-              def row($label; $path):
-                [.data[0].files[] | select(.filename | contains($path)) | .summary.lines] as $parts
+              def components: [
+                {label: "zama-solana-acl", path: "/crates/zama-solana-acl/"},
+                {label: "solana-ed25519-instruction", path: "/crates/solana-ed25519-instruction/"},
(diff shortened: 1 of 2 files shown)


Eikix:
_Automated review — generated by a Claude (Fable) review agent dispatched by the PR author; findings verified against `origin/feature/solana` and the actual CI run._

---

## Overview

Adds a `native-coverage` job to `.github/workflows/solana-tests.yml` that runs `cargo llvm-cov --workspace --exclude zama-solana-runtime-tests --json --summary-only` in the `solana/` workspace and renders a per-component line-coverage table (six components: 3 crates, 3 programs) into `$GITHUB_STEP_SUMMARY` via a jq script. Also documents the equivalent local command and the rationale in `solana/docs/TESTING.md`. Deliberately informational: no coverage floor, runtime-test harness excluded, host-listener/relayer deferred.

The `solana-tests/native-coverage` job **passed** on the PR (29s warm, cache hit from the retrigger commit; the full test suite visibly ran under llvm-cov and the report was produced).

## Correctness

**jq parsing — verified against the llvm-cov export schema and the actual workspace layout; it is correct and defensively solid:**

- Schema usage (`.data[0].files[].summary.lines.{count,covered}`, `.data[0].totals.lines.count`) matches cargo-llvm-cov's `--json --summary-only` output.
- The six hard-coded component paths exactly partition the workspace: every file under `solana/crates/` and `solana/programs/` on the base branch falls inside the six prefixes, and the prefixes are mutually disjoint (no double counting).
- **Missing crate** → `line_parts` empty → `$total` null → `($total // 0) == 0` → `error(...)`, jq exits non-zero, step fails (GitHub's default `bash -e` shell). Good.
- **Zero-line crate** → same error branch, so no division by zero.
- **New/unlisted crate or stray file** → caught by the totals reconciliation (`$selected_total != $workspace_total` → error). This is the third commit's fix and it's the right failure-loud design: the table can never silently under-report.
- `--exclude zama-solana-runtime-tests` matches the actual package name in `solana/runtime-tests/Cargo.toml` — verified. The exclusion rationale (Mollusk runs prebuilt SBF artifacts LLVM can't instrument; including the harness would count test-harness Rust and inflate the number) is sound and is honestly documented in both the summary preamble and TESTING.md. The Mollusk suites still run in `build-and-test`'s `cargo test --workspace`, so no test execution is lost.
- Percent math `(covered*10000/total | round)/100` is fine (float division in jq; the zero-total case is already errored out).

**Shell/injection safety:** the jq program is a static single-quoted string, no GitHub context expressions are interpolated into `run:` bodies, `$GITHUB_STEP_SUMMARY` is quoted, `persist-credentials: false`, `permissions: contents: read`. No injection surface.

**Workflow logic:** `needs: check-changes` + the `workflow_dispatch || changes-solana` gate exactly mirrors the sibling jobs; `cp rust-toolchain.toml ..` mirrors `build-and-test`. Cache paths are workspace-root-relative (`solana/target/llvm-cov-target`) which is correct despite `defaults.run.working-directory: solana`, since `actions/cache` ignores job defaults. `cargo llvm-cov` cleans stale profraw by default (no `--no-clean`), so reusing the cached llvm-cov-target across runs is safe.

**Nits (real but minor):**

1. On a jq failure, the header + preamble lines have already been appended to the step summary before the error, leaving a rowless table stub in the summary. Cosmetic; the red job is the actual signal.
2. The component list duplicates `solana/Cargo.toml`'s member list by hand. The reconciliation check turns any drift into a hard failure (good), but every new workspace crate now requires editing this jq script. Deriving the label from the path segment after `crates/`/`programs/` would remove the maintenance coupling entirely — worth considering, not blocking.
3. `cargo-llvm-cov` is installed unpinned (`tool: cargo-llvm-cov` gets latest) — same as the coprocessor workflow, so it's consistent, but coverage numbers can drift a little across llvm-cov releases. Acceptable for an informational signal.
4. PR title scope `ci(test-suite)` doesn't match the touched area (the commit itself says `ci(solana)`, which is right).

## Fit with fhevm-internal#1657

- Delivers the core ask: cargo-llvm-cov coverage on the Solana crates as a visible CI signal.
- **No floor** — deliberately deferred; the PR body's argument (a single blended floor is meaningless when components range 98.38% → 6.05%, and Mollusk-covered behavior is structurally invisible to LLVM) is verifiable and convincing. But the issue's floor deliverable remains open, and this PR doesn't say where that follow-up is tracked. Per-component floors would be the natural next step and the table structure already supports them.
- **litesvm tier** — the issue said coverage should ideally run with the litesvm tier enabled; that tier doesn't exist in the workspace yet (only referenced in `TESTING_STRATEGY.md` as tracked follow-up #3045), so it couldn't be enabled here. Fine, but worth one line in the PR body linking the two.
- The numbers in the PR body (5,031/11,381 = 44.21%) are consistent with the CI run.

## Conventions, cost, risk

- **Conventions:** action pins (checkout, cache, install-action, rust-toolchain-file) are byte-identical SHAs to `coprocessor-cargo-tests.yml`; cache key pattern matches. It intentionally does *less* than the coprocessor pipeline (no lcov artifact, no diff-coverage PR comment, no `pull-requests: write`) — appropriate for a first informational signal, and it keeps permissions minimal.
- **Cost:** runs only on solana-path changes or manual dispatch; ~29s warm, and it duplicates the `build-and-test` test execution under instrumentation (a cold run is a full workspace build, single ubuntu-latest runner). Modest.
- **Gating:** it's a job in `solana-tests`, so it shows as a PR check; the strict reconciliation means an added crate makes this job red until the table is updated. That's intended, just be aware "informational" describes the *number*, not the job's pass/fail.
- **Summary size:** six-row table, nowhere near the 1 MB step-summary limit.
- **Flakiness:** coverage of deterministic unit tests is stable; the only drift vector is the unpinned llvm-cov version (nit 3).

## Verdict

**Approve** (as a draft ready to un-draft). The workflow is correct, the jq parser is verified against the real schema and workspace and fails loud on every edge case checked, and the scope cuts are honest and well-argued. Highest-value suggestions, none blocking:

1. Derive component labels from file paths instead of hard-coding the six-entry list, so new crates don't require editing the jq script (the reconciliation check already guarantees you can't miss one silently).
2. State in the PR body / TESTING.md where the deferred floor (per-component, once numbers stabilize) and the litesvm-tier coverage (#3045) are tracked, so #1657's remaining deliverables don't get lost.
3. Optionally build the table into a variable and append to `$GITHUB_STEP_SUMMARY` only on success, so a jq failure doesn't leave a rowless table stub in the summary.

Eikix:

No blocking issues; one measurement-honesty gap worth fixing before this number is quoted, plus one heads-up.

1. **Inline `#[cfg(test)]` modules are counted as covered product lines, so the per-component numbers overstate** (`.github/workflows/solana-tests.yml`, "Measure native unit coverage" step). cargo-llvm-cov's default filename ignore only drops `tests/` directories and files named `tests.rs`/`*_tests.rs` — so `zama-host/src/state/tests.rs` is already excluded, but inline `#[cfg(test)] mod tests` blocks inside product files are instrumented and count as ~100%-covered lines in both numerator and denominator. This is material for `zama-solana-acl`: of its ~1,330 raw source lines, roughly half sit in inline test modules in `lib.rs` (from ~L273), `mmr.rs` (~L175), and `lineage.rs` (~L203), so the headline 98.38% is substantially test-code self-coverage. The same applies to zama-host's inline test mods (`eip712.rs`, `instructions/common.rs`, `fhe_eval/*`, `state/encrypted_value.rs`, …) and to the 44.21% aggregate. For a PR whose stated purpose is an honest, non-inflated signal, either (a) move inline test mods into sibling `tests.rs` files (auto-excluded by the default regex, matching the existing `state/tests.rs` pattern), or (b) at minimum state this caveat in `solana/docs/TESTING.md` and the step-summary preamble alongside the Mollusk/SBF one. As written, the doc says excluding `zama-solana-runtime-tests` avoids "counting the Rust test harness" — but in-crate test harness lines are still counted.

2. **Heads-up: the jq totals guard makes any new workspace crate fail this job with only a jq `error(...)` as diagnosis** (workflow, component list in the jq script). The `$selected_total != $workspace_total` check is a good honesty guard (silently unattributed crates would be worse), but the failure mode is a red `native-coverage` job whose only clue is `component rows cover X of Y workspace lines` deep in a step log. A one-line comment in the YAML ("adding a workspace member? add it to this component list") would save the next author a hunt.

Verified specifically: the six hardcoded component paths exactly match the six non-excluded workspace members in `solana/Cargo.toml` at the head ref, with no overlapping path prefixes (no double counting); the cache paths (`solana/target/llvm-cov-target` relative to workspace root vs. `working-directory: solana`) are consistent; and the SBF/Mollusk rationale is accurate — Mollusk executes prebuilt `.so` artifacts, so excluding `zama-solana-runtime-tests` is both necessary (its build needs SBF artifacts this job never produces) and correct for attribution.

PR #3085, opened 2026-07-10
fix(kms-connector): bound Solana decryption retries

Eikix:
## What

- preserve verify-first acceptance for Solana MMR proofs, even when proof and live leaf counts differ;
- classify failed inclusion proofs directionally: proof count below or equal to live state is terminal, while an inclusion-proof mismatch ahead of the KMS finalized view gets an ordinary bounded retry;
- apply the existing 60-claim finalization bound to account observations whose finalized context slot still predates the queued event;
- keep witness storage and decrypt-material release behind `observed_slot >= block_number`.

An ahead-of-live inclusion-proof mismatch uses the existing `ProcessingError::Recoverable` path and therefore consumes the normal `max_decryption_attempts` budget. There is no budget-exempt retry path. Domain, owner, canonical-account, bump, and inconsistent-MMR failures remain terminal regardless of leaf counts.

## Validation

- `SQLX_OFFLINE=true cargo test -p kms-worker core::event_processor::solana_user_decrypt::tests` (24 passed)
- `SQLX_OFFLINE=true cargo test -p host-listener solana_finalized_account_fetcher::tests` (22 passed)
- `SQLX_OFFLINE=true cargo clippy -p kms-worker --all-targets -- -D warnings`
- `SQLX_OFFLINE=true cargo clippy -p host-listener --all-targets -- -D warnings`
- `cargo fmt -p kms-worker -- --check`
- `git diff --check`

The local `attempt_limit` integration suite requires its containerized Postgres fixture; all eight cases stopped at container creation in the restricted environment before exercising code.

Tracks [zama-ai/fhevm-internal#1687](https://github.com/zama-ai/fhevm-internal/issues/1687) and only the stale-finalized-witness overlap with [zama-ai/fhevm-internal#1682](https://github.com/zama-ai/fhevm-internal/issues/1682).

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/solana_finalized_account_fetcher.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/solana_finalized_account_fetcher.rs
@@ -30,6 +30,13 @@ use crate::solana_adapter::{
 /// dropped/reorged transaction) so the queue cannot hot-loop forever.
 const MAX_FINALIZATION_ATTEMPTS: i32 = 60;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum FinalizedAccountFetchAction {
+    Process,
+    Retry,
+    Fail,
+}
+
 #[derive(Clone, Debug)]
 pub struct SolanaFinalizedAccountFetcherConfig {
     pub rpc_url: String,
@@ -632,10 +639,12 @@ where
     // the RPC against not-yet-finalized accounts.
     let mut terminal = 0usize;
     for (job, account) in jobs.iter().zip(accounts) {
-        match account {
-            Some(account)
-                if !finalized_slot_covers_job(job, account.observed_slot) =>
-            {
+        let action = finalized_account_fetch_action(
+            job,
+            account.as_ref().map(|account| account.observed_slot),
+        );
+        match (account, action) {
+            (Some(account), FinalizedAccountFetchAction::Retry) => {
                 retry_finalized_account_fetch(
                     &mut tx,
                     job,
@@ -646,7 +655,19 @@ where
                 )
                 .await?;
             }
-            Some(account) => {
+            (Some(account), FinalizedAccountFetchAction::Fail) => {
+                fail_finalized_account_fetch(
+                    &mut tx,
+                    job,
+                    &format!(
+                        "finalized account context slot {} still predates job block {} after max attempts",
+                        account.observed_slot, job.block_number
+                    ),
+                )
+                .await?;
+                terminal += 1;
+            }
+            (Some(account), FinalizedAccountFetchAction::Process) => {
                 let witness = SolanaFinalizedAccountWitness {
                     account_key: job.account_key,
                     owner: account.owner,
@@ -698,25 +719,27 @@ where
                 }
                 terminal += 1;
             }
-            None => {
+            (None, FinalizedAccountFetchAction::Retry) => {
                 // The allow may simply not have finalized yet: retry until the
                 // attempt budget is spent, then fail hard.
-                if job.attempts < MAX_FINALIZATION_ATTEMPTS {
-                    retry_finalized_account_fetch(
-                        &mut tx,
-                        job,
-                        "account not yet found at finalized commitment; awaiting finalization",
-                    )
-                    .await?;
-                } else {
-                    fail_finalized_account_fetch(
-                        &mut tx,
-                        job,
-                        "account not found at finalized commitment after max attempts",
-                    )
-                    .await?;
-                    terminal += 1;
-                }
+                retry_finalized_account_fetch(
+                    &mut tx,
+                    job,
+                    "account not yet found at finalized commitment; awaiting finalization",
+                )
+                .await?;
+            }
+            (None, FinalizedAccountFetchAction::Fail) => {
+                fail_finalized_account_fetch(
+                    &mut tx,
+                    job,
+                    "account not found at finalized commitment after max attempts",
+                )
+                .await?;
+                terminal += 1;
+            }
+            (None, FinalizedAccountFetchAction::Process) => {
+                unreachable!("a missing account cannot be ready for processing")
             }
         }
     }
@@ -730,6 +753,19 @@ fn finalized_slot_covers_job(
     observed_slot: u64,
 ) -> bool {
     observed_slot >= job.block_number
+}
+
+fn finalized_account_fetch_action(
+    job: &SolanaFinalizedAccountFetchJob,
+    observed_slot: Option<u64>,
+) -> FinalizedAccountFetchAction {
+    if observed_slot.is_some_and(|slot| finalized_slot_covers_job(job, slot)) {
+        FinalizedAccountFetchAction::Process
+    } else if job.attempts < MAX_FINALIZATION_ATTEMPTS {
+        FinalizedAccountFetchAction::Retry
+    } else {
+        FinalizedAccountFetchAction::Fail
+    }
 }
 
 async fn retry_jobs(
@@ -916,6 +952,41 @@ mod tests {
         assert!(!finalized_slot_covers_job(&job, account.observed_slot));
         job.block_number = account.observed_slot;
         assert!(finalized_slot_covers_job(&job, account.observed_slot));
+    }
+
+    #[test]
+    fn finalized_account_fetch_transition_is_bounded_and_fail_closed() {
+        let mut job = acl_record_job(
+            "encrypted_value_created",
+            Some(Handle::from([6; 32])),
+        );
+        job.block_number = 43;
+        job.attempts = MAX_FINALIZATION_ATTEMPTS - 1;
+        assert_eq!(
+            finalized_account_fetch_action(&job, Some(42)),
+            FinalizedAccountFetchAction::Retry
+        );
+        assert_eq!(
+            finalized_account_fetch_action(&job, None),
+            FinalizedAccountFetchAction::Retry
+        );
+
+        job.attempts = MAX_FINALIZATION_ATTEMPTS;
+        assert_eq!(
+            finalized_account_fetch_action(&job, Some(42)),
+            FinalizedAccountFetchAction::Fail
+        );
+        assert_eq!(
+            finalized_account_fetch_action(&job, None),
+            FinalizedAccountFetchAction::Fail
+        );
+
+        // The witness/material path is reached only once the finalized RPC
+        // context covers the queued event, even after the retry budget expires.
+        assert_eq!(
+            finalized_account_fetch_action(&job, Some(43)),
+            FinalizedAccountFetchAction::Process
+        );
     }
 
     #[test]
--- a/kms-connector/crates/kms-worker/src/core/event_processor/error.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/error.rs
@@ -10,12 +10,6 @@ pub enum ProcessingError {
     Irrecoverable(anyhow::Error),
     #[error("Processing failed: {0}")]
     Recoverable(anyhow::Error),
-    /// Recoverable, but exempt from the attempt budget: the failure is provably a stale-proof
-    /// retry (e.g. a Solana MMR proof built against a chain tip the lineage has since advanced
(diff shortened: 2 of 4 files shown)


Eikix:
_Automated review — generated by a Claude (Fable) review agent dispatched by the PR author; findings verified against `origin/feature/solana` (`a8d3c287`)._

---

**Scope reviewed:** 2 commits (`2910ff59`, `97e10f36`), 4 files, +111/−118.

## Overview

The PR closes the two unbounded retry loops from fhevm-internal#1687:

1. **kms-worker** (`solana_user_decrypt.rs`, `processor.rs`, `error.rs`): deletes the `ProcessingError::RecoverableBudgetExempt` variant entirely. `classify_mmr_verification_failure` now maps *every* MMR-proof verification failure to `Irrecoverable` (keeping both leaf counts in the message for diagnostics). The processor's budget-exempt retry arm and the counter-increment exemption in `inner_process` are removed, so `error_counter` now increments on every prepare-request failure.
2. **host-listener** (`solana_finalized_account_fetcher.rs`): introduces a pure decision function `finalized_account_fetch_action(job, observed_slot) -> Process | Retry | Fail` and rewrites the batch loop around it, extending the existing `MAX_FINALIZATION_ATTEMPTS = 60` bound (previously only applied to account-NOT-FOUND) to the account-exists-but-`observed_slot < block_number` case.

## Correctness

### Path (a): stale MMR proof (kms-worker)

- **Terminates: yes.** Verified via `git grep RecoverableBudgetExempt` on the base branch that all producers/consumers of the variant lived in exactly the four touched files; nothing else constructs it, so removal is complete. The only remaining `Recoverable` producers in `solana_user_decrypt.rs` are `fetch_encrypted_value_acl` (RPC error / lineage account not found at finalized) and `dispatch_solana_current` — both genuinely transient, both now unconditionally consume `error_counter` (persisted through `update_status` on every `mark_as_pending`, read back by the picker → restart-safe), and both are cut off by the existing `max_decryption_attempts` arm (default 20) for all four decryption event kinds including `UserDecryptionSolana`. No path maps to a budget-exempt state anymore.
- **Terminal state:** `Irrecoverable` → `mark_as_failed` → row `status = 'failed'` in `user_decryption_requests`. Note that **no error response is sent to the gateway** — the request just parks as `failed` and the client presumably times out. This is the pre-existing semantics for every `Irrecoverable` error in this processor, so the PR is consistent; but see "Risks".
- **Security note (in the PR's favor):** the old taxonomy keyed budget exemption on `auth.proof_leaf_count`, which is *client-supplied wire data*. A malicious requester could deliberately mismatch it to earn infinite budget-exempt retries — the DoS #1687 describes. Removing the exemption closes that; the leaf counts are now diagnostic-only.

- **⚠️ Main finding — terminal is over-broad in one direction.** The rationale ("a failed proof is terminal because the queued request cannot change its proof bytes") is only sound when the *live* state has advanced past the proof (`live_leaf_count > proof_leaf_count`, mountain merged). The opposite skew is transient, not immutable: a client builds a proof against a fresher view than the KMS's finalized RPC read (`proof_leaf_count > live leaf_count`, e.g. a decrypt requested within Solana's ~13 s finalization lag of the allow, or the KMS's RPC node simply lagging). `fetch_encrypted_value_acl` re-reads the lineage live at finalized commitment on every attempt, so on the base branch that request failed verification, got retried, and *succeeded once the node caught up* (the existing, kept test `valid_proof_survives_count_drift_without_merge` shows verification passes once live state covers the proof). After this PR it is **permanently failed on first attempt** even though the identical bytes would verify seconds later. Issue #1687 explicitly allowed "consume the normal attempt budget" as the alternative, and that option handles this race while still terminating. **Suggestion:** map MMR verification failures to `Recoverable` (budget-counted) instead of `Irrecoverable` — bounded by `max_decryption_attempts`, not gameable (budget always consumed), and tolerant of finality lag. At minimum do so when `proof_leaf_count > live_leaf_count`. Caveat: the commitment level at which the client SDK builds proofs was not verified in this review, so the frequency of this race is unquantified — but the code makes it reachable.

### Path (b): behind-slot finalized witness (host-listener)

- **Terminates: yes.** `attempts` is incremented by `claim_pending_finalized_account_fetches` (`SET attempts = attempts + 1`) on *every* claim, before the RPC fetch, identically for exists/absent accounts — so the "claim" semantics are the same for the new case as for not-found, and the counter is DB-persisted and monotonic (no code path resets it; `retry_/fail_/complete_` don't touch it). Restart-safe. `finalized_account_fetch_action` fails once `attempts >= 60` regardless of whether the account is present.
- **Release gate untouched:** `finalized_slot_covers_job` (`observed_slot >= job.block_number`) is unchanged and is the *only* route to `Process`; the added test asserts `Process` at `attempts = MAX` when the slot covers — correctly giving the release condition **priority over the budget**, so a job is never failed on the same poll where the witness actually became valid.
- **No stale read recorded as success (fhevm-internal#1682):** the `Fail` arm calls only `fail_finalized_account_fetch` (row → `status = 'error'`) — no `store_finalized_account_witness`, no `complete_`, no `insert_allowed_handle`. Fail-closed, exactly what #1682 demands.
- **`(None, Process)` `unreachable!`:** genuinely unreachable (`Process` requires `observed_slot.is_some_and(..)`), so it can't panic. Minor style: matching the tuple `(account, action)` manufactures this impossible arm; returning the account inside the `Process` variant (or matching account first) would eliminate it. Non-blocking.
- **Semantics of reusing the 60-claim bound:** correct in mechanism (one claim per poll per job, batch loop paced to `poll_interval` for non-empty batches too). One behavioral consequence worth stating in the PR description: **jobs already stuck behind-slot before this deploy will have `attempts` ≫ 60 and will be terminally failed on their first post-deploy claim.** That drains the poison rows (intended), but affected requesters must resubmit.

## Test coverage

- **host-listener:** `finalized_account_fetch_transition_is_bounded_and_fail_closed` is a good boundary test — Retry at `MAX−1` for both `Some(behind)` and `None`, Fail at `MAX` for both, and Process winning past the budget. However, it exercises only the pure decision function. **No test drives `process_finalized_account_fetch_batch` through the new `(Some, Fail)` arm** to prove the DB effects (row → `'error'`, no witness inserted, `terminal` counted). Since the whole point is "nothing releases authorization prematurely," a batch-level test with a mock client returning a behind-slot account at `attempts = MAX` would be the strongest artifact the issue asked for. Suggest adding one.
- **kms-worker:** `stale_merged_proof_is_terminal` proves the merged-stale proof is `Irrecoverable` and retains both leaf counts. Deleting `invalid_mmr_proof_classification_uses_leaf_count` is right (the classification no longer exists), and renaming the domain-failure test is fine. There is no new test that a stale proof reaches `mark_as_failed` through the processor, but the existing `Irrecoverable` arm is pre-existing and already exercised (`tests/acl.rs` covers the attempt-budget path). Adequate — unless the Recoverable suggestion above is taken, in which case the tests flip.

## Conventions / taxonomy

- Removing a whole error variant instead of guarding it is the right shape — the two-variant `Irrecoverable`/`Recoverable` taxonomy is what the rest of kms-worker uses; the exempt variant was the odd one out.
- **Stale doc comment left behind:** `dispatch_solana_mmr_proof`'s doc (lines 244–245 of the PR head) still says it "classifies a verification failure by `auth.proof_leaf_count` vs `acl.leaf_count` (see the module doc's freshness contract)" — no longer true. Should be updated alongside the module doc (which was updated).
- The retained module-doc paragraph about verify-first accepting count drift is accurate and still load-bearing.

## Risks

- **Legitimate slow finalization killed early:** budget ≈ 60 attempts × `poll_interval` (default 1000 ms) ≈ **~60 s** against Solana's ~13 s (~32-slot) finalization lag — comfortable on a healthy node, thin against a degraded/snapshot-catching-up RPC. More concerning: on an RPC *transport* failure the claim increment has already committed before `retry_jobs` runs, so **a ~60 s RPC outage burns the entire budget with zero real observations**, after which the first successful-but-behind fetch hard-fails. This was pre-existing for not-found; the PR extends the exposure to behind-slot. Consider not counting attempts (or decrementing) on transport-level failures, or raising the constant — as a follow-up, not necessarily here.
- **No recovery from `'error'`:** the claim query selects only `'pending'` and stale `'processing'` rows, so a hard-failed fetch requires manual DB intervention or client resubmission. Intended fail-closed behavior, but ops should know.
- **Silent terminal failures in kms-worker:** `failed` rows emit no gateway response (pre-existing); combined with the main finding, a user whose fresh proof hit a lagging node gets neither a decryption nor an error.
- **Restart behavior:** both counters are DB-persisted (fetcher `attempts` via the claim UPDATE; kms-worker `error_counter` via `update_status`) — no reset-on-restart loophole on either path. Verified.

## Verdict

**Request changes** — the retry-bounding itself is correct, complete (no dangling `RecoverableBudgetExempt` producers), fail-closed per #1682, and restart-safe on both paths. But before merging:

1. **(Blocking discussion)** Reconsider `Irrecoverable` → `Recoverable` (budget-counted) for MMR verification failures, at least for `proof_leaf_count > live_leaf_count`: the current mapping permanently fails *valid fresh proofs* whenever the KMS's finalized view lags the client's — a user-visible regression the "immutable proof bytes" rationale doesn't cover, and the budget-counted option satisfies #1687 equally while staying non-gameable.
2. Add a batch-level test driving `process_finalized_account_fetch_batch` through the new `(Some(account), Fail)` arm, asserting `status = 'error'` and no witness row.
3. Fix the stale `dispatch_solana_mmr_proof` doc comment referencing the removed freshness classification.
4. (Nit) Consider restructuring the `(account, action)` match to remove the `unreachable!` arm; (note for PR body) call out that pre-existing stuck behind-slot jobs with `attempts > 60` will be drained to `'error'` on deploy.

Eikix:

No blocking issues found; the classification change is sound. Three non-blocking observations below.

Verified specifically: no `RecoverableBudgetExempt` / "budget-exempt" references remain anywhere at the head ref; `error_counter` is persisted on `mark_as_pending` and reloaded by the event picker, so the retry budget survives restarts; and the ahead-of-view retry can actually heal, because `check_solana_handles_acl` re-fetches the `EncryptedValue` account at finalized commitment on every attempt. The stale-terminal call is correct for an append-only MMR: at equal leaf counts the peaks are unique, and once the leaf's mountain has merged past the proof's sibling depth the immutable queued proof can never verify again.

1. `solana_user_decrypt.rs:341-371` — the terminal-vs-retry decision hinges on the client/relayer-declared `proof_slot`, not on anything derived from the proof itself. A genuinely ahead-of-view proof whose declared count understates the build-time `leaf_count` (relayer bug, or a client passing the current-ACL default `0`, cf. `userDecrypt.ts:333`) is dropped as "stale and immutable" on the first attempt instead of retried. The relayer does fill `proof_slot = leaf_count` today (`relayer/src/solana_proof/proof.rs:79`), and the inflated-count direction is harmless (bounded retries then failure), so this is informational — but the invariant "declared count must equal build-time leaf_count or ahead-of-view requests drop early" is worth a code comment or SDK-side assertion.

2. Healing-window sizing: the ahead case now tolerates at most `max_decryption_attempts` (default 20, `config.rs:249`) × `db_fast_event_polling` (default 3s) ≈ 60s of finalized-view skew between the proof service's RPC node and the KMS's RPC node; beyond that a valid proof fails terminally. Against a ~13s finality lag that leaves headroom, but the budget is shared with unrelated recoverable failures (network blips, account-not-yet-found in `fetch_encrypted_value_acl`), so a request that spent most of its 20 attempts on RPC errors can drop before catch-up. Likely acceptable; flagging so the sizing is a conscious choice.

3. `solana_finalized_account_fetcher.rs:758-768` — the newly bounded stale-context branch shares the single claim-based `attempts` budget with the not-found phase (60 claims ≈ 60s at the default 1s `poll_interval`). Behavior change to be aware of: a finalized RPC node lagging >~60s now hard-fails allow jobs that previously waited indefinitely. The intended safeguard holds — `Process` is checked before the budget, so a covering snapshot is still stored even after attempt 60 (asserted in `finalized_account_fetch_transition_is_bounded_and_fail_closed`).

Minor test note: the ahead-recoverable case is exercised only at the pure-classifier level; the stale and domain/canonical terminal cases retain full `dispatch_solana_mmr_proof` coverage, which seems sufficient.

PR #3086, opened 2026-07-10
refactor(coprocessor): remove Solana emit listener

Eikix:
## Summary

- Delete the unused Solana JSON-RPC emit listener and remove emit decoding/fallback from Yellowstone ingestion.
- Make `solana-grpc` include reconstruction and expose only the reconstruction CLI path.
- Return reconstructed events directly without the obsolete covered/not-covered wrapper.
- Preserve the relayer/MMR transaction ingestion boundary while deleting the coprocessor's event-decoder path.
- Refresh only the invocation and design text made stale by the deletion.

The approved follow-up architecture in fhevm-internal#1675 does not restore this listener. Ordinary computation and ACL reconstruction remain Yellowstone-only. A separate follow-up will add one narrow host-owned `BornPublicOutputsEvent` batch solely for entropy-derived `make_public=true` handles consumed by the RPC-polling relayer; general operation events remain non-authoritative.

## Validation

- `SQLX_OFFLINE=true cargo check -p host-listener`
- `SQLX_OFFLINE=true cargo check -p host-listener --features solana-grpc --all-targets`
- `SQLX_OFFLINE=true cargo test -p host-listener --features solana-grpc --lib` (134 passed, 1 ignored)
- `SQLX_OFFLINE=true cargo clippy -p host-listener --features solana-grpc --all-targets -- -D warnings`
- `cargo fmt -p host-listener -- --check`
- `bash -n solana/scripts/poc/setup-solana-side.sh`

Context: https://github.com/zama-ai/fhevm-internal/issues/1676


Eikix:
_Automated review — generated by a Claude (Fable) review agent dispatched by the PR author; findings verified against `origin/feature/solana`._

---

## Overview

The PR deletes the emit-decode ingestion path from the coprocessor host-listener, completing the reconstruct-only commitment from #3002 / fhevm-internal#1676:

- Deletes `coprocessor/fhevm-engine/host-listener/src/solana_listener.rs` (806 lines: the RPC `getSignaturesForAddress` polling transport + its ~230 lines of tests) and its `solana-transaction-status` dependency (Cargo.lock pruned accordingly).
- `src/bin/solana_host_listener.rs`: removes the `--transport rpc|grpc`, `--reconstruct`, and `--poll-interval-ms` flags; the binary is now Yellowstone-reconstruction-only, with the HostConfig auto-detection (chain_id, zero-birth-entropy) now unconditional.
- `Cargo.toml`: `solana-grpc` now includes `solana-reconstruct`.
- `src/solana_grpc_listener.rs`: removes emit-decode + the `ReconstructionOutcome::{Complete, NotCovered}` wrapper and its fallback-to-emit branch; `reconstruct_events_for_insert` returns `Vec<SolanaHostEvent>` directly, `NotCovered` becomes `Ok(Vec::new())`. All `#[cfg(feature = "solana-reconstruct")]` gates inside the file dropped (safe, since the file only compiles under `solana-grpc`, which now implies reconstruction).
- Scripts (`setup-solana-side.sh`, `clean-e2e.sh`) and five `solana/docs/*.md` files updated.

## Correctness

**1. Dangling references — none found.** Searched `origin/feature/solana` for every removed symbol/flag: all remaining references to `solana_listener`, `SolanaListenerConfig`, `extract_host_events`, `ReconstructionOutcome`, `--transport`, `--reconstruct` are inside files this PR edits. No other crate (`tfhe-worker`, `test-harness`, `relayer`, `stress-test-generator`) imports the removed module. The `poll_interval_ms` hits elsewhere (`poller.rs`, `solana_finalized_account_fetcher.rs`, relayer configs) are unrelated flags of the same name. `decode_solana_transaction_events` is deliberately *kept* in `solana_adapter.rs` — correctly so: it is still used by `decode_solana_transaction_account_fetches` (solana_adapter.rs:313) and by the LiteSVM in-process tests (`tfhe-worker/src/tests/solana_poc.rs:19,1227`). Not dead code.

**2. Boundary holds.** No on-chain program files or relayer files are in the diff. `emit_cpi!` emission and the relayer's op-event consumers (`relayer/src/solana_proof/decode.rs`, `proof.rs`) are untouched, and the `MMR_ACL_MVP.md` edit states the boundary precisely: listener emit-decode removed now, on-chain events deferred to Carbon indexing (fhevm-internal#1665). This matches the issue's stated boundary exactly.

**3. Covered/not-covered wrapper removal — safe.** The `NotCovered → fall back to emit-decode` branch (and its warn) only had observable effect against a program that actually emits. The supported deployment builds zama-host EMITLESS (setup-solana-side.sh: "EMITLESS program deploy"; solana-e2e.yml: "the emit/RPC transport has been retired"), so `emit_events` was always empty, the fallback ingested nothing, and the warn was guarded by `!emit_events.is_empty()` and never fired. `NotCovered → Ok(Vec::new())` then hits the same `events.is_empty()` early return. No caller, metric, or fallback depended on the signal. Two minor observability notes: the `info!(slot, reconstructed, emit_decoded, "ingesting reconstructed events…")` per-transaction count log is dropped without replacement, and error classification is unchanged (reconstruction errors were already `IngestFailure::fatal` = fail-closed, stop-without-advancing-cursor, on the base branch).

**4. Feature-flag hygiene — correct and necessary.** The grpc listener file now references `solana_reconstruct` items unconditionally, so `solana-grpc` *must* imply `solana-reconstruct`; the Cargo.toml change does that. The PR's validation list includes `cargo check -p host-listener` **without** the feature (which covers the `run_grpc` stub-bail bin) and `--features solana-grpc --all-targets` with clippy `-D warnings`. One combination not listed: `--features solana-reconstruct` **alone** (still a legal standalone feature; it should compile since `solana_reconstruct.rs` is self-contained, but adding that check would close the gap). CI's `solana-tests.yml:142` still passes `--features solana-grpc,solana-reconstruct` — now redundant but valid; could be simplified to `solana-grpc` (optionally in #3089).

**5. CLI/deployment — clean, with one finding.**
- `setup-solana-side.sh` is updated in this PR (drops `--transport grpc --reconstruct`, builds with `solana-grpc` only). The `RECONSTRUCT`/`SKIP_BUILD` env vars removed from `clean-e2e.sh` were already dead: the base `setup-solana-side.sh` never reads either variable (verified by grep). No workflow passes the removed flags.
- **Finding (pre-existing gap made sharper):** `host-listener/Dockerfile` builds and ships `solana_host_listener` with `CARGO_FEATURES=""` by default, and `coprocessor-docker-build.yml:370` builds that image with no Solana features. That shipped binary now *unconditionally bails at runtime* ("requires building with --features solana-grpc"), where before it at least had a working RPC transport. Nothing currently invokes it from the image (the e2e builds locally with cargo), so this isn't a breakage — but the mainline image now carries a stub binary. Worth deciding (possibly in a follow-up): default the Docker Solana bin build to `solana-grpc`, or stop copying `solana_host_listener` into the featureless image.

## Test coverage

Five tests deleted with the module:
- `per_signature_decode_error_advances_cursor_and_continues_batch` — RPC-poll cursor semantics; dead with the transport.
- `extracts_zama_host_cpi_event_from_inner_instruction`, `extracts_zama_host_event_from_log_transport`, `ignores_inner_instruction_from_other_program`, `block_meta_uses_slot_and_block_time` — these tested `getTransaction` JSON wire-shape parsing plus decode. The decode-level assertions survive in `solana_adapter.rs`'s own tests (log-transport decode at ~1605, decode-error at ~1686) and the LiteSVM path in `tfhe-worker/solana_poc.rs`; the JSON wire-shape parsing lost its only consumer. Genuinely redundant.

The reconstruction parity tests (`fhe_eval_acl_tests`) survive intact and, importantly, assert against **explicit expected semantic facts** (derived handles, fetch reasons), not against emit-decode output — so they do not weaken when emit-decode disappears. The `complete_events` unwrap helper removal is mechanical.

## Conventions / docs

- Doc updates are accurate and appropriately scoped: the "Solana poller polls at `confirmed`, open reorg gap" text in DESIGN_DECISIONS/EVM_PARITY/FUTURE_DESIGN was genuinely stale (the listener consumes *finalized* Yellowstone updates) and is fixed consistently across all five docs. Remaining `getSignaturesForAddress`/RPC-polling mentions all refer to the relayer, which is intentionally kept.
- Nit: awkward doc-comment line break in `solana_grpc_listener.rs` — "…from a transaction's instructions.\n/// Covers\n/// `fhe_eval`: …" leaves "Covers" stranded.
- Nit: `setup-solana-side.sh`'s failure message still says "host-listener (grpc,reconstruct) build failed" though the feature list is now just `solana-grpc`. Cosmetic.

## Risks

- **No rollback to emit ingestion.** If reconstruction has a bug, there is no emit path to fall back to. This is acceptable per the issue: the emit path had a real correctness hole (could not schedule decryption for `fhe_eval` outputs without a same-transaction `allow`), and the deployed program is emitless anyway, so the emit path already ingested nothing. The fail-closed `IngestFailure::fatal` behavior (halt, don't advance cursor) is the right failure mode for that world and is unchanged.
- **Stacked ordering:** #3089 (remaining dead emit tests) is stacked on this branch and must merge after; this PR itself is self-contained against `feature/solana`.

## Verdict

**Approve** (as a draft, pending the validation list holding in CI). The deletion is exactly scoped to the issue's boundary, no dangling references were found, and the PR body's claims all check out against the code. Highest-value suggestions, none blocking:

1. Decide the Docker story for `solana_host_listener` in the featureless mainline image (now a runtime stub) — default the feature on for that bin, or stop shipping it there.
2. Add `cargo check -p host-listener --features solana-reconstruct` (alone) to validation to cover the one remaining unexercised feature combination.
3. Simplify `solana-tests.yml`'s `--features solana-grpc,solana-reconstruct` to `solana-grpc` (fine to fold into #3089).
4. Cosmetic: fix the stranded "Covers" doc-comment line and the "(grpc,reconstruct)" build-failure message in `setup-solana-side.sh`.

Eikix:

No blocking issues found; one deployment-surface question and one lost-diagnostic observation below.

Deletion honesty was verified independently at the head ref: `git grep` across the whole tree (CI workflows, charts, docker, test-suite, docs) finds no remaining references to `solana_listener`, `SolanaListenerConfig`, `extract_host_events`, the `--transport`/`--reconstruct`/`--poll-interval-ms` flags, or the `RECONSTRUCT`/`SKIP_BUILD` env vars (prose mentions only). The relayer emit dependency is intact: `solana/programs/zama-host/src/instructions/fhe_eval/event_transport.rs` still `emit_cpi!`s every eval event, the `emit-events` default feature is untouched, and `relayer/src/solana_proof/` is unchanged — born-public handle recovery over RPC (fhevm-internal#1676) is preserved. Only the coprocessor listener path is removed, matching the PR's claim.

1. **Featureless builds now ship a bail-at-startup binary — intended?** `coprocessor/fhevm-engine/host-listener/Dockerfile` unconditionally builds and copies `solana_host_listener` into the image, with default `ARG CARGO_FEATURES=""`, and no workflow/chart in the tree passes `solana-grpc` to that build. Before this PR the featureless binary defaulted to the working RPC transport; after it, `run_grpc` (`src/bin/solana_host_listener.rs:161-168`) bails immediately. Nothing in-tree currently runs the dockerized binary (the e2e builds from source in `setup-solana-side.sh`), so this looks acceptable for the PoC — but if a downstream deployment consumes this image's `solana_host_listener`, it now needs `CARGO_FEATURES=solana-grpc` wired in. Worth a sentence in the PR description or a Dockerfile comment.

2. **Silent no-op on undecodable `fhe_eval` args (lost diagnostic, not a new bug).** In `solana_grpc_listener.rs`, `reconstruct_events_for_insert` gates on `decode_fhe_eval_args(&ix.data).is_some()`; a transaction whose `fhe_eval` instruction data fails to decode (e.g. stale listener vs. a moved instruction ABI) now returns `Ok(Vec::new())` and ingests nothing, with no log. The old `NotCovered` arm at least emitted a "falling back to emit-decode" warning when events were present. Since ingestion is emitless by design (DD-037) the fallback data wouldn't exist anyway, but a `warn!` when a zama-host-touching transaction reconstructs to zero events would preserve the drift signal the IDL golden check doesn't cover at runtime.

Minor: `.github/workflows/solana-tests.yml:142` still passes `--features solana-grpc,solana-reconstruct`; harmless (both features exist, `solana-grpc` now implies the other) but can be shortened to match the new feature graph.

Docs updates (DD-025/DD-028 reorg-gap rewording, `MMR_ACL_MVP.md` relayer-only emit rationale) accurately reflect the code state after this PR.

Eikix:
superseded by #3131

PR #3087, opened 2026-07-10
refactor(sdk): remove fake Solana client inputs

Eikix:
## Purpose

Partially address [zama-ai/fhevm-internal#1592](https://github.com/zama-ai/fhevm-internal/issues/1592) by removing the Solana SDK type fabrications that can be corrected without changing public host-chain metadata or extension APIs.

## Changes

- type the core client’s existing chainless construction path explicitly
- remove unused EVM chain/client parameters from transport-key generation
- pass only the decrypt runtime and resolved TKMS version from both EVM and Solana callers
- allow private TFHE/TKMS version setters to accept the existing hostless core client
- remove the Solana base-client, version-setter, and transport-key double casts / `never` placeholders
- add focused transport-key and hostless Solana-client tests

## Intentionally quarantined

This does not claim to complete [fhevm-internal#1592](https://github.com/zama-ai/fhevm-internal/issues/1592).

- The Solana proof builder still uses an EVM-shaped ACL adapter. Removing it honestly reaches `FheEncryptionKeyMetadata.chainId: number`, while Solana’s RFC-021 host ID is an exact `bigint`. Widening that externally visible metadata family or converting the ID lossily is outside this PR.
- The remaining `.ready` and `init` casts come from the public `FhevmExtension` callback being typed as `FhevmBase`. Changing that public extension contract is outside this PR.

No proof bytes, proof metadata, chain IDs, ACL behavior, public EVM action signatures, or initialization behavior are changed.

## Validation

Passed:

- focused Vitest run: 2 files, 2 tests
- focused Vitest typecheck: no errors
- full unit suite: 791 of 792 tests pass; the only failure is an existing untouched `ZkProofBuilder.test.ts` message mismatch
- `npm run prettier:ext`
- `git diff --check`

Existing branch/environment failures observed:

- full TypeScript build has two errors in untouched files: the TFHE `WasmAssetLoadMode` export and stale `transportPublicKey` Solana test input
- `npm run build:prod` stops on formatting in untouched `src/solana/index.ts`
- ESLint reports three existing errors in untouched Solana files
- cleartext v13 reached Anvil outside the sandbox, then stopped because Solidity dependencies were absent; `forge soldeer install` crashes in the installed Foundry build before installing them

### DIFF
--- a/sdk/js-sdk/src/core/actions/decrypt/generateTransportKeyPair.ts
+++ b/sdk/js-sdk/src/core/actions/decrypt/generateTransportKeyPair.ts
@@ -15,7 +15,7 @@ export async function generateTransportKeyPair(
   fhevm: Fhevm<FhevmChain, WithDecrypt>,
 ): Promise<GenerateTransportKeyPairReturnType> {
   const f = asFhevmWithTkmsVersion(fhevm);
-  return await generateTransportKeyPair_(f);
+  return await generateTransportKeyPair_({ runtime: f.runtime, tkmsVersion: f.tkmsVersion });
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/sdk/js-sdk/src/core/kms/TransportKeyPair-p.test.ts
+++ b/sdk/js-sdk/src/core/kms/TransportKeyPair-p.test.ts
@@ -1,0 +1,28 @@
+import type { WithDecrypt } from '../types/coreFhevmRuntime.js';
+import type { TkmsPrivateKey } from '../types/tkms-p.js';
+import { describe, expect, it, vi } from 'vitest';
+import { generateTransportKeyPair } from './TransportKeyPair-p.js';
+
+describe('generateTransportKeyPair', () => {
+  it('generates a key pair from only the decrypt runtime and TKMS version', async () => {
+    const tkmsPrivateKey = {} as TkmsPrivateKey;
+    const generateTkmsPrivateKey = vi.fn().mockResolvedValue(tkmsPrivateKey);
+    const serializeTkmsPrivateKey = vi.fn().mockResolvedValue(new Uint8Array([1, 2, 3]));
+    const getTkmsPublicKeyHex = vi.fn().mockResolvedValue('0x010203');
+    const runtime = {
+      decrypt: {
+        generateTkmsPrivateKey,
+        serializeTkmsPrivateKey,
+        getTkmsPublicKeyHex,
+      },
+    } as unknown as WithDecrypt;
+
+    const keyPair = await generateTransportKeyPair({ runtime, tkmsVersion: '0.13.20-0' });
+
+    expect(keyPair.publicKey).toBe('0x010203');
+    expect(keyPair.tkmsVersion).toBe('0.13.20-0');
+    expect(generateTkmsPrivateKey).toHaveBeenCalledWith({ tkmsVersion: '0.13.20-0' });
+    expect(serializeTkmsPrivateKey).toHaveBeenCalledWith({ tkmsPrivateKey, tkmsVersion: '0.13.20-0' });
+    expect(getTkmsPublicKeyHex).toHaveBeenCalledWith({ tkmsPrivateKey, tkmsVersion: '0.13.20-0' });
+  });
+});
--- a/sdk/js-sdk/src/core/kms/TransportKeyPair-p.ts
+++ b/sdk/js-sdk/src/core/kms/TransportKeyPair-p.ts
@@ -3,7 +3,7 @@ import type { FhevmRuntime, WithDecrypt } from '../types/coreFhevmRuntime.js';
 import type { Bytes, BytesHex } from '../types/primitives.js';
 import type { TkmsPrivateKey } from '../types/tkms-p.js';
 import type { TkmsVersion } from '../../wasm/tkms/KmsLibApi.js';
-import type { NativeClient, OptionalNativeClient } from '../types/coreFhevmClient.js';
+import type { OptionalNativeClient } from '../types/coreFhevmClient.js';
 import type { FhevmChain } from '../types/fhevmChain.js';
 import { assertIsBytesOrBytesHex, bytesToHexLarge, hexToBytesFaster } from '../base/bytes.js';
 import { InvalidTypeError } from '../base/errors/InvalidTypeError.js';
@@ -183,8 +183,6 @@ export function assertIsTransportKeyPair(
 /** Generates a fresh {@link TransportKeyPair}. */
 export async function generateTransportKeyPair(context: {
   readonly runtime: WithDecrypt;
-  readonly chain: FhevmChain;
-  readonly client: NativeClient;
   readonly tkmsVersion: TkmsVersion;
 }): Promise<TransportKeyPair> {
   const tkmsVersion = context.tkmsVersion;
--- a/sdk/js-sdk/src/core/runtime/CoreFhevm-p.ts
+++ b/sdk/js-sdk/src/core/runtime/CoreFhevm-p.ts
@@ -526,6 +526,21 @@ export type CreateCoreFhevmParameters<
   readonly options?: FhevmOptions | undefined;
 };
 
+export function createCoreFhevm<runtime extends FhevmRuntime>(
+  ownerToken: symbol,
+  parameters: {
+    readonly chain?: undefined;
+    readonly client?: undefined;
+    readonly runtime: runtime;
+    readonly options?: FhevmOptions | undefined;
+  },
+): Fhevm<undefined, runtime, undefined>;
+
+export function createCoreFhevm<chain extends FhevmChain, runtime extends FhevmRuntime, client extends NativeClient>(
+  ownerToken: symbol,
+  parameters: CreateCoreFhevmParameters<chain, runtime, client>,
+): Fhevm<chain, runtime, client>;
+
 export function createCoreFhevm<chain extends FhevmChain, runtime extends FhevmRuntime, client extends NativeClient>(
   ownerToken: symbol,
   parameters: CreateCoreFhevmParameters<chain, runtime, client>,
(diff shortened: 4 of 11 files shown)


Eikix:
_Automated review — generated by a Claude (Fable) review agent dispatched by the PR author; findings verified against `origin/feature/solana` (a8d3c2877) and the PR head._

---

## Overview

Two commits, 8 files, all in `sdk/js-sdk`:

1. **`29748d7` "remove fake Solana client inputs"** — delivers fixes (1) and (2) of fhevm-internal#1592: drops the dead `chain`/`client` params from the internal `generateTransportKeyPair` (`src/core/kms/TransportKeyPair-p.ts`), widens the private `setResolvedTfheVersion`/`setResolvedTkmsVersion` setters to accept a hostless client, and removes the resulting `as unknown as` / `{} as never` fabrications from `src/solana/clients/createFhevmBaseClient.ts` and the two Solana decorators.
2. **`80ee7cc` "require hosted core inputs"** — replaces the single `createCoreFhevm` signature with two overloads in `src/core/runtime/CoreFhevm-p.ts`: a chainless one returning `Fhevm<undefined, runtime, undefined>`, and a hosted one that now **requires** both `chain` and `client`.

## Correctness

### (1) `createCoreFhevm` relaxation — genuinely relax-only ✅

- **Body is undefined-safe.** The only `chain` dereference in the function body is `parameters.chain?.fhevm.relayerUrl` (optional chaining, with `?? fheEncryptionKey.metadata.relayerUrl` fallback); everything else is forwarded opaquely to `new CoreFhevmImpl(...)`. On base, `CreateCoreFhevmParameters` already declared `chain?: chain | undefined` and `client?: client | undefined`, and the Solana path was **already** calling it without chain/client at runtime — only the return type was faked via cast. So runtime behavior is unchanged; this commit is purely type-level.
- **EVM inference unchanged.** All hosted callers (`ethers/clients/createFhevmBaseClient.ts`, `ethers/cleartext/…`, `viem/clients/…`, `viem/cleartext/…`, plus the two `-p` test files) pass both `chain` and `client`, so they can't match the chainless overload (a concrete `FhevmChain` isn't assignable to `chain?: undefined`) and resolve to the hosted overload exactly as before. The instantiation expression in `ProtocolVersionResolver-p.test.ts` (3 type args) can only match the 3-type-param hosted overload — also unchanged.
- One nit: `CreateCoreFhevmParameters` is still exported with `chain?`/`client?` optional, which now understates what the hosted overload demands. Harmless (internal `-p` type, not on any public entrypoint), but worth a follow-up comment or tightening for consistency.

### (2) Removed `generateTransportKeyPair` params were truly dead ✅

Verified the base-branch body: it reads only `context.tkmsVersion` and `context.runtime.decrypt.{generateTkmsPrivateKey, serializeTkmsPrivateKey, getTkmsPublicKeyHex}`, then constructs `TransportKeyPairImpl(PRIVATE_TOKEN, context.runtime, …)`. `chain` and `client` were never touched. Grepped every reference on the base branch: the internal `generateTransportKeyPair_` has exactly two callers — the EVM action wrapper (`src/core/actions/decrypt/generateTransportKeyPair.ts`) and the Solana decrypt decorator — and **both are updated** in this PR. All other hits (docs, examples, `test/fheTest/**`) go through the public action or client method, whose signatures are unchanged. Note `toTransportKeyPair` in the same file still takes `chain`/optional `client` — out of scope here, fine.

### (3) Casts removed, no new dishonesty introduced ✅

Removed at the seam: the `as unknown as Fhevm<undefined, …>` on the Solana base client, both `as unknown as FhevmBase` setter casts, the `as unknown as WithTkmsVersion` read, and the `{} as never` chain/client placeholders. What replaced them is honest: `getResolvedTkmsVersion` already existed on base taking `unknown` with a runtime `asCoreFhevm` check. The one **new runtime behavior** is the `throw new Error('TKMS version is unresolved after client initialization')` guard — previously an `undefined` tkmsVersion would have silently flowed into the WASM calls. This is a strict improvement, but it's the only behavioral delta in the PR and it uses a bare `Error` where the SDK elsewhere has typed error classes (`src/core/base/errors/`) — consider a typed error for consistency.

### (4) "Require hosted core inputs" tightening — safe ✅

It converts "chain and client individually optional" into "both present or both absent" for the internal `createCoreFhevm`. A hypothetical caller passing `chain` without `client` (or vice versa) would now fail both overloads — but no such caller exists anywhere on the branch, and the function is not exported from any public entrypoint (grep of `src/index.ts` and all subpackage indexes: no hits for `createCoreFhevm`, the setters, or `CoreFhevm-p`). No EVM consumer can be broken.

### (5) Quarantine boundary — clean and honestly documented ✅

At PR head, the remaining Solana casts are exactly what the PR body declares:
- `encrypt.ts` lines 76/79: the EVM-shaped ACL adapter (`as unknown as Fhevm<FhevmChain, WithEncrypt> & WithTfheVersion`) — the deferred fix (3), blocked on `FheEncryptionKeyMetadata.chainId: number` vs Solana's exact `bigint` id, matching the issue's own constraint analysis.
- The `(fhevm as Fhevm<undefined, FhevmRuntime, undefined>).ready` narrowings and `init: _initX as (fhevm: FhevmBase) => Promise<void>` — traceable to the public `FhevmExtension.init` callback being typed `(client: FhevmBase<FhevmChain>)` (`coreFhevmClient.ts:114`), declared out of scope.

Each surviving cast has an accurate explanatory comment. The code is not left half-migrated; the boundary is legible.

### Version setter widening ✅

Base signatures were already `FhevmBase` = `FhevmBase<FhevmChain | undefined, FhevmRuntime, NativeClient>` by defaults — the PR only widens `client` from `NativeClient` to `OptionalNativeClient` (and makes the chain widening explicit). `FhevmBase` is a plain readonly-property interface, so hosted arguments remain assignable; the EVM callers `core/clients/decorators/{encrypt,decrypt}-p.ts` are untouched and unaffected. Relax-only confirmed.

## Test coverage

- **`TransportKeyPair-p.test.ts`** pins fix (2) precisely: key-pair generation from only `{runtime, tkmsVersion}`, asserting the three runtime calls receive `tkmsVersion`. Good regression pin.
- **`createFhevmBaseClient.test.ts`** pins fix (1) at both levels: `expectTypeOf(client.chain).toEqualTypeOf<undefined>()` (would have failed against the old cast-free inference), runtime `undefined` checks, exact `bigint` chain id preservation (`9223372036854788153n` > `MAX_SAFE_INTEGER` — deliberately pins the "no lossy Number conversion" constraint), and `client.ready` resolution. The `@ts-expect-error` inside `if (false)` pins the second commit's tightening.
- Nits: (a) the `if (false) { … }` dead-code guard works but is unidiomatic and likely to trip `no-constant-condition` lint someday; a never-invoked arrow or `expectTypeOf` on the function type would be cleaner. (b) `setFhevmRuntimeConfig({})` mutates global state with no reset — fine today with one test file, mildly fragile. (c) The unresolved-tkmsVersion throw path is untested — low value, not blocking.

## Conventions

- `-p.ts` internal naming respected; the new colocated `TransportKeyPair-p.test.ts` matches existing precedent (`HostContractVersion-p.test.ts`, `ProtocolVersionResolver-p.test.ts`).
- No public API churn for EVM consumers: the public action `generateTransportKeyPair(fhevm: Fhevm<FhevmChain, WithDecrypt>)` signature is byte-identical; changed symbols live only in `-p` internals not re-exported from any entrypoint (verified `src/index.ts`, `core/kms/index.ts`, and the ethers/viem/solana indexes).

## Risks

- **Pre-existing failures:** the changes stay clear of the files carrying the declared pre-existing breakage (`ZkProofBuilder.test.ts`, the `WasmAssetLoadMode` export, `solana/index.ts` formatting, ESLint hits) — nothing here worsens them.
- **d.ts output:** internal-file d.ts will change (overloads, widened setter params), but since nothing changed is on a public entrypoint and all signature changes are relax-only or internal-tightening, published consumers see no breaking surface.
- Only behavioral delta is the new throw on unresolved tkmsVersion (failure path only, previously silent-undefined) — an improvement.

## Verdict

**Approve** (as a draft into `feature/solana`). The PR does exactly what it claims, the relax-only/remove-dead constraints from #1592 are respected, all callers were found and updated, and the quarantine is honest and well-commented. Highest-value suggestions, none blocking:

1. Replace the bare `throw new Error('TKMS version is unresolved…')` with the SDK's typed error convention (`src/core/base/errors/`).
2. Tighten or annotate the still-exported `CreateCoreFhevmParameters` (chain/client still individually optional there, now inconsistent with the overloads).
3. Swap the `if (false)` type-pin for a never-invoked arrow or `expectTypeOf` to avoid future `no-constant-condition` lint noise; optionally add a reset for the `setFhevmRuntimeConfig({})` global in the new test.

Eikix:

No blocking issues found; the cast removals are sound and one question about whether the new type-level tests are actually enforced.

Verification done: all five `createCoreFhevm` call sites at the head ref (ethers/viem base + cleartext pass `chain`+`client`, Solana passes neither) match the new overload pair, so the split cannot mis-select for any existing caller. The private `generateTransportKeyPair` in `TransportKeyPair-p.ts` has exactly two callers (the public EVM action and the Solana decrypt decorator), both updated — no remaining caller passes the removed `chain`/`client`. The widened `setResolvedTfheVersion`/`setResolvedTkmsVersion` signatures only accept more (`FhevmChain | undefined`, `OptionalNativeClient`), so the EVM `encrypt-p.ts`/`decrypt-p.ts` decorators still type-check; `FhevmChain` stays imported in `TransportKeyPair-p.ts` because `parseTransportKeyPair`'s context (line 216) still uses it. `getResolvedTkmsVersion` takes `unknown` and validates via `asCoreFhevm`, so the new guarded read in `decrypt.ts:83` is strictly safer than the old `(fhevm as unknown as WithTkmsVersion).tkmsVersion`, which silently yielded `undefined` if init hadn't stored a version.

1. `sdk/js-sdk/src/solana/clients/createFhevmBaseClient.test.ts:21-31` — the compile-time guarantees this test claims (`@ts-expect-error` on hosted-core construction without chain/client, `expectTypeOf(client.chain).toEqualTypeOf<undefined>()`) don't appear to be enforced by any CI pass at the head ref: `src/tsconfig.json` and `tsconfig.build.json` both exclude `src/**/*.test.ts`, and `src/vitest.config.ts` enables `typecheck` without an `include` override, so Vitest's typecheck only picks up `*.test-d.ts` files — of which `src/` has none. At runtime the first test body is `if (false) {}` and passes vacuously, and `expectTypeOf` is a no-op. If the "focused Vitest typecheck" in the PR description ran with a custom include, consider moving these assertions to a `.test-d.ts` file (or adding `typecheck.include`) so the regression guard actually bites. Same applies to the type-only lines in `TransportKeyPair-p.test.ts` (that one's mock-call assertions are real runtime checks, so it still earns its keep either way).

2. Minor, `sdk/js-sdk/src/solana/clients/decorators/decrypt.ts:84-86` — the new "TKMS version is unresolved after client initialization" error duplicates the intent of core's `UNRESOLVED_TKMS_VERSION_MESSAGE`; reusing the constant would keep the failure mode greppable. Not blocking.

No behavior changes to proof bytes, chain IDs, or ACL semantics were found in the diff, consistent with the PR description; the remaining `.ready`/`init` casts flagged as quarantined are indeed still forced by the public `FhevmExtension` contract.

PR #3089, opened 2026-07-10
refactor(coprocessor): remove dead Solana emit tests

Eikix:
## What

- Remove the host-listener's unused ZamaHost and confidential-token emitted-event decoders after #3131 made confirmed Yellowstone reconstruction the sole coprocessor ingress.
- Remove the two on-chain `test_emit_*` instructions and their obsolete ignored worker/event tests.
- Retain one real-TFHE stitched boundary that reconstructs instruction data, computes the result, and decrypts it.
- Generate only the protocol value types still consumed by reconstruction.

## Architecture boundary

This cleanup preserves #3131's confirmed, stateless Yellowstone reconstruction and direct PBS writes. It does not restore the standalone RPC listener, finalized-account queue, witness records, or lineage tracking.

Ordinary operation semantics continue to come from transaction instructions and account state. The narrow produced-public lifecycle event needed by the relayer is a follow-up and is not introduced here.

## Validation

- `cargo test -p zama-host --features poc --lib` (82 passed)
- focused host-listener adapter tests (11 passed)
- `SQLX_OFFLINE=true cargo test -p tfhe-worker --lib confidential_transfer_reconstructs_computes_and_decrypts --no-run`
- `cargo fmt -p host-listener -p tfhe-worker -- --check`
- `cargo fmt --all -- --check` in the Solana workspace
- repository pre-commit checks: full coprocessor `cargo check` and CI-aligned clippy
- structural scan confirmed no finalized fetcher, finalized queue, witness/lineage tracker, or emitted-event decoder was restored

The retained worker boundary requires PostgreSQL and a real TFHE execution environment to run to completion; this repair compile-checked that boundary and leaves its terminal execution to the stack's full vertical validation.

## Stack and exclusions

- Based directly on `feature/solana` after #3122 and #3131.
- Does not include or alter #3083 cleartext work.
- Does not implement the produced-public lifecycle event; that remains the next stack layer.

Tracking: zama-ai/fhevm-internal#1676


Eikix:
_Automated review — generated by a Claude (Fable) review agent dispatched by the PR author; findings verified against the stacked base `origin/elias/solana-remove-emit-listener` (#3086) and the PR head._

---

**Scope reviewed:** the single stacked commit `cb0f36da6`. +38/−2941 over 20 files.

## Overview

This PR completes the cleanup that #3086 started. With the production emit-decode listener gone, it removes:

1. **Host-listener emit-decoding ingress** (`solana_adapter.rs`, −1097): `decode_anchor_*`, `decode_solana_transaction_events`, `merge_solana_transport_events`, the log-stack parser, the entire `ConfidentialTokenEvent` fetch-scheduling path, `SolanaEventDecodeError`, and their unit tests. `build.rs` stops generating `confidential_token_events.rs`; `generated/mod.rs` makes `zama_host_events` a private module and re-exports only the event structs still needed by reconstruction.
2. **The circular parity test** in `solana_grpc_listener.rs`, replaced by `multi_instruction_reconstruction_produces_protocol_facts`.
3. **The ignored tfhe-worker PoC module** (`solana_poc.rs`, −1584) and the two on-chain `test_emit_*` instructions plus their shared `assert_test_shim_authority` guard.
4. Docs (`DESIGN_DECISIONS.md` DD-014/DD-028, `EVM_PARITY.md`, README) and IDL doc strings updated to match.

## Correctness

### On-chain program change — verified safe

- **No dangling references.** Grepped the head commit for every deleted symbol (`test_emit_trivial_encrypt`, `test_emit_fhe_rand`, `TestEmitProtocolEvent`, `assert_test_shim_authority`, all deleted adapter functions, `confidential_token_events`, camelCase `testEmit`): zero hits outside docs describing the removal. On the base branch, all references were confined to the three files this PR deletes/edits.
- **IDL/discriminators.** The vendored `zama_host.json` instruction list is **identical** between base and head (18 instructions; `test_emit_*` were never in it, because `check-zama-host-idl.sh` builds the production IDL without the `poc` feature). The full IDL diff is doc strings only. Anchor instruction discriminators are `sha256("global:<name>")`-based, so removing instructions cannot shift the remaining ones — and since the removed instructions were `#[cfg(feature = "poc")]`, the default production build's instruction set is byte-identical anyway. No redeployment-visible ABI change.
- **Error codes preserved.** `TestShimsDisabled` / `TestShimAuthorityMismatch` remain in `errors.rs` even though nothing raises them anymore. That's the right call: Anchor error codes are order-indexed, so deleting them would renumber every later variant. A `// retained for error-code stability` comment would help the next person who notices they're unraised, but it's optional.
- **HostConfig layout untouched — confirmed.** `host_config.rs` and `state/mod.rs` change only doc comments; `test_authority`, `test_shims_enabled`, `mock_input_enabled` fields all remain. `test_shims_enabled` still gates `zero_birth_entropy_allowed()` (`host_config.rs:79`), matching the new doc strings exactly.
- **Doc-claim check:** `EVM_PARITY.md` now says `set_mock_input_enabled` is `#[cfg(feature = "poc")]` — verified true in `instructions/mod.rs`.
- **`plan_contracts.rs`:** removing the two entries from `production_poc_paths_are_compile_feature_gated_and_runtime_rejected` is correct cleanup. Note `cfg_gated_symbol` returns `true` for absent symbols, so the entries would have passed vacuously either way — the removal is cosmetic, not load-bearing.

### The replacement test — the crux

Assessment: **the new test is at least as strong as what it replaces, and the "circular oracle" claim in the PR body is accurate.**

What the deleted `emit_and_reconstruct_multi_instruction_fact_sets_match` actually compared: its "emit side" (`emit_mode_events_for_multi_instruction_tx`) was **not** an independent execution of the on-chain program. It hand-encoded one CPI `FheBinaryOpEvent` and then built every fetch fact by calling **the reconstruction module's own helpers** (`reconstruct_handle_superseded_fetch`, `decode_encrypted_value_fetch_events`, `reconstruct_acl_record_bound_fetch`). So `assert_eq!(emit_facts, reconstruct_facts)` was substantially reconstruction-vs-itself; a systematic bug in the fetch taxonomy would have passed both sides. The only independent content was the hand-built binary-op payload roundtripped through the emit decoder.

The new test states the six expected facts explicitly and adds `assert_eq!(reconstruct_facts.len(), required_facts.len())` plus containment of all six distinct facts — which, since the facts are distinct, is exact multiset equality. That is *more* exhaustive than the old required-facts loop, and the expected output handle is still anchored to the program's real derivation via `zama_host::state::computed_eval_handle` (`derived_add_output_handle`), so handle derivation is checked against the on-chain crate, not a copy.

What is genuinely no longer checked: there is **no independent decoder oracle at all** anymore — the semantic-fact list is hand-maintained, so if reconstruction and the expectations drift together (e.g. someone updates both in one PR), nothing external catches it. The true independent oracle was always the LiteSVM end-to-end tests, and those were already `#[ignore]`d. This is an accepted consequence of #3086's direction (fhevm-internal#1676), not a defect of this PR — but it should be stated in the PR body as the residual risk, and it raises the value of an eventual live-validator e2e harness on the reconstruct path.

### Relayer / MMR consumer

Verified independent: `relayer/src/solana_proof/decode.rs` defines its **own** `ANCHOR_EVENT_IX_TAG_LE` constant and its own decode — it never imported the host-listener's deleted decoder. The on-chain `emit_cpi!` protocol emissions (`emit-events` feature, default-on) are untouched. The `EVENT_VERSION` lockstep check in `check-zama-host-idl.sh` still works because `build.rs` retains its `EVENT_VERSION` constant. One micro-nit: `relayer/src/solana_proof/decode.rs:131` doc-comments "host-listener's `ANCHOR_EVENT_IX_TAG_LE`", which is no longer a public host-listener export — stale but harmless, and outside this diff.

## Test coverage delta

Of the −2941 lines:

- **Genuinely dead/circular (no real loss):** the emit-decoder unit tests in `solana_adapter.rs` (log-stack parsing, mixed-transport rejection, version gating, confidential-token fetch scheduling, log-line size bound) — they tested code deleted in the same commit. The parity test — replaced by a stronger explicit contract. 4 of 5 `solana_poc.rs` tests were `#[ignore]`d (LiteSVM + real worker + Postgres) and not running in CI.
- **Real, non-ignored coverage deleted:** `solana_worker_replay_shape_preserves_eval_dependencies_and_eager_schedules_compute` in `solana_poc.rs` was **not** ignored and exercised the *retained* `normalize_solana_events_for_db`: log-index assignment across 4 op kinds, eager `is_allowed=true`, `dependence_chain`, and — uniquely — `tfhe_inputs_handle` dependency extraction per op. The retained adapter tests (`builds_existing_db_log_shape`, `compute_is_eager_regardless_of_same_tx_allow_signal`, `normalizes_interleaved_eval_frame_events_for_worker_replay`) cover most of this, but no equivalent assertion of `tfhe_inputs_handle` per-op input extraction was found. Worth either porting that one assertion block into the adapter tests or confirming it's covered elsewhere.
- **Deferred oracle loss:** the ignored LiteSVM end-to-end tests (real ciphertexts, decrypt round-trips, log-size bounds) were the only program-executes → listener-ingests → worker-computes pipeline tests. They were unexercised, so deleting them removes clutter, not CI coverage — but the capability gap they papered over remains open.

## Conventions / hygiene

- **Leftover dev-dependencies (should fix):** with `solana_poc.rs` gone, `tfhe-worker/Cargo.toml` dev-deps `litesvm`, `anchor-lang`, `anchor-spl`, `solana-sdk`, `confidential-token`, `zama-host`, and `k256` are now unused (verified by grep over `src/` and `benches/`; only `host-listener` is still used by benches). Removing them cuts a nontrivial dependency tree from every tfhe-worker test build.
- `#[allow(dead_code)] mod zama_host_events`: `build.rs` still generates the full decoder (`decode_anchor_event`, `decode_anchor_cpi_event`, `anchor_event_discriminator`, `ANCHOR_EVENT_IX_TAG_LE`) that nothing consumes, silenced rather than trimmed. Acceptable for a stacked cleanup; consider trimming the codegen in a follow-up so dead decode logic doesn't linger behind an allow.
- The `generated/mod.rs` module doc ("Consumers should import this module instead of parsing Anchor event bytes directly") is now slightly stale — no decode functions are exported anymore.
- IDL doc-string edits are consistent with the source doc-comment edits (same strings in `host_config.rs`/`state/mod.rs` and both vendored IDLs), consistent with a real `check-zama-host-idl.sh` run. The `confidential_token.json` doc changes come from the shared zama-host types, which explains why that IDL changed without the token program's source being touched.

## Risks

- **Stacked merge ordering:** base is `elias/solana-remove-emit-listener` (#3086, itself a draft targeting `feature/solana`). This PR must not be retargeted to `feature/solana` before #3086 lands, or the diff balloons to include #3086's changes. Standard stack discipline applies.
- **Program redeployment:** none in practice — the removed instructions were `poc`-feature-only, so production artifacts, discriminators, HostConfig layout, and error codes are all unchanged. Local PoC/dev tooling that invoked `test_emit_*` (if any exists outside this repo) would break, but nothing in-repo references them.
- CI on the head commit was still in progress at review time; the validation list in the PR body is consistent with what was verified statically.

## Verdict

**Approve** (once #3086 lands and CI is green). The PR does what it says, the dangerous-sounding parts (on-chain instruction removal, IDL) are verifiably no-ops for deployed programs, and the "circular oracle" rationale checks out against the deleted test's construction. Highest-value suggestions, in order:

1. **Remove the now-unused tfhe-worker dev-dependencies** (`litesvm`, `anchor-lang`, `anchor-spl`, `solana-sdk`, `confidential-token`, `zama-host`, `k256`) — real build-time win, and this PR is the natural place since it deleted their only consumer.
2. **Port the `tfhe_inputs_handle` per-op dependency assertions** from the deleted (non-ignored) `solana_worker_replay_shape_...` test into the retained adapter tests, unless equivalent coverage can be pointed to.
3. Add a one-line comment on `TestShimsDisabled`/`TestShimAuthorityMismatch` noting they're retained for Anchor error-code stability, and note in the PR body that the residual gap is the absence of any independent oracle for reconstruction (tracked toward a live-validator e2e harness).

Eikix:

No blocking issues found; deletion honesty verified, one informational note and one trivial nit below.

**Deletion honesty (verified).** Grepped the whole tree at the head ref for every removed symbol — `decode_anchor_event`/`decode_anchor_cpi_event(s)`/`decode_anchor_log_events`, `decode_solana_transaction_events`, `merge_solana_transport_events`, `decode_solana_transaction_account_fetches`, `decode_confidential_token_anchor_*`, `ConfidentialTokenEvent`, `ZamaHostEvent`, `ANCHOR_EVENT_IX_TAG_LE`, `anchor_event_discriminator`, `SolanaEventDecodeError`, `test_emit_trivial_encrypt`, `test_emit_fhe_rand`, `TestEmitProtocolEvent`, `assert_test_shim_authority` — and found no live references. The two apparent matches are independent: `token_mollusk.rs:124` defines its own local generic `decode_anchor_event`, and `relayer/src/solana_proof/decode.rs:132` deliberately redeclares its own tag constant. Also confirmed at the stacked base (`elias/solana-remove-emit-listener`) that the deleted adapter decoders had no production callers — only the deleted test modules used them — so nothing removed here was still on the ingestion path after #3086.

**Surviving coverage.** The retained boundary test (`confidential_transfer_reconstructs_computes_and_decrypts`) is a genuine upgrade over the deleted ones: it reconstructs the plan from the real CPI instruction data, cross-checks reconstructed step results against on-chain `EncryptedValue.current_handle` reads (account state, not emitted logs), and validates end-to-end via real TFHE decrypt (25/120). The renamed gRPC test tightened from set-comparison against a hand-forged emit path to exact-length plus containment of an explicit fact contract (`solana_grpc_listener.rs:1521`). The deleted `kms_like_user_decrypt_check` was a test-local reimplementation of KMS verification (it validated its own mock); the real independent re-verification lives in kms-worker.

1. **Informational — producer-less fetch kinds.** After this PR, `SolanaFinalizedAccountFetchKind::DisclosureRequest` and `BurnRedemptionRequest` (`coprocessor/fhevm-engine/host-listener/src/solana_adapter.rs:33-34`) have no producer anywhere in the repo — `request_witness_fetch` was their last one, and it was already test-only at the base. Retaining them for persisted-DB-enum stability matches the PR description; worth making sure the quarantined schema work tracks either reintroducing a reconstruction-side producer for disclosure/burn-redemption flows or retiring the kinds.

2. **Nit.** `tfhe-worker/src/tests/mod.rs`: `mod solana_poc;` moved from the alphabetized module list to below `setup_test_key` at the bottom of the file — looks like an accidental placement during the edit; compiles fine but worth restoring to the module block.

Eikix:
*(review drafted by Claude)*

**Verdict: approve.** This is a well-scoped, mechanically-consistent deletion PR. Every removal traces to a real dependency (the emitted-event decode path made obsolete by #3131's Yellowstone reconstruction), the retained real-TFHE test boundary was verified to still compile and exercise the intended path, and the doc/IDL comment updates were cross-checked against the code they describe rather than taken at face value. One small pre-existing-style doc nit was found (non-blocking). Base branch is `feature/solana` (stacked), not `main` — this is not mergeable to main standalone.

## What the PR does

- Removes the Anchor emitted-event decoder generator (`generate_anchor_events` → `generate_anchor_event_types`) from `host-listener/build.rs`, and with it `ANCHOR_EVENT_IX_TAG_LE`, `anchor_event_discriminator`, `decode_anchor_event`/`decode_anchor_cpi_event`, the `ZamaHostEvent` enum, and the hand-rolled `Cursor` byte reader — all now dead since #3131 made Yellowstone-based instruction reconstruction the only coprocessor ingress.
- Removes the corresponding `solana_adapter.rs` decode/merge machinery (`decode_solana_transaction_events`, `decode_anchor_cpi_events`, `decode_anchor_log_events`, `merge_solana_transport_events`, `SolanaEventDecodeError`, log-scope-stack helpers) and ~15 unit tests that only existed to exercise that machinery.
- Removes the two on-chain `test_emit_fhe_rand` / `test_emit_trivial_encrypt` shim instructions from `zama-host` (`instructions/*.rs`, `lib.rs`, `mod.rs`), their `TestEmitProtocolEvent` account struct, and `assert_test_shim_authority`.
- Gates the remaining Anchor `emit!`/`emit_cpi!` call sites behind `#[cfg(feature = "emit-events")]` (adding a no-op stub for `emit_config_updated` when the feature is off), since those events are no longer consumed by anything off-chain.
- Retains exactly one real-TFHE, real-Postgres integration test (renamed `confidential_transfer_reconstructs_computes_and_decrypts`), rewritten to reconstruct the `fhe_eval` plan via `decode_fhe_eval_args`/`reconstruct_fhe_eval_events` instead of decoding emitted events, and prunes the large amount of now-dead fixture/helper code that only the deleted tests used (`kms_like_user_decrypt_check`, `signed_user_decrypt_request`, `seed_authorizing_encrypted_value`, `HostFixture`, `bob: Keypair` field, etc.).
- Updates `zama_host.json` / `confidential_token.json` IDL doc strings, `errors.rs`, `state/host_config.rs`, `state/mod.rs`, `README.md`, and `solana/docs/{DESIGN_DECISIONS,EVM_PARITY,FUTURE_DESIGN,TESTING}.md` to describe `test_authority`/`test_shims_enabled` as retained-for-layout-stability rather than shim-authority fields, and updates `runtime-tests/tests/plan_contracts.rs` and `scripts/check-zama-host-idl.sh` accordingly.

## Correctness checks performed

- **`test_shims_enabled` doc rename** ("Enables the local zero-birth-entropy fallback"): verified against `DESIGN_DECISIONS.md` DD-014, which spells out `zero_birth_entropy_allowed() = test_shims_enabled && is_local_poc_chain()`. Since the field's only remaining consumer is that gate (the `test_emit_*` shim-authority check that used to also read it was deleted), the renamed doc is accurate, not stale.
- **`test_authority` / `TestShimsDisabled` / `TestShimAuthorityMismatch`**: `test_authority` is no longer read anywhere in the diff (`assert_test_shim_authority`, its only reader, was deleted); the two error variants are likewise never constructed anymore. Both are kept as "reserved" rather than deleted — correct, since deleting an `HostConfig` struct field or renumbering an Anchor error enum are both ABI/discriminant breaks; the PR explicitly calls this out (README: "quarantined until an explicit account-ABI break").
- **Test fixture pruning for compile-safety**: `TokenFixture.bob: Keypair` was removed; confirmed every remaining reference is to `output.bob`/`outputs.bob` (a `Pubkey` field on the separate `TransferOutputAccounts` struct), not the removed `Keypair`, so this is not a leftover dangling reference. `solana-tests/build-and-test` and `solana-tests/host-listener-reconstruct` both passed for the head commit, corroborating this.
- **Retained test coverage narrowed slightly**: the surviving `confidential_transfer_reconstructs_computes_and_decrypts` test drops the previous per-output `kms_like_user_decrypt_check`/`output_type` assertions (that helper and the ACL-invariant test that used it were deleted along with the shim-based tests it depended on). The plan-shape and decrypted-value assertions (`events.len() == 5`, per-step `SolanaHostEvent` matches, decrypted values `"25"`/`"120"`) are preserved. This is a real, if modest, reduction in assertion surface for this test — acceptable given the PR's stated scope (dead-code removal, not new coverage), but worth knowing this test now checks less than before.
- **Minor doc bug (found via diff, confirmed against the file at the PR head commit `a7f88218`)**: in `tfhe-worker/src/tests/solana_poc.rs`, `fn coprocessor_signing_key()` now has a truncated doc comment — `/// Coprocessor signing key backing the \`fromExternal\` amount attestations; its EVM address is the` — trailing off mid-sentence, because the diff deleted the doc's continuation line (`` /// `coprocessor_signer` configured on the fixture's `host_config`.``) but left the first line intact. Confirmed present at HEAD via `raw.githubusercontent.com`. Cosmetic only (doc comment on a private test helper), does not affect behavior or compilation.

## CI

- `gh pr checks` shows one `FAILURE` in the status rollup: `coprocessor-cargo-test/cargo-tests (bpr)`. Investigation of that job's log shows it is a gating/summary step whose own body just checks `$CHECK_CHANGES_RESULT`/`$CARGO_TESTS_RESULT`/`$COVERAGE_RESULT` and exits 1 because those upstream jobs were `cancelled` (not `failure`) — i.e. this looks like a stale run superseded by a later push, not a genuine test failure. The current head commit's other coprocessor jobs (`sns-worker`, `common`, `cargo-clippy`) show `pass`, and Solana-specific jobs (`solana-tests/build-and-test`, `solana-tests/host-listener-reconstruct`) both `pass`. `solana-e2e/full-vertical-reconstruct` was still `pending` at review time. Given the ambiguity, this should be double-checked before merge rather than assumed benign, but there is no evidence in the retrieved logs of an actual compile/test regression from this diff.
- PR is `mergeable: true`, `mergeable_state: unstable` (reflects the above pending/ambiguous checks), not a draft.
- Base branch is `feature/solana`, per both `gh pr view` (`baseRefName`) and the PR body ("Based directly on `feature/solana` after #3122 and #3131") — confirming this is a stacked PR, not targeting `main`.

## Security considerations

- Removing the `test_emit_*` shims eliminates an admin-gated but still-present path for forging protocol events on a running validator; this shrinks attack surface (one less privileged instruction pair on `zama-host`), consistent with the PR's stated defense-in-depth rationale in `DESIGN_DECISIONS.md`.
- Preserving `test_authority`/`test_shims_enabled` fields and the two now-unreachable error variants for ABI/account-layout stability is the right call — avoids a silent account-layout or error-discriminant shift for any already-deployed `HostConfig` accounts or client code matching error codes.
- No changes touch signature verification, PDA derivation, or admin-gating logic for any *live* instruction; the diff is a subtractive cleanup plus `#[cfg(feature = "emit-events")]` gating of already-existing `emit!`/`emit_cpi!` calls.

## Suggestions (non-blocking)

1. Fix the truncated doc comment on `coprocessor_signing_key()` in `coprocessor/fhevm-engine/tfhe-worker/src/tests/solana_poc.rs` (currently ends mid-sentence after the partial-line deletion).
2. Before merging, re-run or re-check `coprocessor-cargo-test/cargo-tests (bpr)` to confirm the `FAILURE` in the status rollup is in fact the stale/superseded run described above and not a live signal, since the underlying `cancelled` jobs it gates could not be inspected for pass/fail content directly.

PR #3090, opened 2026-07-10
test(test-suite): make Solana Mollusk flows instruction-first

Eikix:
Implements zama-ai/fhevm-internal#1695 by making the Mollusk confidential-transfer test read as seed → real instruction → execute → evaluate the captured host CPI → assert domain outcomes.

The test-only `CleartextLedger` binds durable clear results through canonical output descriptors and persisted `EncryptedValue.current_handle` values while production token, host, attestation, HCU, ACL/MMR, account, and event paths remain unchanged.

The contract is intentionally outcome-focused—Alice `600`, Bob `500`, transferred amount `400`, and exactly three durable outputs—without freezing the internal five-step plan or remaining-account ordering.

Validated with the focused transfer test, all 32 `token_mollusk` tests, formatting, and clippy with `-D warnings`.

### DIFF
--- a/solana/runtime-tests/tests/token_mollusk.rs
+++ b/solana/runtime-tests/tests/token_mollusk.rs
@@ -40,7 +40,10 @@ use anchor_lang::{
     Discriminator, InstructionData, ToAccountMetas,
 };
 use confidential_token as token;
-use mollusk_svm::{result::Check, Mollusk};
+use mollusk_svm::{
+    result::{Check, InstructionResult},
+    Mollusk,
+};
 use solana_sdk::{
     account::Account,
     instruction::{AccountMeta, Instruction},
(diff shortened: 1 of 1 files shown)


Eikix:

No blocking issues found; the rewrite preserves or strengthens assertion coverage. Notes below on what was specifically verified.

1. **Dropped index-based durable-handle mapping is safe.** The old test bound cleartext values via `output_encrypted_value_index` → `durable_handles_by_index`; the new `CleartextLedger::evaluate_fhe_cpi` (`solana/runtime-tests/tests/token_mollusk.rs:172`) instead derives the canonical PDA from the output descriptors and reads the persisted `current_handle`. This is equivalent: `admit_durable_output_account` in `zama-host/src/instructions/fhe_eval/admission.rs:81-91` already enforces `remaining_accounts[index] == encrypted_value_address(derive_value_key(descriptors))` (`EncryptedValuePdaMismatch`), so `Check::success()` still exercises the index↔descriptor binding on-chain, and the test no longer over-constrains remaining-account ordering.

2. **The only assertion genuinely removed is the exact 5-element intermediate output vector** (`[1, 600, 600, 400, 500]`). The durable outcomes it constrained (Alice 600, Bob 500, transferred 400, exactly 3 durable outputs) are all still asserted at `token_mollusk.rs:1101-1107`, and on this input the removed compare-bit assertion (`1`) could not have distinguished an inverted comparison anyway (400 ≤ 1000 and 400 ≤ 100 differ, but only the former is exercised). No insufficient-balance-path coverage existed before this PR either, so nothing regresses.

3. **Net strengthenings observed:** the CPI capture now filters inner instructions by `program_id == host::id()` before decoding (`token_mollusk.rs:180-188`; the old code `filter_map`-decoded any inner instruction whose bytes happened to parse); the vague `!result.inner_instructions.is_empty()` is replaced by "exactly one fhe_eval CPI" plus `durable_outputs == 3`; and `balance()` now asserts the full chain token account → `balance_encrypted_value` PDA → persisted `current_handle` → cleartext value, with `u64_at` additionally checking `fhe_type` and u64 range.

4. **Conventions:** assertions remain on account state, captured CPIs, and derived PDAs; the retained event checks decode `emit_cpi` inner instructions, unchanged by this PR. All lineage/MMR assertions (supersession leaf counts, subjects, peaks, stable PDAs) are untouched. The PR's claim of 32 `token_mollusk` tests matches the head ref.

Verified against `origin/elias/solana-mollusk-instruction-flows`; nothing blocking surfaced.

PR #3092, opened 2026-07-10
refactor(kms-connector): unify host-chain ACL registration

Eikix:
Unifies EVM and Solana host-chain ACL registration behind one closed backend table, validating duplicate IDs and missing or contradictory fields before backend construction.

Public, legacy EVM, RFC-016, and typed Solana requests now dispatch through the same lookup; unknown IDs remain ordinary bounded-recoverable errors, while configured backend-kind mismatches are terminal.

Removes the split Solana builder and the unused `EncryptedValueSubjectUseRoleMissing` error. `cargo test -p kms-worker --lib` passes 98/98, all-target clippy is clean, and GitHub CI completed the KMS connector tests plus the full Solana vertical successfully; addresses zama-ai/fhevm-internal#1681.

### DIFF
--- a/kms-connector/config/kms-worker.toml
+++ b/kms-connector/config/kms-worker.toml
@@ -167,8 +167,9 @@ chain_id = 12345
 # live event-processing path still needs KMS Core native work-item dispatch and tx-sender/relayer
 # publication.
 chain_kind = "evm"
-# Address of the ACL contract (required for EVM chains, ignored for Solana chains)
+# Address of the ACL contract (required for EVM chains, forbidden for Solana chains)
 # Format: hex string with 0x prefix
 acl_address = "0x0000000000000000000000000000000000000000"
-# Required for Solana chains. Base58 ZamaHost program id expected to own ACL/material witnesses.
+# Required for Solana chains and forbidden for EVM chains. Base58 ZamaHost program id expected to
+# own ACL/material witnesses.
 # solana_host_program_id = "11111111111111111111111111111111"
--- a/kms-connector/crates/kms-worker/src/core/config.rs
+++ b/kms-connector/crates/kms-worker/src/core/config.rs
@@ -168,8 +168,8 @@ pub struct HostChainConfig {
     pub acl_address: Option<Address>,
     /// The expected ZamaHost program id for Solana-owned ACL/material witnesses.
     ///
-    /// Required for Solana ACL verification. If omitted for a Solana chain, the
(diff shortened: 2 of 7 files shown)


Eikix:

No blocking issues found; one deployment-compat note and one minor observation below.

Verification performed against `origin/elias/solana-unified-kms-registration`:

- The KMS-side independent re-verification invariant is intact: `check_user_decryption_request_solana` still runs the pure ed25519 binding check (`verify_solana_user_decrypt_signature`) followed by `check_solana_handles_acl` against the finalized-commitment fetcher, and public decrypt still requires the `PublicDecryptLeaf` MMR proof path. Only the backend lookup changed.
- The terminal "requires EVM" rejection in `check_user_decryption_request_v2` is safe: `processor.rs:179-218` shows RFC-016 (`UserDecryptionV2`) is EVM-only by construction, with Solana handled by its own typed `UserDecryptionSolana` arm — so no legitimate Solana flow routes through the V2 check.
- Deleted `EncryptedValueSubjectUseRoleMissing`: grepped the whole tree at head, zero remaining references.
- EVM behavior for EVM-configured chains is unchanged across all four request families (same calls, same recoverable/irrecoverable classification); the only EVM-visible delta is the unknown-chain error message text.
- No DB migrations are touched; the change is config/startup-scoped only.
- In-repo configs satisfy the new validation: helm `charts/kms-connector/values.yaml` (EVM-only, `aclAddress` set), `test-suite/fhevm/src/generate/solana.ts` (omits `acl_address` for Solana entries), and the staging/template `.env.kms-connector` files.

Findings:

1. Deployment-compat note (kms-connector/crates/kms-worker/src/core/kms_worker.rs:687-729): two previously-tolerated configurations now crash the worker at startup instead of degrading — a Solana chain without `solana_host_program_id` (formerly a documented fail-closed runtime mode, per the old `config.rs` comment) and a Solana chain with a leftover `acl_address` placeholder. All in-repo configs are clean, but any live environment config outside this repo carrying either shape will crashloop on upgrade. Worth confirming before this reaches a deployed environment.

2. Minor (kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs:100-110): a Solana user-decrypt for an entirely unconfigured chain id was previously irrecoverable (terminal fail-closed) and is now recoverable, unified with the EVM convention. Authorization is still never granted, and retries are bounded elsewhere in the stack (#3085), so this looks benign — flagging only because it inverts the old explicit "fail closed at lookup time" comment. If terminal rejection was deliberate for unknown Solana chains, that intent is now gone.

PR #3093, opened 2026-07-10
fix(sdk): fix extraData version check

geoxel:
This occurs when contracts have been upgraded but not the KMS yet.


0xalexbel (approved):
LGTM!

PR #3097, opened 2026-07-10
fix(relayer): preserve Solana replay gaps and pin account order

Eikix:
Stops Solana proof ingestion at the first temporarily unavailable transaction, preserving commits before the gap while preventing both the program cursor and lineage seen-set from advancing past it.

Recovery tests prove restored transactions resume oldest-first without duplicate events, and the existing poll loop retries the typed transient error on its next interval.

The Solana ABI checker now includes each pinned instruction's complete ordered Anchor account tree; exactly 22 instruction schema hashes change while fixture bytes, IDLs, and every non-instruction schema remain identical.

This is a bounded partial implementation of zama-ai/fhevm-internal#1682 and #1683; backfill, HTTP retry classification, endpoint controls, deduplication, finalized-read freshness, and shared-decoder design remain explicitly out of scope.

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/idl/solana_abi_golden.json
+++ b/coprocessor/fhevm-engine/host-listener/idl/solana_abi_golden.json
@@ -140,7 +140,7 @@
       "fixture_len": 8,
       "name": "close_consumed_burn_redemption_request",
       "program": "confidential_token",
-      "schema_hash": "38cbf36bec48f4dd45ce7d1ed99e68c51d155f8667153d517a22aec11c74aace"
+      "schema_hash": "4e43d15dc32003a909f462def5e296d296accf6f1dc4cf1f57000e5ced954bee"
     },
     {
       "category": "instruction_args",
@@ -148,7 +148,7 @@
       "fixture_len": 8,
       "name": "close_consumed_disclosure_request",
       "program": "confidential_token",
-      "schema_hash": "f9c95168d2dfb2ccc22b49ea3d483d752bfbacae262bec52faacd96e3f5fbc0f"
+      "schema_hash": "ec705c2ebc58f892de2f05e6de037a55adc37ef00f617705d7815163c56b84f7"
     },
     {
       "category": "instruction_args",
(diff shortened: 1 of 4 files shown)


Eikix:

No blocking issues found; one operational question below. The fix is sound: both ingestion paths now fail closed at the first unavailable transaction instead of silently skipping it, and per-signature commits (`apply_signature_update` bundles events + replay state + seen-set + cursor in one `update_state` call, `relayer/src/solana_proof/store.rs:351`) mean everything before the gap survives and nothing after it is applied out of order.

Verified specifically:

- The "exactly 22 instruction schema hashes" claim holds: the golden manifest contains exactly 22 `instruction_args` entries (15 confidential_token + 7 zama_host), all changed, and no account/event/type hash or fixture byte moved.
- `RpcChainFetcher::get_transaction` (`relayer/src/solana_proof/chain.rs:299`) returns `Some` with empty instructions for failed transactions, so the new `TransactionUnavailable` error only fires on genuinely absent transactions, not on-chain failures.
- The account-order pin is real: `canonical_json` uses `sort_keys` (object keys only — array order is preserved), and `strip_docs` runs before hashing, so doc-only IDL edits won't churn hashes while account reorders will.
- The poll loop (`relayer/src/solana_proof/poller.rs`) logs and retries on the next interval, and `catch_up_lineage` errors surface as failed proof requests — both transient-safe.
- Tests assert on store state (events, cursor, seen signatures), matching the account-state-not-logs convention, and genuinely exercise the recover-oldest-first and no-duplicate paths.

Findings/questions:

1. **Permanent-gap stall visibility** (`relayer/src/solana_proof/ingest.rs:275`, `poller.rs:69`). If a transaction is permanently unavailable from the configured RPC (e.g. a load-balanced endpoint where one node lacks the slot, or history pruned on a non-archival node), ingestion now halts at the gap forever, surfacing only as a repeated `error!` line per poll interval. Meanwhile the backlog behind the gap keeps growing, so a long-enough stall converts into `BacklogExceeded` — requiring a manual backfill even after the transaction reappears. The PR description scopes retry classification and backfill out (fhevm-internal#1682), so this is not a defect here, but: is there a metric/alert planned on repeated `TransactionUnavailable` for the same signature, so a permanent stall gets noticed before proof requests start failing? Worth a tracking pointer in the error message or the follow-up issue.

2. Minor: the new `plan_contracts.rs:386` guard pins the exact Python source substring `"accounts": instruction.get("accounts", [])` — any harmless refactor of that line breaks the test without an ABI change. Consistent with the file's existing style, so fine to keep; a hash-manifest-based assertion would be sturdier if this pattern grows.

PR #3099, opened 2026-07-10
test(test-suite): add shardable Solana operator conformance

Eikix:
Implements the fast cleartext conformance slice of zama-ai/fhevm-internal#1696 with 379 individually named, Cargo-filterable operator/type/operand tests and a test-owned closed-world admission contract.

The new target covers every shipped positive tuple, arbitrary-type admission drift, width and encoding edges, precise rejection causes, and durable, verified, local, and scalar operand sources while replacing the previous 10-test broad evaluator target.

This PR is deliberately Part A: it does not claim Mollusk account/CPI/persistence, real-TFHE, or full-stack conformance, so zama-ai/fhevm-internal#1696 remains open for those tiers.

Validated with the full 379-test target (~0.5s warm), all Solana runtime-test targets, formatting, clippy with warnings denied, ABI/IDL checks, event-version checks, and pre-commit hooks.


Eikix:

No blocking issues found; one coverage gap in the closed-world sweep and one stale comment worth fixing.

1. **Ternary admission is checked against a test-owned mirror, not the production validator** — `operator_conformance.rs:1637` (`ternary_is_accepted`) drives the closed-world ternary leg through `evaluate()`, which re-implements the ternary type check inside `support/cleartext_fhe_eval.rs` ("Mirrors the host-private `fhe_eval::assert_ternary_operand_types`"). The binary and unary legs call the exported production validators (`assert_binary_operand_types` / `assert_unary_operand_type`) directly, so admission drift in the program fails the test; a change to the host-private `assert_ternary_operand_types` (`programs/zama-host/src/instructions/fhe_eval.rs:282`) would not be caught by this target — the mirror and the program would drift silently until a Mollusk-tier test notices. Consider exporting the ternary validator like the other two so all three closed-world legs exercise production code.

2. **Stale cross-reference in the support module** — `solana/runtime-tests/tests/support/cleartext_fhe_eval.rs:113` still says "keep the malformed-ternary cases in `cleartext_fhe_eval.rs` aligned with that validator", but this PR deletes that test file; the cases now live in `operator_conformance.rs` (`rejected::ternary`). Worth updating the pointer since the comment is the only thing documenting the mirror relationship flagged in (1).

Verified specifically:

- **No silent operator drop**: the closed-world sweep compares the test-owned matrix against the production binary/unary validators across the full 256×256 type space, and `expected_binary`/`expected_unary`/`expected_ternary` use exhaustive `match` over the opcode enums (no wildcard), so adding a new opcode is a compile error in this target, not a silent gap. `expected_binary`/`expected_unary` were cross-checked line-by-line against `assert_binary_operand_types`/`assert_unary_operand_type` at the head ref and match (including Div/Rem scalar-only + zero-after-truncation, Eq/Ne accepting u160, Cast rejecting same-type).
- **Old coverage carried over**: every assertion in the deleted `cleartext_fhe_eval.rs` has an equivalent or stronger replacement (widening MulDiv, u256 high-half preservation, sum/is-in caps now tested at both the exact cap and cap+1, bool encoding edges, seed-type mismatch), with error expectations tightened from generic strings to precise `ZamaHostError` names.
- **CI wiring**: grepped the whole tree at the head ref — no remaining references to the deleted test target in workflows, scripts, or docs; `solana-tests.yml` runs `cargo test --workspace` (and clippy `--all-targets`), so the new target runs unfiltered and no shard list exists today that could drop tests.
- **Evidence honesty**: the module doc and PR body both scope this as cleartext test-evidence only (Mollusk/TFHE/full-stack deferred), consistent with the fhevm-internal#1696 tiering.

PR #3100, opened 2026-07-10
chore(coprocessor): streamline sqlx cli setup

Eikix:
Removes unused host-side SQLx CLI installs from CPU/GPU benchmark and GPU-test workflows, while keeping migrations inside the maintained container path.

Pins the two active migration images and the relayer cache guard to SQLx CLI 0.8.6, matching the workspace dependency and rejecting stale cached binaries.

Deletes the unreferenced worker database recreation script and replaces obsolete host CLI setup documentation with `make init_db`.

Validated workflow parsing, exact/missing/stale version-guard behavior, the migration Make target, locked SQLx CLI feature installation, and reference searches; addresses zama-ai/fhevm-internal#1061.


Eikix:

No blocking issues found; one minor doc inconsistency worth folding in.

**Verification performed at the head ref:**

- **Removed workflow installs are safe.** All three workflows that lost the host `cargo install sqlx-cli` step never invoke `sqlx` elsewhere: both benchmark workflows migrate via `make init_db` (`coprocessor/fhevm-engine/tfhe-worker/Makefile:14`), which runs the `db-migration` docker-compose service; `coprocessor-gpu-tests.yml` runs `cargo test` with `SQLX_OFFLINE=true`, and the test harness applies migrations via the embedded `sqlx::migrate!` macro (`coprocessor/fhevm-engine/test-harness/src/instance.rs:169`) — no CLI needed at runtime.
- **Deletion honesty confirmed.** `git grep recreate_db` across the tree finds only the Makefile's own `recreate_db` target and its README mention — nothing references the deleted `tfhe-worker/scripts/recreate_db.sh`. No `sqlx-cli --version 0.7.2` pins remain anywhere (workflows, Dockerfiles, Makefiles, docs).
- **Version alignment.** The 0.8.6 pin in `Dockerfile.workspace:126` and `db-migration/Dockerfile:10` matches the workspace `sqlx = "0.8.6"` (`coprocessor/fhevm-engine/Cargo.toml:71`) and the relayer's (`relayer/Cargo.toml:70`). The exact-match guard in `relayer-tests.yaml:176` is correct — `sqlx --version` prints `sqlx-cli X.Y.Z`, and `cargo install` replaces a differently-versioned cached binary without `--force`.

**Findings:**

1. Minor — `coprocessor/README.md:56` still instructs `cargo install sqlx-cli` unpinned. After this PR it is the only unpinned install instruction left, and per the rationale this PR itself cites in `relayer-tests.yaml:179` (sqlx-cli 0.9.0 raised MSRV to rustc 1.94 while `rust-toolchain.toml` pins 1.91.1), following it verbatim can fail to build or install a CLI that mismatches the workspace's 0.8.6. Suggest pinning it to 0.8.6 here too, or pointing at `make init_db` like the tfhe-worker README now does.

2. Note, non-blocking — bumping the CLI version now requires touching three literals in `relayer-tests.yaml` (guard string at line 176, echo at line 183, install at line 181). A single env var would keep them from drifting, but this is optional hygiene.

Eikix:
Not yet needed

PR #3101, opened 2026-07-10
test(test-suite): source Solana e2e proofs from relayer

Eikix:
Terminal acceptance slice for zama-ai/fhevm-internal#1675. This PR also partially addresses fhevm-internal#1658.

## Outcome

Make the Solana full-vertical e2e source every historical and public MMR proof from the running relayer. Delete local proof builders, proof-source switches, and builder-only tests while preserving client-side verification, wire encoding, and proof reuse.

## Decisive invariant

The burned lineage must contain exactly two public leaves:

1. leaf 0: the entropy-derived `make_public=true` output recovered from the host lifecycle batch introduced by merged #3136 and validated by merged #3138
2. leaf 1: the later explicit `make_handle_public` performed by `request_disclose_amount`

The client requests `leaf_count - 1` and the terminal flow requires `leaf_count == 2`, requested/returned `leaf_index == 1`, and a proof that authorizes the burned handle against the live account. This prevents the flow from passing if lifecycle-batch ingestion omitted the first leaf.

The mode-free inclusion-proof bytes are captured once and reused unchanged by both redeem and disclose.

## Retry and failure behavior

- retry only HTTP `503` with a structured JSON `status=lagging`
- bound retries to 15 attempts with a 2-second delay inside the proof fetch
- fail immediately on `corrupt_cache`, other HTTP statuses, malformed bodies, transport errors, unverified responses, or response-index mismatch
- preserve non-JSON and invalid-UTF-8 HTTP error bodies as lossy raw diagnostics
- never rerun a stateful live-client operation to hide relayer lag

## Evidence before full vertical

- live-client formatting and tests, including structured-lagging and raw-error-body regressions
- `bash -n` for `full-vertical.sh`
- all 65 focused relayer Solana-proof tests passed on merged #3138
- forbidden local-builder/source-selection symbols absent
- independent adversarial reviews of #3136 and #3138 found no merge-blocking issue

## Merge gate

- base: `feature/solana`, including merged producer #3136 and relayer consumer #3138
- terminal gate: `solana-e2e/full-vertical-reconstruct` on this exact head with a fresh validator and fresh relayer store

This PR remains draft until that run proves both public leaves are served exclusively by the relayer and the same fetched proof succeeds in both on-chain consume paths.

### DIFF
--- a/solana/scripts/poc/full-vertical.sh
+++ b/solana/scripts/poc/full-vertical.sh
@@ -549,7 +549,7 @@ echo "$relout" | grep -q 'OK request_disclose_amount' || fail "request_disclose_
 echo "    disclosure request witness created (KMS context pinned); handle released for public decrypt"
 
 # Public-decrypt the burned handle -> cleartext + KMS PublicDecryptVerification cert. This lineage
-# must contain the born-public lifecycle leaf followed by the explicit make_handle_public leaf; a
+# must contain the produced-public lifecycle leaf followed by the explicit make_handle_public leaf; a
 # proof for a one-leaf lineage would not exercise lifecycle-batch ingestion.
 run_public_decrypt_with_proof "burned" "$BURNED_HANDLE" "$BURNED_ACL" "" 2 1
 cr="$PUBLIC_DECRYPT_JSON"
--- a/solana/scripts/poc/full-vertical.sh
+++ b/solana/scripts/poc/full-vertical.sh
@@ -44,15 +44,14 @@ run_public_decrypt_with_proof() {
   local handle="$2"
   local acl="$3"
   local expected="${4:-}"
-  # Proof source (5th arg, default relayer): the e2e sources the MMR proof from the running relayer
-  # proof service. `local` is reserved for the born-public burned leg until the relayer consumes
-  # the host's lifecycle batch, and drives the in-process PoC builder instead. The client retries
-  # a transient `503 lagging` internally (re-invoking here would be unsafe for stateful steps).
-  local proof_source="${5:-relayer}"
+  local expected_leaf_count="${5:-}"
+  local expected_leaf_index="${6:-}"
+  # The e2e sources MMR proofs from the running relayer proof service. The client retries a transient
+  # `503 lagging` internally (re-invoking here would be unsafe for stateful steps).
   local proof
   proof="$(cd "$ROOT/solana/scripts/poc/live-client" && \
     PUBLIC_DECRYPT_PROOF=1 PUB_HANDLE="$handle" PUB_ACL="$acl" \
-    PROOF_SOURCE="$proof_source" RELAYER_URL=http://127.0.0.1:3000 \
+    RELAYER_URL=http://127.0.0.1:3000 \
     ./target/debug/poc-live-client 2>&1)" \
     || fail "$label public proof: $proof"
   echo "$proof" | grep -E 'PUB H|PUB mmrProofBytes' >/dev/null || fail "$label public proof missing fields: $proof"
@@ -74,6 +73,14 @@ run_public_decrypt_with_proof() {
     [ -n "${!required}" ] || fail "$label public proof missing $required: $proof"
   done
   [ "$pub_h" = "$handle" ] || fail "$label public proof handle $pub_h != $handle"
+  if [ -n "$expected_leaf_count" ]; then
+    [ "$pub_leaf_count" = "$expected_leaf_count" ] \
+      || fail "$label public proof leaf_count $pub_leaf_count != $expected_leaf_count"
+  fi
+  if [ -n "$expected_leaf_index" ]; then
+    [ "$pub_leaf_index" = "$expected_leaf_index" ] \
+      || fail "$label public proof leaf_index $pub_leaf_index != $expected_leaf_index"
+  fi
   PUBLIC_DECRYPT_INCLUSION_PROOF_BYTES="$pub_mmr_inclusion_proof_bytes"
 
   local result
@@ -216,7 +223,7 @@ done
 # here; the client retries a transient `503 lagging` internally, so this is not re-invoked on lag.
 hist_proof="$(cd "$ROOT/solana/scripts/poc/live-client" && \
   HISTORICAL_STEP=supersede TE_VALUE="$VALUE" \
-  PROOF_SOURCE=relayer RELAYER_URL=http://127.0.0.1:3000 \
+  RELAYER_URL=http://127.0.0.1:3000 \
   ./target/debug/poc-live-client 2>&1)" || fail "historical supersede/proof command failed: $hist_proof"
 echo "$hist_proof" | grep -E 'HIST H_new|HIST mmrProofBytes' || fail "historical supersede/proof: $hist_proof"
 HIST_H_OLD2="$(hist_field "$hist_proof" H_old)"
@@ -541,12 +548,10 @@ relout="$(lc CONSUME_SEAL=1 TS_ACL="$BURNED_ACL" TS_HANDLE="$BURNED_HANDLE" \
 echo "$relout" | grep -q 'OK request_disclose_amount' || fail "request_disclose_amount witness: $(echo "$relout" | tail -3)"
 echo "    disclosure request witness created (KMS context pinned); handle released for public decrypt"
 
-# Public-decrypt the burned handle -> cleartext + KMS PublicDecryptVerification cert.
-# PROOF_SOURCE=local: unlike the compute/input-flow/historical legs (relayer-sourced), the
-# born-public burned handle is derived on-chain from slot entropy and is not carried in instruction
-# data. The host now emits it in a narrow lifecycle batch, but this leg stays on the in-process PoC
-# builder until the immediately stacked relayer slice validates and consumes that batch.
-run_public_decrypt_with_proof "burned" "$BURNED_HANDLE" "$BURNED_ACL" "" local
+# Public-decrypt the burned handle -> cleartext + KMS PublicDecryptVerification cert. This lineage
+# must contain the born-public lifecycle leaf followed by the explicit make_handle_public leaf; a
+# proof for a one-leaf lineage would not exercise lifecycle-batch ingestion.
+run_public_decrypt_with_proof "burned" "$BURNED_HANDLE" "$BURNED_ACL" "" 2 1
 cr="$PUBLIC_DECRYPT_JSON"
 # The burned handle's public-decrypt MMR proof (DD-036): both the redeem and disclose consume
 # steps authorize by verifying it against the lineage's on-chain peaks. request_burn_redemption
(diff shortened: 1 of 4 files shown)


Eikix:

One blocking issue: the decisive gate fails deterministically on exactly the leg this PR moves onto the relayer path.

1. **Burned born-public leg cannot be relayer-sourced in the emitless CI arm.** `solana/scripts/poc/full-vertical.sh:542` drops `PROOF_SOURCE=local` for the burned handle, but the only remaining `solana-e2e` arm deploys zama-host EMITLESS (`setup-solana-side.sh:101` builds `--no-default-features`, dropping default-on `emit-events`). With no `emit_cpi!` inner instruction on chain, the relayer cannot resolve the born-public handle (its own `decode.rs:255` comment: "Without the op events … fails closed at proof time"), so its reconstruction diverges from the live peaks. The CI run shows this concretely: `solana-e2e/full-vertical-reconstruct` fails at `FAIL: burned public proof:` after 7 lagging retries with `HTTP 500 {"verified":false,"status":"corrupt_cache","leaf_count":2}` — deterministic fail-closed, not a flake. This is precisely the scenario the comment deleted at `full-vertical.sh:545` documented ("in the emitless arm the relayer cannot resolve it over RPC"). The PR body's "the relayer now reconstructs from its emitted operation event" only holds for emit-enabled builds; the relayer's 52 proof tests construct op-event inner instructions synthetically, so their passing is not evidence for the emitless e2e. The relayer follow-up (sysvar reconstruction or a verified handle hint) needs to land before this deletion, or the burned leg keeps a fallback until then.

2. Minor: `main.rs:730` (post-diff) now uses the last-leaf (`leaf_count - 1`) assumption for the burned lineage too. For the current flow (two historical supersede leaves + one trailing public leaf) that is correct, and the in-client `authorize_public` fail-fast makes any mismatch loud rather than silent — no action needed, noted for when the lineage shape changes.

Verified specifically: `ProofSource`/`PROOF_SOURCE`, `build_historical_access_proof`, `build_public_decrypt_proof`, and `locate_public_decrypt_leaf_index` have zero remaining references anywhere at the head ref (code, scripts, workflows, docs). No proof-verification coverage silently disappears: the deleted unit tests exercised only the deleted local builders, the client keeps its fail-fast `authorize_public`/`mmr_verify` checks against the live account on the relayer path, cross-mode rejection (historical leaf vs public verify) remains covered structurally by domain-separated leaf commitments plus negative tests in `zama-solana-acl/src/lib.rs` (`exact_public_no_roll_forward`, `no_history_rejects_proofs_but_current_works`) and the kms-connector `solana_user_decrypt` accept/reject suite, and the twice-public reconstruction regression now lives in `relayer/src/solana_proof/proof.rs`.

Eikix:
## Review — relayer-only e2e proofs

Net −271 lines. Deletes the in-process PoC proof builders (`build_historical_access_proof`, `build_public_decrypt_proof`), the `PROOF_SOURCE` switch, and their four builder-only unit tests; hardens the relayer fetch path (structured error-body parsing, precise 503-lagging retry match, response `leaf_index` validation) and adds the decisive `leaf_count == 2 / leaf_index == 1` assertion on the burned leg.

### What checks out

- **Shell wiring**: the old 5th positional arg (`proof_source`) is gone from all call sites; only the burned call passes the new args 5/6 (`2 1`). The assertions bind to the right client output — `PUB leafCount` is the live account's `leaf_count`, `PUB leafIndex` the proof's `leaf_index`.
- **Retry semantics match the description**: only a `503` with JSON `status == "lagging"` retries (bounded 15 × 2s); `corrupt_cache` and everything else fail immediately with a distinct message. The `starts_with("relayer proof lagging:")` match is tighter than the old `contains("lagging")`.
- **`dto.leaf_index != leaf_index` check** is a good addition — a wrong-leaf answer previously only failed later at `authorize_public` with a confusing verification error.
- **Test deletion is consistent**: all four removed tests exercised only the deleted local builders; `proof_blob`, `MMR_MODE_*`, and the client-side fail-fast verification paths are retained.

### One real finding — diagnostic regression for non-JSON error bodies

The relayer's `IntoResponse` (`relayer/src/solana_proof/http.rs`) returns a JSON `MmrProofResponse` body **only** for `Lagging` (503) and `CorruptCache` (500). Every other error — invalid address (400), `LeafIndexOutOfRange` (400), `LineageNotFound` (404), other proof errors (500) — returns **plain text** via `other.to_string()`. The new client code unconditionally parses error bodies as `RelayerMmrProofResponse`, so those cases now surface as:

```
relayer proof HTTP 404 returned an invalid error body: <serde error>
```

swallowing the server's actual message, which the old raw-string path preserved. Failure behavior is still correct (immediate, no retry), but debugging a red e2e run gets harder. Suggested fix — read the body once as a string, then try JSON with a raw-text fallback:

```rust
let text = resp.into_string().unwrap_or_default();
match serde_json::from_str::<RelayerMmrProofResponse>(&text) {
    Ok(b) if code == 503 && b.status == "lagging" => { /* lagging error */ }
    Ok(b) => { /* status/leaf_count error */ }
    Err(_) => return Err(format!("relayer proof HTTP {code}: {text}").into()),
}
```

### Minor notes (no change required)

- **Stringly-typed retry signal**: editing the message text in `fetch_relayer_mmr_proof_once` silently breaks the retry match in `fetch_relayer_mmr_proof`. A small typed error enum (`Lagging { leaf_count }` vs `Fatal(String)`) would decouple them; acceptable for a PoC client.
- **Redundant shell assertion**: since the client requests `leaf_count − 1` and verifies the returned `leaf_index`, `expected_leaf_index=1` is implied by `expected_leaf_count=2`. Fine as explicit belt-and-braces in a test.
- **Unreachable loop tail**: `Err("relayer proof still lagging after retries")` can never be reached — the last attempt returns its error directly via the `attempt < MAX_ATTEMPTS` guard. Harmless.

### Verdict

Solid cleanup slice — the deletion is clean, the invariant assertion is meaningful, the retry policy precise. Worth fixing the error-body fallback before merge, since it degrades exactly the diagnostics needed when the terminal `solana-e2e` gate fails.

*Posted by Claude Code on behalf of @Eikix.*

Eikix:
## Review

### Overview

Terminal slice of the relayer-only-proofs stack: deletes the in-process ("local") MMR proof builders and the `PROOF_SOURCE` switch from the PoC live client, making the relayer proof service the sole proof source for all e2e legs — including the born-public burned leg, which previously stayed on the local builder. In exchange it hardens the relayer fetch path: structured 503-lagging detection (replacing a substring match), a response-echo `leaf_index` guard, readable non-JSON error bodies, and shell-level assertions that the burned lineage has exactly `leaf_count == 2` / `leaf_index == 1`. Net −221 lines.

### Correctness — verified against surrounding code

- **Wire shapes match.** The client's `RelayerMmrProofResponse` mirrors the relayer's `MmrProofResponse` (`relayer/src/solana_proof/http.rs:59-65`), and the relayer's 503-lagging and 500-corrupt_cache bodies serialize every field the client's deserializer requires (`mmr_proof: None` is fine — serde defaults missing/null `Option` to `None`). The relayer's "other" error arm returns plain text, which correctly falls through to the lossy-string fallback in `relayer_http_error`.
- **Retry tightening is sound.** Old code retried any error whose message *contained* "lagging" — a false-retry hazard if e.g. an unrelated body mentioned the word. New code retries only when the message starts with `"relayer proof lagging:"`, which is produced solely for HTTP 503 + JSON `status == "lagging"`. A hypothetical 200 with `status=lagging` would fail fast, but the relayer never emits that (200 is always `status: "verified"`), so behavior is consistent with the stated policy.
- **`leaf_index` echo check** is a good defense — a relayer that silently served a different leaf would previously only fail later (or worse, verify against the wrong leaf position).
- **Shell call sites are safe.** The 5th positional arg changed meaning (`proof_source` → `expected_leaf_count`); all four `run_public_decrypt_with_proof` call sites on the PR head were checked — only the burned leg passes args 5/6, the rest pass ≤4. No stale `relayer`/`local` argument left behind.
- **The burned-leg invariant is enforced at two layers**: the client requests `leaf_count - 1` and verifies via `authorize_public` against live peaks; the script separately pins `leaf_count == 2, leaf_index == 1`. This genuinely fails the e2e if lifecycle-batch ingestion drops the first leaf — the acceptance property the PR claims.
- `String::from_utf8_lossy` + shadowed `&str` borrow in `relayer_http_error` is valid Rust (shadowing doesn't drop the `Cow`).

### Suggestions (all minor, none blocking)

1. **Load-bearing string prefix with no compile-time link.** The retry predicate `e.to_string().starts_with("relayer proof lagging:")` and the format string in `relayer_http_error` must stay in sync, but nothing ties them together — the unit test asserts the formatted string, not that the retry loop classifies it as retryable. Cheapest fix: a shared `const LAGGING_PREFIX: &str` used by both the formatter and the predicate. A typed error (small enum implementing `Error`) would be cleaner still, but is arguably overkill for a PoC client.
2. **Dead fallback after the retry loop** (pre-existing): `Err("relayer proof still lagging after retries")` is unreachable — on the final attempt the `attempt < MAX_ATTEMPTS` guard fails and the plain `Err(e)` arm returns first. Could be `unreachable!()` or just deleted; the actual final error (which includes `leaf_count`) is the more useful one anyway.
3. **Untested guards**: the new `leaf_index`-mismatch check and the retry-loop classification have no unit coverage (the tests only exercise `relayer_http_error`'s string output). Fine for a PoC, but the mismatch guard is a one-line test away.
4. **Unbounded `read_to_end` on error bodies** — harmless against the local test relayer; would matter only if this client ever pointed at an untrusted endpoint, which the file header explicitly forbids (RPC pinned to local validator).
5. `run_public_decrypt_with_proof` is now up to six positional args with a `""` placeholder at the burned call site — at the edge of readable for bash. If a seventh arg ever appears, switch to env-var-style named parameters like the function already uses internally.

### Test coverage

Deleting the builder tests alongside the builders is right — they tested code that no longer exists, and the regression they guarded (twice-public lineage reconstruction) is now covered end-to-end by the shell `leaf_count/leaf_index` pins instead. New tests cover 503-lagging vs corrupt_cache classification, non-JSON bodies, invalid UTF-8, and empty bodies. The declared merge gate (fresh-validator `solana-e2e` dispatch on this head, PR stays draft until then) is appropriate for a terminal acceptance slice.

### Security

Nothing concerning: test-only tooling, RPC pinned to localhost, and the change *strengthens* verification (client still runs `authorize_public`/`mmr_verify` against live chain state before handing any relayer-sourced proof to KMS or on-chain consume steps, so a compromised/buggy relayer can't smuggle an unverified proof through the e2e).

### Verdict

Solid PR — deletion-heavy, invariant-driven, with the trust boundary (client-side verification) preserved. The only real nit is the string-prefix coupling between `relayer_http_error` and the retry predicate (suggestion 1); everything else is polish. Looks good to merge pending the `solana-e2e` gate already declared in the description.

PR #3102, opened 2026-07-10
fix(sdk): reject unsupported Solana decrypt batches

Eikix:
Rejects zero- and multi-handle Solana user-decrypt requests before handle conversion, transport-key generation, signing, or relayer access, matching the connector's exact-one contract.

Deletes the now-unreachable proof-specific cardinality branch and updates the public documentation to state the one-handle invariant while retaining the array-shaped wire-compatible API.

The focused Solana SDK suite passes 4/4 and proves invalid counts perform neither transport-key work nor fetches; formatting and diff checks are clean.

This deliberately leaves ACL-key derivation and encrypt submission semantics for separate API decisions; partially addresses zama-ai/fhevm-internal#1684.

### DIFF
--- a/sdk/js-sdk/src/solana/actions/userDecrypt.ts
+++ b/sdk/js-sdk/src/solana/actions/userDecrypt.ts
@@ -40,7 +40,7 @@ export type SolanaUserDecryptContext = {
 };
 
 export type SolanaUserDecryptParameters = {
-  /** The ciphertext handles to decrypt (each a 32-byte handle). */
+  /** The single 32-byte ciphertext handle to decrypt, encoded as a one-element array. */
   readonly handles: readonly EncryptedValueLike[];
   /** Override the chain's ACL domain keys for this request (each a bytes32 0x-hex). */
   readonly allowedAclDomainKeys?: readonly Bytes32Hex[] | undefined;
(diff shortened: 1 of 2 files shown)


Eikix:

No blocking issues found; the guard placement, branch deletion, and test evidence all check out. Two non-blocking observations below.

Verified specifically:

- **Single enforcement point covers all entry points.** The only route to a Solana user-decrypt is `createFhevmDecryptClient` → `solanaDecryptActions` decorator (`sdk/js-sdk/src/solana/clients/decorators/decrypt.ts:82`) → the `userDecrypt` action, and the guard sits at the top of the action (`sdk/js-sdk/src/solana/actions/userDecrypt.ts:194`) before handle conversion, `generateSolanaTransportKeyPair`, signing, and the relayer POST. No other caller of the action exists at the head ref; the encrypt path is untouched.
- **The guard matches the connector contract.** `require_single_handle` in `kms-connector/.../event_processor/solana_user_decrypt.rs:175` rejects anything but exactly one handle, so the SDK previously let zero/multi-handle requests burn a full sign-and-submit round trip only to fail server-side. Fail-fast here is the right fix.
- **Deletion honesty.** The removed proof-cardinality branch (`mmrProof` + `handleBytes.length !== 1`) is genuinely unreachable once `handles.length !== 1` throws first — confirmed by reading the full function at the head ref.
- **Test honesty.** The new test asserts both `generateSolanaTransportKeyPair` and `fetch` are not invoked on zero- and two-handle inputs, which matches the PR's "no transport-key work or fetches" claim; the module is properly `vi.mock`ed and the mock is cleared per-test. The e2e suite (`test-suite/fhevm/solana-userdecrypt-full.ts:75`) already passes a one-element array, so no existing caller breaks.

Observations (non-blocking):

1. `sdk/js-sdk/src/solana/actions/userDecrypt.ts:71-73` — the `mmrProof` JSDoc still says "so `handles` must have length 1 when this is set", which now reads as a proof-specific constraint rather than the universal invariant. Worth rewording to point at the one-handle contract generally, since the proof-specific rejection it referenced was deleted.
2. In the decorator, `await fhevm.ready` runs before the action, so an invalid-count call still triggers TKMS module init (WASM load) even though the request is doomed. Harmless (init is needed for any subsequent valid call), just noting the guard is not literally the first work performed.

PR #3103, opened 2026-07-10
test(test-suite): add representative Solana operator conformance

Eikix:
## Summary

Add a thin, independently filterable Mollusk layer for representative Solana `fhe_eval` admission families.

Each success test:

1. constructs one canonical `FheEvalArgs` plan,
2. executes the real compiled `zama_host` program under Mollusk,
3. evaluates that exact accepted plan with the test-owned cleartext evaluator,
4. asserts a readable domain outcome, and
5. independently verifies the complete emitted handle, event version, subject, and random seed metadata.

The target covers encrypted/encrypted, encrypted/scalar, Bool comparison output, Cast, non-Cast unary, membership, random, and bounded-random families. Three focused negative cases cover mismatched operand types, invalid durable-account ownership, and a read-only durable output.

## Layering

- PR #3099 remains the exhaustive pure semantic/operator contract. This PR deliberately does not repeat its matrix under Mollusk.
- PR #3090 independently owns the confidential-token instruction-first transfer/CPI flow and durable cleartext ledger.
- Existing `host_mollusk` tests remain responsible for durable output creation/supersession, ACL mutation, HCU admission/metering, signer enforcement, and rollback.
- This branch is based directly on `feature/solana`; it does not include either PR.

This advances zama-ai/fhevm-internal#1696 Phase 2 without claiming the later real-TFHE or full-stack phases are complete.

## Validation

- `NO_DNA=1 cargo test -p zama-solana-runtime-tests --test operator_mollusk_conformance`
  - 11 passed in 0.03s locally
- `NO_DNA=1 cargo test --workspace`
- `NO_DNA=1 cargo clippy --workspace --all-targets -- -D warnings`
- `NO_DNA=1 cargo fmt --all -- --check`
- `git diff --check`

### DIFF
--- a/solana/runtime-tests/tests/operator_mollusk_conformance.rs
+++ b/solana/runtime-tests/tests/operator_mollusk_conformance.rs
@@ -1,0 +1,750 @@
+//! Representative real-host conformance for Solana `fhe_eval` operator families.
+//!
+//! Each test executes one canonical instruction against the compiled `zama_host` program, then
+//! evaluates that exact plan in the test-owned cleartext evaluator. The exhaustive semantic
+//! contract stays in `operator_conformance`; this target covers only materially different host
+//! admission and result-binding shapes.
+
+mod support;
+
+use std::{collections::HashMap, path::PathBuf, sync::Once};
+
+use anchor_lang::{
+    prelude::system_program, AccountSerialize, AnchorDeserialize, Discriminator, InstructionData,
+    ToAccountMetas,
+};
+use mollusk_svm::{
+    result::{Check, InstructionResult},
+    Mollusk,
+};
+use solana_program::keccak::hashv as keccak_hashv;
+use solana_sdk::{
+    account::Account,
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+};
+use solana_sha256_hasher::hashv as sha256_hashv;
+use support::cleartext_fhe_eval::{evaluate, ClearInputs, TypedClearValue};
+use zama_host::{
+    self as host, FheBinaryOpCode, FheEvalArgs, FheEvalOperand, FheEvalOutput, FheEvalStep,
+    FheUnaryOpCode,
+};
+
+const CONTEXT_ID: [u8; 32] = [7; 32];
+// The PoC sentinel chain with test shims enabled deliberately substitutes zero birth entropy.
+const PREVIOUS_BANK_HASH: [u8; 32] = [0; 32];
+const UNIX_TIMESTAMP: i64 = 0;
+
+#[test]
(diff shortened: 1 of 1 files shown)


Eikix:

No blocking issues; one behavioral observation this PR pins down for the first time, plus two minor notes.

1. **Bounded-random seed does not bind the upper bound — and this test now locks that in** (`operator_mollusk_conformance.rs:181` / `expected_rand_seed` at :700). `bounded_random_executes_then_binds_bound_seed_and_type` asserts `event.seed == expected_rand_seed()`, i.e. the RandBounded seed is byte-identical to the plain Rand seed (`FHE_eval_seed` tag, no upper bound). That mirrors actual program behavior — `walk.rs:268` reuses `expected_rand_eval_seed` for RandBounded — but the program also ships `computed_eval_rand_bounded_seed` (`state/mod.rs:1256`, tag `FHE_eval_bounded_seed`, binds `upper_bound`), which a tree-wide grep shows is never called anywhere (its only occurrence is its definition, same on `feature/solana`). The result handle does bind the bound, so uniqueness holds; still, worth deciding whether the unused bound-binding derivation is the intended one (in which case this test cements the wrong seed) or dead code to delete. The test name "binds_bound_seed" slightly overstates — the bound is bound into the handle, not the seed.

2. **Entropy path coverage**: `mollusk()` (:436) installs a SlotHashes entry (`[9; 32]` at slot 99), but with `test_shims_enabled: true` the program returns the zero bank hash before reading SlotHashes (`previous_bank_hash_with_test_fallback`, `state/mod.rs:1348`), so every handle assertion in this target rides the zero-entropy shim branch and the sysvar entry is inert. Fine for a shim-chain conformance target (and the `PREVIOUS_BANK_HASH` comment says so), but the real SlotHashes-derived entropy path stays unexercised here — either drop the misleading sysvar setup or add one case with shims off.

3. **Assertion surface**: all positive-path result verification flows through `emit_cpi` inner-instruction payloads (`event()` at :598). That is the allowed inner-CPI surface, not log assertions — but it deepens the test suite's dependence on `emit_cpi`, whose removal is tracked (fhevm-internal#1676). No action needed now; just noting the coupling this adds.

Verified specifically: the test's hand-mirrored handle derivations (binary keccak `FHE_eval`, unary/is_in sha256 `FHE_eval_unary`/`FHE_eval_is_in`, rand `FHE_comp`) match `state/mod.rs` / `fhe_eval/handles.rs` at the head ref, including the Cast-only type-byte inclusion; the `support/cleartext_fhe_eval` module exists on the base branch (this PR genuinely doesn't depend on #3099); cleartext claims are labeled as cleartext throughout — no TFHE evidence is implied; the three negative tests assert exact custom error codes, not just failure.

PR #3104, opened 2026-07-10
refactor(coprocessor): delete unused host-chain address types

Eikix:
Deletes the unused Solana `HostChainAddress` and `HostChainAuxData` abstraction, its re-export, and five tests that exercised only that dead island.

Runtime boundaries already use chain-discriminated EVM and Solana types directly; a repository-wide symbol search found no instruction, account, event, IDL, SDK, listener, or coprocessor consumer.

The change removes 212 lines with no replacement; 101 `zama-host` tests, warnings-denied clippy, formatting, and the canonical Solana IDL/schema/ABI drift check pass.

This supersedes the implementation direction in zama-ai/fhevm-internal#1220 and #1221 without introducing a bare cross-chain `B256` type.


Eikix:

No blocking issues found; the deletion claim was independently verified and holds.

**Deletion verification.** A tree-wide `git grep` at the head ref for `HostChainAddress`, `HostChainAuxData`, `host_chain_address`, `HOST_CHAIN_ADDRESS_LEN`, `HOST_CHAIN_AUX_LEN`, and `EVM_ADDRESS_LEN` — across Rust, the vendored IDL golden (`coprocessor/fhevm-engine/host-listener/idl/solana_abi_golden.json`), sqlx/SQL, docs, CI YAML, and TOML — found zero remaining references. The only textual hits are the unrelated TypeScript `renderHostChainAddresses*` env-file helpers in `test-suite/fhevm/src/generate/addresses.ts`, which predate these types and share nothing but the name. On the base branch, the types were referenced only by the three files this PR touches, so the "dead island" characterization is accurate: no instruction, account, event, IDL, SDK, listener, or coprocessor consumer existed.

**RFC-021 context.** The deletion does not orphan the RFC-021 bytes32 surface. The 128-byte auxiliary layout (`contract || user || acl_contract || chain_id`, addresses left-padded to bytes32) remains implemented and tested where it is actually consumed: `coprocessor/fhevm-engine/zkproof-worker/src/auxiliary.rs` (`SIZE_SOLANA = 128`, with `assembles_solana_bytes32_aux` and negative tests) and the SDK proof-building path (`sdk/js-sdk/src/solana/actions/encryptInput.ts`). The on-program `VerifiedInput` path checks the coprocessor EIP-712 `CiphertextVerification`, not the aux bytes, so the program never needed this type — the deleted duplicate risked drifting from the canonical zkproof-worker encoding rather than protecting it.

**Tests.** The five removed tests exercised only the deleted type's own round-trips and layout; equivalent layout coverage exists in the zkproof-worker test module noted above, so no assertion coverage of live code is lost. The retained tests in `solana/programs/zama-host/src/state/tests.rs` (handle-derivation chain-type bit, keccak rand) are untouched.

One observation, not blocking: the deleted module's doc comment was the only in-tree prose explaining the 92→128-byte widening rationale from the Solana side; `auxiliary.rs` carries a terser version. If DESIGN_DECISIONS.md ever grows an RFC-021 entry, that rationale would be worth preserving there.

PR #3105, opened 2026-07-10
test(test-suite): add representative operator smoke

Eikix:
Adds a serial-by-default standard profile with four explicit encrypted operator flows: encrypted addition, scalar division, comparison, and modular unary negation.

Each case creates an external input proof, executes through the coprocessor, makes the result publicly decryptable, and asserts the readable result; a distinct grep namespace keeps the tests out of the generated heavy operator lane.

The exhaustive 105-file stress catalog remains unchanged, and documentation distinguishes warm-stack feedback from cold build/bootstrap cost and records provenance-correct commands without claiming an unmeasured SLO.

Hardhat compilation, CLI typechecking, and 226 CLI tests pass; the standard full-stack lane is the decisive runtime gate and this partially addresses zama-ai/fhevm-internal#1295.

### DIFF
--- a/test-suite/e2e/contracts/smoke/OperatorSmoke.sol
+++ b/test-suite/e2e/contracts/smoke/OperatorSmoke.sol
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+pragma solidity ^0.8.24;
+
+import "@fhevm/solidity/lib/FHE.sol";
+import {E2ECoprocessorConfig} from "../E2ECoprocessorConfigLocal.sol";
+
+contract OperatorSmoke is E2ECoprocessorConfig {
+    ebool public boolResult;
+    euint32 public uint32Result;
+    euint64 public uint64Result;
+
+    function add(
+        externalEuint64 lhs,
+        externalEuint64 rhs,
+        bytes calldata inputProof
+    ) external {
+        uint64Result = FHE.add(FHE.fromExternal(lhs, inputProof), FHE.fromExternal(rhs, inputProof));
+        FHE.makePubliclyDecryptable(uint64Result);
+    }
+
+    function div(externalEuint64 lhs, uint64 rhs, bytes calldata inputProof) external {
+        uint64Result = FHE.div(FHE.fromExternal(lhs, inputProof), rhs);
+        FHE.makePubliclyDecryptable(uint64Result);
+    }
+
+    function lt(
+        externalEuint64 lhs,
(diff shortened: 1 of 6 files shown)


Eikix:

No blocking issues found; assertions were verified as genuinely end-to-end, with two non-blocking observations below.

Verified specifically:

- All four assertions are non-tautological: each test drives an external input proof through the coprocessor and compares the publicly decrypted result against an independent cleartext constant (`test/operatorSmoke/operatorSmoke.ts:31,46,62,79`). Nothing asserts on logs or on handle-vs-handle identity, and the `neg` expectation `2^32 - 42` is the correct euint32 modular result.
- API usage matches both instance backends at the head ref: `encryptTypedValues` and `publicDecrypt(...).clearValues` exist with identical shapes in `test/sdk/fhevm-sdk/sdk.ts:222,245` and `test/sdk/relayer-sdk/sdk.ts:62,198`, and boolean clear values are asserted as `true` consistent with `fhevmOperations1.ts:414`.
- Grep namespace: `git grep "representative operator smoke"` at the head ref hits only the new test, the CLI test, and `layout.ts` — no collision with the generated catalog. Mocha full names start with `Operator smoke representative operator smoke:`, which the `operators` grep (`test operator|FHEVM manual operations`) does not match, so the smoke set stays out of the heavy lane and vice versa.
- Serial-by-default holds: `TEST_PARALLEL["operator-smoke"]` is absent, and `test.ts:1234` (`options.parallel ?? TEST_PARALLEL[name]`) therefore runs it serially; the CLI test pins this plus standard-only membership.

Observations (non-blocking):

1. Representativeness gap — no select/mux case. The four cases cover binary arithmetic, scalar-operand, comparison-to-ebool, and unary families across two widths, which is reasonable for a smoke. The most distinct family left out is `FHE.select`/ternary (encrypted control flow: ebool condition consumed by a data op), which exercises a different operand shape than any of the four. Worth considering as a fifth case, or a stated reason for exclusion, since "representative" is the PR's claim.
2. The non-overlap guard in `cli.test.ts:165-167` is one-directional and tests the title fragment: it asserts the `operators` grep does not match `"representative operator smoke"`, but not that the `operator-smoke` grep avoids the generated catalog (currently true — verified by grep — but unguarded), and mocha greps full names including the `Operator smoke` describe prefix, so a future `operators` pattern matching the prefix would slip past this assertion. A full-name fixture (`"Operator smoke representative operator smoke: ..."`) on both directions would make the guard airtight.

Eikix:
Should be merged on main it's a operator EVM

PR #3106, opened 2026-07-10
refactor(sdk): remove dead Solana signer adapter

Eikix:
Deletes the final loose Solana signer launcher from the SDK root after the live input, user-decrypt, and public-decrypt harnesses moved to `test-suite/fhevm` and public SDK exports.

The removed adapter imported private `_esm` internals and has no caller on the current branch; its only verified historical use was an ignored helper in a closed reference KMS PR.

The change removes 53 lines with no replacement; frozen dependency installation, harness typechecking, 225 tests, public launcher resolution, reference searches, and diff checks pass.

This completes the remaining implementation residue in zama-ai/fhevm-internal#1519.


Eikix:

No blocking issues found; the deletion claim holds up under independent verification.

Deletion-honesty check performed at `origin/elias/remove-dead-sdk-signer-adapter`:

1. `git grep` for `solana-userdecrypt-sign` (and the looser `userdecrypt-sign`) across the entire head tree — including `.github/` workflows, `test-suite/`, `solana/scripts/`, and docs — returns zero references. Nothing invokes the deleted script.
2. The `UD_*` env-variable names shared with the deleted script do still appear in `solana/scripts/poc/full-vertical.sh:192,241` and `test-suite/fhevm/solana-userdecrypt-full.ts`, but both consume them via `bun run solana-userdecrypt-full.ts` (the pure-SDK harness), not via the deleted `.mjs` — so those are the successors the PR description mentions, not missed callers.
3. The deleted file was the only importer of the built `_esm` output (`src/_esm/core/coprocessor/SolanaUserDecrypt-p.js`); the underlying `buildSolanaUserDecryptRequest` module remains and its surviving consumers (`sdk/js-sdk/src/solana/index.ts`, `actions/userDecrypt.ts`, `userDecryptPreimage.test.ts`, `test/fheTest/solana/userDecrypt.test.ts`) all import the TypeScript source, so removing the script leaves no dangling import path.
4. `sdk/js-sdk/package.json` has no `files` field or script referencing the `.mjs`, so the published package contents and build pipeline are unaffected.

The diff is a pure single-file deletion against `feature/solana` with no stacked-base concerns.

PR #3107, opened 2026-07-10
test(relayer): pin Solana proof request identity

Eikix:
Pins the Solana user-decrypt deduplication contract: otherwise-identical requests with different proof-bearing `extraData` now have different content hashes and job identities, and remain distinct from the EVM request variant.

Removes stale guidance suggesting the relayer could inject an MMR proof after receiving a request; Solana proof discovery and encoding must happen before the Ed25519 signature is produced.

Also corrects EIP-712-only handler wording because v3 conversion selects either the EVM or Solana unified request variant.

The focused test, complete eight-test content-hash module, formatting, and warnings-denied relayer clippy pass; this corrects part of zama-ai/fhevm-internal#1636 without changing runtime behavior.

### DIFF
--- a/relayer/src/core/user_decrypt_request.rs
+++ b/relayer/src/core/user_decrypt_request.rs
@@ -182,7 +182,8 @@ mod tests {
     use crate::core::event::{
         HandleContractPair, HandleEntry, RequestValidity, RequestValiditySeconds,
     };
-    use alloy::primitives::{Address, Bytes, U256};
+    use crate::core::job_id::JobId;
+    use alloy::primitives::{Address, Bytes, B256, U256};
 
     fn sample_legacy_direct() -> UserDecryptRequest {
         UserDecryptRequest::LegacyDirect {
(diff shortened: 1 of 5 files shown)


Eikix:

No blocking issues; one incomplete-cleanup miss and one test-sharpness note.

Verified against the head ref: the new test compiles against real APIs (`JobId` has `From<[u8; 32]>`, `relayer/src/core/job_id.rs:33`); the `solana_proof_extra_data` fixture matches the canonical v0x03 layout `0x03 ‖ context_id(32) ‖ acl_value_key(32) ‖ proof_slot(8 BE) ‖ proof_len(4 BE) ‖ proof` (`kms-connector/crates/utils/src/types/solana_extra_data.rs:186`); and the corrected handler comment matches the actual conversion — `TryFrom<AttestedUserDecryptRequestJson>` does construct `SolanaUnifiedV1` on the Solana attestation type (`relayer/src/core/event.rs:857`). The `SolanaUnifiedV1` content hash does cover `extra_data` and carries a distinct `variant:solana_unified_v1:` tag (`relayer/src/core/user_decrypt_request.rs:136-172`), so the test pins real behavior. The claimed eight-test module count is accurate.

1. **Stale guidance survives in `relayer/src/http/server.rs:182-184`.** The PR removes the "interim internal endpoint until the Solana user-decrypt path lands and calls `build_proof` in-process" narrative from `solana_proof/http.rs` and `solana_proof/mod.rs`, but the same claim remains verbatim where the router is mounted: `// (interim internal endpoint until the Solana user-decrypt path calls it in-process; see relayer/src/solana_proof/http.rs)`. That comment now points at a file that says the opposite (proof discovery must precede the Ed25519 signature, so in-process injection after receipt is impossible). Suggest updating it in this PR since removing exactly this guidance is the PR's stated purpose.

2. **The cross-variant assertion doesn't isolate the variant tag** (`relayer/src/core/user_decrypt_request.rs:375`). `assert_ne!(first_hash, sample_eip712_unified().content_hash())` compares a Solana request carrying a v0x03 proof blob against an EVM request with `extra_data = 0x00`, so it would pass even if the `variant:solana_unified_v1:` prefix were dropped — the differing `extra_data` alone separates them. Feeding both variants the same `extra_data` (e.g. `solana_proof_extra_data(0xaa)` on both) would make the "distinct from the EVM variant" pin actually depend on the tag. Minor, since the field-label domain separation also differs, but this is a pinning test — worth making the pin load-bearing.

3. Optional: unlike the legacy variants, `SolanaUnifiedV1` has no "excluded fields (`signature`, `request_validity`) don't affect the hash" test, so that half of its dedup contract is unpinned. Same pattern as `legacy_direct_excluded_fields_dont_affect_hash` would close it.

PR #3108, opened 2026-07-10
test: remove duplicate ignored test helpers

Eikix:
Removes two ignored tests that were dormant duplicates rather than unique manual coverage.

The relayer's exact EVM input-proof field assertions are merged into its maintained active conversion test, while the TFHE worker's ignored database key-seeding utility is deleted because the maintained `DBInstance` path invokes the same helper and is exercised by active cache tests.

No intentional Solana real-TFHE boundary, SNS integration mode, or stress fixture is changed; the patch is net −32 lines.

Canonical pre-commit formatting, workspace checks, both clippy passes, and reference searches pass.

### DIFF
--- a/coprocessor/fhevm-engine/tfhe-worker/src/tests/mod.rs
+++ b/coprocessor/fhevm-engine/tfhe-worker/src/tests/mod.rs
@@ -14,19 +14,3 @@ mod scheduling_bench;
 mod solana_poc;
 mod test_cases;
 mod utils;
-
-use test_harness::db_utils::setup_test_key as setup_test_key_in_db;
-
-#[tokio::test]
-#[ignore]
-/// setup test data with keys
-async fn setup_test_key() -> Result<(), Box<dyn std::error::Error>> {
-    let pool = sqlx::postgres::PgPoolOptions::new()
-        .max_connections(2)
-        .connect(&std::env::var("DATABASE_URL").expect("expected to get db url"))
-        .await?;
-
-    setup_test_key_in_db(&pool, false).await?;
-
-    Ok(())
-}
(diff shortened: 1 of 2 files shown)


Eikix:

No blocking issues found; both deletions were independently verified as coverage-neutral.

1. **tfhe-worker ignored `setup_test_key` runner** (`coprocessor/fhevm-engine/tfhe-worker/src/tests/mod.rs`): the deleted item was a manual DB-seeding entrypoint, not a test with assertions. The underlying helper `test_harness::db_utils::setup_test_key` remains and is still invoked on three active paths at the head ref: `test-harness/src/instance.rs:175/180` (the `DBInstance` setup used by cache tests), `tfhe-worker/src/tests/utils.rs:151` (the testcontainers `TestInstance` setup that every active tfhe-worker integration test goes through), and `tfhe-worker/benches/utils.rs:134`. A tree-wide grep at the head ref (code, docs, CI workflows, scripts, Makefiles) found no invocation of the deleted test by name and no `--ignored` runner referencing it — the only `--ignored` invocation in the tree is `solana/scripts/poc/adversarial-l4.sh`, which targets an unrelated kms test.

2. **Relayer ignored `test_input_proof_request_conversion_`** (`relayer/src/core/event.rs`): all four of its assertions (chain-id parse, `contract_address`, `user_address`, hex-decoded `ciphetext_with_zk_proof`) are carried verbatim into the now-strengthened active test `evm_input_proof_request_carries_evm_fields_and_leaves_solana_identities_unset` (event.rs:1213). The merged assertions were cross-checked against the `TryFrom<InputProofRequestJson>` impl (event.rs:1093): `CIPHERTEXT` is un-prefixed hex matching the conversion's `hex::decode`, and the 0x-prefixed address constants parse via `Address::from_str`, so the previously-ignored assertions run green as active code. Net effect is a coverage gain — the old active test only asserted `contract_address != Address::ZERO`; the merged one asserts exact values.

One non-blocking observation: neither the deleted test nor the merged one asserts `request.extra_data`, so the `EXTRA_DATA` constant is round-tripped but never checked on the EVM path. Pre-existing, not a regression from this PR.

No Solana-track surface (real-TFHE boundary, SNS mode, fixtures) is touched, matching the PR description.

PR #3113, opened 2026-07-11
feat(host-contracts): auto-resync bridge LZ delegate to ACL owner on acceptOwnership

jatZama:


host-contracts/contracts/ACL.sol
```diff
@@ -207,6 +208,19 @@ contract ACL is
...
+    /**
+     * @notice Accepts pending ownership and re-syncs the ConfidentialBridge's LayerZero
+     *         endpoint delegate to the new owner in the same transaction.
+     */
+    function acceptOwnership() public virtual override {
+        super.acceptOwnership();
+
+        // A bridge address is null if and only if there is no bridge on this chain, by deployment convention.
+        if (CONFIDENTIAL_BRIDGE_ADDRESS != address(0)) {
+            IConfidentialBridge(CONFIDENTIAL_BRIDGE_ADDRESS).syncDelegate();
```
PanGan21, line 220: Would you consider wrapping this in a try catch? 
I am thinking of the case where  `CONFIDENTIAL_BRIDGE_ADDRESS` is still an empty proxy, then `syncDelegate` fails and the whole `acceptOwnership` reverts for that reason.

jatZama: CONFIDENTIAL_BRIDGE_ADDRESS will never be an empty proxy, since when you upgrade ACL you also upgrade UUPSEmpty->ConfidentialBridge in a single atomic tx via the Aragon DAO... try/catch is not needed here. Even when bootstrping via an EOA deployer, there is no scenario when you would transfer ownership mid-first deployment.

PanGan21 (approved):

tawadaa (approved):

PR #3114, opened 2026-07-11
test(coprocessor): add real TFHE operator conformance

Eikix:
## What

Add 16 individually filterable, deterministic real-TFHE conformance tests at the production `tfhe_ops::perform_fhe_operation` seam.

The representative ordinary lane covers:

- Bool XOR and Not
- Uint8 encrypted addition and scalar subtraction wrapping
- Uint8 comparison returning Bool
- distinct Uint8 scalar shift and rotate behavior
- Uint64 multiplication, scalar division, scalar remainder, and unary wrapping
- Uint8 to Uint64 Cast with a big-endian type identifier
- IfThenElse, non-empty Sum, non-empty IsIn, and MulDiv

## Why

The fast catalog validates cleartext semantics, while the Solana Mollusk layer validates instruction admission and account behavior. This layer verifies that the real TFHE implementation produces the expected typed ciphertext and decrypted result without requiring listeners, databases, validators, Docker, or Solana.

## How

- Load the checked-in KMS-shaped `xof-keyset` and matching `xof-cks` fixtures once per test process.
- Install the real server key on every executing test thread.
- Encrypt every value operand with `FheTryEncrypt` and the matching client key.
- Call `perform_fhe_operation` directly.
- Assert the returned ciphertext variant and `type_num`, then decrypt and compare with test-owned constants.
- Serialize real-TFHE execution with a standard-library mutex. An unconstrained debug run left 12 tests executing after 80 seconds; serialization avoids CPU oversubscription while retaining ordinary named Rust tests.

## Feedback loop

Run the full target with the repository's optimized local profile:

```console
NO_DNA=1 cargo test --profile local -p fhevm-engine-common --test real_tfhe_conformance
```

Measured on the implementation machine:

- cold optimized build plus full target: 2m51s wall; the 16 tests themselves took 23.08s
- warm full target: 20.31s wall; the tests took 19.85s
- warm exact Uint8 addition: 3.82s wall; the test process took 3.48s, dominated by fixture loading

The unoptimized test profile is intentionally not the feedback lane: a sampled serialized run was stopped after 240.65s while executing Uint64 multiplication.

## Residual scope

- Rand and RandBounded need an independently specified fixed-seed known-answer contract. Type-only or range-only assertions would not prove seed encoding or randomness semantics, so they are not claimed here.
- Uint128, Uint160, and Uint256 remain outside the ordinary lane. No arbitrary timing threshold or ignored benchmark is introduced by this PR.
- This PR validates the CPU/default-feature `perform_fhe_operation` path. GPU execution and GPU conformance are not claimed and remain residual work.
- Full-stack reconstruction, database/event paths, and Solana admission remain separate layers.

Tracks zama-ai/fhevm-internal#1696 Phase 3.

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/tests/real_tfhe_conformance.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/tests/real_tfhe_conformance.rs
@@ -1,4 +1,7 @@
-use std::{path::PathBuf, sync::OnceLock};
+use std::{
+    path::PathBuf,
+    sync::{Mutex, MutexGuard, OnceLock},
+};
 
 use fhevm_engine_common::{
     tfhe_ops::perform_fhe_operation,
@@ -8,7 +11,7 @@ use fhevm_engine_common::{
 use tfhe::{
     prelude::{FheDecrypt, FheTryEncrypt},
     xof_key_set::CompressedXofKeySet,
-    ClientKey, FheBool, ServerKey,
+    ClientKey, FheBool, FheUint64, FheUint8, ServerKey,
 };
 
 struct TestKeys {
@@ -36,10 +39,14 @@ fn test_keys() -> &'static TestKeys {
     })
 }
 
-fn install_server_key() -> &'static TestKeys {
+fn install_server_key() -> (MutexGuard<'static, ()>, &'static TestKeys) {
+    static EXECUTION: Mutex<()> = Mutex::new(());
+    let execution = EXECUTION
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner());
     let keys = test_keys();
     tfhe::set_server_key(keys.server.clone());
-    keys
+    (execution, keys)
 }
 
 fn run(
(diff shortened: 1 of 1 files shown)


Eikix:

No blocking issues found; the single test is sound and genuinely exercises real TFHE. Two minor notes below.

Verified against `origin/elias/solana-real-tfhe-conformance`: the `xof-keyset` / `xof-cks` fixtures exist as LFS objects in `coprocessor/fhevm-engine/fhevm-keys/` and `coprocessor-cargo-tests.yml` does an LFS checkout and runs `fhevm-engine-common` tests, so this will actually execute in CI; the `perform_fhe_operation` call matches both the CPU and GPU signatures (the `0` third arg is the ignored/`gpu_idx` slot); `FheBool` XOR is handled in `tfhe_ops.rs` (~line 1421); and `type_num() == 0` is the correct FheBool discriminant per `types.rs:466`. The assertion chain (variant + `type_num` + real decrypt against test-owned constants) is honest TFHE evidence, not cleartext conformance re-labelled — consistent with the fhevm-internal#1696 Phase 3 framing.

1. `real_tfhe_conformance.rs:20-36` — the fixture-loading block (read `xof-keyset`, `safe_deserialize_key`, `decompress().into_raw_parts()`, read `xof-cks`) is now the third copy of this pattern, after `sns-worker/src/tests/mod.rs:610-621` and `test-harness/src/db_utils.rs:226-227`. A shared helper is probably blocked by dependency direction (test-harness depends on engine-common), so this is acceptable — but worth a pointer comment or consolidation if a fourth copy appears, since all copies must rotate together with the fixture (the sns-worker doc comment already warns rotations invalidate derived fixtures).

2. `real_tfhe_conformance.rs:44-46` — `install_server_key` clones the full `ServerKey` out of the `OnceLock` on every call because `set_server_key` takes ownership and the key is stored behind a shared reference. With one test this is invisible; once the announced Bool/Uint8/Uint64 families land, each `#[test]` thread pays a full server-key clone. If that becomes noticeable in CI wall time, storing per-thread via `thread_local` + `set_server_key` once per thread (or gating on whether the key is already set) avoids it. Not a correctness issue.

Since the PR is a draft with the remaining operator families still to come, no opinion is offered on breadth yet — the scaffolding (per-process key init, typed `run` helper panicking with the op name) looks like the right seam for extending.

PR #3121, opened 2026-07-13
chore(kms-connector): better docs and logs for raw private key

eudelins-zama:
### Summary

The `TransactionSender` supports a raw hex private key for local/testing setups. This PR wraps it in a small `PrivateKey` newtype so it renders as `PrivateKey(<redacted>)` in debug logs instead of the raw string, and makes the "testing only, use AWS KMS in production" guidance explicit in the config docs.

###  Changes

- New `PrivateKey` newtype (`utils/config/wallet.rs`): `#[serde(transparent)]` over `String`, so the config format is unchanged; custom `Debug` redacts the value. Used for `TransactionSender`'s `private_key` field.
- Docs: field doc and `tx-sender.toml` now note that raw keys are for testing and `aws_kms_config` is the production path.
- Cleanup: gated `Serialize/serialize_pg_interval` imports behind `#[cfg(debug_assertions)]` (they're only used by the debug-only derive), removing release-build warnings in the `tx-sender`, `gw-listener`, and  `kms-worker` configs.

### DIFF
--- a/kms-connector/config/tx-sender.toml
+++ b/kms-connector/config/tx-sender.toml
@@ -42,6 +42,8 @@ database_url = "postgres://postgres:postgres@localhost/kms-connector"
 # Private key as a hex string (optional if `aws_kms_config` is configured)
 # If provided, the connector will use this private key
 # Format: hex string with or without 0x prefix
+# WARNING: Using a raw private key is intended for TESTING PURPOSES ONLY.
+#          Production deployments MUST use `aws_kms_config` instead (see below).
 # IMPORTANT: Keep this secure and never share it!
 # ENV: KMS_CONNECTOR_PRIVATE_KEY
 private_key = "0x3f45b129a7fd099146e9fe63851a71646231f7743c712695f3b2d2bf0e41c774"
--- a/kms-connector/crates/utils/src/config/wallet.rs
+++ b/kms-connector/crates/utils/src/config/wallet.rs
@@ -23,6 +23,44 @@ pub enum KmsWallet {
     AwsKms(AwsSigner),
 }
 
+/// A wallet private key, provided as a hex string (with or without `0x` prefix).
+///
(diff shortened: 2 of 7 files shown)


kms-connector/config/tx-sender.toml
```diff
@@ -42,6 +42,8 @@ database_url = "postgres://postgres:postgres@localhost/kms-connector"
 # Private key as a hex string (optional if `aws_kms_config` is configured)
 # If provided, the connector will use this private key
 # Format: hex string with or without 0x prefix
+# WARNING: Using a raw private key is intended for TESTING PURPOSES ONLY.
```
piizama, line 45: Is it possible to also print a warning log when the connector starts ?

eudelins-zama: Sure, good idea!

eudelins-zama: Done 46c56d396580ca9436f9a1b6e5c54b832909f80a

dd23 (approved):

kms-connector/crates/utils/src/config/wallet.rs
```diff
@@ -23,6 +23,44 @@ pub enum KmsWallet {
...
 
+/// A wallet private key, provided as a hex string (with or without `0x` prefix).
+///
+/// # Testing only
+///
+/// Configuring a raw private key is intended for **testing purposes only**.
+/// Production deployments should use an AWS KMS signer instead (see [`AwsKmsConfig`]).
+///
+/// The inner secret is kept private and the [`Debug`] implementation redacts it, so the key
+/// is never leaked through debug logs.
```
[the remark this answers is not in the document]
eudelins-zama: Nitpicking

eudelins-zama:
@mergifyio queue

PR #3122, opened 2026-07-13
fix(kms-connector): use confirmed Solana decrypt state

Eikix:
Switches Solana KMS ACL reads and relayer proof RPC reads from finalized to explicit confirmed commitment, matching the accepted optimistic authorization semantics and existing EVM-style behavior. Preserves signature, owner, canonical PDA, domain, subject, and MMR verification while treating equal-count confirmed-view proof disagreement as a bounded recoverable failure. Persistent proof-cache divergence remains fail-closed with the existing HTTP 500 contract and requires cache rebuild. This is the preparatory base for a stacked coprocessor cleanup; end-to-end material availability remains finalized until that follow-up removes the separate finalized-account gate.

### DIFF
--- a/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
@@ -139,6 +139,8 @@ where
 
             match self.host_chain_backend(ct_chain_id)? {
                 HostChainAclBackend::Solana(host) => {
+                    // Public access is proven by a PublicDecryptLeaf MMR proof and verified
+                    // against the live confirmed lineage account.
                     check_solana_handles_public_decrypt(host, &[ct.ctHandle.0], extra_data)
                         .await
                         .map_err(|e| {
@@ -491,7 +493,7 @@ where
     /// 2. ed25519 signature over the canonical preimage — binds `publicKey`, handles, identity,
     ///    nonce, allowed domains, and validity window to the claimed Solana identity (closes the
     ///    publicKey-substitution / relayer-bypass bug),
-    /// 3. per-handle ACL read at `finalized` commitment with owner + canonical-PDA checks and the
+    /// 3. per-handle ACL read at `confirmed` commitment with owner + canonical-PDA checks and the
     ///    domain-scoped verifier (identity as subject).
     #[tracing::instrument(skip_all)]
     pub async fn check_user_decryption_request_solana(
@@ -541,7 +543,7 @@ where
         let auth = verify_solana_user_decrypt_signature(request, chain_id)
             .map_err(|e| RequestCheckError::from_processing(RequestCheckKind::Signature, e))?;
 
-        // ACL phase: read each handle's record at finalized commitment and run the domain-scoped
+        // ACL phase: read each handle's record at confirmed commitment and run the domain-scoped
         // verifier with the identity as subject.
         let handles: Vec<HandleBytes> = request.handles.iter().map(|e| e.handle.0).collect();
         check_solana_handles_acl(host, &handles, &auth)
--- a/kms-connector/crates/kms-worker/src/core/event_processor/solana_user_decrypt.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/solana_user_decrypt.rs
@@ -36,11 +36,11 @@
 //! authorizes exactly one handle, so multi-handle Solana requests are out of scope for this
 //! rewrite):
 //!
-//! - **current-handle** (`mmr_proof_bytes` empty): the lineage account, fetched at `finalized`
+//! - **current-handle** (`mmr_proof_bytes` empty): the lineage account, fetched at `confirmed`
 //!   commitment, is checked for canonical PDA derivation + program ownership + `current_handle ==
 //!   handle` + `subject ∈ subjects` ([`dispatch_solana_current`]).
 //! - **historical / public** (`mmr_proof_bytes` non-empty): the request carries an MMR inclusion
-//!   proof, verified against the LIVE finalized peaks (the account is always freshly fetched, never
+//!   proof, verified against the live confirmed peaks (the account is always freshly fetched, never
 //!   a cached/proof-time snapshot) via the shared crate's `authorize_historical` /
 //!   `authorize_public` ([`dispatch_solana_mmr_proof`]). There is no live "is public" flag any
 //!   more — public-ness is only provable via a `PublicDecryptLeaf` MMR leaf.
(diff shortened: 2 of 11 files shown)


Eikix:
## Review: confirmed Solana decrypt state

### What it does

Two coupled semantic changes, plus doc alignment:

1. **Commitment downgrade `finalized` → `confirmed`** for every Solana authorization read: the KMS worker's ACL account fetcher (`solana_v2_fetcher.rs`) and all three relayer proof-service RPCs (`getSignaturesForAddress`, `getTransaction`, `getAccountInfo` in `relayer/src/solana_proof/chain.rs`).
2. **Failure reclassification** in `classify_mmr_verification_failure`: an invalid inclusion proof with `proof_leaf_count == live_leaf_count` was terminal; it's now recoverable (folded into the `>=` branch), on the theory that at `confirmed` the KMS and proof service can legitimately observe different forks with equal counts.

### What's good

- **Mechanically clean and complete.** The rename `SOLANA_COMMITMENT_FINALIZED` → `SOLANA_COMMITMENT_CONFIRMED` is compile-enforced; every remaining `finalized` in the touched crates is Ethereum-side or the intentionally-preserved coprocessor gate. Doc comments, module docs, and the four design docs are updated in lockstep — no stale "finalized" claims left in the Solana auth path.
- **`getSignaturesForAddress` previously sent *no* commitment** (RPC default = finalized) while the other two pinned it explicitly. Extracting `signatures_params`/`transaction_params`/`account_info_params` with a single `SOLANA_PROOF_COMMITMENT` constant fixes that implicit inconsistency and makes the pinning testable — the new `proof_rpc_params_pin_confirmed_commitment` test is exactly the right shape.
- **Branch logic simplification is correct**: `>=` and `<` are exhaustive over the `proof_invalid` case, so deleting the third branch loses nothing.
- Test updates faithfully track the new semantics rather than just silencing failures, and the DD-025 status update honestly says "partially resolved" rather than overclaiming.

### Concerns

**1. This resolves a flagged-open security decision — make sure the sign-off exists.** The deleted comment said *"Authorizing decryption against anything weaker than `finalized` would let a rolled-back slot grant access"*, and DD-025 was explicitly marked "BIG OPEN debate item" with the note that a confirmed-then-reorged release is an **irreversible decrypt**. The PR body says "matching the accepted optimistic authorization semantics" — if that acceptance lives in an RFC or a recorded decision, DD-025 should cite it. On the merits it's defensible (optimistically-confirmed rollback on Solana is essentially a slashing-level event, and most of the ecosystem runs at `confirmed`), but a doc that previously said "open for debate" now says "accepted policy" with no pointer to who accepted it or where.

**2. Equal-count reclassification weakens fail-fast for genuinely bad proofs.** The two updated tests show the trade-off directly: a `PublicDecryptLeaf` for a *rotated-away* handle, or a historical leaf misused for public decrypt, previously failed terminally and now burns the full bounded retry budget before failing. These are exactly the shapes a buggy or malicious proof service would produce. It's still fail-closed and bounded, so it's not a security hole, but it:
   - adds latency and retry noise to every deterministic-bad-proof failure, and
   - makes real proof-service bugs look like transient fork disagreement in logs.

   Worth considering: at equal counts, a *fork* disagreement should resolve within a slot or two, while a wrong-leaf proof never will. If the retry budget is large/long, a tighter budget or a distinct log tag for the equal-count case would keep operability. At minimum it would help to state what the "ordinary bounded attempt budget" actually is in wall-clock terms.

**3. Proof-cache poisoning becomes possible (was impossible).** Ingesting `getSignaturesForAddress`/`getTransaction` at `confirmed` means a (rare) confirmed-rollback can bake orphaned transactions into the append-only leaf cache. The PR body acknowledges this — divergence surfaces as the existing fail-closed HTTP 500 and needs a manual cache rebuild. That converts a previously-impossible failure into a rare operational incident with a manual runbook step. Fine as a trade-off, but it should be written down where operators will find it (TESTING_STRATEGY / a runbook), not only in the PR description.

**4. Split-brain interim state.** Until the stacked coprocessor cleanup lands, authorization is at `confirmed` while material availability is gated at `finalized`. That's conservative (the slower gate still binds end-to-end), so no bug — but it means this PR's latency benefit doesn't materialize yet, which the PR body honestly states. Just confirming the ordering: this is safe to merge *before* the coprocessor change, not after.

### Minor

- `solana_v2_fetcher.rs` test asserts `params[1]["commitment"]` against both the constant *and* the literal `"confirmed"` — preexisting pattern, and actually useful (catches someone editing the constant), so fine.
- The recoverable error message "while confirmed views converge" is good; consider including the account pubkey for debuggability (the stale/terminal branch context has it available upstream).

### Verdict

Solid, well-scoped, internally consistent PR — the code changes are correct given the policy decision. The main ask before merge is **provenance for the policy change** (concern 1) and a decision on whether equal-count bad proofs deserve a tighter retry budget or distinct logging (concern 2). Concerns 3–4 are acknowledged trade-offs that just need operator-facing documentation.

Eikix:
## Re-review after `98b80e72b` (clarify confirmed proof retries)

Two new commits since the previous review: a pure rustfmt/import-ordering commit (`f49a5aedd`, no semantic change) and the substantive fix (`98b80e72b`). The fix resolves three of the four concerns; the fourth was already fine as an acknowledged trade-off.

### How each concern was addressed

**1. Policy provenance — resolved.** DD-025 now records the decision explicitly: accepted by the Solana feature owner during this PR's review on 2026-07-13, with the trade-off spelled out in one honest sentence ("irreversible plaintext release after a valid authorization observed on an exceptionally rolled-back confirmed fork"). That's exactly the citation asked for. It's slightly self-referential (the doc cites the PR that changes the doc), but the decision and its owner are now durable in-repo, which is what matters.

**2. Equal-count vs proof-ahead distinguishability — resolved, via the logging option.** The single `>=` branch is split back into two recoverable branches with distinct greppable tags: `classification=confirmed_equal_count` (fork disagreement or deterministically bad proof) and `classification=confirmed_proof_ahead` (ordinary catch-up). Both tests now assert the tags, so the distinction can't silently regress. The retry budget stays shared rather than being tightened for the equal-count case, but the new TESTING_STRATEGY section quantifies it — 20 attempts at a 3-second fast polling interval, so a deterministic bad proof exhausts in roughly a minute plus load — which answers the wall-clock question and makes the trade-off a documented, informed one. Both defaults were verified against `config.rs` and `kms-worker.toml`; the docs are accurate.

**3. Cache-rebuild runbook — resolved.** TESTING_STRATEGY's new "Confirmed-view operations" section gives operators the concrete recovery procedure for a persistent `corrupt_cache`: stop the relayer, remove the `solana_proof.leaf_store_path` JSON file, restart to replay from `start_signature`, with the correct framing that requests fail closed during replay and the cache is never an authorization anchor.

**4. Split-brain interim state** — unchanged, and was already correctly handled (merge-order safe, honestly documented).

### Remaining nits (non-blocking)

- The `classification=` tag lives inside an `anyhow` message string rather than a structured tracing field, so it's greppable but not usable as a metric label. Fine for now; worth structuring if anyone wants alerting on `confirmed_equal_count` frequency.
- The earlier suggestion to include the account pubkey in the recoverable error message wasn't taken — minor, the upstream context usually carries it.

### Verdict

All substantive review feedback is addressed, the new documentation matches the actual code defaults, and the branch split with asserted classification tags is a strictly better shape than the folded `>=` branch. Good to merge from this review's side.

PR #3123, opened 2026-07-13
feat(coprocessor): add --seed-start-block flag to host-listener poller

Eikix:
## Problem

When the poller starts against a fresh database (new chain onboarding), it has no persisted anchor and falls back to `max(host_chain_blocks_valid)` — a table owned by the WS listener. This cross-service read is a startup race: if the poller initializes before the WS listener has written anything, it anchors at `0` and walks the chain from genesis. Observed on Amoy pollers on devnet (3 of 5 stacks ~2-3 months behind head) and on testnet (Zama + partners).

Context: [Slack thread](https://zama-ai.slack.com/archives/C065F16D5U3/p1783081767401399)

## Fix

Drop the cross-service read entirely and add an `--seed-start-block` flag. The poller's seed sources are now its own, in one linear precedence:

1. **Persisted anchor row** (`host_listener_poller_state`) — always wins; the flag is inert once it exists
2. **`--seed-start-block`** — `>= 0`: absolute height; negative: that many blocks behind head at first startup
3. **Neither → explicit error** — a missing flag on a new chain surfaces as a crash-loop at deploy time instead of a silent weeks-long genesis walk; `--seed-start-block=0` opts into genesis deliberately

### Deliberately NOT the listener's `--start-at-block` semantics

The listener's flag beats persisted state on every startup. Transplanting that onto the poller would mean: every pod restart rewinds the anchor and re-walks; a flag above the anchor silently skips blocks; and it would clobber drift-revert's anchor reset. Hence a distinct name and initialization-only semantics — safe to leave in Helm values permanently.

## Behavior changes when the flag is unset

- Flagless fresh deployment: explicit startup error, instead of a scheduling-dependent race. The e2e compose poller passes `--seed-start-block=0` (fresh local chain: genesis is the honest value).
- Adding a poller to an already-running listener-only deployment now requires `--seed-start-block` explicitly (the removed fallback previously covered that case).
- Existing deployments with an anchor row: untouched.

## Notes for operators

- Already-lagging pollers keep their persisted anchor and still need the manual patch:
  ```sql
  UPDATE host_listener_poller_state SET last_caught_up_block = <target>, updated_at = NOW() WHERE chain_id = 80002;
  ```
- Wiring the flag into helm values (`coprocessor-operator`) is a separate change.

## Testing

- Anchor seeding extracted into a pure `resolve_start_anchor` helper with unit tests (`cargo test -p host-listener --lib start_anchor`).

## Backports

Same commit cherry-picked to `release/0.13.x` (#3124) and `release/0.14.x` (#3125), identical patch-id `e8d7dea5d57db1272a619d5bb1e1863b111cf234`.

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
@@ -52,6 +52,14 @@ struct Args {
 
     #[arg(
         long,
+        default_value = None,
+        help = "Initial sync block used only when no poller state exists yet; an existing anchor always wins (can be negative from last block)",
+        allow_hyphen_values = true
+    )]
+    seed_start_block: Option<i64>,
+
+    #[arg(
+        long,
         default_value_t = 15,
         help = "Depth behind the head considered final (in blocks)"
     )]
@@ -220,6 +228,7 @@ async fn main() -> anyhow::Result<()> {
         max_http_retries: args.max_http_retries,
         rpc_compute_units_per_second: args.rpc_compute_units_per_second,
         health_port: args.health_port,
+        seed_start_block: args.seed_start_block,
         dependence_cache_size: args.dependence_cache_size,
         dependence_by_connexity: args.dependence_by_connexity,
         dependence_cross_block: args.dependence_cross_block,
--- a/coprocessor/fhevm-engine/host-listener/src/poller/mod.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/poller/mod.rs
@@ -71,6 +71,58 @@ fn handle_rpc_failure<E: std::fmt::Display>(
     }
 }
 
+/// Seed for `host_listener_poller_state` when no anchor row exists yet; an
+/// existing row always wins in the caller, so restarts never rewind or skip.
+/// A non-negative value is an absolute height (0 = genesis, explicitly), a
+/// negative value means that many blocks behind the current head. Unset is an
+/// error: the first start for a chain must state where to begin.
+#[derive(Debug, PartialEq, Eq)]
+enum StartAnchor {
+    Block(i64),
+    BehindHead(u64),
+}
+
+fn resolve_start_anchor(seed_start_block: Option<i64>) -> Result<StartAnchor> {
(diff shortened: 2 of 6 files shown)


coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
```diff
@@ -50,6 +50,14 @@ struct Args {
     #[arg(long, help = "PostgreSQL connection URL")]
     database_url: DatabaseURL,
 
+    #[arg(
+        long,
+        default_value = None,
+        help = "Initial sync block used only when no poller state exists yet; an existing anchor always wins (can be negative from last block)",
+        allow_hyphen_values = true
+    )]
+    initial_start_block: Option<i64>,
```
PanGan21: Do we need chart version bump?

coprocessor/fhevm-engine/host-listener/src/poller/mod.rs
```diff
@@ -71,6 +71,48 @@ fn handle_rpc_failure<E: std::fmt::Display>(
...
+enum StartAnchor {
+    Block(i64),
+    BehindHead(u64),
+}
+
+fn resolve_start_anchor(seed_start_block: Option<i64>) -> StartAnchor {
+    match seed_start_block {
+        Some(block) if block >= 0 => StartAnchor::Block(block),
+        Some(delta) => StartAnchor::BehindHead(delta.unsigned_abs()),
+        None => StartAnchor::Block(0),
```
rudy-6-4: should we make it an explicit error, so this never happens unless in test mode ?

Eikix: you mean error in  None => StartAnchor::Block(0),?

Eikix: ok lets do that.

PanGan21 (approved):

antoniupop (approved):

Eikix:
@mergifyio queue

PanGan21 (approved):

PR #3130, opened 2026-07-13
fix(coprocessor): take shared cutover lock for GCS writers in cutover gate

goshawk-3:
GCS-mode (green) writers previously short-circuited cutover_gate and skipped the shared advisory lock entirely. Their writes land in the gcs schema, which execute_cutover merges into the live tables and then drops.  

Move the shared-lock acquisition ahead of the gcs_mode check so green writers join the cutover barrier.


PanGan21 (approved):

goshawk-3:
@mergifyio queue

PR #3131, opened 2026-07-13
refactor(coprocessor): simplify Solana confirmed ingestion

Eikix:
## Why

The Solana coprocessor should prepare ciphertext material promptly from confirmed instruction data. Preparing material does not authorize plaintext release: KMS remains the boundary that validates the live `EncryptedValue` account or MMR proof.

This is stacked on #3122, which moves KMS and relayer proof reads to explicit `confirmed` commitment.

## What changed

- make confirmed Yellowstone instruction reconstruction the only supported Solana listener flow
- insert concrete handle material requests directly into `pbs_computations`
- remove the finalized-account queue, fetcher, witness store, retry state machine, and Docker/setup wiring
- remove cross-transaction encrypted-value lineage state
- remove the unused RPC listener and confidential-token witness event codegen
- retain the previously applied SQL migration and add a forward migration dropping the dead tables
- update the design documents to describe confirmed ingestion and live KMS authorization

The branch removes roughly 4,000 more lines than it adds.

## Validation

- repository formatting and diff checks
- workspace `cargo check`
- CI-aligned Clippy
- all-targets host-listener check with `solana-grpc,solana-reconstruct`
- 61 focused Solana host-listener tests
- 15 encrypted-value instruction/reconstruction tests
- 7 Yellowstone ACL/reconstruction tests
- adversarial review of the final stacked diff: no actionable findings

## Accepted tradeoff

An exceptional rollback of a confirmed fork can leave orphaned append-only computation/material rows and waste work. Those rows cannot authorize a decrypt: KMS re-reads confirmed live ACL state or MMR peaks before releasing plaintext. Reorg cleanup can be added later for resource recovery without becoming an authorization dependency.


Eikix:
## Review (agent-assisted)

### Overview

This PR collapses the Solana ingestion pipeline from a two-stage design (confirmed listener → queued finalized-account fetcher → witness store → pbs rows) into a single stage: confirmed Yellowstone instruction reconstruction inserts `pbs_computations` material requests directly. It deletes the RPC-polling listener, the finalized-account fetcher and its retry state machine, the `EncryptedValueLineageTracker`, and the confidential-token event codegen. The trust-model shift (KMS revalidates live ACL/MMR state at release, so ingestion no longer needs finalized data) is clearly argued and mirrored in the docs.

The deletions are coherent, tests were rewritten rather than deleted wholesale, the migration is forward-only with the original create-migration retained, and `build_subscribe_request` keeps `failed: Some(false)` + `vote: Some(false)`, so only successful confirmed transactions can schedule material — the on-chain program's own checks still gate ingestion.

### Findings

**1. Undecodable covered instructions are now dropped silently (main concern).**
Coverage in `reconstruct_events_for_insert` is decided by `decode_fhe_eval_args(...).is_some()` / `decode_encrypted_value_instruction(...).is_some()`. An instruction with a known discriminator but args that fail Borsh decode (schema drift, future field, malformed payload) is treated as "not covered" → `NotCovered` → `Vec::new()` → silent success. Two regressions vs. the old code:
- Emit-decode used to be the fallback for uncovered transactions, so events still arrived; that safety net is gone by design, but the old `warn!` on fallback is gone too. There is now **no log line at all** for a zama-host transaction that produced zero events.
- The test `make_handle_public_malformed_args_fail_closed` is misnamed: at ingest level a malformed `make_handle_public` doesn't fail closed, it is silently skipped (previously it queued a handle-less fetch resolved from the account). At minimum, a `warn!` when a zama-host instruction carries a known discriminator but undecodable args would make schema drift observable instead of manifesting as "decrypts hang with no logs".

**2. Inconsistent failure posture between the two covered families.**
`fhe_eval` fails closed (`anyhow::bail!` on missing derivation context, unresolvable durable account, malformed plan), but the lifecycle path silently skips when `accounts.get(encrypted_value_index)` is out of range (`accounts.get(...)?` in `decode_encrypted_value_instructions`). A truncated account list on a successful lifecycle instruction should arguably be as loud as the fhe_eval equivalent. Pre-existing shape, but this PR removes the downstream fetcher that would have surfaced the gap.

**3. Docker image no longer ships any Solana binary.**
The Dockerfile drops both binaries, and `solana_host_listener` now has `required-features = ["solana-grpc", "solana-reconstruct"]`, so default builds don't produce it. If intentional (PoC runs from source via `setup-solana-side.sh`), fine — but worth stating in the PR body, since "the only supported Solana listener flow" currently has no deployable image.

**4. Minor.**
- `insert_solana_events` calls `insert_pbs_computations` once per request with a single-element `Vec`; the API takes a slice of handles, so all requests for the transaction could go in one call.
- In `bin/solana_host_listener.rs`, `chain_id` (a `ChainId` for `Database::new`) is later shadowed by the `u64` HostConfig chain id — correct but confusing; distinct names would help. Two "non-obvious why" comments were also deleted (the `from_canonical_u64` RFC-021 rationale and the OTLP no-collector fallback rationale); both still apply to the surviving code and are worth restoring.

### Correctness / tests / security

- Supersession semantics preserved: `Update`/durable-supersede request material for both new and previous handles, pinned by `superseding_fhe_eval_derives_output_handle_without_hint` asserting exactly `[current, previous]`.
- `allow_subjects`/`remove_subject` correctly emit nothing now (material exists from creation; KMS reads live ACL) — tests updated to assert emptiness rather than deleted.
- The reorg tradeoff (orphaned append-only rows from a confirmed-fork rollback) is explicitly accepted and is sound given KMS revalidation; the docs updates encode that invariant.

### Verdict

Solid, well-documented simplification with the right trust-model reasoning. **#1** deserves addressing before merge (an observable signal for covered-but-undecodable instructions); #2 and #3 are judgment calls to confirm as intentional; #4 is cosmetic.

Eikix:
## Re-review after `3f4da867e` (agent-assisted)

### Prior findings — all addressed

- **#1 (silent drop of undecodable instructions) — fixed.** New `is_fhe_eval_instruction` / `is_encrypted_value_instruction` discriminator predicates; any zama-host instruction with a known discriminator but undecodable args now `bail!`s instead of being silently skipped. Covered by the new `known_but_undecodable_instructions_fail_ingest` test.
- **#2 (inconsistent failure posture) — fixed.** A lifecycle instruction whose `encrypted_value` account index is out of range now bails with slot and account count, matching fhe_eval's fail-closed behavior. Test added.
- **#3 (no Docker binary) — documented.** DESIGN_DECISIONS.md now states the shared container is intentionally EVM-only and a production Solana image is future packaging work, not an implicit fallback.
- **#4 (naming/comments) — fixed.** `database_chain_id` / `host_config_chain_id` replace the shadowing, and both deleted rationale comments (RFC-021 bit pattern, OTLP no-collector stall) are restored.

The pre-check loop is well-placed before the `NotCovered` early-return, and the later `ctx.expect(...)` is safe: the pre-check guarantees `is_fhe_eval ⇒ decodable`, and the `has_fhe_eval && ctx.is_none()` bail runs first.

### One new concern introduced by the fix

The new bails are wrapped in `IngestFailure::fatal` → `FatalSameCursor`: the listener **stops entirely without advancing the cursor**, and after restart it re-reads the same transaction and halts again — a permanent ingestion freeze until a human intervenes.

That severity is right for genuine schema drift. But the decoders use Borsh `try_from_slice`, which rejects trailing bytes, while Anchor's generated instruction dispatch deserializes args with `deserialize`, which tolerates trailing bytes (worth double-checking against the Anchor version zama-host pins). If the mismatch holds, a transaction that **succeeds on-chain** — valid args plus appended garbage bytes — fails the listener's decode and permanently halts all Solana ingestion. Anyone who can successfully invoke `fhe_eval` or a lifecycle instruction could trigger it. Before this fix the same mismatch caused a silent miss (arguably worse, since undetectable); now it is a loud halt (detectable, but a DoS lever).

**Suggested resolution:** make the listener's arg decode exactly as lenient as the program's — decode via `AnchorDeserialize::deserialize` on the payload, ignoring trailing bytes — so "known discriminator + undecodable" can only mean real drift, and keep the fatal halt for that case. Small change to `decode_fhe_eval_args` and `decode_encrypted_value_instruction`.

Eikix:
## Re-review at head `58683ce6a` (agent-assisted)

### `fix(coprocessor): match Anchor instruction decoding`

Resolves the decode-leniency finding from the previous round, cleanly:

- All six decoders (`fhe_eval` plus the five `EncryptedValue` lifecycle variants) now route through one helper, `decode_anchor_args<T: AnchorDeserialize>`, using `T::deserialize(&mut &payload[..])` — the same trailing-byte-tolerant semantics as Anchor's generated instruction dispatch.
- **Fail-closed is preserved where it matters:** a truncated or genuinely malformed payload still returns `None` → the pre-check still `bail!`s → fatal halt. "Known discriminator + undecodable" now can only mean real schema drift between listener and program, which is exactly when a loud halt is right. The attacker lever (valid args + appended garbage → successful on-chain tx the listener can't decode) is closed.
- **Tests pin both sides of the contract:** `lifecycle_decode_matches_anchor_trailing_byte_semantics` and the extended fhe_eval round-trip assert trailing bytes decode; `make_handle_public_malformed_args_fail_closed` still asserts truncated args fail.

No findings on this commit.

### Overall

All findings from both prior review rounds are addressed: silent drops → fail-closed with tests, inconsistent failure posture → unified, Docker packaging → documented as intentional in DESIGN_DECISIONS.md, naming/comment nits → fixed, Anchor decode-leniency mismatch → fixed. The only untaken suggestion is the cosmetic `insert_pbs_computations` batching nit. Looks good to merge once CI is green.

PR #3132, opened 2026-07-13
refactor(coprocessor): drop poller seed fallback on WS-listener blocks

Eikix:
Stacked on #3123 — retarget to `main` once it merges. **Main only, deliberately not backported**: this is a behavior change for flagless fresh deployments and does not belong in a patch release (discussed with Rudy).

The poller's first-seed fallback on `host_chain_blocks_valid` (a table owned by the WS listener) was a bootstrap aid from when the poller was introduced into already-running systems. That migration is complete — every live deployment has its own `host_listener_poller_state` row — so the only code path still reaching the fallback is a fresh database, where the cross-service read *is* the startup race.

After this change the poller's seed sources are entirely its own:
1. persisted anchor row (always wins)
2. `--initial-start-block`
3. genesis, with a `warn!`

Consequences:
- flagless fresh deployment: guaranteed genesis walk with a warning, instead of a scheduling-dependent race (deterministic and loud beats silent and lucky)
- adding a poller to an already-running listener-only deployment now requires `--initial-start-block` explicitly
- `resolve_start_anchor` collapses to a single-argument pure function

Tested: `cargo test -p host-listener --lib start_anchor`, clippy clean.

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/bin/poller.rs
@@ -52,6 +52,14 @@ struct Args {
 
     #[arg(
         long,
+        default_value = None,
+        help = "Initial sync block used only when no poller state exists yet; an existing anchor always wins (can be negative from last block)",
+        allow_hyphen_values = true
+    )]
+    initial_start_block: Option<i64>,
+
+    #[arg(
+        long,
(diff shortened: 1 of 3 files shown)


Eikix:
Folded into #3123/#3124/#3125 — decision was to keep all three PRs identical rather than split the fallback removal to main only.

PR #3135, opened 2026-07-13
fix(kms-connector): harden legacy user-decryption calldata sourcing

eudelins-zama:
### Summary

Hardens how the KMS connector sources `calldata` for the legacy user-decryption path.

The ACL check for this path relies on the `ctHandleContractPairs` decoded from the request transaction's `calldata`. We now ensure that this `calldata` is only used when the transaction was sent directly to the `Decryption` contract, so it maps unambiguously to the user-decryption event  we're processing.

### Changes

- `fetch_calldata` now fetches the transaction and verifies that `tx.to()` equals the configured `Decryption` contract address before returning its `calldata`. Transactions sent elsewhere (or contract-creation transactions) are rejected as irrecoverable.
- Added unit coverage as a single parametrized test covering the three cases: a tx sent to the `Decryption` contract (accepted), a transaction sent to an intermediary contract (rejected), and a contract-creation transaction (rejected).

Also include a bump of `crossbeam-epoch` crate

### DIFF
--- a/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
@@ -661,19 +661,35 @@ where
         Ok(sns_ciphertext_materials)
     }
 
+    /// Fetches the calldata of a given transaction.
+    ///
+    /// Only allows transactions sent directly to the `Decryption` contract.
     pub async fn fetch_calldata(
         &self,
         tx_hash: FixedBytes<32>,
     ) -> Result<Vec<u8>, ProcessingError> {
-        self.decryption_contract
+        let decryption_address = *self.decryption_contract.address();
+
+        let tx = self
+            .decryption_contract
             .provider()
             .get_transaction_by_hash(tx_hash)
             .await
             .map_err(|e| ProcessingError::Recoverable(anyhow::Error::from(e)))?
             .ok_or_else(|| {
                 ProcessingError::Irrecoverable(anyhow!("No transaction found with hash {tx_hash}!"))
-            })
-            .map(|tx| tx.input().to_vec())
+            })?;
+
+        if tx.to() != Some(decryption_address) {
+            return Err(ProcessingError::Irrecoverable(anyhow!(
+                "Transaction {tx_hash} was sent to {:?} rather than directly to the Decryption \
+                contract {decryption_address}: its calldata cannot be associated with the user \
+                decryption event.",
+                tx.to(),
+            )));
+        }
+
+        Ok(tx.input().to_vec())
     }
 }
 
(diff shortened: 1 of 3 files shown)


kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
```diff
@@ -1364,4 +1389,96 @@ mod tests {
             .await
             .unwrap();
     }
+
+    // -------------------------------------------------------------------------
+    // `fetch_calldata` hardening: the legacy path recovers the ACL data from the transaction
+    // calldata, so the calldata is accepted only when the transaction targeted the Decryption
+    // contract directly.
+    // -------------------------------------------------------------------------
+
```
dvdplm: I'd remove this.

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
```diff
@@ -1364,4 +1389,96 @@ mod tests {
...
+        asserter.push_success(&mock_legacy_request_tx(tx_hash, to, &calldata));
+
+        let result = processor.fetch_calldata(tx_hash).await;
+        if should_succeed {
+            assert_eq!(result.unwrap(), calldata);
+        } else {
+            match result {
+                Err(ProcessingError::Irrecoverable(_)) => (),
+                other => panic!("Expected Irrecoverable error, got: {other:?}"),
+            }
```
dvdplm, line 1437: I guess you could use `assert!(matches!(… …` here but I do not mind what you did, I think it's pretty readable.

kms-connector/crates/utils/src/config/wallet.rs
```diff
@@ -23,6 +23,44 @@ pub enum KmsWallet {
...
+///
+/// Configuring a raw private key is intended for **testing purposes only**.
+/// Production deployments should use an AWS KMS signer instead (see [`AwsKmsConfig`]).
+///
+/// The inner secret is kept private and the [`Debug`] implementation redacts it, so the key
+/// is never leaked through debug logs.
+#[derive(Clone, Deserialize, PartialEq)]
+#[cfg_attr(debug_assertions, derive(Serialize))]
+#[serde(transparent)]
+pub struct PrivateKey(String);
```
dvdplm: Can we name this `TestingPrivateKey`?

eudelins-zama: Dropped the cherry-pick

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
eudelins-zama: Done 231c8f22580e04798bb3a6f128273b35f265a4a6

kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
eudelins-zama: Keeping it this way, but indeed

dvdplm (approved):
LGTM

PR #3136, opened 2026-07-13
feat(coprocessor): emit produced-public output batches

Eikix:
Part of zama-ai/fhevm-internal#1675.

## Summary

- Emit one versioned `PublicOutputsProducedEvent` after a successful `fhe_eval` frame when it produced public durable outputs.
- Record only each ordered step index, validated host-owned `EncryptedValue` account, and host-derived output handle.
- Remove the obsolete general per-operation event producer and its eight-event transport cap.
- Propagate the lifecycle schema and its independent version through the IDL, ABI golden, and generated host-listener types.
- Update Mollusk operator assertions to use durable output state as the ordinary-operation oracle.

## Why

Instruction reconstruction remains the source for ordinary operation semantics. A `make_public=true` output handle is different: it depends on block entropy and is not present in instruction data. This narrow lifecycle batch exposes only the state required for the relayer to validate an exact ordered one-to-one match against the successful host frame.

## Evidence

- 78 `zama-host` library tests.
- Production IDL build and PoC SBF builds for ZamaHost and confidential-token.
- Solana IDL, schema hashes, event versions, and ABI golden fixtures are in sync.
- Maximum 16-record CPI envelope test remains 1,077 bytes with one signed readonly event-authority account.
- Repository pre-commit formatting, full coprocessor `cargo check`, and CI-aligned clippy passed.
- Structural scan confirmed no finalized fetch queue, witness/lineage machinery, standalone listener, or general emitted-event decoder was restored.

The existing Mollusk suite covers zero, one, gapped multiple, maximum 16, wrong event authority, and transaction rollback against compiled SBF. Remote CI is the combined gate after this restack.

## Protocol change

The protocol-facing names are now:

- `PublicOutputsProducedEvent`
- `ProducedPublicOutput`
- `PUBLIC_OUTPUTS_PRODUCED_EVENT_VERSION`

The event version remains `1`. Renaming the Anchor event intentionally changes its discriminator; the regenerated IDL and ABI golden fixture pin the new value.

## Boundaries

This PR changes the host protocol boundary only. It does not make the relayer trust or consume the event. The stacked relayer slice must reject failed transactions, validate the host program and signed event-authority PDA, reconstruct the enclosing `fhe_eval`, and require an exact ordered one-to-one match before mutating state.

A frame with no produced public output emits no lifecycle event. Mollusk proves the compiled SBF CPI path and maximum-batch heap behavior; it does not claim that a direct maximum-size instruction fits the Solana transaction packet.

## Stack

- Base: #3089.
- Next: relayer validation and consumption for fhevm-internal#1675.
- Terminal: relayer-only two-leaf full vertical after the complete stack is assembled.


Eikix:
# Review — feat(coprocessor): emit Solana born-public output batches

*(review drafted by Claude, checked against Anchor internals and a local CI repro)*

## Overview

This PR replaces the retired per-operation `fhe_eval` event stream with a single versioned `BornPublicOutputsEvent` Anchor self-CPI, emitted once per successful `fhe_eval` frame that produced `make_public` durable outputs. Each record carries only the step index, the host-owned `EncryptedValue` account, and the block-entropy output handle — the one piece of state a relayer cannot reconstruct from instruction data. The old `event_budget` module (8-event CPI cap and the `FheEvalBornPublicFrameTooLarge` fail-closed guard) is deleted, the error slot is kept as a reserved discriminant, and the schema is propagated through the IDL, ABI golden manifest, and generated host-listener types.

Overall a well-executed, well-tested change — but it is currently **red in CI with a real, fixable defect**.

## Blocking: vendored IDL is out of sync (CI is failing)

`solana-tests/build-and-test` fails with:

```
error: zama_host: vendored IDL is out of sync with target/idl/zama_host.json; run solana/scripts/sync-zama-host-idl.sh
```

Root cause: `solana/programs/zama-host/src/errors.rs` changed the `#[msg]` for `FheEvalBornPublicFrameTooLarge` to `"reserved FHE eval born-public transport error"`, but the vendored `coprocessor/fhevm-engine/host-listener/idl/zama_host.json` (line ~1848) still carries the old message `"FHE eval born-public output requires a CPI-transportable frame"`. The check at `check_solana_abi.py:129` is full-JSON equality, so this mismatch trips it. The events/types were hand-added to the vendored IDL but the error message wasn't — running `solana/scripts/sync-zama-host-idl.sh` should fix it.

Two consequences worth flagging:
- In that job, the IDL check runs **before** `cargo test --workspace` and clippy, so **none of the new Mollusk tests actually ran in CI** — the PR's test evidence is local-only until this is fixed.
- The second failing job, `solana-tests/host-listener-reconstruct`, **passes locally at the PR head** (exact CI command, `cargo test -p host-listener --features solana-grpc,solana-reconstruct fhe_eval_acl`: 9 passed, 0 failed). Its failure log has expired, so the CI-side cause is unknown; re-run it after the sync fix and see if it was environmental.

## Correctness — checked and sound

- **Hand-rolled event CPI is faithful.** `emit_born_public_outputs` (`event_transport.rs`) was diffed against Anchor's actual `emit_cpi!` expansion in `anchor-attribute-event`: same `EVENT_IX_TAG_LE` prefix, same single readonly-signer event-authority meta, same `__event_authority` seeds and `crate::EVENT_AUTHORITY_AND_BUMP` (which is Anchor-codegen-generated). Byte-for-byte equivalent.
- **Authority validation is real.** `FheEval` carries `#[event_cpi]`, so Anchor validates the passed event-authority address, and the instruction meta uses the canonical PDA rather than the passed key — wrong-authority fails closed, and the Mollusk test proves it including no durable output surviving.
- **Payload size math checks out**: 8 (tag) + 8 (discriminator) + 1 (version) + 4 (vec len) + 16×66 = 1,077 bytes, matching the unit test. The golden fixture (`fixture_len` 79, discriminator `748fc0d139b7e3a0`) is internally consistent with the IDL discriminator bytes.
- **`build.rs` reuse is fine**: `collect_defined_types`/`enum_variants` already exist on the base branch (shared with instruction-type generation); nothing is missing.
- **Error-slot preservation** (keeping `FheEvalBornPublicFrameTooLarge` as a reserved variant) is the right call for Anchor discriminant stability.
- **Threading `op_index` through `accept_output` and returning the bound pubkey from `bind_eval_output`** is clean; admission correctly ignores the index; ordering is guaranteed by the walk.

## Suggestions (non-blocking)

1. **Consider `emit_cpi!` instead of the hand-rolled CPI.** Since it expands to exactly the code written here and `ctx` is in scope, `emit_cpi!(event)` behind the empty-check would drop ~25 lines of duplicated Anchor internals and track Anchor upgrades automatically. If the motivation is the exact-size unit test on `born_public_outputs_event_instruction`, a one-line comment saying "mirrors `emit_cpi!`; kept manual for the size/meta unit test" would prevent a future reviewer from re-deriving all of this.
2. **`generate_anchor_event_types` hardcodes `zama_host` version keys** while taking `idl_file` as a parameter. It's only called for `zama_host` today, but the generic signature invites a wrong reuse for `confidential_token`. Either pass the version names in or fold the function into `generate_zama_host_events`.
3. **`born_public_events()` in `host_mollusk.rs` filters inner instructions by data prefix only**; `assert_born_public_batch` checks the program id, but the zero-event assertion (`...emits_no_lifecycle_batch`) relies on the prefix alone. Harmless in single-program Mollusk, just slightly weaker than the positive-path assertion.
4. **`#[cfg(any(feature = "emit-events", test))]`** in `common.rs` is a mild smell (test-only import gate); fine as-is.

## Tests & security

Test coverage is genuinely strong: zero/one/gapped/max-16 batches, wrong event authority (fail-closed with no durable output), and transaction rollback, all against the real compiled SBF under Mollusk's CU budget; operator conformance was correctly re-oracled from emitted events to durable output state. Security posture is sound and honestly documented: the event grants no authority, and DD-038 spells out the consumer obligations (validate program, canonical PDA, tx success, ordered one-to-one match) that the stacked relayer slice must implement. The tx-packet caveat for a direct maximum-size instruction is candidly noted rather than papered over.

**Verdict: approve pending the vendored-IDL sync fix** — run `solana/scripts/sync-zama-host-idl.sh`, push, and confirm the workspace tests and the reconstruct job go green (the latter passed in a local repro).

PR #3138, opened 2026-07-13
feat(relayer): validate produced-public output batches

Eikix:
Part of zama-ai/fhevm-internal#1675.

## Summary

- Replace legacy per-operation event correlation with exact `PublicOutputsProducedEvent` validation.
- Preserve top-level instruction indexes and RPC `stackHeight` so each lifecycle batch is bound to its immediate enclosing `fhe_eval`.
- Require the configured host program, canonical event-authority PDA, supported lifecycle version, bounded exact Borsh payload, and ordered one-to-one step/account records.
- Reject failed transactions before instruction decoding and fail closed on malformed execution topology or lifecycle data.
- Preserve the existing atomic replay/store boundary and confirmed MMR peak verification.

## Validation contract

For each successful `fhe_eval`, the relayer derives the expected produced-public `(step_index, encrypted_value)` records from the instruction plan. A frame with no produced public output must have no lifecycle batch; a nonempty frame must have exactly one immediate-child batch. Missing, extra, reordered, duplicate, malformed, oversized, unknown-version, wrongly nested, or unconsumed batches reject the entire transaction decode before store or cursor mutation.

RPC inner instructions do not expose CPI signer/readonly flags. Authentication therefore uses a confirmed successful transaction, the exact host event-dispatch instruction, and the exact seed-derived event-authority PDA. PR #3136 separately proves that the on-chain producer requests one signed readonly account meta.

## Operational contract

- Deploy relayer support before deploying a producer with a new lifecycle version; unknown versions intentionally halt ingestion.
- `stackHeight` is required for every inner instruction and is the RPC compatibility floor for this path.
- The producer emits no event for an empty batch; an encoded batch must contain `1..=MAX_FHE_EVAL_OPS` records.

## Evidence

- 65 Solana-proof subsystem tests passed after restack.
- Relayer formatting passed.
- Failed-transaction decoding and confirmed-RPC parameter tests are both retained.
- The atomic regression still proves lifecycle decode failure leaves the file-store cursor unset.
- The confirmed-peak mismatch regression still rejects a forged event handle.

## Boundaries

This PR consumes the protocol boundary introduced by #3136. It does not change the producer, replay/store architecture, Docker stack, or terminal full-vertical flow. General operation events remain non-authoritative and are not reintroduced.

## Stack

- Base: #3136.
- Next: terminal relayer-only two-leaf proof flow.

### DIFF
--- a/relayer/src/solana_proof/chain.rs
+++ b/relayer/src/solana_proof/chain.rs
@@ -190,6 +190,8 @@ struct CompiledIx {
     program_id_index: usize,
     accounts: Vec<usize>,
     data: String,
+    #[serde(rename = "stackHeight", default)]
+    stack_height: Option<u32>,
 }
 
 #[derive(Deserialize)]
@@ -235,9 +237,56 @@ struct GetTransactionResult {
     meta: Option<Meta>,
 }
 
+fn decode_transaction_result(
+    signature: &str,
+    parsed: GetTransactionResult,
+) -> Result<ChainTransaction, ChainError> {
+    if parsed.meta.as_ref().and_then(|m| m.err.as_ref()).is_some() {
+        // This guard deliberately precedes account-key and instruction decoding:
+        // failed transactions committed no state and must contribute no leaves.
+        return Ok(ChainTransaction {
+            signature: signature.to_string(),
+            slot: parsed.slot,
+            instructions: Vec::new(),
+        });
+    }
+    let mut account_keys: Vec<[u8; 32]> = parsed
+        .transaction
+        .message
+        .account_keys
+        .iter()
+        .map(|s| base58_to_32(s))
+        .collect::<Result<_, _>>()?;
+    if let Some(loaded) = parsed
+        .meta
+        .as_ref()
+        .and_then(|m| m.loaded_addresses.as_ref())
+    {
+        for addr in loaded.writable.iter().chain(loaded.readonly.iter()) {
+            account_keys.push(base58_to_32(addr)?);
+        }
+    }
+    let instructions = flatten_execution_order(
+        &parsed.transaction.message.instructions,
+        parsed
+            .meta
+            .as_ref()
+            .map(|m| m.inner_instructions.as_slice())
+            .unwrap_or(&[]),
+        &account_keys,
+    )?;
+    Ok(ChainTransaction {
+        signature: signature.to_string(),
+        slot: parsed.slot,
+        instructions,
+    })
+}
+
 fn compiled_to_raw(
     ix: &CompiledIx,
     account_keys: &[[u8; 32]],
+    top_level_index: usize,
+    is_inner: bool,
 ) -> Result<RawInstruction, ChainError> {
     let program_id = *account_keys
         .get(ix.program_id_index)
@@ -257,6 +306,11 @@ fn compiled_to_raw(
         program_id,
         accounts,
         data,
+        top_level_index,
+        // RPC omits stackHeight on message instructions; their height is known.
+        // Inner instructions retain the RPC field so missing nesting metadata
+        // can be rejected by the lifecycle decoder.
+        stack_height: if is_inner { ix.stack_height } else { Some(1) },
     })
 }
 
@@ -320,46 +374,7 @@ impl ChainFetcher for RpcChainFetcher {
         }
         let parsed: GetTransactionResult =
             serde_json::from_value(result).map_err(|e| ChainError::Rpc(e.to_string()))?;
-        if parsed.meta.as_ref().and_then(|m| m.err.clone()).is_some() {
-            // A failed transaction never persisted its instructions' effects on-chain.
-            return Ok(Some(ChainTransaction {
-                signature: signature.to_string(),
-                slot: parsed.slot,
-                instructions: Vec::new(),
-            }));
-        }
-        let mut account_keys: Vec<[u8; 32]> = parsed
-            .transaction
-            .message
-            .account_keys
-            .iter()
-            .map(|s| base58_to_32(s))
-            .collect::<Result<_, _>>()?;
-        if let Some(loaded) = parsed
-            .meta
-            .as_ref()
-            .and_then(|m| m.loaded_addresses.as_ref())
-        {
-            for addr in loaded.writable.iter().chain(loaded.readonly.iter()) {
-                account_keys.push(base58_to_32(addr)?);
-            }
-        }
-
-        let instructions = flatten_execution_order(
-            &parsed.transaction.message.instructions,
-            parsed
-                .meta
-                .as_ref()
-                .map(|m| m.inner_instructions.as_slice())
-                .unwrap_or(&[]),
-            &account_keys,
-        )?;
-
-        Ok(Some(ChainTransaction {
-            signature: signature.to_string(),
-            slot: parsed.slot,
-            instructions,
-        }))
+        Ok(Some(decode_transaction_result(signature, parsed)?))
     }
 
     async fn get_lineage_state(
@@ -396,14 +411,44 @@ fn flatten_execution_order(
     let mut by_index: std::collections::HashMap<usize, &Vec<CompiledIx>> =
         std::collections::HashMap::new();
     for group in inner_groups {
-        by_index.insert(group.index, &group.instructions);
+        if group.index >= top_level.len() {
+            return Err(ChainError::Rpc(format!(
+                "inner-instruction group index {} is out of range",
+                group.index
+            )));
+        }
+        if by_index.insert(group.index, &group.instructions).is_some() {
+            return Err(ChainError::Rpc(format!(
+                "duplicate inner-instruction group index {}",
+                group.index
+            )));
+        }
+        let mut previous_height = 1u32;
+        for (position, instruction) in group.instructions.iter().enumerate() {
+            let height = instruction.stack_height.ok_or_else(|| {
+                ChainError::Rpc(format!(
+                    "inner instruction {position} in group {} has no stackHeight",
+                    group.index
+                ))
+            })?;
+            if height < 2
+                || (position == 0 && height != 2)
+                || height > previous_height.saturating_add(1)
+            {
+                return Err(ChainError::Rpc(format!(
+                    "impossible stackHeight {height} at inner instruction {position} in group {}",
+                    group.index
+                )));
+            }
+            previous_height = height;
+        }
     }
     let mut out = Vec::new();
     for (i, ix) in top_level.iter().enumerate() {
-        out.push(compiled_to_raw(ix, account_keys)?);
+        out.push(compiled_to_raw(ix, account_keys, i, false)?);
         if let Some(inner) = by_index.get(&i) {
             for inner_ix in inner.iter() {
-                out.push(compiled_to_raw(inner_ix, account_keys)?);
+                out.push(compiled_to_raw(inner_ix, account_keys, i, true)?);
             }
         }
     }
@@ -455,11 +500,13 @@ mod tests {
                 program_id_index: 0,
                 accounts: vec![],
                 data: "".to_string(),
+                stack_height: None,
             },
             CompiledIx {
                 program_id_index: 1,
                 accounts: vec![],
                 data: "".to_string(),
+                stack_height: None,
             },
         ];
         let inner_groups = vec![InnerIxGroup {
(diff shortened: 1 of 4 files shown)


Eikix:
## Review

### Overview

Replaces the fragile per-operation event correlation (nine `OP_EVENT_NAMES`, count-must-match-plan heuristic) with exact validation of the single `BornPublicOutputsEvent` batch introduced by #3136. The relayer now:

- carries `top_level_index` + RPC `stackHeight` on every `RawInstruction` so each event batch is bound to its *immediate enclosing* `fhe_eval` frame,
- validates the event envelope (host program, exact event-authority PDA, version, bounded exact-length Borsh payload, ordered 1:1 step/account records),
- fails closed on any topology or payload anomaly, rejecting the whole transaction before store/cursor mutation.

Net −440/+685 with most additions being tests. A clear security and robustness improvement over the positional correlation it replaces.

### Correctness — verified sound

- **Frame delimitation** (`decode.rs`, frame scan in `decode_program_instructions`): the contiguous run of instructions with `top_level_index == eval's` and `stack_height > eval_height` is exactly the eval's descendants — valid only because `flatten_execution_order` enforces stack-trace monotonicity (first inner = 2, jumps ≤ +1), which the new `chain.rs` validation provides. The two layers compose correctly.
- **Forged-event paths**: a batch outside any frame hits `UnexpectedBornPublicEvent` at the top of the loop; a batch nested deeper than `eval_height + 1` hits `InvalidBornPublicEnvelope`; two batches hit `UnexpectedBornPublicEvent`; a third party can't produce a *successful* host event-dispatch instruction without the PDA's signer seeds, and failed transactions are discarded before decoding. The signer-flag gap in RPC inner instructions is honestly documented and pinned by #3136.
- **Panic safety**: `&ix.data[8..]` is guarded by `is_fhe_eval` checking `data.len() >= 8`; the `u32` record count is `checked_mul`/`checked_add` guarded before allocation-bearing deserialize.
- **Failed-tx guard moved ahead of key decoding** in `decode_transaction_result` is right — a failed tx can legitimately contain undecodable keys/indexes, and the new regression test covers exactly that.

### Issues / risks (none blocking)

1. **Version bump wedges ingestion — deploy-order coupling** (operational, deliberate). `UnsupportedBornPublicVersion` rejects the transaction, `poll_once` errors, and the cursor never advances. A future on-chain bump of `BORN_PUBLIC_OUTPUTS_EVENT_VERSION` halts every relayer until it's upgraded — the relayer must always be deployed *before* the host emits a new version. Fail-closed is the right call; this contract deserves a sentence in the module doc or on the host constant so a future producer change doesn't discover it in production.

2. **`stackHeight` availability is now load-bearing** (`chain.rs`). Inner instructions without `stackHeight` hard-error. Present on all modern RPC responses, but backfilling old transactions or talking to an old RPC node wedges ingestion on the first inner instruction. Fine for a greenfield subsystem — noting it's a new RPC-version floor.

3. **Empty-batch asymmetry with the producer.** `decode_born_public_event` requires `count >= 1`, and a frame with no born-public outputs must have *zero* batches. This is only consistent if #3136's producer never emits an empty batch — that invariant lives across two PRs with nothing tying them together except the description. A cross-referencing comment on the `1..=MAX_FHE_EVAL_OPS` bound (like the one already there for the signer meta) would make the pairing explicit.

4. **Redundant plan parse** (minor): the `fhe_eval` branch of `decode_program_instructions` inlines `FheEvalArgs::deserialize` on `&ix.data[8..]`, duplicating `decode_fhe_eval_instruction`'s body minus the `debug_assert`. A small `parse_fhe_eval_plan(ix)` helper would keep the two from drifting — purely cosmetic.

5. **Unreachable guard** (informational): in `validate_born_public_event`, the `handles_by_step.get_mut(...).ok_or(BornPublicMismatch)` can't fire — `record.step_index` already matched an `expected` entry built from `enumerate()` over `plan.steps`, so it's always `< step_count`. Harmless defense-in-depth; no change needed.

### Code quality & tests

- Error taxonomy is precise and each variant has a dedicated test — missing, extra, reordered, duplicate (both account and handle), malformed, huge declared count, oversized, unknown version, wrong envelope, wrong nesting, unconsumed, missing stack metadata, host-descendant. An unusually complete negative-path matrix.
- The nested-eval test covers the CPI-wrapper case; `chain.rs` tests cover duplicate/orphan inner groups and impossible stack traces including the valid `[2,3,3,2]` shape.
- Deleting the 9-event test scaffolding (~250 lines) in favor of one batch builder is a real simplification.

### Verdict

Approve-quality. The validation contract is coherent, adversarial paths are closed or documented as pinned by #3136, and coverage is thorough. Only actionable suggestions are documentation-level: state the version-bump deploy-order contract (item 1) and cross-reference the empty-batch/producer invariant to #3136 (item 3).

---
*Review drafted by an agent, posted on request.*

Eikix:
*(review drafted by Claude)*

## Verdict: Approve, non-blocking suggestions only

This PR does what its description claims. It replaces the previous per-op-event correlation heuristic in the relayer's Solana-proof decoder with exact validation of the versioned `PublicOutputsProducedEvent` batch introduced by #3136, and the trust-boundary checks called out in the description were independently verified against both the relayer diff and the producer code on the stacked branch.

Stack note: base is `elias/solana-born-public-output-batch` (the branch backing #3136, not `main`), so this diff is reviewed purely as an increment on top of that work, not against `main`.

### What changed (4 files, +692/-440)

- `relayer/src/solana_proof/chain.rs`: RPC decoding now carries `stackHeight` and `top_level_index` through to each flattened instruction, rejects failed transactions before any account/instruction decoding (`decode_transaction_result`, refactored out of the old inline branch), and validates inner-instruction groups (no duplicate/orphan group index, monotonic-by-one `stackHeight` starting at 2).
- `relayer/src/solana_proof/decode.rs`: the old `op_event_result`/`OP_EVENT_NAMES` heuristic (matching any of 9 op-event kinds and trusting positional count-based correlation) is deleted. In its place, `decode_program_instructions` reconstructs each `fhe_eval`'s CPI frame from `top_level_index`/`stack_height`, requires an exact one-to-one ordered match between the plan's `make_public` durable outputs and the frame's `PublicOutputsProducedEvent` records, and rejects on any deviation (missing/extra/reordered/duplicate/malformed/oversized/unknown-version/wrongly-nested/unconsumed batches, or any non-event host descendant inside the frame).
- `relayer/src/solana_proof/ingest.rs`, `proof.rs`: test-only updates (new fields on `RawInstruction`, new atomicity regression, renamed event builder helpers).

### Correctness analysis (verified against code, not just the PR body)

- **Failed-tx rejection precedes decoding**: `decode_transaction_result` in `chain.rs` checks `meta.err` and returns an empty instruction list before touching `account_keys`/`instructions`, confirmed both by reading the function and by the new test `failed_transaction_is_rejected_before_instruction_decoding`, which feeds a transaction with an invalid base58 key and a malformed instruction that would panic if decoding were attempted.
- **Host program + canonical PDA check**: `decode_born_public_event` requires `ix.program_id == program_id` (the caller-supplied host program) and `ix.accounts.as_slice() == [zama_host::EVENT_AUTHORITY_AND_BUMP.0.to_bytes()]` — an exact one-element match, not just "contains". Confirmed against `solana/programs/zama-host/src/instructions/fhe_eval/event_transport.rs` on the base branch: the producer's `public_outputs_produced_event_instruction` emits exactly one account, `EVENT_AUTHORITY_AND_BUMP.0`, via `invoke_signed` with the `__event_authority` seeds — so the relayer's check matches the producer's real on-chain shape.
- **Frame reconstruction and "immediate child" requirement**: `eval_height = ix.stack_height`, `event_height = eval_height + 1`, and the code walks forward while `child.top_level_index == ix.top_level_index && child_height > eval_height` to bound the frame, then requires the born-public event's `stack_height == Some(event_height)` exactly (not merely "somewhere in the frame"). This matches the producer, which reaches the event via a single `invoke_signed` (one CPI level) directly from `fhe_eval`, confirmed in `event_transport.rs`. A test (`extra_unconsumed_and_wrongly_nested_batches_fail_closed`) explicitly nests the event one level deeper and confirms `InvalidBornPublicEnvelope`.
- **Exact ordered one-to-one match**: `validate_born_public_event` first rejects duplicate accounts/handles via `HashSet`, then requires `actual.len() == expected.len()` and pairwise `(step_index, encrypted_value)` equality in order — genuinely an exact ordered match, not a set comparison. Covered by `reordered_or_mismatched_born_public_batch_fails_closed` and `duplicate_accounts_and_handles_fail_closed`.
- **Bounded, exact Borsh payload**: version is read first and checked against `zama_host::PUBLIC_OUTPUTS_PRODUCED_EVENT_VERSION` before any record parsing; the declared record count is bounds-checked (`1..=MAX_FHE_EVAL_OPS`) and the exact expected byte length is checked via `checked_mul`/`checked_add` *before* `Vec<ProducedPublicOutput>::deserialize` is invoked, so a huge declared count fails on the length check rather than triggering a large allocation attempt — confirmed by `huge_declared_record_count_fails_before_record_decode` (count = `u32::MAX`) and by reading the `checked_mul`/`checked_add` guard directly. `RECORD_BYTES = 66` was checked against the actual struct on the base branch (`step_index: u16` + `encrypted_value: Pubkey` + `output_handle: [u8;32]` = 2+32+32 = 66) — it's correct today but is a hand-maintained constant with no static assertion tying it to the struct (see suggestion below). Trailing-byte check (`!body.is_empty()`) after deserialization catches over-long payloads that satisfy the count-based length check by coincidence.
- **Fail-closed on unexpected host descendants**: any host-program instruction inside an `fhe_eval` frame that isn't the born-public event triggers `UnexpectedHostDescendant`, which also has the effect of disallowing nested `fhe_eval` calls inside another `fhe_eval`'s frame — consistent with the producer not being expected to self-CPI into `fhe_eval` recursively.
- **Atomic replay/store boundary preserved**: the new `ingest.rs` test `decode_failure_does_not_advance_cursor_or_store_state` confirms `poll_once` leaves the cursor unset (`None`) when decode fails, matching the PR's atomicity claim.
- **Confirmed-peak / forged-handle regression retained**: `proof.rs`'s existing test asserting a forged `TrivialEncryptEvent`-equivalent handle is still rejected at the MMR-peak check was retained and only renamed to use the new event builder (`born_public_event_ix`), not weakened.

No correctness issues were found in the diff itself.

### Test coverage

Strong and directly maps to the threat model listed in the PR body: missing event, extra/unexpected event on an empty-output frame, reordered records, duplicate account/handle, malformed payload, unknown version, wrong envelope (wrong account), oversized batch, huge declared count, unconsumed/misplaced batch, wrongly-nested batch, non-event host descendant inside the frame, missing `stackHeight` metadata, and the pre-existing failed-tx / atomicity / forged-handle regressions. `chain.rs` also gained direct unit tests for duplicate/orphan inner-instruction group indices and impossible `stackHeight` sequences.

### Security considerations

- The relayer's authentication model (successful tx + exact host program + exact `__event_authority` PDA, since RPC compiled inner instructions don't expose CPI signer flags) is explicitly documented as relying on #3136 to guarantee the producer requests a signed, readonly account meta for that PDA. This is a real cross-PR trust dependency — the relayer-side check alone (account list equality) cannot itself prove the CPI was actually signed; it trusts that only the host program's `fhe_eval` code path can reach that exact self-CPI shape via `invoke_signed`. That was spot-checked and holds against the base-branch producer code, but reviewers of #3136 should confirm signer-seed derivation can't be replicated by anything other than the host program's PDA-signing invocation.
- `decode_program_instructions`'s upfront check (`instructions.iter().any(|ix| matches!(ix.stack_height, None | Some(0)))`) rejects the **entire transaction's instruction set** if *any* instruction — including ones unrelated to the host program — is missing `stackHeight`. This is a broad blast radius by design (per the "operational contract" section: `stackHeight` is treated as an RPC compatibility floor for this whole path), but worth flagging explicitly as a deliberate availability/strictness trade-off rather than an oversight.
- Defense-in-depth is preserved: the PR correctly notes the confirmed MMR peak check at proof time (DD-035) remains the ultimate authority even though the new validation is much stricter than the old op-event heuristic.

### Non-blocking suggestions

1. `RECORD_BYTES: usize = 66` in `decode.rs` hardcodes the serialized size of `zama_host::events::ProducedPublicOutput`. Nothing ties it to the struct definition, so a future field change in that struct (in a different crate) would silently desynchronize the two. A `static_assertions::const_assert_eq!` (or equivalent compile-time check) tying `RECORD_BYTES` to `core::mem::size_of`/a serialized-length constant exported by `zama_host` would remove the manual-sync risk.
2. Consider a short comment on `RawInstruction::stack_height: Option<u32>` documenting that `None`/`Some(0)` are both treated as fatal for decode (not just "unknown") — it's enforced in `decode_program_instructions` but a reader of `decode.rs`'s struct definition alone wouldn't see that invariant.
3. `relayer-test/integration-test` was still `pending` at review time (unit-test and clippy both passed). Worth confirming it's green before merge, since it's the job most likely to exercise the new frame-reconstruction path end-to-end.

PR #3044, opened 2026-07-07
feat(coprocessor): create initial implementations of both upgrade-controller and consensus-detector

goshawk-3:
- Extracts the `consensus-detector` and `upgrade-controller` crates plus a new e2e test suite from the feature branch: https://github.com/zama-ai/fhevm/pull/2604
-  Extending the set of tables merged back into public at cutover.
- Hardens the latest-main image resolver so branch-only/unpublished packages no longer stall resolution, with unit tests.
- Adds sqlx query caching for the consensus-detector state_hash queries.

####  Fixes

-  fixes https://github.com/zama-ai/fhevm-internal/issues/1598
- fixes https://github.com/zama-ai/fhevm-internal/issues/1603

PR #3139, opened 2026-07-13
feat(coprocessor): move upgrade_state INSERT to host-listener

PanGan21:
## Summary

Moves the `upgrade_state` write from the upgrade-controller to the host-listener. This makes the upgrade state reliable even if the controller is stopped or misses a notification.

**host-listener**

- Validates `CoprocessorUpgradeProposed` and saves the GCS upgrade row in the same transaction.
- Sends the notification only after the row has been saved.
- Ignores duplicate events, protects an upgrade already in progress, and accepts only newer proposals after a finished or failed upgrade.
- Stores the event’s block number to reject old replayed proposals and clears state left by the previous upgrade.

**upgrade-controller**

- Reads the saved row instead of rebuilding it from the notification.
- Checks the row at startup and every poll, allowing it to restart readiness checks or resume an interrupted cutover.
- Ties every background task and state change to the matching proposal and event block, so an old task cannot change a newer upgrade.
- Cleans up host and Gateway readiness listeners when their task ends.

**cleanup safety**

- Before removing old computations and proofs, the controller checks that they still belong to the active proposal.
- Cleanup uses the same database lock as cutover and rollback, so they cannot overlap.
- Readiness tasks wake up after a rollback and stop when their proposal is no longer active.

Related: https://github.com/zama-ai/fhevm-internal/issues/1679

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
@@ -1120,9 +1120,7 @@ mod tests {
         use sqlx::Row;
         use test_harness::instance::{setup_test_db, ImportMode};
 
-        let instance = setup_test_db(ImportMode::WithKeysNoSns)
-            .await
-            .expect("test db");
+        let instance = setup_test_db(ImportMode::None).await.expect("test db");
         let pool = PgPoolOptions::new()
             .max_connections(4)
             .connect(instance.db_url())
@@ -1190,9 +1188,7 @@ mod tests {
         use sqlx::Row;
         use test_harness::instance::{setup_test_db, ImportMode};
 
-        let instance = setup_test_db(ImportMode::WithKeysNoSns)
-            .await
-            .expect("test db");
+        let instance = setup_test_db(ImportMode::None).await.expect("test db");
         let pool = PgPoolOptions::new()
             .max_connections(4)
             .connect(instance.db_url())
--- a/coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
@@ -213,7 +213,6 @@ pub async fn ingest_block_logs(
     confidential_bridge_address: &Option<Address>,
     options: IngestOptions,
 ) -> Result<(), sqlx::Error> {
-    let gcs_mode = db.gcs_mode();
     let Some(mut tx) = db.new_transaction().await? else {
         info!("cutover completed — host-listener skipping block ingest (retired stack)");
         return Ok(());
@@ -318,8 +317,7 @@ pub async fn ingest_block_logs(
                 .as_ref()
                 .is_some_and(|addr| &log.inner.address == addr);
         if is_protocol_config_address {
-            handle_protocol_config_log(&mut tx, chain_id, gcs_mode, log)
-                .await?;
+            handle_protocol_config_log(&mut tx, chain_id, log).await?;
             continue;
         }
 
@@ -612,15 +610,21 @@ const UPGRADE_ACTIVATED_CHANNEL: &str = "event_upgrade_activated";
 async fn handle_protocol_config_log(
     tx: &mut sqlx::Transaction<'_, sqlx::Postgres>,
     chain_id: ChainId,
-    gcs_mode: bool,
     log: &Log,
 ) -> Result<(), sqlx::Error> {
     match ProtocolConfig::ProtocolConfigEvents::decode_log(&log.inner) {
         Ok(event) => match &event.data {
             ProtocolConfig::ProtocolConfigEvents::CoprocessorUpgradeProposed(proposed) => {
-                let proposal_block = log.block_number.and_then(|b| i64::try_from(b).ok());
-                notify_coprocessor_upgrade_proposed(tx, chain_id, gcs_mode, proposed, proposal_block)
-                    .await?;
+                let Some(proposal_block) =
+                    log.block_number.and_then(|b| i64::try_from(b).ok())
+                else {
+                    warn!(
+                        proposal_id = %proposed.proposalId,
+                        "Ignoring CoprocessorUpgradeProposed without a valid block number"
+                    );
+                    return Ok(());
+                };
+                notify_coprocessor_upgrade_proposed(tx, chain_id, proposed, proposal_block).await?;
             }
             other => {
                 warn!(
@@ -643,9 +647,8 @@ async fn handle_protocol_config_log(
 async fn notify_coprocessor_upgrade_proposed(
     tx: &mut sqlx::Transaction<'_, sqlx::Postgres>,
     chain_id: ChainId,
-    gcs_mode: bool,
     event: &ProtocolConfig::CoprocessorUpgradeProposed,
-    proposal_block: Option<i64>,
+    proposal_block: i64,
 ) -> Result<(), sqlx::Error> {
     let listener_chain_id = chain_id.as_u64();
 
@@ -693,8 +696,10 @@ async fn notify_coprocessor_upgrade_proposed(
         return Ok(());
     };
 
-    // Guarded singleton upsert in the decode tx (durable); NOTIFY below fires only when a row applied.
-    let stack_role = if gcs_mode { "GCS" } else { "BCS" };
+    sqlx::query("DELETE FROM upgrade_state WHERE stack_role = 'BCS'")
+        .execute(&mut **tx)
+        .await?;
+
     let result = sqlx::query(
         r#"
         INSERT INTO upgrade_state (
@@ -702,7 +707,7 @@ async fn notify_coprocessor_upgrade_proposed(
             start_block, end_block, gw_start_block, host_chain_id,
             host_consensus_reached, gw_consensus_reached, proposal_block, updated_at
         )
-        VALUES ($1, 'UpgradeActivated', 'in_progress', $2, $3, $4, $5, $6, $7, FALSE, FALSE, $8, NOW())
+        VALUES ('GCS', 'UpgradeActivated', 'in_progress', $1, $2, $3, $4, $5, $6, FALSE, FALSE, $7, NOW())
         ON CONFLICT (stack_role) DO UPDATE
         SET state              = EXCLUDED.state,
             status             = EXCLUDED.status,
@@ -721,16 +726,13 @@ async fn notify_coprocessor_upgrade_proposed(
             gw_dry_run_started     = FALSE,
             last_error         = NULL,
             updated_at         = NOW()
-        WHERE (upgrade_state.state IN ('LIVE', 'PAUSED')
-            OR upgrade_state.status IN ('completed', 'failed'))
-          AND upgrade_state.proposal_id IS DISTINCT FROM EXCLUDED.proposal_id
-          -- Re-arm only from a later block; older replays are rejected. NULL stays permissive.
-          AND (upgrade_state.proposal_block IS NULL
-            OR EXCLUDED.proposal_block IS NULL
+        WHERE (upgrade_state.proposal_block IS NULL
             OR EXCLUDED.proposal_block > upgrade_state.proposal_block)
+          AND (upgrade_state.status = 'failed'
+            OR (upgrade_state.status = 'completed'
+              AND upgrade_state.proposal_id IS DISTINCT FROM EXCLUDED.proposal_id))
         "#,
     )
-    .bind(stack_role)
     .bind(&proposal_id_bytes[..])
     .bind(&event.softwareVersion)
     .bind(start_block)
@@ -743,9 +745,8 @@ async fn notify_coprocessor_upgrade_proposed(
 
     if result.rows_affected() == 0 {
         warn!(
-            stack_role,
             proposal_id = %proposal_id_hex,
-            "Rejected event_upgrade_activated: another upgrade already in flight for this stack role"
+            "Rejected event_upgrade_activated: proposal is active, completed, or not newer"
         );
         return Ok(());
     }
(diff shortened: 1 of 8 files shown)


coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -169,15 +171,231 @@ fn decode_hex(s: &str) -> Result<Vec<u8>, Error> {
...
 }
 
-/// Handle an `event_upgrade_activated` notification: parse payload and upsert
-/// the FSM row with `state='UpgradeActivated'`, `status='in_progress'`. For
-/// GCS, additionally spawns the dry-run readiness loop.
+/// One `upgrade_state` row read by the FSM driver; column names match fields for `sqlx::FromRow`.
+#[derive(Debug, Clone, sqlx::FromRow)]
+struct UpgradeRow {
+    proposal_id: Vec<u8>,
+    chain_id: Option<i64>,
```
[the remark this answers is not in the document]
PanGan21: Addressed in: https://github.com/zama-ai/fhevm/pull/3139/changes/6f51ce77da5d614cc408553380cd45743b8a8969

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -169,15 +171,231 @@ fn decode_hex(s: &str) -> Result<Vec<u8>, Error> {
...
+
+/// Proposals with a drive in flight, so the same proposal is never driven twice at once.
+type InFlight = Arc<Mutex<HashSet<Vec<u8>>>>;
+
+/// Claim `proposal_id` for driving; `true` if acquired (release when done), `false` if already held.
+fn claim_in_flight(set: &InFlight, proposal_id: &[u8]) -> bool {
+    set.lock()
+        .expect("in-flight mutex poisoned")
+        .insert(proposal_id.to_vec())
+}
```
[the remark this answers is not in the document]
PanGan21: Addressed in: https://github.com/zama-ai/fhevm/pull/3139/changes/6f51ce77da5d614cc408553380cd45743b8a8969

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -169,15 +171,231 @@ fn decode_hex(s: &str) -> Result<Vec<u8>, Error> {
...
+          ORDER BY id DESC",
+    )
+    .bind(stack_role)
+    .fetch_all(pool)
+    .await?;
+    Ok(rows)
+}
+
+/// Proposals with a drive in flight, so the same proposal is never driven twice at once.
+type InFlight = Arc<Mutex<HashSet<Vec<u8>>>>;
```
PanGan21: Addressed in: https://github.com/zama-ai/fhevm/pull/3139/changes/6f51ce77da5d614cc408553380cd45743b8a8969

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -169,15 +171,250 @@ fn decode_hex(s: &str) -> Result<Vec<u8>, Error> {
...
+    let gw_start_block = row
+        .gw_start_block
+        .ok_or_else(|| Error::Payload("GCS row missing gw_start_block".into()))?;
+
+    // gw gate for the zkproof-worker (releases on gw_start_block, a different clock than host start_block); spawned to run concurrently.
+    {
+        let pool = pool.clone();
+        let cancel = cancel.child_token();
+        let proposal_id = proposal_id.clone();
+        tokio::spawn(async move {
```
PanGan21: [AUTOMATED] **Bug — poll-fallback re-dispatch leaks one gw-readiness watcher (and one pooled DB connection) per tick, exhausting the pool and stalling cutover**

While a GCS row sits in `DryRunStarted` with `gw_dry_run_started = FALSE` (host clock satisfied, Gateway not yet at `gw_start_block`), every poll tick (default 30s) re-dispatches the row through `reconcile_in_progress → dispatch_upgrade_state → drive_gcs_dry_run`:

1. `drive_gcs_dry_run` **unconditionally** spawns this `wait_until_gw_dry_run_ready` watcher, which opens a dedicated `PgListener` and blocks on `listener.recv()` until `gw_dry_run_started` flips true (no short-circuit while it is FALSE).
2. The inline `wait_until_dry_run_ready` returns `Ok(false)` almost immediately for a `DryRunStarted` row, so the `InFlightGuard` is released within milliseconds — it does **not** bound the spawned watchers, and their `CancellationToken` is a child of the global `run()` token, so nothing tears them down between ticks.
3. `PgListener::connect_with(&pool)` acquires and **holds** a connection from the bounded pool for the watcher's whole lifetime (sqlx-postgres 0.8.6 `listener.rs:59-62`, `pool.acquire()`), and the default `database_pool_size` is 4 (one already held by the main `run()` listener).

So the watchers accumulate one per tick and exhaust the pool within ~2–3 ticks (~60–90s) of any DryRunStarted→gw-ready window longer than that — after which `execute_cutover` (`pool.begin()`) and the reconcile scans block on `acquire` and time out: the poll fallback built to recover the upgrade ends up stalling its own cutover, with only generic pool-timeout errors in the logs.

**Suggested fix:** ensure at most one gw watcher exists per proposal — e.g. gate the spawn on a dedup keyed by `proposal_id` that is held for the *watcher's* lifetime (a second `InFlightGuard`-style set, or fold the gw axis into the existing `in_flight` claim), and/or skip the spawn when the row was dispatched in `DryRunStarted` with the watcher already running. Reading `gw_dry_run_started` into `UpgradeRow` and skipping the spawn when it is already TRUE would also avoid pointless spawns after the gate clears.

Confidence: 90/100 (validated: full path traced at head, sqlx pooling behavior confirmed against the locked 0.8.6 source)

PanGan21 (commented):
[AUTOMATED]

## Code Review — PR #3139 (head `04a99b1`)

Reviewed with parallel specialized agents (2× bug hunters, guideline compliance, error-handling audit), followed by independent validation of each candidate finding.

### Findings

**1 confirmed issue** (details inline): the poll-fallback re-dispatch leaks one gateway-readiness watcher — each holding a pooled DB connection — per poll tick while a GCS row sits in `DryRunStarted` waiting for the Gateway clock, exhausting the default 4-connection pool within ~2–3 ticks and stalling `execute_cutover` itself. Validated at 90/100 confidence against the sqlx 0.8.6 source.

### Checked and cleared

- ✅ Migration `20260713120000`: constraint name, identity-PK on populated table, `DELETE NULL` → `SET NOT NULL` ordering, and UNIQUE compatibility with legacy rows all verified correct
- ✅ Host-listener INSERT: column/bind mapping, `ON CONFLICT` arbiter matches the migration's UNIQUE constraint, notify gated on `rows_affected` inside the ingest tx (row + notify commit/rollback atomically)
- ✅ Payload round-trip (`proposal_id` hex ↔ BYTEA), all `upgrade_state` reads either proposal-scoped or newest-in-progress (`ORDER BY id DESC LIMIT 1`), no state-string mismatches vs the CHECK constraint
- ✅ Cutover concurrency: advisory lock + re-read-under-lock correctly serializes the unanimity path vs reconcile dispatch; `InFlightGuard` is panic-safe and never held across `.await`
- ✅ Telemetry style, rustfmt (80-col), and migration conventions
- ℹ️ Sub-threshold notes (not blocking, no action required): `.lock().unwrap()` on the guard mutex technically conflicts with the coprocessor "never panic" guideline though the critical section cannot panic in practice; a pre-migration in-flight GCS row would have `chain_id = NULL` and fail dispatch with a logged error (only relevant if migrating mid-upgrade).

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
PanGan21: Addressed in: https://github.com/zama-ai/fhevm/pull/3139/commits/896bfc428b8ad5937c2121c95afb6a9ecad0d92d

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -668,14 +672,49 @@ async fn notify_coprocessor_upgrade_proposed(
         return Ok(());
     };
 
+    // stack_role follows this listener's own gcs_mode: green binary → 'GCS', blue → 'BCS'.
+    let stack_role = if gcs_mode { "GCS" } else { "BCS" };
+
+    // Persist the FSM row in the ingest tx; ON CONFLICT (stack_role, proposal_id) makes duplicate decodes idempotent.
+    let result = sqlx::query(
+        r#"
+        INSERT INTO upgrade_state (
```
goshawk-3, line 705: As a side note, when we run a HL per host-chain they will overwrite each other as the primary key is `stack_role, proposal_id`. 

This should be fine until we implement  multi-chain support as per https://github.com/zama-ai/fhevm-internal/issues/1530

coprocessor/fhevm-engine/db-migration/migrations/20260713120000_upgrade_state_per_proposal.sql
```diff
@@ -0,0 +1,11 @@
...
+ALTER TABLE upgrade_state DROP CONSTRAINT upgrade_state_pkey;
+ALTER TABLE upgrade_state
+    ADD COLUMN id BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY;
+ALTER TABLE upgrade_state
+    ADD COLUMN chain_id BIGINT NULL CHECK (chain_id >= 0);
+DELETE FROM upgrade_state WHERE proposal_id IS NULL;
+ALTER TABLE upgrade_state ALTER COLUMN proposal_id SET NOT NULL;
+ALTER TABLE upgrade_state
+    ADD CONSTRAINT upgrade_state_stack_role_proposal_id_key
+    UNIQUE (stack_role, proposal_id);
```
goshawk-3: Why do we need this refactoring? 

We should always have a single GCS stack_role in the DB. We can make sure this with "singleton" column pattern.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -169,15 +171,281 @@ fn decode_hex(s: &str) -> Result<Vec<u8>, Error> {
...
+    .await?;
+    Ok(rows)
+}
+
+/// Per-proposal tasks in flight, so the same task is never run twice at once.
+type InFlight = Arc<Mutex<HashSet<(InFlightKind, Vec<u8>)>>>;
+
+/// Kind of per-proposal task an [`InFlightGuard`] claims.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+enum InFlightKind {
```
goshawk-3: Why do we need `kind of per-proposal task`? Do we need to support multiple proposals at the same time?

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -191,135 +459,20 @@ pub async fn handle_upgrade_activated(
         start_block = payload.start_block,
         end_block = payload.end_block,
         gw_start_block = payload.gw_start_block,
-        "event_upgrade_activated received — inserting upgrade_state row"
+        "event_upgrade_activated received — dispatching upgrade_state row"
     );
 
-    let result = sqlx::query(
-        r#"
-        INSERT INTO upgrade_state (
```
goshawk-3, line 203: Maybe, instead of refactoring   the upgrade-controller, we could simply change this to  `ON CONFLICT DO NOTHING` . That would make the PR shorter and give us less merge conflicts.

goshawk-3 (changes requested):

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
PanGan21: This code is only going to run from the host listener of the canonical chain (eth) so it will be skipped from other host listeners. I think in that case there will be no ovewrite

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
PanGan21: The enum distinguishes two concurrent tasks for the same proposal: the FSM action (drive or cutover dispatch) and the spawned gw watcher. It is not about supporting multiple proposals at the same time, the doc might be a bit missleading and I will change it

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
PanGan21: In that case I am not sure why we even need this pr, we can just implement this comment and then the refactor is not needed at all: https://github.com/zama-ai/fhevm/pull/3044#discussion_r3551159086

coprocessor/fhevm-engine/db-migration/migrations/20260713120000_upgrade_state_per_proposal.sql
PanGan21: Probably you are right that we should keep singleton. I am checking how to revert this. I thought we wanted multiple rows

PanGan21: I updated the logic and only moved the insert

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
PanGan21: This is removed now

Eikix (dismissed):

Eikix (approved):

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -615,7 +615,16 @@ async fn handle_protocol_config_log(
     match ProtocolConfig::ProtocolConfigEvents::decode_log(&log.inner) {
         Ok(event) => match &event.data {
             ProtocolConfig::ProtocolConfigEvents::CoprocessorUpgradeProposed(proposed) => {
-                notify_coprocessor_upgrade_proposed(tx, chain_id, proposed).await?;
+                let Some(proposal_block) =
```
PanGan21, line 618: e2e tests are failing: https://github.com/zama-ai/fhevm/actions/runs/30083541819/job/89450348430

so blocking the merge

PanGan21:
https://github.com/Mergifyio dequeue

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
PanGan21: successful run
https://github.com/zama-ai/fhevm/actions/runs/30091612503/job/89475954984

PR #3142, opened 2026-07-13
chore(kms-connector): docs follow-up for raw private key

eudelins-zama:
Follow-up of https://github.com/zama-ai/fhevm/pull/3121

- `Serialize` impl for `Config` only when compiled for `test` (not `debug_assertions`)
- Better name for the `TestingPrivateKey` struct matching its intent


dvdplm (approved):
LGTM

eudelins-zama:
@mergifyio queue

PR #3143, opened 2026-07-13
test(test-suite): rolling CU cost snapshots for Solana Mollusk profiles

Eikix:
## Why

Solana transaction-packing and budget decisions depend on compute-unit measurements that otherwise go stale silently. The confidential-swaps plan cited a pre-#3002 transfer cost from a test that no longer exists, so neither later regressions nor improvements could be reviewed against the decisions that used that number.

This adds committed Mollusk cost snapshots, following the same review model as a gas snapshot: any compute-unit change in either direction fails. A more expensive instruction may be a regression; a cheaper instruction may invalidate conservative packing or budget assumptions. Intentional changes update the JSON in the same diff, leaving reviewable and bisectable evidence.

## What

- `tests/support/cost_snapshot.rs` records exact compute units, unique account count, and instruction-data bytes in per-test-binary JSON snapshots.
- Dedicated tests keep cost drift separate from behavior assertions.
- Fixed fixture keys make PDA bump-search cost stable across runs.
- `bash scripts/update-cost-snapshots.sh` is the supported regen path: verifies CI-pinned Solana CLI 2.1.0 / Anchor 1.0.2, cleans, rebuilds SBF, then rewrites JSON only under `ZAMA_UPDATE_COST_SNAPSHOT=1`.
- Scripts hygiene in the same PR: `scripts/poc` → `scripts/e2e`, plus `scripts/README.md` catalog so entrypoints are obvious. Cargo `--features poc` is unchanged (test shims, not the live harness folder).

Current profiles on `feature/solana` after #3144:

- `confidential_transfer/direct`: 332,805 CU · 15 accounts · 223 B
- `confidential_transfer/steady_state`: 350,274 CU · 15 accounts · 223 B
- `initialize_token_account`: 65,393 CU · 12 accounts · 16 B
- `fhe_eval/three_op_frame`: 61,504 CU · 9 accounts · 375 B

Mollusk is not a mainnet cost oracle. These values include arbitrary but fixed PDA bump-search costs; the useful signal is the exact delta between commits under the pinned build/runtime toolchain.

## Update workflow

From `solana/`:

```text
bash scripts/update-cost-snapshots.sh
```

Use `--no-clean` only after a trusted rebuild. Prefer the script over setting `ZAMA_UPDATE_COST_SNAPSHOT=1` by hand (that escape hatch skips the toolchain/clean guardrails). Script catalog: `solana/scripts/README.md`.

## Validation

- clean SBF rebuild with Solana CLI 2.1.0 and Anchor 1.0.2;
- 76/76 host Mollusk tests;
- 34/34 confidential-token Mollusk tests;
- repeated exact snapshot assertion runs;
- `ZAMA_UPDATE_COST_SNAPSHOT=0` remains assertion-only;
- runtime-tests all-targets warnings-denied Clippy;
- formatting, IDL/schema/ABI synchronization, and diff checks;
- `rg 'scripts/poc'` clean after the e2e rename (CI + test-suite paths updated).


Eikix:
## Review (agent-generated, posted at Elias's request)

### Overview

Adds committed, gas-snapshot-style cost baselines for Solana Mollusk profiles: a small support module (`tests/support/cost_snapshot.rs`) that measures compute units, unique account count, and instruction-data bytes per profile, compares them exactly against per-binary JSON files under `cost-snapshots/`, and rewrites entries when `ZAMA_UPDATE_COST_SNAPSHOT=1`. Four profiles are snapshotted across the host and confidential-token test binaries, with fixed fixture keys so PDA bump-search cost stays stable.

### What works well

- The design decisions are sound and well-documented in-module: exact equality in both directions (cheaper also fails, since it invalidates packing assumptions), dedicated `cost_snapshot_*` tests so drift never fails behavior tests, and the honest caveat that Mollusk is a delta signal, not a mainnet oracle.
- Concurrency is handled correctly for the actual hazard: a per-process `Mutex` with poison recovery serializes read-modify-write within a binary, and distinct binaries write distinct files (host vs token), so parallel test binaries can't clobber each other.
- The update path is safely gated (`=1` exactly), still runs `Check::success()`, and inserting into the existing map preserves other profiles' entries.
- Fixed keys via `Pubkey::new_from_array([0x11; 32])` etc. are a clean way to pin bump-search cost; the rationale comment on `with_keys`/`with_block_cap_keys` explains exactly why the variant exists.
- `BTreeMap` + pretty JSON + trailing newline gives stable, diff-friendly snapshot files.

### Suggestions / minor issues

1. **Stale profiles are never pruned.** Update mode inserts/overwrites the current profile but nothing removes entries whose test was renamed or deleted — an orphan entry lingers in the JSON forever and silently reads as a live baseline. Not worth blocking on with 4 profiles, but consider either a comment acknowledging it, or (later) a completeness check that each JSON key was asserted at least once per run.
2. **Corrupted-file recovery in update mode.** If a snapshot file contains invalid JSON, `read_entries` panics even under `ZAMA_UPDATE_COST_SNAPSHOT=1`, so the documented update workflow can't regenerate it — the file has to be deleted by hand. In update mode it may be friendlier to treat unparseable content like a missing file.
3. **`&PathBuf` parameters** (`read_entries`, `write_entries`) would conventionally be `&Path` (clippy's `ptr_arg` usually flags this — worth double-checking why the warnings-denied run accepted it).
4. **`confidential_transfer/direct` masks `steady_state`.** Both snapshots are asserted sequentially in one test, so a drift in the first assertion hides whether the second also drifted until the first is fixed. Understandable since steady-state needs the first transfer's account creation; a line noting that coupling would help the next person who sees one failure become two.
5. **`unique_accounts` naming.** It counts unique pubkeys in the instruction's account metas plus the program id — a static property of the instruction shape, not something measured from execution. Fine as a tripwire; a one-word rename (`instruction_accounts`?) or a comment would prevent misreading it as runtime-loaded accounts.
6. **Module inclusion consistency.** `token_mollusk.rs` now has both `mod support;` and a `#[path = "support/cost_snapshot.rs"] mod cost_snapshot;` while `support/mod.rs` doesn't declare `cost_snapshot`. That works (each binary compiles its own copy, and host_mollusk avoids dragging in `cleartext_fhe_eval`), but it's worth a short comment near the `#[path]` attribute saying this is deliberate, since the obvious "cleanup" of moving it into `support/mod.rs` would force `host_mollusk` to compile the rest of the support module.

### Risks

- **Toolchain sensitivity is the main operational risk.** Exact CU equality means any local Solana CLI / Anchor / rustc drift produces false failures, and stale `target/deploy` artifacts produce wrong baselines. The PR mitigates this well (pinned versions in the doc comment, `cargo clean` + rebuild in the workflow), but expect occasional contributor friction; if it recurs, a version check in the update path could help.
- No security concerns: test-only code, no production surface. The new `serde` derive dependency is dev-scoped to the runtime-tests crate.

### Verdict

Solid, well-scoped PR that does what the description says. Nothing blocking — the suggestions above are polish.

Eikix:
_Agent-assisted review (Claude Code), posted on Elias's behalf._

## Overview

Adds a gas-snapshot-style regression harness for Solana compute costs: a shared test-support module (`tests/support/cost_snapshot.rs`) records exact compute units, unique account count, and instruction-data bytes per profile into committed JSON under `runtime-tests/cost-snapshots/`; four dedicated `cost_snapshot_*` tests fail on drift in either direction. A guarded regen script (`scripts/update-cost-snapshots.sh`) enforces the CI-pinned toolchain before rewriting baselines. The same PR renames `scripts/poc` → `scripts/e2e`, adds a `scripts/README.md` catalog, and adds usage headers to existing scripts.

## What holds up well

- **The design is sound and the key claims check out.** CI (`solana-tests.yml`) builds SBF via `check-zama-host-idl.sh` with pinned Solana CLI 2.1.0 / Anchor 1.0.2 before `cargo test --workspace`, so the snapshot tests run in CI under the same toolchain the update script enforces. CU codegen comes from the Solana CLI's bundled platform-tools, so pinning that (plus `rust-toolchain.toml` and `Cargo.lock`) is the right determinism lever.
- **No leftover references:** `git grep 'scripts/poc'` on the PR head is clean, including the e2e workflow.
- **No module duplication:** `support/mod.rs` declares only `cleartext_fhe_eval`, so the deliberate `#[path]` includes in both test binaries are safe and the rationale comments are accurate.
- Careful details throughout: `BTreeMap` for stable JSON ordering, poison-recovering mutex so one panicking cost test doesn't cascade, fixed fixture keys because PDA bump search is part of measured CU, direction-aware failure messages ("improvement — re-check packing decisions"), behavior checks (`Check::success()`) asserted before the snapshot so a functional break fails as itself.
- Snapshot tests are correctly separated from behavior tests, and per-binary snapshot files avoid cross-binary write races (the in-process mutex handles intra-binary ones).

## Issues

**1. Orphaned profiles accumulate silently (medium).** Update mode "inserts or overwrites, never deletes" — documented, with manual cleanup as the answer. But nothing ever *detects* an orphan: rename a test and the stale JSON key persists forever, looking like a live baseline. Since `update-cost-snapshots.sh` is the sanctioned full-regen path and already does a clean rebuild, it could simply delete `runtime-tests/cost-snapshots/*.json` before running the tests. That makes the script's output authoritative (exactly the live profiles) at zero cost, while the library's no-delete behavior stays correct for the manual env-var escape hatch.

**2. `--help` prints `set -euo pipefail` (minor).** In `update-cost-snapshots.sh`, the header comment spans lines 2–15 but the help handler does `sed -n '2,16p'`, so line 16 (`set -euo pipefail`) leaks into the help text. Change to `2,15p`, or better, print up to the first non-comment line so the range can't rot when the header grows.

**3. Loose Solana version match (minor).** `[[ "$solana_ver" == *" ${EXPECTED_SOLANA}"* ]]` is a substring test — `" 2.1.0"` would also accept a hypothetical `2.1.05`. Also, the second branch (`*" ${EXPECTED_SOLANA}-"*`) is dead: any string matching it already matches the first pattern. Anchoring on the version token (e.g. matching `solana-cli ${EXPECTED_SOLANA}` followed by space/`-`/end) would tighten this; low stakes since CI independently verifies its own toolchain.

**4. Unused derives (nit).** `Cost` derives `PartialEq, Eq` but comparisons are all field-wise for the sake of the per-field failure messages. Drop the derives or use them.

**5. Env gate accepts only exactly `"1"` (nit).** `ZAMA_UPDATE_COST_SNAPSHOT=true` (or `=yes`) silently runs in assert mode. Since the PR explicitly advertises `=0` as assertion-only, consider panicking on values other than `0`/`1`/unset so a typo'd update attempt fails loudly instead of looking like unexplained drift. Low priority given the script is the supported path.

## Accepted risks (worth stating, not blocking)

- **Exact-match snapshots are only as stable as SBF build reproducibility.** The PR's own validation (repeated exact runs, clean rebuild) plus toolchain pinning addresses this, and any toolchain bump will legitimately require a reviewed snapshot update — that's the intended behavior, but expect that PR to be noisy.
- **`direct` and `steady_state` share one test**, so a `direct` mismatch masks whether `steady_state` also moved until re-run. The tradeoff is documented in a comment and correct (steady-state needs the first transfer's account creation); fine as is.

## Verdict

High-quality PR — approve with the minor fixes above. Only #1 (orphan cleanup in the regen script) is worth addressing before merge; the rest are polish. Documentation is unusually thorough: the module doc, script headers, and `scripts/README.md` catalog all state scope, non-goals ("Mollusk is not a mainnet cost oracle"), and regen workflow clearly, and the `--features poc` vs `scripts/poc` disambiguation in the README preempts a real confusion.

PR #3149, opened 2026-07-13
fix(sdk): harden Solana config and proof hex

Eikix:
Solana runtime configuration is now compared across every public field after cloning nested logger, module-version, and auth values. Repeating an equivalent configuration with the same logger callbacks remains idempotent, while changed nested values are rejected. The exported Solana proof-hex parser now rejects non-hex characters instead of allowing `NaN` to become zero bytes, while preserving prefixed and unprefixed inputs. Focused Vitest coverage exercises both contracts, including valid round trips, malformed characters, prefixes, and odd lengths.

Related to zama-ai/fhevm-internal#1709.

### DIFF
--- a/sdk/js-sdk/src/solana/internal/config.test.ts
+++ b/sdk/js-sdk/src/solana/internal/config.test.ts
@@ -1,0 +1,60 @@
+import type { FhevmRuntimeConfig } from '../../core/types/coreFhevmRuntime.js';
+import { describe, expect, it, vi } from 'vitest';
+
+type ConfigModule = typeof import('./config.js');
+
+async function loadConfigModule(): Promise<ConfigModule> {
+  vi.resetModules();
+  return import('./config.js');
+}
+
+describe('Solana runtime config', () => {
+  it('allows an identical normalized config with the same logger callbacks', async () => {
+    const { setFhevmRuntimeConfig } = await loadConfigModule();
+    const debug = vi.fn();
+    const warn = vi.fn();
+    const error = vi.fn();
+    const locateFile = (file: string): URL => new URL(file, 'https://example.test/');
+    const config: FhevmRuntimeConfig = {
+      locateFile,
+      wasmAssetLoadMode: 'verified-blob',
+      moduleVersions: { tfhe: '1.6.1', kms: '0.13.20-0', checkCompatibility: 'warn' },
+      logger: { debug, warn, error },
+      singleThread: false,
+      numberOfThreads: 4,
+      auth: { type: 'ApiKeyHeader', header: 'x-api-key', value: 'secret' },
+    };
+
+    setFhevmRuntimeConfig(config);
+
+    expect(() =>
+      setFhevmRuntimeConfig({
+        ...config,
+        logger: { debug, warn, error },
+        moduleVersions: { tfhe: '1.6.1', kms: '0.13.20-0', checkCompatibility: 'warn' },
+        auth: { type: 'ApiKeyHeader', header: 'x-api-key', value: 'secret' },
+      }),
+    ).not.toThrow();
+  });
+
+  it('rejects a changed logger callback', async () => {
+    const { setFhevmRuntimeConfig } = await loadConfigModule();
+    const logger = { debug: vi.fn(), error: vi.fn() };
+
+    setFhevmRuntimeConfig({ logger });
+
+    expect(() => setFhevmRuntimeConfig({ logger: { ...logger, debug: vi.fn() } })).toThrow(
+      'FhevmRuntime config has already been set and cannot be changed.',
+    );
+  });
+
+  it('rejects changed nested configuration', async () => {
+    const { setFhevmRuntimeConfig } = await loadConfigModule();
+
+    setFhevmRuntimeConfig({ auth: { type: 'ApiKeyHeader', value: 'first' } });
+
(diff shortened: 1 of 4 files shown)


Eikix:
## Adversarial review (agent-generated, dispatched by Elias) — head `c7b422a`

### (a) Independent statement of intended flow

Two separate boundary defects in the Solana SDK adapter:

1. **Config idempotency.** `setFhevmRuntimeConfig` (`sdk/js-sdk/src/solana/internal/config.ts`) is contractually idempotent: callable repeatedly with identical parameters, throws on a genuine change. On first call it stores a **frozen clone** of `logger`. The old repeated-call guard compared `stored.logger !== config.logger` — but `stored.logger` is the clone, so it can never be reference-equal to the caller's original object. Result: a second call with the *same* logger throws. Fix: compare by field/callback identity instead of object reference, and extend comparison to all config fields (the old guard also silently ignored `wasmAssetLoadMode`, `moduleVersions`, and `auth`).

2. **Proof hex parsing.** `hexToBytes` in `sdk/js-sdk/src/solana/proof.ts` (re-exported publicly as `solanaProofHexToBytes` via `solana/index.ts:16`) validated only length. `parseInt("zz",16)` → `NaN` → written into a `Uint8Array` as `0`, so malformed hex silently produced zero bytes. Fix: reject any non-`[0-9a-fA-F]` input before parsing.

Both defects confirmed real on `origin/feature/solana`; the fixes address them.

### Verification of correctness

- **Idempotency fix works.** On first set, `logger`/`auth` are cloned via spread, which preserves callback and value references; `moduleVersions` via `cloneModuleVersions`. On a repeated call `loggersAreEqual` compares `debug`/`warn`/`error` identities (all three fields of `Logger`), `authConfigsAreEqual` compares discriminant + fields, `moduleVersionsAreEqual` handles `'auto'`/`undefined`/object. `runtimeConfigsAreEqual` covers **all 7** fields of `FhevmRuntimeConfig` (`coreFhevmRuntime.ts`) — a genuine improvement over the base guard.
- **Undefined/missing handling is sound.** `loggersAreEqual(undefined, defined)` → `false`; `authConfigsAreEqual` returns `false` when exactly one side is `undefined`. You cannot "unset" logger/auth on a second call.
- **Security of lenient comparison: not exploitable.** The comparison only gates whether a second call throws; the first frozen config always wins and is never overwritten. Worst case of an over-lenient match is a silently-ignored second call, never config injection.
- **`authConfigsAreEqual` switch compiles and lints.** `AuthType` is a closed 3-member union, all covered — satisfies `@typescript-eslint/switch-exhaustiveness-check: error` (`eslint.config.js:178`).
- **Hex fix is correct and tests match behavior.** Odd-length is checked before the char regex, so `'0x123'` → odd-length error (as tested); `'0xzz'`/`'0y00'` → invalid-char error; mixed-case and `0x`/`0X` prefixes accepted. Regex-then-`parseInt` guarantees no `NaN`.
- **No API drift inside the repo.** The hardened `hexToBytes` in `proof.ts` has **no internal callers** — only tests + the public re-export. The widely-used lenient `hexToBytes` in `core/base/bytes.ts` is a *different* function (its own tests at `bytes.test.ts:692` still assert `hexToBytes('zz')` → `[0]`) and is untouched.
- **Scope is clean.** Only the two source files + their tests changed.

### (b) BLOCKING findings

**None.** No concrete failure scenario could be constructed. The two fixes are correct, field-complete, and well-scoped; the tests are meaningful (they pass the same callback references across calls, which is exactly the regression the base code broke).

### (c) OPTIONAL improvements

1. **Missed generalization — the identical logger bug is live in the ethers and viem adapters.** `sdk/js-sdk/src/ethers/internal/config.ts:28` and `sdk/js-sdk/src/viem/internal/config.ts:29` both clone the logger on first set and then compare `stored.logger !== config.logger` on repeat — the exact defect fixed here. Additionally, both omit `auth` from their equality check entirely, so they silently accept a *changed* `auth` on a second call. Since `core/runtimeConfig-p.ts` already hosts the shared `cloneModuleVersions`/`moduleVersionsAreEqual`, the natural home for `loggersAreEqual`/`authConfigsAreEqual`/`runtimeConfigsAreEqual` is that module, fixing all three adapters. Legitimately out of scope for an issue-1709 Solana PR — a follow-up, not a blocker — but the three adapters now diverge in behavior.
2. **Over-strict auth default-header equality.** `AuthApiKeyHeader.header` defaults to `x-api-key`, so calling once with `{type:'ApiKeyHeader', value:'k'}` and again with `{type:'ApiKeyHeader', header:'x-api-key', value:'k'}` — semantically identical — throws on the second call. Minor; arguably acceptable under a strict "identical parameters" reading.
3. **Test gaps (non-blocking).** Not exercised: `BearerToken`/`ApiKeyCookie` branches of `authConfigsAreEqual`; `moduleVersions` mismatch; `wasmAssetLoadMode` mismatch (the fields the base guard silently ignored). `proof.test.ts` doesn't cover empty-string / bare `0x` (both yield an empty array — likely intended, but undocumented).

### (d) Verdict

**no-actionable-findings** (for the PR as scoped). The only substantive observation — the same latent logger bug and an auth-comparison gap in the ethers/viem adapters — is real but outside the issue's declared scope and best handled as a follow-up that lifts the equality helpers into `core/runtimeConfig-p.ts`.

PR #3150, opened 2026-07-13
fix(kms-connector): preserve empty Solana ACL domain semantics

Eikix:
## Summary

- preserve Gateway/EVM permissive semantics when the signed Solana ACL-domain list is empty
- keep non-empty domain membership enforcement unchanged for current and historical authorization
- document the empty-list behavior in the connector and SDK chain configuration

## Validation

- `cargo test -p kms-worker solana_encrypted_value_acl::tests` (9 passed)
- `cargo clippy -p kms-worker --lib --tests -- -D warnings`
- `cargo fmt -p kms-worker --check`
- `git diff --check`

Refs zama-ai/fhevm-internal#1707.

### DIFF
--- a/kms-connector/crates/kms-worker/src/core/solana_encrypted_value_acl.rs
+++ b/kms-connector/crates/kms-worker/src/core/solana_encrypted_value_acl.rs
@@ -100,8 +100,9 @@ impl SolanaAclVerifier {
         Ok(())
     }
 
-    /// Current decrypt: live handle + membership, within the request's domain scope. Reads the
-    /// account fetched at `confirmed` commitment — never a snapshot.
+    /// Current decrypt: live handle + membership, within the request's domain scope. An empty
+    /// domain scope is permissive. Reads the account fetched at `confirmed` commitment — never a
+    /// snapshot.
     pub fn verify_current_user_decrypt(
         &self,
         account_key: SolanaPubkeyBytes,
@@ -113,7 +114,9 @@ impl SolanaAclVerifier {
     ) -> Result<(), SolanaAclVerificationError> {
         let acl = &decoded.acl;
         self.verify_canonical(account_key, owner, acl)?;
-        if !allowed_acl_domain_keys.contains(&acl.acl_domain_key) {
+        if !allowed_acl_domain_keys.is_empty()
+            && !allowed_acl_domain_keys.contains(&acl.acl_domain_key)
+        {
             return Err(SolanaAclVerificationError::DomainNotAllowed);
         }
         authorize_current(acl, handle, subject).map_err(map_acl_error)?;
@@ -121,7 +124,8 @@ impl SolanaAclVerifier {
     }
 
     /// Historical decrypt: a valid historical-access MMR proof against the live confirmed peaks
-    /// (the account passed in, read fresh — not a cached/snapshotted proof-time state).
+    /// (the account passed in, read fresh — not a cached/snapshotted proof-time state). An empty
+    /// domain scope is permissive.
     pub fn verify_historical_user_decrypt(
         &self,
         target: EncryptedValueTarget<'_>,
@@ -130,7 +134,9 @@ impl SolanaAclVerifier {
         proof: &MmrProof,
     ) -> Result<(), SolanaAclVerificationError> {
         self.verify_canonical(target.account_key, target.owner, target.acl)?;
-        if !allowed_acl_domain_keys.contains(&target.acl.acl_domain_key) {
+        if !allowed_acl_domain_keys.is_empty()
+            && !allowed_acl_domain_keys.contains(&target.acl.acl_domain_key)
+        {
             return Err(SolanaAclVerificationError::DomainNotAllowed);
         }
         authorize_historical(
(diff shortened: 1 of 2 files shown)


Eikix:
## Adversarial review (agent-generated, dispatched by Elias) — head `62208d6`

### (a) Independent intended-flow statement — with signed-payload binding traced

**Intended change:** On Solana, an empty signed `allowedAclDomainKeys` list must mean "accept any ACL domain" (permissive), matching Gateway/EVM semantics, while all other authorization (owner, canonical PDA, bump, ed25519 signature, chain routing, subject membership, handle match, MMR proof) stays intact. Before this PR, `verify_current_user_decrypt` / `verify_historical_user_decrypt` did an unconditional `!allowed_acl_domain_keys.contains(&acl.acl_domain_key)` → an empty list rejected everything.

**Binding trace (the critical adversarial question):** The empty list is **not independently attacker-controllable.** Chain of custody:

1. `payload.allowedAclDomainKeys` → `allowed_acl_domain_keys` in `verify_solana_user_decrypt_signature` (`solana_user_decrypt.rs:129`).
2. That exact Vec is committed into the ed25519 signing preimage via `solana_user_decrypt_signing_preimage` (`utils/src/types/solana_extra_data.rs:132-135`), **length-prefixed** (`(len as u32).to_be_bytes()` then each 32-byte key). Empty encodes unambiguously as `0x00000000`, so empty vs non-empty can never collide.
3. `identity_key.verify(&preimage, signature)` (`solana_user_decrypt.rs:154`) checks the signature against `payload.userIdentity` — the requester's own Solana pubkey.
4. Only on success is `VerifiedSolanaAuth { allowed_acl_domain_keys, .. }` returned; `dispatch_solana_current`/`dispatch_solana_mmr_proof` pass the same signed bytes to the verifier (`solana_user_decrypt.rs:238, 272`).

So a relayer/MITM stripping the domain list to empty invalidates the ed25519 signature and is rejected *before* the verifier runs. Permissiveness is a choice bound to the identity's private key — exactly analogous to EVM's EIP-712-signed `allowedContracts`. Broadening the domain scope only broadens the requester's *own* reach; `authorize_current`/`authorize_historical` still require `identity ∈ acl.subjects` and the correct handle, so a non-member gains nothing.

**EVM/gateway parity (verified, not taken on faith):**
- EVM: `inner_allowed_contracts_check_for_user_decryption_v2` (`decryption.rs:490-493`) returns `Ok(())` when `allowed_contracts.is_empty()`; per-handle ownership still runs. Semantic: empty = skip the *scoping refinement*, not the ownership check.
- Gateway: `Decryption.sol:670` ("Empty `allowedContracts` is valid (permissive mode)") and `:743-745` for Solana ("an empty domain-key list is valid (permissive mode); only the upper bound is enforced").
- The Solana fix mirrors this precisely: empty skips only the domain filter; `authorize_*` (subject + handle + MMR) still runs.

**Both paths covered:** current (`verify_current_user_decrypt`, line 116-120) and historical (`verify_historical_user_decrypt`, line 136-140) both get the identical `!is_empty() && !contains(...)` guard. `verify_public_decrypt_exact` has no domain filter (public decrypt is not user/domain-scoped) — correctly untouched.

### (b) BLOCKING findings

**None.**

- No weakening: `verify_canonical` (owner + PDA + bump) runs first, unconditionally, unchanged. When the list is non-empty, behavior is byte-identical to before (`&&` short-circuit). When empty, only the domain filter is skipped; `authorize_current`/`authorize_historical` still run in every case. Signature/chain checks live upstream and are untouched.
- Tests are comprehensive and adversarial for both paths (`solana_encrypted_value_acl.rs`):
  - Current: empty accepted (`OWNER, &[]` ok); empty does **not** bypass subject auth (`STRANGER, &[]` → `EncryptedValueSubjectMissing`); non-empty wrong domain rejected (`&[[9;32]]` → `DomainNotAllowed`); canonical check still fires under empty (`wrong_account, &[]` → `NonCanonicalEncryptedValueAcl`).
  - Historical (`post_rotation_then_historical_proof`): empty accepted; non-empty wrong domain rejected; empty does **not** bypass the MMR proof's sealed subject (`STRANGER, &[]` → `HistoricalAccessProofInvalid`) or handle (`h(99), &[]` → `HistoricalAccessProofInvalid`).

### (c) OPTIONAL improvements

1. Defense-in-depth SDK parity: EVM warns on permissive mode (`allowedContracts=[]`) combined with a long validity window (`SignedDecryptionPermitV2-p.ts:204`, `MAX_DURATION_SECONDS`). No equivalent warning exists for empty Solana `domainKeys` + long `durationSeconds`. Not required by this PR's scope, but would complete the parity story.
2. Minor test-coverage note (not a gap): the historical test replaced the old non-empty `STRANGER, &[DOMAIN]` rejection with `STRANGER, &[]`. Non-empty wrong-domain rejection is still covered by the new `&[[9;32]]` case — no net loss.

### (d) Verdict

**no-actionable-findings.** The fix is minimal, correct, and security-sound: the empty list is bound under the requester's ed25519 signature (length-prefixed in the preimage, so unforgeable and unambiguous), semantically equivalent to EVM/gateway permissive mode, applied to both current and historical paths, with no weakening of owner/PDA/signature/subject/handle/MMR checks, and with adversarial tests proving permissive mode does not bypass subject or handle authorization on either path.

Key files: `kms-connector/crates/kms-worker/src/core/solana_encrypted_value_acl.rs` (fix + tests), `.../event_processor/solana_user_decrypt.rs` (signature binding + dispatch), `kms-connector/crates/utils/src/types/solana_extra_data.rs:95` (length-prefixed preimage), `.../event_processor/decryption.rs:490` (EVM parity reference), `gateway-contracts/contracts/Decryption.sol:670,743` (gateway permissive-mode semantics).

PR #3151, opened 2026-07-13
refactor(host-contracts): remove inert Solana authorities

Eikix:
## What

- remove the inert `input_verifier_authority` and `material_authority` fields from `HostConfig`;
- remove them from initialization arguments/events, clients, fixtures, generated IDLs, and ABI goldens;
- shrink `HostConfig::SPACE` from 215 to 151 bytes.

## Why

Repository-wide tracing found no production instruction that reads either field or names an operation protected by them. Retaining them implied authorization boundaries that did not exist and forced every initializer and fixture to supply meaningless pubkeys.

This is a focused breaking pre-release cleanup from zama-ai/fhevm-internal#1693. It does not implement or change governance, pausing, KMS rotation, input-attestation verification, material scheduling, HCU identity, or the high-bit chain-ID follow-up.

## Validation

- `bash scripts/check-zama-host-idl.sh`
- `cargo test -p zama-host` — 77 passed
- focused Mollusk HostConfig initialization test — passed
- PoC live-client `cargo check` — passed
- Solana workspace formatting — passed
- repository pre-commit coprocessor check and CI-aligned Clippy — passed
- repository-wide stale-reference scan and `git diff --check` — clean


Eikix:
## Adversarial review (agent-generated, dispatched by Elias) — head `7f92fa5`

### (a) Independent reader-trace conclusion: NO protected operation — deletion justified

- Whole-repo grep for `input_verifier_authority` / `material_authority` on `7f92fa5` returns **zero** hits (on base they existed only in the files this PR edits).
- On base, the only code touching the fields was `initialize_host_config.rs`: write-at-init, emit-in-event, and a `require!(… != Pubkey::default())`. **No `has_one` constraint, no instruction reads them, no state transition gates on them.**
- Offset/indirect-consumer trace (the primary risk of deleting mid-struct fields): the **only** off-chain decoder of `HostConfig` is `coprocessor/fhevm-engine/host-listener/src/solana_reconstruct.rs:296` `parse_host_config`, which calls `zama_host::state::HostConfig::try_deserialize` via a **path dependency on the on-chain crate** (`host-listener/Cargo.toml:84`) and reads only `chain_id`. It tracks the struct definition, so no fixed-offset breakage is possible. The KMS connector (`kms-worker/src/core/solana_acl.rs`) only derives the PDA **address** from the seed — it never decodes account data. Relayer references (`relayer/src/solana_proof/*`) are instruction **account-ordering**, independent of the struct's data fields. No hand-rolled off-chain `HostConfig` field mirror exists anywhere.

### (b) BLOCKING findings: NONE

Synchronization verified complete and internally consistent:

- **SPACE**: `151` is arithmetically correct (`32+8+8+20+20+20+8+1+1+8+8+8+8+1`); unit test updated `215 → 151` (−64) and asserts both.
- **ABI golden**: the reviewer independently re-implemented `check_solana_abi.py`'s hashing/fixture logic and recomputed all four affected entries from the head vendored IDL — every `schema_hash` and `fixture`/`fixture_len` matches the checked-in golden exactly (HostConfig 159, HostConfigInitializedEvent 81, InitializeHostConfigArgs type 77, initialize_host_config instr 85). The apparent anomaly (instruction `schema_hash` `e09039…` unchanged while its fixture shrank) is **correct**: the instruction schema references `InitializeHostConfigArgs` by name, so only the type's hash changes while the serialized fixture correctly shrinks.
- Both vendored IDLs cleaned (`zama_host.json` all 3 struct defs; `confidential_token.json`), events, init args (`state/mod.rs`), all four Mollusk/PoC fixtures, `tfhe-worker/.../solana_poc.rs`, and **both** live-client call sites (`main.rs:461`, `:1687`) updated. Zero leftover references.
- Scope is tight — no governance/pauser/KMS parity added (matching the issue's explicit non-goal).

### (c) OPTIONAL improvements (non-blocking)

- Dead error variant `InputVerifierMismatch` (`solana/programs/zama-host/src/errors.rs:29` + IDL `zama_host.json:1677`) remains. Confirmed **already dead on base** — never thrown even while the field existed — so not a regression, and leaving it is ABI-safer than removing it (removal would shift subsequent Anchor error codes). Same situation for `confidential-token`'s `MaterialCommitmentMismatch`.
- Doc/comment mentions of "material authority" (`solana/docs/DESIGN_DECISIONS.md:186/1100`, `host_mollusk.rs:16`) are accurate superseded/historical narrative — line 1100 is the very decision this PR implements. No action needed.

### (d) Verdict: **no-actionable-findings**

The reader trace independently confirms both fields protect no operation; the field removal is layout-safe because the single off-chain decoder reuses the on-chain type; and SPACE, IDLs, events, fixtures, and ABI goldens are all provably consistent (recomputed, not just eyeballed). The only optional item is a pre-existing dead error variant that predates this PR.

PR #3152, opened 2026-07-13
fix(listener): reject malformed Solana accounts

Eikix:
## Summary

- validate every static and loaded Solana account key before transaction reconstruction
- reject invalid program and instruction account indexes instead of substituting zero keys or compacting account positions
- cover malformed keys/indexes and valid top-level/inner instruction resolution

This is the first correctness slice of zama-ai/fhevm-internal#1686. It intentionally leaves transaction-signature validation, inner-instruction group validation, transport extraction, and provider retry policy unchanged.

## Validation

- `NO_DNA=1 SQLX_OFFLINE=true cargo test -p host-listener --features solana-grpc,solana-reconstruct solana_grpc_listener:: -- --nocapture`
- `NO_DNA=1 cargo fmt -p host-listener -- --check`
- `NO_DNA=1 SQLX_OFFLINE=true cargo clippy -p host-listener --features solana-grpc,solana-reconstruct --all-targets -- -D warnings`
- repository pre-commit checks (format, workspace check, CI-aligned clippy)

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/src/solana_grpc_listener.rs
+++ b/coprocessor/fhevm-engine/host-listener/src/solana_grpc_listener.rs
@@ -246,6 +246,52 @@ fn fhe_eval_durable_encrypted_value(
         .copied()
 }
 
+fn validated_account_keys<'a>(
+    keys: impl IntoIterator<Item = &'a Vec<u8>>,
+) -> Result<Vec<[u8; 32]>> {
+    keys.into_iter()
+        .enumerate()
+        .map(|(index, key)| {
+            <[u8; 32]>::try_from(key.as_slice()).map_err(|_| {
+                anyhow!(
+                    "account key {index} has invalid length {}, expected 32 bytes",
+                    key.len()
+                )
+            })
+        })
+        .collect()
+}
+
+fn resolve_instruction(
+    account_keys: &[[u8; 32]],
+    top_level_index: u32,
+    is_inner: bool,
+    program_id_index: u32,
+    data: &[u8],
+    account_indices: &[u8],
+) -> Result<crate::solana_reconstruct::DecodedInstruction> {
+    let program =
+        account_keys.get(program_id_index as usize).ok_or_else(|| {
+            anyhow!("program_id_index {program_id_index} out of range")
+        })?;
+    let accounts = account_indices
+        .iter()
+        .map(|&index| {
+            account_keys.get(index as usize).copied().ok_or_else(|| {
+                anyhow!("instruction account index {index} out of range")
+            })
+        })
+        .collect::<Result<Vec<_>>>()?;
+
+    Ok(crate::solana_reconstruct::DecodedInstruction {
+        program: bs58::encode(program).into_string(),
+        data: data.to_vec(),
+        accounts,
+        top_level_index,
+        is_inner,
+    })
+}
+
 #[allow(clippy::too_many_arguments)]
 async fn subscribe_loop(
     db: &Database,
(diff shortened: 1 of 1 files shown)


Eikix:
## Adversarial review (agent-generated, dispatched by Elias) — reviewed at `4c5db2a`, identical tree at current head `c650f55` (message-only amend for commitlint scope)

### (a) Independent intended-flow statement + rejection-disposition analysis

First correctness slice of fhevm-internal#1686. In the Yellowstone gRPC reconstruction path (`ingest_transaction`), the base branch had four places that turned raw provider bytes into keys/positions *infallibly*, substituting sentinels on bad data:

1. `account_keys_bytes`: `<[u8;32]>::try_from(k.as_slice()).unwrap_or([0u8; 32])` — truncated pubkey silently became the **zero key** (base line 542).
2. `decode` closure program id: `.get(program_id_index as usize).cloned().unwrap_or_default()` — out-of-range program index silently became the **empty string**.
3. `resolve` closure: out-of-range account index was silently **dropped, compacting/shifting the remaining positional accounts**.
4. A dead `inner: Vec<(String, Vec<u8>)>` block (base 518-532) built but never consumed.

The PR replaces all of this with two fallible helpers — `validated_account_keys` (rejects any key whose length ≠ 32) and `resolve_instruction` (rejects out-of-range program-id and account indexes), deletes the dead block, and collapses the duplicated base58 list + byte list into a single validated `Vec<[u8;32]>` (removing a latent drift risk). All error paths map to `IngestFailure::permanent`.

**Rejection disposition (the key question).** `IngestFailure::permanent` → `CursorDecision::Advance` → the caller error-logs `"permanently failed to ingest gRPC transaction; skipping"` and advances past the tx. Analysis:

- **The state-divergence scenario requires a dishonest/buggy feed.** The subscription filters `vote: Some(false), failed: Some(false)` — only successfully executed txs touching zama-host. For any tx a real validator executed, pubkeys are fixed 32 bytes and all indices are in range *by construction*. **A consensus-valid tx from an honest feed can never hit these rejection paths.** They defend against a buggy/malicious Yellowstone source or a proto-decoding bug.
- **Under a compromised/buggy feed**, the drop is real and permanent, but (i) it is error-logged, not silent, and (ii) KMS independently revalidates owner/PDA/domain/signature/MMR against confirmed live accounts before plaintext release. A dropped tx is therefore an **availability** problem (ciphertext material not eagerly prepared), not an **authorization bypass**.
- **Disposition is internally consistent** with the established taxonomy: the base already classifies every raw-feed-data defect as `permanent`/advance (`tx has no meta`, `no transaction`, `no message`, `invalid block_time`), while genuine reconstruction *logic* errors halt via `IngestFailure::fatal`. This PR sits on the correct side of the "our bug halts / their bad data skips" split.

### (b) BLOCKING findings

None. All four infallible/sentinel paths are now fallible, the disposition is consistent with sibling handling, tests cover the three rejection classes plus valid top-level/inner resolution with order preservation, and the diff is scoped to exactly one file with no half-built fixture framework (the issue's shared-golden-fixtures ambition is correctly not started here).

### (c) OPTIONAL improvements

1. **Surviving zero-key substitution one layer down — the same anti-pattern #1686 names.** `reconstruct_events_for_insert` still does `ix.accounts.get(COMPUTE_SUBJECT_INDEX).copied().unwrap_or([0u8; 32])` at `solana_grpc_listener.rs:783-785`. If a matched `fhe_eval` instruction has a truncated account list, the compute *subject* silently becomes the zero key → wrong handle derivation / wrong material request, rather than a rejection. Likely intentionally out of this slice's scope, but should be an explicit follow-up making it fallible too. (Adjacent: line 780 `.expect("covered fhe_eval requires reconstruction context")` is a hard panic in production reconstruction code; pre-existing, worth a note.)
2. **Disposition rationale is undocumented.** An argument exists for `retryable` (a transient transport glitch may deliver clean bytes on re-fetch) vs `permanent` (poison-pill stall protection + KMS revalidation boundary). `permanent` is defensible; a one-line comment on *why* malformed-provider-data is `permanent` and not `retryable`/`fatal` would prevent future re-litigation.
3. **Test gap: loaded-addresses (ALT) boundary.** No test that an account/program index pointing into the ALT region resolves correctly, nor that the `static ++ loaded_writable ++ loaded_readonly` concatenation order is preserved end-to-end — the subtle correctness-critical property of this path.
4. **Minor style:** the `all_instructions` block's trailing `std::result::Result::<_, IngestFailure>::Ok(ordered)` + outer `?` is redundant (it's a bare block, not a closure); could just evaluate to `ordered`.

### (d) Verdict

**no-actionable-findings** (no blocking issues). The slice does exactly what #1686's first bullet asks — fallible pubkey conversion and fallible program/account-index resolution, rejecting the whole tx instead of substituting zero keys / empty program ids / shifted positions — with a rejection disposition consistent with the listener's existing raw-feed-defect handling that does not drop consensus-valid txs under an honest feed. The optional items (esp. the surviving `unwrap_or([0u8;32])` at line 783-785 and the missing disposition rationale) are worth logging as explicit follow-ups but do not block this slice.

PR #3153, opened 2026-07-14
feat(sdk): add e2e load test tool

zmalatrax:
## fhevm-sdk CLI + load-test tool

Adds `sdk/cli-js-sdk`, a pnpm workspace providing a command-line tool for exercising `@fhevm/sdk` flows against arbitrary contracts, the FHETest contract, plus ERC-7984 confidential token utilities. Also adds a load-test runner for exercising the protocol paths at scale. Pins `@fhevm/sdk` to `0.13.2-1` (the v13 release line).

### Packages

- `packages/toolkit` — importable library wrapping `@fhevm/sdk` (viem): network/client config, encryption, decrypt flows, offline verification of saved user-decrypt results, FHETest and ERC-7984 helpers. No CLI dependencies.
- `packages/cli` — the `fhevm-sdk` commander CLI, consuming the toolkit.
- `packages/load-test` — the `load-test` CLI: plans and executes batched runs across the protocol's cost units (local CPU proof generation on worker threads, on-chain writes, relayer-served public/user decrypts), with preflight checks, handle pooling, interrupt-safe execution, and machine-readable reports. See `packages/load-test/README.md` and `docs/OPERATIONS.md`.

### Commands

- `input-proof` — encrypt values and request a relayer-verified input proof
- `public-decrypt` / `user-decrypt` / `delegated-user-decrypt` — each a family of subcommands: `direct` decrypts any existing handle (`--handle`, `--contract` for ACL pairing), `fresh`/`stored` run FHETest demo flows (plus `make-public` for public-decrypt)
- `verify-user-decrypt` — re-verifies a relayer user-decrypt response offline from a saved validation artifact (`--artifact`, written by any user-decrypt flow). The relayer GET URL is derived from the artifact (network, flow, job id); `--job-id`/`--url` override. Shares are KMS-signature-verified against on-chain signers and decrypted with the saved transport key; plaintexts are compared when the artifact records expected values.
- `fhe-test` — FHETest info, inspection, handle initialization, and on-chain FHE operation demos
- `token` — ERC-7984 confidential transfers (`--verify` decrypts sender balance before/after to confirm the transfer) and balance handle reads
- `completion` — shell tab completion (bash, zsh, fish, pwsh)

Supports `testnet` (Sepolia, default) `testnet-amoy` (Polygon Amoy), `devnet`, `devnet-amoy` (Polygon Amoy), and `mainnet`, with RPC/relayer URL overrides and `ZAMA_FHEVM_API_KEY` relayer auth. Progress logs go to stderr, results to stdout as JSON, so commands are pipeable. See `sdk/cli-js-sdk/README.md` for full usage.

### Notes

- Commands with subcommands declare no options of their own (direct modes live under `direct`), so flags always bind to the subcommand that uses them.
- User-decrypt validation artifacts (schema v2) contain the transport private key — treat them as key material. Offline verification is pinned fail-closed to the exact `@fhevm/sdk` version it was built against.


0xalexbel (approved):
LGTM!

PR #3160, opened 2026-07-14
feat(test-suite): make e2e SDK selection explicit

Eikix:
## Summary

Makes the E2E SDK family, source, and exact version explicit, and routes normal and multi-chain tests through one test-suite-owned adapter factory.
The harness can reproduce the incident clients (`relayer-sdk@0.4.2`, `relayer-sdk@0.4.4`, and `fhevm-sdk@1.1.0-alpha.4`) without changing SDK or KMS connector source or claiming an SDK support policy.
Build-time selection validation, package-version checks, and a runtime realpath guard prevent a test image from silently claiming npm provenance while executing the vendored workspace SDK; same-version npm/workspace ambiguity fails closed.
Validation passed for the CLI checks, 260 unit tests, compatibility smoke, seven selector tests, and adapter compilation; immutable SDK-specific images and digest receipts remain follow-up work for [fhevm-internal#1712](https://github.com/zama-ai/fhevm-internal/issues/1712).

### DIFF
--- a/test-suite/e2e/Dockerfile
+++ b/test-suite/e2e/Dockerfile
@@ -1,9 +1,15 @@
 ARG NODE_VERSION=22.22.2-alpine3.22
 ARG RELAYER_SDK_VERSION=""
+ARG E2E_SDK_FAMILY=""
+ARG E2E_SDK_SOURCE=""
+ARG E2E_SDK_VERSION=""
 
 # --- Builder Stage ---
 FROM node:${NODE_VERSION} AS builder
 ARG RELAYER_SDK_VERSION=""
+ARG E2E_SDK_FAMILY
+ARG E2E_SDK_SOURCE
+ARG E2E_SDK_VERSION
 
 WORKDIR /app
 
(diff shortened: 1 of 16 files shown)


Eikix:
🤖 Automated review — HEAD d7f139b

**TL;DR:** Solid, well-scoped change. The selection contract (`selection.ts`) is cleanly validated and unit-tested, the legacy `RELAYER_SDK_VERSION` path is preserved with an explicit warning, and the new preflight composes correctly with the existing rollout runbooks (both already pass `--override test-suite`, and `upgradeRuntimeGroup` reuses persisted overrides, so mid-rollout phase locks don't trip it). No blocking issues found.

### Should-fix

1. **npm workspace shadowing can silently defeat `E2E_SDK_SOURCE=npm` for `fhevm-sdk`** — `test-suite/e2e/Dockerfile` (npm-pin branch) + `test-suite/e2e/test/sdk/factory.ts:258`. The e2e workspace depends on `@fhevm/sdk: "*"`, and `sdk/js-sdk/src` is a root workspace named `@fhevm/sdk` (currently `1.1.0-alpha.7`). If a caller requests `E2E_SDK_FAMILY=fhevm-sdk E2E_SDK_SOURCE=npm E2E_SDK_VERSION=<version equal to the workspace version>`, `npm install` satisfies the pin with the workspace symlink instead of the registry tarball. The runtime guard in `factory.ts` compares versions only, so it passes and the image claims npm provenance while actually running workspace code. A cheap hardening: when `selection.source === 'npm'`, also assert `fs.realpathSync(require.resolve(...))` does not point inside the repo (`/app/sdk/js-sdk`). The PR body defers full provenance to follow-up work, but this specific failure is silent and the check is one line.

2. **Dockerfile partial-selection errors don't match the runtime contract** — `test-suite/e2e/Dockerfile:32-51`. The build-time triage keys off `[ -z "$sdk_family$sdk_source$sdk_version" ]`, so a partial selection (e.g. only `E2E_SDK_FAMILY=relayer-sdk`, or a typo'd `E2E_SDK_SOURCE`) falls through to the `workspace selection requires fhevm-sdk@workspace` error (or, for `family=fhevm-sdk`/`version=workspace` with a bad source, builds a workspace image that then throws at runtime). The precise diagnostics live only in `selection.ts`, which runs after the image is built. Consider validating the triple in the RUN step with messages mirroring `selectSdk` (or reusing it via a small node invocation) so misconfigurations fail at build with the right message.

### Nits

- `test-suite/e2e/Dockerfile` runs `npm run tsc:sdk-selection` (good — typechecks adapters against the actually-installed package) but not `npm run test:sdk-selection`; the selection unit tests appear unwired in CI in this diff. They're fast — worth running in the builder stage or a workflow.
- `test-suite/e2e/test/sdk/factory.ts:214` — `selectSdk(process.env)` and `packageIdentity` execute (and can throw / print the legacy warning) at module import time. Fine for the test image, but it means any local hardhat import of `instance.ts` without env prints the legacy warning; intentional per the design, just noting the side-effectful import.

### Context

- Adapter compat shims are reasonable: dual `durationDays`/`durationSeconds` for `signDecryptionPermit` across fhevm-sdk versions, and positional `extraData` gated on a `getExtraData` capability probe for relayer-sdk 0.4.x vs newer — matches how the two published API generations differ.
- Rollout locks (`rollouts/v0.12-to-v0.13/versions.ts`, `rollouts/v0.13.0-testnet/versions.ts`) switch to the explicit triple; `rollout-v013-testnet.test.ts` and the new `stack.test.ts` case cover the pinning and the preflight respectively. `compose.ts` keeps `RELAYER_SDK_VERSION` as a defaulted build arg for the legacy interface, so older invocations still build.

Eikix:
🤖 Automated review — HEAD c658be7 (follow-up since d7f139b)

**TL;DR:** LGTM. The single new commit (`fix(test-suite): fail closed on SDK provenance`) directly and correctly addresses everything flagged in the previous review. No new blocking or should-fix issues found in the delta.

### Prior findings — all addressed

1. **npm workspace shadowing of `@fhevm/sdk` (should-fix)** — Fixed in `test-suite/e2e/test/sdk/factory.ts:65-79`. The new guard realpaths the resolved entrypoint and throws if `family=fhevm-sdk source=npm` resolves inside the vendored `sdk/js-sdk` tree. Since the workspace is `sdk/js-sdk/src`, prefix-matching on the realpathed parent dir (`sdk/js-sdk` + `path.sep`) covers it, and a genuinely registry-installed copy nests under `test-suite/e2e/node_modules`, so no false positives. This is exactly the fail-closed check suggested.
2. **Dockerfile partial/typo'd selection errors (should-fix)** — Fixed via the new `validate:sdk-selection` script (`test-suite/e2e/package.json`) run in the builder before the shell triage (`test-suite/e2e/Dockerfile:44`). It calls `selectSdk(process.env)` directly, so a partial triple or a typo'd `E2E_SDK_SOURCE` now fails the build with the precise `selection.ts` message instead of the misleading workspace-branch error. ARG defaults of `""` are correctly treated as unset by `selectSdk`. Reusing `selectSdk` rather than duplicating the rules in shell is the right call.
3. **Selection unit tests not wired into the image build (nit)** — Fixed: `test-suite/e2e/Dockerfile:70` now runs `npm run test:sdk-selection` before the typecheck.
4. **Side-effectful module-import selection (nit)** — Unchanged; previously noted as intentional per the design, no action expected.

The new negative test in `test-suite/e2e/test/sdk/selection.test.ts:31-39` (valid family + version but `source=registry`) pins the typo'd-source diagnostic — good coverage of the exact failure mode from finding 2.

### Nit (new, non-blocking)

- `test-suite/e2e/test/sdk/factory.ts:68` — `fs.realpathSync(path.resolve(__dirname, '../../../../sdk/js-sdk'))` is computed unconditionally at import, including for relayer-sdk selections; if the vendored dir is ever absent (e.g. the e2e workspace consumed outside the full repo/image layout), this throws ENOENT before any selection-relevant check. In the supported Docker/repo layouts the dir always exists, so this is theoretical — could be guarded with `fs.existsSync` or moved inside the `fhevm-sdk && npm` branch if that layout ever changes.

### Verdict

Follow-up delta is a focused, correct response to the prior review; the overall change (reviewed at d7f139b) remains sound. Good to merge from this reviewer's perspective.

PR #3161, opened 2026-07-14
feat(test-suite): plan progressive KMS node upgrades

Eikix:
## Summary

Adds a deterministic progressive KMS upgrade planner built from the resolved threshold serving committee, with canonical node identities, a derived `2t+1` reconstruction quorum, explicit upgrade order, and minimal quorums to test after selected steps.
The standard path tests every step through half, every partial step between the not-yet-upgraded nodes losing a standalone quorum and the upgraded nodes gaining one, and the final partial step.
Completed upgrade operations are tracked independently from planned node versions; observed runtime versions belong to executor receipts, and provisioned spares outside the serving committee are excluded.
Validation passed in the 261-test `test-suite/fhevm` suite, type-checking, and the compatibility smoke test; exact node isolation, restoration, receipts, and larger response-set coverage remain follow-up work for [fhevm-internal#1714](https://github.com/zama-ai/fhevm-internal/issues/1714).


Eikix:
## Review — feat(test-suite): model progressive KMS quorum boundaries

### Overview

Adds two new files to `test-suite/fhevm/src`: `threshold-rollout.ts`, a pure in-memory model of a progressive KMS upgrade (plan validation, feasible pending/upgraded quorum compositions, "dangerous" partial-upgrade counts, and deterministic representative quorums per checkpoint), plus a test file. No runtime integration — planning scaffolding for fhevm-internal#1714.

### Correctness — verified

- The math checks out: `quorumCompositionRange` is the correct hypergeometric support (`max(0, Q−(N−u))` … `min(Q, u)`), and `dangerousPartialUpgradeCounts` produces the right boundary sets for the tested cases (N=4/Q=3 → 1..3; N=7/Q=5 → 1..6; N=5/Q=4 → 1..4) and behaves sensibly at edges (Q=1, Q=N).
- Both files were extracted and run standalone: **11/11 tests pass** under bun 1.3.6.
- Validation in `defineThresholdUpgradePlan` is thorough (duplicate ids/identities, permutation check on `upgradeOrder`, empty versions), and the defensive copy is tested.

### Main concerns

1. **Parallel abstraction next to the existing KMS model (biggest issue).** This suite already has `src/kms-party.ts` (party naming: `kmsCoreName`, quorum: `reconstructionThreshold`) and `resolveKmsTopology` in `src/scenario/resolve.ts`, which enforces the real protocol constraint `parties = 3t+1` with reconstruction quorum `2t+1`. The new module accepts any `quorum` from 1 to N and as few as 2 parties — topologies the rest of the suite (correctly) rejects — and the test hand-spells `"kms-core"`, `"kms-core-2"`… instead of calling `kmsCoreName`, which `kms-party.ts` explicitly says never to do inline. Suggestion: derive party identities via `kmsCoreName`, and either build plans from a `resolveKmsTopology` result or document why arbitrary N/Q is deliberately allowed here.

2. **`upgradeState` is positional, not version-derived — and this can misstate quorum danger.** A party already at `targetVersion` before the rollout still counts as `"pending"` (the third `thresholdUpgradeCheckpoint` test enshrines this). Since the mixed-quorum danger being modeled is about *effective versions responding*, a quorum reported as `pendingResponders: 2` could actually be version-homogeneous. If "operation step" vs. "effective version" is an intentional distinction, it deserves a doc comment on `upgradeState`; otherwise consider rejecting plans where a party already matches `targetVersion`, or deriving responder composition from versions.

3. **Nothing consumes this yet.** The module is only imported by its own test. That's stated in the PR body, so it's a scoping decision rather than a defect — but combined with point 1, there's real drift risk: by the time the executor lands, the topology/naming assumptions may have to be reconciled with `kms-party.ts` anyway. Doing that reconciliation now is cheap; later it's a migration.

### Minor

- In `dangerousPartialUpgradeCounts`, the second loop's upper bound (`≤ quorum`) overlaps with the explicit `counts.add(quorum)` and `counts.add(parties - quorum + 1)` two lines later — the set dedups it, but tightening the loop to the forced-mixed interval (`≤ quorum - 1`) would make the inline comments and code tell the same story.
- `thresholdUpgradeCheckpoint` rebuilds parties via `parties.find(...)!` inside two maps — O(N²) but N is tiny; fine as-is.
- Test descriptions are unusually good — they state the invariant, not the mechanics.

### Verdict

Well-built, well-tested pure model with correct quorum math. The ask before merge: reconcile with the existing `kms-party.ts` / `resolveKmsTopology` conventions (or explicitly document the divergence) and clarify the positional-vs-version `upgradeState` semantics — both are cheap now and expensive after the executor builds on this.

Eikix:
🤖 Automated review — HEAD 200b719

**TL;DR:** Solid, self-contained planning module with correct quorum math and thorough validation; no blocking issues. The concerns from the earlier review (parallel abstraction vs `kms-party.ts`/`resolveKmsTopology`, ambiguous `upgradeState` semantics) are addressed at this HEAD: the plan is now derived from a `ResolvedKmsTopology`, identities come from `kmsCoreName`, the quorum is the canonical `2t+1` via `reconstructionThreshold`, and `upgradeState` is documented and tested as operation-tracking independent of `version`.

### Findings

**Blocking** — none.

**Should-fix** — none. Verified independently:
- `possibleUpgradedNodesInQuorum` is the correct hypergeometric support: `max(0, Q−(N−u)) … min(Q, u)` (`test-suite/fhevm/src/threshold-rollout.ts:139-144`).
- `defineKmsUpgradePlan` re-enforces `committeeSize === 3t+1` on top of `resolveKmsTopology` (which already guarantees it, `test-suite/fhevm/src/scenario/resolve.ts:100-104`) — redundant but justified since tests hand-construct topologies via spread.
- `buildKmsUpgradeStep` quorum membership is deterministic (order-prefix from both sides of the upgrade boundary) and consistent with the declared `upgradeOrder`; the spares exclusion (`kmsPartyIds(committeeSize)`) matches the suite's convention that cores beyond `committeeSize` are spares.

**Nits**
- `threshold-rollout.ts:154-168` (`upgradeStepsToTest`): all three tests use committees where the union of "through half" + "boundary interval" + "N−1" happens to be *every* partial step (N=4/Q=3, N=7/Q=5, N=5/Q=4). The first valid committee where a gap appears is t=3 (N=10, Q=7 skips step 8). Since the gap is the documented design, a test at N=10 would pin it and stop a future reader from "fixing" it into full coverage. Also, the N=5/Q=4 case isn't a valid 3t+1/2t+1 committee — fine for a generic helper, slightly misleading in a KMS-named test.
- `upgradeStepsToTest` takes raw `(totalNodes, quorum)` and nothing ties it to a `KmsUpgradePlan`; the executor will presumably call it with `(plan.nodes.length, plan.quorum)`. A thin plan-accepting overload (or one test exercising exactly that call shape) would prevent argument-order drift when fhevm-internal#1714 lands.
- `threshold-rollout.ts:81-86`: the extra-key check coerces with `Number(key)`, so a key like `" 1"` passes the outside-committee check and later surfaces as `missing KMS node 1` — still an error, just a slightly misleading one. Not worth changing unless you're touching the function anyway.

### Context

Scoped as pure planning scaffolding for fhevm-internal#1714 — only the test imports the module, which the PR body states explicitly. Two files, no runtime integration, no compatibility surface. Follow-up executor work (isolation, restoration, receipts, larger response sets) is correctly deferred.

PR #3162, opened 2026-07-14
ci(test-suite): restore stateful rollout workflow

Eikix:
## Summary

Restores a manually dispatched stateful rollout workflow.

The workflow accepts one runbook path, installs the same runtime dependencies and registry credentials as the existing test-suite E2E workflow, runs the rollout with a bounded timeout, prints the receipt produced by `fhevm-cli`, uploads the Markdown and JSONL receipts even on failure, and always cleans the stack.

The workflow is dispatch-only. It does not infer runbooks from PR changes or run through normal pull-request validation.

The receipt artifact uses the repository's existing `actions/upload-artifact` pin, includes the hidden `.fhevm` directory explicitly, fails when evidence is missing, and uses a rerun-safe artifact name.

This is the CI entry point for [fhevm-internal#1718](https://github.com/zama-ai/fhevm-internal/issues/1718) and the isolated execution prerequisite for [fhevm-internal#1714](https://github.com/zama-ai/fhevm-internal/issues/1714).

## Relationship to #3170

[#3170](https://github.com/zama-ai/fhevm/pull/3170) is based directly on `main` and has no code dependency on this PR. This workflow must land before #3170 can satisfy its acceptance criteria: dispatch #3170's exact head with `rollouts/mainnet-v0.11-to-v0.12-kms/run.ts`, then inspect the retained receipt.

## Validation

- `zizmor --persona auditor --offline .github/workflows/test-suite-stateful-rollout.yml`
- `git diff --check`
- pre-commit workflow checks
- adversarial review of artifact paths, hidden-file handling, permissions, rerun behavior, and cleanup ordering

The rollout itself was not run locally. Because this is a new `workflow_dispatch` workflow, GitHub can register and execute it only after this PR lands on the default branch.

### DIFF
--- a/.github/workflows/test-suite-stateful-rollout.yml
+++ b/.github/workflows/test-suite-stateful-rollout.yml
@@ -1,0 +1,252 @@
+name: test-suite-stateful-rollout
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+  workflow_dispatch:
+    inputs:
+      runbook:
+        description: "TypeScript rollout runbook path"
+        required: true
+        type: string
+      test-profile:
+        description: "Rollout test depth"
+        default: "rollout-standard"
+        required: true
+        type: choice
+        options:
+          - rollout-standard
+          - rollout-heavy
+
+permissions: {}
+
+concurrency:
+  group: test-suite-stateful-rollout-${{ github.ref }}
+  # Label-only events must not cancel a rollout that is already using the same PR ref.
+  cancel-in-progress: ${{ github.event_name != 'pull_request' || github.event.action != 'labeled' }}
+
+jobs:
+  fhevm-rollout:
+    name: fhevm-rollout
+    if:
+      ${{ github.event_name != 'pull_request' || (github.event.pull_request.head.repo.full_name == github.repository &&
+      contains(github.event.pull_request.labels.*.name, 'rollout') &&
+      (github.event.action != 'labeled' || github.event.label.name == 'rollout')) }}
+    permissions:
+      contents: "read" # Required to checkout repository code.
+      packages: "read" # Required by fhevm-cli when resolving GitHub package metadata.
+    runs-on: large_ubuntu_32
+    # Repository prerequisite: configure this environment with required reviewers
+    # and environment-scoped registry secrets. Declaring it here does not enforce
+    # either protection by itself.
+    environment: stateful-rollout
+    timeout-minutes: 300
+    env:
+      GH_TOKEN: ${{ github.token }}
+      ROLLOUT_TEST_PROFILE: ${{ inputs.test-profile || 'rollout-standard' }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: "false"
+          fetch-depth: 0
+
+      - name: Setup Docker
+        uses: docker/setup-buildx-action@d70bba72b1f3fd22344832f00baa16ece964efeb # v3.3.0
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@ecf28ddc73e819eb6fa29df6b34ef8921c743461 # v2.1.3
+
+      - name: Install foundry
+        uses: foundry-rs/foundry-toolchain@82dee4ba654bd2146511f85f0d013af94670c4de # v1.4.0
+
+      - name: Resolve rollout runbook
+        env:
+          INPUT_RUNBOOK: ${{ inputs.runbook }}
+          BASE_REF: ${{ github.base_ref }}
+        run: |
+          set -euo pipefail
+          if [ -n "$INPUT_RUNBOOK" ]; then
+            runbook="$INPUT_RUNBOOK"
+            if [ ! -f "$runbook" ]; then
+              runbook="${GITHUB_WORKSPACE}/${INPUT_RUNBOOK}"
+            fi
+            if [ ! -f "$runbook" ]; then
+              echo "Runbook not found: ${INPUT_RUNBOOK}" >&2
+              exit 1
+            fi
+            echo "ROLLOUT_RUNBOOK=$runbook" >> "$GITHUB_ENV"
+            exit 0
+          fi
+
+          if [ -z "$BASE_REF" ]; then
+            echo "workflow_dispatch requires the runbook input" >&2
+            exit 1
+          fi
+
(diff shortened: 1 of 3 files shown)


Eikix:
🤖 Automated review — HEAD 5be9342

**TL;DR:** Solid, well-hardened CI workflow. Action pins, registry logins, safe-chain, and runner match the neighboring test-suite E2E workflows as claimed; the fork/label gate, template-injection-safe input handling, and unconditional diagnostics/cleanup are all correct. No blocking issues — a few should-fix/nit items below.

### Verified against the codebase

- `fhevm-cli rollout run` / `clean` subcommands exist (`test-suite/fhevm/src/cli.ts:285`, `:218`), and `check`, `test src`, `compat-smoke` are real package scripts.
- Receipt path is correct: `STATE_DIR` resolves to repo-root `.fhevm` (`test-suite/fhevm/src/layout.ts:10-11`), so `../../.fhevm/rollout/receipt.{md,jsonl}` from `test-suite/fhevm` matches `receiptDir()` in `src/commands/rollout-receipt.ts:55-58`.
- Threshold container globs match generated names: `kms-core-2..N` + `kms-core-init` (`src/generate/kms-core.ts:226-227`) and per-party `kms-connector-{i}-*` (`src/generate/compose.ts:460-461`) are all caught by the diagnostics loops, with party 1 covered by the explicit base-name snapshots.
- Runbook inference awk correctly maps any change under `test-suite/fhevm/rollouts/<dir>/…` to `<dir>/run.ts`, and both existing runbooks (`v0.12-to-v0.13`, `v0.13.0-testnet`) fit that layout.
- Injection surface: `inputs.runbook` is passed via `env` (never interpolated into the script), the `if` gate blocks fork PRs before any secret use, and the concurrency expressions only reference trusted context. `zizmor: ignore[secrets-outside-env]` is scoped and justified.
- The relative-path fallback in "Execute stateful rollout" is needed and correct: `fhevm-cli` `cd`s to its own dir, so the `GITHUB_WORKSPACE`-prefixed absolute path is what makes PR-inferred repo-root-relative paths work.

### Should-fix

1. **`ROLLOUT_TEST_PROFILE` is only consumed by one runbook** — `rollouts/v0.12-to-v0.13/run.ts` reads it, but `rollouts/v0.13.0-testnet/run.ts` does not. A manual dispatch of the testnet runbook with `test-profile: rollout-heavy` silently behaves as standard. Consider either wiring the profile into that runbook or failing fast/logging when the selected runbook ignores the profile, so a "heavy" run's green check isn't misleading.

### Nits

2. **Diagnostics skip several default coprocessor services** (`.github/workflows/test-suite-stateful-rollout.yml`, "Collect rollout diagnostics") — `coprocessor-host-listener-poller`, `coprocessor-upgrade-controller`, `coprocessor-consensus-detector`, and `coprocessor-db-migration` exist in the compose set (the first three are recent additions) but aren't snapshotted. For an upgrade-rollout workflow specifically, `upgrade-controller`/`consensus-detector` logs seem like the ones you'd most want on failure.
3. **`GITHUB_ENV` write from a dispatch input** — `echo "ROLLOUT_RUNBOOK=$runbook" >> "$GITHUB_ENV"` would let a newline-containing `runbook` input define extra env vars. Only actors with dispatch rights can trigger it, so this is theoretical; using the heredoc `GITHUB_ENV` syntax (`ROLLOUT_RUNBOOK<<EOF`) would close it cheaply.
4. **Sanitizer is pattern-based** — bare hex private keys or unlabeled credentials in container logs won't be redacted. Fine as best-effort (and better than the neighboring E2E workflows, which don't sanitize at all), just don't treat the artifacts as secret-free by construction.

### Context assessment

PR description is accurate on every claim I could check statically (pins, receipt paths, threshold coverage, gating semantics). As the description notes, the real proof point is the first actual rollout run — nothing here has executed end-to-end yet, and issues like registry scope for `GH_TOKEN` package resolution will only surface at runtime (though the pattern is identical to the working E2E workflows, so risk is low).

Eikix:
🤖 Automated review — HEAD 798d1e3 (follow-up since 5be9342)

**TL;DR:** Follow-up review of the single commit since 5be9342 (`fix(ci): harden stateful rollout dispatch`). All actionable findings from the prior review are addressed, and the fixes are correct. LGTM — one harmless nit below.

### Prior findings — addressed

1. **Should-fix: `ROLLOUT_TEST_PROFILE` ignored by the testnet runbook** — addressed. `rollouts/v0.13.0-testnet/run.ts` now calls `resolveRolloutTestProfile(process.env.ROLLOUT_TEST_PROFILE)` as the first statement of `run()`, so a `rollout-heavy` dispatch fails fast before any rollout side effect, and the resolved profile is threaded through all six `ctx.test(...)` calls. Covered by a new unit test in `src/rollout-v013-testnet.test.ts`.
2. **Nit: missing coprocessor diagnostics** — addressed and then some. The diagnostics step now snapshots `host-listener-poller`, `db-migration`, `consensus-detector`, and `upgrade-controller` per coprocessor instance, plus `relayer-db-migration`, `kms-connector-db-migration`, and dynamically named extra-host-chain listeners.
3. **Nit: `GITHUB_ENV` newline injection via `runbook` input** — addressed. `write_runbook_env` rejects any `\n`/`\r` in the path before writing, in both the dispatch and PR-inference branches.
4. **Nit: pattern-based sanitizer** — informational; unchanged, as expected.

### New findings

**Nit**
- `.github/workflows/test-suite-stateful-rollout.yml` (extra-chain listener loop, ~line 262): the regex `^coprocessor[0-9]*-host-listener(-poller)?-.+$` also matches the base `coprocessor[0-9]*-host-listener-poller` containers — the `-.+$` tail consumes `-poller` via the non-poller alternative — so those logs are snapshotted twice (once in the per-instance loop, once as "Extra Host Chain Listener"). Extra-chain suffixes are `-<chain-key>` (e.g. `coprocessor-host-listener-poller-chain-b`, per `hostChainSuffix` in `test-suite/fhevm/src/layout.ts:420`), so appending `| grep -vE '^coprocessor[0-9]*-host-listener-poller$'` (alongside the existing `-consumer` exclusion) would dedupe. Purely cosmetic — duplicate log groups in the run output, no correctness impact.

Sanity check of the full diff vs `main` found nothing new beyond the prior review; the delta is narrowly scoped to the three feedback items.

.github/workflows/test-suite-stateful-rollout.yml
PanGan21, line 1: Let's also change [this comment in readme](https://github.com/zama-ai/fhevm/blob/3ccb7ee4781b56bf97666a3a1b452ee72c92a198/test-suite/fhevm/README.md?plain=1#L214) so it is aligned with this pr as well

PR #1684, opened 2025-12-30
fix(protocol-contracts): mitigate ERC4626 inflation attack with decimal offset (L-06)

melanciani:
## Summary
- Set `_decimalsOffset` to 2 in OperatorStaking vault to mitigate the ERC4626 inflation attack
- This creates 100 virtual shares per asset unit, making the attack economically unfeasible
- Updated tests to account for the new share-to-asset conversion ratio

refs https://github.com/zama-ai/fhevm-internal/issues/826

PR #3163, opened 2026-07-14
feat(sdk): submit Solana input proofs

Eikix:
## Outcome

Add a public Solana SDK action that submits a previously built input proof, waits for the relayer result, and verifies that the returned handles exactly match the locally constructed proof.

`buildInputProof` remains pure. Network ownership is explicit through `submitInputProof`.

## Changes

- add `submitInputProof({ inputProof, options })` to the existing Solana encrypt client;
- derive canonical base58 contract and user identities from the proof’s bytes32 identities;
- reuse the SDK’s existing asynchronous relayer machinery and runtime authentication;
- reject empty proofs, proof/client chain-ID mismatches, and returned-handle cardinality, order, or value mismatches;
- document that the Solana host program verifies the returned coprocessor signatures when they are consumed;
- expose the typed result through the existing decorator and public Solana entrypoint;
- replace three duplicated launcher/shell POST-and-poll implementations;
- remove duplicate base58 identity environment plumbing.

`@scure/base@1.2.6` is promoted from a transitive dependency to a direct published SDK runtime dependency for canonical base58 encoding.

The shared relayer client serializes `contractChainId` as hexadecimal (`0x8000000000003039` for the test Solana chain), whereas the removed shell path serialized it as decimal. The previous exact head passed the Solana full vertical with the hexadecimal payload.

## Validation

- focused submit-input-proof tests: 9/9
- focused ESLint and configured Prettier
- test-suite/fhevm typecheck
- `bash -n` for both modified shell scripts
- `git diff --check`
- pre-commit hooks
- independent exact-head adversarial review; actionable findings addressed in follow-up

The repository-wide SDK TypeScript build retains two errors in untouched base files. Neither appears in the changed-file validation.

## Remaining gates

- current-head CI and relevant Solana full vertical;
- verify the pending issue-level KMS-impact classification; KMS behavior is unchanged;
- reconcile #1684’s Remaining A wording with the already-landed explicit current/historical `aclValueKey` ownership before closing the issue.

Implements the remaining input-proof submission and handle-verification portion of zama-ai/fhevm-internal#1684.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22390,7 +22390,6 @@
     },
     "sdk/js-sdk/node_modules/@scure/base": {
       "version": "1.2.6",
-      "devOptional": true,
       "license": "MIT",
       "funding": {
         "url": "https://paulmillr.com/funding/"
@@ -26113,6 +26112,7 @@
       "dependencies": {
         "@noble/curves": "2.0.1",
         "@noble/hashes": "2.0.1",
+        "@scure/base": "1.2.6",
         "wasm-feature-detect": "^1.8.0"
       },
       "engines": {
--- a/sdk/js-sdk/src/package.json
+++ b/sdk/js-sdk/src/package.json
@@ -106,6 +106,7 @@
     }
   },
   "dependencies": {
+    "@scure/base": "1.2.6",
     "@noble/hashes": "2.0.1",
     "@noble/curves": "2.0.1",
     "wasm-feature-detect": "^1.8.0"
--- a/sdk/js-sdk/src/solana/actions/submitInputProof.test.ts
+++ b/sdk/js-sdk/src/solana/actions/submitInputProof.test.ts
@@ -120,6 +120,35 @@ describe('submitInputProof', () => {
     });
   });
 
+  it('rejects an input proof without handles before contacting the relayer', async () => {
+    const inputProof = new Proxy(proof(), {
+      get(target, property, receiver) {
+        return property === 'getInputHandles' ? () => [] : Reflect.get(target, property, receiver);
+      },
+    });
+    globalThis.fetch = vi.fn();
+
+    await expect(submitInputProof(context(), { inputProof })).rejects.toThrow(
+      'Input proof must contain at least one external handle',
+    );
+    expect(globalThis.fetch).not.toHaveBeenCalled();
+  });
+
+  it('rejects an input proof for a different Solana chain before contacting the relayer', async () => {
+    const inputProof = proof();
+    const fhevm = context();
+    const mismatchedContext = {
+      ...fhevm,
+      solanaChain: { ...fhevm.solanaChain, id: CHAIN_ID + 1n },
+    };
+    globalThis.fetch = vi.fn();
+
+    await expect(submitInputProof(mismatchedContext, { inputProof })).rejects.toThrow(
+      `Input proof chain id ${CHAIN_ID} does not match Solana client chain id ${CHAIN_ID + 1n}`,
+    );
+    expect(globalThis.fetch).not.toHaveBeenCalled();
+  });
+
   it('uses the existing queued GET path until the request succeeds', async () => {
     const inputProof = proof();
     globalThis.fetch = vi
--- a/sdk/js-sdk/src/solana/actions/submitInputProof.test.ts
+++ b/sdk/js-sdk/src/solana/actions/submitInputProof.test.ts
@@ -1,0 +1,192 @@
+import type { FhevmRuntime } from '../../core/types/coreFhevmRuntime.js';
+import type { InputHandle } from '../../core/types/encryptedTypes-p.js';
+import type { EncryptionBits } from '../../core/types/fheType.js';
+import type { Bytes32Hex } from '../../core/types/primitives.js';
+import type { SolanaZkProof } from '../../core/types/zkProof-p.js';
+import { hexToBytes32 } from '../../core/base/bytes.js';
+import { fetchCoprocessorSignatures } from '../../core/modules/relayer/module/fetchCoprocessorSignatures.js';
+import { toSolanaZkProof } from '../../core/coprocessor/SolanaZkProof-p.js';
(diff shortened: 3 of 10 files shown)


Eikix:
🤖 Automated review — HEAD f78e3b9

**TL;DR:** Solid, well-scoped change. The new `submitInputProof` action correctly reuses the existing relayer async machinery, the handle-verification is real (cardinality/order/value via `assertHandleArrayEquals`), the test coverage is meaningful, and the shell-script dedup is a clear win. No blocking findings; a few should-verify/should-fix items below, mostly around parity with the EVM path and one wire-format change the PR description understates.

## Should-fix / should-verify

1. **`contractChainId` wire format changed decimal → hex — verify against the live relayer, and the PR description is inaccurate on this point.** The old launcher POSTed `contractChainId: SID.toString()` (decimal, e.g. `"9223372036854788153"`); the new path goes through `fetchCoprocessorSignatures` which serializes `uintToHex0x(chainId)` (`sdk/js-sdk/src/core/modules/relayer/module/fetchCoprocessorSignatures.ts:27`), so the POST now carries `"0x8000000000003039"` (the new test at `sdk/js-sdk/src/solana/actions/submitInputProof.test.ts:113` pins exactly this). The EVM path already sends hex so the relayer likely accepts it, but the Solana `/v2/input-proof` path has only ever been exercised with decimal by these scripts, and Solana chain ids are >2^63. The PR body claims "the request wire ... unchanged" — it isn't. Since the full vertical is listed as a remaining draft gate, make sure that gate actually runs at this head before merge; it is the only thing that validates this format change end-to-end.

2. **Parity gap with the EVM path: no empty-handles rejection.** `fetchVerifiedInputProof` (the EVM counterpart, `sdk/js-sdk/src/core/coprocessor/fetchVerifiedInputProof.ts:45-49`) throws `Input proof must contain at least one external handle`; `submitInputProof` doesn't. A zero-value `SolanaZkProof` would round-trip "successfully" with a vacuous handle check (`0 == 0`) and return empty handles/signatures. Cheap guard, worth adding for parity.

3. **No cross-check that `inputProof.chainId === fhevm.solanaChain.id`** (`sdk/js-sdk/src/solana/actions/submitInputProof.ts:37`). The action targets the client's chain's relayer but forwards the proof's own chainId as `contractChainId`, so a proof built against a different chain is submitted silently and only fails (or worse, succeeds) relayer-side. A one-line assert would catch caller mix-ups early.

## Nits

4. **The synthetic `as unknown as ZkProof` / `as unknown as Parameters<...>[0]` casts** (`submitInputProof.ts:36-46`, `:56-65`) couple this action to the *current* field usage of `fetchCoprocessorSignatures` with no compiler protection — if that implementation starts touching other `ZkProof` members or `RelayerClientWithRuntime.client`, this breaks silently at runtime. Consider a narrow structural parameter type on the relayer side (it only reads `chainId`, `contractAddress`, `userAddress`, `ciphertextWithZkProof`, `getExtraData` and `chain.fhevm.relayerUrl`) instead of double-casting, even as a follow-up.

5. **Signatures are returned unverified** — unlike the EVM path, which runs `verifyInputProof` against the coprocessor signer set. This appears deliberate (the Solana host verifies the attestation in-frame on consumption, per the comments in `solana/scripts/e2e/full-vertical.sh`), but the `submitInputProof` doc comment says only that it "verifies the returned handles"; a one-line note that signature verification is delegated to on-chain consumption would prevent callers assuming EVM-equivalent guarantees.

## Context assessment

- The base58 identity derivation (`base58.encode(hexToBytes32(...))`) is byte-for-byte equivalent to the `solana address -k` values the scripts previously plumbed through `IN_*_B58`, so dropping that env plumbing is safe; `hexToBytes32` left-pads short inputs rather than rejecting them, but both identities originate from full 32-byte proof fields here.
- `@scure/base@1.2.6` promotion to a direct runtime dependency is correctly reflected in `package-lock.json` (drops `devOptional`), and pinning matches the existing `@noble/*` pinning convention.
- The launcher's stdout contract change (relayer raw JSON → `{handles, signatures, extraData}`) is updated consistently in both consuming scripts, including the failure branches.
- Tests exercise the queued-GET continuation, schema rejection, all three mismatch classes, and terminal relayer errors — good coverage for the new surface.

Eikix:
🤖 Automated review — HEAD 617ea17 (follow-up since f78e3b9)

**TL;DR:** Approve-quality follow-up. The single new commit (`617ea17 fix(sdk): validate Solana input proof submission`) directly and correctly addresses the actionable findings from the previous review. No new findings.

## Prior findings — status

1. **`contractChainId` decimal → hex wire change / inaccurate PR description** — addressed. The PR description now states the hex serialization explicitly (`0x8000000000003039` for the test chain) and notes the previous exact head passed the Solana full vertical with the hex payload. The remaining ask is unchanged and the PR body itself lists it as a gate: run CI + the Solana full vertical at *this* head before merge.
2. **Missing empty-handles rejection (EVM parity)** — fixed. `submitInputProof` now throws `InputProofError` with the same message as `fetchVerifiedInputProof` (`sdk/js-sdk/src/solana/actions/submitInputProof.ts:40-44`), with a test asserting the relayer is never contacted (`submitInputProof.test.ts:123-135`).
3. **No proof-vs-client chain-id cross-check** — fixed. A mismatch now throws before any network call (`submitInputProof.ts:45-49`), also covered by a no-fetch test (`submitInputProof.test.ts:137-150`).
4. **Double `as unknown as` casts coupling to `fetchCoprocessorSignatures` internals** — not addressed; it was flagged as a nit with a follow-up suggested, so deferring is acceptable. Still worth a narrow structural parameter type on the relayer side eventually.
5. **Doc comment silent on signature verification** — fixed. The doc comment now states signatures are verified by the Solana host program on consumption, not by this SDK action (`submitInputProof.ts:28-32`).

## New findings

None. The two new guards run before any relayer traffic, reuse the existing `InputProofError` type consistently with the EVM path (`fetchVerifiedInputProof.ts`), and the new tests pin both error messages and the absence of network calls. The rest of the diff vs `feature/solana` is byte-identical to the previously reviewed head.

PR #3164, opened 2026-07-14
test(test-suite): run current Solana decrypt through fhevm-cli

Eikix:
## Outcome

Make current Solana user decryption independently discoverable and runnable through a named fhevm-cli profile while keeping signing, relayer transport, proof verification, and de-signcryption inside the public SDK.

## Changes

- add `fhevm-cli test solana-current-user-decrypt`;
- keep the flow in a focused Solana module rather than expanding the generic test command;
- accept the explicit current handle and `aclValueKey` already required by the SDK;
- assert exactly one expected plaintext and preserve SDK errors unchanged;
- replace the current-decrypt Bash launcher block with the named profile;
- reduce and rename the remaining loose launcher to historical-only for #1636;
- leave the shared encrypted-value producer modes intact because they have other callers.

## Validation

- all fhevm-cli unit tests: 236 passed
- focused CLI and current-decrypt tests: 43 passed
- `bun run check`
- profile discovery through `fhevm-cli test list`
- historical launcher static build
- `bash -n solana/scripts/e2e/full-vertical.sh`
- `git diff --check`
- pre-commit hooks
- two adversarial review passes; no remaining actionable finding

## Remaining draft gates

- current-head CI and the real Solana full vertical;
- verify the pending issue-level KMS-impact classification (the SDK request and connector wire behavior are unchanged).

Implements zama-ai/fhevm-internal#1711. The issue should close only after the named profile passes the real vertical and the change is merged.


Eikix:
🤖 Automated review — HEAD 62bb1cd

**TL;DR:** Solid, well-scoped refactor: the current-decrypt flow moves into a named, unit-tested `fhevm-cli` profile with a clean dependency seam, and the loose launcher is correctly reduced to historical-only. No blocking issues found; two behavioral points worth checking before the real vertical run.

## Findings

### Should-fix

1. **Removed `process.exit(0)` may reintroduce a hang** — `test-suite/fhevm/solana-userdecrypt-historical.ts` (end of file): the old `solana-userdecrypt-full.ts` ended with an explicit `process.exit(0)`, which usually exists because something (vendored TKMS WASM, relayer HTTP keep-alive) keeps the event loop alive after success. The rename drops it, and the new CLI path relies on natural drain too (`src/cli.ts` only sets `process.exitCode = 1` on failure, never force-exits on success). If that exit was load-bearing, both the historical leg and the new `fhevm-cli test solana-current-user-decrypt` leg of `full-vertical.sh` will hang after printing OK. Worth verifying on the real vertical, or keeping the explicit exit.

2. **`PreflightError` used for post-run assertion failures** — `test-suite/fhevm/src/solana/current-user-decrypt.ts:448,453`: "returned N clear values" and "cleartext X != expected Y" are runtime test failures, not preflight conditions. Since `src/errors.ts` frames these classes as distinct user-facing categories, a plain `Error` (or a dedicated assertion error) would keep failure taxonomy honest. Env/format validation using `PreflightError` (lines 355–395) is fine.

### Nits

- `test-suite/fhevm/src/solana/current-user-decrypt.ts:365-367`: the error message says "must be an even-length hex value" but the same branch also fires for odd characters (non-hex input). Message could say "even-length hex string".
- Env-parsing helpers (`required`/`bytes`/hex-csv splitting) are now duplicated between `src/solana/current-user-decrypt.ts` and the root `solana-userdecrypt-historical.ts`. Acceptable given the root script sits outside the tsconfig `include` (so it can't share typed code with `src/`), but worth a pointer comment if the historical script lives long past #1636.

### Verified / no action

- `defineFhevmSolanaChain` import move from `@fhevm/sdk/chains` to `@fhevm/sdk/solana` is valid — exported from `sdk/js-sdk/src/solana/index.ts:33`.
- The dynamic-import seam (`await import(solanaModule)`) is justified: `tsconfig.json` includes `src/**/*.ts`, and the SDK `exports` map points at generated `_types`, absent in clean checkouts. The comment in the file explains this.
- Profile dispatch in `src/commands/test.ts:871` correctly runs before the `loadState()` bootstrap check, so the profile doesn't require `fhevm-cli up` — appropriate for the Solana vertical.
- New strict validation (0x-prefixed 32-byte handle/keys, nonce as bytes32, validity fields required together) matches what `full-vertical.sh` produces (`$USER`/`$H`/`$VK` are all 0x-hex); stricter than the old launcher but reasonable. The SDK's nonce is documented as 32 bytes, so the tightened `bytes32` check is correct.
- Unit tests cover env completeness, request shape (including absence of `mmrProof`), result-count and plaintext assertions, and unchanged SDK error propagation.

### Context

PR is honest about remaining gates (real Solana full vertical not yet run against this head; issue-level KMS-impact classification pending on zama-ai/fhevm-internal#1711). Given the vertical is the only thing that exercises `runPublicSdkUserDecrypt` and the renamed historical script end-to-end, the exit-behavior question above should be answered by that run.

Eikix:
🤖 Automated review — HEAD 83d1b34 (follow-up since 62bb1cd)

**TL;DR:** The single new commit (`fix(test-suite): classify decrypt assertion failures`) cleanly addresses the prior review's actionable feedback. No new findings; LGTM on the delta.

## Prior findings status

1. **Should-fix: `PreflightError` used for post-run assertion failures** — **addressed.** The result-count and cleartext-mismatch checks in `test-suite/fhevm/src/solana/current-user-decrypt.ts:124,129` now throw plain `Error`, keeping `PreflightError` for genuine preflight/env validation. Verified the CLI's top-level handler (`src/errors.ts` `formatCliError`) prints plain `Error` messages identically and still sets `process.exitCode = 1`, and the unit tests assert on message text so they remain valid.
2. **Nit: "even-length hex value" message** — **addressed** ("hex string" at `current-user-decrypt.ts:42`).
3. **Should-verify: removed `process.exit(0)` / possible event-loop hang** — **not code-changed, still open as a runtime check.** This was flagged for verification on the real Solana full vertical, which the PR body still lists as a pending draft gate. No action needed in the diff; just confirm both the historical launcher and `fhevm-cli test solana-current-user-decrypt` exit cleanly when the vertical runs.
4. **Nit: env-helper duplication with the historical script** — unchanged; was accepted as-is given the tsconfig boundary.

## New findings

None. The delta touches only three lines in one file; the whole-diff-vs-base sanity check shows nothing else changed since the prior review.

PR #3159, opened 2026-07-14
test(test-suite): thin Solana full vertical operator coverage

Eikix:
## Outcome

Keep the expensive Solana full vertical representative instead of repeating operator semantics already proven by the pure catalog, compiled-host Mollusk suite, and direct real-TFHE tests.

## Changes

- reduce the networked operator sweep from 28 executions to 8 materially distinct integration shapes;
- retain encrypted/encrypted, encrypted/scalar, cast, ternary, bounded-randomness, vector, membership, and composite wiring;
- leave verified input, public output, MMR proof, current/historical decrypt, and confidential-token composition flows untouched;
- document the evidence boundary and the pre-change runtime baseline.

The patch changes three files with 17 additions and 64 deletions.

## Validation

- pre-commit hooks
- `bash -n solana/scripts/e2e/full-vertical.sh`
- `git diff --check`
- static retained-mode verification
- independent exact-head adversarial review: no actionable findings

Historical baseline from four successful Actions runs:

- vertical script: 4m15s–4m29s
- complete job: 44m49s–53m53s, dominated by setup

## Remaining draft gates

- current-head ordinary CI and `solana-e2e/full-vertical-reconstruct`;
- record the post-thinning vertical-script duration;
- verify the pending issue-level KMS-impact classification (repository evidence currently indicates no KMS change).

Closes zama-ai/fhevm-internal#1696 after the remaining runtime evidence is recorded and the change is merged.

PR #3165, opened 2026-07-14
feat(sdk): add Solana public decrypt certificate action

Eikix:
## Outcome

Add a public Solana SDK action that retrieves and validates the shape and MMR binding of a public-decrypt certificate claim without treating the returned plaintext as authoritative before on-chain verification.

This PR is stacked on #3164 so the loose public-decrypt launcher can be replaced through the same focused fhevm-cli flow structure.

## Changes

- add signer-free `publicDecryptCertificate` SDK action/client support;
- accept one canonical mode-prefixed MMR transport blob and reject non-public mode;
- require `proofSlot == leafCount` and verify MMR inclusion before network access;
- build version-`0x03` extra data from the same proof bytes;
- require nonempty ABI cleartext, at least one exact 65-byte signature, and byte-identical returned extra data;
- return an explicitly untrusted certificate claim for later `disclose_*_secp` verification;
- expose the same zero-logic action on the existing decrypt client for mixed flows;
- add `fhevm-cli test solana-public-decrypt`;
- migrate the full vertical while preserving its consume-compatible JSON;
- delete the 109-line loose `solana-publicdecrypt.ts` launcher.

No Codama, transaction client, token-specific interpretation, off-chain threshold verification, or KMS wire change is introduced.

## Validation

- focused SDK tests: 14/14
- focused SDK lint and type analysis
- focused CLI/public-decrypt tests: 46/46
- `bun run check`
- `bash -n solana/scripts/e2e/full-vertical.sh`
- `git diff --check`
- commit hook
- independent adversarial review: no actionable findings

The wider SDK build retains two parent/base test-type errors outside this diff.

## Remaining draft gates

- #3164 must land, then this branch must be rebased and retargeted to `feature/solana`;
- current-head CI and the real Solana full vertical must prove on-chain certificate consumption;
- the pending issue-level KMS-impact classification must be recorded (repository evidence indicates no KMS change).

Implements the public-decrypt half of zama-ai/fhevm-internal#1624. The issue remains open for the separate Codama-backed `confidentialTransfer` action.


Eikix:
🤖 Automated review — HEAD d1e352d

**TL;DR:** Solid, well-tested change. The SDK action does local MMR verification before any network access, accepts only the canonical mode-prefixed blob, and correctly matches the relayer's documented wire format (no-0x `decryptedValue`/`signatures`, 0x `extraData`). No blocking issues; one compatibility question about the hard `extraData` requirement, plus a few e2e-script nits.

### Should-fix

1. **`extraData` hard requirement vs. relayer optionality** — `sdk/js-sdk/src/solana/actions/publicDecryptCertificate.ts` (missing-`extraData` throw): the core guard `RelayerPublicDecryptSucceeded.ts` explicitly says "In v11 extraData is optional" and treats it as such, and the e2e consume path historically used `.get('extraData', fallback)`. The new action throws whenever the relayer omits `extraData`. If the certificate flow genuinely needs the echoed extra data for `disclose_*_secp` (likely, since it's part of what the KMS signs over), the strictness is right — but it's worth a one-line comment stating the relayer-version assumption, and per the PR's own draft gates this hasn't yet been proven against the real vertical at head. Just confirm the deployed relayer for the Solana v2 path always echoes `extraData`.

2. **Dead 4th positional argument in `run_public_decrypt_with_proof`** — `solana/scripts/e2e/full-vertical.sh:48`: the function no longer reads `$4` (the old `expected`), but callers at lines 180, 302, and 549 still pass a value in that slot while the function reads `${5:-}`/`${6:-}`. Cleartext assertions are preserved at the callsites (lines 182, 283, 304, 555), so no coverage is lost, but a silently ignored positional arg between used ones is a trap for future edits — drop the arg at callsites and renumber, or comment why the slot is reserved.

### Nits

3. **`| tail -1` loses failure diagnostics** — `solana/scripts/e2e/full-vertical.sh:97`: `set -o pipefail` (line 16) keeps the exit code, so `|| fail` still fires, but on failure `$result` now holds only the last output line, whereas the old launcher surfaced the whole error. Consider capturing full output and extracting the JSON line.

4. **`solana-public-decrypt` bypasses `runLogged`** — `test-suite/fhevm/src/commands/test.ts:884`: unlike `solana-current-user-decrypt` (line 881), this profile is run bare — presumably deliberate so stdout's last line stays the machine-readable JSON envelope for the e2e's `tail -1`. A short comment would stop someone "fixing" the inconsistency and breaking the vertical.

5. **Fractional byte count in signature error** — `publicDecryptCertificate.ts` signature loop: `got ${signature.length / 2} bytes` prints e.g. `1.5 bytes` for odd-length input; report the hex length or round instead.

### Context assessment

The PR replaces the 109-line loose `test-suite/fhevm/solana-publicdecrypt.ts` launcher with a dependency-injected CLI flow mirroring the `solana-current-user-decrypt` structure from #3164 (the base of this stack), and the JSON envelope preserved for the consume steps matches what the script reads (`result.decryptedValue/signatures/extraData` only — verified no other fields of the old relayer body were consumed). Validation logic checked against `proof.ts` (`decodeMmrProofTransportBlob` enforces exact-length Borsh with mode byte; `verifyPublicDecryptProof` recomputes the leaf commitment) and against the relayer wire-format guard — all consistent. Test coverage on both the SDK action (mode/trailing-byte/inclusion/slot/malformed-certificate/terminal-error) and the CLI flow (env plumbing, missing-env, error passthrough) is thorough.

Reminder from the PR's own gates: base is currently `elias/solana-cli-current-user-decrypt`; retarget to `feature/solana` after #3164 lands.

Eikix:
🤖 Automated review — HEAD 0633767 (follow-up since d1e352d)

**TL;DR:** LGTM. The single new commit (`fix(sdk): harden Solana public decrypt flow`) directly addresses the prior review's findings; the delta is small, well-tested, and I found no new issues.

### Prior findings — status

1. **`extraData` hard requirement vs. relayer optionality (should-fix)** — **addressed.** `sdk/js-sdk/src/solana/actions/publicDecryptCertificate.ts` now treats the response `extraData` as optional (matching the core guard's "In v11 extraData is optional"): if present it must still be byte-identical to the request, and if absent the claim falls back to the request's own `requestExtraDataHex`. That fallback is semantically sound — the request extra data is what the flow submitted and what `disclose_*_secp` will verify against — and a new test covers the omitted-field path.
2. **Dead 4th positional arg in `run_public_decrypt_with_proof` (should-fix)** — **addressed.** `solana/scripts/e2e/full-vertical.sh:49-53`: parameters renumbered to `${4:-}`/`${5:-}`, all four callsites updated (including the `burned` call passing `2 1`), and the now-unused `public_decrypt_expected` plumbing in `assert_decrypt` removed. Cleartext assertions at the callsites are unchanged.
3. **`| tail -1` losing failure diagnostics (nit)** — **addressed.** The CLI invocation now captures full output, fails with the whole output on error, and extracts the JSON line only on success (`full-vertical.sh:91-100`).
4. **Missing comment on `solana-public-decrypt` bypassing `runLogged` (nit)** — **not addressed.** `test-suite/fhevm/src/commands/test.ts:884-888` still runs the profile bare with no comment explaining that stdout's last line must remain the machine-readable JSON for the e2e's `tail -1`. Fine to leave, but still a small trap for future edits.
5. **Fractional byte count in signature error (nit)** — **addressed.** The error now reports `got ${signature.length} hex characters`, with a new odd-length test case.

### New findings

None. The relaxed `extraData` guard keeps the mismatch check when the field is present, the returned claim's `extraData` is always populated, and the shell changes were checked against all callers.

Standing reminders from the PR's own draft gates (unchanged): retarget to `feature/solana` once #3164 lands, and the real full vertical at head still needs to prove on-chain certificate consumption.

Eikix:
🤖 Automated review — HEAD 7a8571f (follow-up since 0633767)

**TL;DR:** LGTM — this is a pure rebase. `git range-diff` shows both PR commits are content-identical to the previously reviewed head (`d1e352d = 625bbcf`, `0633767 = 7a8571f`); the only change is the new base: the stack (#3159, #3163, #3164, #3166) has landed on `feature/solana` and the PR now targets it directly. No new code to review.

### Rebase composition check

- The retarget gate from the PR description is done: base is now `feature/solana`, PR is out of draft, and GitHub reports it mergeable.
- Checked for semantic conflicts with #3166 (which deleted unused live-client modes): every mode `solana/scripts/e2e/full-vertical.sh` still invokes (`BOOTSTRAP`, `TRIVIAL_ENCRYPT_EVAL`, `HISTORICAL_STEP`, `PUBLIC_DECRYPT_PROOF`, `CONSUME_*`, `FHE_EVAL_*`) survives in `solana/scripts/e2e/live-client/main.rs` at this head. No drift.

### Prior findings — status

All prior should-fixes and nits were already resolved at 0633767 except one, and it remains open:

- **`solana-public-decrypt` bypasses `runLogged` without a comment (nit)** — `test-suite/fhevm/src/commands/test.ts:884-886` still runs the profile bare (vs. `runSolanaCurrentUserDecrypt` at line 881 which is wrapped). Presumably deliberate so stdout's last line stays the machine-readable JSON envelope for the e2e's `tail -1`, but a one-line comment would stop someone "fixing" the inconsistency and breaking the vertical. Not blocking.

### New findings

None.

Remaining gate from the PR's own description: the `solana-e2e/full-vertical-reconstruct` check is still pending at this head — the real full vertical proving on-chain certificate consumption is the last thing to watch before merge.

PR #3169, opened 2026-07-14
feat(sdk): add Solana confidential transfers

Eikix:
## Summary

- Adds the public `@fhevm/sdk/solana` `confidentialTransfer` action and exposes it through the encrypt client.
- Uses injected Solana Kit RPC and subscriptions with explicit owner and fee-payer signers.
- Generates the minimal confidential-token client closure from the committed Anchor IDL through Codama and checks generated drift during builds.

## Stack and scope

- Implements `zama-ai/fhevm-internal#1624`.
- Stacked on #3163 (`elias/solana-sdk-submit-input-proof`); this PR must be reviewed and merged after that base.
- KMS impact: none. No change is required in `zama-ai/kms` or its `feature/solana` branch.
- The full vertical and two-holder flow tracked by `zama-ai/fhevm-internal#1625` remain later acceptance gates.

## Invariants

- The returned relayer handles are re-bound to the original proof before any transaction is built.
- The selected input must be an `euint64` for the client's exact high-bit Solana chain, ACL program, owner, and mint compute-signer domain.
- The configured ACL program must equal the host program compiled into the confidential-token deployment. The canonical host address is generated from the committed IDL, and mismatches are rejected before RPC.
- Codama retains only `confidentialTransfer`, its required attestation type, and compute-signer PDA. Generated files and IDL provenance are committed; `codegen:solana:check` rejects drift.
- Transactions use a fixed 400,000 compute-unit limit, are simulated with signature verification, and are sent and confirmed at `confirmed` commitment.

## Validation

- Clean CommonJS, ESM, and declaration builds.
- Solana SDK suite: 47 tests passed with no type errors.
- Focused confidential-transfer action and generated-instruction tests: 14 passed.
- Codama generated snapshot check passed.
- Prettier check passed.
- Two adversarial reviews completed; no actionable findings remain.


Eikix:
🤖 Automated review — HEAD bdf7b646

**TL;DR:** Solid, well-tested addition. The client-side attestation binding checks mirror the on-chain program faithfully (value-key derivation, seeds, labels, self-transfer/deny-record rule all verified against `solana/programs/confidential-token` and `zama-solana-acl` on `feature/solana`). No blocking issues; a few small should-fix/nit items below.

## Findings

### Blocking
None.

### Should-fix
1. **Double simulation per transfer** — `sdk/js-sdk/src/solana/actions/confidentialTransfer.ts:181-198`: the action explicitly `simulateTransaction`s with `sigVerify: true`, then sends with `skipPreflight: false`, so the RPC node runs preflight simulation a second time. Consider `skipPreflight: true` after the explicit simulation (which gives the better error), or drop the explicit call. Also note the blockhash is fetched before simulation, so a slow simulation eats into the transaction lifetime — minor, but the duplicate simulation makes it slightly worse.

### Nits
2. **`aclProgramAddress` is effectively a checksum, not a parameter** — `confidentialTransfer.ts:87-89` rejects anything but the IDL-baked `ZAMA_HOST_PROGRAM_ADDRESS`, and `CONFIDENTIAL_TOKEN_PROGRAM_ADDRESS` cannot be overridden at all (the generated builder supports `config.programAddress` but the action doesn't expose it). This matches the stated PR invariant, so fine for the PoC — but it means the public action can't target a localnet/devnet deployment with different program keys. Worth a TODO/issue for post-PoC configurability.
3. **Silent pad/truncate of malformed signatures** — the generated encoder uses `fixEncoderSize(getBytesEncoder(), 65)` for `signatures` (`generated/confidentialToken/types/coprocessorInputAttestation.ts`), so a wrong-length signature from the relayer would be zero-padded/truncated and fail on-chain instead of client-side. A cheap length assert in the action (next to the other binding checks) would give a clearer error.
4. **Codegen check couples the SDK build to the monorepo layout** — `scripts/build/codegen-solana-confidential-token.mjs` resolves the IDL at `../../../../coprocessor/fhevm-engine/host-listener/idl/confidential_token.json`, and `build:profile` (hence `prepublishOnly`) now hard-depends on it. Fine while the SDK is only built in-tree; will break any standalone/out-of-tree build of the package.
5. Redundant guard: `confidentialTransfer.ts:100` (`inputHandle === undefined`) can't be reached after the index-range check on line 96 — harmless noise.

## What was verified against the on-chain program
- `deriveValueKey` prefix/ordering matches `zama_solana_acl::derive_value_key` (`zama-encrypted-value-key-v1`, acl-domain-key = mint), and `transferred_amount______________` matches `transferred_amount_label()`.
- PDA seeds match: `encrypted-value`, `hcu-authority` + mint, `fhe-compute` + mint (compute signer), `__event_authority` on both programs.
- The self-transfer + deny-records rejection mirrors `assert_no_remaining_accounts` on the program's `from_key == to_key` no-op path.
- Attestation encoding (u8 handle index, u64 chain id, u32-prefixed extraData, 65-byte sigs) matches the committed IDL.

## Context
- Base PR #3163 is merged into `feature/solana`; this branch is one commit behind the base (#3165, unrelated public-decrypt certificate action) — a trivial rebase.
- Test coverage is good: attestation-binding rejection matrix, decoded-wire assertions on compute-budget + account roles, simulation-failure short-circuit, and a Codama drift check wired into every build.

sdk/js-sdk/src/solana/actions/confidentialTransfer.ts
```diff
@@ -0,0 +1,196 @@
...
+import { deriveValueKey } from '../proof.js';
+import { getConfidentialTransferInstructionAsync } from '../internal/generated/confidentialToken/instructions/confidentialTransfer.js';
+import { findComputeSignerPda } from '../internal/generated/confidentialToken/pdas/computeSigner.js';
+import {
+  CONFIDENTIAL_TOKEN_PROGRAM_ADDRESS,
+  ZAMA_HOST_PROGRAM_ADDRESS,
+} from '../internal/generated/confidentialToken/programAddress.js';
+
+const EVENT_AUTHORITY_SEED = new TextEncoder().encode('__event_authority');
+const HCU_AUTHORITY_SEED = new TextEncoder().encode('hcu-authority');
```
Eikix, line 43: I thought HCU authority was going to be removed and merged into compute_subject?

PR #3170, opened 2026-07-14
feat(test-suite): test partial KMS upgrades

Eikix:
## Summary

Adds a rollout API that applies a KMS core version to any chosen serving nodes while preserving the rest of the fleet:

```ts
await ctx.upgradeKmsNodes([1, 2], { lockFile: targetLock });
```

Nodes are recreated sequentially, mixed fleets survive resume and repair, and each checkpoint records the observed Docker images. The runbook chooses the node combinations and tests; the executor does not prescribe a rollout matrix.

This PR also includes the dispatch-only stateful rollout workflow previously proposed in [#3162](https://github.com/zama-ai/fhevm/pull/3162), so the implementation, its runbooks, and its CI entry point can be reviewed together.

## Progressive KMS rollout

The generic runbook upgrades threshold KMS nodes one by one. After each replacement it:

1. Leaves exactly a `2t+1` quorum online that must include the node just upgraded.
2. Runs one deterministic user-decryption probe, proving that node supplied a usable share.
3. Restores the stopped KMS parties, including when the probe fails.
4. Runs `rollout-standard` against the complete healthy fleet.

The baseline also runs `rollout-standard`. Receipts retain the live and stopped node IDs, applied versions, Docker evidence, test results, and restoration outcome.

## Mainnet incident reproduction

`rollouts/mainnet-v0.11-to-v0.12-kms/run.ts` pins the deployed component versions from the 2026-07-13 incident, including KMS core `v0.13.3` to `v0.13.10` and relayer-sdk `0.4.2`.

It exercises the required checkpoints:

1. Four old nodes pass `rollout-standard`.
2. Nodes 1 and 2 are upgraded. The complete 2-old/2-new fleet must fail one deterministic `ebool` user-decryption probe with the relayer-sdk `0.4.2` reconstruction wrapper.
3. Nodes 3 and 4 are upgraded. Four target-version nodes pass `rollout-standard`.

The test-suite override makes the SDK pin effective without changing SDK source. Expected-failure output and Docker image evidence are retained in the receipt. If recreation or readiness fails after a node may have changed, the failed-upgrade evidence is recorded before the original error is rethrown.

## CI execution

`.github/workflows/test-suite-stateful-rollout.yml` accepts a checked-in runbook path through manual dispatch. It uses the same pinned setup actions and registry credentials as the existing E2E workflows, bounds the rollout step to 185 minutes, uploads Markdown and JSONL receipts even on failure, and always attempts cleanup.

Because GitHub registers a new `workflow_dispatch` workflow only after it exists on the default branch, the first isolated rollout must be dispatched on the resulting `main` commit immediately after merge. It cannot run against this PR head before merge while the workflow remains dispatch-only.

## Scope

- Changes are confined to the test suite and its CI entry point.
- No SDK, KMS Connector, or KMS source is changed.
- The generic runbook temporarily stops only enough KMS parties to require the newly upgraded node in the reconstruction quorum, then restores them.
- No broad quorum enumeration, upgrade planner, label trigger, or normal-PR rollout execution is introduced.
- Rollout integration is not run by normal local validation commands.

## Validation

- `bun run check`
- `bun test src` — 302 tests, 985 assertions
- `bun run compat-smoke`
- `zizmor --persona auditor --offline .github/workflows/test-suite-stateful-rollout.yml`
- repository pre-commit workflow checks
- `git diff --check`
- adversarial review of failure isolation, required-node participation, evidence retention, dispatch security, action pins, timeouts, and cleanup

Rollout integration was not run locally. Post-merge acceptance requires dispatching `rollouts/mainnet-v0.11-to-v0.12-kms/run.ts` and inspecting the retained receipt for the old healthy quorum, expected mixed-fleet failure, final healthy quorum, and responder/image evidence at every audit-critical checkpoint.

Tracks zama-ai/fhevm-internal#1714.

### DIFF
--- a/test-suite/fhevm/rollouts/kms-core-progressive/run.ts
+++ b/test-suite/fhevm/rollouts/kms-core-progressive/run.ts
@@ -1,0 +1,51 @@
+import type { RolloutRunContext } from "../../src/commands/rollout-run";
+import { from, scenario, to, versionSources } from "./versions";
+
+const testRollout = (ctx: RolloutRunContext) => ctx.test("rollout-standard", { parallel: false });
+const testRequiredNode = (ctx: RolloutRunContext) =>
+  ctx.test("user-decryption", { grep: "test user decrypt ebool$", parallel: false });
+
+type KmsCoreVersions = Record<string, string> & { CORE_VERSION: string };
+
+export type ProgressiveKmsCoreRunbook = {
+  from: KmsCoreVersions;
+  overrides?: Parameters<RolloutRunContext["up"]>[0]["overrides"];
+  scenario: string;
+  to: KmsCoreVersions;
+  versionSources: string[];
+};
+
+export const runProgressiveKmsCoreRollout = async (
+  ctx: RolloutRunContext,
+  config: ProgressiveKmsCoreRunbook,
+) => {
+  const baselineLock = await ctx.writeVersionLock("00-kms-core-baseline", {
+    versions: config.from,
+    sources: config.versionSources,
+  });
+  const targetLock = await ctx.writeVersionLock("01-kms-core-target", {
+    versions: config.to,
+    sources: config.versionSources,
+  });
+
+  await ctx.up({ lockFile: baselineLock, overrides: config.overrides, scenario: config.scenario });
+  await testRollout(ctx);
+
+  const state = await ctx.readState();
+  const nodeIds = Array.from({ length: state.scenario.kms.committeeSize }, (_, index) => index + 1);
+  for (const nodeId of nodeIds) {
+    await ctx.upgradeKmsNodes([nodeId], { lockFile: targetLock });
+    await ctx.withRequiredKmsNode(nodeId, () => testRequiredNode(ctx));
+    await testRollout(ctx);
+  }
+};
+
+export const run = (ctx: RolloutRunContext) =>
+  runProgressiveKmsCoreRollout(ctx, {
+    from,
+    scenario,
+    to,
+    versionSources,
+  });
+
+export default run;
--- a/test-suite/fhevm/src/commands/rollout-run.ts
+++ b/test-suite/fhevm/src/commands/rollout-run.ts
@@ -4,7 +4,7 @@
 import path from "node:path";
 import { pathToFileURL } from "node:url";
 
-import { PreflightError } from "../errors";
+import { CommandError, PreflightError } from "../errors";
 import { runContractTask, snapshotContractSources } from "../flow/contracts";
 import { waitForTestSuite } from "../flow/readiness";
 import { composeUp } from "../flow/runtime-compose";
@@ -12,12 +12,15 @@ import {
   applyVersionLock as applyStackVersionLock,
   refreshDiscovery as refreshStackDiscovery,
   up,
+  upgradeThresholdKmsNode,
   upgradeRuntimeGroup as upgradeStackRuntimeGroup,
 } from "../flow/up-flow";
 import { STATE_DIR, composePath, hostChainRuntimes } from "../layout";
+import { reconstructionThreshold } from "../kms-party";
 import { loadState } from "../state/state";
 import type { LocalOverride, State, UpOptions, VersionBundle, VersionTarget } from "../types";
 import { ensureDir, writeJson } from "../utils/fs";
+import { partyContainers, setRunning, waitForPartiesRunning, waitForPartiesStopped } from "./kms-generation";
 import { type RolloutReceipt, createRolloutReceipt } from "./rollout-receipt";
 import { test as runTest } from "./test";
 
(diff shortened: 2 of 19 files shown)


test-suite/fhevm/rollouts/kms-core-progressive/run.ts
```diff
@@ -0,0 +1,47 @@
...
+    versionByNodeId: Object.fromEntries(nodeIds.map((nodeId) => [nodeId, from.CORE_VERSION])),
+    targetVersion: to.CORE_VERSION,
+    upgradeOrder: nodeIds,
+  });
+  const stepsToTest = new Set(upgradeStepsToTest(plan.nodes.length, plan.quorum));
+
+  for (const [index, nodeId] of plan.upgradeOrder.entries()) {
+    await ctx.upgradeKmsNode(nodeId, { lockFile: targetLock });
+    const completedUpgradeSteps = index + 1;
+    if (!stepsToTest.has(completedUpgradeSteps)) {
```
Eikix: [AUTOMATED]

[P1] A final forced quorum containing node 4 is required. `upgradeStepsToTest(4, 3)` excludes step 4, so after `upgradeKmsNode(4)` only the unisolated decryption at the end runs. Nodes 1-3 already satisfy the 3-of-4 reconstruction threshold, while the upgrade health gate checks only the metrics endpoint; the rollout can therefore complete with node 4 unable to participate in MPC/decryption. An explicit final three-node quorum containing node 4 should be isolated and tested. Calling `buildKmsUpgradeStep(plan, 4)` alone would still select nodes 1-3.

Confidence: 99/100

test-suite/fhevm/src/commands/rollout-run.ts
```diff
@@ -86,6 +94,56 @@ const refreshTestSuiteContainer = async () => {
...
+
+  const stopped = kmsPartyIds(committeeSize).filter((nodeId) => !selected.has(nodeId));
+  try {
+    await operations.setRunning(stopped.flatMap(partyContainers), "stop");
+    await operations.waitForPartiesStopped(stopped);
+    await operations.waitForPartiesRunning([...nodeIds]);
+    return await task();
+  } finally {
+    await operations.setRunning([...stopped].reverse().flatMap(partyContainers), "start");
+    await operations.waitForPartiesRunning(stopped);
```
Eikix: [AUTOMATED]

[P2] The original quorum/probe error needs to be preserved when restoration also fails. If `task()` throws and this readiness wait throws from `finally`, JavaScript replaces the pending task error with the restoration error. The outer failure handler then records only the replacement message, losing the reason the quorum test failed. Both failures should be retained, for example with an `AggregateError`.

Confidence: 99/100

test-suite/fhevm/src/commands/rollout-run.ts
```diff
@@ -170,10 +228,31 @@ export const createRolloutContext = (receipt: RolloutReceipt = createRolloutRece
...
     await upgradeStackRuntimeGroup(group, options);
     await receipt.record("upgrade-runtime", group, { lockFile: options.lockFile });
   },
+  async withKmsNodes(nodeIds, task) {
+    const state = await loadState();
+    if (!state) {
+      throw new PreflightError("Stack is not running; run ctx.up(...) first");
+    }
+    return runWithKmsNodes(state, nodeIds, async () => {
+      await receipt.record("kms-quorum", `KMS nodes ${nodeIds.join(",")}`, {
```
Eikix: [AUTOMATED]

[P2] These audit-critical Docker snapshots need to fail closed. `inspectContainers()` converts `docker inspect` and container-listing failures into `dockerInspectError`, `record()` still resolves, and no later completeness check prevents a `complete` receipt. The run can therefore finish without the observed responder/image IDs required by fhevm-internal#1714. A failed snapshot should fail the rollout, or final receipt validation should reject missing runtime evidence.

Confidence: 92/100

Eikix (commented):
[AUTOMATED]

## Code Review

Blocking findings were identified: one rollout correctness defect and two failure/evidence-handling defects, detailed inline.

The isolated integration acceptance criterion is also unmet: fhevm-internal#1714 requires automated integration coverage on an isolated CI runner, but this PR only exercises mocked contexts and states that the rollout was not run. PR #3162 supplies the dispatch-only workflow, but it remains open and is not declared as a prerequisite. #3162 should land first, then this exact runbook should be dispatched successfully and its receipt retained before this PR is considered ready.

dd23:
>  four-node threshold KMS fleet whose cores are upgraded one at a time.

Note that we're in any case able to tolerate a single node failing out of 4, so the real test we want to do here is upgrading 2 out of the 4, as the (1,3) and (3,1) cases will work regardless of versions used if we just look at compute outputs. We could however inspect concrete KMS responses and see if we always get 4 good ones. In that case the 1,3 and 3,1 cases would make sense as well.

obatirou (approved):

PanGan21 (approved):

PR #3174, opened 2026-07-15
fix(host-contracts): use fully-qualified EmptyUUPSProxy name in bridg…

PanGan21:
### Summary

Qualify the `EmptyUUPSProxy` lookup in `task:deployBridge` and `buildUpgradeProposal` with its fully-qualified name so the bridge deploy and upgrade paths resolve unambiguously during upgrades, where both contracts/ and oldContracts/ are compiled and the bare name fails with HH701.

### DIFF
--- a/host-contracts/tasks/taskDeploy.ts
+++ b/host-contracts/tasks/taskDeploy.ts
@@ -500,7 +500,10 @@ task('task:deployBridge').setAction(async function (_, { ethers, upgrades }) {
     return;
   }
 
-  const currentImplementation = await ethers.getContractFactory('EmptyUUPSProxy', deployer);
+  const currentImplementation = await ethers.getContractFactory(
+    'contracts/emptyProxy/EmptyUUPSProxy.sol:EmptyUUPSProxy',
+    deployer,
+  );
   const newImplem = await ethers.getContractFactory('ConfidentialBridge', deployer);
 
   const proxy = await upgrades.forceImport(proxyAddress, currentImplementation);
--- a/host-contracts/tasks/utils/upgradeProposal.ts
+++ b/host-contracts/tasks/utils/upgradeProposal.ts
@@ -47,7 +47,10 @@ export async function buildUpgradeProposal(
   const { ethers, upgrades } = hre;
   const privateKey = getRequiredEnvVar('DEPLOYER_PRIVATE_KEY');
   const deployer = new ethers.Wallet(privateKey).connect(ethers.provider);
-  const currentImplementation = await ethers.getContractFactory('EmptyUUPSProxy', deployer);
+  const currentImplementation = await ethers.getContractFactory(
+    'contracts/emptyProxy/EmptyUUPSProxy.sol:EmptyUUPSProxy',
+    deployer,
+  );
   const newImplementation = await ethers.getContractFactory(params.contractName, deployer);
   await upgrades.forceImport(params.proxyAddress, currentImplementation);
   const newImplementationAddress = String(


host-contracts/tasks/taskDeploy.ts
```diff
@@ -500,7 +500,10 @@ task('task:deployBridge').setAction(async function (_, { ethers, upgrades }) {
     return;
   }
 
-  const currentImplementation = await ethers.getContractFactory('EmptyUUPSProxy', deployer);
+  const currentImplementation = await ethers.getContractFactory(
```
isaacdecoded, line 503: Do we also need this change for the deployment flow? I mean, during deployment, we shouldn't have the previous and new contracts scenario you are targeting. And in case of adding this change, why not applying also to the other contracts deployment tasks?

host-contracts/tasks/utils/upgradeProposal.ts
```diff
@@ -47,7 +47,10 @@ export async function buildUpgradeProposal(
   const { ethers, upgrades } = hre;
   const privateKey = getRequiredEnvVar('DEPLOYER_PRIVATE_KEY');
   const deployer = new ethers.Wallet(privateKey).connect(ethers.provider);
-  const currentImplementation = await ethers.getContractFactory('EmptyUUPSProxy', deployer);
+  const currentImplementation = await ethers.getContractFactory(
```
isaacdecoded, line 50: Maybe we can try to align it with the current implementation at the [upgradeContracts.ts file](https://github.com/zama-ai/fhevm/blob/main/host-contracts/tasks/upgradeContracts.ts#L68-L70) where we explicitly pass the currentImplementation and the newImplementation paths, so we avoid any ambiguity.

host-contracts/tasks/taskDeploy.ts
PanGan21: > Do we also need this change for the deployment flow?

This is the task called for the deployment flow of zws.

> And in case of adding this change, why not applying also to the other contracts deployment tasks?

Not sure if I should change all references since I don't know what could be the effect. I am targetting to fix the current failure.
The repo itself is inconsistent with the pattern - see how ProtocolConfig was deployed - something that I don't think should change in this pr.

obatirou: Please also note that this should be shared with OZ asap as this file is in scope for current audit.
imo we should fix the bug only and do an issue for standardization but they could provide direction if needed.

host-contracts/tasks/utils/upgradeProposal.ts
PanGan21: There is no real currentImplementation and the newImplementation for the bridge contract. This is new deployment.
The flow is:
- Deploy empty proxy
- Upgrade ACL
- Deploy implementation of bridge and link it with the deployed proxy

host-contracts/tasks/taskDeploy.ts
PanGan21: Ticket for standardizing the fully qualified naming in the repo: https://github.com/zama-ai/fhevm-internal/issues/1731

host-contracts/tasks/utils/upgradeProposal.ts
PanGan21: Maybe i missunderstood. Do you mean to pass the currentImplementation and newImplementation to the buildUpgradeProposal?
In that case some unrelated changes to the fix will be added to `protocolConfigMirror.ts`

host-contracts/tasks/taskDeploy.ts
isaacdecoded: Still not sure how is that the full qualified path is needed in the `task:deployBridge`, but not for other deploy tasks, e.g., the `task:deployProtocolConfig`. These are deployment tasks, so as I asked above, the current vs new implementations shouldn't take effect - in other words, the deployment tasks should be executed in an environment containing a single implementation source.

I'm just asking to understand the solution in this PR.

host-contracts/tasks/utils/upgradeProposal.ts
isaacdecoded: > There is no real currentImplementation and the newImplementation for the bridge contract. This is a new deployment.

AFAICT, upgrade tasks are for contracts already deployed, meaning that there must be a current and a new implementation. For a new deployment, there's a dedicated task for it, i.e., the `task:deployBridge` below

host-contracts/tasks/taskDeploy.ts
PanGan21: I went through the history, and that's why the other deploy tasks never hit this:

- `task:deployProtocolConfig` was called directly after the `deployEmptyProxy` step and before any upgrade task so `oldContracts/` didn't exist yet -> single source -> the bare lookup resolves fine.
- `task:deployBridge` has to run after the ACL upgrade. By that point the ACL upgrade step has already created/compiled `oldContracts/` -> two sources -> HH701.

host-contracts/tasks/utils/upgradeProposal.ts
PanGan21: `task:deployBridge` is for EOA owners. This is related to deploying the bridge with a DAO proposal similarly to how the protocol config was deployed

PanGan21: Ref here: https://github.com/zama-ai/fhevm/blob/d2899a73c8bdbbd8e94305ba3d8d1dc4abfb3dfa/host-contracts/tasks/upgradeContracts.ts#L363

PanGan21: What i can do to improve readability is to pass `currentImplementation` and `newImplementation` to buildUpgradeProposal. This will introduce an unrelated change on the protocol config calling this function. Imo this can be done in the refactor task

isaacdecoded: Ok, that's the part that is confusing to me: for a contract deployment from scratch, we have never used DAO proposals -  I might be missing something though cc @obatirou.

In fact, taking a look at the v0.13.x when we introduced the ProtocolConfig contract, [we didn't have an upgrade task](https://github.com/zama-ai/fhevm/blob/v0.13.2-0/host-contracts/tasks/upgradeContracts.ts) at all for this contract (so I don't get what you mean above about having a DAO proposal to [deploy the ProtocolConfig](https://github.com/zama-ai/fhevm/blob/v0.13.2-0/host-contracts/tasks/taskDeploy.ts#L537)).

Anyway, it seems to me that it's a matter of clarifying at the implementation level. Maybe we should rename the `task:prepareUpgradeConfidentialBridge` to something like `task:prepareDaoDeployConfidentialBridge` and move it to the `taskDeploy.ts` file since in the `upgradeContracts.ts` file we have only upgrade tasks for already deployed contracts - at least this is what I found confusing during this PR review.

isaacdecoded (approved):
Just some nitpicks based on a personal confusion, but LGTM.

host-contracts/tasks/utils/upgradeProposal.ts
obatirou: I think this was misleading for protocolConfig as it was a one time thing for protocolConfig due to the migration from gateway. https://github.com/zama-ai/fhevm/blob/129333b9d82161407541d6932e10a49703a5abe9/host-contracts/tasks/taskMigrate.ts#L207

We used a DAO proposal for the upgrade with call to `initializeFromMigration`.
When we deploy an empty proxy, they are owned by ACL owner, so the DAO proposal is mandatory on ethereum. On new chains, this is different (we first deploy everything with deployer key then transfer ownership)

If you have any doubt about how a deployment was done for a specific tag, you should always report to the [gitops repository](https://github.com/zama-zws/gitops/blob/70256202e2144415cfa31926fed40f2e38e13d1d/values/eth-blockchain/eth-sc-deploy/values-zws-testnet.yaml#L280) which is the source of truth


We can see for 0.12 -> 0.13 in testnet: 
1. deployEmptyProxiesProtocolConfigKMSGeneration
2. prepareDeployProtocolConfigFromMigration
3. the runbook for 0.12 to 0.13 with protocol Config [initializeFromMigration](https://app.notion.com/p/zamaai/Ethereum-host-upgrades-0-13-0-3825a7358d5e8070b2d0f1a45c9baf3e?source=copy_link)

obatirou: > task:prepareDeployConfidentialBridge and move it to the taskDeploy.ts

I agree with this, especially as it will call `initializeFromEmptyProxy` vs `reinitializerVX` for upgrades.
But not sure this should be introduced here.

PanGan21: Thanks for the clarification.
@isaacdecoded  i agree that the naming might be a bit confusing. I am just thinking that the whole deploy flow has been tested using this task name. Do you think we should refactor the name in this pr?

obatirou: I agree, this might be for a next release refactor. Else we will have divergence between devnet and testnet process.

isaacdecoded: @PanGan21 @obatirou also agree to consider as a follow-up refactor - probably it's worth explaining and linking this conversation in https://github.com/zama-ai/fhevm-internal/issues/1730 or create a new one, as you prefer. Thanks to both of you for helping me to understand.

PanGan21: Added the conversation reference to this ticket which is about the general refactor: https://github.com/zama-ai/fhevm-internal/issues/1731

obatirou (approved):

PR #3189, opened 2026-07-15
ci: add preview env based on label

fegmorte:
## What & why

Adds **ephemeral, per-PR end-to-end preview environments** for the full fhEVM stack, deployed on the `zws-dev` Kubernetes cluster and driven entirely by PR labels. Envionments can also be triggered with a workflow_dispath event manually.

Each labeled PR gets its own isolated namespace (`fhevm-ci-<pr-author>-<pr-number>`) running the complete stack from real Helm charts then optionally runs the automated e2e DAG (Argo Workflow) and reports pass/fail back as a PR comment.

### The SDLC gap this fills

Today our e2e coverage has a hole in the middle of the software delivery lifecycle:

- **Early / CI:** the `docker-compose` e2e suite is fast and runs on every change, but it exercises a *simplified, single-host* composition that does **not** reflect how the stack is actually packaged and wired in production (Helm charts, per-party topology, Crossplane-provisioned infra, real KMS keygen, S3, Redis brokers).
- **Late / QA:** the first time the **real Helm charts** and the multi-party topology get exercised together is on shared, long-lived environments (**devnet / testnet**) — well after merge, where a wiring regression is expensive to trace back and blocks everyone on the shared environment.

This PR closes that gap: it brings **real-chart, production-shaped, multi-party e2e** *forward* in the lifecycle, to a **disposable per-PR environment** available before merge. It catches chart-wiring, packaging, KMS-generation and cross-component integration regressions at PR-review time — the same class of issues that otherwise only surface on devnet/testnet — without the cost or contention of a shared environment.

## How it works

Three PR labels gate the behavior:

| Label | Effect |
| --- | --- |
| `pr-preview-e2e` | Deploy the pinned versions; build no images. |
| `pr-preview-e2e-build` | Also build fresh images from the PR branch. |
| `pr-preview-e2e-tests` | Additionally run the automated e2e DAG and comment results on the PR. |

- `preview-env-deploy.yml` builds images (reusable workflows), resolves tags (`resolve-tags.cjs`), then deploys the whole stack into the per-PR namespace. Centralized version pins live in the top-level `env:` block; `workflow_dispatch` exposes the same knobs for manual runs.
- `preview-env-destroy.yml` tears the namespace down on PR close, on removal of the last preview label, or via manual dispatch — releasing Crossplane claims so the underlying AWS resources (S3 buckets, enclave nodegroups) are deprovisioned instead of leaking.
- The dedicated KMS reuses `zama-ai/kms`'s own deploy pipeline as-is.
- Docs: `ci/preview-env/README.md` (architecture / rationale) and `ci/preview-env/101-preview-env.md` (practical how-to).


ci/preview-env/host-chain/values-anvil-host-e2e.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: ghcr.io/foundry-rs/foundry
+  tag: v1.3.5
+network:
+  blockTime: "1"
+  host: "0.0.0.0"
+  chainId: "12345"
+  accounts: "120"
+  # Well-known Foundry/Hardhat test mnemonic. Deterministic accounts only —
+  # this cluster is ephemeral and never holds real funds.
+  mnemonic: "test test test test test test test test test test test junk"
```
eudelins-zama, line 13: We also use set `--slots-in-an-epoch 1` [here](https://github.com/zama-ai/fhevm/blob/main/test-suite/fhevm/docker-compose/host-node-docker-compose.yml#L21-L24), because kms-connector listen to finalized event on Ethereum.

This speed up finalization of the test setup, so maybe we want it there too?

fegmorte: Thank you @eudelins-zama I bumped the chart [here](https://github.com/zama-ai/fhevm/pull/3284) and add this config

.github/workflows/preview-env-deploy.yml
```diff
@@ -0,0 +1,1081 @@
...
+        # (coprocessor-redis-<i>). Image tags are shared across parties.
+        env:
+          COPROC_WALLETS_JSON: ${{ steps.coproc-wallets.outputs.wallets_json }}
+        run: |
+          for i in $(seq 1 "${NB_COPROCESSOR}"); do
+            privkey=$(jq -r --argjson party "${i}" '.[] | select(.party == $party) | .privateKey' <<<"${COPROC_WALLETS_JSON}")
+            if [[ -z "${privkey}" || "${privkey}" == "null" ]]; then
+              echo "::error::no derived coprocessor wallet for party ${i} in coproc_wallets_json=${COPROC_WALLETS_JSON}"; exit 1
+            fi
+            helm upgrade --install "coprocessor-${i}" "${COPROCESSOR_CHART}" --version "${COPROCESSOR_CHART_VERSION}" \
```
PanGan21, line 809: Ideally we should be able to run optionally two coprocessor stacks per operator with different releases sharing the same DB/S3. This is a good first step but probably it might require some changes in the future

.github/workflows/preview-env-deploy.yml
```diff
@@ -0,0 +1,1081 @@
...
+              --set-string "hostListenerShared.image.tag=$(jq -r .coprocessor_host_listener <<<"${TAGS_JSON}")" \
+              --set-string "hostListenerPollerShared.image.tag=$(jq -r .coprocessor_host_listener <<<"${TAGS_JSON}")" \
+              --set-string "hostListenerCatchupOnlyShared.image.tag=$(jq -r .coprocessor_host_listener <<<"${TAGS_JSON}")" \
+              --set-string "hostListenerConsumerShared.image.tag=$(jq -r .coprocessor_host_listener <<<"${TAGS_JSON}")" \
+              --set-string "snsWorker.image.tag=$(jq -r .coprocessor_sns_worker <<<"${TAGS_JSON}")" \
+              --set-string "tfheWorker.image.tag=$(jq -r .coprocessor_tfhe_worker <<<"${TAGS_JSON}")" \
+              --set-string "txSender.image.tag=$(jq -r .coprocessor_tx_sender <<<"${TAGS_JSON}")" \
+              --set-string "zkProofWorker.image.tag=$(jq -r .coprocessor_zkproof_worker <<<"${TAGS_JSON}")"
+          done
+          sleep 60
```
PanGan21, line 834: I am wondering if the sleep here could result to a race. Would it be better to use `--wait` with a timeout instead?

.github/workflows/preview-env-deploy.yml
PanGan21: No need to be done now, but it is for a future improvement

ci/preview-env/gateway-chain/values-anvil-gateway-e2e.yaml
```diff
@@ -0,0 +1,35 @@
...
+  chainId: "54321"
+  # 120 (not the default 10) to pre-fund preview-env-deploy.yml's "Derive KMS
+  # tx-sender wallets" step's HD indices 10..10+NB_KMS_CORE-1 (up to 13 parties)
+  # on top of every low-index account already used by the gateway-contracts/
+  # relayer values files - matches anvil-host's own headroom for the same reason.
+  accounts: "120"
+  # Same rationale as anvil-host: the gw-listener consumes *finalized* gateway
+  # events, so shrink the epoch to make finalization near-instant instead of
+  # lagging anvil's default ~2 epochs.
+  slotsInAnEpoch: "1"
```
eudelins-zama: Nit: the comment is not true, at least from the kms-connector PoV.
We only wait for events finalization on Ethereum, not on Gateway.
That being said, this is not a blocker...

.github/workflows/preview-env-deploy.yml
fegmorte: Will do that in the next pr

.github/workflows/preview-env-deploy.yml
fegmorte: the `--wait` only for the helm applied correctly but the pods can still be starting. We can use a kubectl command but I prefer just keeping things simple with a single sleep and will see

PanGan21 (dismissed):

ci/preview-env/gateway-chain/values-anvil-gateway-e2e.yaml
fegmorte: removed the comment 👍

eudelins-zama (approved):

eudelins-zama:
@mergifyio refresh

eudelins-zama:
@mergifyio queue

PR #3193, opened 2026-07-16
fix(host-contracts): allow pending status for epoch destruction

isaacdecoded:
## Summary

`destroyKmsEpoch()` only accepted `Active` epochs, which left one state unrecoverable: a pending same-set rotation whose activation votes diverged. This PR extends it to also destroy a `Pending` epoch whose context is `Active`, restoring a governance recovery path. No ABI or access-control changes.

## Background

Epochs activate via `confirmEpochActivation()`: every KMS node of the context submits EIP-712-attested results, and activation requires **unanimity on an identical result hash**. Confirmations are **one-shot per signer** (votes must not be mutable), so a single divergent result permanently blocks activation — the epoch is stuck in `Pending` forever. Neither destruction path could clear it: `destroyKmsEpoch()` required `Active`, and `destroyKmsContext()` reverts for the latest-active context — which is exactly the context a same-set rotation (`defineNewEpochForCurrentKmsContext()`) opens its pending epoch under.

Found during review of the epoch documentation ([tech-spec#494 discussion](https://github.com/zama-ai/tech-spec/pull/494#discussion_r3586121829)).

Closes https://github.com/zama-ai/fhevm-internal/issues/1786

## The change

`destroyKmsEpoch()` (still `onlyACLOwner`) accepts exactly two cases; everything else reverts (`InvalidKmsEpoch`, fails closed):

- **`Active` (non-latest)** — unchanged: retire a superseded epoch.
- **`Pending` under an `Active` context** — new: abort a stuck same-set rotation. Recovery: destroy, then re-trigger the rotation, which always mints a fresh epoch id (ids are never reused, so the destroyed epoch's leftover vote bookkeeping is unreachable).

Deliberately still rejected: the latest active epoch (destroying it would halt decryption), and the pending epoch of an in-flight context switch — that pair is settled by `destroyKmsContext()`, which clears the paired epoch itself.

## Tests

New tests reproduce the divergent-vote deadlock end-to-end (stuck epoch → abort → re-trigger → activate), cover the abort path, and confirm the context-switch pair is never stuck. Full `ProtocolConfig` suite: 118/118 passing.


host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -393,6 +393,8 @@ interface IProtocolConfig {
 
     /**
      * @notice Destroy a superseded (non-current) KMS epoch, preventing it from being used.
+     *         Also aborts a stuck Pending epoch of an Active context — a same-set rotation whose
```
jot2re:
```suggestion
     *         Also used to abort a stuck Pending epoch of an Active context — a same-set rotation whose
```

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -511,7 +513,19 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        // - Pending under an Active context: abort a stuck same-set rotation. Activation
+        //   confirmations are one-shot per signer, so a divergent vote can never reach unanimity
+        //   and the epoch would otherwise stay Pending forever — with its context still current,
+        //   neither destroyKmsContext() (reverts for the latest-active context) nor completion
+        //   could ever settle it. A pending context switch is settled by destroyKmsContext()
+        //   instead, which clears its paired epoch itself.
+        EpochState state = $.epochState[epochId];
+        bool destroyable = state == EpochState.Active ||
+            (state == EpochState.Pending && $.contextState[$.contextForEpoch[epochId]] == ContextState.Active);
+        if (!destroyable) {
```
obatirou: could you use early returns instead of this condition?
Also, do we agree that we do not need to destroy an epoch when the context is in `Created` or `Pending` state?

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -511,7 +513,19 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        // - Active: retire a superseded epoch so its old shares stop being served.
+        // - Pending under an Active context: abort a stuck same-set rotation. Activation
+        //   confirmations are one-shot per signer, so a divergent vote can never reach unanimity
+        //   and the epoch would otherwise stay Pending forever — with its context still current,
+        //   neither destroyKmsContext() (reverts for the latest-active context) nor completion
+        //   could ever settle it. A pending context switch is settled by destroyKmsContext()
+        //   instead, which clears its paired epoch itself.
+        EpochState state = $.epochState[epochId];
+        bool destroyable = state == EpochState.Active ||
+            (state == EpochState.Pending && $.contextState[$.contextForEpoch[epochId]] == ContextState.Active);
```
jot2re: Could it happen that the context is also pending and hence block any destruction?

jot2re (approved):
Thanks for taking care of this! Just a couple of comments

host-contracts/contracts/interfaces/IProtocolConfig.sol
isaacdecoded: Suggestion added [here](https://github.com/zama-ai/fhevm/pull/3193/changes/d2e31f104b172158f876749f1a30b03108076497#diff-46eb54d4359fd5be9ee85fb292f7a594d7f52c9429506e71b1a745f3075be0a7R396).

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: No, a Pending context is destroyable since it's never the latest-active one - liveness only requires exists-and-not-destroyed. Destroying it clears the paired pending epoch in the same call, so no combination is stuck. Added a [test for this exact case](https://github.com/zama-ai/fhevm/pull/3193/changes/d2e31f104b172158f876749f1a30b03108076497#diff-278c68bc3f6dea554e94044d5a81ccdf9c9faa2420e4cd13ca1dbbad34d2e6bcR1149).

host-contracts/contracts/ProtocolConfig.sol
isaacdecoded: Refactored [here](https://github.com/zama-ai/fhevm/pull/3193/changes/d2e31f104b172158f876749f1a30b03108076497#diff-ba77c9fc763d9a63be867533f531523aa9b6601802eb7fa378a41c9e1f43092fR527-R531). Let me know if it was what you had in mind.

> Also, do we agree that we do not need to destroy an epoch when the context is in Created or Pending state?

I think so, when the context is Pending/Created it means it's an in-flight context switch, and destroyKmsContext() is the settlement path - it's allowed for non-latest-active contexts and clears the paired epoch itself

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -511,7 +513,22 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
...
+        //   Pending/Created) is settled by destroyKmsContext() instead, which clears its paired
+        //   epoch itself.
+        EpochState state = $.epochState[epochId];
+        if (state == EpochState.Pending) {
+            if ($.contextState[$.contextForEpoch[epochId]] != ContextState.Active) {
+                revert InvalidKmsEpoch(epochId);
+            }
+        } else if (state != EpochState.Active) {
             revert InvalidKmsEpoch(epochId);
         }
```
obatirou:
```suggestion
        bool isEpochActive = state == EpochState.Active;
        bool isEpochPendingUnderActiveContext = state == EpochState.Pending &&
            $.contextState[$.contextForEpoch[epochId]] == ContextState.Active;
        if (!isEpochActive && !isEpochPendingUnderActiveContext) {
            revert InvalidKmsEpoch(epochId);
        }
```

host-contracts/test/protocolConfig/protocolConfig.t.sol
```diff
@@ -1061,19 +1061,114 @@ contract ProtocolConfigTest is HostContractsDeployerTestUtils {
...
+        emit IProtocolConfig.KmsEpochDestroyed(pendingEpochId);
+        vm.prank(owner);
+        protocolConfig.destroyKmsEpoch(pendingEpochId);
+
+        assertFalse(protocolConfig.isValidEpochForContext(contextId, pendingEpochId));
+        (, uint256 epochAfter) = protocolConfig.getCurrentKmsContextAndEpoch();
+        assertEq(epochAfter, activeEpochId);
+
+        // Governance can re-trigger the rotation and complete it.
+        uint256 retriedEpochId = EPOCH_COUNTER_BASE + 3;
```
obatirou, line 1075: it is not retried, it is a new one?

isaacdecoded: From the test case perspective, it's a "retried" key resharing flow. From the contract perspective, is a new epoch ID, yes. Feel free to suggest the change.

isaacdecoded:
@mergifyio queue

obatirou (approved):

obatirou:
> @Mergifyio queue

Thanks again for implementing the fix!
This is on release branch, I do not think mergify is necessary.
But we need a port on main

obatirou:
Please do not merge until @tawadaa approval

isaacdecoded:
https://github.com/Mergifyio dequeue

obatirou (approved):

PR #3208, opened 2026-07-16
fix(host-contracts): allow explicit subject rotation on durable supersede

Eikix:
Fixes zama-ai/fhevm-internal#1741.

## What

On `feature/solana`, a sender's second confidential transfer to a **different** recipient reverts with `PreviousStateMismatch` (6053): the transferred-amount lineage is per-sender, its ACL is built per-call (including the recipient), and the host froze membership through eval binding.

This PR keeps that freeze as the default and adds **explicit subject rotation at durable-output supersede**:

- `previous_handle` / `previous_subjects` must still match the stored state exactly (unchanged — this is what lets the relayer/indexers reconstruct MMR leaves from instruction data alone).
- `output_subjects` may now differ from the stored set. The outgoing handle's historical leaves are sealed for the **old** audience first, then the new set is installed. Net semantics: each handle has exactly one ACL, fixed at birth, sealed at supersession (EVM parity — per-handle ACL epochs).
- Every rotation-**added** subject passes the grant deny-list, like `allow_subjects` (kept subjects are not re-checked, consistent with existing grant semantics). The `fhe_eval` preflight pass learns to account for the added subjects' deny-record witnesses so the all-accounts-consumed guard holds with the deny-list enabled.
- Authorization unchanged: the app-account authority signer gate on durable outputs.

The confidential-token transfer then works for arbitrary recipient sequences with **no app-side change**. Relayer replay rotates its reconstructed state at the identical point in the leaf sequence, and its idempotent-recovery path gains an arm for the fhe_eval supersede shape.

## Tests

- Regression: Alice→Bob then Alice→Charlie now succeeds; asserts exact post-rotation subjects and that the first receipt's sealed leaves carry the old audience `{alice, bob, compute}`.
- Rotate-back (Alice→Bob→Charlie→Bob), audience shrink (host-level 3→2), self-transfer no-op, deny-list **enabled** rotation (positive and denied-added-subject negative), relayer replay of rotation+make_public in one output, relayer ingest idempotent recovery.
- Full suites green: zama-host unit 83, zama-fhe 27, relayer replay 12 + ingest 12, host_mollusk 79, token_mollusk 39.

## Cost snapshots

Regenerated under the pinned 2.1.0 toolchain; `compute_units` only, no shape change (accounts / instruction bytes / CPI fields identical):

| profile | before | after |
|---|---|---|
| fhe_eval/three_op_frame | 61,398 | 61,438 |
| fhe_eval/max_op_frame | 151,083 | 151,136 |
| confidential_transfer/direct | 332,592 | 333,137 |
| confidential_transfer/steady_state | 350,061 | 350,816 |
| initialize_token_account | 65,180 | 65,268 |

## Open questions for reviewers

1. **Subject continuity.** The removed equality check was also a structural guarantee that a supersede could never change a lineage's audience. With rotation, the host no longer guarantees that any particular subject (owner, compute signer) survives — audience continuity becomes the app's responsibility. All current call sites keep owner+compute invariant (audited), but a host-level "must-keep" rule (e.g. the superseding authority must remain in `output_subjects`) would restore defense-in-depth. Deliberately not implemented here — semantics call for the host team.
2. **Deny-after-grant.** A subject deny-listed after being granted keeps standing membership across rotations (deny gates grants, not membership; removal stays `remove_subject`). This matches existing `allow_subjects` semantics and is unchanged here — flagged so it's a conscious choice.
3. Create-path subjects remain un-deny-checked (pre-existing asymmetry, unchanged): a denied subject can still enter a lineage at birth or via `allow_subjects`' member set; rotation is now the strictest of the three grant routes.

Design discussion and alternatives considered (per-pair lineages, per-party split, born-sealed receipts, CPI reconciliation) are in fhevm-internal#1741.

### DIFF
--- a/coprocessor/fhevm-engine/host-listener/idl/zama_host.json
+++ b/coprocessor/fhevm-engine/host-listener/idl/zama_host.json
@@ -2500,7 +2500,9 @@
                 "name": "output_subjects",
                 "docs": [
                   "Subjects on the output lineage. On create these are the initial",
-                  "subjects; on supersede they must equal the stored subjects exactly."
+                  "subjects; on supersede they become the new audience, which may rotate",
+                  "away from the stored set (the outgoing audience is sealed into",
+                  "historical leaves first; added subjects pass the grant deny-list)."
                 ],
                 "type": {
                   "vec": {
--- a/relayer/src/solana_proof/ingest.rs
+++ b/relayer/src/solana_proof/ingest.rs
@@ -173,6 +173,32 @@ fn recovered_already_applied_event(
             )])
         }
         (
+            DecodedInstruction::FheEvalUpdateEncryptedValue {
+                encrypted_value,
+                previous_handle,
+                previous_subjects,
+                output_subjects,
+                make_public_handle,
+            },
+            ReplayError::PreviousStateMismatch(error_lineage),
+        ) if error_lineage == encrypted_value
+            && state.current_handle == *make_public_handle
+            && state.subjects.as_slice() == output_subjects.as_slice() =>
+        {
+            // Already applied: the lineage is at this supersede's post-state (a rotation moved
+            // subjects to `output_subjects`, so a re-apply trips PreviousStateMismatch). Replay
+            // the same leaves the fresh apply would have — the outgoing audience sealed first,
+            // then the born-public leaf if any — mirroring the raw UpdateEncryptedValue arm.
+            let mut events = vec![LineageEvent::handle_superseded(
+                *previous_handle,
+                previous_subjects,
+            )];
+            if let Some(handle) = make_public_handle {
+                events.push(LineageEvent::MarkedPublic { handle: *handle });
+            }
+            Some(events)
+        }
+        (
             DecodedInstruction::RemoveSubject {
                 encrypted_value,
                 subject,
@@ -432,6 +458,35 @@ mod tests {
                 previous_subjects,
             },
         )
+    }
+
+    #[test]
+    fn recovers_already_applied_fhe_eval_rotation_idempotently() {
+        let ev = pk(0x40);
+        let owner = pk(0x30);
+        let old_recipient = pk(0x31);
+        let new_recipient = pk(0x32);
+        // State already advanced to the post-rotation audience by a prior application.
+        let mut state = Some(LineageReplayState {
+            current_handle: None,
+            subjects: vec![owner, new_recipient],
+        });
+        let instruction = DecodedInstruction::FheEvalUpdateEncryptedValue {
+            encrypted_value: ev,
+            previous_handle: pk(0x10),
+            previous_subjects: vec![owner, old_recipient],
+            output_subjects: vec![owner, new_recipient],
+            make_public_handle: None,
+        };
+
+        let events = apply_or_recover_instruction(&mut state, &instruction).unwrap();
+
+        // Replays the outgoing-audience leaves without double-advancing the state.
+        assert_eq!(
+            events,
+            vec![LineageEvent::handle_superseded(pk(0x10), &[owner, old_recipient])]
+        );
+        assert_eq!(state.unwrap().subjects, vec![owner, new_recipient]);
     }
 
     fn make_public_ix(program_id: [u8; 32], lineage: [u8; 32], handle: [u8; 32]) -> RawInstruction {
--- a/relayer/src/solana_proof/ingest.rs
+++ b/relayer/src/solana_proof/ingest.rs
@@ -484,7 +484,10 @@ mod tests {
         // Replays the outgoing-audience leaves without double-advancing the state.
         assert_eq!(
             events,
-            vec![LineageEvent::handle_superseded(pk(0x10), &[owner, old_recipient])]
+            vec![LineageEvent::handle_superseded(
+                pk(0x10),
+                &[owner, old_recipient]
+            )]
         );
         assert_eq!(state.unwrap().subjects, vec![owner, new_recipient]);
     }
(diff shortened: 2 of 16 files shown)


Eikix:
> [!NOTE]
> Code review generated by Claude Code, requested by @Eikix. Reviewed at head `bd7c21c8db` (CI fully green).

## Overview

The PR fixes the `PreviousStateMismatch` (6053) revert on a sender's second confidential transfer to a different recipient. It removes the host-side requirement that a durable-output supersede's `output_subjects` equal the stored subject set, replacing it with explicit audience rotation: the outgoing audience is sealed into historical MMR leaves first, then the new set replaces current membership, and every rotation-*added* subject must clear the grant deny-list. The relayer's replay and idempotent-recovery paths learn the new shape, and the preflight all-accounts-consumed accounting covers the added subjects' deny witnesses.

## Correctness — verified, looks sound

Traced the full path on the PR head; the load-bearing invariants hold:

- **Sealing order is correct.** `supersede_current_handle` appends leaves from `value.subjects` and is called *before* `value.subjects = output_subjects` in `bind_eval_output`, so historical leaves always carry the pre-rotation audience. The born-public leaf lands after, matching DD-036 order and the relayer's reconstruction.
- **Rotated sets stay well-formed.** `assert_output_acl_metadata` runs at the top of `bind_eval_output` for create *and* supersede, so an empty set, a set over `MAX_ACL_SUBJECTS`, a default pubkey, and duplicate subjects are all rejected on rotation too. Good that the new unit test pins the empty/over-cap cases.
- **Deny-list gate is consistent across all three passes.** Admission checks `check_rotation_grants_not_denied` against stored subjects (no marking), execution checks again with marking, and preflight marks the added subjects' deny records computed from instruction data. The three added-sets agree because `previous_subjects` must equal stored subjects (validated *before* the deny check in both admission and bind), and a lying `previous_subjects` dies with `PreviousStateMismatch` before any deny record is consumed — so trusting instruction data in preflight is safe, as the comment claims.
- **Duplicate added-subject across steps is harmless.** Remaining accounts are unique (`assert_unique_remaining_accounts`), `mark_deny_record`/`deny_subject_record_for` locate by address and set `used` idempotently, so two steps adding the same subject share one witness cleanly.
- **Relayer replay mirrors on-chain order** (seal old audience → rotate → optional MarkedPublic), and the new ingest recovery arm's guard (`current_handle == make_public_handle && subjects == output_subjects`) is exactly as tight as the pre-existing raw-`UpdateEncryptedValue` arm, gated behind the seen-signatures dedup.

## Findings

**1. Stale comments + a subtle in-frame semantics change in `resolve_durable_operand` (fhe_eval.rs, `superseded_in_frame` field doc and the operand-resolution comment) — worth fixing.**
Both still say "supersession never edits membership, so only the current-handle equality is exempted here." That's now false — this PR is precisely about supersession editing membership. And it's not just the comment: that branch re-reads the account and checks `value.has_subject(self.subject)` against *post-rotation* membership. So if a frame rotates the compute subject out of a lineage and a later step in the same frame uses the superseded handle as an operand, execution now fails with `SubjectNotAllowed` even though admission authorized it against frame-entry state — contradicting the stated EVM parity ("a handle stays usable as a value within the transaction that rotated it"). No soundness hole (admission still gates against frame-entry state, so rotation can't *escalate* an operand authorization; the tx just reverts atomically), but the intended semantics should be decided explicitly, the comments updated, and ideally a mollusk test pinning whichever behavior is chosen.

**2. Pre-existing (not introduced here, but adjacent): a *non-rotating* fhe_eval supersede re-apply isn't caught by the recovery arm — it silently double-applies.**
After a non-public supersede, `current_handle` is `None`, and `validate_previous_state` skips the handle check when `current_handle` is `None`. So re-applying a supersede whose `previous_subjects` equal current subjects *passes* validation and re-appends the same leaves, rather than tripping `PreviousStateMismatch` into the recovery path. The new arm only ever fires for rotations (where subjects moved). The seen-signatures dedup makes this hard to hit in practice, but since this PR extends the recovery machinery it's a good moment for the relayer owners to confirm that crash window is acceptable.

**3. Nit: `rotation_added_subjects` (zama-fhe lib.rs) and the host compute the added set from different baselines** — the client uses instruction-data `previous_subjects`, the host uses stored `value.subjects`. They're provably equal given the validation order, so no action needed; a one-line comment on the client helper noting why this is safe would help the next reader.

## Code quality, tests, docs

- Code follows the existing conventions well: the deny check reuses `deny_subject_record_for`/`check_grant_not_denied_info`, the new `fhe_eval_step_output` match-all helper is clean, and the token program's rename to `validate_deny_subject_records_for_grant_subjects` accurately reflects the widened contract.
- **Test coverage is excellent** — the exact regression (Alice→Bob→Charlie), rotate-back with full MMR peak verification, audience shrink with the no-realloc-shrink assertion, self-transfer no-op, deny-list-enabled rotation both positive and negative, relayer replay of rotation+make_public, and ingest idempotent recovery. The negative deny test also asserting the error is `AclSubjectDenied` rather than an unconsumed-account error is a nice touch.
- Docs (DD amendment, MMR_ACL_MVP, IDL, state doc comments) are updated consistently; cost deltas (~0.03–0.2%) are negligible with no instruction-shape change.

## On the PR's open questions

1. **Subject continuity:** the host-level "superseding authority must remain in `output_subjects`" guard seems worth adding as defense-in-depth — it's one cheap check, restores a structural invariant (an authority can't accidentally brick its own lineage), and doesn't constrain the confidential-token use case. Owner/compute continuity beyond that is reasonably the app's job.
2. **Deny-after-grant** keeping standing membership matches `allow_subjects`; consistent, fine as a conscious choice.
3. Create-path asymmetry is pre-existing; rotation being the strictest grant route is the right direction.

## Verdict

No blocking issues. The core mechanism (seal-then-rotate, deny-gated additions, exact previous-state pinning) is correct and well-tested across host, token, and relayer. Finding 1 (stale comments + explicit decision on in-frame operand use after a self-removing rotation) would be good to address before merge; the rest are notes.

PR #3209, opened 2026-07-16
feat(solana-proof-service): rehome Yellowstone block source

Eikix:
## Summary

- Scaffold a top-level `solana-proof-service/` Cargo workspace (sibling of `relayer/` / `listener/`) with a lean `solana-proof-source` lib crate.
- Rehome the confirmed Yellowstone completed-block source (+ 13 unit tests) from #3207 (`b53a762ea`) into that crate, with a field-only `RawInstruction` type so the source does not pull in the full host decoder.
- Remove the temporary Yellowstone adapter and unused `tonic` / `yellowstone-grpc-proto` deps from `fhevm/relayer`, so durable proof-source code is no longer embedded in the retiring relayer.

**Out of scope (later slices):** SQL store, HTTP API, recovery, multi-replica, readiness.

**Provenance:** #3207 (`elias/solana-relayer-yellowstone-source` @ `b53a762ea`) remains the source of truth for the original adapter; this PR only rehomes it.

## Test plan

- [x] `NO_DNA=1 cargo test -p solana-proof-source` — 13 Yellowstone tests pass
- [x] `NO_DNA=1 cargo fmt --check` / `cargo clippy -- -D warnings` in `solana-proof-service/`
- [x] `NO_DNA=1 cargo test -p fhevm-relayer --lib solana_proof` — 64 remaining tests pass (77 − 13 Yellowstone)


Eikix:
Addressed review findings:
1. **P1 TLS** — `tonic` now enables `tls-webpki-roots` so `https://` Yellowstone endpoints work.
2. **P2 token leak** — custom `Debug` for `YellowstoneSourceConfig` redacts `x_token` (covers nested `YellowstoneBlockSource` Debug too), with a unit test.

PR #3211, opened 2026-07-16
feat(solana-proof-service): atomic PostgreSQL store and sequential runner

Eikix:
## Summary
- Adds service-owned Postgres schema (`solana_proof_progress/blocks/transactions/lineages/leaves`) and `SqlProofStore` with atomic `apply_completed_block`, plus checkpoint / proof snapshot / integrity status APIs (Bootstrap A).
- Mines decode+replay into the service, reduces `CompletedBlock` → deterministic leaves/peaks in one SQL transaction, and runs a cancellation-safe sequential Yellowstone ingest loop that does not own persistence.
- Commits compile-checked SQLx offline metadata; unit coverage always runs; real Postgres integration tests via `make test-db` (`DATABASE_URL`).

## Test plan
- [x] `cd solana-proof-service && make test` (unit + ignored DB tests)
- [x] `cd solana-proof-service && make clippy`
- [x] `DATABASE_URL=postgres://… make test-db` (empty block, vote/fail identities, leaf order, exact replay no-op, conflict halt, rollback, snapshot isolation, incomplete bootstrap, post-bootstrap birth, unknown pre-bootstrap, MMR reconstruction)
- [ ] CI path-filtered solana-proof-service workflow on this PR

Refs: fhevm-internal #1682 (PR2 store slice)


Eikix:
Addressed the two findings:

1. **P1 subject rotation** — `FheEvalUpdateEncryptedValue` no longer requires `output_subjects == previous subjects`. Historical leaves still seal against the old audience; `state.subjects` is then replaced with `output_subjects`.
2. **P2 backoff** — reconnect backoff is no longer reset on successful subscribe; it resets only after `Applied` / `AlreadyApplied` stream progress.

Added `fhe_eval_subject_rotation_seals_old_audience_then_adopts_output_subjects`. Lib tests: 35 passed.

PR #3212, opened 2026-07-16
fix(host-contracts): meter HCU on the signed compute subject

Eikix:
Fixes zama-ai/fhevm-internal#1708. Follow-up tracked in zama-ai/fhevm-internal#1744.

## What

`fhe_eval` metered the per-app-per-slot HCU block cap on a caller-supplied `hcu_authority` signer, with the meter and trust PDAs derived from it alone. Nothing bound that signer to anything, so a caller could present a fresh `hcu_authority` per call and mint a fresh zero-usage meter each time — bypassing a finite `hcu_block_cap_per_app`. (Dormant today only because the cap ships `u64::MAX`.)

This PR **deletes `hcu_authority` entirely** and meters the already-mandatory, already-signed `compute_subject` — the same identity that must clear durable-input ACL admission (the msg.sender analog). The meter and trust PDAs re-key on `compute_subject`. Substituting it is not free: for any frame consuming a durable or verified input, `compute_subject` is pinned by ACL membership / attestation, so the rotation bypass is closed by construction. For the confidential-token program the subject is the CPI-signed `["fhe-compute", mint]` PDA (1:1 with the mint), so per-mint metering is unchanged and the token program is fully covered.

Bonus: removing the account drops one account-meta from every eval and each of the six token instructions.

## Scope of the guarantee (see DD-039)

Closed: the confidential-token program, and any frame consuming a durable/verified input.
**Not** closed: input-free frames (`Rand`/`TrivialEncrypt`/scalar-only with transient output) from a *direct* caller, where `compute_subject` is an unconstrained signer. That residual is latent (needs a finite cap to be set) and does not affect the token program. Tracked as fhevm-internal#1744; a reviewer confirmed there is no clean minimal fix (an input-free frame has no stable identity to meter) and no cross-mint griefing vector.

## Blast radius (intended ABI break)

`hcu_authority` removed from `fhe_eval` (9 named accounts now) and all six confidential-token instructions + the deposit-app CPI. Regenerated IDLs (`zama_host.json`, `confidential_token.json`, `solana_abi_golden.json`) and resynced the host-listener copies. Downstream index constants updated: host-listener and relayer `FHE_EVAL_REMAINING_BASE` (both 10 → 9); js-sdk `confidential_transfer` client regenerated (18 → 17 accounts).

## Tests

- New regression `mollusk_fhe_eval_same_compute_subject_accumulates_across_varied_accounts_and_trips_cap`: two frames in one slot, same `compute_subject`, different `payer` AND `app_account_authority`, cap sized for one frame → the second accumulates onto the same meter and trips `HcuBlockLimitExceeded`. Proves no other caller-controlled account yields a fresh meter.
- Adapted the HCU fixture/tests to model `compute_subject` as the metered identity; removed the obsolete missing/non-canonical-`hcu_authority` tests.
- Suites: zama-host 80, zama-fhe 27, host_mollusk 77, token_mollusk 33, operator conformance 11, relayer solana_proof 65, js-sdk solana 66 — all green.

## Cost snapshots (regenerated, pinned 2.1.0)

| profile | compute_units | unique_accounts |
|---|---|---|
| host `fhe_eval/three_op_frame` | 61398 → 61259 | 9 → 9 |
| host `fhe_eval/max_op_frame` | 151083 → 150944 | 9 → 9 |
| token `confidential_transfer/direct` | 332592 → 329820 | 15 → 14 |
| token `confidential_transfer/steady_state` | 350061 → 347289 | 15 → 14 |
| token `initialize_token_account` | 65180 → 62273 | 12 → 11 |

Token profiles show the genuine −1 account and ~2.8k CU reduction; host profiles' account count is flat because the fixture now models `compute_subject` as a distinct signer (the realistic shape).


Eikix:
## Review

### Overview

Removes the standalone `hcu_authority` signer from `fhe_eval` and keys the HCU block-cap meter and trust PDAs on the already-mandatory, already-signed `compute_subject`. This closes a real metering bypass: the old `hcu_authority` was bound to nothing, so a direct caller could rotate a fresh keypair per call and get a fresh per-slot meter, defeating any finite `hcu_block_cap_per_app`. The change ripples through the host program, the `zama-fhe` CPI crate, all six confidential-token instructions, the deposit app, regenerated IDLs, host-listener/relayer account-index constants (10 → 9), and the js-sdk client.

### Correctness — looks right

- **The security fix is sound.** `compute_subject` is `Signer<'info>` in `FheEval`, and `block_cap.rs` `check`/`charge` both key off `ctx.accounts.compute_subject.key()`. For any frame consuming durable/verified inputs, the subject is pinned by ACL admission/attestation, so rotation costs input access; for the token program it's the CPI-signed `["fhe-compute", mint]` PDA, so per-mint metering is preserved with one fewer account.
- **Index bookkeeping is consistent.** `FHE_EVAL_REMAINING_BASE` is 9 in both the host-listener and the relayer, both sides' comments spell out the same 9-account layout, and the tests that pin the base (`durable_output_as_sole_remaining_account_resolves`, the relayer layout fixtures) were updated in lockstep. Counting the enumeration by hand checks out: payer, compute_subject, app_account_authority, host_config, system_program, hcu_block_meter, hcu_trusted_app_record, event_authority, program = 9. ✓
- **The residual gap is honestly scoped.** DD-039 documents that input-free frames (`Rand`/`TrivialEncrypt`/scalar-only, transient output) from a direct caller still have an unconstrained subject; it's dormant (cap ships `u64::MAX`) and tracked in fhevm-internal#1744. Good practice to write this down rather than overclaim.

### Test coverage — strong

- The new regression test is well-designed: same `compute_subject`, **varied** `payer` and `app_account_authority` each binding fresh output lineages, cap sized for exactly one frame → second frame trips `HcuBlockLimitExceeded`, meter unchanged after the breach, no output created. That directly falsifies "some other caller-controlled account yields a fresh meter."
- `mollusk_fhe_eval_unsigned_compute_subject_is_rejected` preserves the meter-drain/bypass-theft guard from the old design.
- The fixture now deliberately makes `compute_subject` distinct from `app_account_authority` — good, that keeps the "meter never keys on the output authority" assertions meaningful (and explains the flat 9→9 host account count in the cost snapshots).
- One small gap worth a thought: nothing in this PR's tests directly demonstrates the load-bearing claim that *rotating* `compute_subject` on a durable-input frame fails ACL admission (the "closed by construction" leg of the argument). If an existing ACL-admission test already rejects a non-member subject, fine — otherwise a one-test companion to the regression test would pin the other half of the invariant.

### Minor issues (nits, none blocking)

1. **Stale comments left behind:**
   - `coprocessor/fhevm-engine/tfhe-worker/src/tests/solana_poc.rs` (`transfer_ix`): "…The mint's HCU authority is mandatory." — the account it describes was deleted on the next line.
   - `solana/scripts/e2e/live-client/main.rs`, two sites (`trivial_encrypt_eval_with_label`, `fhe_eval_verified_input_add`): "The mandatory HCU authority is the payer itself for this wallet-driven PoC leg." — same.
2. **Dead error variant:** `ConfidentialTokenError::HcuAuthorityMismatch` still exists in `errors.rs` (its only construction site and test were removed). If it's being kept to avoid renumbering Anchor error codes, a one-line `/// retired:` note would prevent someone reusing it; on a pre-release feature branch it could also just be deleted.
3. **Decoder/version skew:** `FHE_EVAL_REMAINING_BASE` is a bare constant with no tie to the deployed program version — a listener/relayer at this commit replaying *pre-upgrade* transactions would resolve the wrong account (off by one into the remaining accounts). On `feature/solana` pre-launch this is presumably fine with a coordinated redeploy + reindex from the upgrade slot, but it's worth confirming that's the plan since the failure mode is silent misattribution, not an error.

### Style / conventions

- Doc comments were updated consistently across all six token instructions, the CPI crate, and both state accounts; DD-039 is a genuinely good design record (states the rejected Option B and why).
- js-sdk changes are mostly regenerated code plus a clean manual edit in `confidentialTransfer.ts`; the generated-file churn matches the IDL change exactly.

### Verdict

The core change is a correct, well-tested fix for a real (if dormant) metering bypass, with the ABI break intentional and propagated everywhere it needs to go. Only nits: the three stale comments, the dead error variant, and a deployment-ordering confirmation for the decoder constant.

*(Review generated by Claude Code at Elias's request.)*

PR #3182, opened 2026-07-15
ci(common): bump pinned forge version from v1.3.1 to v1.7.1

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/1669

PR #3217, opened 2026-07-17
chore(kms-connector): alloy and rust bump

eudelins-zama:
### What

Bumps `alloy` to `2.2.0` and the KMS connector Rust toolchain to 1.97.1.

- kms-connector: bump alloy =1.0.38 → =2.1.1, Rust 1.92.0 → 1.97.1.
- shared crates & rust bindings: relax the alloy requirement to allow both `1.x` and `2.x`, so parent crates can pick their version.
- tx-sender / provider: simplify provider setup against the new alloy API — the utils/provider.rs helper shrinks significantly (~390 lines removed) and the recursion-limit bump is no longer needed.
- copro & relayer lockfiles updated.
  - they were outdated because of this [PR](https://github.com/zama-ai/fhevm/pull/3182), and I wanted to make sure crates update were not breaking anything
  - confirmed thanks to the CI :slightly_smiling_face: 

###  Notes

Housekeeping only, no functional changes intended

### DIFF
--- a/kms-connector/crates/utils/src/provider.rs
+++ b/kms-connector/crates/utils/src/provider.rs
@@ -1,35 +1,14 @@
 use alloy::{
-    consensus::TrieAccount,
-    eips::{BlockId, BlockNumberOrTag, Encodable2718},
-    network::{Ethereum, Network, TransactionBuilder},
-    primitives::{
-        Address, B256, BlockHash, BlockNumber, Bytes, StorageKey, StorageValue, TxHash, U64, U128,
-        U256,
-    },
+    network::{Network, TransactionBuilder},
+    primitives::{Address, Bytes},
     providers::{
-        EthCall, EthCallMany, EthGetBlock, FilterPollerBuilder, GetSubscription,
-        PendingTransaction, PendingTransactionBuilder, PendingTransactionConfig,
-        PendingTransactionError, Provider, ProviderCall, RootProvider, RpcWithBlock, SendableTx,
-        fillers::{
-            BlobGasFiller, CachedNonceManager, FillProvider, GasFiller, JoinFill, NonceManager,
-            TxFiller,
-        },
+        PendingTransactionBuilder, Provider, RootProvider, SendableTx,
+        fillers::{BlobGasFiller, CachedNonceManager, GasFiller, JoinFill, NonceManager},
     },
-    rpc::{
-        client::NoParams,
-        types::{
-            AccessListResult, Bundle, EIP1186AccountProofResponse, EthCallResponse, FeeHistory,
-            Filter, FilterChanges, Index, Log, SyncStatus, TransactionReceipt, TransactionRequest,
-            erc4337::TransactionConditional,
-            pubsub::{Params, SubscriptionKind},
-            simulate::{SimulatePayload, SimulatedBlock},
-        },
-    },
-    transports::{TransportError, TransportResult},
+    transports::TransportResult,
 };
 use futures::lock::Mutex;
-use serde_json::value::RawValue;
-use std::{borrow::Cow, sync::Arc};
+use std::sync::Arc;
 
 pub type FillersWithoutNonceManagement = JoinFill<GasFiller, BlobGasFiller>;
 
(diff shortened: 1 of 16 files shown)


eudelins-zama:
@claude /pr-review

jot2re (dismissed):
LGTM! But might also be good to have someone with greater knowledge of the provider to have a look.

jot2re (commented):
The bot came with a couple of other small findings:

1. *Gateway send path changed abstraction*. The gateway sender's send_transaction_sync ([gateway.rs:172](vscode-webview://1hujdn068sh756d5jkadu9mjrecdgfcagnaais0oaq1ojqi9cr7u/kms-connector/crates/tx-sender/src/core/gateway.rs#L172)) previously hand-built an envelope and issued a raw eth_sendRawTransactionSync request; it now delegates to alloy 2.x's native send_transaction_sync. Intent is equivalent and the old code already required eth_sendRawTransactionSync node support, so no expected regression — but since the RPC abstraction changed under it, this path should be exercised end-to-end against the real gateway node before release, not just via mocks. This is the one change I couldn't fully verify from static analysis.

2. *alloy-primitives pin asymmetry*. In [shared/ciphertext-attestation/Cargo.toml](vscode-webview://1hujdn068sh756d5jkadu9mjrecdgfcagnaais0oaq1ojqi9cr7u/shared/ciphertext-attestation/Cargo.toml), alloy-primitives stays "1" while alloy-signer is widened to ">=1, <3". Consistent today (alloy 2.1.1 → primitives 1.6.0), but if a future alloy 2.x bumps its internal primitives to 2.x you'd get two incompatible Address/B256 types. Consider widening alloy-primitives to ">=1, <3" too for symmetry, or leaving a comment noting the coupling.

eudelins-zama:
> Gateway send path changed abstraction. The gateway sender's send_transaction_sync (gateway.rs:172) previously hand-built an envelope and issued a raw eth_sendRawTransactionSync request; it now delegates to alloy 2.x's native send_transaction_sync. Intent is equivalent and the old code already required eth_sendRawTransactionSync node support, so no expected regression — but since the RPC abstraction changed under it, this path should be exercised end-to-end against the real gateway node before release, not just via mocks. This is the one change I couldn't fully verify from static analysis.

Indeed but was done this way previously **only** because it was not supported by `alloy`. Current behavior is what we wanted to do initially, so now that we can do it we should :slightly_smiling_face: .
Only caveat is that a bug there would be caught only at devnet level, but I'm pretty confident on it, and worst case the fix would be trivial without any major impact as would be caught at devnet level, not mainnet

> alloy-primitives pin asymmetry. In shared/ciphertext-attestation/Cargo.toml, alloy-primitives stays "1" while alloy-signer is widened to ">=1, <3". Consistent today (alloy 2.1.1 → primitives 1.6.0), but if a future alloy 2.x bumps its internal primitives to 2.x you'd get two incompatible Address/B256 types. Consider widening alloy-primitives to ">=1, <3" too for symmetry, or leaving a comment noting the coupling.

Won't fix as consistent today, and not known if alloy-primitives will ever have a 2.x bump

kms-connector/crates/utils/src/provider.rs
```diff
@@ -109,360 +43,114 @@ where
...
+    ) -> TransportResult<N::ReceiptResponse> {
         let nonce = self
             .nonce_manager
             .lock()
             .await
             .get_next_nonce(&self.inner, self.signer_address)
             .await?;
         tx.set_nonce(nonce);
-        let res = self.inner.send_transaction(tx).await;
+        let res = self.inner.send_transaction_sync(tx).await;
```
dvdplm, line 76: Was this always sync behind the scenes and they renamed the method?

dvdplm (dismissed):
As a nightly user I *really* appreciate the upgrade to get rid of the recursion issue.

The code looks sane to me, but I do have one concern: why do we have to be flexible with the alloy version? I think it's a little bit dangerous and I'd like to understand why there is no other way (e.g. consuming users could stay on a previous version until they can upgrade?).

I'm curious: did removing that extra generic parameter have any impact on compilation speed?

kms-connector/crates/utils/src/provider.rs
eudelins-zama: No, before `send_transaction_sync` was not part of the `Provider` trait, and we had our manual impl of `send_transaction_sync`.

Now it's there so we are moving (and adapting lightly) our impl here.

Btw, note that we are using:
- `send_transaction_sync` for Gateway tx (as fast as possible for decryption)
- `send_transaction` (not sync) for Ethereum tx (slow protocol management op, because `_sync` is not supported by all RPC provider on Ethereum)

antoniupop (dismissed):

zmalatrax (dismissed):

eudelins-zama:
@mergifyio queue

eudelins-zama:
@mergifyio queue

jot2re (approved):
LGTM

zmalatrax (approved):

antoniupop (approved):

eudelins-zama:
@mergifyio queue

manoranjith (approved):
LGTM.

PR #3236, opened 2026-07-17
feat(coprocessor): support rerunning the upgrade process

PanGan21:
### Summary

A blue-green upgrade whose window never reaches unanimity previously stalled forever in `DryRunStarted`. 
This PR handles the failure path: the consensus-detector emits `event_unanimity_consensus_timeout` when a track is still un-anchored `commitment_timeout` after `end_block`, and the upgrade-controller rolls the dry-run back so the upgrade can be re-proposed.

The rollback:
-  flips the GCS row to `PAUSED/failed` 
- clears the consensus latches and `gw_dry_run_started` 
- drops and recreates the `gcs-<version>` schema as empty

It only applies while still dry-running. A late or duplicate timeout is a no-op and can never undo a committed cutover.

### Recovering from missed signals

The controller no longer depends on catching a notification at the exact moment it fires. The detector keeps re sending its consensus and timeout signals every pass until the window closes, and the controller re-reads the upgrade row on startup and on a periodic tick to pick up wherever it left off. So if the controller restarts (or misses a signal for any reason) mid-upgrade, it still finishes the cutover or the rollback  instead of stalling. Re-running the cutover is safe: it re-checks the row and does nothing if the upgrade has already gone through.

### Fencing writes against the reset

Dropping and recreating the schema races with the workers still writing to it: a worker mid operation or one reconnecting after a database drop, could finish an old write into the freshly reset schema, poisoning the next attempt and, at cutover, leaking into live data.

Two changes close the race:

- The rollback runs under an exclusive advisory lock and does the pause, schema reset, and notify in one transaction. - Every green write takes the shared form of the same lock, so a write either commits before the reset (and is dropped by it) or runs after it, never during.
- Each write also re checks whether the dry-run is paused before it runs. Raw chain ingestion keeps going, so the stack can rerun without a restart. The other workers wait for the next attempt.

Related: https://github.com/zama-ai/fhevm-internal/issues/1528

### DIFF
--- a/coprocessor/fhevm-engine/consensus-detector/src/lib.rs
+++ b/coprocessor/fhevm-engine/consensus-detector/src/lib.rs
@@ -19,7 +19,7 @@
 //! each pass only re-requests operators it hasn't heard from — a slow operator
 //! simply fills in on a later tick. The FIRST block on a track whose blob is
 //! unanimous across all operators anchors that track: the service stops polling
-//! it and emits `unanimity_consensus(chain_id, block_height, block_hash)` — and
+//! it and emits a proposal-scoped unanimity event — and
 //! re-emits it each pass until the window closes, so a controller that missed the
 //! NOTIFY still latches. One unanimous non-trivial block is a sufficient anchor
 //! because a determinism-breaking upgrade is systematic — any such block reveals
@@ -138,6 +138,7 @@ pub struct NewBlockPayload {
 /// Payload emitted on `unanimity_consensus` / `unanimity_consensus_timeout`.
 #[derive(Debug, Clone, Serialize)]
 struct UnanimityPayload<'a> {
+    proposal_id: &'a [u8],
     chain_id: i64,
     block_height: i64,
     block_hash: &'a str,
@@ -238,28 +239,34 @@ fn all_slots_filled(slots: &[Option<Vec<u8>>]) -> bool {
     !slots.is_empty() && slots.iter().all(Option::is_some)
 }
 
-/// Returns `(start_block, end_block)` for the GCS dry-run when it's active.
+/// Returns `(proposal_id, start_block, end_block)` for the active GCS dry-run.
 /// `None` otherwise. Scoped to `stack_role = 'GCS'` because BCS also stays
 /// `status='in_progress'` during the upgrade and doesn't own the replay window.
-pub(crate) async fn active_upgrade_window<'e, E>(executor: E) -> Result<Option<(i64, i64)>, Error>
+pub(crate) async fn active_upgrade_window<'e, E>(
+    executor: E,
+) -> Result<Option<(Vec<u8>, i64, i64)>, Error>
 where
     E: Executor<'e, Database = Postgres>,
 {
-    let row: Option<(String, Option<i64>, Option<i64>)> = sqlx::query_as(
-        "SELECT state, start_block, end_block FROM upgrade_state
+    type ActiveWindowRow = (String, Option<Vec<u8>>, Option<i64>, Option<i64>);
+    let row: Option<ActiveWindowRow> = sqlx::query_as(
+        "SELECT state, proposal_id, start_block, end_block FROM upgrade_state
           WHERE stack_role = 'GCS' AND status = 'in_progress'",
     )
     .fetch_optional(executor)
     .await?;
 
-    let Some((state, start_block, end_block)) = row else {
+    let Some((state, proposal_id, start_block, end_block)) = row else {
         return Ok(None);
     };
     if !matches!(state.as_str(), "UpgradeActivated" | "DryRunStarted") {
         debug!(state = %state, "GCS not in UpgradeActivated/DryRunStarted — ignoring");
         return Ok(None);
     }
-    Ok(start_block.zip(end_block))
+    Ok(proposal_id
+        .zip(start_block)
+        .zip(end_block)
+        .map(|((proposal_id, start), end)| (proposal_id, start, end)))
 }
 
 /// Per-track eager consensus state. We only need ONE unanimous block to anchor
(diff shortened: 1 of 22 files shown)


coprocessor/fhevm-engine/consensus-detector/src/lib.rs
```diff
@@ -401,23 +424,34 @@ async fn active_host_chain_id(pool: &Pool<Postgres>) -> Result<Option<i64>, Erro
...
+/// anchor per track per window, and at most one `unanimity_consensus_timeout` per
+/// window if the deadline elapses before both tracks anchor.
 async fn consensus_pass(
     pool: &Pool<Postgres>,
     http: &reqwest::Client,
     s3_urls: &Arc<RwLock<Vec<String>>>,
     host: &mut ConsensusTrack,
     gateway: &mut ConsensusTrack,
-    current_window: &mut Option<(i64, i64)>,
+    ws: &mut WindowState,
```
Eikix: nit: ws is already known for websocket

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
```diff
@@ -401,23 +424,34 @@ async fn active_host_chain_id(pool: &Pool<Postgres>) -> Result<Option<i64>, Erro
     Ok(row.and_then(|(v,)| v))
 }
 
+struct WindowState {
+    window: Option<(i64, i64)>,
+    timeout_deadline: Option<Instant>,
```
Eikix: not sure I understand:

once the host reaches the end block: If there is no unanimity, you trigger some grace period "timeout_deadline" and then when this timeout is reached, GCS DB is reset?

what is the use of this grace period?

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
```diff
@@ -401,23 +424,34 @@ async fn active_host_chain_id(pool: &Pool<Postgres>) -> Result<Option<i64>, Erro
     Ok(row.and_then(|(v,)| v))
 }
 
+struct WindowState {
```
Eikix, line 436: my intuition says WindowState might be a bit overloaded as a struct

it carries the window of blocks, but also a piece of state machine that indicates whether we are in a failure path?

coprocessor/fhevm-engine/fhevm-engine-common/src/gcs_activation.rs
```diff
@@ -70,29 +102,25 @@ pub async fn run_gcs_activation_watcher(
...
-                        prev,
-                        "GCS DryRunStarted observed; releasing worker at start_block"
-                    );
-                }
+        let current = state.load(Ordering::SeqCst);
+        let next = row.as_ref().map_or(current, |(fsm_state, start_block)| {
+            host_gate_value(fsm_state, *start_block, current)
+        });
+        if next != current {
+            state.store(next, Ordering::SeqCst);
```
obatirou, line 113: Is it correct that the SNS worker check the flag only at startup ?
https://github.com/zama-ai/fhevm/blob/35beb157e8fe2d65fad3bfca344518e82e86d8fc/coprocessor/fhevm-engine/sns-worker/src/lib.rs#L720-L733

Hence, continue to process work

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
PanGan21: The grace period gives the operators time to finish agreeing async, since reaching end_block on chain doesn't mean all operators submissions are collected yet.
There is some lag between reaching the end_block on host chain and having all operators uploading their state hash on s3. The timeout_deadline is the cutoff for how long we wait for that lag before giving up and resetting for a retry.

coprocessor/fhevm-engine/fhevm-engine-common/src/gcs_activation.rs
PanGan21: Yes correct observation it has been addressed in: https://github.com/zama-ai/fhevm/pull/3236/commits/a832f3c632d6c393fa04f678f455b8e939a713cc

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
PanGan21: Addressed here: https://github.com/zama-ai/fhevm/pull/3236/commits/a832f3c632d6c393fa04f678f455b8e939a713cc

coprocessor/fhevm-engine/consensus-detector/src/lib.rs
PanGan21: I refactored it here: https://github.com/zama-ai/fhevm/pull/3236/commits/e8cc48554e103adaf1bea32a252165c2bebaf247
It still carries some state, however using the enum it might be a bit cleaner. Let me know if you think otherwise

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -250,95 +255,109 @@ pub async fn handle_upgrade_activated(
...
+                            error!(error = %e, "failed to prune gcs.verify_proofs; skipping gw release");
+                            return;
                         }
                     }
-                    Ok(false) => info!(
-                        gw_start_block,
-                        "gw readiness loop exited without satisfying readiness — skipping prune and release"
-                    ),
-                    Err(e) => error!(error = %e, "GCS gateway dry-run readiness loop failed"),
+                    if let Err(e) = transition_to_gw_dry_run_started(&pool).await {
```
obatirou: Is this is ok to not input the proposal id?
This is not symmetric with transition_to_dry_run_started

PanGan21: I missed this. Addressed here: https://github.com/zama-ai/fhevm/pull/3236/commits/65a0ef1399d738c8c003381891cca5205b5b9632

Eikix (approved):

PR #3250, opened 2026-07-20
chore(relayer): alloy and rust bump

eudelins-zama:
### Main changes

Follow up of https://github.com/zama-ai/fhevm/pull/3217

The goal is to bump `alloy` to `2.x` in both `coprocessor` and `relayer` ASAP, so we can get rid of the multi-versions supported previously introduced in shared sub-crates (ex [here](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/rust_bindings/Cargo.toml#L13-L17)), and accepts only `alloy = 2` there.

This PR focuses only on `relayer`. The Rust version bump was done only because latest `alloy` [MSRV](https://docs.rs/alloy/latest/alloy/#supported-rust-versions-msrv) is `1.94.1`.

### CI changes

Took this PR as an opportunity to improve Rust setup in relayer CI. We used to rely on a 3rd party `dsherret/rust-toolchain-file` action. But now `rustup` is pre-installed by default on Github runners, so we don't need any 3rd party action anymore.
We can have our own shared setup action, that just rely on `rustup`.
The goal is to reuse it for coprocessor (WIP [here](https://github.com/zama-ai/fhevm/pull/3256)) and other workflows relying on Rust.

Also, we now use the `rustup-toolchain.toml` to build the Relayer's docker image. Switch to the shared rust golden image for the build stage (with support of Rust 1.97.1, as not already the case for chainguard rust image).
Build run example: https://github.com/zama-ai/fhevm/actions/runs/29833729227

### DIFF
--- a/relayer/docker/relayer/Dockerfile
+++ b/relayer/docker/relayer/Dockerfile
@@ -1,11 +1,13 @@
 # syntax=docker/dockerfile:1
-# Base from https://docs.docker.com/guides/rust/develop/
-ARG RUST_VERSION=1.93
+# RUST_IMAGE_VERSION is intentionally left without a default so the Rust toolchain
+# version has a single source of truth: relayer/rust-toolchain.toml.
+ARG RUST_IMAGE_VERSION
 ARG APP_NAME=fhevm-relayer
 
-FROM cgr.dev/zama.ai/rust:${RUST_VERSION}-dev AS build
+FROM ghcr.io/zama-ai/fhevm/gci/rust-glibc:${RUST_IMAGE_VERSION} AS build
 ARG APP_NAME
 WORKDIR /app
+USER root
 
 # Build the application.
 # Leverage a cache mount to /usr/local/cargo/registry/
@@ -14,9 +16,6 @@ WORKDIR /app
 # Leverage a bind mount to the src directory to avoid having to copy the
 # source code into the container. Once built, copy the executable to an
 # output directory before the cache mounted /app/target is unmounted.
-USER root
-RUN apk add --no-cache openssl-dev=3.6.1-r4
-
 RUN --mount=type=bind,source=relayer/Cargo.toml,target=Cargo.toml \
     --mount=type=bind,source=relayer/Cargo.lock,target=Cargo.lock \
     --mount=type=bind,source=relayer/build.rs,target=build.rs \
--- a/relayer/test-support/ethereum_rpc_mock/src/fhevm.rs
+++ b/relayer/test-support/ethereum_rpc_mock/src/fhevm.rs
@@ -1060,7 +1060,6 @@ fn build_user_decrypt_response(
         vec![
             Decryption::UserDecryptionResponse::SIGNATURE_HASH,
             B256::from(decryption_id),
-            B256::from(U256::from(0)), // indexShare topic
         ],
     )
 }
@@ -1110,11 +1109,6 @@ fn build_public_decrypt_response(
         vec![
             Decryption::PublicDecryptionResponse::SIGNATURE_HASH,
             B256::from(decryption_id),
-            if success {
-                B256::from([1u8; 32])
-            } else {
-                B256::ZERO
-            },
         ],
     )
 }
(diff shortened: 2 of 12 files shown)


relayer/test-support/ethereum_rpc_mock/src/fhevm.rs
```diff
@@ -1060,7 +1060,6 @@ fn build_user_decrypt_response(
         vec![
             Decryption::UserDecryptionResponse::SIGNATURE_HASH,
             B256::from(decryption_id),
-            B256::from(U256::from(0)), // indexShare topic
```
eudelins-zama, line 1063: **Removed the extra log topics in ethereum_rpc_mock**

The `alloy` 2.0 bump made `SolEvent::decode_log_data` strict about topic count: a log must have exactly `1 (signature) + <indexed fields>` topics. The mock was emitting extra topics that Solidity never would — a success flag on `PublicDecryptionResponse`, and `indexShare` (a non-indexed field) on `UserDecryptionResponse`. Alloy 1.7 silently ignored the surplus; 2.0 rejects it with `topic list length mismatch`, so the relayer failed to decode the mocked Gateway responses and the success-path tests timed out.

These events each have only `decryptionId` indexed, so the correct log carries 2 topics. The removed values were never read from topics anyway (the relayer reads them from the decoded body), so this just aligns the mock with real contract output.

relayer/test-support/ethereum_rpc_mock/src/fhevm.rs
```diff
@@ -1110,11 +1109,6 @@ fn build_public_decrypt_response(
         vec![
             Decryption::PublicDecryptionResponse::SIGNATURE_HASH,
             B256::from(decryption_id),
-            if success {
-                B256::from([1u8; 32])
-            } else {
-                B256::ZERO
-            },
```
eudelins-zama, line 1117: Same as [here](https://github.com/zama-ai/fhevm/pull/3250/changes#r3615694795)

relayer/test-support/ethereum_rpc_mock/src/fhevm.rs
```diff
@@ -1206,7 +1200,6 @@ fn build_individual_user_decrypt_response(
         vec![
             Decryption::UserDecryptionResponse::SIGNATURE_HASH,
             B256::from(decryption_id),
-            B256::from(U256::from(index_share)),
```
eudelins-zama, line 1209: Same as [here](https://github.com/zama-ai/fhevm/pull/3250/changes#r3615694795)

zmalatrax (dismissed):

relayer/docker/relayer/Dockerfile
```diff
@@ -1,11 +1,13 @@
 # syntax=docker/dockerfile:1
-# Base from https://docs.docker.com/guides/rust/develop/
-ARG RUST_VERSION=1.93
+# RUST_IMAGE_VERSION is intentionally left without a default so the Rust toolchain
+# version has a single source of truth: relayer/rust-toolchain.toml.
+ARG RUST_IMAGE_VERSION
 ARG APP_NAME=fhevm-relayer
 
-FROM cgr.dev/zama.ai/rust:${RUST_VERSION}-dev AS build
+FROM ghcr.io/zama-ai/fhevm/gci/rust-glibc:${RUST_IMAGE_VERSION} AS build
```
zmalatrax, line 7: Should we care about the relayer-docker-build warning?
> **Default value for global ARG results in an empty or invalid base image name**
Check warning: Default value for global ARG results in an empty or invalid base image name
InvalidDefaultArgInFrom: Default value for ARG ghcr.io/zama-ai/fhevm/gci/rust-glibc:${RUST_IMAGE_VERSION} results in empty or invalid base image name
More info: https://docs.docker.com/go/dockerfile/rule/invalid-default-arg-in-from/

eudelins-zama: This is on purpose to not have a default value. Otherwise it would be a version that we need to manually increase from time to time (or it would just be noise).
But the goal is to rely solely on the `rust-toolchain.toml` as the single source of truth, and never use a manual `RUST_IMAGE_VERSION`.
Not sure why the warning appears for only 1 of the 2 updated `Dockerfile` though

zmalatrax: Okay, makes sense, we can disregard this warning then

zmalatrax (approved):

eudelins-zama:
@mergifyio queue

PR #3255, opened 2026-07-21
feat(host-contracts): confidential-batcher deposit path (join, quit, dispatch, settle, claim)

Eikix:
Closes zama-ai/fhevm-internal#1757. Part of the confidential-vault demo epic (fhevm-internal#1754).

## What

The confidential batcher's deposit direction — the product's core: privacy through aggregation. Users `join` a batch with encrypted amounts; only the batch **total** is ever revealed (by burning the batch's encrypted balance and having the KMS certify the burned amount); the total goes into `demo-vault`; users `claim` encrypted shares at the batch's frozen public rate. New program `confidential-batcher`, replacing the `confidential-deposit-app` reference it evolves (nothing depended on it; all references retargeted).

Design record: DD-042 (+ implementation addendum) and `solana/docs/CONFIDENTIAL_VAULTS.md`.

## Design

- **State**: `Batcher` config; per-batch `Batch` with its **own confidential token account** (so the revealed total is exactly this batch's sum); per-(batch,user) `DepositRecord`. One PDA — `batch_authority` — owns every per-batch account, is the compute subject and app authority of every batcher eval, and signs all CPIs (one seed set).
- **`join`** (user-signed, one transaction): attested `confidential_transfer` into the batch account — the transfer's recipient rule gives the batch authority read admission on the transferred amount by construction, no token changes — then the batcher's own eval re-materializes the deposit into the user's `DepositRecord` lineage *in the same instruction* (nothing ever reads the transferred lineage later, so supersede races are structurally impossible). Repeated joins accumulate.
- **`quit`** (user-signed, Pending only): exact refund via `confidential_transfer_from_value` of the deposit lineage, which is then reset to encrypted zero — re-join accumulates from zero.
- **`dispatch`** (permissionless after `min_batch_age`): one `confidential_burn_from_value` CPI of the batch's full balance (the whole-balance alias goes through the token's dedup); the born-public burned handle is pinned on the batch.
- **`settle`** (permissionless): `redeem_burned_amount` CPI (KMS cert + exact-handle MMR proof + write-once redemption marker) releases the plain total → `demo_vault::deposit` → wrap received shares back to confidential (public amount = the already-public total) → freeze `rate = floor(shares·10⁹/total)`, the one plaintext division. Zero-total batches cancel; the only branch is on the KMS-certified public cleartext.
- **`claim`** (permissionless pull, destination pinned to the record's user): one `mul_div(deposit, rate, 10⁹)` frame, then `transfer_from_value` of the result. Floor both ways ⇒ `Σ claims ≤ shares_received` (unit-tested adversarially incl. u64-scale); per-user rounding loss < 1 share unit; nobody profits from rounding.
- **Privacy audiences**: deposit lineage = {user, batch authority, deposit-mint compute signer}; claim lineage = {user, batch authority, shares compute signer}. The PDAs can't sign KMS decrypt requests, so only the user can decrypt their amounts. The only born-public value in the whole flow is the burned batch total. Events carry no amounts except the intentionally public `BatchSettled{total, shares, rate}`.
- Recorded EVM lessons hold: no participant-count gates, no noiser, nothing branches on encrypted state, no operator custody, quit always available pre-dispatch.
- Divergence from the issue text: `join` uses the *attested* transfer arm (a joining wallet has a fresh coprocessor-attested input, not a pre-existing lineage); the issue's `request_burn_redemption`/`redeem_burned_amount_secp` naming maps to the post-#3243 single-instruction seam. No host or token changes.

## Settle transaction size (measured, not assumed)

Settle composes redeem + vault-deposit + wrap atomically: 34 accounts. A dedicated test builds the real instruction and serializes it against the 1232-byte packet limit:

| cert threshold | legacy tx | v0 + one lookup table |
|---|---|---|
| t=1 (test) | 1369 B — overflows | **382 B — fits** |
| t=7-of-13 (production-like) | 1760 B — overflows | **773 B — fits** |

**Legacy settle is impossible; a v0 transaction with one ALT suffices — no split needed.** The batcher's burned lineage always holds exactly one MMR leaf (proof depth 0, structurally); the test also records the out-of-domain depth-20 bound (overflows even with ALT) as the point where a split would become necessary.

## Known limitations (documented in-code and in DD-042)

- **Dust-total batches strand**: if the certified total floors to zero vault shares, `demo_vault::deposit` reverts and the batch is stuck `Dispatched` forever (the demo vault's share price can only rise, so the revert is not retryable-to-success; a near-free harvest-pump grief exists; loss bounded below one share's worth per batch). Pinned by a dedicated test asserting the exact `ZeroShares` revert and the atomic rollback. The fix (cancel-and-refund: wrap the redeemed total back to the batch's confidential account, refund each user their encrypted deposit via the quit mechanism — no new ACL grants needed) is sketched in DD-042 for a follow-up issue, deliberately out of this PR.
- Single-user batches reveal that user's amount (aggregation of one) — documented and honestly tested.
- Settle assumes the grant deny-list is disabled (POC fixtures); if enabled, settle fails closed rather than bypassing the check.
- `authority_funding_lamports` left on the batch-authority PDA are unrecoverable by design (no sweep).
- No settle cost snapshot: the redemption-marker PDA is seeded by the runtime-derived burned handle, making its bump-search cost order-dependent — the exact failure mode the fixed-key snapshot rule exists to avoid.

## Tests

12 Mollusk tests driving the real program (five programs registered; every CPI a genuine `invoke_signed`), with a cleartext ledger that replays **all** `fhe_eval` CPIs per instruction and asserts exact eval counts: full two-user lifecycle; lifecycle with yield (rate rounds down, claims sum exactly); single-user reveal caveat; repeat-join accumulate + exact quit refund (aliasing regression); zero-total cancel; dust-total stuck-batch pin; min-age reject; status-machine rejects; double-claim reject; open-batch ordering; settle wire-size; cost snapshot with fixed fixture keys (open_batch 250 964 CU, join 429 421, dispatch 470 073, claim 327 538). Plus unit tests for space invariants, rate rounding, and Σ claims ≤ shares.

Full suite on the rebased tip: **782 passed, 0 failed**. Clippy zero warnings, fmt clean, IDL/ABI goldens in sync (host/token surfaces untouched), snapshots byte-identical on re-run, `program_autofixer` clean. Independently reviewed (privacy, authority/fund-safety, aliasing, math, CPI seams, tests): no code defects; the review's documentation findings are folded in.


solana/programs/confidential-batcher/src/instructions/settle.rs
```diff
@@ -0,0 +1,300 @@
...
+                vault_token_account: ctx.accounts.vault_token_account.to_account_info(),
+                depositor_shares: ctx.accounts.batch_share_tokens.to_account_info(),
+                token_program: ctx.accounts.token_program.to_account_info(),
+            },
+            &[&authority_seeds],
+        ),
+        cleartext_total,
+    )?;
+    ctx.accounts.batch_share_tokens.reload()?;
+    let shares_received = ctx.accounts.batch_share_tokens.amount;
```
Eikix: [P1] Compute shares from the deposit delta

The token account's entire balance is treated as newly received shares here. SPL token destinations do not authorize incoming transfers, so shares can be transferred into this PDA-owned account before settlement. For a small batch, a sufficiently large preloaded balance causes `freeze_share_rate` to overflow, leaving every settlement attempt reverted after the deposits were burned. The balance should be captured before the vault CPI, reloaded afterward, and checked subtraction should be used as `shares_received`; only that delta should be wrapped and priced.

PR #3256, opened 2026-07-21
chore(coprocessor): alloy and rust bump

eudelins-zama:
### Main changes

Follow up of https://github.com/zama-ai/fhevm/pull/3217

The goal is to bump `alloy` to `2.x` in both `coprocessor` and `relayer` ASAP, so we can get rid of the multi-versions supported previously introduced in shared sub-crates (ex [here](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/rust_bindings/Cargo.toml#L13-L17)), and accepts only `alloy = 2` there.

This PR focuses only on `coprocessor` (and `listener`, because deps versions of these two entities are coupled). The Rust version bump was done only because latest `alloy` [MSRV](https://docs.rs/alloy/latest/alloy/#supported-rust-versions-msrv) is `1.94.1`.

Include few fix of new clippy lints.

### CI changes

Took this PR as an opportunity to improve Rust setup in copro/listener CI. We used to either:
- hardcode Rust version in CI files (whereas it's already hardcoded in `rust-toolchain.toml`)
- or use a 3rd party `dsherret/rust-toolchain-file` action

But now `rustup` is pre-installed by default on Github runners, so we don't need any 3rd party action anymore.
We can have our own shared setup action, that just rely on `rustup`.

Also, we now parametrized the Rust base image via `ARG RUST_IMAGE_VERSION` so CI injects the toolchain version from `rust-toolchain.toml` (single source of truth), removing the hardcoded version from the coprocessor's per-service Dockerfiles.
We merged `Dockerfile` and `Dockerfile.ci` for the listener as it seems the distinction was not providing a clear benefit, and we moved to using the shared rust golden image as the base image of the build step.
The local-only `Dockerfile.workspace` keeps a hardcoded default (docker-compose builds don't get the injection) and the listener, whose CI and local Dockerfiles were unified into one, keeps that default inline for the same reason.

Docker build run example:
- copro: https://github.com/zama-ai/fhevm/actions/runs/29839354360
- listener: https://github.com/zama-ai/fhevm/actions/runs/29844168906


listener/Dockerfile
```diff
@@ -1,35 +1,40 @@
...
-COPY . .
-RUN cargo chef prepare --recipe-path recipe.json
-
-# ── Builder (cook deps, then build app) ──────────────────────────────
-FROM rust:1.92-slim-bookworm AS builder
-RUN apt-get update \
-    && apt-get install -y pkg-config libssl-dev \
-    && rm -rf /var/lib/apt/lists/* \
-    && cargo install cargo-chef
+COPY listener/ /app/
```
obatirou, line 15: was this tested?
docker-compose do 
```
build:
    context: .
    dockerfile: Dockerfile
```

so not sure this will work now that listener folder is copied in app/ ?

coprocessor/fhevm-engine/fhevm-engine-common/src/db_keys.rs
```diff
@@ -489,11 +489,9 @@ async fn read_large_object_in_chunks(
...
             // calculate the bandwidth of the read operation
             let elapsed = started_at.elapsed().as_secs();
-            let bandwidth = if elapsed > 0 {
-                bytes.len() as u64 / elapsed
-            } else {
-                bytes.len() as u64
-            };
+            let bandwidth = (bytes.len() as u64)
+                .checked_div(elapsed)
+                .unwrap_or(bytes.len() as u64);
```
obatirou: if we refactor, we could have an intermediate variable to increase readability:  `let total_bytes = bytes.len() as u64;`

listener/Dockerfile
eudelins-zama: Let's see: https://github.com/zama-ai/fhevm/actions/runs/29844168906 :slightly_smiling_face:

eudelins-zama: :white_check_mark:

obatirou (dismissed):

listener/crates/listener_core/src/blockchain/evm/evm_block_computer.rs
```diff
@@ -527,6 +527,8 @@ impl EvmBlockComputer {
             excess_blob_gas: rpc_header.excess_blob_gas,
             parent_beacon_block_root: rpc_header.parent_beacon_block_root,
             requests_hash: rpc_header.requests_hash,
+            block_access_list_hash: rpc_header.block_access_list_hash,
```
nboisde, line 530: Quick question here, are we sure this deserialization is compatible with all evm chains ?

coprocessor/fhevm-engine/fhevm-engine-common/src/db_keys.rs
eudelins-zama: Fixed 1592a21864a140f90d416b28e26a23f6a6bdc1e2

listener/crates/listener_core/src/blockchain/evm/evm_block_computer.rs
eudelins-zama: Both new fields are optional, plus the whole line is just forwarding what alloy already decoded from the RPC header, so I think we're good there.

Made Claude do a quick analysis on this, not issue found

charts/coprocessor/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: coprocessor
 description: A helm chart to distribute and deploy Zama fhevm Co-Processor services
-version: 0.13.9
+version: 0.13.10
```
eudelins-zama, line 3: Had to bump this just for comment update in `values.yaml` below 🥲

nboisde (approved):

obatirou (approved):

eudelins-zama:
@mergifyio queue

PR #3260, opened 2026-07-21
chore(kms-connector): destroy epoch tied to context

eudelins-zama:
### Summary

Follow-up of #3073.

Aligns the kms-connector with the new KMS Core gRPC contract for context/epoch destruction, where destroyed epochs are now reported back in the gRPC response.
Also adds a startup cache-reconciliation for `gw-listener` and pins all context/epoch validity view calls to finalized blocks.

Bumps `kms-grpc` (`b9087af` → `d27c3b52`, `0.14.0` → `0.15.0`).

### What changed

1. **Destroyed epochs flow back via the response.**
   - `DestroyMpcContextRequest` drops its `epoch_ids` field; `DestroyMpcContextResponse` now returns the epoch IDs destroyed alongside a context.
   - New `SendResponse` enum (`Empty` | `DestroyedEpochs(Vec<U256>)`) carries this through the send path; `send_request_with_retries` gains a `map_response` closure.
   - The connector invalidates the returned epochs locally in the processor.

2. **Startup cache reconciliation.** New `EthereumListener::revalidate_context_cache()` re-checks every cached-valid context/epoch against on-chain `ProtocolConfig` and invalidates stale rows. Wired into gw-listener startup as a backstop for missed events or a KMS-Core restart mid-destruction.

3. **View calls pinned to `BlockId::finalized()`** across all five context/epoch validity call sites, plus a test-setup helper that waits for deployments to finalize.

### DIFF
--- a/kms-connector/crates/gw-listener/src/core/ethereum.rs
+++ b/kms-connector/crates/gw-listener/src/core/ethereum.rs
@@ -415,6 +415,7 @@ mod tests {
         let db = test_instance.db();
         clear_context_cache(db).await;
 
+        // Arbitrary context/epoch IDs chosen to test the revalidation logic.
         let destroyed_context = U256::from(1_u64);
         let destroyed_context_epoch = U256::from(10_u64);
         let valid_context = U256::from(2_u64);
@@ -426,9 +427,14 @@ mod tests {
             .await
             .unwrap();
 
-        // Rows are audited in `ORDER BY id` order: `isValidKmsContext(#1)` -> false,
-        // `isValidKmsContext(#2)` -> true, then `isValidEpochForContext(#2, #20)` -> true.
-        // Epoch #10 must not trigger a call, as its context is already known destroyed.
+        // The `Asserter` replies to each on-chain `.call()` with the next queued response, in
+        // order. `revalidate_context_cache` audits contexts first (both are cached as valid, so
+        // both are checked), then epochs, each `ORDER BY id`. So the calls happen in this order:
+        //   1. `isValidKmsContext(#1)`         -> false: context #1 is destroyed on-chain
+        //   2. `isValidKmsContext(#2)`         -> true:  context #2 is still valid on-chain
+        //   3. `isValidEpochForContext(#2,#20)`-> true:  epoch #20 (of valid ctx #2) still valid
+        // Note there is no call for epoch #10: its context (#1) was just found destroyed, so the
+        // audit invalidates the epoch without an on-chain check.
         let asserter = Asserter::new();
         asserter.push_success(&false.abi_encode());
         asserter.push_success(&true.abi_encode());
@@ -457,7 +463,10 @@ mod tests {
             .await
             .unwrap();
 
-        // `isValidKmsContext(#1)` -> true, `isValidEpochForContext(#1, #10)` -> false
+        // The `Asserter` replies to each on-chain `.call()` with the next queued response, in
+        // order. In this scenario, the context is still valid but its epoch has been destroyed:
+        //   1. `isValidKmsContext(#1)`          -> true:  context #1 is still valid
+        //   2. `isValidEpochForContext(#1,#10)` -> false: epoch #10 is destroyed, so invalidated
         let asserter = Asserter::new();
         asserter.push_success(&true.abi_encode());
         asserter.push_success(&false.abi_encode());
--- a/kms-connector/crates/gw-listener/src/core/ethereum.rs
+++ b/kms-connector/crates/gw-listener/src/core/ethereum.rs
@@ -158,7 +158,7 @@ where
             let Some(context_id) = epoch.context_id.map(|i| U256::from_le_slice(&i)) else {
                 // A valid epoch row always carries its context association; without it the epoch
                 // cannot be checked on-chain. This should be unreachable, but we delete the row
-                // just in case so the kms-worker's on-chain is able to fix the cache.
+                // just in case so the kms-worker's on-chain check is able to fix the cache.
                 warn!("KMS epoch #{epoch_id} was cached as valid without any context. Deleting...");
                 sqlx::query!("DELETE FROM kms_epoch WHERE id = $1", epoch.id)
                     .execute(&self.db_pool)
--- a/kms-connector/crates/gw-listener/src/core/ethereum.rs
+++ b/kms-connector/crates/gw-listener/src/core/ethereum.rs
@@ -6,7 +6,7 @@ use crate::{
     monitoring::metrics::{EVENT_LISTENING_ERRORS, EVENT_RECEIVED_COUNTER},
 };
 use alloy::{
-    eips::BlockNumberOrTag,
+    eips::{BlockId, BlockNumberOrTag},
     network::Ethereum,
     primitives::{B256, U256},
     providers::Provider,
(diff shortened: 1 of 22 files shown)


kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -110,12 +114,78 @@ where
...
+                .fetch_all(&self.db_pool)
+                .await?;
+        for epoch in epochs {
+            let epoch_id = U256::from_le_slice(&epoch.id);
+            let Some(context_id) = epoch.context_id.map(|i| U256::from_le_slice(&i)) else {
+                // A valid epoch row always carries its context association; without it the epoch
+                // cannot be checked on-chain. This should be unreachable, but we delete the row
+                // just in case so the kms-worker's on-chain check is able to fix the cache.
+                warn!("KMS epoch #{epoch_id} was cached as valid without any context. Deleting...");
+                sqlx::query!("DELETE FROM kms_epoch WHERE id = $1", epoch.id)
```
jot2re, line 163: I get that we delete the epoch if it is basically invalid, but how could it happen in the first place that there is no context associated with an epoch?

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -347,6 +407,107 @@ mod tests {
...
+        let valid_epoch = U256::from(20_u64);
+        publish_context_and_epoch(db, destroyed_context, destroyed_context_epoch)
+            .await
+            .unwrap();
+        publish_context_and_epoch(db, valid_context, valid_epoch)
+            .await
+            .unwrap();
+
+        // Rows are audited in `ORDER BY id` order: `isValidKmsContext(#1)` -> false,
+        // `isValidKmsContext(#2)` -> true, then `isValidEpochForContext(#2, #20)` -> true.
```
jot2re: I am not sure i get why context Id 2 is valid but 1 is not. It seems they are both published. I guess it is because of the boolean value pushed to the asserter but a little comment would be great, in particular since there are only 3 pushes and 4 asserts

kms-connector/crates/gw-listener/src/core/ethereum.rs
```diff
@@ -347,6 +407,107 @@ mod tests {
...
+    #[timeout(Duration::from_secs(90))]
+    #[tokio::test]
+    async fn test_revalidate_context_cache_invalidates_destroyed_epoch() {
+        let test_instance = TestInstanceBuilder::db_setup().await.unwrap();
+        let db = test_instance.db();
+        clear_context_cache(db).await;
+
+        let context_id = U256::from(1_u64);
+        let epoch_id = U256::from(10_u64);
+        publish_context_and_epoch(db, context_id, epoch_id)
```
jot2re, line 462: Same here. I am confused why this is correct. Can you add some comments?

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
```diff
@@ -264,8 +258,19 @@ impl<GP: Provider + Clone + 'static, HP: Provider, C: ContextManager> DbEventPro
         if !event.already_sent {
             let (error_count, result) = self.kms_client.send_request(&request).await;
             event.error_counter += error_count;
-            result?;
+            let send_response = result?;
             event.already_sent = true;
+
+            // Invalidate epochs associated to destroyed context returned by the KMS Core.
+            // A failure must not prevent the event from completing, as retrying would result in an
+            // `AlreadyExists` from KMS Core with no epoch IDs.
```
jot2re: A repeated call to the KMS core should not give `AlreadyExists`. Instead it would give `NotFound` (in case everything has been successfully deleted). If there are parts that have not been successfully deleted (i.e. the KMS core returns an error), then a repeated call that succeeds would give Ok and all the epoch ids as response.

jot2re (changes requested):
Generally looks great! Thanks for the quick update!
I just have a couple of tiny nitpicky comments

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: It should be only possible if kms-connector receive an epoch destruction event for an epoch that was not already in his cache.
In such case, it only receive the epoch_id, not context_id

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: > I am not sure i get why context Id 2 is valid but 1 is not

The `asserter.push_success(true/false.abi_encode());` is what decide what is valid or not.
They are the mock responses of the `isValidKmsContext`/`isValidEpochForContext` calls

eudelins-zama: > I guess it is because of the boolean value pushed to the asserter but a little comment would be great, in particular since there are only 3 pushes and 4 asserts

There's already a comment explaining the overall thing + the "Epoch #10 must not trigger a call, as its context is already known destroyed." specific one for the 3 pushes vs 4 asserts...

It would be very verbose if we add more no?

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: OK fair, I'll try to clarify a bit these :innocent:

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Fixed 9355681018bc910579444bd008e232f3ac51af15

kms-connector/crates/gw-listener/src/core/ethereum.rs
eudelins-zama: Fixed 9355681018bc910579444bd008e232f3ac51af15

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
eudelins-zama: Fixed 9355681018bc910579444bd008e232f3ac51af15

jot2re (commented):
A few more comments from Claude:
The processor invalidating the epochs returned in `DestroyMpcContextResponse` — is untested. The existing integration test mocks `DestroyMpcContext` to return `Empty::default()` integration_tests.rs, which decodes to `epoch_ids = []`, so `map_destroyed_epochs` yields `DestroyedEpochs(vec![])` and the invalidation loop never runs. There is:
- no unit test for map_destroyed_epochs (including the `invalid-request_id` → Irrecoverable branch), and
- no positive test asserting that a non-empty epoch_ids response actually flips the corresponding `kms_epoch` rows to `is_valid = FALSE` in the DB.

`send_request` is `pub` and its return type changed; rule 6 wants a positive test for the new path. Recommend a test where the mock returns a DestroyMpcContextResponse with two epoch IDs and asserts both rows are invalidated.

eudelins-zama:
> no positive test asserting that a non-empty epoch_ids response actually flips the corresponding kms_epoch rows to is_valid = FALSE in the DB.

Fair, added an integration test in d814a05537dbc7c43d439149423746078500c208

> no unit test for map_destroyed_epochs (including the invalid-request_id → Irrecoverable branch)

Would not add value IMO, most important case is tested in the integration test above

jot2re (approved):
LGTM!

PR #3261, opened 2026-07-21
feat(relayer): add extradata tag validation

zmalatrax:
Closes https://github.com/zama-ai/fhevm-internal/issues/1771

### DIFF
--- a/relayer/openapi.yml
+++ b/relayer/openapi.yml
@@ -1321,10 +1321,7 @@ components:
           example: '1'
         extraData:
           type: string
-          description: |-
-            Extra data forwarded verbatim to the gateway contract. Accepts `"0x00"`, version `0x01`
-            (`0x01` + 32-byte contextId), or version `0x02` (`0x02` + 32-byte contextId + 32-byte epochId).
-            contextId must be 0x07-tagged and epochId must be 0x08-tagged (first byte of each).
+          description: Extra data forwarded verbatim to the gateway contract. Accepts `"0x00"`, version `0x01` (`0x01` + 32-byte contextId), or version `0x02` (`0x02` + 32-byte contextId + 32-byte epochId). contextId must be 0x07-tagged and epochId must be 0x08-tagged (first byte of each).
           example: '0x00'
         handleContractPairs:
           type: array
@@ -1368,11 +1365,7 @@ components:
           - "0x1234567890123456789012345678901234567890"
         extraData:
           type: string
-          description: |-
-            Extra data forwarded verbatim to the gateway contract. Accepts `"0x00"`,
-            version `0x01` (`0x01` + 32-byte contextId), or version `0x02`
-            (`0x02` + 32-byte contextId + 32-byte epochId). contextId must be
-            0x07-tagged and epochId must be 0x08-tagged (first byte of each).
+          description: Extra data forwarded verbatim to the gateway contract. Accepts `"0x00"`, version `0x01` (`0x01` + 32-byte contextId), or version `0x02` (`0x02` + 32-byte contextId + 32-byte epochId). contextId must be 0x07-tagged and epochId must be 0x08-tagged (first byte of each).
           example: '0x00'
         handles:
           type: array
@@ -1636,10 +1629,7 @@ components:
           minItems: 1
         extraData:
           type: string
-          description: |-
-            Extra data forwarded verbatim to the gateway contract. Accepts `"0x00"`, version `0x01`
-            (`0x01` + 32-byte contextId), or version `0x02` (`0x02` + 32-byte contextId + 32-byte epochId).
-            contextId must be 0x07-tagged and epochId must be 0x08-tagged (first byte of each).
+          description: Extra data forwarded verbatim to the gateway contract. Accepts `"0x00"`, version `0x01` (`0x01` + 32-byte contextId), or version `0x02` (`0x02` + 32-byte contextId + 32-byte epochId). contextId must be 0x07-tagged and epochId must be 0x08-tagged (first byte of each).
           example: '0x00'
     PublicDecryptResponseJson:
       type: object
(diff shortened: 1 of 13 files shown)


relayer/src/http/endpoints/v2/types/public_decrypt.rs
```diff
@@ -17,6 +17,7 @@ pub struct PublicDecryptRequestJson {
     pub ciphertext_handles: Vec<String>,
     /// Extra data forwarded verbatim to the gateway contract. Accepts `"0x00"`, version `0x01`
     /// (`0x01` + 32-byte contextId), or version `0x02` (`0x02` + 32-byte contextId + 32-byte epochId).
+    /// contextId must be 0x07-tagged and epochId must be 0x08-tagged (first byte of each).
```
manoranjith: Since the info is duplicated in many places and prone to drift from each other, could we move it to validate func's docs and or rename it as `validate_kms_extra_data_v0_v1_v2`. Wdyt ?

relayer/src/http/utils/validations.rs
```diff
@@ -113,35 +113,51 @@ pub fn validate_0x_hex_allow_empty(hex_str: &str) -> Result<(), ValidationError>
...
+    const CONTEXT_ID_TAG: u8 = 0x07;
+    const EPOCH_ID_TAG: u8 = 0x08;
+
     match extra_data {
         "0x00" => Ok(()),
         s if s.len() == 68 && s.starts_with("0x01") => {
             // Version 1: [0x01 | contextId(32B)] = 33 bytes = 66 hex chars + "0x" prefix = 68 chars
-            decode_versioned_extra_data(s)
+            let bytes = decode_versioned_extra_data(s)?;
+            validate_id_tag(&bytes[1..33], CONTEXT_ID_TAG)
```
manoranjith: Nit: Since `validate ID tag is only a single byte comparision, could we inline it ?  Since we are comparing just a byte, we could also `-> lower_hex` and directly compare the ascii if it one liner is preffered.

relayer/openapi.yml
```diff
@@ -1637,6 +1639,7 @@ components:
           description: |-
             Extra data forwarded verbatim to the gateway contract. Accepts `"0x00"`, version `0x01`
             (`0x01` + 32-byte contextId), or version `0x02` (`0x02` + 32-byte contextId + 32-byte epochId).
+            contextId must be 0x07-tagged and epochId must be 0x08-tagged (first byte of each).
```
manoranjith: Even if we move the comments into docs of validatation fn or its name, we could retain it in the open api comments, that are intended for end user.

manoranjith (dismissed):
LGTM. Added a minor and a nit comment.

relayer/src/http/endpoints/v2/types/public_decrypt.rs
zmalatrax: Indeed, I've refactored to a single shared `extra_data_decryption_schema()` referenced by
`#[schema(schema_with = ...)]` in the endpoints 

Note that utoipa deprecated `.example` in favor of `.examples`, but I've kept `.example` for consistency, hence the `allow(deprecated)`

I don't think that we need to rename with the version, it would require renaming it every time a new extra data version is added, introducing a potential drift

relayer/src/http/utils/validations.rs
zmalatrax: I'd rather avoid comparing hex strings

As we use it 3 times in the function, inlining it would bloat the function with 3x the error construction, wdyt? 🤔 
Though I've simplified the validate_id_tag to only tag a single byte rather than a 32-byte slice

relayer/openapi.yml
zmalatrax: The openapi.yml is generated from `make openapi-generate`)

It surfaced a small inconsistency on the VersionResponse example, it was still hardcoded in health.rs, linked it to `CARGO_PKG_VERSION` (otherwise `make openapi-generate` overwrites the hand-writter 0.14.0 back to 0.9.0)

manoranjith: On CARGO_PKG_VERSION, we need a ci check to ensure it's upto date before a release. Maybe a task out of scope for this PR.

relayer/src/http/endpoints/health.rs
```diff
@@ -25,7 +25,7 @@ pub struct HealthResponse {
 pub struct VersionResponse {
     #[schema(example = "fhevm-relayer")]
     name: String,
-    #[schema(example = "0.9.0")]
+    #[schema(example = json!(env!("CARGO_PKG_VERSION")))]
```
manoranjith, line 28: Thanks for the additional fix 💛

relayer/src/http/utils/validations.rs
```diff
@@ -107,41 +107,79 @@ pub fn validate_0x_hex_allow_empty(hex_str: &str) -> Result<(), ValidationError>
...
+#[allow(deprecated)]
+pub fn extra_data_decryption_schema() -> utoipa::openapi::schema::Object {
+    utoipa::openapi::schema::ObjectBuilder::new()
+        .schema_type(utoipa::openapi::schema::SchemaType::Type(
+            utoipa::openapi::schema::Type::String,
+        ))
+        .description(Some(
+            "Extra data forwarded verbatim to the gateway contract. Accepts `\"0x00\"`, version `0x01` \
+             (`0x01` + 32-byte contextId), or version `0x02` (`0x02` + 32-byte contextId + 32-byte epochId). \
+             contextId must be 0x07-tagged and epochId must be 0x08-tagged (first byte of each).",
```
manoranjith, line 124: Extracting the comment out to a function makes it much cleaner 👍

manoranjith (approved):
Thanks for the review fixes 🙏 . Looks cleaner.

zmalatrax:
@mergify queue

manoranjith (commented):
lgtm

PR #1773, opened 2026-01-12
ci(common): fix re-tag race conditions

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/845

PR #3263, opened 2026-07-21
feat(host-contracts): confidential-batcher redeem direction (join, dispatch, settle, claim)

Eikix:
Closes zama-ai/fhevm-internal#1758. Part of the confidential-vault demo epic (fhevm-internal#1754). Also fixes fhevm-internal#1774 items 1–2.

## What

The confidential batcher's **redeem direction**, mirroring deposits: users join a batch with encrypted vault shares; only the batch's share **total** is ever revealed (confidential burn + KMS certificate); settle redeems the certified total via `demo_vault::withdraw`, wraps the received underlying back into confidential tokens, and freezes an informational rate; claims pay each user encrypted underlying proportional to their joined shares. Same privacy property both ways: individual amounts stay encrypted, only the batch total is born public.

## Design

- **One program, direction fixed per `Batcher` instance.** Each batcher is initialized as `Deposit` or `Redeem` — mirroring the EVM's two deployments, so pending batches in one direction never block the other — while all account layouts and instructions are shared under direction-neutral **join/payout** vocabulary (join mint = what users batch in; payout mint = what claims pay). The batch lifecycle's only direction branch is settle's vault CPI (`deposit` vs `withdraw`); `initialize_batcher` additionally validates the mint wiring per direction at setup. Join, quit, dispatch, and claim are literally direction-free shared code. Rationale in the DD-042 addendum: the two flows differ in exactly one CPI, so duplicating every accounts struct to encode one branch would double review surface for zero clarity.
- **Exact proportional claims (fixes #1774 item 1, both directions):** claims now pay `mul_div(joined, payout_received, total_joined)` — the exact proportional floor, same FHE op count, intermediate < 2¹²⁸. The old two-step rate math double-rounded; a pinning unit test verifies it stranded exactly 6,148,914,726 raw units on the u64-scale case vs ≤1 unit per claim now. The frozen `payout_rate` is informational/event-facing only (grep-verified: written, never read by an instruction) and saturates at u64::MAX — a display value must not be able to brick settle.
- **Delta accounting preserved as an invariant:** the settle payout leg in both directions snapshots the SPL balance before the vault CPI, reloads after, and wraps only the `checked_sub` delta — preloading the payout account is inert (adversarial tests in both directions at rate-breaking scale).
- **Dust asymmetry (analyzed, tested):** the deposit-side dust brick (#1773) has **no redeem analog** — the vault price never drops below 1:1, so withdrawing ≥1 share always pays ≥1 underlying; pinned by a test at an extreme share price. Deposit dust behavior unchanged and still pinned.
- **Host-policy envelope documented (#1774 item 2):** settle/claim module docs and DD-042 now state the POC assumption that deny-list records and HCU accounts are hardcoded `None`.
- Cancel-after-dispatch stays out of scope (quit-before-dispatch only), per the issue; deadline-cancel is tracked in fhevm-internal#1773.

## Breaking renames (deliberate, POC — no back-compat)

The direction-neutral vocabulary renames merged deposit-direction identifiers: PDA seeds `deposit-record`→`join-record`, `batch-underlying`→`batch-join-underlying`, `batch-share-tokens`→`batch-payout-underlying`; pending label `batcher-pending-join`; `DepositRecord`→`JoinRecord`; event `SharesClaimed`→`PayoutClaimed`; `Batcher`/`Batch` layout changes. `APP_EVENT_VERSION` bumped to 2. Any batcher deployed from the previous commit is orphaned and old event decoders break — accepted cost on this branch, nothing consumes the old deployment. Error enum strictly append-only (retired variants kept in place, marked legacy).

## Tests

24 batcher Mollusk tests (all genuine `invoke_signed` CPIs with the cleartext-ledger replay asserting exact eval counts): full redeem lifecycle ×2 users, redeem with yield (rounding), redeem repeat-join/quit-rejoin aliasing, redeem zero-total cancel, redeem dust-at-extreme-price (settles, no brick), redeem preload attack (inert), claim-after-quit (pays zero, record closed, others unaffected), deposit+redeem batchers interleaved end-to-end over shared mints (joins/dispatches/settles/claims alternating, escrow/vault/user balances checked at every stage), swapped-wiring reject, redeem settle wire-size (legacy overflows 1232 B; v0 + one lookup table fits at 7-of-13 cert threshold), redeem cost snapshots (fixed fixture keys) — plus all existing deposit tests, updated to the exact-proportional claim math.

Full suite: **796 passed, 0 failed**. Clippy `-D warnings`, fmt, IDL/ABI goldens in sync (no host/token changes), cost snapshots byte-identical on re-run, `program_autofixer` clean. Independently reviewed (privacy audiences, delta accounting, fund safety, state machine, cross-direction substitution, math, docs honesty, wire size): no P1/P2 findings; all review nits folded in.

### DIFF
--- a/solana/docs/CONFIDENTIAL_VAULTS.md
+++ b/solana/docs/CONFIDENTIAL_VAULTS.md
@@ -60,13 +60,18 @@ everything below uses Solana-native building blocks.
    share tokens, and wraps them into **confidential shares**. It fixes one
    public exchange rate for the whole batch: shares received ÷ total
    deposited.
-5. **Claim.** Each user collects their shares. Their encrypted deposit is
-   multiplied by the public rate — an encrypted number times a public number,
-   which FHE does cheaply. No one ever divides encrypted by encrypted.
+5. **Claim.** Each user collects their shares: their encrypted deposit's
+   exact proportional part of the batch's shares — an encrypted number times
+   a public number, divided by a public number, which FHE does cheaply. No
+   one ever divides encrypted by encrypted, and the floor rounding means the
+   claims can never add up to more than the batch received.
 
-Withdrawing works the same way in mirror: join a redeem batch with encrypted
-shares, the batch total of shares is revealed and redeemed from the vault,
-and users claim encrypted tokens back at the batch's rate.
+Withdrawing works the same way in mirror, in the same program: a *redeem*
+batcher's batches are joined with encrypted shares, the batch total of shares
+is revealed and withdrawn from the vault, and users claim their exact
+proportional part of the returned tokens, encrypted. Deposit and redeem
+batchers run side by side — one pending batch each — so exits never wait on
+entries.
 
 ## What is public and what is private
 
--- a/solana/docs/DESIGN_DECISIONS.md
+++ b/solana/docs/DESIGN_DECISIONS.md
@@ -1825,3 +1825,45 @@ the loss per batch is bounded below one share's worth. Behavior is pinned by
 cancel-and-refund settle branch: wrap the redeemed underlying back into the batch's confidential
 account and refund each user's encrypted deposit via `confidential_transfer_from_value` (quit's
 mechanism) — not implemented in the deposit-path PR.
+
+Redeem path implemented (fhevm-internal#1758), as an addendum to the deposit path above. **One
+program serves both directions, with the direction on the `Batcher` config** — each config is a
+Deposit or a Redeem instance, mirroring the EVM's two batcher deployments: a pending deposit batch
+never blocks a redeem batch (each batcher serializes only its own batches), yet every account
+layout and every instruction is shared. The vocabulary is direction-neutral — a JOIN confidential
+mint (what users batch in: cUnderlying for deposits, cShares for redeems) and a PAYOUT confidential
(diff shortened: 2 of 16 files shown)


Eikix:
Found one documentation issue; no blocking code defects.

- [P2] The overview still describes claims as `encrypted(amount) × public rate`, but this PR changes claims to exact proportional `joined × payout_received / total_joined`. The same stale description remains in `solana/README.md`. These should be updated so integrators do not implement the superseded double-rounding calculation.

Validation:

- `confidential-batcher`: 9/9 tests passed.
- `batcher_mollusk`: 24/24 tests passed.
- `git diff --check`: clean.
- GitHub checks: green.
- No GitHub comments were submitted.

::code-comment{title="[P2] Update the superseded claim formula" body="The diagram still says claims use `e(amount) × public share rate`, while this PR changes claims to `joined × payout_received / total_joined` and makes the stored rate informational only. The overview in `solana/README.md` repeats the old formula, so both descriptions should be updated to prevent clients from reproducing the removed double-rounding behavior." file="/Users/work/code/zama/fhevm/.claude/worktrees/batcher-redeem/solana/docs/CONFIDENTIAL_VAULTS.md" start=43 priority=2}

PR #3265, opened 2026-07-21
fix(coprocessor): guard green-stack writes against dry-run rollback reset

PanGan21:
### Problem

When a GCS dry-run is rolled back, the controller wipes the `gcs` schema and notifies the workers to pause. Workers learn about the pause from an in-memory flag that a background task flips when the notification arrives, so there's a window where the rollback has already committed (schema reset) but a worker hasn't seen the flag yet. A worker mid-computation can then finish its old batch and write stale rows into the freshly reset schema, which can poison the next upgrade attempt.

### Fix

Order every green write against the reset, and have each write re-check the durable state in its own transaction, so there's no window to race:

- The rollback runs in one transaction holding the **exclusive** advisory lock: it marks the upgrade row `PAUSED`, wipes and recreates the schema, and notifies the workers.
- Every guarded write takes the **shared** form of the same lock. A write therefore either commits before the reset (and is dropped by it) or runs after the reset — it can never interleave with the wipe.
- A write that runs after the reset re-reads the paused state inside its own transaction, so a worker that missed the flag still sees `PAUSED` and holds off. Raw chain ingestion (host-listener, gw-listener) keeps tailing after a rollback; derived-output workers (tfhe, sns, zkproof, `state_hash`, tx-sender digest/ct128) skip while the row is `PAUSED`.

Because the write path itself re-checks and is ordered against the reset, the protection holds for every writer without depending on the in-memory flag arriving in time.

### Options considered

- **In-memory flag check before writing**: cheap, but only narrows the window; the flag lags the durable rollback, so stale writes still go through.
- **Rely on the timeout**: the stale row is from the *old* window, so it never enters the new comparison, yet cutover still merges it into live data (whole-table merge, overwrite on conflict). Nothing catches it, so this isn't a safe fallback.
- **Restart + wait before reset**: reset the schema only while all workers are stopped and waiting. Race-free, but it restarts every green worker on each rollback and needs extra machinery (durable reset flag + self-restart +
  startup barrier).
- **Durable check in each write transaction (this PR)**: each write re-checks the paused state in its own transaction, ordered against the reset by a shared/exclusive advisory lock. Race-free with no worker restart; costs a stack-specific check (raw vs derived) on the write paths.

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
@@ -310,19 +310,51 @@ pub async fn begin_guarded_conn(
     Ok(tx)
 }
 
-/// PostgreSQL advisory-lock key serializing BCS writes against `execute_cutover`.
-/// `execute_cutover` (upgrade-controller) takes the **exclusive** form; every BCS
-/// write transaction takes the **shared** form via [`cutover_gate`]. Chosen to be
+/// PostgreSQL advisory-lock key serializing writes against cutover and rollback.
+/// The upgrade-controller takes the **exclusive** form; every guarded write
+/// transaction takes the **shared** form via [`cutover_gate`]. Chosen to be
 /// recognizable in logs (`0x4648_4556_4355_5456` ~ ASCII "FHEVCUTV").
 pub const CUTOVER_LOCK_ID: i64 = 0x4648_4556_4355_5456;
 
-/// Cutover safety gate for a BCS write transaction.
+/// Result of opening a guarded write transaction.
+pub enum WriteGuard<'a> {
+    Proceed(Transaction<'a, Postgres>),
+    Stop,
+    Skip,
+}
+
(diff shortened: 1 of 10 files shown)


coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
```diff
@@ -370,27 +396,37 @@ pub async fn cutover_gate(
...
     let mut tx = begin_guarded_pool(pool).await?;
-    if cutover_gate(&mut tx, gcs_mode).await? {
-        tx.rollback().await?;
-        return Ok(None);
+    match cutover_gate(&mut tx, gcs_mode).await? {
+        None => Ok(WriteGuard::Proceed(tx)),
+        Some(reason) => {
+            tx.rollback().await?;
+            Ok(match reason {
+                SkipWrite::Retired => WriteGuard::Retired,
```
Eikix: Looks a bit weird in terms

PR #3275, opened 2026-07-22
feat(test-suite): e2e additional multisig coverage

maksymsur:


test-suite/e2e/contracts/erc1271/ERC1271MultisigWallet.sol
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.24;
+
+import "@fhevm/solidity/lib/FHE.sol";
+import {IERC1271} from "@openzeppelin/contracts/interfaces/IERC1271.sol";
+import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import {E2ECoprocessorConfig} from "../E2ECoprocessorConfigLocal.sol";
+
+/// @notice M-of-N multisig ERC-1271 mock mirroring Safe's STATIC (ECDSA-only)
```
manoranjith: @maksymsur When analysing these assumptions with Claude, the mock seems to deviate from actual safe behaviour (as you have already documented in the comments). 

Afaiu: If we want to assert if off-chain components are forwarding the Safe signature without truncation, the tests in current form assert that. But this doesn't necessarily assert compatibility with safe signatures. Could you confirm if these are acceptable ?

  - **SafeMessage re-hashing**. A real Safe's isValidSignature(bytes32 dataHash, bytes) does not verify owner sigs over dataHash. It wraps it —  getMessageHash(dataHash) = an EIP-712 hash of a SafeMessage(bytes) struct under the Safe's own domain (its address + chainId) — then checks owner sigs against that. The mock verifies owners over the raw digest (:66). So the bytes a real Safe owner signs are different from what this test signs. If the SDK/KMS need to reconstruct or account for that wrapper, this test can't catch a mismatch.
  - **Safe's v-byte overloading.** Safe supports v ∈ {27,28} (ECDSA), v>30 (eth_sign), v=1 (pre-approved hash), and v=0 (nested contract-signature owner, with appended dynamic data). The mock handles only v ∈ {27,28} via ECDSA.tryRecover — so a real 2-of-3 Safe that mixes an approved-hash or contract-sig part  is unrepresentable here.
  - **Blob layout.** The mock hard-assumes length % 65 == 0 (:57). Real Safe blobs with contract-signature parts append dynamic data past the static 65-byte section, so that invariant is false for them.

test-suite/e2e/contracts/erc1271/ERC1271MultisigWallet.sol
```diff
@@ -0,0 +1,76 @@
...
+        value = FHE.asEuint64(v);
+        FHE.allowThis(value); // isAllowed(value, address(this)) == true
+    }
+
+    /// @inheritdoc IERC1271
+    /// @dev At least `threshold` parts, and the blob must be an exact multiple of
+    ///      65 bytes — the alignment requirement is a MOCK-ONLY simplification of
+    ///      the ECDSA-static case (real Safe only requires length >= threshold*65
+    ///      and ignores trailing bytes of any length). As in Safe, only the FIRST
+    ///      `threshold` parts are validated; further parts are ignored. Every
```
manoranjith: Nit: This setup is interesting. Should we also add a test case where we pass an unusual length (i.e not a multiple of 65) to assert if it passes fine through all components, reaches the mock contract and extra bytes are ignored ?

test-suite/e2e/test/sdk/unified/unifiedUserDecrypt.ts
```diff
@@ -237,10 +238,44 @@ export function computeUnifiedDigest(cfg: UnifiedConfig, req: UnifiedDecryptRequ
...
+  const domain = domainOf(cfg, chainId);
+  const message = messageOf(req);
+  const parts = await Promise.all(
+    owners.map(async (owner) => ({
+      address: (await owner.getAddress()).toLowerCase(),
+      signature: await owner.signTypedData(domain, UNIFIED_USER_DECRYPT_TYPES, message),
+    })),
+  );
+  // Lowercase-hex string order equals numeric address order for equal-length hex.
+  parts.sort((a, b) => a.address.localeCompare(b.address));
```
manoranjith:
```suggestion
parts.sort((a, b) => (BigInt(a.address) < BigInt(b.address) ? -1 : 1));
```

Nit: Safe expects sorting based on numeric order, where as `localeCompare` would sort based on dictionary order.  Here the difference doens't matter since the addresses are equal length. Hence, marking this suggestion as a Nit.

manoranjith (commented):
Thanks for the additional tests @maksymsur 

Have one question regarding assumptions we make and a couple of nit comments.

test-suite/e2e/contracts/erc1271/ERC1271MultisigWallet.sol
maksymsur: Good points — addressed in the latest commit https://github.com/zama-ai/fhevm/pull/3275/commits/2146680950790543938703fa1083802d5c7b73ee:

- **SafeMessage re-hashing** - confirmed acceptable: the wrapper is wallet-internal. Every protocol layer (SDK → relayer → gateway → connector) passes the digest and signature blob through opaquely and calls isValidSignature(digest, blob) per ERC-1271; nothing on our side reconstructs or depends on what the wallet's owners actually signed over. A real Safe signs the SafeMessage wrap of our digest inside the wallet — invisible to the pipeline. The mock documents this and signs the raw digest to stay self-contained; bytecode-level fidelity is the tracked real-Safe interop follow-up.
- **v-byte overloading** - addressed: the mock now supports v=1 pre-approved-hash parts (r = approving owner, consumed against approvedHashes via a new approveHash) and v>30 eth_sign parts (verified over the eth_sign wrap, v−4), alongside v=27/28 ECDSA. A mixed 2-of-3 blob (ECDSA part + approved-hash part) and an eth_sign-parts blob are now e2e positives. v=0 contract-signature parts (dynamic appendix) stay out of scope for the mock → real-Safe follow-up.
- **Blob layout** - addressed: the mock now uses Safe's actual rule (`length >= threshold*65`, trailing bytes of any length ignored); the alignment requirement is gone.

test-suite/e2e/contracts/erc1271/ERC1271MultisigWallet.sol
maksymsur: Thnx! Added in https://github.com/zama-ai/fhevm/pull/3275/commits/2146680950790543938703fa1083802d5c7b73ee: a positive submits two valid parts + 35 trailing junk bytes (165 B, not a multiple of 65) and decrypts successfully - proving every component forwards the unusual length untouched and the wallet ignores the extra bytes.

maksymsur:
closing in favor of https://github.com/zama-ai/fhevm/pull/3278 where work is being continued

PR #3278, opened 2026-07-22
feat(sdk): ERC-1271 user-decryption signatures

manoranjith:
Adds smart-contract-wallet (ERC-1271) user-decryption to `@fhevm/sdk`, together with the e2e coverage that exercises it end to end.

**SDK** — no new public API surface. A wallet permit is assembled off-SDK and submitted through the existing `parseSignedDecryptionPermit`, whose `signature` now accepts a variable-length blob (concatenated multisig, or the empty `0x` pre-approved-hash flow) and whose `eip712.message.userAddress` names the wallet. There is no new client method and no `signatureKind` discriminator — an earlier iteration added both; they were removed in favour of auto-detection.

Verification is a precautionary client-side port of the shared `verify_signature` crate: it auto-detects EOA vs ERC-1271 — a 65-byte `ecrecover` against `userAddress` that returns before any RPC, otherwise an `isValidSignature` STATICCALL fall-through. A normal EOA permit therefore behaves exactly as before (no on-chain call). The check is not authoritative: the KMS Connector runs the same algorithm and gates key release, so a definitive local rejection throws an `Erc1271Error` while a missing read provider or transport error warns and forwards.

**Test suite** — the acceptance coverage for the above: Safe-faithful multisig and approveHash wallet mocks, an `EncryptedValueHolder` for the realistic wallet-as-`userAddress` decrypt shape, and the SDK-client suite driving the multisig / below-threshold / approveHash flows through the public surface. Docker-compose and CLI plumbing for the `latest-main --override test-suite` run mode ride along.

Compatibility: the only public-facing change is widening the permit `signature` type from `Bytes65Hex` to `BytesHex`, a backward-compatible superset. Only the /v3 unified route is affected; the /v2 EOA-only path and the signing path (`signDecryptionPermit`, still EOA/self-only) are untouched.

Test plan:
- [x] `@fhevm/sdk` unit suite (locally 798/798), `lint` (eslint + `tsc -b --noEmit`), and `prettier:check` — all green.
- [x] e2e `erc1271-user-decryption` against a full stack. From `fhevm/test-suite/fhevm`:

```
bun install && ./fhevm-cli up --target latest-main --override test-suite
./fhevm-cli test erc1271-user-decryption
```


manoranjith:
Changes were ported in PR #3358 by @geoxel

PR #3283, opened 2026-07-22
ci: opt-in Polygon Amoy second host chain for preview env (deploy_polygon)

fegmorte:
## Summary
- Adds a `deploy_polygon` workflow input (default `false`) that spins up a **second host chain — Polygon Amoy (80002)** alongside the ETH one, sharing a single KMS key.
- Polygon is a fresh local `anvil` (no fork). Its host-contracts deploy with `--with-kms-generation false --protocol-config-source canonical`, mirroring the ETH ProtocolConfig/KMS context, so there is **no second keygen ceremony**.
- Coprocessor / poller / kms-connector go multichain via **additive overlay values** applied with an extra `-f` only when `deploy_polygon=true` (avoids `dbMigration` crashing on a `HOST_CHAIN_<i>_ACL` for a ConfigMap that doesn't exist when Polygon is off). The Polygon `chains[]` entry uses `useLegacyName: false` so its consumer/poller don't collide with the ETH ones.
- Each party gets a Polygon listener producing to the **same** per-party Redis (filtered by `--chain-id`) with its own cursor DB; the relayer and kms-connector gain a second host chain for ACL checks; chain `80002` is registered into the shared GatewayConfig.
- With `automated_tests` on, a Polygon e2e suite (network `polygonAmoy`) also runs and is folded into the report.
- Also trims duplicated Bitnami-rationale comments from the two postgres overlay values files.

## Test plan
- [ ] Manual dispatch of `preview-env-deploy.yml` with `deploy_polygon=true` and `automated_tests=true`; confirm the Polygon anvil, host-contracts (canonical mirror), listener, coprocessor consumer/poller, kms-connector second host chain, and gateway registration all come up.
- [ ] Confirm the Polygon `e2e-test-suite-polygon` workflow runs and appears in the report.
- [ ] Confirm a default dispatch (`deploy_polygon=false`) is unchanged — no Polygon resources, `dbMigration` succeeds.

## TODO
- Choose which kms version we want to deploy when we deploy from the PR

🤖 Generated with [Cursor](https://cursor.com)


ci/preview-env/host-chain/values-anvil-host-polygon-e2e.yaml
```diff
@@ -0,0 +1,28 @@
...
+image:
+  name: ghcr.io/foundry-rs/foundry
+  tag: v1.3.5
+network:
+  blockTime: "1"
+  host: "0.0.0.0"
+  # Polygon Amoy chain id. Mimics Amoy without forking its state.
+  chainId: "80002"
+  accounts: "120"
+  mnemonic: "test test test test test test test test test test test junk"
```
PanGan21, line 22: I think `slotsInAnEpoch` is missing here. I think the default for Anvil will be 32 if you don't set this, unless this was on purpose.

.github/workflows/preview-env-deploy.yml
PanGan21, line 420: Is 150 enough time to run 3 test suites within the same job?

ci/preview-env/coprocessor/values-coprocessor-e2e.yaml
```diff
@@ -190,6 +190,20 @@ snsWorker:
...
+  # (test-suite/fhevm/docker-compose/coprocessor-docker-compose.yml:
+  # --signer-type=private-key --private-key=${TX_SENDER_PRIVATE_KEY}) - the SAME
+  # coprocessor key txSender uses (both the tx-sender and attestation signer are
+  # the one on-chain COPROCESSOR_SIGNER_ADDRESS_<i-1> identity). secret.value is
+  # overridden per party via --set-string in preview-env-deploy.yml; the
+  # placeholder below (account #4, matching txSender.wallet) keeps this file
+  # valid standalone.
+  wallet:
+    enabled: true
+    secret: {value: "0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a"}
```
PanGan21, line 206: This seems like a duplicate. I see wallet entry in line 227

.github/workflows/preview-env-deploy.yml
```diff
@@ -902,21 +1046,44 @@ jobs:
...
+          kms_verifier=$(kubectl get configmap host-sc-addresses -n "${NAMESPACE}" -o jsonpath='{.data.kms_verifier\.address}' 2>/dev/null || echo "")
           require_nonempty "gw-sc-addresses[decryption.address]" "${decryption}"
           require_nonempty "gw-sc-addresses[gateway_config.address]" "${gateway_config}"
           require_nonempty "host-sc-addresses[kms_generation.address]" "${kms_generation}"
-          require_nonempty "host-sc-addresses[acl.address]" "${acl}"
           require_nonempty "host-sc-addresses[protocol_config.address]" "${protocol_config}"
+          require_nonempty "host-sc-addresses[acl.address]" "${acl}"
+          kms_verifier_args=()
+          if [[ -n "${kms_verifier}" ]]; then
+            kms_verifier_args=(--set-string "commonConfig.ethereumContractAddresses.kmsVerifier=${kms_verifier}")
```
PanGan21: Not sure if this is really needed. Could you verify that the templates also read the kmsVerifier?

ci/preview-env/README.md
```diff
@@ -206,6 +214,48 @@ self-contained `hostListener`. Per party `i`:
 > CPU/memory on the `coprocessor` nodepool, so `nb_coprocessor` > 1 multiplies the
 > cluster capacity needed. Default stays `1`.
 
+## Multichain: second Polygon host chain (`deploy_polygon`)
+
+`pr-preview-deploy.yml` takes a `deploy_polygon` input (default `false`) that adds a
```
PanGan21: It seems correct file name is `preview-env-deploy.yml`

ci/preview-env/host-chain/values-anvil-host-polygon-e2e.yaml
fegmorte: [🤖 Generated with Cursor] Good catch — the ETH host anvil sets `slotsInAnEpoch: "1"` and this one didn't, so Polygon anvil was defaulting to 32-slot epochs (finalized ~32-64 blocks behind head). Fixed in 0c1bfe175 by adding `slotsInAnEpoch: "1"` to match the ETH host.

ci/preview-env/coprocessor/values-coprocessor-e2e.yaml
fegmorte: [🤖 Generated with Cursor] Confirmed — `snsWorker.wallet` was defined twice under the same mapping (last-wins, identical values, so harmless at runtime but a dead duplicate key). Removed the second block in 0c1bfe175.

.github/workflows/preview-env-deploy.yml
fegmorte: [🤖 Generated with Cursor] You're right, it's a no-op: `charts/kms-connector/templates/_helpers.tpl` only renders `kmsGeneration`, `acl` and `protocolConfig` from `ethereumContractAddresses` — `kmsVerifier` is consumed nowhere in the chart. Removed the fetch, the `--set-string …kmsVerifier` and the misleading warning in 0c1bfe175.

ci/preview-env/README.md
fegmorte: [🤖 Generated with Cursor] Fixed in 0c1bfe175 — renamed the stale `pr-preview-deploy.yml` reference to `preview-env-deploy.yml`.

.github/workflows/preview-env-deploy.yml
fegmorte: [🤖 Generated with Cursor] Fair question. The 150m is for the deploy job; the e2e suites run as separate Argo workflows (`e2e-test-suite-fhevm-sdk` / `-relayer-sdk` / `-polygon`), so the job timeout mostly covers deploy + keygen wait rather than the suites themselves. With the default (fast) KMS params a full green matrix has completed inside this budget; only a real Default-param keygen (hours) would exceed it, and that's not the default path. Left as-is; happy to bump if you'd prefer headroom.

fegmorte:
[🤖 Generated with Cursor]

## Code Review

Reviewed the Polygon second-host-chain changes (bugs, guideline compliance, error handling) and validated @PanGan21's inline comments — all five were legitimate. Addressed in `0c1bfe175`:

| # | Severity | Item | Resolution |
| --- | --- | --- | --- |
| 1 | Medium | `slotsInAnEpoch` missing on the Polygon anvil (defaulted to 32-slot epochs) | Added `slotsInAnEpoch: "1"` to match the ETH host so finality doesn't lag |
| 2 | Low | 150m deploy-job timeout vs 3 suites | Answered inline — suites run as separate Argo workflows; budget covers deploy + keygen wait |
| 3 | Low | Duplicate `snsWorker.wallet` key | Removed the duplicate block |
| 4 | Low | `ethereumContractAddresses.kmsVerifier` set-string is a no-op | Removed dead code + misleading warning (chart never renders it) |
| 5 | Trivial | Stale `pr-preview-deploy.yml` in README | Renamed to `preview-env-deploy.yml` |

No additional bugs found beyond these; a separate Bugbot pass also came back clean. Design is sound and consistent with the existing ETH overlay patterns.

One thing to confirm before merge: CI test history on this PR is mixed (the SDK matrix went fully green once, incl. Polygon 8/8, but later runs hit the known HEAD-coprocessor vs pinned-KMS S3 ciphertext-format skew). Worth reproducing a clean matrix on the latest commit.

PanGan21 (approved):

PR #3285, opened 2026-07-22
fix(test-suite): resolve sha baselines against published image tags

Eikix:
## Problem

The orchestrated e2e workflow generates its baseline lock with `resolve --target sha --sha $BASE_SHA`, which pins every repo-owned image to the base sha without checking that the tag actually exists in GHCR.

The docker-build workflows normally guarantee it does: on every push, each component is either rebuilt or its previous image is re-tagged with the new sha. But that invariant can break — on `release/0.13.x`, the `re-tag-gw-listener-image` job of [kms-connector-docker-build run 29908109541](https://github.com/zama-ai/fhevm/actions/runs/29908109541) hit a transient DNS timeout during `docker login ghcr.io`, so `kms-connector/gw-listener:d77a041` was never published. Every e2e run whose baseline landed on `d77a041` then failed ~20 minutes in with `manifest unknown`, on PRs unrelated to the flake (#3273).

Beyond flakes, two normal behaviors also leave base shas without a complete tag set: a baseline resolved while the base push's builds are still running, and `is-latest-commit` gating skipping docker builds entirely for commits superseded before their workflows ran.

## Fix

`--target sha` now resolves against observed registry state instead of trusting the invariant:

- fetch the published GHCR tag set for each repo-owned package (same primitive the `latest-main` target already uses);
- any component missing a tag at the requested sha independently falls back to its newest published tag on the requested sha's ancestry (`commitsFrom` generalizes `mainCommits` to walk from any ref);
- each substitution is recorded in the lock's `sources` and printed by the resolve step, e.g. `CONNECTOR_GW_LISTENER_VERSION=2cde6c9 (fallback: d77a041 unpublished)`;
- a fallback deeper than **50 commits** fails resolution instead: that lag means the component's publishing is broken rather than flaked, and a fallback must not paper over it;
- components with no published tag in the walked window (brand-new images, feature-branch-only packages) keep the requested sha, matching how `latest-main` skips gating on them;
- if GitHub metadata is unavailable (offline, missing `read:packages`), resolution keeps the previous unverified behavior and records `published-image-check=skipped`.

The fallback can never skip a published feature: if a component changed at some commit at-or-before the base sha and its build succeeded, that tag exists and the ancestry walk stops there first. Since re-tags publish byte-identical images under new names, a flake-induced fallback points at exactly the image the re-tag would have produced (verified on the incident: both `gw-listener` tags share digest `sha256:c71f99b2…`).

## Trade-off

If a component's image is missing because a *real build* of changed code failed on the base branch, a shallow fallback substitutes the previous image instead of failing resolution — visible in the resolve log and lock sources, but not fatal. Detecting that case precisely would require duplicating the per-component path filters from the docker-build workflows, which would drift; the build failure itself is already loud on the base branch, and the 50-commit cap bounds how stale a substitution can silently get. Chosen behavior: keep unrelated PRs unblocked, keep substitutions visible, fail on structural lag.

## Follow-ups

- Backport to `release/0.14.x` and `release/0.13.x` after review (the resolver has drifted between branches, so these are adaptations, not cherry-picks).
- Separately, the re-tag jobs could retry `docker login`/push so a single DNS blip can't strand a branch tip in the first place.

## Test plan

- `bun run check` (tsc) clean
- `bun test src`: 307 pass, including new coverage for `findPublishedAncestorIndex`, `resolveMissingRepoTagFallbacks` (fallback, depth-cap failure, unverified-pin cases), and per-component overrides in `presetBundle`
- `compat-smoke` runs in CI (requires a Docker daemon locally)

### DIFF
--- a/test-suite/fhevm/src/resolve.test.ts
+++ b/test-suite/fhevm/src/resolve.test.ts
@@ -15,7 +15,9 @@ import {
   SIMPLE_ACL_MIN_SHA,
   SHA_RUNTIME_COMPAT_MIN_SHA,
   applyVersionEnvOverrides,
+  findPublishedAncestorTag,
   presetBundle,
+  resolveMissingRepoTagFallbacks,
   selectSupportedMainSha,
   shaRuntimeCompatFloor,
   simpleAclFloor,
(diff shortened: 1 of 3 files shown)


tawadaa (dismissed):

Eikix:
@mergifyio queue

tawadaa (approved):

Eikix:
@mergifyio queue

PR #3295, opened 2026-07-22
feat(kms-connector): add generic host chain support to kms-worker chart

haroldsphinx:
## Summary
  - Add `bnbTestnetUrl`, `bnbTestnetChainId`, `bnbTestnetContractAddresses` to `commonConfig`
  - Add `KMS_CONNECTOR_BNB_TESTNET_*` env vars into kms-worker deployment (conditional on `bnbTestnetUrl` being set)
  - Resolve BNB contract addresses via the contracts helper, same pattern as polygon
  - Bump chart version to 1.5.4
  - Backwards compatible: no BNB env vars rendered when `bnbTestnetUrl` is empty

## Test plan
  - [ ] `helm lint charts/kms-connector/` passes
  - [ ] `helm template` with BNB config renders three chains in `KMS_CONNECTOR_HOST_CHAINS`
  - [ ] `helm template` without BNB config renders only ETH + Polygon (backwards compat)


charts/kms-connector/templates/_helpers.tpl
```diff
@@ -37,15 +37,18 @@ consume with `fromYaml`.
 {{- $gw := .Values.commonConfig.gatewayContractAddresses | default dict -}}
 {{- $eth := .Values.commonConfig.ethereumContractAddresses | default dict -}}
 {{- $pol := .Values.commonConfig.polygonContractAddresses | default dict -}}
+{{- $bnb := .Values.commonConfig.bnbTestnetContractAddresses | default dict -}}
```
tawadaa, line 40: Can you also update configs for devnet

tawadaa (changes requested):

charts/kms-connector/templates/_helpers.tpl
haroldsphinx: Done

tawadaa (approved):

PR #1658, opened 2025-12-26
feat(protocol-contracts): add CLI tool for common FHEVM uses

jatZama:

PR #3302, opened 2026-07-23
test(test-suite): seed the Solana e2e scenario harness and port the confidential-transfer leg

Eikix:
First row of the `full-vertical.sh` deletion checklist. Standing plan: [architecture on #1656](https://github.com/zama-ai/fhevm-internal/issues/1656#issuecomment-5055206861), [end state + checklist on #1658](https://github.com/zama-ai/fhevm-internal/issues/1658#issuecomment-5055206711).

## WHAT

**Commit 1 — the harness seed** (`test-suite/fhevm/e2e/harness/`, independently green): `loadEnv()` builds an environment-injected `TestEnv` (urls, RFC-021 chain id, capability flags `{faucet, freshMints, fastSlots}`; every default traced to the current e2e runtime, env-var overridable; written `defaults <- overrides` so the demo-config/devnet manifest source slots in as a partial later); `personas` (capability-gated funding — throws loudly where no faucet exists; addresses derived natively via @solana/kit, no `solana` CLI subprocess); `until()` polling. Zero protocol knowledge — scenarios reach the protocol solely through `@fhevm/sdk` actions. Shared ACL/context constants hoisted to `layout.ts` (one copy; `two-holder-transfer.ts` and the harness both import it).

**Commit 2 — the first ported leg**: the `[sdk-transfer]` confidential-transfer arc moves from bash into `e2e/scenarios/`, driving the *same* orchestrator (`runSolanaTwoHolderTransfer`) through a new env-injection seam (`TwoHolderConfig`, defaults byte-identical — the 14 existing tests and the fhevm-cli profile are unchanged). Assertion set is identical by construction (same code path): Alice 1000 → transfer(400) → both handles rotate → Alice 600 / Bob 400, each a live SDK current user-decrypt. The scenario adds readiness gates (relayer `/liveness`, proof-service `/health/readiness`) the bash leg never had. The bash leg is deleted in the same commit; CI runs the scenario suite as a new step against the still-up stack, ordered before the `if: failure()` log-collection steps.

Deviation, recorded in TESTING.md: **bun:test, not vitest** — the reused orchestrator's `layout.ts` is bun-native (`import.meta.dir`); vitest remains the target if/when layout.ts is ported off bun APIs.

Also documented: the SDK gaps that keep the disclose/consume leg in bash for now (mint init, wrap, balance-state reads, burn/seal/redeem instructions, MMR proof sourcing — the last one is in flight as fhevm-internal#1721).

## Review trail

Independent adversarial round verified: the arc is fully self-provisioning (fresh mints/keypairs per run — moving it after the vertical creates no ordering flakes), CI failure propagation + log capture ordering correct, orchestrator refactor byte-neutral, commit 1 independently green, zizmor clean. Verdict APPROVE-WITH-NITS; its P3s plus a code-quality lookback (subprocess use, business-endpoint probe, filler assertion, constant drift, dead field) all folded.

## Test plan

- [x] `bun run check` / `bun run test:e2e:harness` (9) / `bun test src/solana/*.test.ts` (14) / `bash -n full-vertical.sh`
- [x] Commit 1 checked out alone: harness tests + typecheck green
- [ ] `solana-e2e` on this PR — the scenario step's first live execution IS this PR's CI run

### DIFF
--- a/.github/workflows/solana-e2e.yml
+++ b/.github/workflows/solana-e2e.yml
@@ -197,6 +197,14 @@ jobs:
       - name: Drive the Solana decrypt vertical (full-vertical.sh)
         run: bash solana/scripts/e2e/full-vertical.sh
 
+      - name: Run the Solana e2e scenario suite (SDK-driven, against the up stack)
+        # The stack from clean-e2e is still up. These bun:test scenarios drive product arcs through
+        # @fhevm/sdk Solana actions only; the confidential-transfer arc (encrypt input ->
+        # submitInputProof -> confidentialTransfer -> current user-decrypt) was moved here from
+        # full-vertical.sh's [sdk-transfer] leg. See solana/docs/TESTING.md ("Scenario layer").
+        working-directory: test-suite/fhevm
+        run: bun run test:e2e
+
       - name: Collect stack logs on failure
         if: ${{ failure() }}
         run: |
--- a/sdk/js-sdk/src/solana/actions/confidentialTransfer.ts
+++ b/sdk/js-sdk/src/solana/actions/confidentialTransfer.ts
@@ -170,7 +170,9 @@ export async function confidentialTransfer(
     createTransactionMessage({ version: 0 }),
     (m) => setTransactionMessageFeePayerSigner(feePayer, m),
     (m) => setTransactionMessageLifetimeUsingBlockhash(latestBlockhash, m),
-    (m) => setTransactionMessageComputeUnitLimit(400_000, m),
+    // A live transfer has been observed to exceed 400k CU (PDA bump search and
+    // emit_cpi! overhead vary per run); 800k keeps headroom under the 1.4M/tx cap.
+    (m) => setTransactionMessageComputeUnitLimit(800_000, m),
     (m) => appendTransactionMessageInstructions([instruction], m),
   );
   const transaction = await signTransactionMessageWithSigners(message);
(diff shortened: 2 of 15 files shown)


Eikix:
## Review

Read the full diff against the audit context I have on this branch. Clean PR — the harness is genuinely three primitives, the seam is backward-compatible, and the CI ordering is correct.

### What's good

- **Zero-protocol-knowledge rule is held.** The harness assembles endpoints and capability flags; it never touches FHE bytes. The scenario reaches the protocol only through `runSolanaTwoHolderTransfer` + SDK actions.
- **`TwoHolderConfig` seam is minimal.** `resolveConfig` defaults reproduce the local stack byte-for-byte; the 14 existing tests and the fhevm-cli profile are untouched. The export of `createRealTwoHolderDependencies` is the only new public surface.
- **Constants hoisted to `layout.ts`** — one source of truth for ACL program + user-decrypt context, shared by the orchestrator and the harness. Eliminates the drift risk I flagged in the audit (finding #2's spirit: duplicated constants across crates/packages).
- **`until()` edge-case handling** is correct: only `false`/`undefined`/`null` mean "keep waiting"; `0` and `""` are valid results. The throwing-probe recovery is the right behavior for readiness gates.
- **CI step placement** (after `full-vertical.sh`, before `if: failure()` log collection) means a scenario failure still triggers log capture. Correct.
- **Bash leg deleted in the same commit** — no double coverage. The comment in `full-vertical.sh` points to the new location.

### Nits (P3, non-blocking)

1. **`until.test.ts` line 437 test name is inverted.** "does not treat 0 or empty string as ready" — but the assertion `expect(value).toBe(0)` proves 0 *is* returned (i.e. treated as a valid result). Suggest: `"treats 0 and empty string as valid results, not as not-ready"`.

2. **`personas.fund()` still shells out to `solana airdrop`.** The PR body correctly notes address derivation is native (`@solana/kit`), but funding is a subprocess. This is fine for local (the only source today) and the capability gate throws loudly on non-local. Just noting it for the record — a future `@solana/kit` airdrop RPC call would remove the last CLI dep from the harness.

3. **`resolveConfig` default for `userDecryptContext` calls `solanaUserDecryptContext()` (no arg)**, which reads `process.env.SOLANA_UD_CONTEXT_ID` at call time. This is the same behavior as before the refactor (the old code also read the env var), so it's not a regression — but it means the "defaults reproduce the local stack" contract has a hidden env-var channel. The scenario passes the value explicitly, so the e2e path is deterministic. Fine.

### Verdict

LGTM. The harness is the right size, the seam is the right shape, and the documentation (TESTING.md scenario-layer section) is honest about what's not yet portable and why.

PR #1825, opened 2026-01-22
refactor(coprocessor): remove ACL propagate ops

dartdart26:
Since MultichainACL will eventually be removed, we don't need to propagate ACL events from the host chains to the GW anymore.

PR #3303, opened 2026-07-23
feat(solana-proof-service): semantic MMR proof lookup — ask by handle/subject, never leaf_index

Eikix:
Closes zama-ai/fhevm-internal#1721 (rescoped [here](https://github.com/zama-ai/fhevm-internal/issues/1721#issuecomment-5055283191)). One atomic cut: the mechanical `leaf_index` lookup is deleted; clients ask the two questions a product can actually ask, and the service resolves the MMR position internally.

## WHAT

| Endpoint | Key | Leaf kind |
|---|---|---|
| `GET /internal/solana/access-proof?encrypted_value&handle&subject` | `(lineage, handle, subject)` | `ZAMA_HIST_ACCESS_LEAF_V1` |
| `GET /internal/solana/public-proof?encrypted_value&handle` | `(lineage, handle)` | `ZAMA_PUBLIC_DECRYPT_LEAF_V1` |

**Invariant this buys:** leaf index/count are outputs clients pass through into wire fields (`proofLeafCount` stays the connector's retry diagnostic) — never inputs clients compute, assume, or demand from callers. dApps never see an MMR term.

- **Store**: migration adds semantic columns (`leaf_kind`, `handle`, `subject`) + a composite index; ingest populates them from the same decoded values that produce the leaf commitment (they cannot disagree). Resolution + proof build + leaf count all run in ONE `REPEATABLE READ` transaction — no tear between resolve and prove. New `.sqlx` metadata regenerated against a live DB.
- **Not-found race, both directions**: not-found while the snapshot trails the on-chain leaf count → retryable `503 lagging` (a just-sealed leaf the store hasn't ingested); not-found at parity → terminal `404 leaf_not_found`. False-404 for this-build data is structurally impossible (leaf row + count commit atomically); the liveness caveat for querying a never-existing key against a continuously-appending lineage is documented (client retry caps bound it).
- **Stale-DB safety**: a store populated before this migration would carry NULL-semantic rows that silently 404 at parity — startup now **refuses to boot** on such a store (`has_pre_semantic_leaf_rows` guard + DB-gated test), with the rebuild-from-genesis rule documented in the migration and README.
- **SDK** (vault module): `fetchSolanaPublicDecryptProof(cfg, encryptedValue, handle)` replaces the index-based fetch; the `leaf_index: 0` burned-lineage assumption and the `proof_slot` wire field are deleted; the leaf-count guard now compares two *fetched* values; 404 terminal / 503 bounded-retry.
- **Live client**: proof fetches move to the semantic endpoints; the client-side index bookkeeping (`leaf_count - 1` computation, hardcoded historical index 0, request/response index-match assertion) is deleted — #1658-aligned shrinkage. The post-fetch leaf-commitment verification is preserved, keyed on the resolved index.
- Zero references to the old route remain repo-wide.

Unblocks the discovery half of #1636 (historical decrypt needs exactly this primitive). Composes with #1825 (persisted nodes) and #1828 (chain context, already in the DTO).

## Review trail

Independent adversarial round, key verdicts: duplicate-semantic-key resolution safe (the leaf index is bound inside the commitment hash, so any resolved leaf MMR-verifies; earliest-match is correct and defensive); ingest column/commitment coherence proven same-iteration; false-404 impossible for this-build data (atomic ingest + REPEATABLE READ); the stale-DB P3 upgraded from prose to the startup guard; 404-liveness caveat documented. APPROVE-WITH-NITS, both nits folded.

## Test plan

- [x] `make check && make fmt && make clippy && make test` — 143 unit tests
- [x] `make test-db` — 18/18 DB-gated integration tests, incl. semantic resolution (kind separation, wrong-subject miss, un-ingested lineage) and the pre-semantic-rows guard
- [x] SDK: `vitest src/solana/vault` 24/24, eslint/prettier clean; live client: `NO_DNA=1 cargo check --tests` + tests 2/2
- [ ] `solana-e2e` on this PR — the live client's public + historical proof legs exercise both semantic endpoints end-to-end


Eikix:
## Review

Read the full 2400-line diff. This is the cleanest way to kill an entire class of client-side knowledge (leaf index computation). The invariant — leaf_index is an OUTPUT, never an INPUT — is held end-to-end.

### What's good

- **REPEATABLE READ atomicity.** `proof_snapshot_for_leaf` resolves the semantic key AND reads the snapshot in one transaction. The resolved index and the leaves the proof is built from cannot tear against a concurrent apply. This is the critical correctness property.

- **Not-found classification is sound.** `snapshot.leaf_count == on_chain.leaf_count` → terminal 404; any other ordering → retryable 503. The `_` arm correctly covers both `Less` (store behind) and `Greater` (store ahead of a stale on-chain read — safe: client retries, next attempt sees updated chain). The liveness caveat (nonexistent key against a continuously-appending lineage stays perpetually `Lagging`) is documented and correctly bounded by client retry caps.

- **Startup guard (`has_pre_semantic_leaf_rows`)** fails closed on NULL-semantic rows. Prevents the silent-wrong-404 trap a stale store would create. The migration comment, README operational rule, and DB-gated test all agree.

- **Ingest column/commitment coherence.** `StagedLeaf.kind/handle/subject` are populated from the same decoded values that produce the leaf commitment hash (`sha256(lineage || leaf_index || handle [|| subject])`). They cannot disagree — the semantic columns ARE the preimage.

- **`try_build_from_snapshot` now returns `CorruptStore` on build failure** instead of `LeafIndexOutOfRange`. Correct: since the index was resolved from a real row in the same snapshot, a build failure means snapshot corruption, not a client error.

- **Live client `historical_supersede_step`** now uses `proof.leaf_index` (service output) for the leaf commitment verification instead of hardcoded `0`. Strictly better — catches any resolution bug at the commitment check.

- **SDK `proofSlot` deletion** is correct. The old wire field was really the leaf count (the #1828 comment said so); `proof.leafCount` is the honest replacement.

- **Zero references to the old route remain repo-wide.** Docs, comments, code, OpenAPI — all updated in one atomic cut.

### Nits (P3, non-blocking)

1. **`has_pre_semantic_leaf_rows` is a full-table scan** (`WHERE leaf_kind IS NULL OR handle IS NULL`). Fine for POC (one-time boot check, store rebuilt from genesis). For production, a partial index on `leaf_kind IS NULL` would make this O(1), but that's premature now.

2. **`EVM_PARITY.md` and `MMR_ACL_MVP.md`** get slightly long lines from the endpoint rename. Pre-existing style in those files (bullet-point prose), not introduced by this PR.

3. **The composite index `(lineage, leaf_kind, handle, subject)\)** serves both query shapes: access-proof uses all four columns; public-decrypt uses the first three + `subject IS NULL` (B-tree indexes NULLs, so the prefix scan is efficient). Just noting the reasoning is sound — no action needed.

### Verdict

LGTM. The design is tight, the atomicity is correct, the failure taxonomy is complete (lagging / leaf_not_found / lineage_not_found / corrupt_cache), and the test coverage (143 unit + 18 DB-gated) exercises the resolution semantics directly. The independent adversarial round's verdicts (duplicate-key safety, false-404 impossibility, ingest coherence) are confirmed by the code.

PR #3307, opened 2026-07-23
ci(solana): cache workspace dep compile via cargo-chef + registry buildcache

Eikix:
## What

Adds a cargo-chef cook layer to `Dockerfile.workspace` and a registry buildcache writer for it, so the ~28-minute coprocessor workspace image build in Solana e2e drops to ~10–12 min whenever Cargo manifests are unchanged.

Branch-local experiment: all three touched areas (`Dockerfile.workspace`, `solana-images-publish.yml`, the compose generator) exist only on `feature/solana` or are additive, so the whole commit is droppable at merge-back.

## How

- **`coprocessor/fhevm-engine/Dockerfile.workspace`** — new `chef`/`planner` stages produce `recipe.json`; the builder runs `cargo chef cook` before any source `COPY`, so the dependency-compile layer is keyed only on the recipe + `CARGO_PROFILE`. sccache wiring moves after cook (and is removed entirely from `sqlx_builder`) so the cached layers are credential-independent — writer and all readers share the same key. The old `--mount=type=cache` target/registry mounts on the final `RUN` are removed: on fresh CI runners they were always empty, and they would shadow the cooked layers. The serial `cargo build -p gw-listener` solc guard stays (cargo-chef stubs `build.rs` during cook, so cook doesn't cover it).
- **`.github/workflows/solana-images-publish.yml`** — new `workspace-cache` job: a cache-only `docker buildx build --target db-migration` exporting `mode=max` to `ghcr.io/zama-ai/fhevm/coprocessor/workspace:buildcache-feature-solana`. One build suffices: `db-migration` depends on both `builder` and `sqlx_builder`, so the full graph exports under the one ref.
- **`test-suite/fhevm/src/generate/compose.ts` + `build-cache.ts`** — the existing `withRegistryBuildCache` guard that skipped `Dockerfile.workspace` builds now redirects coprocessor workspace builds to the new cache ref (kms-connector workspace stays skipped, as before — no writer for it).

`solana-e2e.yml` is untouched — it already sets `FHEVM_BUILDCACHE_TAG=buildcache-feature-solana` and the compose generator injects `cache_from`.

## Expected behavior

- Manifest-unchanged PRs (the common case): cook layer hits, only workspace source recompiles.
- `Cargo.toml`/`Cargo.lock` changes: cook recompiles deps from scratch, without sccache — accepted tradeoff, since sccache measured flat on this build (756s→844s in the branch's own timing data).
- Fork PRs benefit too: the cache key carries no credential-dependent ARGs.

## Verification

- `bun test` (test-suite/fhevm): 253 pass
- `tsc --noEmit`, `actionlint`, `zizmor`: clean
- Real cook/build behavior is validated by CI itself on this PR (no local Docker daemon); the writer job runs on merge to `feature/solana`, so the first cache hit lands on the PR after this one merges.

Reviewed by an independent fresh-context pass (cache-key writer≡reader parity, planner COPY-set completeness, workflow, compose/TS all verified; both P2 findings addressed in this commit).

### DIFF
--- a/.github/workflows/solana-images-publish.yml
+++ b/.github/workflows/solana-images-publish.yml
@@ -214,3 +214,84 @@ jobs:
             --tag "ghcr.io/zama-ai/$IMAGE:feature-solana" \
             --tag "ghcr.io/zama-ai/$IMAGE:feature-solana-$SHA" \
             --push "$CONTEXT"
+
+  # Cache WRITER for the BRANCH-LOCAL coprocessor workspace Dockerfile
+  # (coprocessor/fhevm-engine/Dockerfile.workspace). Separate from the per-image
+  # `publish` job above: that job builds the per-binary Dockerfiles (like main's
+  # release CI); this one builds the single-cargo-pass workspace Dockerfile that
+  # solana-e2e's --override compose builds use, and exports its layers so the
+  # ~28-min cold dependency compile (tfhe-rs dominates) is paid once per branch
+  # tip instead of on every PR run. READERS are solana-e2e's compose builds,
+  # which add the shared cache ref via FHEVM_BUILDCACHE_TAG (see
+  # withRegistryBuildCache + WORKSPACE_BUILDCACHE_IMAGE in test-suite/fhevm). This
+  # job is BRANCH-LOCAL and dies at merge-back to main, alongside the Dockerfile.
+  #
+  # mode=max exports intermediate (builder-stage) layers — the cargo-chef `cook`
+  # layer that holds the compiled dependencies, which is the whole point. The
+  # cook layer is keyed only on recipe.json (+ the chef layer + CARGO_PROFILE),
+  # all of which precede the per-PR-varying source/contract COPYs, so a PR that
+  # touches no Cargo manifest reuses it. A SINGLE `--target db-migration` build
+  # covers everything: db-migration's stage graph COPYs from both `builder`
+  # (/out/resolve_database_url) and `sqlx_builder`, so building it pulls the full
+  # chef -> planner -> builder (cook) chain plus sqlx_builder, and mode=max
+  # exports every one of those layers under the one shared cache ref — which is
+  # what the worker services (target tfhe-worker/host-listener/...) then read.
+  # CARGO_PROFILE is left at the Dockerfile default (release) — fhevm-cli passes
+  # none, so writer and reader must match by both defaulting. No sccache args are
+  # passed: sccache sits AFTER the cook layer in the Dockerfile, so the exported
+  # cook layer is credential-independent and shared with the credential-bearing
+  # e2e reads.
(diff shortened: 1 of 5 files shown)


Eikix:
## Review

Read the full diff. Textbook cargo-chef split with a well-reasoned cache-key design.

### What's good

- **Cook layer cache key is minimal and credential-independent.** Keyed only on `recipe.json` + chef layer + `CARGO_PROFILE`. sccache args/ENV live AFTER cook, so fork PRs (empty bucket), the publish writer (no sccache args), and credential-bearing e2e readers all share the same cooked-dependencies layer. This is the critical design decision and it's correct.

- **`--target db-migration` single-build trick.** db-migration's stage graph COPYs from both `builder` and `sqlx_builder`, so one build pulls the full chef → planner → builder (cook) chain plus sqlx_builder, and `mode=max` exports every intermediate layer under the one shared ref. Clean.

- **Old `--mount=type=cache` removal is correct.** The target/registry mounts were always empty on fresh spot runners, and they would SHADOW the cooked `target/` that lives in the layer. The comment explains this explicitly.

- **`BUILD_ID` moved after cook.** Prevents a per-PR-varying ARG from perturbing the cook cache key. `CARGO_PROFILE` is the only ARG before cook, and both writer and reader default to `release`.

- **Planner COPY set matches the builder's Rust source set.** `cargo chef prepare` only parses manifests (compiles nothing), so contract artifacts aren't needed. The set is identical to the builder's COPY lines.

- **Compose generator routing is correct.** Coprocessor workspace → shared `WORKSPACE_BUILDCACHE_IMAGE:tag`; kms-connector workspace → skipped (no writer); per-binary Dockerfiles → per-image ref. The test verifies all three cases.

- **sccache removal from `sqlx_builder`** is justified: one `cargo install sqlx-cli` doesn't benefit from sccache, and removing it makes the layer credential-independent (shared with the publish writer).

### Nits (P3, non-blocking)

1. **`cargo-chef` version pin (`0.1.77`)**: a bump reshapes the cook layer (different chef binary). The comment notes this. Just flagging that a dependabot-style bump would invalidate all existing caches — worth a note in the commit that bumps it.

2. **The `workspace-cache` job's first run** (no cache exists) pays the full cold cook (~28 min) plus the build. The 90-min timeout covers this. The comment notes "a missing cache ref on the first run is a non-fatal BuildKit warning." Correct.

3. **`volume=160gb`** on the writer runner: justified by the workspace target dir + exported mode=max layers. If the cook layer grows (more deps), this may need bumping — but that's a future concern.

### Verdict

LGTM. The writer≡reader cache-key parity holds (both default CARGO_PROFILE, neither passes sccache to cook), the planner COPY-set is complete, and the compose routing is correct. The 28→10-12 min improvement on manifest-unchanged PRs is the right tradeoff.

PR #3311, opened 2026-07-23
refactor(host-contracts): align coprocessor-upgrade tasks with repo

PanGan21:
### Summary

The coprocessor upgrade has two tasks: one that prints the call data to submit as a DAO proposal, and one that sends the same call directly with the deployer key (for devnet / test-suite). This renames and cleans them up to match the conventions the other host-contract governance tasks already use, and drops an unused workflow.

- Rename the DAO task `task:prepareCoprocessorUpgrade` → `task:buildProposeCoprocessorUpgradeCalldata`, following the same naming and paterns as the KMS-context tasks.
- Both tasks now print the `target` (ProtocolConfig) and the `calldata`. A DAO proposal needs both  and the DAO task accepts `--use-internal-proxy-address`.
- Rename the RPC env vars: to the names already used across the repo so operators reuse the endpoints they already have set.
- Remove the unused `host-contracts-prepare-coprocessor-upgrade` CI workflow and update `COPROCESSOR_UPGRADE_RUNBOOK.md`.


PanGan21:
E2E test run for multi copro multi chain: https://github.com/zama-ai/fhevm/actions/runs/34374551076

obatirou (approved):

PR #3312, opened 2026-07-23
chore(common): faster mergify pr cancelation

eudelins-zama:
### Summary

Mergify's merge queue was no longer failing fast when a docker build broke in a batch PR.
Root cause (confirmed with Mergify support): `test-suite-orchestrate-e2e-tests.yml` listened to `labeled` events, since that's how the `e2e orchestrate test` label triggered e2e runs. When cla-bot applies `cla-signed` to a merge-queue PR, it spawns a second run of the workflow where every job is skipped. Since a check reported in two check suites shows only the newest suite's status, the skipped run masked the real failure, Mergify then kept waiting on `run-e2e-tests` (up to the 5h `checks_timeout`) instead of de-queuing the batch.

### Changes

- Remove `labeled` from the `pull_request` triggers and drop the `e2e orchestrate test` label path entirely.
- Add a `workflow_dispatch` trigger (with an opt-in `run-threshold-legs` input) as the new manual way to run the orchestrated e2e suite on a PR.
- Simplify `cancel-in-progress` now that the `labeled`-event special case is gone.
- Resolve the diff base-sha via the compare API when running via `workflow_dispatch` (no PR payload available).
- Update `test-suite-e2e-threshold-post-merge.yml` comment to reflect the new manual-dispatch path.

### Impact

There is no longer a PR label to trigger e2e orchestrate tests — use `workflow_dispatch` on `test-suite-orchestrate-e2e-tests.yml` instead (optionally with `run-threshold-legs: true` for the non-gating threshold-KMS legs).

### Run example

Workflow run example: https://github.com/zama-ai/fhevm/actions/runs/30076498359

### DIFF
--- a/.github/workflows/test-suite-e2e-threshold-post-merge.yml
+++ b/.github/workflows/test-suite-e2e-threshold-post-merge.yml
@@ -5,8 +5,8 @@ name: test-suite-e2e-threshold-post-merge
 # non-gating but competed with the gating e2e shards for the same `large_ubuntu_32`
 # runner pool (observed adding 25+ min of runner-wait to the merge-critical path).
 # Running them here keeps the same per-merge coverage cadence without blocking the
-# queue. They remain available pre-merge for release branches and PRs labeled
-# `e2e orchestrate test` via test-suite-orchestrate-e2e-tests.yml.
+# queue. They remain available pre-merge for release branches and manual dispatches
+# of test-suite-orchestrate-e2e-tests.yml with `run-threshold-legs: true`.
 #
 # Image freshness: the stack resolves `latest-main`, i.e. the newest published
 # main-tagged bundle. The push-triggered docker builds for this same commit may still
--- a/.github/workflows/test-suite-orchestrate-e2e-tests.yml
+++ b/.github/workflows/test-suite-orchestrate-e2e-tests.yml
@@ -2,36 +2,33 @@ name: test-suite-orchestrate-e2e-tests
 
 on:
   pull_request:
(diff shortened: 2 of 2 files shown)


Eikix (approved):
LGTM!
You can delete the orchestrated label

PS: does this apply also to e2e label?

eudelins-zama:
> You can delete the orchestrated label
> PS: does this apply also to e2e label?

I don't understand @Eikix , the only label I'm aware of (and that the workflow was using AFAIK) is `e2e orchestrate test`.
I'll delete it if this PR get merged ofc :slightly_smiling_face: 
There's also `e2e multi kms` but it was not used AFAIU (AI generated?), will remove it as well.
But I'm not sure what you call "orchestrated label" and "e2e label"...

Eikix:
> > You can delete the orchestrated label
> > PS: does this apply also to e2e label?
> 
> I don't understand @Eikix , the only label I'm aware of (and that the workflow was using AFAIK) is `e2e orchestrate test`. I'll delete it if this PR get merged ofc 🙂 There's also `e2e multi kms` but it was not used AFAIU (AI generated?), will remove it as well. But I'm not sure what you call "orchestrated label" and "e2e label"...

Ahh interesting, it was deleted.
Historically there was an e2e test label that would trigger the e2e tests (not orchestrated workflow)

Afaik the second label was used by @isaacdecoded for multi kms workflow test

eudelins-zama:
Can't find any use of `e2e multi kms` by doing a grep on the whole repo...

eudelins-zama:
@mergifyio queue

PR #3315, opened 2026-07-23
fix(test-suite): fix race for keygen, crsgen abort

eudelins-zama:
### Problem

The `kms-generation-abort` e2e profile was flaky: the on-chain keygen/crsgen trigger and its abort were sent back-to-back, so both events could land in the same kms-connector gw-listener poll batch (30s polling). The abort could then reach KMS Core before the generation request it targets:
- KMS Core answers the abort with `NotFound`, so nothing is actually aborted.
- The ceremony dies when its MPC peers drop out, and in other parties, the request cycles pending/under_process forever and the test times out.

### Fix (test setup only)

- Sleep 3s between the trigger and the abort (`TRIGGER_TO_ABORT_SLEEP_MS`), so the two events land in different listener poll batches. 3s is safely below the ceremony duration (~24s for crsgen with Test params in CI), so the abort still arrives mid-flight.
- Set `KMS_CONNECTOR_KEY_MANAGEMENT_POLLING=1s` in the test-suite env template so the listener batch window is well under that sleep.

In production this race is essentially irrelevant (keygen ceremonies last days, and crsgen abortion is slated for removal), hence a test-setup fix rather than a protocol change.

### Tests

- Ran `./fhevm-cli test kms-generation-abort` 10 times locally: :white_check_mark: 
- CI [run](https://github.com/zama-ai/fhevm/actions/runs/30016644493): :white_check_mark: 

### Misc

- Noticed the `test-suite/fhevm/env/staging` files were not used anymore, replaced by `test-suite/fhevm/templates/env` so I removed it
- Updated/removed a few other outdated config lines in the kms-connector env file


test-suite/fhevm/env/staging/.env.coprocessor
isaacdecoded, line 1: So, all these env/staging files are unused? 😮

isaacdecoded (approved):
LGTM! 👍

eudelins-zama:
@mergifyio queue

test-suite/fhevm/env/staging/.env.coprocessor
eudelins-zama: Yes...

PR #3318, opened 2026-07-23
fix(host-contracts): require a signature check before recording an epoch activation confirmation

obatirou:
Fixes zama-ai/fhevm-internal#1782


host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -304,6 +304,10 @@ interface IProtocolConfig {
     /// @param epochId The epoch ID.
     error EpochActivationAlreadyConfirmed(address signer, uint256 epochId);
 
+    /// @notice The epoch activation payload has no keys and no CRS, so no attestation can be verified.
+    /// @param epochId The epoch ID.
+    error EmptyEpochActivationPayload(uint256 epochId);
```
isaacdecoded: nit: 
```suggestion
    error EmptyEpochActivationAttestation(uint256 epochId);
```

isaacdecoded (approved):
LGTM! 👍

host-contracts/contracts/ProtocolConfig.sol
```diff
@@ -416,6 +416,12 @@ contract ProtocolConfig is IProtocolConfig, UUPSUpgradeableEmptyProxy, ACLOwnabl
             revert InvalidKmsContext(contextId);
         }
 
+        // An empty payload skips both attestation loops below, so no signature would be verified before the vote is
+        // recorded. Reject it so `_requireExpectedSigner` always checks at least one attestation from the signer.
+        if (keys.length == 0 && crsList.length == 0) {
```
isaacdecoded: Shouldn't it be an OR, i.e., `keys.length == 0 || crsList.length == 0`?

obatirou: Currently, if _at least_ one is different from 0, there will be a signature check.
Your suggestion is stricter by requiring both to be different from 0, do you think this is more relevant?

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: fixed with 01112cddf3ea33230d668353a1d5de6f1d327eda

host-contracts/contracts/ProtocolConfig.sol
obatirou: Discussed in IRL: KMS should resign all previous keys and CRS, hence none should be empty
Done with 3eab218f974fd41427d343a3a4d7f8317232a825

isaacdecoded (approved):

PR #3328, opened 2026-07-24
docs(host-contracts): clarify KMS context-validity gates on lifecycle reads (N-05)

obatirou:
Documents the deliberately looser live-context gate on KMS lifecycle reads and the active-only signer gate invariant, resolving audit finding N-05.
Closes https://github.com/zama-ai/fhevm-internal/issues/1793

### DIFF
--- a/host-contracts/contracts/KMSGeneration.sol
+++ b/host-contracts/contracts/KMSGeneration.sol
@@ -869,10 +869,9 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
         address signerAddress,
         address txSenderAddress
     ) internal view virtual {
-        // This signer check requires an `Active` context. The sibling response-path reads
-        // (tx-sender, kmsGen threshold, node) accept any live context. They agree today because
-        // every request pins its context from `getCurrentKmsContextAndEpoch`, which returns the
-        // active context. The live gates also permit a context that is not yet `Active`.
+        // This signer check requires an `Active` context. The other reads this contract makes on
+        // `PROTOCOL_CONFIG` accept any live context. Both gates agree today because every request
+        // pins its context from `getCurrentKmsContextAndEpoch`, which returns the active context.
         if (!PROTOCOL_CONFIG.isKmsSignerForContext(contextId, signerAddress)) {
             revert NotKmsSigner(signerAddress);
         }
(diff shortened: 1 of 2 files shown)


host-contracts/contracts/interfaces/IProtocolConfig.sol
```diff
@@ -607,6 +607,8 @@ interface IProtocolConfig {
 
     /**
      * @notice Returns the kmsGen threshold for a given context.
+     * @dev Unlike the peer threshold getters, this returns a value for an `Active`, `Pending` or
```
isaacdecoded: Not sure I understand the full docs:
- What are the peer threshold getters exactly?
- It says "`Active`, `Pending` or `Created`" which are the only three possible states, right? So maybe we can just say: "this returns a value for any context's state".

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -869,6 +869,10 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
         address signerAddress,
         address txSenderAddress
     ) internal view virtual {
+        // This signer check requires an `Active` context. The sibling response-path reads
```
isaacdecoded: Sorry because it's probably me forgetting things, but what are these "sibling response-path reads" exactly?

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: * peer could have been swapped with 'others':  I was talking about `getPublicDecryptionThresholdForContext`, `getUserDecryptionThresholdForContext` etc which have a `_requireValidContext` guard which means a state `Active` only. Will rephrase 
* Yes good point will make the change, no need for explicit state names.

host-contracts/contracts/KMSGeneration.sol
obatirou: Sorry will rephrase: it is the others reads from KMSGeneration to ProtocolConfig

obatirou: Done with bb332fa0979e9f1130bbd087e118059664fb2117

host-contracts/contracts/interfaces/IProtocolConfig.sol
obatirou: Done with bb332fa0979e9f1130bbd087e118059664fb2117

isaacdecoded (approved):
Clear enough to me now! 🙏

PR #3331, opened 2026-07-24
refactor(test-suite): simplify rollout runbooks

Eikix:
## Summary

- keep two executable rollout runbooks with names that state their version range and purpose
- add a KMS-only v0.13.21 to v0.13.22 runbook that upgrades every threshold node individually
- require each upgraded KMS node to participate in a 3-of-4 quorum, then run the normal rollout profile after every step
- resolve the surrounding current stack once so only the selected component versions change during a rollout
- remove the unsupported v0.13.3 to v0.13.10 incident reproduction and the historical v0.13.0 testnet runbook
- retain bounded KMS logs, connector database status, image, and quorum evidence in failure receipts

## Runbooks

- `v0.12-to-v0.13-protocol-upgrade`
- `v0.13.21-to-v0.13.22-kms-node-by-node`

The v0.12 to v0.13 protocol runbook keeps its existing upgrade sequence and version matrix under a clearer name.

The KMS runbook upgrades nodes 1 through 4 one at a time. After each replacement, the upgraded node is forced into the serving quorum for targeted decryption before `rollout-standard` exercises the complete mixed fleet. The surrounding stack is resolved once and remains identical across the baseline and target locks.

v0.13.21 and default-configured v0.13.22 use the same fixed PRSS schedule. The compatibility thresholds introduced in v0.13.22 are intentionally omitted: v0.13.21 rejects those unknown config fields, and the thresholds are only needed when migrating from a node version that still uses the legacy schedule.

## Supporting compatibility and reliability changes

- replace and verify the exact requested relayer SDK when building a rollout e2e image
- allow legacy stacks without `ProtocolConfig`, while enforcing it in the decryption test that requires it
- keep localhost MinIO URLs reachable from the e2e container
- select legacy key-generation parameters and host pauser flags from the active release
- allow enough time for real four-party threshold preprocessing during bootstrap
- cache resolved current-stack bundles when deriving multiple rollout locks
- capture bounded KMS logs and per-party connector request status in failure receipts

## Validation

- `bun run check`
- `bun test` — 307 passing
- `bun run compat-smoke`
- `git diff --check`
- ordinary PR checks — passing
- isolated rollout dispatch on every pushed head:
  - `bc2abbff9` — [passed](https://github.com/zama-ai/fhevm/actions/runs/30381057034) in 46m52s
  - `b55f1602c` — [passed](https://github.com/zama-ai/fhevm/actions/runs/30448221259) in 50m26s, bootstrap 210s
  - `4bb6e9b2b` — [passed](https://github.com/zama-ai/fhevm/actions/runs/30454625844) in 47m49s, bootstrap 264s

A later review pass found one real defect that the earlier adversarial review missed: the host
pauser compatibility boundary was set at v0.12.5 when `task:addHostPausers` actually took
`useInternalProxyAddress` from v0.12.1, so host contracts v0.12.1 through v0.12.4 were still being
sent a flag those tasks reject. Fixed in `d922fa8f3`, together with scoping the raised KMS bootstrap
budget to pre-v0.13.20 cores only (`c81b2528b`) and removing runbook indirection that had no second
caller (`b55f1602c`, `4bb6e9b2b`). See the review comments for the full list, including four findings
that were checked and withdrawn.

The retained receipt confirms:

- the homogeneous v0.13.21 baseline passed `rollout-standard`
- nodes 1–4 were upgraded individually to v0.13.22
- each upgraded node was forced into a 3-of-4 quorum and passed targeted user decryption
- `rollout-standard` passed after every mixed-fleet checkpoint
- the final homogeneous v0.13.22 fleet passed `rollout-standard`


Eikix:
## Automated review pass

Agent-generated review of this PR (2 passes: correctness, then a dedicated pass on redundancy / unnecessary complexity). Every version claim below was checked by reading tagged sources rather than inferred from the diff.

### Verified before reviewing

- Container/DB naming used by the new diagnostics (`kms-core`, `kms-core-N`, `kms-connector[-N]-*`, `kmsConnectorDbName`) matches `kms-party.ts` and the compose `container_name`s — the regexes are correct, and party 1's unnumbered name is handled.
- `waitForBootstrap` does poll every 2s (`readiness.ts:472`), so the corrected comment is accurate and 1200 attempts is a 40-minute floor.
- `KEYGEN_PARAMS_TYPE` is already produced by `generate/env.ts:648` for threshold+Test; the gateway-sc change brings the legacy gateway keygen path in line with host-sc. Correct and consistent.
- `@zama-fhe/relayer-sdk` exports `./package.json` at 0.4.2/0.4.3/latest, so the Dockerfile's `node -p require(...)` probe resolves.
- The `protocolConfigAddress` guard narrows in the same function body, so there is no type break; `sdk/fhevm-sdk/sdk.ts` and `multiChainHelper.ts` already treat it as optional.
- `resolveVersionLock`'s resolved bundle does **not** carry `RELAYER_SDK_VERSION` (not a repo package key), so the new runbook exercises the `@fhevm/sdk` path.

---

## Correctness

### 1. High — the host pauser-flag boundary is still wrong (off by four patches)

`src/compat/compat.ts:377` moves `requiresLegacyHostPauserTaskFlag` from `<0.13.0` to `<0.12.5`. The rename actually landed at **v0.12.1**:

```
v0.11.0  useInternalPauserSetAddress      v0.12.1  useInternalProxyAddress
v0.12.0  useInternalPauserSetAddress      v0.12.2-v0.12.5  useInternalProxyAddress
```

(`git show <tag>:host-contracts/tasks/addPausers.ts`)

So HOST_VERSION `v0.12.1`–`v0.12.4` still receives `--use-internal-pauser-set-address`, which hardhat rejects as an unrecognized param. The boundary should be `[0, 12, 1]`. Nothing in-tree pins those versions any more (the two runbooks that did are deleted here), so this is latent rather than breaking — but the comment "still expect the legacy internal PauserSet flag name" is wrong for four tags, and the next person reconstructing a v0.12.1 baseline hits it. The gateway boundary at `<0.13.0` **is** correct (gateway renamed only at v0.13.0), so the host/gateway asymmetry is real — just misnumbered on the host side.

### 2. High — `network_mode: "container:fhevm-minio"` lands in the shared base compose

`docker-compose/test-suite-docker-compose.yml:6` applies to *every* scenario, not just legacy rollouts, to solve a legacy-only problem (pre-v0.13 KMS publishing `http://localhost:9000` key URLs). Consequences:

- The e2e container cannot be created unless `fhevm-minio` is running, and `refreshTestSuiteContainer` uses `composeUp(..., { noDeps: true })` (`rollout-run.ts:110`), so compose will not start minio for it. There is no `depends_on: minio` either (permitted alongside `network_mode`; only `networks`/`ports`/`hostname` conflict).
- Recreating minio orphans the e2e container's netns; the e2e container must then be recreated too, and nothing enforces that ordering.
- The e2e container loses its own network identity and shares minio's `localhost`, where 9000/9001 are now occupied.

Suggested: emit this from the compat policy as a generated override for the legacy path (alongside the existing `requiresLegacyGatewayKmsGenerationAddress`-style detection) rather than unconditionally in the base, and add a comment in the yml — the only rationale currently lives in a test name and the PR description.

### 3. Medium — `PROTOCOL_CONFIG_CONTRACT_ADDRESS` goes from required to silently optional

`e2e/test/instance.ts:72,99` drops the hard requirement and only `unifiedUserDecryption.ts:416` re-asserts it. On a modern stack where the env is accidentally missing, `createInstance` now quietly builds a non-ProtocolConfig instance and most of the suite passes with weaker coverage instead of failing fast. Prefer gating on the thing that actually varies — require it unless the stack is legacy (an explicit flag set by the rollout, or a HOST_VERSION check) — so the escape hatch is scoped and visible.

### 4. Medium — the bootstrap budget triples for every threshold scenario

`flow/up-flow.ts` (`THRESHOLD_BOOTSTRAP_ATTEMPTS`), 450 → 1200 attempts = 15 → 40 minutes. This is global: the threshold post-merge burn-in and every threshold CI job now spend 40 minutes before declaring a genuinely hung DKG dead. The comment attributes the need to "older KMS releases", so key it off `CORE_VERSION` (legacy → 1200, current → 450), or fail early on a liveness signal from the core logs instead of flat wall-clock.

### 5. Medium — the Dockerfile hotswap path is changed but not exercised by this PR's validation

`e2e/Dockerfile:37-42`. The cited rollout dispatch runs the new runbook, which pins no `RELAYER_SDK_VERSION`, and the resolved bundle does not add one — so the `if` branch never executed, and `test-suite-docker-build` is `skipping` on this PR. Two robustness points: `test "$installed" = "$RELAYER_SDK_VERSION"` hard-fails for any dist-tag or range (the workflow input is free text, and `prerelease`/`latest` are live tags), and a bare `test` fails with no message — print expected vs actual.

### 6. Medium — diagnostics fetch ~5x more log than they keep, all at once

`rollout-receipt.ts`: `diagnosticLogArgs` uses `--tail 10000` while `diagnosticLogOutput` retains 2000 lines, and `Promise.all` fans out over every `kms-core*` + `kms-connector*` container — for four-party that is ~21 containers (4 cores + init + 4x4 connector services) buffered simultaneously, on the failure path where the host is already unhealthy. Drop `--tail` to match the retained bound. Separately, joining `stdout` then `stderr` loses interleaving (kms-core logs mostly to stderr), so the receipt's log order is misleading — `--timestamps` plus a merge, or a note, would fix it.

### 7. Low — the new runbook's surrounding stack is unpinned

`versions.ts` supplies only `CORE_VERSION` and `resolveVersionLock` fills the rest from `latest-main`. Good for "only KMS changes", but a release runbook that drifts with main is not rerunnable: in a month it can fail for reasons unrelated to v0.13.22. The receipt records the resolved bundle, so it stays traceable after the fact — consider passing an explicit `target` once v0.13.22's paired release exists.

### 8. Low — deleting the incident reproduction discards the reasoning with the file

`mainnet-v0.11-to-v0.12-kms` was the only in-repo regression for the 2026-07-13 mixed-fleet decryption incident. Dropping it is a defensible call, but the *why* should survive somewhere durable (README rollouts section or a tracking issue), otherwise the next person re-derives the v0.4.2 boundary from scratch.

---

## Slop and unnecessary complexity

### 9. The new runbook keeps an abstraction that has exactly one caller

`rollouts/v0.13.21-to-v0.13.22-kms-node-by-node/run.ts` exports `KmsNodeByNodeRunbook` + `runKmsNodeByNodeUpgrade(ctx, config)`, then calls it once, in the same file, passing that module's own imports. Worth noting the deleted `mainnet-v0.11-to-v0.12-kms/run.ts` had its own inline body and never called the shared helper — so the parameterization was already single-use before this PR, and a commit titled "simplify rollout runbooks" *renames* it instead of collapsing it.

Inlining the body into `run` deletes the exported type, the `config.` prefixes, the optional-`overrides` plumbing, and the `Parameters<RolloutRunContext["up"]>[0]["overrides"]` indirection — roughly 20 lines, with no loss of capability. If a second node-by-node runbook ever appears, extracting then is cheap and will be informed by what actually differs.

### 10. `KmsCoreVersions` earns nothing

`Record<string, string> & { CORE_VERSION: string }` guards a `from`/`to` pair that is now a single-key literal, already constrained by `satisfies Env` in `versions.ts`. It disappears with #9.

### 11. `to = { ...from, CORE_VERSION: "v0.13.22" }` where `from` has one key

The spread reads as "inherit the rest" when there is no rest. `export const to = { CORE_VERSION: "v0.13.22" } satisfies Env` states the fact directly.

### 12. Three overlapping assertions where the PR replaced one, under a title the test cannot support

In `rollout-v01321-v01322-kms-node-by-node.test.ts`:

- `expect(baseline.RELAYER_SDK_VERSION).toBeUndefined()` is strictly implied by `expect(Object.keys(from)).toEqual(["CORE_VERSION"])` two lines below, which is in turn implied by the diff assertion between them.
- `const baseline = from as Record<string, string>` exists only to index a key the type says is absent — the cast is there to enable the redundant assertion.
- The title "uses the threshold CI SDK path and changes only KMS Core" promises an SDK-path behaviour the body never observes.

The pre-PR form (`expect("RELAYER_SDK_VERSION" in from).toBe(false)`) was one honest assertion. `Object.keys(from)).toEqual(["CORE_VERSION"])` plus the diff check, under a title naming only what is checked, covers it.

### 13. `createRolloutContextWithCache` is an extra module-level function for one closure variable

The split exists only because the context is returned as an object literal from an arrow expression. A block body holds the memo directly:

```ts
export const createRolloutContext = (receipt = createRolloutReceipt(), overrides = {}): RolloutRunContext => {
  const resolvedBundles = new Map<VersionTarget, VersionBundle>();
  return { /* ... */ };
};
```

Important caveat, and the reason to keep the memo itself: it is load-bearing for **correctness**, not performance. `targetUsesCache` (`bundle-store.ts:137`) returns true only for `"sha"`, so `latest-main` re-resolves live on every call (`bundle-store.ts:147`), and two live resolves mid-run could return different shas — which would silently break the "baseline and target differ only in `CORE_VERSION`" invariant the whole runbook rests on. It currently reads like a perf cache; it deserves a comment saying that. The existing `resolveCalls === 1` test does cover the property.

### 14. Two lock-writing methods whose names do not distinguish them

`writeVersionLock` (exact env) and `resolveVersionLock` (merge onto the resolved current stack) both write a lock file; the names differ in the wrong dimension. Either name for what they do (`writeExactVersionLock` / `writeVersionLockOverCurrentStack`) or collapse to one method with `{ base: "current" }`. Also `README.md:206` still documents only `writeVersionLock` — the new method is missing from the section that exists to list the ctx API.

### 15. Unexplained magic filter

`line.includes("BatchSpanProcessor.ExportError")` in `diagnosticLogOutput` has no comment saying what it is or why it is dropped (OTEL exporter noise from containers with no collector). Every comparable legacy/compat hack in this codebase carries its rationale; this one should too.

### 16. Head+tail split in `diagnosticLogOutput`

`Math.floor(maxLines / 2)` head, remainder tail, plus an omission marker and a test case. For failure triage the tail is what matters, and `--since 1h` already tends to include startup. Tail-only removes the arithmetic and the marker. Judgement call rather than a defect — flagging it as the one place the bounding logic could be halved.

### 17. Ad-hoc name regexes next to the naming authority they duplicate

`rollout-receipt.ts` imports `kmsConnectorDbName` from `kms-party.ts`, then hand-rolls `/^kms-core(?:-|$)/`, `/^kms-connector(?:-|$)/`, and `/^kms-connector(?:-(\d+))?-db-migration$/` beside it, while `kmsCoreName` (`kms-party.ts:29`) and `kmsConnectorPrefix` (`kms-party.ts:34`) already encode that convention. Two sources of truth for one naming rule, in one file, in one diff. The first two regexes also collapse into `/^kms-(core|connector)(?:-|$)/`.

### 18. `overrides: [{ group: "test-suite" }]` added without a comment

It makes the rollout build the e2e image from the working tree — which is why the Dockerfile hotswap path never runs here (#5) and means the rollout validates *local* e2e code against *released* KMS cores. Deliberate, but it needs one line saying so.

### 19. Nits

`versionSources` in the retained `v0.12-to-v0.13-protocol-upgrade/versions.ts` uses backtick template literals for two entries that interpolate nothing; the entries `fhevm=latest-main` / `sdk=@fhevm/sdk` in the new `versions.ts` are redundant or misleading, since the resolved bundle's own sources (carrying the real main sha) are prepended and no SDK is pinned. `"latest-main"` is now hardcoded in three places (`upOptions`, `writeRolloutVersionLock`, `resolveVersionLock`) and wants a constant.

---

## What is good

The new receipt helpers (`diagnosticLogArgs`, `diagnosticLogOutput`, `kmsConnectorPartyIds`) are pure and directly unit-tested, including the noise-filter-before-bounding ordering and partial-Docker-state party discovery. The `resolveVersionLock` test asserts both locks differ only by `CORE_VERSION`, which is exactly the invariant the runbook depends on, and it does so through an injectable resolve. Renames follow the descriptive-naming convention and the directory/test-file names now state their version range. Test deletions are all paired with their deleted subjects.

## Coverage and security

No CI job on this PR runs the fhevm-cli `bun test` suite or builds the e2e image (`build/test-suite`, `test-suite-docker-build`, all e2e jobs `skipping`), so findings 1 and 5 rest on local runs plus one dispatch that touched neither path. If label-gating is what skips them, a test-suite-only PR of this size probably wants them on.

No security concerns: no new secret handling, the diagnostics dump container logs and aggregate counts into a CI artifact (no ciphertext or key material in the new SQL), and `network_mode` does not widen exposure since minio already publishes 9000/9001.

Eikix:
## Review pass 3 — corrections, and the fixes now on the branch

Every finding from the earlier review was re-checked against the code and the branch history before acting. **Four did not survive**, and they are withdrawn below. The rest are fixed in `d922fa8f3`, `c81b2528b`, `b55f1602c`, `4bb6e9b2b`.

### Withdrawn

- **`depends_on: minio` (was finding 2).** Wrong suggestion — `composeFiles()` passes only the component's own compose file, so `minio` is not a defined service in the `test-suite` invocation and `depends_on` would be a hard error. The same constraint is why `network_mode` must name the raw container instead of `service:minio`. The implementation was effectively forced; only the missing explanation was a real gap, and it is now a comment. Downgrading this from High to a nit.
- **`PROTOCOL_CONFIG_CONTRACT_ADDRESS` plumbing (was finding 3).** No silent degradation is possible on a modern stack: `discovery.ts` already refuses to produce the env without the address once host contracts reach v0.13 (`requiresModernHostAddressArtifacts`). The layering — discovery enforces it, the test that reads it on-chain asserts it — is correct as written. Recorded as a comment so it does not get re-flagged.
- **Removing `npm uninstall` (was part of finding 5).** Load-bearing. `e7e2ea799` shipped the install without it and `6a0051e7d` ("replace pinned rollout SDK") added it specifically, i.e. installing over the existing dependency left the lockfile's version in place. Left alone, with a comment saying why.
- **Lowering `--tail 10000` (was finding 6).** The over-fetch is deliberate: the noise filter runs before the bound, so fetching more than is kept is what makes the retained 2000 lines signal rather than exporter spam. Only the missing rationale was real, and it is now documented.

### Fixed

**`d922fa8f3` — host pauser boundary.** Confirmed by reading `host-contracts/tasks/addPausers.ts` at every tag: exactly one param per tag, renamed at **v0.12.1** (v0.11.0/v0.12.0 take `useInternalPauserSetAddress`, v0.12.1 onward take `useInternalProxyAddress`). The boundary was `<0.12.5`, so v0.12.1–v0.12.4 were still being sent a flag those tasks reject; now `<0.12.1`. The gateway boundary stays `<0.13.0`, verified separately, and the divergence is now stated in the comment. No in-tree runbook changes behaviour (both use v0.12.5+ or latest-main), so this closes a latent trap rather than an active break.

**`c81b2528b` — bootstrap budget.** The history settles it: 450 → 600 (`d887237f2`, "allow legacy KMS bootstrap to finish") → 1200 (`6be224e8a`, "wait for legacy KMS preprocessing"), reverted to 450 in `e1171bfdf`, re-raised to 1200 in `15a4f6da3` while the runbook was still paired with the v0.12 stack. Every raise was for a legacy core, and later commits moved the runbook to v0.13.21/latest-main without re-tuning. Pre-v0.13.20 cores now get 1200; everything else is back to the historical 450, so a hung DKG on a modern threshold job reports in 15 minutes instead of 40.

**`b55f1602c` — the simplifications that survived.** The node-by-node helper was parameterised for a runbook that never called it (the deleted mainnet one had its own inline body), so it is folded back into `run()`, dropping the exported type and the `Parameters<...>` indirection. `createRolloutContextWithCache` collapses into `createRolloutContext`; the memo it held stays, now with a comment recording that it is load-bearing for correctness and not just for API calls — `latest-main` is never cached on disk (`targetUsesCache`), so a second resolve could return a different main sha and make a runbook's baseline and target locks differ in components it never meant to change. The runbook test drops the assertion implied by its neighbour and the cast that existed only to enable it, and its title now names what it checks. `versionSources` loses the two entries the resolved bundle already supplies with the real sha. The SDK version check reports expected vs actual instead of failing on a bare `test`. README documents `resolveVersionLock` and keeps the removed incident runbook's version coordinates.

**`4bb6e9b2b` — the same rule applied to the three commits above.** The pauser test looped over
v0.12.1/v0.12.4/v0.12.5, but the last two sit on the same side of the boundary and asserted nothing
extra, and the loop hid which case failed; it is back to the two adjacent cases. Four of eight
budget assertions were likewise redundant. The README paragraph recording the removed incident
runbook's coordinates is dropped too — git history and this PR already hold them, and a README
section describing a file that no longer exists only rots.

### On the CI observation

The earlier note ("no job runs the test-suite unit tests on this PR") was right about the fact and wrong about the cause — it is not label-gating gone astray. `bun run check`, `bun test src`, and `compat-smoke` are steps of `fhevm-e2e-test`, which for a direct PR requires the **`e2e` label**; this PR carries only `cla-signed`. Adding `e2e` would run all three in CI.

### Validation

`bun run check` clean, `bun test src` 307 pass / 0 fail, `bun run compat-smoke` passed, `git diff --check` clean. The `rollout-run.ts` diff is mostly re-indentation from the collapsed wrapper; it was verified inert by diffing the file against its parent with leading whitespace stripped — the only substantive changes are the constant, the three call sites, the wrapper collapse and the comment.

Two isolated rollout dispatches, one per pushed head, both green and both exercising the restored budget on a real four-party threshold baseline:

- [run 30448221259](https://github.com/zama-ai/fhevm/actions/runs/30448221259) on `b55f1602c` — success in 50m26s, `wait-for-materials done (210s)`
- [run 30454625844](https://github.com/zama-ai/fhevm/actions/runs/30454625844) on `4bb6e9b2b` — success in 47m49s, `wait-for-materials done (264s)`

That is 105 and 132 polls against a 450-attempt budget, so the modern path keeps 3.4–4.3x headroom and the previous global 1200 was over-provisioned by roughly an order of magnitude. Both receipts show the full sequence: nodes 1 through 4 upgraded individually, each forced into a 3-of-4 quorum with a passing targeted user decryption, and `rollout-standard` passing after every checkpoint.

### Not fixed

`from[key as keyof typeof from]` in the runbook test trades one cast for another rather than removing it; `expect(to).toEqual({ CORE_VERSION: "v0.13.22" })` is the cast-free equivalent given the `Object.keys(from)` assertion above it. Left for the next push rather than spending a dispatch cycle on one line.

obatirou (approved):

Eikix:
@mergifyio queue

tawadaa (approved):

Eikix:
@mergifyio queue

PR #719, opened 2025-08-13
chore(kms-connector): fix flaky tests

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/343

PR #3332, opened 2026-07-24
chore: merge main into feature/solana (rebase clean)

Eikix:
## Summary

Merges `origin/main` (827c5ee3a) into `feature/solana` (954d40302) using `--allow-unrelated-histories`. All 742 conflicts resolved by taking main as the base and selectively restoring Solana-specific additions.

## Key merge decisions

| Area | Decision |
|------|----------|
| **Alloy** | Main's 2.2.0 as base; kms-worker Solana code still on 1.x API (needs follow-up) |
| **SDK permits** | Feature/solana's unified `SignedDecryptionPermit` model; main's V1/V2 files kept |
| **ProtocolConfig.sol** | Main's `txSender` model + `CoprocessorUpgrade` features |
| **Decryption.sol** | Main's dual emission + Solana `userDecryptionRequestSolana()` |
| **KMS connector** | Main's base + Solana `HostChainAclBackend` dispatch |
| **Relayer** | Main's base + Solana `keyurl_handler`, `ed25519-dalek`, `tower-http` |
| **Coprocessor** | Main's base + Solana host-listener modules (feature-gated) |

## Solana-only directories preserved intact

- `solana/` (Anchor workspace)
- `solana-proof-service/`
- `sdk/js-sdk/src/solana/` (vault module)
- Coprocessor host-listener `solana_*.rs` modules
- KMS connector `solana_*.rs` modules
- 3 Solana DB migrations

## Known follow-ups

- [ ] Upgrade kms-worker Solana code to alloy 2.2.0 API
- [ ] Reconcile SDK V1/V2 permit files with unified model
- [ ] Verify e2e test suite passes


Eikix:
## Fold of `main` into `feature/solana` — hard-case resolution log

This branch has been rebuilt as a **true 2-parent merge** of `main` into `feature/solana` (the previous single-parent squash had dropped Solana files because the clone was shallow and had no merge base). Base `B = 96d83c05`, HEAD = `feature/solana`, theirs = `origin/main`.

**271 hand-conflicts** were resolved under a strict **union rule** (keep BOTH main's evolution AND feature/solana's Solana content; consensus/wire surfaces — PDA seeds, domain tags, discriminators, wire formats, error-enum discriminant order — kept byte-identical to `feature/solana`; error enums append-only). ~56 generated artifacts (rust_bindings, selectors, `.sqlx`, Cargo.lock, package-lock) were **regenerated from the merged sources**, not hand-merged.

The overwhelming majority of conflicts were clear-cut unions and are not listed. Below are only the **genuinely hard cases** — where `main` introduced a feature that structurally collided with Solana code, or a security/consensus/wire surface was involved. Each is `WHAT MAIN INTRODUCED → THE TENSION → HOW IT WAS RESOLVED → residual risk`.

**🚧 One upstream blocker (#10) prevents `kms-worker` from building and needs a fold in `zama-ai/kms`. One item (#9) needs a decision from the decryption-lane owner; three MEDIUM items (#3, #6, #8) want a targeted review.**

---

### 10. `kms-grpc` pin — no published KMS revision satisfies both sides  🚧 UPSTREAM BLOCKER
- **main pins** `d27c3b52` (kms **#696**, "handle epoch destruction in connection with context destruction", merged to kms `main` 2026-07-21) — adds `DestroyMpcContextResponse { repeated RequestId epoch_ids }`, which this branch's `map_destroyed_epochs` / `SendResponse::DestroyedEpochs` depends on.
- **feature/solana pins** `b14d45dc` (kms **#719**, "require typed user decrypt identity", merged to kms `feature/solana` 2026-07-16) — adds `optional bytes solana_pubkey = 11` on `UserDecryptionRequest`, which the Solana user-decrypt path sets at `decryption.rs:741`.
- **Verified against the actual protos in both checkouts:** main's rev has `DestroyMpcContextResponse` and **zero** Solana references; the solana rev has `solana_pubkey` and **no** `DestroyMpcContextResponse` at all. The KMS repo mirrors this repo's split, and `gh api compare` confirms kms `feature/solana` (HEAD `c9866731`) is **diverged** from main's rev — 22 ahead / 64 behind. **No published kms revision contains both.**
- **Resolved as:** kept **main's pin** — strictly less breakage (one call site) than taking the Solana rev, which would break main's entire epoch-destruction path. This leaves `kms-worker` with **one known compile error** (`decryption.rs:741`, `no field solana_pubkey`).
- **What unblocks it:** a **`main` → `feature/solana` fold in `zama-ai/kms`** — the exact analogue of this PR — then repin here. It is a one-line change once that rev exists. Everything else in this PR is independent of it: all other `kms-connector` crates (`connector-utils`, `gw-listener`, `tx-sender`) compile cleanly.

---

### 9. kms-connector ciphertext manager — RFC-023 integrated attestation removes Solana's bypass  ⚠️ DECISION NEEDED (HIGH)
- **Main introduced (RFC-023):** a rewrite of `kms-worker/.../ciphertext/manager.rs`. Old design ("verify all attestations, then retrieve") → per-handle `verify_and_retrieve_handle()` where attestation **consensus is integrated into retrieval** — the off-chain consensus verdict picks the *winning bucket* the ciphertext is fetched from. Attestation is no longer separable; it's always-on by construction.
- **feature/solana had:** a separable `verify_attestations()` gated by `if !self.config.enabled { return Ok(()) }` — a runtime switch to **skip attestation consensus entirely** (`ct_attestation.enabled = false`), with retrieval on an independent path. Plausibly needed where the coprocessor attestation infra isn't present (solana-e2e / local dev).
- **Tension:** architecturally incompatible — you can't "skip consensus" in main's design because consensus is what selects the retrieval bucket.
- **Resolved:** took main's integrated design wholesale. The `enabled=false` bypass is **gone** (serde ignores the now-unknown TOML field, so no parse error, but the *capability* is removed). Not restored, because that would mean re-architecting decryption-lane code — and not blocked, because main's design is the newer/hardened one.
- **Open question for the decryption-lane owner:** does Solana still need to run **without** off-chain attestation consensus?
  - **No** (Solana always has attestation infra) → main's design stands, nothing to do.
  - **Yes** (e.g. solana-e2e / local without coprocessor attestation) → a **follow-up PR** must re-add a consensus-free retrieval fallback on top of main's design, gated by a restored `enabled` flag. This is net-new work, not a merge resolution.

### 3. relayer extra-data validation — SECURITY surface (MEDIUM)
- **Main:** tagged extra-data enforcement (0x07-tagged contextId / 0x08-tagged epochId), `decode_versioned_extra_data` → `Vec<u8>`, tag-rejection tests.
- **Tension:** feature/solana added a 0x03 Solana MMR-proof path in the same logic; main also used 0x03 as its "unsupported version" negative-test example — but 0x03 is a **valid Solana version** post-merge.
- **Resolved:** 0x01/0x02 → main's tag checks; 0x03 → feature/solana's Solana MMR validator; negative tests re-based (0x04 is now the "unsupported version" example; the untagged-EVM test switched to tagged constants).
- **Review:** confirm 0x03 doesn't accidentally weaken EVM tag enforcement (Solana 0x03 bypasses tag checks *by design*, via its own validator).

### 6. sdk/js-sdk permit subsystem — main's V1/V2 split vs feature/solana's unified impl (MEDIUM)
- **Main:** versioned decryption-permit subsystem split by protocol version (V1/V2), dispatching on `getResolvedProtocolVersion` / `isSemverStrictlyBefore`.
- **feature/solana had:** a unified single-implementation permit (which the resolution guide nominally preferred).
- **Tension:** feature/solana's unified impl imports files that main's refactor **removed** from the merged tree — keeping it → dangling imports → SDK won't compile.
- **Resolved:** landed on main's V1/V2-split structure throughout (forced by compilability — it's the only self-consistent version in the tree). **Contradicts the guide's stated preference.**
- **Review:** confirm Solana user-decrypt actions still function through the V1/V2 dispatch (Solana handles resolve to a protocol version that routes to a permit shape they accept). If Solana needs a shape the split doesn't emit → rework.

### 8. coprocessor sns-worker S3 upload — dual-key storage vs attestation+reorg (MEDIUM)
- **Main:** attestation-verified S3 object-existence checks (`check_attested_object_exists`), deferred-enqueue provenance, block-reorg validation on the ciphertext upload path.
- **feature/solana had:** a dual-key storage scheme (per-ct **context-key** + backward-compat **digest-key**, with `Ct-Format` metadata) — a Solana storage/wire refactor.
- **Tension:** both refactored the SAME upload surface independently over a base that had neither; they overlap on object-existence/enqueue. Naively taking one side drops the other's guarantee.
- **Resolved:** **nested composition** — feature/solana's dual-key `check_ct128_objects_exist` wraps main's attestation `check_attested_object_exists`. Storage/wire layout follows feature/solana; main's attestation + deferred-enqueue + reorg preserved inside. Cross-checked against `sns-worker/src/tests/mod.rs` and `lib.rs`.
- **Review:** confirm attestation is enforced for BOTH context-key and digest-key objects; reorg/deferred provenance still gates enqueue with dual-key writes; `.sqlx` regen matches.

---

<details>
<summary>Clear-cut structural collisions (resolved, low residual risk) — #1, #2, #4, #5, #7</summary>

**1. gateway `InputVerification.sol` — proof-response reassembly (LOW).** `verifyProofResponse` is byte-identical both sides, but git misaligned main's RFC-023 priority-sender finalization block against feature/solana's new `verifyProofResponseSolana`. Reassembled so both functions are intact; ERC-7201 storage order preserved.

**2. gateway `IDecryption.sol` — event overload union (LOW).** main added two `UserDecryptionRequest` overloads (legacy `bytes32[]`, unified `HandleEntry[]`); feature/solana added `UserDecryptionRequestSolana`. Kept all 5 events; test assertions switched to full-signature lookups (bare-name now ambiguous). Bindings/selectors regenerated against all 5.

**4. `shared/ciphertext-attestation/consensus.rs` — Byzantine tie-break (LOW-MED).** main made winner selection deterministic (`valid_signers: usize` → `signers: HashSet<Address>`, `max_by_key(len)` → deterministic `max_by` on group size then material byte-compare). Took main's correctness fix wholesale; old count available as `.signers.len()`; forced matching `.signers.len()` in kms-worker `manager.rs`. In-memory eval types only — wire/discriminants untouched.

**5. relayer `event.rs` — dead keyurl-event block (LOW).** main's keyurl *poller* replaced the keyurl *event* mechanism (`KeyUrlEventId` + `RelayerEventData::KeyUrl` removed). feature/solana left an orphan block referencing the deleted enum → uncompilable. Dropped the orphan (matches main); kept feature/solana's `SOLANA_CHAIN_TYPE_BIT` / `is_solana_host_chain_id` helpers.

**7. sdk verified-input-proof files — investigated, not a dropped Solana feature (LOW).** main relocated `verifyInputProof.ts` / `fetchVerifiedInputProof.ts` (`actions/base/` → `coprocessor/`, alive in merged tree) and retired `createVerifiedInputProofFromRawBytes.ts` (a pre-Solana shared helper with no internal consumer; its only reference was a barrel re-export). All Solana coprocessor files retained. Accepted main's removal of the raw-bytes helper — a public-API-surface reduction only. Can be resurrected with a one-line import-path fix if an external Solana consumer needs it.

</details>

---

<details>
<summary>Defects found by the central build and fixed (worth a look — they were resolver mistakes, not upstream issues)</summary>

**Two add/add duplication defects.** Both were blocks present on *both* sides but not in the base, so git raised add/add and a resolver "unioned" them by pasting both copies. Neither was caught by the per-domain agents; the build caught both.
- `host-contracts/contracts/interfaces/IProtocolConfig.sol` — 9 error declarations duplicated, breaking `forge build` with "Identifier already declared". The copies differed only by main's signer→txSender rename; **kept main's `txSender` copy** (matches the regenerated bindings), deleted the stale one. Solana-unique `InvalidEpoch` and `EpochNotUnderActiveContext` explicitly preserved.
- `kms-connector/.../event_processor/kms_client.rs` — `request_new_mpc_context`, `request_new_mpc_epoch`, `poll_epoch_result_response` each defined twice (E0592). The copies were two API generations; **kept the newer `SendResponse`-returning copy** (also the one carrying main's `request_destroy_mpc_*` methods), deleted the 99-line stale copy.

A repo-wide sweep for this defect class found no others. The apparent duplicates in `IDecryption.sol` / `DecryptionMock.sol` are legitimate Solidity **event overloads** (distinct parameter types — verified by extracting full signatures; they compile and bind cleanly).

**Two consequences of main's alloy 1.0.38 → 2.2.0 bump in kms-worker.**
- *reqwest split.* `kms_worker.rs` imports `use alloy::transports::http::reqwest`. alloy 1.0.38 re-exported reqwest 0.12.24 (matching the workspace pin `=0.12.24`); alloy 2.2.0 re-exports **0.13.4**, so the client passed to `SolanaV2Fetcher::new` stopped type-checking. Fixed at the two Solana call sites by naming the crate explicitly (`::reqwest::Client::new()`), leaving `CiphertextManager` on alloy's client. **Flag:** the connector now compiles two reqwest versions — worth unifying in a follow-up.
- *handle-only decryption.* main's RFC-023 changed `prepare_decryption_request` to take `handles: &[B256]`, but the Solana arm still passed `&req.snsCtMaterials`. Fixed by mirroring the sibling `UserDecryptionV2` arm — the Solana event carries its own `handles: HandleEntry[]` field, so this is the intended adaptation rather than a workaround.

</details>

---

*Regeneration note:* generated artifacts (host+gateway `rust_bindings` & selectors, kms-connector `.sqlx`, all Cargo.lock, package-lock) were regenerated from merged sources and verified against CI's own conformance checks. The migration set union-merged cleanly — both main's tables (`abort_*_requests`, `kms_context/epoch_destroyed`) and feature/solana's (`solana_native_decryption_*`, `solana_user_decrypt_responses_v0`) coexist.

PR #3341, opened 2026-07-27
feat(test-suite): implement destroy context/epoch test cases

isaacdecoded:
E2E tests are passing [here](https://github.com/zama-ai/fhevm/actions/runs/32713278358). More precisely, the destroying context and epoch cases [here](https://github.com/zama-ai/fhevm/actions/runs/32713278358/job/97391250506#step:22:1009).

Closes https://github.com/zama-ai/planning-blockchain/issues/1254


test-suite/fhevm/docker-compose/coprocessor-docker-compose.yml
```diff
@@ -19,6 +19,7 @@ services:
   coprocessor-db-migration:
     container_name: coprocessor-db-migration
     image: ghcr.io/zama-ai/fhevm/coprocessor/db-migration:${COPROCESSOR_DB_MIGRATION_VERSION}
+    platform: linux/amd64
```
eudelins-zama, line 22: Curious, why was this needed?

isaacdecoded: It seems we stopped building ARM images at some point for all components. So, the test-suite deployment is failing locally on Apple Silicon users like me 🥲

eudelins-zama: Indeed, but what does this `platform: linux/amd64` changes about this?
Does this make it work on your end? (is Apple Silicon able to run AMD64 images using Apple Rosetta or Idk?) :thinking: 
If yes, good news!

Only question I would have:
- are we close to hardware limitations with this approach (is it slow, making your laptop crashes or whatever)?

isaacdecoded: > Does this make it work on your end? (is Apple Silicon able to run AMD64 images using Apple Rosetta or Idk?)

Yes, exactly, `platform: linux/amd64` tells Docker to pull the amd64 image and run it emulated. And yes, I just googled it and Docker Desktop uses Rosetta 2 for x86/amd64 images on Apple Silicon

> are we close to hardware limitations with this approach (is it slow, making your laptop crashes or whatever)?

Since these are emulated, the execution is slower than native, yes. But still I didn't have any issues while running the E2E tests locally.

test-suite/fhevm/src/resolve/presets.ts
```diff
@@ -2,7 +2,7 @@
  * Defines hardcoded companion version pins layered onto non-network targets during bundle resolution.
  */
 export const MAINLINE_COMPANIONS = {
-  CORE_VERSION: "b9087af",
+  CORE_VERSION: "nightly-25c2778",
```
isaacdecoded, line 5: @jot2re I used this nightly image here to test the changes you made to fix the session ID derivation from the previous epoch ID. But ideally we'd like to use a "stable" image version here. Do you know how to build this image, or should we wait for a new release?

jot2re: We can set a flag on any branch in the kms GH to compile the docker images; that way you can create a "stable" version.
I that that [here](https://github.com/zama-ai/kms/pull/761) so hopefully that should trigger a build you can use

dd23: I'm assuming this should go into a fix on the v0.14 branch, right?
We can make a new pre-release in any case and then decide if it should go into the stable v0.14.0 release (I would think so), or if we make a v0.14.1.
Depends on what QA prefers. 

So I guess https://github.com/zama-ai/kms/pull/752 needs to go into the v0.14.x branch @jot2re

jot2re: I did not realize this was a fix for the existing release to be published before 0.15. 
Maybe @isaacdecoded can clarify the goal?

isaacdecoded: > I did not realize this was a fix for the existing release to be published before 0.15.

Initially, for this PR, we decided that we don't need these E2E tests to be delivered in 0.14, so this PR targets `main` (meaning it will be released `>= 0.15`). But thinking twice, it seems we should include it, and the KMS core fix in 0.14.x indeed. I mean, in case we have plans to start executing key resharings from 0.14.0 and/or the QA team may need them to run E2E testing.

cc @jot2re @dd23

jot2re: I created a PR porting the KMS fix to 0.14. Maybe @dd23 can have a look? https://github.com/zama-ai/kms/pull/762

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -1,13 +1,27 @@
...
  * proves the KMS reacts to the emitted events:
  *   1. NewKmsContext — broadcast `defineNewKmsContextAndEpoch` (same committee, so no new signing
  *      keys are needed; on a spare-core cluster, a node swap with the dropped node's tx-sender
  *      stopped first), wait for the new context to become the on-chain active one, then decrypt.
  *   2. NewKmsEpoch — broadcast `defineNewEpochForCurrentKmsContext` (same context, new epoch), wait
  *      for the new epoch to activate, then decrypt again.
+ *   3. Destruction (same-committee clusters only) — the switch and rotation each retire a
+ *      still-live-but-no-longer-current entry (the baseline context, the switch's epoch). Destroy
+ *      each via `destroyKmsContext` / `destroyKmsEpoch` and prove: the current context/epoch cannot
+ *      be destroyed and unknown / already-destroyed ids revert; `KmsContextDestroyed` /
```
jot2re: What does revert mean here?

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -1,13 +1,27 @@
...
+ *      `KmsEpochDestroyed` fire and the target stops reading valid without moving the active pointer;
+ *      every party's connector forwards `DestroyMpcContext` / `DestroyMpcEpoch` to its core and
+ *      invalidates its validation cache; and the current context/epoch keeps serving. Skipped on a
```
jot2re: Do the epochs associated with a context also gets destroyed on the contract? Because they do in the KMS

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -1,13 +1,27 @@
...
+ *      node swap: the retired context's committee differs from the current party set, so dropped /
+ *      spare nodes can't decommission it (the KMS "at least one must remain" guard).
+ *   4. Recovery + abort (same-committee clusters only) — a further context switch after the destroy
```
jot2re: the wording is really hard to follow, is the point you are making here that if an abort happens, a new context/epoch switch must be done from the non-aborted context/epoch?

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -219,6 +239,299 @@ const proveSpareInQuorum = async (state: State, runDecryption: DecryptionRunner)
...
+    parties, "destroyed context invalidated in the validation cache",
+    // Postgres renders boolean::text as 'true'/'false' (not the 't'/'f' psql shows for the raw type).
+    columnQuery("kms_context", "id", "is_valid", oldContextId), ["false"],
+  );
+  // #696: DestroyMpcContext returns the epoch ids it decommissioned; the connector invalidates them.
+  // The retired context's epoch must therefore flip to invalid off the back of the context destroy,
+  // without a separate destroyKmsEpoch call.
+  await pollConnectors(
+    parties, "destroyed context cascaded to its epoch in the validation cache",
+    columnQuery("kms_epoch", "id", "is_valid", contextEpochId), ["false"],
```
jot2re, line 423: Would it make sense to add 2 epochs to the context for this test to validate that both gets invalidated? Just to make sure we test there there is no mistake s.t. only the first or last gets invalidated.

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -219,6 +239,299 @@ const proveSpareInQuorum = async (state: State, runDecryption: DecryptionRunner)
...
+      "open a second lifecycle op while a rotation is in flight",
+      "KmsLifecycleOperationInFlight(uint256,uint256)",
+      "defineNewEpochForCurrentKmsContext()",
+    );
+    await assertCurrentUnchanged(rpcUrl, configAddress, baseline, "stalling the epoch rotation");
+    console.log(
+      `[kms-context-switch] rotation stuck Pending (epoch ${pendingEpochId}); a second lifecycle op reverts KmsLifecycleOperationInFlight and the active pointer held`,
+    );
+
+    // A stuck rotation is one that reshared but never activated (the node down blocks the final
```
jot2re: So reshared but not activated is stored with a `completed` status, right?

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -219,6 +239,299 @@ const proveSpareInQuorum = async (state: State, runDecryption: DecryptionRunner)
...
+  const recovered = await waitForActivation(
+    rpcUrl, configAddress, "recovery epoch rotation after abort",
+    (current) => current.contextId === baseline.contextId && current.epochId > baseline.epochId,
+  );
+  console.log(`[kms-context-switch] recovery rotation activated: epochId ${baseline.epochId} -> ${recovered.epochId}`);
+  if (!(await runDecryption(`kms-context-switch: decrypt after abort + recovery (epochId=${recovered.epochId})`))) {
+    throw new PreflightError(
+      "kms-context-switch: user-decryption failed after aborting the stuck rotation and recovering",
+    );
+  }
```
jot2re: Would it make sense to also add a test for this at the context level? I.e.
Have Context A with epoch A1, make Context B with epoch B1 and abort during the construction of B1. Then make a new context C with epoch C1 derived from A1.

jot2re (changes requested):
Great work! Just some smaller comments!
NOTE: that I also primarily focused on the tests since I am not super involved in this part of the code base so I have a hard time to judge the other code changes.

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: Yes, we disable all epochs associated with a destroyed context via this [isValidEpochForContext](https://github.com/zama-ai/fhevm/blob/release/0.14.x/host-contracts/contracts/ProtocolConfig.sol#L726-L732) method. But also, we're iterating on the way we track the epochs of a context since it seems we can improve it according to a last minute AI audit.

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: Same as [pointed above](https://github.com/zama-ai/fhevm/pull/3341#discussion_r3828382458), there is a validation method that checks that any epoch associated with a destroyed context is disabled (see [_isValidKmsContext](https://github.com/zama-ai/fhevm/blob/release/0.14.x/host-contracts/contracts/ProtocolConfig.sol#L991-L994)). In other words, epoch validity derives from context validity, so all epochs of a destroyed context go invalid in a single write.

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: Correct, `completed` status on the KMS connector DB means the epoch result landed from KMS core independently of its activation, which is a `KMS connector - txSender <-> ProtocolConfig contract` step.

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: Rephrased and detailed a bit more the testing case [here](https://github.com/zama-ai/fhevm/pull/3341/changes/8f04c8eb647a0e8aecf3e78f9e79a88387ba4500#diff-5f6f216e91265a0cf179d166e5e837d0cfd9dd3bfb84fe0d206dbd71ee9f25a1R22-R28).

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -12,16 +12,20 @@
...
- *      rotation then recovers.
+ *      `KmsEpochDestroyed` fire and the target stops reading valid without moving the active pointer
+ *      (destroying a context also stops its epochs reading valid: `isValidEpochForContext` requires
+ *      a valid context, so the contract needs no per-epoch loop); every party's connector forwards
+ *      `DestroyMpcContext` / `DestroyMpcEpoch` to its core and invalidates its validation cache; and
+ *      the current context/epoch keeps serving. Skipped on a node swap: the retired context's
+ *      committee differs from the current party set, so dropped / spare nodes can't decommission it
+ *      (the KMS "at least one must remain" guard).
+ *   4. Recovery + abort (same-committee clusters only) — two independent checks:
+ *      a) one more context switch after the destroy must still reshare and activate, so destroying
```
jot2re: Sorry, I still find the text extremely hard to read, is the meaning the following:
"a) it must still be possible to do context switches (which includes resharing and then activation) after destroying a context. That is, destroying a context should never leave the system in a stuck state where it is not possible to construct a new context"

test-suite/fhevm/src/commands/kms-context-switch.ts
jot2re: I am sorry, but I still find the extremely hard to parse the text. I think my main issue is that the text seems to mention multiple distinct points in a single sentence with as few words as possible. This makes it very hard for a human to read and understand

jot2re (dismissed):
Looks good, although I still find some of the documentation very hard to read/follow.

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: So, in an attempt to improve readability, I figured out that some parts of the tests could be refactored based on the fixes we have worked on during this PR iterations. This means that the documentation mentioned above as hard to parse, [should now be simpler](https://github.com/zama-ai/fhevm/pull/3341/changes/8f6a931e3f7debf8d6989c346097ec2ab8affc1d#diff-5f6f216e91265a0cf179d166e5e837d0cfd9dd3bfb84fe0d206dbd71ee9f25a1) (hopefully).

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: The [KMS context switch profile](https://github.com/zama-ai/fhevm/pull/3341/changes/8f6a931e3f7debf8d6989c346097ec2ab8affc1d#diff-5f6f216e91265a0cf179d166e5e837d0cfd9dd3bfb84fe0d206dbd71ee9f25a1) is now refactored and should have some docs improvements. Please, don't hesitate to continue pushing to make things clearer.

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -1,52 +1,71 @@
...
- * so this doubles as the discovery test for whether the cluster reshares at all. Threshold-only:
- * the switch tasks act on the host ProtocolConfig.
+ *   1. Same-committee context switch (NewKmsContext). Broadcast `defineNewKmsContextAndEpoch` with
+ *      the same committee, so no new signing keys are needed. Wait until the new context becomes
+ *      the active one on chain, then decrypt.
+ *   2. Epoch rotation (NewKmsEpoch). Broadcast `defineNewEpochForCurrentKmsContext` (same context,
+ *      new epoch). Wait until the new epoch activates, then decrypt again.
+ *   3. Destruction. Steps 1 and 2 leave two entries that are still live but no longer current: the
+ *      baseline context and the epoch from step 1. Destroy them with `destroyKmsContext` /
+ *      `destroyKmsEpoch` and check:
```
jot2re: Nitpick: it seems this starts as a list of checks and then it becomes actions. Maybe have this line instead be "`destroyKmsEpoch` by proceeding with the following steps"?

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -115,42 +135,45 @@ const waitForActivation = async (
...
- * own removal. The rest of the node stays up because the reshare still needs its core (the KMS
- * core cannot yet reshare around an absent outgoing party); upgrade to a full party stop once it
- * can. Without a spare it is a same-committee reshare. Activation gates on ALL new-committee
- * signers, so the context id only advances once the spare has reshared and confirmed.
+ * NewKmsContext step, in one of two modes.
+ * Same-committee (swap = false): the new context keeps the committee, so this is a pure reshare.
+ * Node swap (swap = true): the new context drops one committee node and promotes the spare, so n
+ * stays committeeSize. The dropped node's tx-sender is stopped before the broadcast. The switch
+ * must then complete on the n − t quorum of the previous committee, without the dropped node's
+ * confirmation — a node that cannot transact must not veto its own removal. The rest of the
```
jot2re: Are we assuming that we switch at most 1 node out at the time?

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -443,22 +496,23 @@ const destroyContextAndEpoch = async (
...
 /**
- * Aborts a stuck same-context epoch rotation and recovers, exercising the single-in-flight guard and the
- * Pending-epoch destroy. Stopping one committee node's tx-sender starves the rotation of its final
- * activation confirmation (activation needs every signer), so the epoch reshares but stays Pending
- * under the still-Active context — the case destroyKmsEpoch now settles. While it is Pending, any
- * new lifecycle op reverts KmsLifecycleOperationInFlight; destroying the Pending epoch clears it; and
- * once the node is back a fresh rotation activates. Same-committee clusters only.
+ * Aborts a stuck epoch rotation and recovers from it. Stopping one committee node's tx-sender
+ * blocks that node's activation confirmation, and activation needs every signer. The new epoch
+ * therefore reshares but stays Pending under the still-Active context. This checks three things:
```
jot2re: Nitpick: it seems more like 3 actions than 3 checks.

Also from the comments it is a bit unclear whether an abort triggers a new epoch rotation or just allows the ability to do so

jot2re (commented):
Still looks good. Added a few nitpicky comments

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: Indeed. Rephrased it as [a list of checks](https://github.com/zama-ai/fhevm/pull/3341/changes/9917c8cd65043344a01cf0e006f956bd69b3cf14#diff-5f6f216e91265a0cf179d166e5e837d0cfd9dd3bfb84fe0d206dbd71ee9f25a1R17-R26).

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: [Not an assumption](https://github.com/zama-ai/fhevm/pull/3341/changes/9917c8cd65043344a01cf0e006f956bd69b3cf14#diff-5f6f216e91265a0cf179d166e5e837d0cfd9dd3bfb84fe0d206dbd71ee9f25a1R153-R160), but the current testing scenario swaps 1 dropped node by 1 spare one. It can be configured from a yaml file to increase the number of committee, dropped, and spare nodes.

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: [Numbered the list](https://github.com/zama-ai/fhevm/pull/3341/changes/9917c8cd65043344a01cf0e006f956bd69b3cf14#diff-5f6f216e91265a0cf179d166e5e837d0cfd9dd3bfb84fe0d206dbd71ee9f25a1R504-R509) to be clearer what the steps does.

test-suite/fhevm/src/commands/kms-context-switch.ts
```diff
@@ -14,22 +14,20 @@
...
- *          harmless — on-chain epoch validity derives from context validity;
- *        - the current context/epoch keeps serving.
+ *      `destroyKmsEpoch`. Each destroy must:
+ *        - be rejected for the current context and epoch, for unknown ids, and for ids already
+ *          destroyed;
+ *        - emit `KmsContextDestroyed` / `KmsEpochDestroyed`;
+ *        - make the target invalid on chain without moving the active context/epoch (destroying a
+ *          context also makes its epochs invalid, because `isValidEpochForContext` needs a valid
+ *          context, so the contract writes nothing per epoch);
+ *        - reach every committee node: its connector forwards `DestroyMpcContext` /
```
jot2re, line 24: Actually destruction on more than 2t+1 should be sufficient to consider it successful. For destruction we don't need full quorum

test-suite/fhevm/src/commands/kms-context-switch.ts
jot2re: Thank you

jot2re (dismissed):
Looks good! Thanks for the clarifying text! 

I left one comment, which is not really for this PR, but something I only just realized today. So sorry I did not mention it earlier.

obatirou:
Rebased after conflict and launched the E2E again https://github.com/zama-ai/fhevm/actions/runs/34053305696

test-suite/fhevm/src/commands/kms-context-switch.ts
isaacdecoded: At the contract level, we don't have any "consensus logic" for context/epoch destruction, i.e., destroying a context/epoch is a single call from the ACL owner (the governor). In the docs above, when it says "reach every committee node" it means the test checks that the `DestroyMpcContext` event is propagated to all KMS nodes (from connectors to cores).

jot2re: Ah I see. All good then.

jot2re (approved):
All gone!

obatirou (approved):

isaacdecoded:
@mergifyio queue

isaacdecoded:
@mergifyio refresh

isaacdecoded:
@mergifyio queue

PR #3345, opened 2026-07-27
feat(gateway-contracts): remove the priority coprocessor feature

Eikix:
Removes the priority coprocessor feature. Deactivating it was not enough: a re-activation during an incident would silently break the KMS connector's off-chain consensus check.

## Why

The priority short-circuit let one designated tx-sender finalize consensus alone, regardless of the configured threshold. That makes `coprocessorThreshold` inert and breaks the property the off-chain consensus check depends on — `threshold > numCoprocessors / 2`. Without it, two disjoint groups of coprocessors can each "reach consensus" on a different value for the same handle, with no deterministic tie-break across MPC parties.

In v0.14 the off-chain check runs in shadow mode (warning log + error counter, non-blocking), so the mismatch was tolerable. In v0.15 off-chain consensus becomes authoritative, hence deleting the code path in this release rather than leaving it dormant.

## What changes

| Contract | Version | Reinitializer |
| --- | --- | --- |
| `GatewayConfig` | v0.7.0 → **v0.8.0** | 9 → **10**, `reinitializeV9()` |
| `CiphertextCommits` | v0.5.0 → **v0.6.0** | 6 → **7**, `reinitializeV6()` |
| `InputVerification` | v0.5.0 → **v0.6.0** | 6 → **7**, `reinitializeV6()` |

Removed from the ABI: `setPriorityCoprocessorTxSender`, `removePriorityCoprocessorTxSender`, `getPriorityCoprocessorTxSender`, the `UpdatePriorityCoprocessorTxSender` event, `PriorityCoprocessorTxSenderNotRegistered` / `PriorityCoprocessorNotInNewCoprocessors` / `PriorityCoprocessorSignerChanged`, and `reinitializeV8`.

`CiphertextCommits.addCiphertextMaterial` and `InputVerification.verifyProofResponse` / `rejectProofResponse` now finalize purely on `_isConsensusReached`. The `_canFinalizeCoprocessorConsensus` helper and the priority guard in `updateCoprocessors` (with its `_findCoprocessorSigner` helper) are gone, as are the two hardhat ops tasks, the `--priority-coprocessor-tx-sender` upgrade param, and the `priority-coprocessor` e2e profile with its CI step.

## Storage

The four priority storage fields are **kept, not deleted**, so no future variable can land on a slot that still holds a live address on testnet. They keep their original names, matching the convention already used for the nine deprecated fields in `GatewayConfigStorage`.

Renaming them to `deprecated*` with `@custom:oz-renamed-from` was tried first and does not work: the annotation is not honoured for members of an ERC-7201 namespaced struct, and `validateUpgrade` rejects the layout as incompatible.

`reinitializeV9` zeroes the `GatewayConfig` slot during the upgrade rather than assuming ops already ran `removePriorityCoprocessorTxSender`. Testnet Phase 3 does remove priority in a batched tx, but making the upgrade self-healing means the ordering cannot be got wrong.

## Historical records keep their meaning

The three per-record markers in `CiphertextCommits` / `InputVerification` are **still read**, only never written again.

They exist because the raw sender list is append-only: a late coprocessor is still appended to it after finalization. A record finalized under priority reached consensus through one sender, and the `AddCiphertextMaterialConsensus` / `VerifyProofResponse` event already emitted on chain carries that sender alone. Dropping the read would make `getAddCiphertextMaterialConsensusTxSenders`, `getCiphertextMaterials`, `getSnsCiphertextMaterials` and the two proof-consensus getters return every matching submitter for those records — the contract contradicting its own emitted history, with no transaction to explain the change.

The markers are empty for everything finalized since the feature was removed, so new records fall through to plain threshold consensus.

## Upgrade ordering

The new `GatewayConfig` drops `getPriorityCoprocessorTxSender()`, which the *currently deployed* `CiphertextCommits` and `InputVerification` call on every response. Upgrading `GatewayConfig` first would make those responses revert on the missing selector until both consumers are upgraded.

In practice this window does not open: gateway upgrades reach the Gateway chain as one `GovernanceOAppSender.sendRemoteProposal` batch, relayed in a single LayerZero message and executed by the Gateway Safe in one transaction — v0.13.1 shipped its three `upgradeToAndCall` calls exactly this way. Keeping all four calls in one proposal closes it entirely. `upgrade-manifest.json` is reordered so the sequential CI upgrade job also exercises the safe order, and the constraint is documented on `reinitializeV9`, where the proposal-review checklist already sends reviewers.

Separately, before this upgrade every host-chain `InputVerifier` must already accept the full gateway coprocessor signer set at the gateway threshold. While priority mode was active the gateway emitted a single signature; afterwards it emits threshold-sized bundles unconditionally, and the host reverts on the first unrecognized signer (`InvalidSigner`, `InputVerifier.sol:465`) — it does not skip unknown signers. The mirror-image failure is `SignatureThresholdNotReached` if the host threshold is raised too early. Per host chain, at every step: host signer set ⊇ gateway signer set, and host threshold ≤ the number of signatures the gateway emits. This is the same invariant the staged multi-coprocessor rollout already relies on; the difference is that "remove priority" now happens at upgrade time, so it must hold before the upgrade lands.

## Follow-up, deliberately not in this PR

Nothing yet enforces `threshold > numCoprocessors / 2`. `_setCoprocessorThreshold` and the host `InputVerifier._setThreshold` still validate only `1 <= t <= n`, and the getter is named `getCoprocessorMajorityThreshold` while guaranteeing no majority. Removing priority mode is what unblocks enforcing it. Left out here because it is an ops-visible behaviour change: it would reject an emergency `t=1, n=5` config, and it needs a call on whether the deployed sets satisfy it first.

## Testing

- `gateway-contracts`: 374 passing, 0 failing.
- Slot clearing is tested by planting a value with `hardhat_setStorageAt` and asserting `reinitializeV9` zeroes it — no test-only contracts needed. The slot index was determined empirically and mutation-checked (pointing it one slot off makes it fail).
- The historical-marker behaviour is tested the same way: reach consensus with all three senders, plant a priority marker, assert all three views collapse to the singleton. Self-validating on the slot index — a wrong slot leaves the getters returning all three and fails.
- Verified the deletions return the two consensus contracts to their pre-feature shape. Priority entered `gateway-contracts` in exactly one commit on main's lineage (`8bf24548b`, the v0.13.1 forward-port) and nothing touched these files between it and main, so `8bf24548b^` is an exact pre-feature reference. Diffing against it leaves only the version/reinitializer bumps, the retained slots, the historical-marker reads, one comment, and repo-wide quote style — no logic differences. This caught behaviour-preserving residue that grepping for "priority" would not: nested `if`s that existed only to host the priority branch, and a scoped block added purely to stay under the stack limit.
- `test-suite/fhevm`: 308 pass, 0 fail; `tsc --noEmit` clean.
- Bindings and selectors regenerated with `make update-conformance`; `cargo check` clean on the bindings crate. ABI symbol delta verified as exactly the 8 priority symbols removed, `reinitializeV9` added, and `reinitializeV5` → `reinitializeV6` on the two other contracts.


gateway-contracts/contracts/interfaces/IGatewayConfig.sol
```diff
@@ -502,43 +471,13 @@ interface IGatewayConfig {
      * @notice Update the coprocessor threshold.
      * @dev The new threshold must verify `1 <= t <= n`, with `n` the number of coprocessors currently registered.
      *      Applied live — `InputVerification` no longer has to be paused first (the pause coupling was removed,
-     *      see fhevm-internal#1487); input verification reads this threshold on every call. While priority mode
-     *      is active the threshold is inert (the priority sender alone finalizes), so a batched
-     *      `updateCoprocessorThreshold(n)` + `removePriorityCoprocessorTxSender()` flips to true threshold
-     *      consensus atomically. Raise each host chain's threshold only after that flip, never before —
-     *      otherwise the host reverts with `SignatureThresholdNotReached`.
+     *      see fhevm-internal#1487); input verification reads this threshold on every call, so any in-flight
+     *      per-proof tally sees the new value mid-flight. Raise each host chain's threshold only after this
```
obatirou: tally -> count
I think the overall phrase can be improved to something like 
```
            .... `InputVerification` reads the threshold each time a
     *      coprocessor responds. For a proof that is still pending, the next response is compared against
     *      the new threshold. Raise each host chain threshold only after this one. If you raise a host
     *      chain threshold first, the host reverts with `SignatureThresholdNotReached`.
```

gateway-contracts/contracts/CiphertextCommits.sol
```diff
@@ -291,6 +291,32 @@ contract CiphertextCommits is ICiphertextCommits, UUPSUpgradeableEmptyProxy, Gat
...
+     * @notice Returns the coprocessor transaction senders exposed as consensus participants.
+     * @dev The raw sender list keeps growing after finalization, since a "late" valid coprocessor
+     *      still gets appended to it. For handles finalized while the priority coprocessor feature
+     *      was active, consensus was reached by that single sender and the emitted
+     *      `AddCiphertextMaterialConsensus` event carries it alone, so the historical marker is read
+     *      to keep these views consistent with that event. The marker is never written again: it is
+     *      empty for every handle finalized since the feature was removed, which fall through to the
+     *      raw list built under threshold consensus.
+     */
+    function _getAddCiphertextMaterialConsensusTxSenders(
```
obatirou: It seems this function was moved but should not

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -241,25 +243,32 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
...
+     *      `removePriorityCoprocessorTxSender` ops call having run.
+     * @dev UPGRADE ORDER: this implementation drops `getPriorityCoprocessorTxSender()`, which the
+     *      previously deployed `CiphertextCommits` and `InputVerification` call on every response.
+     *      Upgrading `GatewayConfig` before them — the order used up to v0.13.x — would make those
+     *      responses revert on the missing selector until both are upgraded. Gateway upgrades reach
+     *      this chain as one `sendRemoteProposal` batch relayed in a single LayerZero message, so
+     *      putting every `upgradeToAndCall` in one proposal closes the window entirely. If they are
+     *      ever split across proposals, upgrade `CiphertextCommits` and `InputVerification` first and
+     *      `GatewayConfig` last; the new implementations no longer read that getter.
+     * @dev Before upgrading, every host-chain `InputVerifier` must already accept the full gateway
```
obatirou: this could be more concise

gateway-contracts/contracts/GatewayConfig.sol
```diff
@@ -160,7 +160,9 @@ contract GatewayConfig is IGatewayConfig, Ownable2StepUpgradeable, UUPSUpgradeab
         mapping(uint256 contextId => bool isDestroyed) destroyedKmsContexts;
         /// @notice Whether a registered host chain has been disabled.
         mapping(uint256 chainId => bool isDisabled) disabledHostChains;
-        /// @notice The coprocessor transaction sender that can finalize consensus alone.
+        /// @dev Deprecated by the removal of the priority coprocessor feature: coprocessor consensus
```
obatirou: nit: Convention is 
`/// @dev Deprecated. <Justification>`

This applies to other natspec for deprecated field

test-suite/fhevm/src/commands/test.ts
```diff
@@ -1397,13 +1396,6 @@ export const test = async (testName: string | undefined, options: TestOptions) =
       ? undefined
       : "multi-chain-isolation requires a multi-chain topology; rerun `fhevm-cli up --scenario multi-chain` first";
 
-  const priorityCoprocessorRequirement = () => {
```
obatirou, line 1400: Could you look for dead env variables, comments, helpers in test-suite related to this feature please?

obatirou (dismissed):
ltgm 
You will need approval from gateway cc @isaacdecoded

isaacdecoded (approved):
LGTM! 👍

Eikix:
@mergifyio queue

Eikix:
@mergifyio queue

Eikix:
@mergifyio queue

PR #3349, opened 2026-07-27
feat(host-contracts): add missing scalar type check for fheMulDiv HCU

jatZama:


library-solidity/codegen/src/hcuLimitGenerator.ts
```diff
@@ -162,17 +162,17 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
     const functionName = `checkHCUFor${operation.charAt(0).toUpperCase() + operation.slice(1)}`;
```
obatirou, line 162: Most of the changes seems to be formatting, could we do this in a new PR if necessary and keep only the actual change added please?

host-contracts/contracts/HCULimit.sol
```diff
@@ -1651,6 +1651,8 @@ contract HCULimit is UUPSUpgradeableEmptyProxy, ACLOwnable {
             }
 
             _adjustAndCheckFheTransactionLimitTwoOps(opHCU, caller, factor1, factor2, result);
+        } else {
+            revert OnlyScalarOperationsAreSupported();
```
obatirou, line 1655: shouldn't this be `InvalidMulDivScalarByte` like in `FHEVMExecutor`?

library-solidity/codegen/src/hcuLimitGenerator.ts
jatZama: fixed

PanGan21: Let me do a small pr for the format and I will also add a ci step on library solidity to require formatting, which is missing and creates these problems.

PanGan21: Here is the pr: https://github.com/zama-ai/fhevm/pull/3353
@obatirou

host-contracts/contracts/HCULimit.sol
jatZama: No I think your suggestion is wrong. We should use different errors in HCULimit vs FHEVMExecutor to make debugging easier. This is already the case btw: compare fheDiv errors in HCULimit ( `OnlyScalarOperationsAreSupported` ) vs FHEVMExecutor ( `IsNotScalar` ) for the scalarByte value: those are already different on purpose, so it makes debugging easier.

obatirou: Indeed make sense

obatirou (approved):

PR #3351, opened 2026-07-27
feat(solana): add confidential vault demo dapp

Eikix:
Implements the Solana confidential-vault demo journey for #1761 and the owned local lifecycle required by #1760.

## What changed

- Adds the full shield → deposit → automatic dispatch → KMS-certified settlement → sponsored cShare claim → yield → redeem → sponsored cUSDC claim journey.
- Presents one product intent for shielding and depositing, while announcing and tracking the two current wallet transactions before submission.
- Supports repeat deposits, percentage-based redemption before or after yield, repeatable independent fast-forwarding, and correct full-redemption state.
- Uses Wallet Standard for Phantom and a deterministic burner reference path. Fee SOL and mock USDC funding are automatic and idempotent.
- Simulates every unsigned transaction before opening the wallet, then simulates the signed transaction with signature verification before journaling or submission.
- Uses one-time `signMessage` authorization for current Solana user decryption behind product-level reveal actions. The UI remains compatible with the future reusable-permit adapter.
- Keeps protocol construction, proofs, derivation, settlement verification, confidential transfers, and user decryption in public Solana SDK actions; demo orchestration remains local to the dApp.
- Adds compact success feedback and expandable activity instead of permanent notification blocks.
- Adds a collapsible Developer Evidence surface with confirmed status, slot, actual compute consumption, signatures, Explorer links, top-level program IDs, ciphertext handles, encrypted-value accounts, and recorded wallet-to-cleartext timings.
- Adds optional boot-owned Prometheus and Jaeger services through `bun run demo serve --observability`. Ciphertext handles correlate dApp evidence with connector traces through ACL validation, ciphertext verification, KMS RPC, response persistence, and forwarding.
- Adds `bun run demo doctor|up|serve|status|logs|reseed|down` with exact per-boot ownership, strict collision handling, supervised reseed, partial-start recovery, and exact teardown.
- Keeps all multi-arch services native on Apple Silicon. Only centralized `kms-core` is pinned to `linux/amd64`, because its current image has no arm64 manifest.
- Keeps the boot capability in a mode-0600 server-side file. The stable browser URL is always `http://127.0.0.1:5173/`.

## Protocol and demo boundaries

Real local-stack behavior includes browser TFHE encryption, input proofs, coprocessor attestation, Solana transactions, encrypted batch accounting, KMS public-decryption certificates verified on-chain, confidential-token claims, ACL-checked user decryption, KMS signcryption, and browser reconstruction.

Demo-only behavior includes mock USDC, automatic funding, seeded actors, the local keeper, the illustrative 7% APY, repeatable yield donation, and centralized KMS topology.

The connector trace is intentionally component-scoped. It represents KMS Core as an external RPC boundary and is correlated to earlier coprocessor work by ciphertext handle; it is not presented as one browser-to-coprocessor-to-KMS distributed trace.

## Verification

- dApp: 127 tests, TypeScript, oxlint, production build, and live Vite hot reload.
- Demo tooling: 66 tests total, including 39 focused lifecycle ownership, collision, observability, and Apple Silicon tests.
- KMS worker: 103 library tests, including Solana authorization and latency instrumentation.
- Connector: focused ciphertext-handle trace-attribute test.
- Formatting: targeted Rust formatting and `git diff --check`.
- Live owned stack: 38 exact containers, 8 volumes, one network, validator/listener/faucet/dApp supervision, and all readiness gates healthy.
- Live Browser:
  - repeat deposit completed and reopened the form with cleared input and success toast;
  - confirmed CU metadata for shield, deposit, dispatch, settlement, and claim;
  - redemption completed before any fast-forward through automatic settlement and sponsored cUSDC claim;
  - two independent fast-forwards changed share price from 1.00 → 1.07 → 1.14;
  - real user decryption returned 100 cShares and recorded the exact ciphertext handle and wallet-to-cleartext timing.
- Live observability:
  - the copied cShares handle resolved to a Jaeger trace spanning gateway intake, Solana ACL checks, ciphertext digest verification, KMS Core request/result polling, response persistence, and forwarding;
  - Prometheus target health is checked continuously;
  - percentile guidance requires a visible sample count and at least 20 successful observations.
- Independent adversarial reviews covered async ordering, sponsored claims, repeat journeys, evidence accuracy, observability scope, and lifecycle ownership.

## Remaining manual rehearsal

- Complete one final human-operated Phantom recording through redemption. Phantom may show its remote scanner warning for a private validator; the dApp independently simulates locally before and after signing and labels the localnet developer boundary.


Eikix:
*Agent-assisted review (Claude Code), posted at the author's request. Findings below were read from the code, not from the PR description; the crypto claim was re-derived independently.*

## How this was reviewed

Three passes, deliberately in this order:

1. **Follow the blast radius, not the line count.** 163 files, but only a handful leave the demo. Sorted by "what does production see?": `kms-connector/`, `relayer/`, `solana/programs/`, the shared `test-suite/fhevm/src` and compose files, the Dockerfiles, then the SDK, then the dApp.
2. **Verify the crypto independently rather than trusting either side's test.** Both the Rust and TS suites assert the same vectors, so agreeing with each other proves nothing about the spec. The canonical commitment was rebuilt from the module docs and hashed out-of-band: `25a475893b5f352d6b0c45eac781c83523a07808f3a7184a57ed7e4f5ed7ac46` (no proof tail) and `3008dc63526a0c181d463c5d32b651945cdfdcde0a2e919fd8509aded4db1685` (with MMR tail). Both match Rust and TS byte-for-byte. The scheme change is correct.
3. **Read the money path and the trust boundaries in full**, not the diff: deposit / redeem / settlement / reveal, the wallet session, the dev-server endpoints, the lifecycle owner's bring-up and teardown.

## The flow this PR installs (and why it reaches so far)

The causal chain is the whole story, and it starts with a wallet constraint:

> Phantom refuses `signMessage` on arbitrary binary (those bytes could be a serialized transaction) → the v2 raw preimage can never be signed by a real wallet → wrap the same canonical commitment in a UTF-8 digest envelope (`v3`) → bump the domain tag → bump the relayer's wire discriminator to `-v2` → change the connector's verifier.

So a demo requirement propagated a protocol-visible change into three production components. Everything else layers on top: the program gains a `payer`/`owner` split so the keeper can sponsor accounts mid-journey; the SDK grows harvest/metrics/batch-pinning; the dApp orchestrates; and a new lifecycle owner (`bun run demo up|serve|status|logs|reseed|down`) takes exclusive ownership of the local stack, replacing `clean-e2e.sh` as the CI lane's bring-up.

---

## Risk register

### 🔴 Decide before merge

| # | Finding |
|---|---|
| 1 | **The v3 wallet prompt is unverifiable.** The user sees a fixed sentence plus an opaque sha256 — not the handle, the transport `publicKey`, the expiry, or the requesting page. A hostile page can present a byte-identical-looking prompt whose commitment names *its own* transport key over the user's handles. v2 wasn't verifiable either, so this is not a regression — but v3 is the moment to fix it, and the digest envelope makes it cheap: put security-relevant fields in the cleartext envelope *and* keep them committed in the digest (`Reveal: 1 value / Expires: … / Key: <8 hex> / Request: <digest>`). If deferred, it deserves a recorded decision. |
| 2 | **Release coordination.** `validations.rs:496` asserts `…-v1` is now *rejected* — there is no dual-accept window, so relayer + connector + SDK must ship atomically. Relatedly, `solana_extra_data.rs` still says "the relayer, the client SDK, and **kms-core** must reproduce this message byte-for-byte" while the same file says verification lives in kms-worker. If kms-core really has an implementation this is an uncoordinated cross-repo break; if not, delete the claim. |
| 3 | **Three version numbers for one scheme**: domain tag `v3`, wire attestation type `v2`, envelope `Version: 1`. Painful during incident triage — align them or document the mapping in one place. |
| 4 | **`# syntax=docker/dockerfile:1` removed from three production Dockerfiles** (`relayer/`, `relayer-migrate/`, `connector-db/`) for local Apple Silicon ergonomics. Release image builds now depend on whatever frontend the building daemon has, instead of a pinned one. |
| 5 | **Local tag shadowing.** `clean-e2e.sh`'s `ensure_native_rust_builders` builds an arm64 image and tags it `ghcr.io/zama-ai/fhevm/gci/rust-glibc:$version`. Label-guarded, but every later build on that machine — including unrelated work — silently prefers the local image. A distinct local tag plus a build-arg override costs nothing. |
| 6 | **`coprocessor-cargo-test/gw-listener` and `/sns-worker` are failing.** Logs have expired, so the cause is unknown. Neither crate is touched here, so likely flake or pre-existing — but that needs confirming rather than assuming. |

### 🟠 Worth fixing in this PR

| # | Finding |
|---|---|
| 7 | **Every dev-server endpoint is gated only by forgeable headers — including the one that serves a private key.** All eight handlers in `demoServerPlugin.ts` call `hasDemoPageContext` and nothing else; `authorizeDemoHeaders` is never invoked there. `/api/demo-session` (:472) returns Alice's 64-byte secret key; `/api/demo-operator` (:354) and `/api/demo-harvest` (:434) perform keeper-signed actions. Because `Sec-Fetch-*` are forbidden headers, this is a sound *anti-CSRF* gate — it stops a hostile web page. It is not authorization: any local process sending four fixed headers to `127.0.0.1:5173` gets all of it. Note the browser deliberately never holds the boot token (`authorizedServiceEnv` passes only the token *file path*), so "require the token" isn't available — either state the boundary in the README, or inject a per-page nonce at load. |
| 8 | **FHE public key + CRS are injected unverified.** `encryptionKey.ts:59-76` trusts server-supplied ids and an ETag-derived "fingerprint" and never checks that the bytes hash to the advertised id — after calling `clearSolanaEncryptionKeyCache` to discard the SDK's own verified copy. Key substitution is the exact attack class the v3 binding exists to prevent. Verify `hash(bytes) == id` client-side; and reconsider shipping the cache-clearing helper as *public* SDK surface. |
| 9 | **Throw-after-sign.** `writeActiveDeposit` (deposit.ts:167) and `writeActiveRedeem` (redeem.ts:65) call `localStorage.setItem` unguarded, and both run inside `onTransactionSigned` (deposit.ts:519, redeem.ts:396). In a storage-blocked profile they throw *after* the wallet approved and *before* submission, and the SDK hook contract aborts the send. The sibling `writeShieldJournal` (deposit.ts:147) does guard — the inconsistency looks unintentional. Worth documenting on the SDK type that a throwing hook means "signed, not sent". |
| 10 | **`reads.ts` drift guard was loosened** from strict `offset === body.length` to "trailing bytes are a multiple of 32". The realloc-high-water rationale is sound, but a future layout shift of any multiple of 32 now passes silently — the popcount invariant only catches it when `peaks`/`leafCount` land wrong. Name that residual gap, or derive the exact expected trailing capacity. (`trailingCapacity < 0` is unreachable.) |
| 11 | **Production observability reshaped for local needs.** `DECRYPTION_LATENCY_BUCKETS` extended to 60/120/300s invalidates anything built on the old layout. `tracing-docker-compose.yml` lost the `fhevm_default external: true` network, Jaeger's OTLP `4317/4318`, the container names, and the comment documenting standalone use — the old documented path now yields a Prometheus that cannot resolve its targets. `platform: linux/amd64` dropped from relayer/relayer-db/relayer-migration/solana-proof-{db,service}: confirm every tag in use publishes an arm64 manifest, or Apple Silicon hard-fails where it previously emulated. |

### 🟡 Follow-up is fine

- **Unbounded scan on the connect path**: `hasClaimedDeposit` (deposit.ts:291-307) walks every batch from 0 with 1–2 sequential RPC calls each, on every connect. `redeem.ts:31` bounds the equivalent with `RECENT_BATCH_SCAN_LIMIT = 32`.
- **Mixed commitment**: `batchProvisioning.ts` pins `confirmed` everywhere except two bare `getCurrentBatch(rpc, roots)` calls made right after writing at `confirmed` — which can read back staler state. Those two calls are also identical duplicates; hoisting `batch` once fixes the race and the round trips.
- **`writeRegistry` uses a fixed `${registryPath}.tmp`** while every other atomic write here suffixes pid + UUID (`authorization.ts:48`). Safe today only because provisioning is serialized.
- **Asymmetric confidentiality**: redeem carefully redacts (`redactRedeemPosition`, `writeActiveRedeem` hardcodes `'0'`) so the confidential amount never hits disk; deposit persists the plaintext amount to localStorage (deposit.ts:173). Not rendered anywhere, but inconsistent in the artifact whose purpose is demonstrating confidentiality.
- **Host-dependent generated artifacts**: `compose.ts:52` pins kms-core to amd64 only on darwin/arm64, so the generated override differs per machine. `layout.ts:109` throws during module *import* on a malformed env var.
- **`clean-e2e.sh` is now a second bring-up path with no CI coverage**, while `solana/README.md`, `solana/docs/TESTING.md`, and `solana-proof-service/README.md` still point people at it. That will rot silently.
- **`materialize-test-sdk.sh` copies the SDK into consumer `node_modules`** — the copy goes stale the moment the SDK is rebuilt without re-running it. Six commits on this branch fight this resolution problem; the packaging approach looks like the actual bug.
- **The browser calls the proof service's `/internal/…` route** (settlement.ts:107), requiring permissive CORS on an internally-namespaced endpoint.
- **Scope**: extracting the v3 bump + relayer rename into its own PR would make the protocol-visible change reviewable by its owners and independently revertable.

---

## Credit where it's due

- **The v3 mechanism is right**, and the cross-impl parity is genuinely locked: independent re-derivation matched both vectors exactly, and the rename left no stale `signing_preimage` / `…-v1` references anywhere.
- **The `payer`/`owner` split is safe for the right reason** — init cannot assign a nonzero balance, and the PDA, stored owner, and ACL audience all derive from `owner`. `mollusk_initialize_token_account_allows_distinct_sponsor_and_owner` asserts exactly that, *including* that the payer does not become an ACL subject and that re-init fails.
- **The dApp's async discipline holds up under adversarial reading.** The reducer drops any update whose generation doesn't match (`demoStateReducer:124`) *and* every async read carries its own monotonic request counter, so stale responses cannot overwrite fresh state. Cross-tab exclusion via Web Locks with `ifAvailable`.
- **`readExactMessageSignature` (demoSession.ts:173) verifies byte-for-byte that the wallet did not modify the v3 message** before accepting the signature — precisely the right check for `modifyAndSignMessages`.
- **The lifecycle owner is careful where it counts.** `ownedProcessBaseEnv` allowlists 14 env vars instead of inheriting the parent environment. The boot token never reaches the browser and nothing is `VITE_`-prefixed. Teardown is guarded twice — compose-project prefix (lifecycle.ts:618) *and* the exact recorded ID set — so a foreign resource cannot be selected. The supervisor socket pins uid, dev, and ino at 0700/0600. Partial-failure recovery records only identities it can prove.
- **`readiness.ts` is a real fix to shared tooling**: the old `handle ([a-zA-Z0-9]+)` fallback could match a CRS or FHE key handle instead of the signing key; it's now an exact `SigningKey/<64hex>` match plus an S3 lookup for restart cases, with the reasoning in the comment.
- **`setup-solana-side.sh` no longer blind-`pkill`s** `solana-test-validator` / `solana_host_listener` in lifecycle mode — it refuses instead.
- **The rewired lane passes**: `solana-e2e/full-vertical-reconstruct` green at 48m41s, which is the meaningful signal for the bring-up rewrite.

## Coverage boundary

Read in full: the Rust and SDK signing path, the relayer plumbing, the program change and its tests, the shared `test-suite/fhevm/src` diffs, the three shell scripts, `deposit/redeem/settlement/revealShares/demoSession/encryptionKey/demoApi/batchProvisioning/mutationLock/sendTransaction/transactionSimulation`, the controller's state machine, `authorization.ts`, `faucet.ts`, and ~700 lines of `lifecycle.ts` (up, teardown, env, collision, preflight, supervisor).

Not read closely: the React components and CSS (~2,200 lines), the ~3,000 lines of new tests, the remaining ~1,600 lines of `lifecycle.ts` (status, logs, reseed, doctor internals), and the lockfiles.

Eikix:
**Correction to the review above**, after an adversarial pass over its own findings. Two are withdrawn, one was mis-tiered, two get stronger, and one premise was inherited rather than verified.

**Finding 7 is withdrawn as a security finding.** The demo keypairs (`alice.json`, `bob.json`, `keeper.json`, `mint-authority.json`) are committed to the repository, and the dev server reads Alice's key from the checkout to serve it. So `/api/demo-session` exposes nothing a caller doesn't already have — reaching `127.0.0.1:5173` implies being on a machine with the repo — and the keeper endpoints grant no privilege a local process lacks, since it could sign keeper transactions directly. The "hands out a private key" framing was wrong. What remains is narrow and non-security: the PR text describes a mode-0600 capability model that does not actually cover the dApp's own API surface. That's a documentation-accuracy nit (🟡), not 🟠.

**Finding 8 is withdrawn.** It claimed the demo replaces a *verified* key-acquisition path with an unverified one. There is no verified path: no hash, digest, or id-vs-bytes check exists anywhere in `sdk/js-sdk/src/core/key/`. The SDK's normal fetch trusts the relayer/MinIO transport exactly as much as the demo's injection does, so `clearSolanaEncryptionKeyCache` is a benign cache-eviction helper for local validator restarts, not a provenance bypass. The residual observation — that nothing verifies FHE key bytes against their advertised ids — is a **pre-existing SDK gap**, unrelated to this PR, and deserves its own issue.

**Finding 3 was mis-tiered** and belongs in 🟡 rather than 🔴. Steelmanned: envelope format (`Version: 1`), request commitment (`v3`), and wire attestation type (`v2`) version three genuinely different things, so three numbers is defensible design rather than a defect.

**Two findings get stronger.**

- *Finding 4*: all three de-pinned Dockerfiles use `RUN --mount=type=bind`, `type=cache,sharing=locked`, and `type=secret` — frontend-version-dependent features, so those builds now inherit whatever frontend the building daemon vendors instead of a pinned one. Compounding it: `build/relayer` and `build/kms-connector` were **skipped** on this PR, so the change to production image builds is unexercised by its own CI.
- *Finding 1*: the EVM path uses EIP-712 typed data, which wallets render field-by-field, so an EVM user does see the transport `publicKey` and contract scope. v3 therefore leaves the Solana path strictly *less* transparent than its EVM equivalent — not merely "no worse than before".

**One premise was inherited, not verified.** "Phantom refuses `signMessage` on arbitrary binary" is the PR's own stated rationale (`solana_extra_data.rs` module docs), repeated above as established fact. It is plausible and consistent with known wallet anti-blind-signing behaviour, but it was not independently confirmed — and the entire justification for a protocol-visible version bump rests on it. Worth confirming against Phantom's actual behaviour before the scheme change lands.

**Unchanged:** findings 2, 5, 6, 9, 10, 11 and the full 🟡 list stand as written. On 9, one honesty note — the trigger (`localStorage` throwing) is narrow, so the durable half of that finding is the inconsistency with `writeShieldJournal` and the undocumented SDK hook contract, not the likelihood of the failure.

*Also since posting: a third check is now red — `coprocessor-cargo-test/cargo-tests (bpr)`, failing in 5s, most likely the roll-up gate reporting the two lane failures rather than an independent break.*

PR #3355, opened 2026-07-28
feat(common): bash sha as preview env default

eudelins-zama:
### Summary

Removes the hardcoded version pins for fhevm's own artifacts in the PR-preview e2e environment, so it no longer drifts behind `main` and needs no manual pin bumps.

### Changes

- **Images resolve from the PR's base commit** instead of pinned tags: every push to `main`/`release/*` already tags each image with its short SHA, so `resolve-tags.cjs` walks back from the change-detection base commit to find published tags (failing loudly if GHCR has pruned them, with a warning for stacked PRs whose base is not on `main`). The run summary's Images table shows each tag and its provenance (`built`, `base-sha`, `dispatch-override`).
- **In-repo charts install straight from the branch checkout** rather than published OCI releases, we use local charts from the branch as-is without a publish/version bump. A `*_chart_version` dispatch input can still force a published release.
- **Listener build added** to the preview build matrix; its image now resolves from the same commit as the coprocessor consumer, so broker-payload compatibility holds by construction.
- **KMS pin bumped** to `d27c3b5` (external dep, so it stays a pin: it has no fhevm commit to derive from).
- **Follow-up fixes for recent `main` changes** surfaced by running the preview against the head charts/binaries: inject the kms-connector's `ethereumContractAddresses.protocolConfig` and move `kmsGeneration` to its new values key (and drop the dead `kmsVerifier` wiring), enable the sns-worker wallet now that the binary unconditionally requires an attestation signer, and pass `--seed-start-block=0` to the host-listener poller, which is now mandatory on first startup.
- Docs (`ci/preview-env/README.md`, `101-preview-env.md`) updated to match, including the new gotchas (plain `pr-preview-e2e` tests your charts but the base commit's code — use `-build` for code changes; stacked PRs resolve images from the merge-base with `main`).

### Tests

- Label based deployment with latest setup: https://github.com/zama-ai/fhevm/actions/runs/30360690688/job/90279565198?pr=3355
  - :white_check_mark: All good
- Manual workflow dispatch deployment with images/charts overriding (using what versions that were pinned before this PR): https://github.com/zama-ai/fhevm/actions/runs/30374063344/job/90325216374
  - :x: Fails because values.yaml files are not backward/forward compatible. In the future, we could simply starts the workflow from the tag ref we want, instead of the current branch with overriding of the component versions, this would use the correct versions and yaml files automatically.
  - So I would say this is the expected behavior
- Label based deployment with latest setup + e2e tests:https://github.com/zama-ai/fhevm/actions/runs/30925132167/job/92045522888
  - All tests passing, except the "random-subset" one. AFAIU, it is expected because we're using a test parameters during keygen, not real ones


piizama (dismissed):

piizama (approved):

isaacdecoded (approved):

eudelins-zama:
@mergifyio queue

PR #3362, opened 2026-07-29
chore(kms-connector): bound per-handle fan-out in kms-worker request checks

eudelins-zama:
### What

The `KmsWorker` fans out one attestation `HEAD` per Coprocessor bucket and one ACL call per host chain for every decryption request it processes. With up to `task_limit` (1000) requests in flight, nothing capped how many of those outbound requests could pile up, nor how long each could hang.

This PR puts a ceiling and a timeout on both paths:

- **Host chain RPC**: `connect_to_rpc_node_with_bounds` builds the provider with a `tower` `GlobalConcurrencyLimitLayer` and a reqwest client timeout. One ceiling per configured host chain.
- **S3**: a `BoundedClient` wrapper holding a semaphore per bucket for `HEAD`s and one global semaphore for ciphertext `GET`s.

### Config

| Field | Default |
| --- | --- |
| `host_rpc_max_concurrent_calls` | 128 |
| `host_rpc_call_timeout` | 10s |
| `s3_max_concurrent_heads_per_bucket` | 64 |
| `s3_max_concurrent_gets` | 16 |
| `s3_get_timeout` | 10s |
| `s3_head_timeout` | 5s |
| `copro_registry_refresh` | 60s |

All optional, all documented in `config/kms-worker.toml`.

The two concurrency ceilings sitting high (128 ACL calls, 64 `HEAD`s per bucket) bound cheap requests: an `eth_call` and a header-only `HEAD` cost the worker almost nothing, so the point is to stay within a provider's rate limits without throttling the fan-out. `s3_max_concurrent_gets` is much lower because it bounds memory rather than requests: each `GET` buffers a whole SNS ciphertext, so 16 at once is already a lot of the worker's heap.

The timeouts are all sized well above a healthy response, as they only exist to release a hung request — neither alloy nor reqwest sets one by default. `s3_head_timeout` is the tightest since a `HEAD` returns headers only and sits in the consensus critical path, while `s3_get_timeout` has to cover a full ciphertext download.

### Issue

Closes https://github.com/zama-ai/fhevm-internal/issues/1846

### DIFF
--- a/kms-connector/config/kms-worker.toml
+++ b/kms-connector/config/kms-worker.toml
@@ -56,7 +56,7 @@ database_url = "postgres://postgres:postgres@localhost/kms-connector"
 # ENV: KMS_CONNECTOR_GRPC_REQUEST_RETRIES
 # grpc_request_retries = 3
 
-# The maximum number of decryption attempts before a decryption request is permanently removed (optional, defaults to 20)
+# The maximum number of decryption attempts before a decryption request is considered failed (optional, defaults to 20)
 # ENV: KMS_CONNECTOR_MAX_DECRYPTION_ATTEMPTS
 # max_decryption_attempts = 20
 
@@ -71,6 +71,19 @@ database_url = "postgres://postgres:postgres@localhost/kms-connector"
 # Timeout to connect to a S3 bucket (optional, defaults to 3s)
 # ENV: KMS_CONNECTOR_S3_CONNECT_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
 # s3_connect_timeout = "3s"
+
+# Maximum number of calls in flight on a single host chain endpoint, across all requests (optional, defaults to 128)
+# ENV: KMS_CONNECTOR_HOST_RPC_MAX_CONCURRENT_CALLS
+# host_rpc_max_concurrent_calls = 128
+
+# Maximum number of attestation HEAD requests in flight on a single Coprocessor bucket (optional, defaults to 64)
+# ENV: KMS_CONNECTOR_S3_MAX_CONCURRENT_HEADS_PER_BUCKET
+# s3_max_concurrent_heads_per_bucket = 64
+
+# Maximum number of ciphertext GET requests in flight, across all requests and buckets (optional, defaults to 16)
+# Kept low: SNS ciphertexts are big, so too many buffered at once would OOM the worker.
+# ENV: KMS_CONNECTOR_S3_MAX_CONCURRENT_GETS
+# s3_max_concurrent_gets = 16
 
 # Gas cap for the host-chain `IERC1271.isValidSignature` static call (RFC-012).
 # Bounds resource use when verifying signatures from smart-account users (Safe, Argent,
--- a/kms-connector/config/kms-worker.toml
+++ b/kms-connector/config/kms-worker.toml
@@ -72,9 +72,13 @@ database_url = "postgres://postgres:postgres@localhost/kms-connector"
 # ENV: KMS_CONNECTOR_S3_CONNECT_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
 # s3_connect_timeout = "3s"
 
-# Maximum number of calls in flight on a single host chain endpoint, across all requests (optional, defaults to 128)
-# ENV: KMS_CONNECTOR_HOST_RPC_MAX_CONCURRENT_CALLS
-# host_rpc_max_concurrent_calls = 128
+# Timeout of a single attestation HEAD request on a Coprocessor bucket (optional, defaults to 5s)
+# ENV: KMS_CONNECTOR_S3_HEAD_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
+# s3_head_timeout = "5s"
+
+# Timeout of a single ciphertext GET request, covering the whole response body (optional, defaults to 10s)
+# ENV: KMS_CONNECTOR_S3_GET_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
+# s3_get_timeout = "10s"
 
 # Maximum number of attestation HEAD requests in flight on a single Coprocessor bucket (optional, defaults to 64)
 # ENV: KMS_CONNECTOR_S3_MAX_CONCURRENT_HEADS_PER_BUCKET
@@ -84,6 +88,18 @@ database_url = "postgres://postgres:postgres@localhost/kms-connector"
 # Kept low: SNS ciphertexts are big, so too many buffered at once would OOM the worker.
 # ENV: KMS_CONNECTOR_S3_MAX_CONCURRENT_GETS
 # s3_max_concurrent_gets = 16
+
+# Refresh interval of the Coprocessor registry read from the GatewayConfig contract (optional, defaults to 60s)
+# ENV: KMS_CONNECTOR_COPRO_REGISTRY_REFRESH (format: https://docs.rs/humantime/latest/humantime/)
+# copro_registry_refresh = "60s"
+
+# Maximum number of calls in flight on a host chain endpoint, across all requests (optional, defaults to 128)
+# ENV: KMS_CONNECTOR_HOST_RPC_MAX_CONCURRENT_CALLS
+# host_rpc_max_concurrent_calls = 128
+
+# Timeout of a single host chain call (optional, defaults to 10s)
+# ENV: KMS_CONNECTOR_HOST_RPC_CALL_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
+# host_rpc_call_timeout = "10s"
 
 # Gas cap for the host-chain `IERC1271.isValidSignature` static call (RFC-012).
 # Bounds resource use when verifying signatures from smart-account users (Safe, Argent,
@@ -147,15 +163,6 @@ address = "0x0000000000000000000000000000000000000000"
 # EIP-712 domain version for ProtocolConfig contract (optional, defaults to "1")
 # ENV: KMS_CONNECTOR_PROTOCOL_CONFIG_CONTRACT__DOMAIN_VERSION
 # domain_version = "1"
-
-# Off-chain ciphertext-attestation verifier configuration
-# [ct_attestation]
-# Per-bucket S3 attestation HEAD request timeout in seconds (optional, defaults to 5)
-# ENV: KMS_CONNECTOR_CT_ATTESTATION__HEAD_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
-# head_timeout = 5s
-# Coprocessor registry snapshot refresh interval in seconds (optional, defaults to 60)
-# ENV: KMS_CONNECTOR_CT_ATTESTATION__REGISTRY_REFRESH (format: https://docs.rs/humantime/latest/humantime/)
-# registry_refresh = 60s
 
 # List of supported host chains (required)
 # ENV: KMS_CONNECTOR_HOST_CHAINS
--- a/kms-connector/config/kms-worker.toml
+++ b/kms-connector/config/kms-worker.toml
@@ -76,9 +76,10 @@ database_url = "postgres://postgres:postgres@localhost/kms-connector"
 # ENV: KMS_CONNECTOR_S3_HEAD_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
 # s3_head_timeout = "5s"
 
-# Timeout of a single ciphertext GET request, covering the whole response body (optional, defaults to 10s)
+# Timeout of a single ciphertext GET request (optional, defaults to 20s)
+# A retrieval retries over every winning-group buckets, so a slow bucket is better abandoned for one of its replicas.
 # ENV: KMS_CONNECTOR_S3_GET_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
-# s3_get_timeout = "10s"
+# s3_get_timeout = "20s"
 
 # Maximum number of attestation HEAD requests in flight on a single Coprocessor bucket (optional, defaults to 64)
 # ENV: KMS_CONNECTOR_S3_MAX_CONCURRENT_HEADS_PER_BUCKET
--- a/kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
@@ -9,7 +9,6 @@ use crate::core::{
 use alloy::{
     primitives::{Address, B256, U256},
     providers::Provider,
-    transports::http::Client,
 };
 use anyhow::anyhow;
 use ciphertext_attestation::{
@@ -18,7 +17,6 @@ use ciphertext_attestation::{
 };
 use futures::future::try_join_all;
 use kms_grpc::kms::v1::TypedCiphertext;
-use std::time::Duration;
 use tokio::task::JoinSet;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, info, warn};
@@ -34,12 +32,6 @@ pub struct CiphertextManager<P: Provider> {
 
     /// HTTP client for the attestation HEAD fan-out and the ciphertext retrieval.
     s3_client: BoundedClient,
-
-    /// Timeout of a single attestation `HEAD` on a Coprocessor bucket.
-    head_timeout: Duration,
-
-    /// Number of attempts for S3 ciphertext retrieval.
-    s3_ciphertext_retrieval_attempts: u8,
 }
 
 impl<P> CiphertextManager<P>
@@ -48,7 +40,6 @@ where
 {
     pub async fn connect(
         provider: P,
-        client: Client,
         config: &Config,
         cancel_token: CancellationToken,
     ) -> anyhow::Result<Self> {
@@ -56,13 +47,7 @@ where
 
         Ok(Self {
             registry,
-            s3_client: BoundedClient::new(
-                client,
-                config.s3_max_concurrent_heads_per_bucket,
-                config.s3_max_concurrent_gets,
-            ),
-            head_timeout: config.s3_head_timeout,
-            s3_ciphertext_retrieval_attempts: config.s3_ciphertext_retrieval_attempts,
+            s3_client: BoundedClient::from_config(config)?,
         })
     }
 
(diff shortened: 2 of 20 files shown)


kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -27,9 +27,97 @@ use sha3::{
...
+        bucket: &str,
+        handle: B256,
+        head_timeout: Duration,
+    ) -> Result<CiphertextAttestation, FetchAttestationError> {
+        let bucket_semaphore = self.bucket_semaphore(bucket);
+        let _permit = bucket_semaphore
+            .acquire()
+            .await
+            .expect("S3 HEAD semaphore closed");
+        fetch_single_attestation(&self.client, bucket, handle, head_timeout).await
```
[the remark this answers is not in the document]
eudelins-zama: I think panicking is the right call here, as it would be totally unexpected, and handling the error would not provide any benefit.
It's possible only if the `Semaphore` gets closed, but we don't use such feature at all in the whole codebase, so this should be unreachable

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -339,4 +353,45 @@ mod tests {
         let result = aggregate_resolved_handles(vec![]);
         assert!(matches!(result, Err(ProcessingError::Recoverable(_))));
     }
+
+    /// No S3 request is issued without a permit: with the bucket's only one held, the resolution
```
jot2re: I am not sure I get this ` with the bucket's only one held,`.
Isn't the point of the test that the bucket is not reachable and hence a permit cannot be acquired?

jot2re (changes requested):
The overall changes I see looks good to me, but since I don't know the code base well, I asked Claude for some help checking the PR as well. It found quite a few things (which I _think_ are relevant):

1. 10s whole-body GET timeout vs 32 MiB ciphertexts. The coprocessor's own code says SNS ciphertexts reach 32 MiB (aws_upload.rs:43). 10s for the full body needs ≥3.3 MB/s sustained per stream, with 16 streams sharing the link. There was no total GET deadline before, so this converts "slow but successful" into hard failure — then s3_ciphertext_retrieval_attempts=3 × buckets, then max_decryption_attempts=20, then the request is dropped. Either raise the default (30–60s) or state the measured p99 body size/throughput it's based on.

2. Unbounded response body — the one resource this PR leaves unbounded. retrieve_ciphertext_via_http buffers response.bytes() with no content-length check and no cap, and the digest is verified after the full body lands. A winning-group bucket serving a huge body OOMs the worker before it can be rejected. Worse, body.to_vec() copies the Bytes, so peak is ~2× body: 16 × 32 MiB × 2 ≈ 1 GiB — the s3_max_concurrent_gets=16 rationale under-counts by 2×. Pre-existing, but squarely in this PR's scope (rule 14, ASVS resource limits). Cheap fixes: reject on oversized content-length, stream with a byte budget, drop the to_vec().

3. GET permit held across the entire retry loop. One of only 16 permits is held for attempts × winning_buckets sequential attempts — up to 3 × B × 10s while a stuck bucket is retried. Memory is only consumed during the body read, so acquiring per attempt keeps the OOM bound and cuts hold time by up to 3×B. The doc comment states the behaviour but not its cost.

4. Test gaps. connect_to_rpc_node_with_bounds — the actual pub entry point — has no test, only the private bounded_rpc_client. call_timeout is never exercised. test_load_from_env wasn't extended, so env deserialization of the new NonZeroUsize keys is unverified (it should work via try_parsing(true), and a 0 value is a good negative test). And test_load_valid_config_from_file passes trivially because every new key is commented out in kms-worker.toml — a typo'd key name there would not be caught.

## Nits and consistency
5. head_semaphores is never evicted — an entry lingers for every bucket URL ever seen, even after a coprocessor is deregistered. Governance-bounded, so low risk; deriving the per-bucket semaphores from the registry snapshot would also get the Mutex off the per-HEAD hot path.

6. CiphertextManager.head_timeout sits next to s3_ciphertext_retrieval_attempts — one drops the s3_ prefix, one keeps it (rule 8, trivial).

kms-connector/crates/kms-worker/src/core/kms_worker.rs
```diff
@@ -149,6 +155,7 @@ impl
         let kms_health_client = KmsHealthClient::connect(&config.kms_core_endpoints).await?;
         let s3_client = reqwest::Client::builder()
             .connect_timeout(config.s3_connect_timeout)
+            .timeout(config.s3_get_timeout)
```
jot2re: Claude: 
s3_get_timeout is enforced out-of-band and is absent in 2 of 3 construction sites. 

The GET deadline exists only because kms_worker.rs:158 calls .timeout() on the reqwest client. BoundedClient neither takes it nor asserts it, so CiphertextManager::for_test and the new s3.rs unit tests — both Client::new() — have no GET deadline. The new timeout is therefore untested, tests/s3.rs had to duplicate the builder, and any future caller silently gets an unbounded GET. Pass get_timeout into BoundedClient::new and apply it per-request the way head_timeout already is, or add one build_s3_client(&Config) helper.

kms-connector/crates/utils/src/conn.rs
```diff
@@ -89,9 +123,26 @@ pub async fn connect_to_rpc_node_with_wallet(
     Ok(NonceManagedProvider::new(provider, wallet.address()))
 }
 
+/// An `RpcClient` which never has more than `max_concurrent_calls` JSON-RPC calls in flight.
+fn bounded_rpc_client<T: IntoBoxTransport>(
```
jot2re, line 127: Claude: 

host_rpc_call_timeout does not bound queue wait, and nothing else does. 

The reqwest timeout is innermost, under the concurrency limiter (conn.rs:127-140), so a call waiting for one of 128 permits has no deadline. There is no request-level deadline anywhere in the worker either — grep tokio::time::timeout across kms_worker.rs, processor.rs, decryption.rs returns nothing. With task_limit=1000, events_batch_size=50, and O(handles × checks) ACL calls per request (try_join_all at decryption.rs:97 and :165 and :382), the queue can go very deep. Same story for the 16 GET permits. This is the gap between bounded resource use (achieved) and bounded latency (not). Put a TimeoutLayer outside the concurrency limit so the config key means what it says, and/or add LoadShed to fail fast — otherwise reword the config comments to "per HTTP attempt, excludes queueing".

kms-connector/config/kms-worker.toml
```diff
@@ -135,15 +164,6 @@ address = "0x0000000000000000000000000000000000000000"
 # ENV: KMS_CONNECTOR_PROTOCOL_CONFIG_CONTRACT__DOMAIN_VERSION
 # domain_version = "1"
 
-# Off-chain ciphertext-attestation verifier configuration
-# [ct_attestation]
-# Per-bucket S3 attestation HEAD request timeout in seconds (optional, defaults to 5)
-# ENV: KMS_CONNECTOR_CT_ATTESTATION__HEAD_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
```
jot2re, line 141: Claude: 

Env rename is breaking. 

KMS_CONNECTOR_CT_ATTESTATION__HEAD_TIMEOUT → KMS_CONNECTOR_S3_HEAD_TIMEOUT, ...__REGISTRY_REFRESH → KMS_CONNECTOR_COPRO_REGISTRY_REFRESH. Config has no deny_unknown_fields, so a deployment still setting the old names gets defaults with no warning. Nothing in charts/ or test-suite/ sets them today (checked), so no config file needs updating — but the PR needs an upgrade note.

kms-connector/crates/kms-worker/src/core/kms_worker.rs
jot2re, line 131: Claude:

Siblings not updated (rule 9). 

Only host-chain providers get bounds. The Gateway and Ethereum providers two lines above (kms_worker.rs:132-134), both gw-listener providers (gw_listener.rs:71-73), and connect_to_rpc_node_with_wallet in tx-sender all still pass None. Defensible if host chains are the intended scope — say so in the PR description, or make bounds the default with an opt-out. Related asymmetry: the S3 client sets connect_timeout and timeout; the RPC client sets only timeout (no host_rpc_connect_timeout mirroring s3_connect_timeout).

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -27,9 +27,97 @@ use sha3::{
...
+            let semaphore = Arc::new(Semaphore::new(self.max_concurrent_heads_per_bucket.get()));
+            semaphores.insert(bucket.to_string(), Arc::clone(&semaphore));
+            semaphore
+        }
+    }
+
+    /// See [`retrieve_verified_ciphertext`].
+    ///
+    /// One permit for the whole call: it retries the winning buckets sequentially.
+    pub(crate) async fn retrieve_verified_ciphertext(
```
jot2re: Claude: should be tighter when it comes to access. Only used within the same module.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -27,9 +27,97 @@ use sha3::{
...
+        Self {
+            client,
+            head_semaphores: Arc::new(Mutex::new(HashMap::new())),
+            max_concurrent_heads_per_bucket,
+            get_semaphore: Arc::new(Semaphore::new(max_concurrent_gets.get())),
+        }
+    }
+
+    /// See [`fetch_single_attestation`].
+    pub(crate) async fn fetch_single_attestation(
```
jot2re: Claude: Should be tighter access. Only used within the same module

kms-connector/crates/utils/src/conn.rs
```diff
@@ -61,12 +69,38 @@ pub type WalletProviderFillers = JoinFill<
     WalletFiller<EthereumWallet>,
 >;
 
+/// The ceilings of a RPC connection: how many calls it may have in flight, and how long any single
+/// one of them may take.
+#[derive(Clone, Copy, Debug)]
+struct RpcCallBounds {
+    pub max_concurrent_calls: NonZeroUsize,
```
jot2re: Nitpick Claude: Public access on a private struct. Should be changed to private access.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
```diff
@@ -27,9 +27,97 @@ use sha3::{
...
+        &self,
```
jot2re, line 83: Claude: this method along with `retrieve_verified_ciphertext` resolves associated items first, so these very likely link each method to itself. Use [fn@self::fetch_single_attestation].

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
jot2re, line 188: Nitpick: Claude: this method logs "Connected to RPC node successfully" without issuing a single request (CONNECTION_RETRY_NUMBER only guards the DB path). Pre-existing, but this function was touched.

kms-connector/crates/utils/src/conn.rs
eudelins-zama: Indeed the concurrency limit is the outer layer and the reqwest timeout the inner one, so time spent waiting for a permit (waiting in the "queue") is not counted.

That's intended, though. In this config a "call" is the HTTP request, and `host_rpc_max_concurrent_calls`, and I think it's the most intuitive behavior tbh.

Same for the S3 keys: `s3_get_timeout` bounds one GET while a retrieval makes several, and `s3_head_timeout` bounds one HEAD. So the wording is consistent with the rest of the file and I'd rather keep it.

kms-connector/config/kms-worker.toml
eudelins-zama: Should not be used in production anyway.
Most config stuff should rely on default values. Env is only there to override in case of emergency or whatever.

Closing this one.

kms-connector/crates/kms-worker/src/core/kms_worker.rs
eudelins-zama: We wants to bound RPC calls on host chain because we're doing a lot of them concurrently (ACL checks, context/epoch validation if not in cache...).

That is not the case for the Gateway, where we mainly do event listening / transaction sending.

Don't think we should act on this

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: Let's keep it out of this scope of this already not so small PR. Agree with the findings, but it really is very low impact...

kms-connector/crates/kms-worker/src/core/kms_worker.rs
eudelins-zama: Fixed c5c5f58a4bff14469e890f27148d33579cfd801e

kms-connector/crates/utils/src/conn.rs
eudelins-zama: Fixed 76d6ad246530df80d2b2a720f141b8c38c2f3f08

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: Fixed c5c5f58a4bff14469e890f27148d33579cfd801e

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: Fixed c5c5f58a4bff14469e890f27148d33579cfd801e

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: I updated the docs in c5c5f58a4bff14469e890f27148d33579cfd801e

Will probably merge the inner functions in a follow-up (for the bounds of the ciphertext download) anyway

eudelins-zama:
> 1. 10s whole-body GET timeout vs 32 MiB ciphertexts. [...] Either raise the default (30–60s) or state the measured p99 body size/throughput it's based on.

Fixed, and measured: the default is now 20s.

I checked where the 32MiB comes from, since it drives the number: a single comment in `sns-worker/src/aws_upload.rs:43`, explaining why the upload batch size is 10. It isn't a constant and isn't enforced anywhere, so I measured our mainnet bucket instead. In a 1000-object sample there are two size clusters: 1,919 B objects with no format metadata (the regular ciphertexts), and 98,844 B objects carrying `x-amz-meta-ct-format: compressed_on_cpu` — the SNS ones we GET. So the bodies we actually fetch today are ~96KiB, not 32MiB.

I still sized the timeout for the large case, because `CiphertextFormat` also defines the uncompressed variants, and an uncompressed ct128 is plausibly in the tens of MiB — most likely what that comment describes. 20s covers 32MiB at ~1.7MB/s per stream.

I didn't go to 30–60s though. The deadline gets multiplied by `s3_ciphertext_retrieval_attempts × winning buckets` on the failure path, so 60s means a ~15 minute worst case while holding one of the 16 GET permits. And since consensus already tells us several buckets vouch for the same digest, abandoning a slow bucket for a replica beats waiting it out.

> 2. Unbounded response body — the one resource this PR leaves unbounded.

Agreed, and the digest can't save us here since it's only checkable once the body has landed. The fix is to reject an oversized `content-length`, then stream with a byte budget so a missing or lying header can't get around it, which also removes the `to_vec()` copy. It needs a new config key and a decision on the ceiling value, so it's cleaner on its own. Will do in this [follow-up](https://github.com/zama-ai/fhevm-internal/issues/1856)

> 3. GET permit held across the entire retry loop.

Keeping as is. Real, but small here: the winning group is ~5 buckets, so a stuck one delays a few retrievals rather than a queue. The memory half of the argument also disappears once the body is capped, which is where a per-attempt permit would belong if we want one.

> 4. Test gaps. `connect_to_rpc_node_with_bounds` — the actual pub entry point — has no test [...] `call_timeout` is never exercised. `test_load_from_env` wasn't extended [...] `test_load_valid_config_from_file` passes trivially because every new key is commented out in `kms-worker.toml`.

Fixed, except the last one.

`connect_to_rpc_node_with_bounds` now has a test, which also exercises `call_timeout`: a socket that accepts connections and never answers, then asserting the call ends in an error *and* that at least the timeout elapsed. Without that second assert the test would pass just as happily on an instant connection refusal.

`test_load_from_env` now sets and checks all seven new keys, plus a negative test asserting a `0` is refused for each `NonZeroUsize` bound. That also confirms `try_parsing(true)` really does route env strings into `NonZeroUsize`.

The commented-out keys in `kms-worker.toml` I'm leaving as is. Every documented default in that file is commented out, so catching a typo'd name needs a separate fixture rather than the example config.

> 5. `head_semaphores` is never evicted — an entry lingers for every bucket URL ever seen, even after a coprocessor is deregistered.

Keeping as is. Bounded by governance, and an entry is a URL string plus a semaphore. Deriving them from the registry snapshot is a redesign for a bound we already have.

> 6. `CiphertextManager.head_timeout` sits next to `s3_ciphertext_retrieval_attempts` — one drops the `s3_` prefix, one keeps it.

Fixed. The keys are all `s3_*` / `host_rpc_*` now, and `head_timeout` moved into `BoundedClient`, which is S3-only, so the prefix would be redundant there.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
jot2re: I think you did the opposite, right? Now the functions are fully public. 
I think Claude's comment were more like changing to `pub(in crate::core::event_processor::ciphertext)`. But maybe that is your intend since you made the other version of this function private.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
jot2re: idem

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: It's now used in integration test, (we were using the nested function before), so yes it must be pub

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs
eudelins-zama: It's now used in integration test, (we were using the nested function before), so yes it must be pub

jot2re (approved):
LGTM! Just one tiny nitpicky comment still open

eudelins-zama:
@mergifyio queue

eudelins-zama:
@mergifyio refresh

PR #3364, opened 2026-07-29
feat(solana): add the canonical user-decrypt permit and its revocation

KitHat:
First PR of the fhevm-internal#1689 series: one implementation of the permit canon, the
normative vectors, and the on-chain revocation lever. Everything else in the series
consumes this.

## What's here

- `crates/zama-solana-permit` — typed fields, strict decoding, the canonical text, the
  sRFC 38 envelope, Ed25519 verification, the transport-key fingerprint. Reads no chain
  state and no clock: the validity window, the revocation watermark, deployment identity
  and scope are the authorization layer and live in the consumer that owns live state.
- `test-fixtures/permit/permit_v1.json` — 29 normative records plus the schema the other
  implementations include.
- `zama-host: revoke_permits` — writes `max(stored, Clock.unix_timestamp)`. One
  transaction kills every outstanding permit of the signer, in constant work.

## Worth a reviewer's attention

- **The API cannot express an unverifiable path.** `PermitFields` has no public
  constructor, the renderer returns `String` without a `Result`, and verification takes
  typed fields plus a signature — never a text or an envelope. Three compile-fail doctests
  hold the same shapes from the other side.
- **Strictness is pinned, not inherited.** `verify_strict` semantics: the same three checks
  Solana applies to transaction signatures. Canonical public-key encoding is required on
  top of that, deliberately stricter than Solana — no wallet produces a non-canonical one,
  and libraries disagree about accepting them, so the disagreement is removed here rather
  than left for five implementations to find.
- **Goldens were computed outside the crate** (independent base58, SHAKE-256, Ed25519 and
  calendar), so a bug shared between the renderer and a helper it uses cannot make them
  pass.
- **Tests were written before the code.** Two of their own claims turned out to be wrong
  and are fixed here: the per-domain size step at the digit boundary of the domain count,
  and a record marked `valid` whose user pubkey was not a usable key.

## Verify

cargo test -p zama-solana-permit
cargo build-sbf --manifest-path programs/zama-host/Cargo.toml
cargo test -p zama-solana-runtime-tests --test permit_invalidation_mollusk
python3 scripts/check_solana_abi.py --root .

## Not here

No revocation event: there is no consumer yet — the Connector reads the account from an
atomic snapshot, not from logs — and adding one now would freeze its shape in the vendored
IDL before the consuming PRs specify it. The vectors are not byte-frozen either, because
the chain-id derivation is still an open protocol question; the file says so. The
authorization rules, the TS renderer, the linker and relayer validation are later PRs.


Eikix:
**Here's an automated review.
Overall looks very solid**


<html><head></head><body><p>I re-reviewed PR #3364 from scratch using this rule:</p><blockquote><p>Match the EVM outcome. Use Solana-native encoding and state. Follow the current KitHat gist unless there is a concrete reason not to.</p></blockquote><p>The earlier <code>extra_data</code> and revocation P1 findings are withdrawn.</p><h2>Findings</h2><h3>[P2] Pin the revocation account’s serialized layout</h3><p>The future Connector will read <code>PermitInvalidation</code> directly from a Solana account snapshot, equivalent to the EVM Connector calling:</p><pre><code>decryptionSignatureInvalidatedBefore(user)</code></pre><p>But <code>PermitInvalidation</code> is supplied as an <code>UncheckedAccount</code>, so its type is absent from the generated IDL. Current tests pin:</p><ul><li>discriminator;</li><li>total size;</li><li>PDA derivation.</li></ul><p>They do not pin the exact bytes and field order:</p><pre><code>discriminator | user | watermark | bump</code></pre><p>A same-size field change could therefore pass the current ABI checks while causing the Connector to misread the watermark.</p><p><a href="https://github.com/zama-ai/fhevm/blob/bdde63af16b8c63f5d365920e72b69216200b351/solana/programs/zama-host/src/state/permit_invalidation.rs#L13-L27"><span><span><code>PermitInvalidation</code></span></span></a> · <a href="https://github.com/zama-ai/fhevm/blob/bdde63af16b8c63f5d365920e72b69216200b351/solana/programs/zama-host/src/instructions/revoke_permits.rs#L28-L40"><span><span><code>RevokePermits</code></span></span></a></p><p>Before merging the canonical state definition, I would add either:</p><ul><li>a small shared codec used by the host and Connector; or</li><li>a byte-exact serialized account fixture consumed by both.</li></ul><p>This is the only material implementation gap I found.</p><h3>[P2] State the EVM revocation limitation next to the implementation</h3><p>The mechanism is correct for EVM parity. The documentation is too absolute when it says one transaction kills “every outstanding permit.”</p><p>A permit signed in advance with a future <code>start_timestamp</code> can survive an earlier revocation and become usable when its window opens. This is an accepted cross-ecosystem weakness in the current gist, not something Solana should unilaterally redesign. <a href="https://gist.github.com/KitHat/296fdc99aaf2a49d6703401c61f4731b"><span><span>The gist explicitly records that decision and keeps the timestamp watermark.</span></span></a></p><p>The implementation should therefore say:</p><ul><li>usable permits starting before the watermark are invalid;</li><li>future starts are rejected at request time;</li><li>pre-signed future permits remain the accepted EVM-parity limitation.</li></ul><p>That prevents later consumers from assuming the PDA provides a stronger guarantee than EVM.</p><h2>Feature-by-feature result</h2><div><div>
Feature | EVM outcome | PR #3364 | Review
-- | -- | -- | --
Reusable permit | Handles excluded | Handles excluded | Correct
User identity | EVM address | Native 32-byte Solana pubkey | Correct adaptation
Transport key | Full key cryptographically committed | Full key committed through SHAKE-256 fingerprint | Equivalent outcome
Application scope | Empty = permissive; max 10 | Empty = permissive; max 10 | Correct
Validity | Start, duration, max 365 days | Same | Correct
Deployment binding | Chain ID + verifying contract | Chain ID + verifying program | Correct adaptation
KMS routing | extraData 0x02 context + epoch | Same bytes, rendered readably | Accepted and coherent
Wallet signing | EIP-712 | Human-readable sRFC38 v1 | Correct adaptation
Revocation | Per-user timestamp threshold | Per-user PDA timestamp threshold | Correct
Smart accounts | ERC-1271 | Later on-chain delegation | Correct Solana substitute
Signature verification | ECDSA/ERC-1271 | Strict Ed25519 | Correct
Handles and linker | Per-request, outside permit | Deferred | Correct sequencing

</div></div><p>The sRFC38 envelope is exact: preamble, version 1, one signer, signer pubkey, then UTF-8 content. It matches the <a href="https://docs.anza.xyz/proposals/off-chain-message-signing"><span><span>Anza off-chain signing specification</span></span></a>.</p><h2>Required behavior in later PRs</h2><p>These are deliberately outside #3364, but they are required for EVM outcome parity:</p><ul><li>SDK start:</li></ul><pre><code>max(floor(now / 60) × 60, current watermark)</code></pre><ul><li>Gateway/relayer/Connector reject future starts.</li><li>Expiry should match current EVM code: reject when <code>now &gt; start + duration</code>. The gist’s <code>now ≥ end</code> wording differs by one second and should not be copied.</li><li>Connector rejects when:</li></ul><pre><code>start_timestamp &lt; invalidation_watermark</code></pre><p>Equality remains valid, matching EVM.</p><ul><li>Live deployment, KMS context/epoch, scope and delegation checks must use one consistent Solana account snapshot.</li><li>The TS renderer must reproduce the Rust vectors byte-for-byte.</li><li>Phantom and Ledger must be exercised through <code>solana:signOffchainMessage</code>; no raw <code>signMessage</code> fallback.</li><li>Authorization and linker vectors still need to be added when those layers land.</li></ul><h2>Open decisions, not defects</h2><ul><li>The final genesis-hash-to-<code>chain_id</code> derivation remains unsettled. The PR correctly says its current vectors are not yet byte-frozen.</li><li>The reference 800-byte transport key is not a canonical ML-KEM key. That matches EVM’s permit layer, which treats the key as signed bytes, but using a genuinely generated ML-KEM-512 public key would make the fixtures easier for KMS implementations to consume.</li><li>Solana revocation is idempotent while EVM rejects a non-increasing threshold. The resulting authorization state is identical; the Solana behavior is preferable.</li></ul><h2>Verdict</h2><p>The permit design itself is aligned with EVM outcomes and the current gist. I would request one implementation change before merge: pin the <code>PermitInvalidation</code> account layout for the Connector. I would also correct the unconditional revocation claim.</p><p>The PR currently conflicts with <code>feature/solana</code>, so it still needs a rebase and regenerated ABI/cost snapshots. Existing applicable CI is green at <code>bdde63a</code>, but it must run again after that rebase.</p><p>No GitHub review or comment was posted.</p></body></html>

Eikix:
Additional nits:

- Trim comments when they repeat an adjacent test name or implementation.
- Use a real generated ML-KEM-512 public key for the positive fixture.

Eikix (approved):

PR #3372, opened 2026-07-30
chore(common): prefixed s3 path for ct

eudelins-zama:
### Problem

`sns-worker` uploads the ct64 and ct128 ciphertexts under the same S3 key (`{handle}/{context_id}`), relying on the bucket to differentiate them.
In production the chart wires `--bucket-name-ct64` and `--bucket-name-ct128` to the same bucket, so the two concurrent uploads race for one object and the last writer wins.
`kms-worker`s then fail retrieval with a digest mismatch.

###  Fix

Prefix the S3 keys by artifact type so a single bucket is safe by construction:

- ct128: `ct128/{handle}/{context_id}`
- ct64: `ct64/{handle}/{context_id}`

### Changes:

- coprocessor (sns-worker): prefixed keys for uploads; S3 migration and dry-run target the prefixed layout (migration has not been run in any environment yet, so no legacy un-prefixed data exists).
- kms-connector (kms-worker): RFC-023 URL builder fetches `{bucket_url}/ct128/{handle}/{context_id}`. The digest-URL fallback is untouched, so buckets written by v0.13 coprocessors remain readable.
- test-suite: ct64/ct128 now share a single copro bucket, mirroring the production wiring. The e2e suite would now reproduce the collision on an unfixed sns-worker.

### Follow-ups

RFC 023 needs a follow-up amendment for the new layout; same for main (re-application, since the digest fallback/compat copy were removed there)

### DIFF
--- a/coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
+++ b/coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
@@ -567,8 +567,14 @@ async fn upload_ct(
     Ok(result)
 }
 
-pub(crate) fn s3_ciphertext_key(handle: &[u8], context_id: U256) -> String {
-    hex::encode(handle) + "/" + &context_id.to_string()
+/// S3 key of the ct128 (SNS) ciphertext object.
+pub(crate) fn s3_ct128_key(handle: &[u8], context_id: U256) -> String {
+    format!("ct128/{}/{}", hex::encode(handle), context_id)
+}
+
+/// S3 key of the compressed ct64 ciphertext object.
+pub(crate) fn s3_ct64_key(handle: &[u8], context_id: U256) -> String {
+    format!("ct64/{}/{}", hex::encode(handle), context_id)
 }
 
 fn b256_from_bytes(field: &str, bytes: &[u8]) -> anyhow::Result<B256> {
@@ -725,7 +731,7 @@ async fn upload_ciphertexts(
     };
 
     if *ct128_digest != NO_SNS_CIPHERTEXT_DIGEST.to_vec() {
-        let key = s3_ciphertext_key(&task.handle, context_id);
+        let key = s3_ct128_key(&task.handle, context_id);
         let digest_key = hex::encode(ct128_digest);
 
         if preserve_legacy_s3_format {
@@ -826,7 +832,7 @@ async fn upload_ciphertexts(
     {
         let ct64_compressed = task.ct64_compressed.as_ref();
         let ct64_digest = &upload_material.ct64_digest;
-        let key = s3_ciphertext_key(&task.handle, context_id);
+        let key = s3_ct64_key(&task.handle, context_id);
         let ct64_checksum_sha256 = conf
             .verify_sha256_checksum
             .then(|| match ct64_compressed.is_empty() {
@@ -1848,7 +1854,7 @@ mod tests {
         assert_eq!(row.ciphertext128.as_deref(), Some(ct128_digest.as_slice()));
         assert_eq!(row.s3_format_version, Some(S3_FORMAT_VERSION_LEGACY));
 
-        let ct64_key = s3_ciphertext_key(&handle, COPROCESSOR_CONTEXT_ID_1);
+        let ct64_key = s3_ct64_key(&handle, COPROCESSOR_CONTEXT_ID_1);
         client
             .head_object()
             .bucket(bucket_ct64)
@@ -1863,10 +1869,13 @@ mod tests {
     fn s3_key_v1() {
         let handle = B256::ZERO;
         let coprocessor_context_id = U256::ZERO;
-        let s3_key = s3_ciphertext_key(handle.as_ref(), coprocessor_context_id);
         assert_eq!(
-            "0000000000000000000000000000000000000000000000000000000000000000/0",
-            &s3_key
+            "ct128/0000000000000000000000000000000000000000000000000000000000000000/0",
+            &s3_ct128_key(handle.as_ref(), coprocessor_context_id)
+        );
+        assert_eq!(
+            "ct64/0000000000000000000000000000000000000000000000000000000000000000/0",
+            &s3_ct64_key(handle.as_ref(), coprocessor_context_id)
         );
     }
 }
--- a/coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
+++ b/coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
@@ -13,8 +13,8 @@ use aws_sdk_s3::Client;
 use base64::Engine;
 use bytesize::ByteSize;
 use ciphertext_attestation::{
-    CiphertextAttestation, CiphertextAttestationPayload, CiphertextFormat, Version,
-    S3_METADATA_ATTESTATION_KEY,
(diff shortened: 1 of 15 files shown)


coprocessor/fhevm-engine/sns-worker/src/tests/mod.rs
```diff
@@ -940,13 +940,21 @@ async fn assert_ciphertext128(
         assert_ciphertext_uploaded(
             test_env,
             &test_env.conf.s3.bucket_ct128,
+            &crate::aws_upload::s3_ct128_key(handle, crate::aws_upload::COPROCESSOR_CONTEXT_ID_1),
             handle,
             Some(ct.len() as i64),
             Some((&expected_ct_format, expected_attestation_format)),
         )
         .await?;
```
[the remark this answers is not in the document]
eudelins-zama: Hallucination, see CI successful compilation.
Test code anyway

coprocessor/fhevm-engine/sns-worker/src/tests/s3_migration.rs
```diff
@@ -419,7 +421,7 @@ async fn test_before_and_quit_records_corrupted_ct64_object_error() {
     assert_object_missing(
         &env.s3_client,
         &env.conf.s3.bucket_ct64,
-        &crate::aws_upload::s3_ciphertext_key(&handle, crate::aws_upload::COPROCESSOR_CONTEXT_ID_1),
+        &crate::aws_upload::s3_ct64_key(&handle, crate::aws_upload::COPROCESSOR_CONTEXT_ID_1),
     )
```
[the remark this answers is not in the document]
eudelins-zama: Hallucination, see CI successful compilation.
Test code anyway

dd23 (approved):
LGTM

coprocessor/fhevm-engine/sns-worker/src/tests/mod.rs
```diff
@@ -940,13 +940,21 @@ async fn assert_ciphertext128(
         assert_ciphertext_uploaded(
             test_env,
             &test_env.conf.s3.bucket_ct128,
```
obatirou, line 962: `&test_env.conf.s3.bucket_ct128` is used here and  `&test_env.conf.s3.bucket_ct64` below.
As we are now assuming there is only one bucket (even if we keep the option), we should ensure the test config reflect it

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -567,8 +567,14 @@ async fn upload_ct(
...
-pub(crate) fn s3_ciphertext_key(handle: &[u8], context_id: U256) -> String {
-    hex::encode(handle) + "/" + &context_id.to_string()
+/// S3 key of the ct128 (SNS) ciphertext object.
+pub(crate) fn s3_ct128_key(handle: &[u8], context_id: U256) -> String {
+    format!("ct128/{}/{}", hex::encode(handle), context_id)
+}
+
+/// S3 key of the compressed ct64 ciphertext object.
+pub(crate) fn s3_ct64_key(handle: &[u8], context_id: U256) -> String {
+    format!("ct64/{}/{}", hex::encode(handle), context_id)
```
obatirou: nit: we should document the invariant that is assumed.
ct64 and ct128 may share the same bucket and prefixes are what keep them distinct

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
```diff
@@ -567,8 +567,14 @@ async fn upload_ct(
     Ok(result)
 }
 
-pub(crate) fn s3_ciphertext_key(handle: &[u8], context_id: U256) -> String {
-    hex::encode(handle) + "/" + &context_id.to_string()
+/// S3 key of the ct128 (SNS) ciphertext object.
+pub(crate) fn s3_ct128_key(handle: &[u8], context_id: U256) -> String {
+    format!("ct128/{}/{}", hex::encode(handle), context_id)
```
obatirou: should `S3_CT128_KEY_PREFIX` / `S3_CT64_KEY_PREFIX` be added to `ciphertext-attestation` as KMS connector and coproc depends on it and avoid hardcoding the prefix?

coprocessor/fhevm-engine/sns-worker/src/tests/mod.rs
obatirou: We might to need update s3_utils: assert_ciphertext_s3_object_count use s3_utils which rely on bucket name to count the different cts uploaded

eudelins-zama: I would prefer to do this properly in `main` but keep it like this in `v0.14` to do the least amount of change possible.
Thoughts?

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
eudelins-zama: In its current form, we the setup allow the use of same bucket OR two different buckets, so there's no invariant assumed here AFAICT

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
eudelins-zama: Indeed, will do

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
eudelins-zama: Actually, I think it's better this way in the sense that it offers more flexibility (two or one buckets, both are fine in current setup, so better?)

Then no need to handle the above comment in `main` (and/or in this PR)?

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
eudelins-zama: Fixed 1dddae45d5aad1498d2682592e5e6181fd598217

coprocessor/fhevm-engine/sns-worker/src/tests/mod.rs
eudelins-zama: See [here](https://github.com/zama-ai/fhevm/pull/3372#discussion_r3682503735), I think I prefer current setup in the end

coprocessor/fhevm-engine/sns-worker/src/aws_upload.rs
obatirou: Ok I agree

coprocessor/fhevm-engine/sns-worker/src/tests/mod.rs
obatirou: The current fix is done to ensure things can work under a unique bucket.
Not modifying unit test or not adding news one go against this no?
The bug went though because unit tests assumed 2 buckets and we missed to cover a case that was not scheduled. There is a tradeoff to be done.

eudelins-zama: Fixed 38189254ce47734f1785db9df03f3dcf4fc13887

obatirou (approved):

PR #3377, opened 2026-07-30
refactor(solana): fhe_eval cleanup — interned wire format, measured op cap (fhevm-internal#1853)

Eikix:
Implements zama-ai/fhevm-internal#1853 (fhe_eval ground-up cleanup, W1–W12) as one folded PR, commits ordered for review per the issue's landing plan. On merge: closes fhevm-internal#1722, fhevm-internal#1723, fhevm-internal#1725 (absorbed/superseded per the issue's disposition table).

## Commits (review order = landing order)

| Commit | Waves | What |
|---|---|---|
| `51878a9c7` | W1+W2 | Delete the dead fhe_eval derivation surface |
| `d618b0d6d` | W5 | `EvalAccountTable` — one owner for every remaining-accounts invariant |
| `560df19f5` | W6 | Collapse admission + execution into one walk |
| `e65abd57e` | W3+W4 | Keccak content-addressed handles, durable-anchor rand seeds, `context_id` deleted |
| `ba5f71bd2` | W7+W10 | Interned constant-pool wire format; zama-fhe split into 10 modules |
| `342488f13` | W8+W9+W12 | `MAX_FHE_EVAL_OPS` = 32 from measured budgets; HCU inverse conformance; sysvar tracking note |

## Headline results

- **Wire format (W7):** `FheEvalArgs` carries an interned 32-byte constant pool; steps reference handles/scalars/ACL keys/labels/subjects by `u8` index. Max-op frame instruction data **889 → 306 bytes at the old cap (−66%)**, three-op token frame 343 → 189 B; CU flat. `AclSubjectEntry` deleted; frames self-describe `account_count` (DD-033), stamped at CPI-invoke time because deny-record witnesses are per-transaction.
- **Cap (W8):** derived from measurement only. Marginal chained step ≈ 9 B / ≈3,700 CU. At 32 ops the max frame measures **450 B / 150,354 CU** — the full signed transaction fits the 1,232-byte packet with ≥150 B envelope headroom (asserted by a new test) and needs no compute-budget instruction. 48 ops would break both budgets.
- **Derivation ABI (W3/W4):** deterministic handles are content-addressed (no `context_id`, no op-index salt — EVM `FHEVMExecutor` parity); rand seeds anchor to the frame's durable writes (`(account key, previous_handle)` pairs in wire order), making them compulsorily fresh without a caller-supplied counter.
- **Co-change lanes:** host-listener reconstruction, solana-proof-service decode, vendored IDL/ABI goldens, cost snapshots (regenerated via the CI artifact flow), tfhe-worker `solana_poc`. All suites green: 104 host unit / 32 SDK / 621 runtime / proof-service / 139 listener; clippy + fmt clean.

## Problems and boundaries surfaced by this work

1. **Hard 32KB heap boundary on all-durable frames.** The W8 heap re-check found that the heap-heaviest legal frame shape — every step a born-public durable create — fits **20 creates** on the SBF bump heap; the 21st OOMs and reverts cleanly (nothing commits). This is a *hard* limit today: Anchor's default allocator serves a fixed 32KB region, so a compute-budget heap-frame request does not help. Pinned by `mollusk_fhe_eval_born_public_heap_boundary`. Realistic frames (many transient steps, few durable outputs) are unaffected — the 32-op max frame with one durable output passes with wide margin. Raising the boundary requires a custom allocator, deliberately out of scope here.
2. **`e65abd57e`'s commit message overstates verification.** It claims the listener/proof-service lanes were verified in that commit; they were not (the listener would not have compiled against it). Both lanes were repaired and genuinely verified in `ba5f71bd2`. Kept as-is to preserve the review-ordered history; the misstatement is corrected here.
3. **The sysvar read stays manual, but is now safe code (W12, resolved in `6daf43e8d`).** `previous_bank_hash` hand-reads `SlotHashes` via the `sol_get_sysvar` syscall because `PodSlotHashes::fetch()` is broken upstream: it allocates an align-1 `Vec<u8>` and then fails its own 8-byte alignment check whenever the SBF bump allocator's cursor is odd — verified byte-identical at solana-sysvar 3.1.1, 4.2.0, and master, and reproduced under Mollusk (real SBF; it passes only when the buffer lands 8-aligned by allocation-history luck). Both former `unsafe` blocks are deleted: the syscall copies raw bytes with no alignment requirement and entries parse via `from_le_bytes`, so plain byte buffers suffice. The manual windowed read (16 entries, 648 bytes) also stays deliberately — `fetch()` would allocate the full 20,488-byte sysvar, 2/3 of the 32KB heap, per call. No upstream anza-xyz/solana-sdk issue exists as of 2026-07; an issue draft is ready to file.
4. **W7's round-trip gate is composite, not one test.** The listener imports zama-host's own `computed_*` functions and `FheEvalArgs` types (byte-identical by construction); listener fixtures assert pooled-frame reconstruction including rand anchors; Mollusk asserts MMR reconstruction via the shared `zama-solana-acl` crate; and the `solana_poc` e2e (SDK frame → LiteSVM → listener reconstruct → decrypt) covers the full loop but is `#[ignore]`d — it needs a live Postgres + TFHE stack and runs in the e2e lane, not in this PR's CI. Pulling host-listener into the runtime-tests workspace for a single always-on test would drag sqlx/alloy/tokio into the Solana build.
5. **HCU numbers remain placeholders.** W9 adds inverse conformance (every combination validation admits has a cost row, so a validated step can never die mid-frame with `HcuUnknownCost`), but calibration is deferred and limits ship disabled. The metering pass runs before the walk's type validation, so it prices combinations validation later rejects — fail-closed, by design.
6. **Deferred on purpose** (per the issue's non-goals): `CurrentOf` operand mode (fhevm-internal#1854), registered plans (#1724, breaks DD-033), `previous_subjects` compression (gated on #1844/#1654), v0+ALT packing (#1726 — account keys only, orthogonal to instruction-data bytes).

## Notes for reviewers

- `previous_handle`/`previous_subjects` stay verbatim in instruction data (not interned): they are what lets indexers reconstruct MMR leaves from instruction data alone, and the rand anchor hashes them in wire order.
- Cost snapshots were refreshed against the CI-canonical artifacts (`scripts/check-zama-host-idl.sh` build flags); token/batcher CU drift of ±25 CU comes from the rebuilt program, not from logic changes.
- The stale DD-024 citation for the event-batch/step-cap debate (flagged in #1725) now points at DD-038.


Eikix:
## Automated deep review (Claude, dispatched by @Eikix)

An in-depth AI review pass ran against the full diff. Posting the results here for the record; treat it as a pre-review, not a substitute for human review.

### Checked and confirmed clean

- **Pool index bounds-checking** on all four consumers of the interned wire format — the on-chain program, the `zama-fhe` SDK, the proof-service decoder, and the host-listener reconstruction. Out-of-bounds indexes reject the frame in preflight, before any state is touched.
- **Duplicate-account rule and one-durable-write-per-account rule** hold across the walk.
- **Deny-subject records are located by derived address**, never by a caller-supplied index — a caller cannot point the check at the wrong record.
- **Rand-step freshness**: a rand step must bind a durable output, so seed reuse across frames is impossible.
- **In-frame supersession exemption** behaves as documented.
- **`make_public` MMR leaf ordering** is identical across all three paths that compute it (program, listener, proof service).
- **Listener round-trip**: the listener re-derives output handles from raw transaction bytes; both sides of that equivalence were rewritten in this PR and the fixture + e2e lanes confirm byte-identical results.

### Open findings (deliberately not fixed in this PR)

1. **Deny-list create-path gap** — inherited from the base branch, not introduced here. Per-subject deny checks run only on rotation-added subjects; initial subjects on the create path and `allow_subjects` additions are not per-subject checked, so a denied subject can enter via create and persist. Disclosed in the rotation-check doc comment. Decision pending: fix here, file an issue, or leave documented.
2. **Error-code reuse** — a few diagnostic codes are reused for the wrong defect (e.g. `FheEvalPoolIndexOutOfBounds` also fires for "pool entry never referenced"). Cosmetic; fixing requires new error variants and an IDL regen.

### Findings already addressed during the review

- PR description refreshed to match the final state of the diff.
- Three stale/incorrect code comments fixed.

### Notes for human reviewers

- The cost-snapshot JSON churn at the tip of the branch is CI-toolchain drift (platform-tools 2.1.0 on CI vs 2.1.21 locally, +6/+12/+24 CU offsets), converged on CI's canonical numbers — not a correctness signal.
- Listener reconstruction fixtures compile only with `--features solana-grpc,solana-reconstruct`; a plain `cargo test` silently skips them.
- The highest-value place to look for bugs is the handle-derivation equivalence between `fhe_eval` and the listener — both sides changed.

PR #3394, opened 2026-07-30
feat(relayer): add ciphertext readiness check based S3 storage URLs

manoranjith:
Implements the Relayer side of RFC-023 Part 2 — [zama-ai/fhevm-internal#1622][issue].

**Breaking for deployments.** `gateway.contracts.gateway_config_address` and both `gateway.readiness_checker.gw_ciphertext_check` timings are mandatory with no defaults. Relayer values live in `zama-zws/gitops`, outside this repo, and must gain all three before rollout or the relayer will not start.

One commit is unrelated to readiness: the local dev Postgres moves to 15433, which collided with the e2e stack's `relayer-db`.

## Deviations from the issue

- There is no `fhevm_ct_commits` crate. Consensus evaluation already lives in `shared/ciphertext-attestation`; this adds a sibling, `shared/ciphertext-attestation-client`, holding only the parts that touch the network and the chain — the S3 `HEAD` fetch and the `GatewayConfig` registry mirror. It sits under `shared/` so kms-worker could adopt it if that team finds it useful; step 5 is not done here.
- The registry is cached, not read per request: signers, buckets and threshold come from `GatewayConfig` at startup, refreshed on a timer. The per-handle attestation check itself stays stateless as specified. Since no per-request contract call remains that can fail, the three `test_readiness_contract_error_returns_failure_v2` tests became `test_readiness_no_consensus_returns_failure_v2`.
- Two failure modes, not one user-facing error. Attestations not yet published retry, since Coprocessors publish asynchronously; Coprocessors that answer and disagree are terminal under a new `no_attestation_consensus` label, HTTP 500, because waiting cannot resolve it. `openapi.yml` is regenerated.
- The S3 mock is wiremock, not MinIO: the client does a bare unauthenticated `reqwest.head()`, and MinIO cannot express the timeout or ordered-response cases.

## Test plan

- [ ] CI: 207 unit, 279 integration, 26 support-crate tests — all green locally.
- [ ] CI: `make clippy`, `cargo fmt --check`, `make openapi-check`.
- [ ] e2e (needs the `e2e` label), which cannot confirm the gate fails closed — no e2e spec asserts a readiness rejection.

## Follow-ups

- Before this leaves draft: fail-closed unit tests in `shared/ciphertext-attestation-client` — under-threshold, disagreement, unregistered signer, `HEAD` timeout, wrong handle, plus a positive control. Nothing currently proves the gate rejects.
- `relayer/tests/recovery_test.rs` is red and was red before this branch: the `/v2/keyurl` poller has gated startup fatally since `1796dd217`, that file has never mocked `getActiveKeyId`, and no CI job runs it. Not fixed here.

[issue]: https://github.com/zama-ai/fhevm-internal/issues/1622


shared/ciphertext-attestation-client/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+authors = ["Zama"]
+name = "ciphertext-attestation-client"
```
eudelins-zama, line 3: Why a whole new crate and not simply a `client` module within the `ciphertext-attestation` crate ? (potentially gated by a `client` feature)

manoranjith: Yeah, would to fine organise as a module within the original crate as well.

manoranjith: Just put it in a dedicated place to start with.

eudelins-zama: This is slightly harder to review this way IMO (new Cargo.toml/lock files, edit of the workflows etc.), but that's a detail tbh

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,546 @@
...
+pub struct ValidAttestation {
+    material: ConsensusMaterial,
+    signer: Address,
+}
+
+impl ValidAttestation {
+    /// Gate 1: signature recovers to the embedded signer ([`CiphertextAttestation::verify`]).
+    /// Gate 2: that embedded signer equals the signer this bucket is registered to on-chain —
+    /// without this gate a bucket could serve another bucket's (validly signed) attestation and
+    /// have it counted as its own.
```
eudelins-zama, line 27:
```suggestion
    // Gate 1: signature recovers to the embedded signer ([`CiphertextAttestation::verify`]).
    // Gate 2: that embedded signer equals the signer this bucket is registered to on-chain —
    // without this gate a bucket could serve another bucket's (validly signed) attestation and
    // have it counted as its own.
```

eudelins-zama: Or a proper docstring format instead

shared/ciphertext-attestation-client/src/fetch.rs
```diff
@@ -0,0 +1,138 @@
...
+
+/// Why a handle has no attestation consensus this round.
+///
+/// The two variants call for opposite responses, which is why they are distinguished here rather
+/// than collapsed into one opaque error: [`Self::Starved`] is the expected early state, since
+/// Coprocessors publish attestations asynchronously, so it is worth retrying. [`Self::Split`]
+/// means the signers that answered disagree and no vote still to come could break the tie —
+/// retrying re-reads the same disagreement.
+#[derive(Debug, thiserror::Error)]
+pub enum ConsensusCheckError {
```
eudelins-zama, line 25: We have `ConsensusCheckError`, `AttestationError`, `ValidationError`...

We should try to unify this into a single one if possible

PR #3397, opened 2026-07-31
chore(common): update codeowners

eudelins-zama:
Mainly did this update to share the ownership of `test-suite` to both fhevm-devs and mpc-devs.

AFAIU, in CODEOWNERS, the ordering matters: the last line to match a file wins the ownership. So I moved a bit things to be more coherent.
For ex, host-contracts was first assigned to gateway-team, then to copro-team. Only the last one (copro-team) was applied in reality, so I removed the first gateway-team ownership line.

I moved the shared "broad" shared ownership at the top of the file so it can be overriden more easily (ex: `test-suite` owned by mpc+fhevm-devs, but `test-suite/gateway-stress` is mpc-devs only now), and moved the Dockerfile ownership by `fhevm-devops` at the bottom (it was overriden by sub-team ownership, thus not applied)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,17 +3,13 @@
 # All pull request should be reviewed by at least one of the members of fhevm-devs
 *    @zama-ai/fhevm-devs
 
-# Production Dockerfiles should be reviewed by members of fhevm-devops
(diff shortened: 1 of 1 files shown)


isaacdecoded (approved):
LGTM!

PR #3414, opened 2026-07-31
feat(kms-connector): permit-based Solana user-decrypt authorization

KitHat:
# What

The v1 Solana user-decrypt authorization module for `kms-worker` (`core/solana`): strict request decoding, permit-envelope verification (via `zama-solana-permit` from #3364), deployment identity, an atomic host-state snapshot, and the per-handle rules — lineage resolution, handle binding, delegation freshness, invalidation watermark, KMS context/epoch servability — with an exhaustively-matched failure taxonomy.

Every check is a pure function of `(typed request, snapshot, deployment, now)`; `snapshot` is the only reader of chain state, so a race scenario is two snapshot values rather than two moments in time.

## Not wired in yet

The module is additive and unreachable from the runtime flow: the existing PoC path (ad-hoc wallet signature, one handle per request) keeps serving requests unchanged. `evm_path_unchanged.rs` pins the shared EVM surface (EIP-712 digest, type string, `extraData` container) byte-for-byte. Wiring the pipeline in and deleting the PoC verifier is a follow-up.

## Tests

- Per-rule integration suites under `crates/kms-worker/tests/solana_*`, including a race matrix and permit-vector parity (the normative set from #3364 run through the Connector's own `decode` + `check_signature`).
- A new normative Connector-authorization vector set (`solana/test-fixtures/connector-auth/`, regenerated via `solana/scripts/update-connector-auth-vectors.sh`) for non-Rust consumers to converge on.

## Plumbing

`zama-solana-permit` is a new path dependency, so the four kms-connector Dockerfiles copy it into the build context (same standalone-crate pattern as `zama-solana-acl`).

The PR size is huge mainly due to the tests. I tried to keep functionality in rather small separated files.


Eikix (commented):
## Automated review — consolidated findings

Reviewed at `5e82d626b5945b4cf086fdd7da88523138945034` against EVM `main`, the current Solana host, and the agreed permit design.

### Confirmed

- The Gateway already applies the 2048-bit request limit to `userDecryptionRequestSolana` through the same check used by EVM. The Connector should not repeat this check or copy the FHE type-size table.
- Keeping `extraData` is correct. It carries the KMS context and, when needed, the historical proof.
- The permit binds the user, deployment, validity window, allowed domains, handles, and KMS selection in the intended places.

### Recommended before this becomes the live authorization path

1. **Accept wildcard delegation.** The host and EVM accept either a delegation for the exact application account or a wildcard delegation. The new Connector reads only the exact account. Both must be checked; the signed domain restriction still applies.

2. **Handle a pre-funded permit-revocation account like the host does.** Anyone can send a small amount of SOL to the predictable address before it is initialized. The host treats a System Program-owned, empty, non-executable account as uninitialized. The Connector currently rejects it. The same rule should be used in both places; every other foreign or malformed account must still be rejected.

3. **Match the EVM expiry boundary.** A permit is expired when `now > end`, not when `now >= end`. It remains valid at the exact end timestamp.

4. **Split large Solana account reads without lowering the 2048-bit limit.** Keys should be deduplicated and split into requests of at most 100 accounts. Those requests may be sent as one JSON-RPC batch or with bounded concurrency. Every response used for one decision must have the same slot; otherwise the whole read must be retried. The deciding read must not be older than the earlier discovery read.

5. **Define one chain-ID rule before runtime wiring.** Recommended rule:

   ```text
   digest = SHA-256(ASCII("zama-solana-chain-id-v1") || base58_decode(genesis_hash))
   chain_id = 0x8000000000000000 |
              (big_endian_u64(digest[0..8]) & 0x7fffffffffffffff)
   ```

   Mainnet, Devnet, and Testnet genesis hashes are known and can be kept in a reviewed registry, so deployment does not need an RPC call. Custom and local clusters should provide an explicit genesis hash. The Connector should still compare its RPC endpoint's genesis hash with the selected cluster once at startup. Shared Rust and TypeScript test vectors are required.

### Non-blocking design choices

- The current two-read delegation flow is safe: the first read discovers the application account and the second makes the decision. If one read is preferred, the request can carry an untrusted application-account hint. It must then be compared with the validated lineage and used to recompute the value key. A false hint must only cause rejection.
- Checking stored PDA bumps is common and already done elsewhere in this repository. Adding the checks would be consistent hardening, but the canonical address, owner, and contents already provide the authorization boundary.
- EVM allows an older threshold-KMS epoch to be destroyed but protects the latest active epoch. A known destroyed epoch is terminal. Solana should reuse that shared lifecycle when wired. Centralized KMS has no threshold epoch to destroy.
- Request/response linking and the final runtime connection remain follow-up work. The chain-ID rule must be settled first.

The first four recommendations affect authorization or availability and should be resolved in this PR or before any runtime path can call this module. Chain-ID derivation and runtime linking can remain explicit follow-ups while this PR stays an unwired foundation.

KitHat:
**1. Wildcard delegation** — done. `check_delegation` now evaluates two rows: the lineage's
app-specific record and the delegator's wildcard record (reserved app-context sentinel, same
derivation). Either live row authorizes; neither can veto the other. Both addresses are planned
into the deciding read unconditionally, so the flow stays at two reads. When both rows exist and
neither is live, the failure reports both reasons (`NoLiveGrant`). The stored PDA bump is now also
checked against the canonical one on delegation and invalidation records.

**2. Pre-funded revocation account** — done. A System-owned, empty account at the invalidation
address reads as "never revoked" (watermark 0), same rule as the host program. System-owned with
data is still rejected; the executable combination is unreachable for the reason the host states.
Vector added.

**3. Expiry boundary** — done. The window is the closed interval `[start, start + duration]`;
expired only when `now > end`. Connector and Gateway now agree second for second. Vectors updated.

**4. Large reads** — the ordering half is done: the deciding read must not be older than the
discovery read (`DecidingReadOlderThanDiscovery`, transient — a retry on a caught-up node
succeeds). Chunking we'd rather not do: every rule is evaluated against one snapshot, and N chunks
are N slots — equal `context.slot` across chunks on a standard RPC is reachable only through retry
loops, and a merged observation is not one observation. Open question — enforce the account budget
at admission instead: with the wildcard row the worst case is 3 accounts per entry plus one
(`3N+1 ≤ 100` → `handles.length ≤ 33`), checked in `userDecryptionRequestSolana` next to
`EmptyHandles`, before the fee. That keeps the Connector's "never re-adjudicate an accepted
request" contract and doesn't depend on per-operator RPC config (`--rpc-max-multiple-accounts`
defaults to 100; managed providers pin it or lower). If agreed, this lands as a gateway-contracts
change.

**5. Chain id** — adopted your derivation verbatim as a deployment-time rule (now in the spec):
applied once per cluster, configured everywhere. The Connector consumes the configured value and
checks only the chain-kind bit — recomputing the rule at startup would be a second implementation
of it, and a wrong configuration fails loudly (every permit and every handle mismatches from the
first request). Open items: the startup genesis-hash cross-check belongs to runtime wiring, and
the shared Rust/TS vectors for the rule are still to be added alongside the permit fixtures.

Eikix (approved):

PR #3424, opened 2026-08-03
fix(coprocessor): park stalled dependency chains

Eikix:
Targets `release/0.14.x`, the repository's actual branch corresponding to the requested `release/v0.14.x`.

After the configured no-progress budget, the TFHE worker drops only its in-memory reference to the dependency-chain lock and stops extending that lease.

The PostgreSQL row remains untouched: it stays `processing` with the same owner, expiry, and scheduling timestamp, so the chain is neither forgotten nor marked processed.

Another available chain can run immediately, while the parked chain becomes eligible for the existing `ExpiredLock` work-stealing path after its existing TTL.

The no-progress counter, cross-priority early-lock fallback, and medium-happy-path retry remain unchanged; formatting, Cargo check, CI-aligned Clippy, and adversarial review pass, while the focused test compiles but local setup hits the existing `CompressedXofKeySet` fixture error before its body.

### DIFF
--- a/coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
+++ b/coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
@@ -582,6 +582,20 @@ impl LockMngr {
         self.lock.as_ref().map(|(lock, _)| lock.clone())
     }
 
+    /// Stop extending the current lock without modifying its database row.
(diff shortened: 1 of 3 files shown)


antoniupop (approved):

Eikix:
@mergifyio rebase

PR #3427, opened 2026-08-03
fix(coprocessor): sns-worker, S3 migration, sleep when processing only stalled errors

rudy-6-4:
## Summary

- Sleep on mostly done migration with only remaining stalled error
- Make sleep cancellable


obatirou (approved):

rudy-6-4:
@mergifyio queue

PR #3411, opened 2026-07-31
chore(solana): coherence pass — one vocabulary, no dead surface, APIs that say the truth

Eikix:
Executes zama-ai/fhevm-internal#1859 (spec and state; the issue's checkboxes track progress).

Targets `feature/solana` directly. (#3377 merged into it on 2026-07-31; this branch already contains it.)

## Waves (commit order)

1. **Docs scaffold** — GLOSSARY.md (normative), INVARIANTS.md (Part I: what the system guarantees / Part II: sizes, limits, and how it is run), README rewrite. ✅
2. **Pure deletions** — dead error variants, never-emitted events, tombstones, cosmetic constructors. ✅
3. **Decoder consolidation** — replace hand-mirrored structs and hand-computed discriminators with a shared `zama_host::decode` module built from the Anchor-generated `zama_host::instruction::*` types (host-listener and proof-store rewired; the listener's dead IDL-codegen'd decoder deleted). ✅
4. **Vocabulary renames** — execution (the old `batch`), dictionary, fhe_execute, persistent, update, create/created-public, domain/account/label (full rename including wire names, per recorded decision); trust-triad error names (attestation vs input proof); the nine never-emitted `Fhe*Event` structs reclassified as decoded op records in `zama_host::records` (listener rewired off its IDL events codegen); deny-list error names de-stratified; js-sdk internal vocabulary (wire keys and signed preimages untouched). ✅
5. **Type rewrites** — v3 relayer wire enum (wire-compatible), newtypes, API-shape fixes, decided items (ALT deactivate+close, vault move). Pulled forward and already landed: S1 deny-check on all grant paths, S2 pool-entry hygiene, S3 flat-authority docs, HCU u64::MAX convention.

Each thematic chunk gets an adversarial review pass before moving on; findings are fixed in dedicated commits referencing "§0c review gate".

## Operator note: HCU limit convention is a state break for existing deployments

[...]

PR #3432, opened 2026-08-03
fix(solana): harden confidential token settlement

Eikix:
## Summary

- Serialize confidential burns per `(mint, token account)` with one `PendingBurn`. A second burn is rejected until the first is redeemed or cancelled; no lanes or caller-selected burn IDs remain.
- Add owner-authorized cancellation that atomically restores encrypted balance and encrypted total supply, closes `PendingBurn`, and lets a cancelled batch return to refund-only withdrawal.
- Make deny-list enforcement grant-only. Token-account ACL/public-sealing operations are exposed through owner-authorized wrappers; encrypted total-supply operations are exposed through mint-authority wrappers.
- Support classic SPL Token and Token-2022 through `TokenInterface`, with fail-closed extension checks, frozen-account rejection, and token-program-aware vault validation.
- Bind disclosure to the exact token state field: mint, optional token account, encrypted value account authority, encrypted value label, handle, kind, encrypted value account, and cleartext amount are emitted and checked coherently.
- Keep wrapper mint authority separate from Zama Host upgrade governance. Governance can later be installed behind the wrapper authority without coupling it to Host code upgrades.
- Refresh generated clients, IDLs, ABI goldens, demo flows, glossary-aligned design docs, exact compute snapshots, and the live/coprocessor callers.

Closes tracking for [zama-ai/fhevm-internal#1862](https://github.com/zama-ai/fhevm-internal/issues/1862). The branch is rebased on `chore/solana-coherence-pass` (#3411).

## Safety properties

- At most one unsettled burn exists per confidential token account.
- Redeem and cancel require the `PendingBurn` handle to equal the current burned-amount handle and consume the same PDA exactly once.
- A failed redeem or batch settle leaves the burn, vault balances, encrypted supply, and batch state unchanged.
- Cancellation restores exactly the encrypted delta removed by burn and releases no underlying tokens.
- Token-account subject rotation and public sealing require the token owner; total-supply rotation and public sealing require the mint authority.
- Token-2022 mint extensions are rejected and token-account extensions are rejected except `ImmutableOwner`; frozen sources and destinations fail closed.
- Disclosure kind, account shape, authority, domain, label, canonical encrypted value address, and handle are checked before the host proof is consumed.

## Validation

- [x] `cargo fmt --check`
- [x] `cargo test --workspace`
- [x] `cargo test -p zama-host --features decode`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo clippy -p zama-host --features decode --all-targets -- -D warnings`
- [x] live e2e client `cargo check --all-targets`
- [x] exact Solana 2.1.0 compute snapshot regeneration and verification
- [x] IDL, schema-hash, ABI-golden, proof-store ingest, dead-surface, glossary-alias, and e2e-grep checks
- [x] JS SDK production build, codegen drift check, formatting, lint, types, and tests
- [x] demo-dapp frozen install, typecheck, lint, 239 tests, and production bundle
- [x] affected KMS Solana authorization suites and strict Clippy


Eikix:
Changes are needed before this PR can be considered complete for `fhevm-internal#1862`.

The app-authority ACL gate, removal of `initial_balance`, pre-funded PDA handling, `host_config` constraints, wrap negative tests, KMS runbook, and decision not to add new registry/governance machinery address substantial parts of the tracker. The remaining problems and proposed solutions are below.

## 1. Replace concurrent burn lanes with one sequential pending burn per token account

The current lane design permits several burns, but only the shared `burned_amount` account's current handle can be recovered. A later burn therefore turns every earlier lane into a KMS-only redemption path. This does not provide independent `redeem | recover` lanes and makes the lifecycle harder to reason about.

A simpler invariant should be used for this PoC:

`Ready -> PendingBurn -> Redeemed | Cancelled -> Ready`

Suggested change:

- Derive one pending PDA from `["pending-burn", mint, token_account]`.
- Reject a new burn while that pending account is active.
- Assert that no pending burn is active before the FHE burn is executed, then create it atomically in the same transaction. A pre-funded system account should remain safe to allocate/assign as in the current defense.
- Keep the shared burned-amount value current until the pending burn is settled.
- Require both settlement paths to match the pending account's mint, owner, token account, encrypted-value account, and current burned handle.
- Rename `recover_pending_burn` to `cancel_pending_burn`. Redemption releases the underlying tokens; cancellation restores the confidential balance and encrypted total supply. Both close the pending account.
- Remove `burn_id` unless a concrete client-correlation use remains. If retained, it should be documented as metadata rather than a uniqueness or replay mechanism.

The pending PDA is writable, so competing burns and competing redeem/cancel attempts will be serialized. The first successful settlement closes the account; later attempts fail. Tests should cover a second burn being rejected, a new burn after redeem, a new burn after cancel, redeem/cancel races, stale settlement data, and recovery by program-PDA owners such as the batcher.

Applications needing parallel exits can aggregate them into one app operation or use separate app-owned confidential token accounts. Multiple independently pending burns from the same token account can be deferred until a concrete use case requires the additional state. The parent issue's current “concurrent burns = concurrent lanes” decision must be updated to record this deliberate simplification.

Cancellation is a compensating operation, not a historical rollback: the original burn and events remain on-chain, and an amount already made public cannot become private again. That should be stated in the documentation.

## 2. Keep host deny enforcement on grant paths only

The documented policy says that host deny is grant-path only, but deny is still checked during redemption and during `remove_subject`. This turns the host grant denylist into token cashout policy and prevents a denied application authority from reducing an existing ACL.

Suggested change:

- Remove the deny account/check from `redeem_burned_amount`.
- Remove the deny check from `remove_subject` and from its token wrapper.
- Keep deny enforcement when membership is added through persistent create/update or `allow_subjects`.
- Add tests proving that a denied subject cannot be newly granted, a denied authority can still remove a subject, host deny alone does not block redemption, and an underlying frozen account does block the relevant token transfer.

## 3. Implement the accepted underlying-token interface honestly

The current `Program<Token>`, `Account<SplMint>`, and classic token-account types are still SPL-only. Token-2022 and fail-closed extension handling are documented as long-term even though they are required by the parent issue.

Suggested change:

- Use `Interface<TokenInterface>` and `InterfaceAccount` for the underlying mint and token accounts in initialize, wrap, and redeem.
- Make canonical vault derivation use the underlying mint's token-program ID instead of hard-coding `spl_token::ID`.
- Support plain classic SPL and plain Token-2022 mints.
- Reject unsupported Token-2022 mint shapes, including transfer fee, transfer hook, non-transferable, and Token-2022 confidential-transfer extensions.
- Cover classic SPL and Token-2022 wrap/redeem, wrong-program inputs, frozen source/destination/vault accounts, and each rejected extension.

Freeze enforcement at the underlying asset boundary comes from the token program's checked transfer. Mirroring a user's underlying freeze across internal confidential transfers is a separate and larger product decision; it should be recorded as deferred rather than implied by `TokenInterface` support.

## 4. Bind disclosure events to recognized token state

`disclose_secp` currently accepts any canonical host `EncryptedValue` whose domain equals the mint. An unrelated same-domain value can therefore produce an apparently token-scoped `HandleDisclosedEvent`. Indexer guidance cannot make that event trustworthy.

Suggested change:

- Make the disclosed value kind explicit, for example balance, total supply, or burned amount.
- Validate the complete shape for that kind: canonical encrypted-value address, application authority, label, and its relationship to the supplied mint, token account, or pending burn.
- Include the validated kind and relevant application account in the event.
- Add same-domain negative tests for the wrong label, application authority, encrypted-value address, and token/pending account.

Alternatively, the event can be made explicitly generic and include the raw authority and label, but it must no longer claim token-scoped semantics.

## 5. Complete total-supply auditor rotation using the existing mint authority

Token-account auditor rotation is wrapped correctly, but the total-supply PDA has no equivalent allow/remove path. `ConfidentialMint.authority` already exists and is set at initialization, but it is not currently used by any later instruction.

Suggested change:

- Add `allow_total_supply_subjects` and `remove_total_supply_subject`.
- Require the external signer to equal `mint.authority`.
- Validate the canonical total-supply PDA and `mint.total_supply_encrypted_value`, including its domain, label, and application authority.
- Invoke the host signed by the total-supply PDA.
- Keep auditors decrypt-only and keep the compute signer from becoming an ACL administrator.
- Add allow/remove success tests and wrong-authority/value-shape failures. Runtime coverage should also be added for the existing token-account remove wrapper.

`mint.authority` is the per-mint policy authority. It can later be initialized to, or controlled by, a multisig/DAO execution PDA. It is distinct from the authority that can upgrade or administer Zama Host. Host upgrade governance and mint-authority rotation are separate concerns and should remain out of scope for this PR.

## 6. Finish the parent-item and documentation coherence pass

- Reconcile the sections that still describe the removed permanent `BurnRedemption` marker or say that redemption was untouched.
- Replace lane/claim/recover terminology consistently with sequential pending burn/redeem/cancel.
- Expose an actual reusable ATA-like derive/create-for/get-or-create helper if SDK-level UX is intended; otherwise narrow the parent item and documentation to the demo-only behavior that exists.
- Record whether `leaf_index` is intentionally unnecessary rather than leaving the item without a decision.
- Update every parent checklist item with direct code, test, documentation, or explicit deferral evidence. Accepted requirements should not be silently converted into long-term items in the implementation PR.
- Regenerate IDLs and clients after the instruction shapes stabilize, then complete the checked test plan and require the final CI run to be green.
- Remove placeholder comments and reconcile the large design documents so that one current lifecycle and authority model is described.

For reviewability, the final commits should be grouped by decision: sequential burn lifecycle; deny policy; TokenInterface/freeze; disclosure binding; ACL wrappers; then tests/generated artifacts/docs. This will keep the security-relevant handwritten changes separate from generated churn.

Eikix:
## Deep review — agent pass against every point of fhevm-internal#1862

Reviewed by an agent (Claude Code) against the parent issue's 22 findings, 5 product decisions, and the 5-wave execution plan. Verdict: **sound core design, but not mergeable as-is** — CI is red, and three checklist items are reported done when they are not.

---

### Blocking

**B1. `solana-tests/build-and-test` fails.** Two cost-snapshot mismatches: `open_batch` 218662 → 218572 and `redeem_open_batch` 215663 → 215573, both −90 CU. The committed snapshots read ~90 CU *high*, i.e. they were regenerated incrementally rather than through `scripts/update-cost-snapshots.sh` (which runs `cargo clean` first) — the exact trap `docs/TESTING.md` documents. Mechanical fix. The PR's own test plan also has all 6 boxes unchecked.

**B2. This PR is 5 commits behind its own base.** `chore/solana-coherence-pass` has moved on, including `77ac54e2d refactor(solana): emit every host event through the event CPI, delete emit-events` — which touches the event surface this PR adds two events to. Rebase before merge.

---

### Issue coverage

| # | Item | Status |
|---|---|---|
| 1, 4, 19 | PendingBurn lane, seeds incl. `token_account`, rent freed | ✅ done well |
| 13 | ACL mutate = `encrypted_value_account_authority` | ✅ host + token wrappers + tests |
| 6 | dead `initial_balance` | ✅ removed, error variant retired in place |
| 21 | `host_config` constraints | ✅ complete — but see D2 |
| 14 | disclose indexer note | ✅ |
| 15 | KMS rotation runbook | ✅ |
| 16 | `wrap_usdc` negatives | ✅ 4 negatives added |
| 3, 9–12, 20 | doc notes + #1692 / #1634 links | ✅ |
| 2 | no token-owned denylist | ✅ stance documented |
| Wave 3 | mint-PDA consolidation | ✅ deferred *with* a written reason |
| 5, 18 | wrap amount event, `amount > 0` | ✅ correctly skipped |
| **7** | ATA-like doc **+ SDK get-or-create** | ⚠️ doc claims it exists; it does not (D3) |
| **8** | TokenInterface **or drop the Cargo feature** | ⚠️ neither done (D1) |
| **17** | recipient grief "covered by #13" | ⚠️ not closed (C2) |

---

### Correctness

**C1. The batcher — the only real consumer — still has no recover path.** `recover_pending_burn` takes `owner: Signer`, and a batch's owner is `batch_authority`, a PDA that can only sign via CPI. No batcher instruction CPIs recover. So a batch whose KMS cert never arrives has its joined funds stuck exactly as before. Finding #1 was accepted specifically *"(liveness/ops)"*; on the product surface that liveness is unchanged. Either wire a `recover` on the batcher, or state plainly that PendingBurn buys liveness only for direct-owner burns.

**C2. #17 should not be ticked.** `make_handle_public` is still `has_subject`-gated (`zama-host/src/instructions/encrypted_value.rs:182`). Any decrypt subject — including an auditor added through the new wrapper — can permanently seal the owner's *current balance handle* public. That is arguably the higher-severity half of #13: an auditor is trusted to *see* a balance, not to publish it irreversibly. The PR is honest about this (new `ANTI-11b`), which is the right instinct — but the issue checkbox and the "Partial — fixed by #13" note are now wrong.

**C3. Total-supply rotation is impossible, not "deferred."** Wave 1 made mutation authority-only, and nothing can sign as `["total-supply", mint]` for `allow_subjects`. Currently harmless — the audience is `compute_only(compute_signer)` (`initialize_mint.rs:73`), so there are no auditors to rotate — but it also means none can *ever* be added. "Deferred" understates it; better phrasing is "no auditor can be added to total supply until a wrapper exists."

**C4. The safety argument for closing the lane is asserted, not written.** The deleted comment warned: *"the marker is never closed, because closing it would make the payout replayable."* Closing **is** safe here, and the review confirmed why: per-burn keying, plus `pending.burned_handle == burned_handle` (`redeem_burned_amount.rs:167`), plus handles being content-addressed over `op/operands/type + slot entropy` (`fhe_execute/walk.rs:22-26`) — so every payout requires its own real burn and is pinned to that burn's cleartext. But that three-step argument appears nowhere. A future reader sees a warning deleted and a claim in its place. The argument belongs where the warning was.

---

### Decision-making

**D1. #8's cheap half was skipped.** `anchor-spl`'s `token_2022` feature is still enabled on `confidential-token` and `confidential-batcher` (`Cargo.toml:22` / `:28`) with **zero** `TokenInterface` / `token_2022` usage in either. The issue explicitly offered "*or drop Cargo feature until done*". Leaving it contradicts the standard the immediate parent branch just set in `f940acb3b`: *"A feature declaration that gates nothing is a claim that a knob exists when it does not."*

**D2. #21 was resolved in the expensive direction.** Adding `seeds` / `seeds::program` / `bump` to `host_config` on 7 token instructions is redundant for all 7: they only forward it into `fhe_execute`, which already constrains it (`fhe_execute.rs:55`). The cost is visible in the snapshots — roughly +1.8k CU on every affected instruction (`confidential_transfer` +1832, `claim` +1803, `join` +1842; `open_batch` +3230 ≈ 2× the `initialize_token_account` CPI), consistent with one extra `create_program_address` per instruction. Plus ABI churn: Anchor emits `pda` metadata into the IDL, so 7 pinned schema hashes changed despite unchanged args. It is load-bearing in exactly one place — `allow_token_account_subjects`, which reads `grant_deny_list_enabled` itself. Worth either a sentence justifying the defense-in-depth, or dropping it where the callee already checks.

**D3. A false doc claim.** The new Wave-4 stance in `DESIGN_DECISIONS.md` says ATA-like UX includes "*demo get-or-create in `demo-dapp`*". There is no get-or-create anywhere in demo-dapp — `buildInitializeTokenAccountInstruction` is an unconditional builder with no existence check. Finding #7 asked for "Doc + SDK"; the SDK half is not done and the doc says it is. Same class of defect as the base branch's §0c gate commit (*"correct what the event CPI actually guarantees, and five other false claims"*).

**D4. Hardcoded e2e `burn_id`.** `live-client/main.rs` uses `burn_id = 0x00..01` with "this PoC issues at most one burn per run." A run that dies between burn and redeem leaves the lane open, and the next run then fails with `PendingBurnAlreadyInitialized` — a confusing error for an unrelated cause. Deriving from the slot or keypair costs nothing.

---

### Style

- **S1. Commit hygiene regression.** 7 of 10 commits have **empty bodies**, and 6 are rebase-repair churn ("retarget X onto Y", "clear CI blockers"). The base branch's commits carry multi-paragraph rationale plus verification notes. Squash the retarget commits into the waves they repair, and write bodies.
- **S2. `INVARIANTS.md` `11b.`–`11f.` are not valid markdown list items** — they render as paragraphs and break list continuity. No other entry in the file uses letter suffixes. Renumber, or use nested bullets under 11.
- **S3.** `assert_encrypted_value_account_is_token_account` fails with `CurrentEncryptedValueMismatch` — it is an *authority* mismatch, not a current-handle one. Five precise `PendingBurn*` errors were added; this one deserves the same care.
- **S4.** `redeem_burned_amount` keeps `system_program` "for CPI/client compatibility" though claim no longer inits anything — a dead meta in a PR that already breaks that instruction's arg list.
- **S5.** In live-client `allow_for_decryption`, after the early return, `authority: account` and `authority: payer.pubkey()` are identical; using `account` implies flexibility that does not exist.

---

### What is genuinely good

- **`burn_id`-seeded lanes are a better design, not just a rename.** Making the PDA derivable pre-execution let `pendingBurn` move *into* the settle ALT and out of the static account list (`derive.ts`) — the old handle-seeded marker could not. A real structural win.
- `open_pending_burn` uses fund/allocate/assign, so a pre-funded lamport balance cannot grief creation.
- Wave 3/4 deferrals are documented with reasons rather than silently dropped.
- `ANTI-11b` / `11d` honestly disclose what was *not* fixed.
- `mollusk_redeem_historical_burned_handle_after_supersession_then_rejects_double_redeem` pins the exact act-once property, and the `allow_subjects` negatives cover peer-authority rejection on both host instructions.

---

### Verification notes

`dead-surface-check.sh` reported clean, and CI's own `solana-tests/dead-surface`, `demo-dapp`, `native-coverage`, and `host-listener-reconstruct` all pass — `build-and-test` is the sole red. Mollusk could not be run locally: the review machine's `solana` rustup toolchain link is broken and the 3.0.13 SBF toolchain cannot build `base64ct v1.8.3` (pre-existing; this PR touches no Cargo manifests), so test results were taken from the CI log.

Eikix:
# Deep review — audit triage verification against fhevm-internal#1862

Reviewed the full diff (66 files, +3694/−978) plus the head state after the retarget fixups, cross-checked against every triage item in the parent issue, and ran an adversarial pass over the new PendingBurn lane mechanics.

**Verdict: the PR is sound and addresses every actionable triage item, but should not merge as-is** — two stale doc passages contradict the new design, two leftover placeholder comments remain, and a handful of cheap negative tests are missing.

---

## 1. Parent-issue checklist verdict

### Wave 1 — #13 (BLOCKING): ACL mutate gate ✅
- Host gate changed from "any current subject" to `require_keys_eq!(authority, value.encrypted_value_account_authority, ...)` in both `allow_subjects` and `remove_subject` — same gate as persistent create/update. Exactly what the triage asked: decrypt subjects are no longer co-admins.
- Token wrappers (`allow_token_account_subjects` / `remove_token_account_subject`) CPI host via `invoke_signed` as the token-account PDA; owner signer gates both; deny-list forwarded only when enabled.
- Tests repurposed correctly: `mollusk_allow_subjects_rejects_unallowed_authority` now proves a **current subject** is rejected, plus a new remove-side peer-subject negative.
- Docs updated consistently (INVARIANTS HOLDS-11 / ANTI-11b split, EVM_PARITY, DD-032 amendment, ACL crate doc comment). #17's total-supply rotation is explicitly deferred and documented.

### Wave 2 — #1/#4/#19: PendingBurn lanes ✅
- `PendingBurn` replaces the forever `BurnRedemption` marker; seeds include `token_account` (**#4**), rent returns to owner via `close = owner` on both claim and recover (**#19**), unwrap liveness restored via `recover_pending_burn` (**#1**).
- The client-supplied `burn_id` design is the right call: it makes the lane derivable pre-tx so it can ride in the batcher's settle ALT (static set 14→13, lane added to the table), verified end-to-end in `derive.test.ts` / `settleBatch.test.ts` / `provision.test.ts` (count pinned at 23).
- Recover is genuinely tip-only (`pending.burned_handle == burned_value.current_handle`), restores **both** balance and total supply via dual FHE `add`, cross-checks all lane fields, and the execution authority retarget onto the token account is consistent at head.
- Batcher `burn_id = batch.key().to_bytes()` gives uniqueness by construction; settle re-derives the canonical lane address before the redeem CPI.

### Waves 3/4 — deferrals ✅ properly recorded
Both appear in DESIGN_DECISIONS "Open Product Decisions" with real rationale (mint-as-PDA consolidation: churn outweighs attack-surface reduction; freeze/TokenInterface: no token-owned sanctions list, SPL freeze on the wrapped mint, ATA-like UX intentional). These are decisions, not TODOs.

### Hygiene items
| # | Item | Verdict |
|---|---|---|
| 6 | dead `initial_balance` removal | ✅ Removed everywhere (instruction, batcher CPI, tfhe-worker fixture, live-client, demo-dapp builder + tests). Zero leftovers repo-wide. |
| 16 | wrap Mollusk negatives | ✅ Happy path + 4 negatives; `amount=0` documented as valid no-op. |
| 21 | `host_config` seeds alignment | ✅ Constraints added across token instructions. |
| 14 | disclose indexer note | ✅ |
| 15 | KMS rotation/destroy runbook | ✅ Four-step runbook under DD-040 (rotate ≠ revoke; `destroy` is the lever). |
| 3 | operator gap tracked → #1692 | ✅ EVM_PARITY rows annotated. |
| 2, 7, 8, 9–12, 20, 22 | docs/stance | ✅ |
| 5, 18 | skipped per triage | ✅ Correctly untouched. |

**Downstream consumers: no breakage.** SDK codegen deliberately prunes `redeemBurnedAmount`/`confidentialBurn` (batcher-internal CPI, never built by the SDK); test-suite demo files reference redeem only in comments about the unchanged vault-escrow requirement. IDLs and `solana_abi_golden.json` are in sync; `check_solana_abi.py` gained the three new instructions + the `PendingBurn` account check.

---

## 2. Adversarial review — no vulnerabilities found

- **Lane pre-funding grief:** `fund_allocate_assign` (transfer shortfall + allocate + assign signed as the program PDA) tolerates attacker-pre-funded lamports instead of Anchor `init`'s fail-closed check, with explicit post-checks replicating init's safety. Correct trade-off, documented.
- **Lane squatting:** attacker cannot pre-init — `allocate`/`assign` require the program-owned PDA signature; open path also verifies system-owned + empty + exact `data_len` post-write.
- **Claim/recover races:** both paths `close = owner`, so claim-after-recover and recover-after-claim fail on a closed lane. No double-spend.
- **burn_id reuse after lane GC:** a reopened lane stores the *new* `burned_handle`; an old KMS cert fails the exact-handle proof. Safe.
- **Rent asymmetry:** permissionless payer funds, owner receives rent on close — documented as "crank subsidy, not theft". Acceptable.
- **e2e CONSUME_SEAL with no re-grant:** safe — the burned delta is created `PersistentAudience::for_owner(owner, compute_signer)`, so the owner is already a subject at seal time, and `make_handle_public` remains subject-gated (ANTI-11b).

---

## 3. Issues to fix before merge

### 3a. Stale docs contradict the new design (blocking per the doc-code truthfulness rule)
1. **DESIGN_DECISIONS.md ~L1335 (DD-036):** *"Double-redeem is still prevented by the per-handle `burn-redemption` marker PDA (DD-022)"* — false after this PR; the lane PDA closed on claim/recover is now the replay boundary.
2. **DESIGN_DECISIONS.md ~L1738 (DD-040):** *"Marker retained as the sole persistent bit: … write-once, never-closed per-handle `BurnRedemption` PDA at `["burn-redemption", mint, burned_handle]` … a second redeem of the same handle fails on the `init`"* — the account no longer exists. The DD-032 amendment was updated; these two DDs were not. They need dated amendment notes (the established pattern).
3. **TESTING_STRATEGY.md:29:** *"the burn-redemption consume-once replay marker and expired-request rejection"* — stale test-surface description.

### 3b. Leftover scaffolding
- `token_mollusk.rs` L4770 and L5631: bare `// TAG60 placeholder` comments with no content — remove or expand.

### 3c. Test gaps (cheap, worth adding)
- No **double-recover** rejection test (second recover on a closed lane).
- No **recover-then-claim / claim-then-recover** mutual-exclusion test — the lane's defining property is only asserted one direction at a time.
- No `PendingBurnAlreadyInitialized` test (burn with an occupied lane).
- `remove_token_account_subject` has no **token-level** Mollusk coverage (only host-level remove tests exercise the gate).
- No negative for a **non-owner caller** of the token wrappers.

### 3d. Nit
- In `RecoverPendingBurn`, `burned_amount_value` is `#[account(mut)]` while its doc says "read as a persistent operand (left unchanged)". If the host CPI truly only reads it, drop `mut` (write-lock reduction); otherwise fix the comment.

---

## 4. Style & decision-making

**Decision-making: strong.** The load-bearing calls are well-reasoned and documented at every layer (code comments → INVARIANTS → DDs):
- Client-supplied `burn_id` instead of waiting for `burned_handle` — the only design that makes the lane derivable at ALT-provision time.
- `fund`/`allocate`/`assign` over Anchor `init` — correct grief-tolerance analysis with safety-preserving post-checks.
- Tip-only recover is the right boundary: no KMS dependency for undoable burns, historical burns keep the cert path.
- Deferral discipline is good — Waves 3/4 are recorded decisions with rationale, and total-supply rotation is scoped out of Wave 1 rather than half-done.

**Commit hygiene: acceptable for a triage PR.** The five `fix(solana): retarget …` fixups are rebase-absorption work against the coherence base; ideally squashed before review, but each is clearly messaged and the final state is coherent.

**Style nits:** the `TAG60 placeholder` comments are the only real noise; naming is otherwise consistent with the repo vocabulary (`encrypted_value_account_authority`, "encrypted value account", lane terminology), error ordinals preserved on retirement, test naming follows `mollusk_<flow>_<behavior>`.

---

## Recommendation

**Approve with required follow-ups:** fix the two DD-036/DD-040 stale passages and TESTING_STRATEGY.md:29, delete the `TAG60 placeholder` lines, and ideally add the recover/claim mutual-exclusion + lane-collision negatives. Note `solana-tests/build-and-test` was still pending at review time — wait for it before merging.

Eikix:
The PR has been reworked around a smaller settlement model and the unresolved points from the three review comments.

## Problems that were present

1. Concurrent burn lanes did not provide independent redeem/cancel semantics because every lane shared one current burned-amount handle. Older lanes could become KMS-only, and the extra lane and burn-ID state made replay reasoning harder.
2. Cancellation was not exposed by the batcher, so its real failure mode still left joined funds stuck.
3. Host ACL mutation and public sealing exposed too much power to decrypt subjects. Deny checks were also applied outside grant paths. Total-supply ACL administration had no mint-authority wrapper.
4. The implementation claimed underlying-token support that was still classic SPL-only and did not fail closed for Token-2022 extensions or frozen accounts.
5. Disclosure events were not bound tightly enough to recognized confidential-token state and could give unrelated same-domain values token-scoped meaning.
6. Generated clients, ABI fixtures, callers, cost snapshots, lifecycle documentation, glossary terms, and several adversarial cases had drifted from the intended design.

## Solutions and resulting direction

- One `PendingBurn` PDA is now derived per `(mint, token account)`. A second burn is rejected until the first is either redeemed or cancelled. No lanes or caller-selected burn IDs remain.
- Redeem and cancel both require the pending burned handle to equal the current burned-amount handle. Both consume the same PDA exactly once. Redeem releases the underlying tokens; cancel atomically restores encrypted balance and encrypted total supply without releasing underlying tokens.
- The batcher now has an owner-authorized cancel dispatch. A cancelled batch enters refund-only withdrawal, so the liveness fix reaches the actual consumer.
- Deny enforcement is grant-only. Token-account ACL audience management and public sealing require the token owner. Total-supply ACL audience management and public sealing require `ConfidentialMint.authority`. This wrapper authority can later be controlled by governance, but it remains separate from Zama Host upgrade authority.
- Classic SPL Token and plain Token-2022 are handled through `TokenInterface`. Unsupported mint extensions fail closed; token-account extensions fail closed except `ImmutableOwner`; frozen sources and destinations are rejected.
- Disclosure is checked against an explicit `Balance`, `TransferredAmount`, `BurnedAmount`, or `TotalSupply` kind and the exact mint, optional token account, domain, encrypted value account authority, encrypted value label, canonical encrypted value account, and current handle. The emitted event carries that validated shape.
- The reusable demo get-or-create helper now handles absent, system-prefunded, existing, and cross-owner confidential token accounts. Redundant wrapper-side `host_config` PDA checks were removed where the typed Host CPI already validates the account.
- Current documentation now describes one sequential burn lifecycle, uses `cancel` for the compensating operation, records that cancellation does not erase burn history or restore privacy, and keeps multi-way burns deferred until a concrete application requires them.

## Parent issue disposition

| Parent point | Resolution |
|---|---|
| #1, #4, #19 — burn liveness, seeds, rent | Replaced by the sequential `PendingBurn` lifecycle above. Seeds contain mint and token account; redeem/cancel close the account and return rent. `leaf_index` is intentionally not duplicated in the burn event because it is carried by the public-decrypt proof. |
| #2 — compliance/deny policy | No token-owned sanctions list was added. Underlying token freeze is enforced at wrap/redeem, and Host deny remains grant-path only. |
| #3 — operators/callbacks | Kept out of scope and linked to #1692 in the parity/design documentation. |
| #5 — wrap amount event | Intentionally skipped because the clear amount is already instruction data and is reconstructible without another event. |
| #6 — dead `initial_balance` | Removed from the instruction and every caller/generated surface. |
| #7 — ATA-like confidential account UX | Canonical derive/create/get-or-create behavior is implemented in the demo client, including idempotent and prefunded-account cases. |
| #8 — honest underlying-token interface | Implemented with classic SPL + Token-2022 coverage, program-aware vaults, extension rejection, and frozen-account negatives. |
| #9–#12 — registry, pause, observer, governance | No speculative registry/observer/governance machinery was added. Pause interaction and the explicit deferrals remain documented. |
| #13 — peer ACL administration | Host mutation requires the encrypted value account authority; token wrappers map that authority to token owner or mint authority. Decrypt subjects cannot administer peers. |
| #14 — disclosure forgery | Replaced the same-domain-only check with the exact kind/shape binding described above. |
| #15 — live KMS contexts | The rotation/destroy runbook is retained, and tests cover live rotated-out contexts plus destroyed-context rejection. |
| #16 — wrap negatives | Wrong owner, mint, vault, total-supply authority, token program, extensions, and frozen-state cases are covered. |
| #17 — recipient/auditor grief | Both ACL mutation and public sealing are authority-gated; an auditor can decrypt but cannot manage peer subjects or publish the owner's current value. |
| #18 — zero wrap | Kept as the documented valid no-op; no unsupported metering policy was invented. |
| #20 — low-64 context ID | Kept as the intentional EVM-compatible extraction and documented as such. |
| #21 — `host_config` consistency | Local checks remain only where wrapper code reads Host configuration; forwarded-only accounts rely on the typed Host CPI's canonical validation, avoiding duplicate CU and ABI churn. |
| #22 — registry/mint authority | No registry was added. The existing mint authority is now used for total-supply policy wrappers and can later be assigned to governance without becoming Host upgrade authority. |
| Mint-PDA consolidation decision | Deferred: collapsing mint, compute, and total-supply authorities would add churn without improving the sequential settlement invariant. The authority boundaries are explicit and usable without that migration. |

## Adversarial acceptance

- **Double/triple spend:** a second pending burn, double redeem, double cancel, redeem-after-cancel, and cancel-after-redeem are rejected. A new burn succeeds only after the prior pending PDA has been consumed.
- **Stale or substituted state:** wrong handles, labels, authorities, domains, token accounts, value kinds, account shapes, and non-canonical encrypted value addresses are rejected before proof consumption.
- **PDA grief/replay:** system-prefunded pending PDAs are allocated safely; occupied or foreign accounts fail closed; settlement is pinned to the exact current burn and the pending PDA closes once.
- **Authority separation:** non-owners cannot mutate or publish token-account state; non-mint authorities cannot mutate or publish total supply; auditors remain decrypt-only; denied subjects cannot be newly granted but existing access can be removed and redemption is not blocked by the grant denylist.
- **Asset boundary:** wrong token programs, unsupported Token-2022 extensions, frozen accounts, wrong vaults, and mismatched mint/account ownership are rejected.
- **Batch rollback:** failed settle leaves the burn, encrypted balances, encrypted supply, and batch state unchanged; cancel restores the exact encrypted delta and permits only refunds.

The branch is a single commit rebased on the current `chore/solana-coherence-pass` tip (`e51e23dd5e`). The 89 confidential-token Mollusk tests, 25 batcher Mollusk tests, full Solana workspace tests, KMS Solana authorization tests, strict Clippy/format checks, IDL/schema/ABI reconstruction, dead-surface and glossary checks, exact Solana 2.1.0 cost snapshots, JS SDK checks, live client, and demo application all pass locally. The GitHub Actions run for `8eb43a9bb6` is green.

Eikix:
Follow-ups carried over from the test-integrity review on #3411, collected here because this PR is where several of them land. None of them is a blocker for this PR; the first one is the only one that this PR actively adds to.

### A cross-language check for seed literals — this PR adds a new one

`pending-burn` is now written in Rust (`state/mod.rs:64`, plus `confidential_burn.rs:616`, `redeem_burned_amount.rs:65`, `cancel_pending_burn.rs:45`) and independently in TypeScript twice (`batcherPdas.ts:13` as `PENDING_BURN_SEED`, and `dispatchBatch.test.ts:91` as a hand restatement). Nothing compares any TS copy to the Rust.

That is the general shape across the dapp — every PDA seed exists twice in TypeScript and once in Rust. The goldens are built honestly enough to catch account reordering, a wrong label, or a changed program id, but a changed Rust seed leaves all 240 dapp tests green and fails first as a live `ConstraintSeeds` error. With `pending_burn` now inside the settle transaction, that failure would land in settle rather than in CI.

A small script asserting the TypeScript seed constants equal the Rust byte literals would close it for all of them at once. #3411 lists it as a follow-up; this PR is where the newest instance appears.

### `check-e2e-greps.py` producer sets are one blob, and this PR edits the live client

The checker's producer set was widened to `solana/programs`, so a grep is satisfied by the literal appearing in *any* producer file — including a doc comment. 20 of the 55 labels are currently vouched for that way.

The one that matters is `output encrypted value`, the capture behind six ACL assertions at `full-vertical.sh:349-438`: it appears both in the live client and in a doc comment in `zama-host/src/state/mod.rs`. Since this PR changes 127 lines of `main.rs`, it is worth knowing the guard would not notice if one of those prints were dropped — the failure would revert to a silent abort deep into the vertical, which is exactly what the guard exists to prevent. Making the producer set per-grep rather than one concatenated blob is the fix.

### `solana-e2e` does not run for a PR based on `chore/solana-coherence-pass`

Its trigger is `branches: [feature/solana, elias/solana-proof-service-recovery]`, so this PR's green check set contains no live vertical, despite it changing `full-vertical.sh` and 127 lines of `live-client/main.rs`. A run has been dispatched manually at `8eb43a9bb6` (`actions/runs/31084262800`). Adding the coherence-pass branch to the trigger would cover stacked PRs generally.

### Three test surfaces run in no workflow

`bun test demo` and `test:e2e:harness` are invoked nowhere in `.github/`, no shell script, and no other package — roughly 90 test cases, including `demo/lifecycle.test.ts`, which is where #3411 put a new exports-map guard written specifically to catch a deleted `@fhevm/sdk` subpath. Separately, `test-suite/fhevm`'s only `tsc --noEmit` and only `bun test src` sit behind an `e2e` pull-request label. Anything this PR adds to those files will not be executed by CI either.

### The reconstruction-first negative control no longer exists, and three comments still claim it

Deleting the `emit-events` feature removed the `--no-default-features` host build that used to prove the coprocessor was fed by Yellowstone reconstruction rather than by emitted events. `setup-solana-side.sh:46` still says "disable optional host events", `:245` says "(the program emits nothing)", and `solana-e2e.yml:8` says "zama-host disables optional events". The listener still has exactly one event decoder, so nothing is broken today — what is gone is the test for it.

### Small inherited cleanups, if this PR is touching those files anyway

- `docs/TESTING.md:62` still prints `cargo test -p zama-solana-runtime-tests --test plan_contracts`; the target is `execution_contracts`, so the command as written errors. The alias sweep cannot see it because `_` is a word character to grep.
- The `lineage` rename left doubled words in seven connector test names, e.g. `a_encrypted_value_account_account_owned_by_another_program_is_terminal`.
- Three comments in the dapp's vault tests (`settleBatch.test.ts:185-194`, `provision.test.ts:143-145`, `derive.test.ts:70-72`) describe a stronger binding than the assertion beneath them performs.

Eikix:
Second review pass, focused on the substance of the hardening: the invariants register, the burn/redeem/cancel state machine, the FHE re-credit, the Token-2022 handling, the new Mollusk suites, and the glossary. Process and test-infrastructure follow-ups are already collected in the previous comment and are not repeated here.

**Verdict.** The state machine is sound as reviewed: every new invariant entry (11c–11f, 55–60) corresponds to an enforced check with a matching negative test, cancel restores exactly what burn removed in both balance and total supply, the authority separation holds, and the Token-2022 checks fail closed. Two defects were found, both cheap to fix: the rewritten INVARIANTS #24 pins a test that does not exist, and the live client's disclosure-kind mapping uses the wrong label for transferred amounts, in a branch no CI path exercises. Below those, a short list of specific missing negative tests and glossary observations.

---

### 1. INVARIANTS #24 pins a nonexistent test — CONFIRMED

`solana/docs/INVARIANTS.md:170` anchors the act-once rule to
`mollusk_redeem_pending_burn_after_encrypted_value_update_then_rejects_double_redeem`. No test by that name exists anywhere in the tree (`grep -r` over `solana/` finds only the register itself). The test that actually covers this is `mollusk_redeem_current_pending_burn_then_rejects_double_settlement` (`solana/runtime-tests/tests/token_mollusk.rs:3491`), and the name difference is not cosmetic: the retired test redeemed a *historical* burned handle after an update, which the new model forbids (`redeem_burned_amount.rs:160-163` requires the pending handle to be current). The register entry was reworded for the new model but given a test name that was never created. Since the whole point of the "pinned from both sides by" convention is that the named tests are the enforcement anchor, the name should be corrected; the other two pinned names (`mollusk_two_sequential_burns_each_redeemable_exactly_once`, `mollusk_disclose_secp_is_idempotent_no_replay_marker`) both exist.

### 2. Live client maps the wrong label for TransferredAmount disclosure — CONFIRMED

`solana/scripts/e2e/live-client/main.rs:1259-1261` classifies a value for `make_token_account_handle_public` by comparing its label against `encrypted_transfer_amount_label()` (`"transfer_amount_________________"`) and mapping it to `DisclosedValueKind::TransferredAmount`. But a real transferred-amount value carries `encrypted_transferred_amount_label()` (`"transferred_amount______________"`) — that is the label the transfer output is created with (`common.rs:203`) and the label both the wrapper (`allow_token_account_subjects.rs:174`) and `disclose_secp` (`disclose_secp.rs:113`) require for that kind. Consequences:

- A genuine transferred-amount account falls through to the `Err("TS_ACL is not a token-account disclosure field")` arm, so the live client cannot seal a transferred amount at all.
- The branch as written can never match anything the program accepts: no program instruction creates a persistent value with the `transfer_amount` label (its only production caller in the entire tree is this line), and if one existed, the wrapper would reject it with `DisclosedValueBindingMismatch`.

CI cannot see this because `full-vertical.sh:503` only ever seals the burned amount (`BurnedAmount` arm). The fix is a one-word change to `encrypted_transferred_amount_label()`. Related cleanup: `encrypted_wrap_amount_label()` (`state/mod.rs:96`) now has zero production callers — only a test fixture uses it.

### 3. Missing negative tests — specific cases

- **Token-2022 account-extension allowlist has no negative.** `assert_supported_underlying_token_account` (`common.rs:516-534`) rejects any account extension other than `ImmutableOwner`, but no test presents a disallowed one. The mint-extension arm has one negative (`mollusk_wrap_rejects_token_2022_mint_extensions`, NonTransferable); the account arm has only the positive `ImmutableOwner` fixture. The interesting adversarial case is an extension the *account owner* can add after creation regardless of the mint — `CpiGuard` or `MemoTransfer` — since the extension-free-mint rule does not exclude those. One wrap or redeem negative with such an account would pin the arm.
- **Token-level cancel by a non-owner is untested.** The gate exists (`cancel_pending_burn.rs:83-93`: `OwnerMismatch` plus the token-account PDA shape) and the composed batcher path has an unauthorized-cancel negative (`batcher_mollusk.rs:1778-1791`), but nothing exercises a stranger signing `cancel_pending_burn` directly against someone else's pending burn. Given that cancel moves encrypted value, the direct negative is worth its lines.
- **Frozen classic-SPL accounts are only tested through Token-2022 fixtures.** Low risk — the check (`common.rs:512-515`) is shared and program-independent — noted for completeness.

Otherwise the suites test what their names claim. Spot-checks for wrong-reason passes came back clean: `CleartextLedger` re-evaluates the *emitted* `fhe_execute` CPI bytes rather than pinning expected values (`token_mollusk.rs:174-231`), so the cancel test's 750→1000 / 4750→5000 restoration (`token_mollusk.rs:4035`) is genuinely derived from the instruction the program produced; `mollusk_two_sequential_burns_each_redeemable_exactly_once` runs real burns and asserts the on-chain MMR peaks equal an independent reconstruction (`token_mollusk.rs:3862-3866`) before proving against them; the dapp's `cancelDispatch.test.ts` re-derives every PDA from raw seeds rather than calling the code under test. Cancel-after-redeem and redeem-after-cancel are both covered (`token_mollusk.rs:3541-3549`, `:4106-4122`); stale-handle cancel is covered with full atomicity assertions (`:4127`); the occupied-PDA rejection checks both a foreign-owned and a dirty system-owned account (`:3396`).

### 4. Invariants — verified against the code

Each entry this PR adds or rewrites was checked for an enforcing `require` rather than prose:

- **#11 / 11b:** `zama-host` now gates `allow_subjects`, `remove_subject`, and `make_handle_public` on signer == `encrypted_value_account_authority` (`encrypted_value.rs:43-47`, `remove_subject.rs:29-33`, `encrypted_value.rs:180-184`), with host-level negatives (`host_mollusk.rs:742`, `:780`, `:1536`) and token-wrapper negatives (`token_mollusk.rs:1110`, `:1143`, `:1165`).
- **11c:** the availability check runs before the FHE execution (`confidential_burn.rs:391` precedes `fhe::execute` at `:525`) and again inside `open_pending_burn` (`:611`); the PDA is created fund/allocate/assign so pre-funding cannot grief (`:654-681`, tested at `token_mollusk.rs:3368`).
- **11d:** `cancel_pending_burn.rs:128-131`. **11e:** the cancel execution is `add(balance, burned)` + `add(total_supply, burned)` (`cancel_pending_burn.rs:169-177`), the exact inverse of burn's `sub` pair (`confidential_burn.rs:480-486`); since burn defines `burned = balance - select(ge, balance-amount, balance)`, the re-credit is exact whether the burn debited `amount` or 0. **11f:** `fhe_execute` is pause-gated host-side (`fhe_execute.rs:80`), redeem/disclose additionally locally (`common.rs:432-448`, tested at `token_mollusk.rs:3935`).
- **#55–57:** the disclosure binding checks domain, kind→authority, kind→label, and canonical PDA (`disclose_secp.rs:79-137`), with a kind/label-mismatch negative (`token_mollusk.rs:4891`); the token-program pin is ownership-based and re-checked on initialize, wrap, and redeem (`initialize_mint.rs:49`, `wrap_usdc.rs:73-81`, `redeem_burned_amount.rs:95-103`).
- **#58–60:** total-supply mutation requires `has_one = authority` plus the canonical value shape (`total_supply_subjects.rs:18`, `:69-95`); batcher cancel requires the wrapper authority (`cancel_dispatch.rs:74-78`), and `Refunding` rejects dispatch, settle, join, and repeat cancel (`dispatch.rs:68`, `settle.rs:170`, `join.rs:91`, `cancel_dispatch.rs:66`) while `quit.rs:80-86` admits refunds — all exercised in `mollusk_cancel_dispatch_restores_burn_and_allows_refunds`.
- The `fhe_execute` deny-list narrowing (output authority deny-checked only when the output actually adds a subject, `fhe_execute.rs:520-529`, `:563-569`) is mirrored in preflight account-marking (`preflight.rs:365-385`), tested from both sides (`token_mollusk.rs:2124`; the cancel-under-deny sections of `token_mollusk.rs:4077-4092` and `batcher_mollusk.rs:1764-1775`), and — closing an open question from the prior pass — the product rationale for grant-only denial is now recorded in DD-045 (`DESIGN_DECISIONS.md`: "Applying it to settlement can trap funds after policy changes").

Entry #10 is textually unchanged by this PR and still holds under the new code (new grants are still per-subject deny-checked on all three membership paths, `encrypted_value.rs:48-76`).

### 5. Glossary

The four new burn-lifecycle rows match the code they describe. Three observations:

- **"settled" and "cancel" are now homonyms across the two reference programs.** The glossary defines *settled* as the terminal adjective for a pending burn, but `BatchStatus::Settled` (`confidential-batcher/src/state.rs:58`) means "payout received; claims open", and the batcher has both `BatchStatus::Canceled` (zero-total settle — nothing restored) and `cancel_dispatch` (burn restored → `Refunding`). A reader applying the glossary's *cancel* row ("restores the encrypted balance…") to `BatchStatus::Canceled` gets the wrong prediction. A parenthetical in the two rows, or batcher rows of their own, would close it.
- **The new retired aliases are unenforced.** The rows retire *burn lane*, *claim burn*, *recover*, and *finalized*, but `dead-surface-check.sh`'s alias sweep gained no entries for them (the script is untouched by this PR). A manual sweep finds the retired terms absent from the owned trees today, so this is only a durability note — and the sweep self-describes as expiring scaffolding, so skipping may be deliberate; if so, that is worth saying in the row.
- `DisclosedValueKind` and `Refunding` have no rows. Both are cross-language vocabulary now (Rust enum, IDL, generated TS, event field). Minor.

One doc-comment inaccuracy in the same vein: `confidential_burn.rs:176-181` and `confidential_transfer.rs:177-181` say a value "from any app" is spendable "once its owner has granted the mint's compute subject via `allow_token_account_subjects`" — that wrapper signs only as this program's token-account PDA, so a genuinely foreign app's value must be granted by that app's own authority through host `allow_subjects` directly. The "cross-app" test (`token_mollusk.rs:5646`) uses a token-account-authority value, so the truly foreign case is neither described accurately nor tested.

---

### What was checked, and what could not be

Checked on the PR tree (`git diff e51e23dd5e..origin/fix/solana-audit-1862`, extracted to a scratch copy): full read of the changed confidential-token, confidential-batcher, and zama-host instruction code; `cargo check -p confidential-token -p confidential-batcher` passes; `check-pda-seeds.py` and its `--self-test` pass and the script is wired into `solana-tests.yml:146-147`, closing the cross-language seed follow-up for the one seed this PR adds; INVARIANTS, GLOSSARY, DESIGN_DECISIONS, and EVM_PARITY read against the code they describe.

Not verified: the Mollusk suites were reviewed statically only — executing them requires `cargo build-sbf` artifacts, which must not be built on this machine while the demo validator is running (the manually dispatched `solana-e2e` run 31084262800 at this head is the executed evidence). Of the demo-dapp vitest suite, 136 tests across 22 files pass in the review environment; the other 19 files import the built `@fhevm/sdk` package, which was not rebuilt here, so their pass/fail status rests on the PR's CI claim. The overflow behavior of the cancel-path FHE `add` was reasoned about (bounded by 1:1 SPL backing, per EVM_PARITY's `tryIncrease` divergence note) but not tested anywhere; no violating path was found.

PR #3433, opened 2026-08-03
fix(coprocessor): replay quorum-accepted input proofs

Eikix:
## Problem

During a rolling upgrade, one coprocessor can reject an input proof while the Gateway accepts it with the configured quorum. The accepted input can then be used on-chain, but the rejecting coprocessor has no local ciphertexts for computations that depend on it.

A local rejection alone cannot tell us whether the Gateway ultimately accepted or rejected the proof. The final `VerifyProofResponse` and `RejectProofResponse` events are therefore used as the source of truth.

## Change

- Retain a locally rejected proof after its rejection response is sent.
- Keep retained proofs out of both the transaction-sender queue and the ZK-worker queue while the Gateway outcome is unknown.
- On final Gateway acceptance, replay verification locally and require the reconstructed handles to exactly match the Gateway-approved handles.
- Insert reconstructed ciphertexts without sending another Gateway response.
- On final Gateway rejection, delete the retained proof.
- Make final-acceptance handling idempotent so event redelivery does not reset replay backoff.
- Reuse the existing `verify_proofs` columns; no migration or new replay table is introduced.

The retained-state encoding is deliberately small:

- local result ready to send: `verified = false`, non-NULL empty `handles`;
- rejection sent, awaiting Gateway: `verified = false`, `handles = NULL`;
- Gateway accepted, replay pending: `verified = NULL`, consensus `handles` present.

## Retry policy

The first failed reconstruction is retried after one minute. Delay then doubles up to a 24-hour cap. Automatic retries remain eligible for seven days. After that window the proof payload is retained but is no longer selected automatically.

Replay uses the currently loaded key and CRS. A proof that no longer verifies fails closed, no ciphertext is inserted, and the payload remains retained. This minimal version does not persist the original key/CRS identities, so it cannot proactively disable attempts at the exact rotation boundary.

## Rollout requirement

This is a coordinated three-service change. Deployment order must be:

1. `gw-listener` and wait for it to catch up;
2. `transaction-sender`;
3. `zkproof-worker`.

The listener-to-sender transition must happen with new proof traffic paused or otherwise drained. An old transaction sender deletes locally rejected payloads, while an old listener ignores final response events. An unordered rolling deployment can therefore still lose the replay payload or consume the final event before the new behavior is active.

## Limitations

- This prevents recurrence after deployment; it does not recover historical proofs whose payload was already deleted.
- The listener does not rescan final events before its saved cursor.
- Proofs with no observed final Gateway outcome are retained intentionally.
- #3424 remains complementary worker-fairness protection while an input is unavailable.

Related context:

- https://github.com/zama-ai/fhevm-internal/issues/1860
- https://github.com/zama-ai/fhevm-internal/issues/1861

## Validation

- Gateway-listener integration tests cover acceptance, rejection, and duplicate acceptance delivery.
- ZK-worker integration tests cover successful local replay, handle mismatch retention/backoff, and malformed NULL input rejection.
- Transaction-sender tests cover retaining a sent local rejection, including the already-rejected contract response.
- The complete transaction-sender suite and both retained-rejection signer variants pass.
- SQLx metadata was regenerated from the release schema.
- Repository pre-commit formatting, workspace `cargo check`, and CI-aligned Clippy checks pass.


Eikix (commented):
## Review

Traced the full state machine across all three services, checked the race windows, idempotency, channel wiring, and the SQL backoff math. The design is tight and the edge cases hold up — approving with a few minor findings.

### Verified correct

**State encoding is consistent across all services.**
- Worker SELECT filter, `get_remaining_tasks` mirror query, and tx-sender SELECT all agree on what's actionable; the count query was updated in lockstep.
- `is_replay = verified_at.is_some()` is sound: fresh inserts never set `verified_at`, and the only writers are the worker's local-result path and the acceptance handler.

**Idempotent final-acceptance.** The WHERE clause `verified IS DISTINCT FROM TRUE AND (verified = FALSE OR handles IS NULL)` is exactly right. Enumerated all redelivery states: replay-pending and replay-failed rows (verified=NULL, handles non-NULL) don't match, so duplicate events can't reset backoff; verified=TRUE rows are protected; deleted rows no-op.

**Race windows are safe:**
- Acceptance landing while tx-sender has a rejection in flight: acceptance UPDATE wins (verified=FALSE matches), and both landed/reverted reject-tx paths resolve — on revert, `retain_rejected_proof`'s `WHERE verified = FALSE` no-ops against the already-transitioned row.
- Worker holds the `FOR UPDATE` lock during replay: gw-listener's UPDATE/DELETE blocks until commit, then re-evaluates WHERE against committed state. No lost updates.
- Late local success after quorum acceptance: verified=TRUE rows are never converted; the late-response revert paths (`CoprocessorAlreadyVerified`/`CoprocessorAlreadyRejected`) now branch on `locally_verified` and self-heal.
- Ciphertext insertion + row deletion share one transaction, so a concurrent final-reject DELETE can't strand state either way.

**Backoff math checks out**: retry 1 → 60s, doubling via `1 << LEAST(GREATEST(retry_count-1,0),11)`, capped at 86400s; 7-day eligibility anchored to acceptance time (`verified_at` untouched by the failure UPDATE); `last_retry_at IS NULL` makes freshly accepted rows immediately eligible.

**Other details**: event signatures match `IInputVerification.sol` exactly; the NULL-input path setting `handles = ''::BYTEA` is a necessary fix (otherwise those rows fall out of tx-sender's new `handles IS NOT NULL` predicate); `verify_proof_request` stays `ON CONFLICT DO NOTHING` idempotent, so batch re-processing on handler error is safe; `begin_write_guarded` fencing is applied uniformly. Suppressing the tx-sender channel notify on replay correctly avoids waking it for work it can't select.

### Findings (non-blocking)

1. **Minor — new event arms lack error logging/metrics.** `verify_proof_accepted`/`verify_proof_rejected` propagate errors with a bare `?`, while the `VerifyProofRequest` arm wraps with `inspect_err` + `VERIFY_PROOF_FAIL_COUNTER`. A DB blip on a final event restarts the batch silently. An `inspect_err` log at minimum would give operational parity.

2. **Minor edge — NULL-input row in replay.** A replay-pending row with NULL input hits the NULL branch, which writes `verified=false, handles=''` regardless of `is_replay`, converting it back to local-reject state; tx-sender then fires a pointless `rejectProofResponse` at an already-finalized Gateway (reverts → retained). Self-healing, and can't realistically occur since a NULL input could never reach quorum acceptance.

3. **Operational — retained rows without a final outcome accumulate indefinitely.** `remove_proofs_by_retry_count` now only GCs `verified = TRUE`, so rows whose final event was missed before the listener cursor live forever. Acknowledged as intentional in the description — suggest a follow-up for observability, e.g. a gauge or audit query on `verified IS DISTINCT FROM TRUE AND verified_at < NOW() - interval '7 days'`.

4. **Deployment-order fragility is real but documented.** The runbook (gw-listener → tx-sender → zkproof-worker, proof traffic drained between steps 1 and 2) is correctly identified as mandatory: an old tx-sender deletes retained payloads, and an old worker treats replay-pending rows as fresh work and overwrites consensus handles. Since this targets `release/0.14.x`, make sure the rollout order lands somewhere operators will see it (chart notes or the coprocessor upgrade runbook), not just the PR body.

Eikix (commented):
## Review of `921a0761f4` (trace proof replay recovery)

Reviewed the delta against `51f4ee90c9`. This is a pure observability/test-hardening follow-up — no state-machine changes — and it addresses finding #1 from my earlier review. LGTM.

### gw-listener
- Both new event arms now wrap with `inspect_err` + `error!` including `zk_proof_id` before propagating, matching the `VerifyProofRequest` arm's logging parity. ✓
- The new `debug!` logs on the 0-rows-affected branches (acceptance found no eligible retained proof / rejection found nothing to discard) are well placed — that's exactly the signal needed to diagnose duplicate or out-of-cursor final-event delivery.

### zkproof-worker
- `is_replay` threaded through the `Process zk-verify request` log, `verify_task`/`db_insert` spans, and `Completed` — replay attempts are now cleanly separable from fresh verification in traces.
- Demoting replay verification failures from `error!` to `warn!` ("Gateway-accepted proof replay failed; retained for retry") is the right call: these are expected and retriable (e.g., key/CRS rotation), while fresh verification failures stay `error!`.
- `replayed_ciphertext_count` is only set inside the verified branch (past the handle-mismatch guard) and logged after `txn.commit()`, so no success log can fire on a rolled-back path.

### Tests
The strengthened `test_quorum_accepted_proof_is_replayed_only_locally` is the standout: it now deletes the ciphertexts from both `ciphertexts` and `ciphertexts_branch` before forcing the replay and asserts the full count is restored afterward. The previous version ran with ciphertexts already present from the initial verification, so it never actually exercised the restoration path this PR exists for — now it genuinely simulates the rejecting coprocessor's missing-ciphertext state and proves recovery. Bindings are correct (`handles` as `BYTEA[]` for the delete/count, concatenated blob for `verify_proofs.handles`).

### CI at head
All coprocessor suites pass: `zkproof-worker`, `gw-listener`, `transaction-sender`, `common`, clippy.

No new findings. Prior non-blocking items (#2 NULL-input-in-replay edge, #3 retention observability follow-up, #4 runbook visibility) remain as noted.

Eikix:
### Proof lifecycle

```mermaid
flowchart TD
    A[Proof request] --> B{Local verification}

    B -->|Accepted locally| C[Send acceptance response]
    C --> D[Ciphertexts available locally]
    D --> E[Remove proof row]

    B -->|Rejected locally| F[Send rejection response]
    F --> G[Retain proof while awaiting Gateway]

    G -->|Gateway rejects| H[Remove proof row]
    G -->|Gateway accepts| I[Replay proof locally]

    I -->|Verification succeeds<br/>and handles match| J[Insert local ciphertexts]
    J --> E

    I -->|Verification fails<br/>or handles differ| K[Retain and back off]
    K -->|Retry due within 7 days| I
    K -->|7 days elapsed| L[Stop automatic retries<br/>retain for investigation]
```

The row encoding used to implement this flow is:

| State | `verified` | `handles` | Meaning |
|---|---:|---|---|
| Fresh request | `NULL` | `NULL` | Await local verification |
| Acceptance ready | `TRUE` | Computed handles | Send acceptance response |
| Rejection ready | `FALSE` | Empty, non-NULL | Send rejection response |
| Awaiting Gateway | `FALSE` | `NULL` | Rejection sent; preserve proof |
| Replay pending | `NULL` | Gateway handles | Gateway accepted; reconstruct locally |

`is_replay` only distinguishes the first and last rows where `verified = NULL`:

- Fresh verification sends a Gateway response.
- Replay verifies against the Gateway-approved handles and never sends another response.

If Gateway finalization races local processing, the listener moves the retained row directly into either removal or replay; the same terminal paths apply.

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
```diff
@@ -486,61 +516,122 @@ async fn execute_verify_proof_routine(
...
-                    if let Some(bn) = block_number {
-                        insert_input_handles(&mut txn, cts, bn).await?;
+                    if is_replay && handles_bytes != expected_handles {
+                        error!(
+                            request_id,
+                            expected_handles = expected_handles.len() / 32,
+                            computed_handles = handles_bytes.len() / 32,
+                            "Replayed proof produced handles that differ from Gateway consensus"
+                        );
+                        verification_error =
```
goshawk-3, line 526: Should we add a counter metric to track the number of these errors? We would expect it to increase mainly during rolling upgrades. Otherwise, an increase could indicate an issue.

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
```diff
@@ -518,6 +535,109 @@ impl<P: Provider<Ethereum> + Clone + 'static> GatewayListener<P> {
...
+               AND verified IS DISTINCT FROM TRUE
+               AND (verified = FALSE OR handles IS NULL)",
+            zk_proof_id,
+            handles,
+        )
+        .execute(tx.as_mut())
+        .await?;
+
+        if result.rows_affected() > 0 {
+            info!(
```
goshawk-3: This should be also treated as a warning as it's kind of abnormal behavior.

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
```diff
@@ -518,6 +535,109 @@ impl<P: Provider<Ethereum> + Clone + 'static> GatewayListener<P> {
...
+                zk_proof_id,
+                "Cutover completed — skipping proof replay scheduling"
+            );
+            return Ok(());
+        };
+        let result = sqlx::query!(
+            "UPDATE verify_proofs
+             SET verified = NULL,
+                 verified_at = NOW(),
+                 handles = $2,
```
goshawk-3, line 568: Why don't we just check if these (input) handles already exist in the `ciphertexts` table as non-NULL ciphertexts?

Eikix: I'm not sure I understand what it saves

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
Eikix: I think the metric should never increase right?
It'd be quite catastrophic if handles computation differ between copros?

coprocessor/fhevm-engine/gw-listener/src/gw_listener.rs
goshawk-3: Since `verify_proofs` is ephemeral, e.g  its rows are removed after the retry policy is exhausted. The single source of truth is the existence of the input handles in the `ciphertexts` table.

goshawk-3: E.g if it happens that we receive `verify_proof_accepted` for handles that are missing in "ciphertexts" table and there is no entry in `verify_proofs` for them then we should report it as a critical error as it's unrecoverable edge case.

coprocessor/fhevm-engine/zkproof-worker/src/verifier.rs
goshawk-3: My point is about counting inputs that have been rejected and needs to be replayed during the rolling upgrade.

goshawk-3 (approved):
LGTM

PR #3436, opened 2026-08-04
refactor(host-contracts): reduce canonical snapshot export to the env map

obatirou:


host-contracts/tasks/utils/upgradeProposal.ts
```diff
@@ -92,6 +92,8 @@ export async function executeUpgradeProposal(hre: HardhatRuntimeEnvironment, pre
   console.log(
     `Executing prepared upgrade on ${prepared.proxyAddress} (implementation ${prepared.newImplementationAddress})...`,
   );
+  // Print the arguments this payload encodes, so an operator sees the state that gets deployed.
+  console.log('decodedArgs:', toJsonString(prepared.decodedArgs));
```
isaacdecoded: nit: maybe make the `'decodedArgs'` more explicit, e.g., `'Decoded payload arguments'`

host-contracts/README.md
```diff
@@ -143,9 +143,9 @@ npx hardhat task:exportCanonicalProtocolConfig \
   --out canonical-protocol-config-snapshot.json
 ```
 
-The artifact records the canonical chainId, the block number the read was pinned to, the
-contract address, the active KMS context id, the active epoch id, the KMS node set, and all
-four thresholds (bigints serialized as strings).
+The artifact holds a single `export` object. That object is the snapshot as a flat `KEY=value` map,
+with bigints serialized as decimal strings. A deployment platform patches those keys into a
+ConfigMap. The ConfigMap then sets the environment variables that the apply tasks in step 3 read.
```
isaacdecoded: Mentioning the `ConfigMap` concept isn't a bit of coupling with k8s? I mean, the end user shouldn't necessarily use ConfigMaps.

isaacdecoded (approved):
LGTM! 👍

isaacdecoded (approved):

PR #3446, opened 2026-08-05
feat(common): add compressed key material migration

Eikix:
## Context

This is the complete 0.15 implementation of the [same-key compressed material migration](https://github.com/zama-ai/tech-spec/pull/478), tracked by [the delivery plan](https://github.com/zama-ai/fhevm-internal/issues/1621).

Contracts, KMS Connector, and coprocessor support are combined because the changed ABI is not safe or testable as separate deployments. The migration ceremony remains deferred to the approved 0.15.1 window.

## Changes

- extend the existing key-generation lifecycle with `existingKeyId`;
- keep fresh generation unchanged with `existingKeyId = 0`;
- use KMS `UseExisting + CompressedAll` for a non-zero active key ID and copy the result under that key ID;
- publish migration completion through `ActivateKey` without changing the active protocol key;
- stage the compressed material only after host-chain finality and verify its signed digest before updating the existing key row;
- keep Blue workers on legacy material when `FORCE_LEGACY_SERVER_KEY=true`;
- default the Helm chart to the safe legacy setting, while fresh e2e and GPU validation explicitly use `false`.

## Compatibility boundary

The `keygen` selector and the `PrepKeygenRequest`, `KeygenRequest`, and `ActivateKey` topics deliberately change. No key generation may be in flight during upgrade, and governance must wait until Contracts, every Connector, and every authoritative coprocessor listener use this version.

The full 0.15 to 0.15.1 Blue-Green rehearsal remains in [the rollout PR](https://github.com/zama-ai/fhevm/pull/3464). KMS completion semantics and production readiness gates remain tracked in [the delivery plan](https://github.com/zama-ai/fhevm-internal/issues/1621).

## Validation

- 91 `KMSGenerationTest` tests pass;
- the previously failing Connector `PrepKeygenRequest` integration test passes on the combined tree;
- Rust checks pass for the common key loader, listener, and all three workers;
- Helm renders legacy selection by default and the fresh preview override disables it;
- Docker e2e and GPU CI explicitly select compressed-capable behavior.

The component reviews are preserved in [the Connector PR](https://github.com/zama-ai/fhevm/pull/3447) and [the coprocessor PR](https://github.com/zama-ai/fhevm/pull/3448).


host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -68,19 +68,20 @@ interface IKMSGeneration {
     /**
      * @notice Emitted to trigger an FHE key generation.
      * @param prepKeygenId The ID of the preprocessing keygen request.
-     * @param keyId The ID of the key to generate.
+     * @param requestId The ID of the key generation request.
+     * @param existingKeyId Zero for a fresh key, or the existing key whose compressed material is requested.
      * @param extraData Additional context data.
      */
-    event KeygenRequest(uint256 prepKeygenId, uint256 keyId, bytes extraData);
+    event KeygenRequest(uint256 prepKeygenId, uint256 requestId, uint256 existingKeyId, bytes extraData);
```
eudelins-zama: I would avoid the `keyId` -> `requestId` renaming.

It would break consistency with other events, without much benefit.
It also minimizes the amount of changes needed in kms-connector as well (for ex in current draft PR, you forgot to rename the DB column `key_id` -> `request_id`, but we don't necessarily want it because it would break backward compat' of the migration)

Eikix: good point! 

the only thing is that the reader such as @jatZama and @isaacdecoded will need to know that keyId is either:
- in a normal key generation: the keyId variable points to the new id of the newly generated key 
- in the compressed key migration: the keyId variable points to a migration request id which will be used to store the migrated material in S3, but the underlying protocol key id is `existingKeyId` since it's a same key id migration as per RFC28 and 29

host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -55,7 +55,7 @@ interface IKMSGeneration {
      * @param paramsType The type of the parameters to use.
      * @param extraData Additional context data (0x01 || contextId, or 0x02 || contextId || epochId).
      */
-    event PrepKeygenRequest(uint256 prepKeygenId, ParamsType paramsType, bytes extraData);
+    event PrepKeygenRequest(uint256 prepKeygenId, ParamsType paramsType, uint256 existingKeyId, bytes extraData);
```
isaacdecoded, line 59: `existingKeyId` is not listed in the NatSpec above.

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -257,31 +251,31 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
...
-        // The counter is initialized at deployment such that keyId's first byte uniquely
+        // Generate a globally unique requestId for the key generation.
+        // The counter is initialized at deployment such that requestId's first byte uniquely
         // represents a keygen request, with format: [0000 0100 | counter_1..31]
-        // We generate the keyId in the preprocessing step in order to anticipate the introduction
-        // of key lifecycle: the keyId will be set to `Generating` status here
-        // See https://github.com/zama-ai/fhevm-internal/issues/185
         $.keyCounter++;
-        uint256 keyId = $.keyCounter;
+        uint256 requestId = $.keyCounter;
```
isaacdecoded: We can also keep the current `keyId` naming for variables, as Simon E. suggested for the event. I can see you are renaming it in several parts of the contract, but not sure this renaming is really needed/meaningful - it seems more like a renaming absorbed based on the KMS side gRPC endpoints.

host-contracts/contracts/interfaces/IKMSGeneration.sol
Eikix: nice catch

host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -53,9 +53,10 @@ interface IKMSGeneration {
      * @notice Emitted to trigger an FHE key generation preprocessing.
      * @param prepKeygenId The ID of the preprocessing keygen request.
      * @param paramsType The type of the parameters to use.
+     * @param existingKeyId Zero for a fresh key, or the existing key whose compressed material is requested.
      * @param extraData Additional context data (0x01 || contextId, or 0x02 || contextId || epochId).
      */
-    event PrepKeygenRequest(uint256 prepKeygenId, ParamsType paramsType, bytes extraData);
+    event PrepKeygenRequest(uint256 prepKeygenId, ParamsType paramsType, uint256 existingKeyId, bytes extraData);
```
isaacdecoded, line 59: nit: just an idea I had, maybe `compressingKeyId` is a naming that explains better what this "existing key ID" is included/propagated for. Although, I'm not fully aware of the current `existingKeyId` choosing reasons.

host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -68,27 +69,29 @@ interface IKMSGeneration {
...
+     * @param existingKeyId Zero for a fresh key, or the existing key whose compressed material is requested.
      * @param extraData Additional context data.
      */
-    event KeygenRequest(uint256 prepKeygenId, uint256 keyId, bytes extraData);
+    event KeygenRequest(uint256 prepKeygenId, uint256 keyId, uint256 existingKeyId, bytes extraData);
 
     /**
      * @notice Emitted when a KMS node has responded to a keygen request.
-     * @param keyId The ID of the key.
+     * @param keyId The key generation request ID.
```
isaacdecoded: nit: this is not only an ID of the keygen request, but also the ID of the key itself. Maybe we can be more explicit here.

host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -282,13 +319,26 @@ interface IKMSGeneration {
...
      * @param signature The signature of the KMS node that has responded.
      */
     function keygenResponse(uint256 keyId, KeyDigest[] calldata keyDigests, bytes calldata signature) external;
 
+    /**
+     * @notice Get the compressed materials added by a completed migration.
+     * @dev This getter is only for material produced by `keygen` with a non-zero existing key ID.
+     * Normal key generation continues to use `getKeyMaterials`.
+     * @param keyId The migrated key ID.
+     * @return keyMaterialId The KMS request and storage ID of the compressed material.
```
isaacdecoded: nit: or `compressedKeyId` instead of `keyMaterialId`

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -137,12 +137,12 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
...
         // Keygen state variables:
         // ----------------------------------------------------------------------------------------------
-        /// @notice The number of keygen, used to generate the keyIds.
+        /// @notice The number of keygen requests, used to generate their IDs.
         uint256 keyCounter;
-        /// @notice Bidirectional mapping between preprocessing request IDs and key IDs
+        /// @notice Bidirectional mapping between preprocessing and keygen request IDs
         mapping(uint256 id => uint256 pairedId) keygenIdPairs;
-        /// @notice The digests of the generated keys
-        mapping(uint256 keyId => KeyDigest[] keyDigests) keyDigests;
```
isaacdecoded, line 145: nit: I'd keep the current naming `uint256 keyId` since requestId could be a little ambiguous, i.e., the key digests are tied to a key ID, while a request ID sounds like any kind of request.

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -167,6 +167,13 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
         uint256[] completedKeyIds;
         /// @notice CRS IDs that reached consensus.
         uint256[] completedCrsIds;
+        // ----------------------------------------------------------------------------------------------
+        // Compressed key material production variables:
+        // ----------------------------------------------------------------------------------------------
+        /// @notice The durable key augmented by a migration request.
+        mapping(uint256 requestId => uint256 existingKeyId) existingKeyIdByRequestId;
```
isaacdecoded, line 174: nit: what about something like:

```
mapping(uint256 compressedKeyId => uint256 keyId) keyIdsByCompressedKeyId;

mapping(uint256 keyId => uint256 compressedKeyId) compressedKeyIdsByKeyId;
```

Btw, from older state mapping variables, we use plural like `keyDigests`, i.e., requestIdByExistingKeyId should be then request**Ids**ByExistingKeyId

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -215,37 +222,24 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
...
      */
-    function keygen(ParamsType paramsType) external virtual onlyACLOwner {
+    function keygen(ParamsType paramsType, uint256 existingKeyId) external virtual onlyACLOwner {
+        if (existingKeyId != 0) {
+            _checkMigrationRequest(paramsType, existingKeyId);
+        }
         KMSGenerationStorage storage $ = _getKMSGenerationStorage();
 
-        // Check that the previous keygen request has reached consensus
-        // Exception for the first keygen request, which has no previous key (counter is KEY_COUNTER_BASE)
```
isaacdecoded, line 248: nit: why remove this comment line?

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -271,17 +262,20 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
         $.keygenIdPairs[prepKeygenId] = keyId;
         $.keygenIdPairs[keyId] = prepKeygenId;
 
-        // Store the FHE params type, used for both the preprocessing and the key generation
-        // This value can later be read through the `getKeyParamsType` function, once the key
-        // has been generated
+        // Store the FHE params type used by both request phases.
```
isaacdecoded:
```suggestion
        // Store the FHE params type used by both request phases, prepKeygen and keygen.
```

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -351,6 +344,11 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
             revert EmptyKeyDigests(keyId);
         }
 
+        uint256 migrationKeyId = $.existingKeyIdByRequestId[keyId];
```
isaacdecoded: IMO, these many ways to name variables are very confusing: migrationKeyId or existingKeyId? [The suggestion here](https://github.com/zama-ai/fhevm/pull/3446/changes#r3765553399) can probably improve this.

host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -282,13 +319,26 @@ interface IKMSGeneration {
...
+     * @return kmsNodeStorageUrls The storage URLs of the KMS nodes that reached consensus.
+     * @return keyDigests The digests of the compressed material.
+     */
+    function getCompressedKeyMigrationMaterials(
```
isaacdecoded:
```suggestion
    function getCompressedKeyMaterialsFromMigration(
```

Can I ask why we need a separate view function only for compressed key materials from migration? Why not just reusing the current `getKeyMaterials` and let the compression be abstract/transparent?

host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -282,13 +319,26 @@ interface IKMSGeneration {
...
+        uint256 keyId
+    ) external view returns (uint256 keyMaterialId, string[] memory kmsNodeStorageUrls, KeyDigest[] memory keyDigests);
```
isaacdecoded:
```suggestion
    ) external view returns (uint256 keyId, string[] memory kmsNodeStorageUrls, KeyDigest[] memory keyDigests);
```

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -503,14 +528,14 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
...
-        // Keep abort available until the key lifecycle itself is done.
+        // The prep request reaches consensus before the paired keygen request does.
+        // Keep abort available until the keygen request itself is done.
         uint256 keyId = $.keygenIdPairs[prepKeygenId];
         if ($.isRequestDone[keyId]) {
             revert AbortKeygenAlreadyDone(prepKeygenId);
         }
 
-        // Mark both the prep-keygen and its associated key as done to unblock
+        // Mark both request phases as done to unblock the next keygen.
```
isaacdecoded:
```suggestion
        // Mark both request phases, prepKeygen and keygen, as done to unblock the next keygen.
```

host-contracts/contracts/KMSGeneration.sol
```diff
@@ -351,6 +344,11 @@ contract KMSGeneration is IKMSGeneration, EIP712Upgradeable, UUPSUpgradeableEmpt
             revert EmptyKeyDigests(keyId);
         }
 
+        uint256 migrationKeyId = $.existingKeyIdByRequestId[keyId];
+        if (migrationKeyId != 0) {
```
isaacdecoded: Shouldn't we check the compressed key digests for non-migration flows? I mean, the compressed key digests are generated only from the "migration path," or will all keygens include the compressed key digest from now on?

Eikix: Nice catch! I agree that starting 0.15, every new key will be compressed so we shouldcheck too

Eikix: Is this correct @dd23

host-contracts/contracts/interfaces/IKMSGeneration.sol
Eikix: Thinking again, I think you are right! the only consumer of that getter is the relayer,
we dont need one for the migration

host-contracts/contracts/interfaces/IKMSGeneration.sol
Eikix: existingKeyId is become the migration does not change the active key id. it's still the same compact public key, underlying private key
it's jsut that we add one more representation to that same, already existing, key


and the thing is in keygen() and KMS semantics, when we trigger a key generation event, the operation is identified by its key ID

so what's a bit hard at first to understnad is that:
- migration keygen() generates a new keyId, that new key Id is used to identify the migration request, but the key Id itself will not be "used" in the sense that it will not replace the active key id.
- so existingKeyId is to say -> "we are migrating that already active keyId"

host-contracts/contracts/KMSGeneration.sol
Eikix: I wouldn't do that because there is no distinct compressed key ID. The mappings express request ↔ existing key, and each lookup stores one value, so singular names are appropriate.

host-contracts/contracts/interfaces/IKMSGeneration.sol
```diff
@@ -68,27 +69,29 @@ interface IKMSGeneration {
     /**
      * @notice Emitted to trigger an FHE key generation.
      * @param prepKeygenId The ID of the preprocessing keygen request.
-     * @param keyId The ID of the key to generate.
+     * @param keyId The ID used for the key generation request.
+     * @param existingKeyId Zero for a fresh key, or the existing key whose compressed material is requested.
      * @param extraData Additional context data.
      */
-    event KeygenRequest(uint256 prepKeygenId, uint256 keyId, bytes extraData);
+    event KeygenRequest(uint256 prepKeygenId, uint256 keyId, uint256 existingKeyId, bytes extraData);
```
jatZama, line 76: I lack context here, but are we sure this event was never used anywhere else already? because changing events signature should be avoided otherwise since this is a breaking change, especially here as it could be avoided since the extraData argument could be used exactly in this case to avoid breaking signatures changes.

Eikix: I think we settled on the fact that this event is config only, listened to only by kms-connector
And @eudelins-zama accepted that breaking chance since there'll be no key generation event during the upgrade window

jatZama: ok just wanted to make sure this was already considered by mpc team, fine then.

isaacdecoded (dismissed):
LGTM! ✅

antoniupop (dismissed):

eudelins-zama (approved):

antoniupop (approved):

Eikix:
@mergifyio queue

PR #3447, opened 2026-08-05
feat(kms-connector): add compressed key material migration

Eikix:
## Context

This is the KMS Connector part of the [same-key compressed material migration](https://github.com/zama-ai/tech-spec/pull/478), tracked by [the delivery plan](https://github.com/zama-ai/fhevm-internal/issues/1621).

It is based on [the host-contract change](https://github.com/zama-ai/fhevm/pull/3446). Its diff contains only `kms-connector/**` files.

This PR completes the Connector side of the breaking contract-binding boundary. No key generation may run until the contract, every Connector, and every authoritative coprocessor listener understand the new request and `ActivateKey` topics.

## Changes

- persist `existingKeyId` from `PrepKeygenRequest` and `KeygenRequest`;
- preserve the existing `keyId` event field and `key_id` database column;
- use zero for fresh key generation and a non-zero `existingKeyId` for migration;
- send `GenerateAll + CompressedAll` for fresh key generation;
- send `UseExisting + CompressedAll` for migration, preserving the existing key tag;
- forward `keyId` as the KMS `request_id` and copy the migrated result to `existingKeyId`;
- reuse the current KMS RPC, polling, and response paths.

No migration-specific listener event, response path, or routing layer is added.

## Validation

- focused fresh and migration request-construction cases, including the copy-to-existing flag;
- offline checks for all affected Connector targets;
- mock-contract compilation;
- SQLx metadata checked against the additive schema.

The production ceremony remains gated on confirming that the active key metadata is accepted by KMS `UseExisting` and that KMS does not expose a successful result before the copy to the existing key ID commits.


kms-connector/crates/kms-worker/src/core/event_processor/kms.rs
```diff
@@ -132,3 +172,24 @@ const COMPRESSED_XOF_KEY_SET_CONFIG: KeySetConfig = KeySetConfig {
...
+        assert_eq!(
+            config.secret_key_config,
+            KeyGenSecretKeyConfig::UseExisting as i32
+        );
+        assert_eq!(
+            config.compressed_key_config,
+            CompressedKeyConfig::CompressedAll as i32
+        );
+    }
+}
```
eudelins-zama, line 202: Please no 🫠

kms-connector/crates/tx-sender/src/core/kms_response_picker/picker.rs
```diff
@@ -192,7 +192,8 @@ impl DbKmsResponsePicker {
                     LIMIT $1 FOR UPDATE SKIP LOCKED
                 ) AS resp
                 WHERE keygen_responses.key_id = resp.key_id
-                RETURNING resp.key_id, key_digests, signature, created_at, otlp_context
+                RETURNING resp.key_id, key_digests, signature, keygen_responses.created_at,
+                    keygen_responses.otlp_context
```
eudelins-zama: Why this change?

kms-connector/crates/tx-sender/src/core/ethereum.rs
```diff
@@ -119,13 +119,13 @@ where
...
-            response.signature.into(),
-        );
+        let key_digests = response.key_digests.into_iter().map(|k| k.into()).collect();
+        let signature = response.signature.into();
+        let call = self
+            .kms_generation_contract
+            .keygenResponse(response.key_id, key_digests, signature)
+            .into_transaction_request();
 
-        let call = call_builder.into_transaction_request();
```
eudelins-zama, line 128: Why this change?

kms-connector/connector-db/migrations/20260805100000_rfc029_compressed_key_migration.sql
```diff
@@ -0,0 +1,7 @@
+-- Keep the existing key identifier available through both keygen phases.
+-- A NULL value identifies an ordinary fresh-key request.
```
eudelins-zama: While it's good to allow null for backward compatibility of DB migration, I think this claim is wrong.

What identify an ordinary fresh-key request from now on is if this field is `0`, right?

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -268,13 +268,16 @@ async fn publish_prep_keygen_request<'e>(
     created_at: DateTime<Utc>,
     otlp_ctx: PropagationContext,
 ) -> anyhow::Result<PgQueryResult> {
+    let existing_key_id = (!request.existingKeyId.is_zero())
+        .then(|| request.existingKeyId.to_le_bytes::<32>().to_vec());
```
eudelins-zama: Why this extra logic at listener level? Let's KISS, and just insert it with `.as_le_slice()` like other IDs, and let the kms-worker handle the logic (check if 0 etc, it already have to check if it's null or not anyway...)

kms-connector/crates/gw-listener/src/core/publish.rs
```diff
@@ -292,11 +295,14 @@ async fn publish_keygen_request<'e>(
...
```
eudelins-zama: Same

kms-connector/crates/gw-listener/tests/common/mod.rs
```diff
@@ -181,6 +181,19 @@ pub async fn mock_event_on_gw(
...
+                existingKeyId: rand_key_id,
+                ..Default::default()
+            };
+            let tx = test_instance
+                .kms_generation_contract()
+                .keygen(ParamsTypeDb::Test as u8, rand_key_id)
+                .send()
+                .await?;
+            (tx, event.into())
+        }
```
eudelins-zama: Not sure this `TestEventType::CompressedKeyMigrationKeygen` makes sense at gw-listener level.

With unified keygen events approach, migration and keygen are handled the same way on listener side, no need to duplicate keygen tests here.

kms-connector/crates/gw-listener/tests/common/mod.rs
```diff
@@ -376,8 +391,13 @@ pub fn check_event_in_db(rows: &[PgRow], event: ProtocolEventKind) -> anyhow::Re
...
             for r in rows {
-                if e.prepKeygenId
-                    == U256::from_le_bytes(r.try_get::<[u8; 32], _>("prep_keygen_id")?)
+                if e.existingKeyId
+                    == r.try_get::<Option<[u8; 32]>, _>("existing_key_id")?
+                        .map(U256::from_le_bytes)
+                        .unwrap_or_default()
+                    && (e.prepKeygenId.is_zero()
+                        || e.prepKeygenId
+                            == U256::from_le_bytes(r.try_get::<[u8; 32], _>("prep_keygen_id")?))
```
eudelins-zama: This look weird to me (why check `e.prepKeygenId.is_zero()`?)

Let's revert to how it was

kms-connector/crates/gw-listener/tests/catchup.rs
```diff
@@ -13,7 +13,10 @@ use tracing::info;
 #[case::user_decryption_v2(TestEventType::UserDecryptionV2)]
 #[case::prep_keygen(TestEventType::PrepKeygen)]
 #[case::keygen(TestEventType::Keygen)]
+#[case::compressed_key_migration(TestEventType::CompressedKeyMigrationKeygen)]
```
eudelins-zama: To remove IMO

kms-connector/crates/gw-listener/tests/integration_test.rs
```diff
@@ -19,6 +19,7 @@ use tracing::info;
 #[case::user_decryption_v2(TestEventType::UserDecryptionV2)]
 #[case::prep_keygen(TestEventType::PrepKeygen)]
 #[case::keygen(TestEventType::Keygen)]
+#[case::compressed_key_migration(TestEventType::CompressedKeyMigrationKeygen)]
```
eudelins-zama: To remove IMO

kms-connector/crates/utils/src/types/grpc.rs
```diff
@@ -14,7 +14,10 @@ use tonic::Response;
 pub enum KmsGrpcRequest {
     PublicDecryption(PublicDecryptionRequest),
     UserDecryption(UserDecryptionRequest),
-    PrepKeygen(KeyGenPreprocRequest),
+    PrepKeygen {
+        request: KeyGenPreprocRequest,
+        routing_id: RequestId,
+    },
```
eudelins-zama: I'd avoid this change if possible, don't see why this would be required tbh

kms-connector/crates/kms-worker/src/core/event_processor/kms.rs
```diff
@@ -76,17 +76,24 @@ where
...
-            keyset_config: Some(COMPRESSED_XOF_KEY_SET_CONFIG),
-            keyset_added_info: None,
+            keyset_config: Some(keyset_config(existing_key_id)),
+            keyset_added_info: (!existing_key_id.is_zero()).then(|| KeySetAddedInfo {
+                from_keyset_id_decompression_only: None,
+                to_keyset_id_decompression_only: None,
+                existing_keyset_id: Some(u256_to_request_id(existing_key_id)),
+                use_existing_key_tag: true,
+                copy_compressed_key_to_original: false,
+            }),
```
eudelins-zama: Nit: a helper function to choose the `keyset_added_info` value, with a (very) short comment to explain why it's `None` for fresh, why it's this for a migration would be nice

kms-connector/crates/kms-worker/src/core/event_processor/kms.rs
```diff
@@ -132,3 +148,85 @@ const COMPRESSED_XOF_KEY_SET_CONFIG: KeySetConfig = KeySetConfig {
...
+        async fn validate_context(&self, _extra_data: &ExtraData) -> Result<(), RequestCheckError> {
+            Ok(())
+        }
+    }
+
+    #[rstest]
+    #[case::fresh(U256::ZERO, false)]
+    #[case::migration(U256::from(42), true)]
+    #[tokio::test]
+    async fn prepares_keygen_request(#[case] existing_key_id: U256, #[case] is_migration: bool) {
```
eudelins-zama: Nit: `is_migration` could (should?) be derived from `existing_key_id` instead of manually defined here.

The results comparison is a bit tautological, but I guess it's still good enough to test the migration branching, which is why this test is useful

kms-connector/tests/contracts/KMSGenerationMock.sol
```diff
@@ -14,16 +14,18 @@ contract KMSGenerationMock {
 
     enum KeyType {
         Server,
-        Public
+        Public,
+        Reserved,
```
eudelins-zama, line 18: What is this `Reserved` case? Hallucination?

eudelins-zama: Ah no it's from the real contracts indeed. Closing

kms-connector/crates/utils/src/types/event.rs
```diff
@@ -383,7 +386,10 @@ pub fn from_keygen_row(row: &PgRow) -> anyhow::Result<ProtocolEvent> {
     let kind = ProtocolEventKind::Keygen(KeygenRequest {
         prepKeygenId: U256::from_le_bytes(row.try_get::<[u8; 32], _>("prep_keygen_id")?),
         keyId: U256::from_le_bytes(row.try_get::<[u8; 32], _>("key_id")?),
-        existingKeyId: U256::ZERO,
+        existingKeyId: row
+            .try_get::<Option<[u8; 32]>, _>("existing_key_id")?
+            .map(U256::from_le_bytes)
+            .unwrap_or_default(),
```
dd23, line 392: not saying this is wrong, just want to double-check that fallback to a default value (instead of throwing an error) is what we want if the key id is none, or the map to U256 fails.

eudelins-zama: This is coming from the fact we allow `NULL` in DB for this field, for backward compat of the DB migration.

But we could argue that we could drop this backward compat' (then refuse `NULL`, thus no need to fallback to a default value), because we won't trigger a keygen anyway if something gets wrong during the upgrade process.

eudelins-zama: In that specific case, defaulting to `0` means we default to a fresh keygen, not a migration, which looks fair to me, but is non-intuitive for an external reader.

So I would be in favor of enforcing `NOT NULL` at DB level, and use no default here.

But fine with both approach tbh

dd23: I'm also fine with both, but in if we go with the `default` route, we should document the implications in a comment at least.

eudelins-zama: Fixed a167a175d12ffe9d7305f652ab0be05d6967e5aa

kms-connector/crates/kms-worker/src/core/event_processor/kms.rs
```diff
@@ -108,3 +118,85 @@ const COMPRESSED_XOF_KEY_SET_CONFIG: KeySetConfig = KeySetConfig {
...
+fn keyset_config(existing_key_id: U256) -> KeySetConfig {
+    if existing_key_id.is_zero() {
+        COMPRESSED_XOF_KEY_SET_CONFIG
+    } else {
+        COMPRESSED_MIGRATION_KEY_SET_CONFIG
+    }
+}
+
+fn keyset_added_info(existing_key_id: U256) -> Option<KeySetAddedInfo> {
+    // Fresh keygen has no source key. Migration reads the existing shares and preserves their tag.
```
eudelins-zama: Nit: I don't know what "preserves their tag" means here

eudelins-zama (approved):

PR #3448, opened 2026-08-05
feat(coprocessor): prepare 0.15 key migration

Eikix:
## Context

This is the 0.15 coprocessor prerequisite for [the same-key compressed material migration](https://github.com/zama-ai/tech-spec/pull/478), tracked by [the delivery plan](https://github.com/zama-ai/fhevm-internal/issues/1621).

It is stacked on [the KMS Connector support](https://github.com/zama-ai/fhevm/pull/3447), which is itself stacked on [the host-contract change](https://github.com/zama-ai/fhevm/pull/3446). All three must ship before the migration request.

## Changes

- keep ordinary `ActivateKey` processing when `existingKeyId = 0`;
- stage migration request `keyId` and non-zero `existingKeyId` in the existing activation table;
- wait for both the existing canonical-block finality and the canonical RPC's `finalized` height;
- download `/CompressedXofKeySet/{existingKeyId}`, then verify its signed digest and format;
- update only the existing key's compressed column and retain all legacy material;
- let production Blue CPU workers select legacy material with `FORCE_LEGACY_SERVER_KEY=true`;
- keep unforced CPU workers compressed-first with legacy fallback;
- reject the force-legacy setting on GPU workers, which require compressed material;
- emit one structured post-commit log and retain the activated staging row as durable evidence.

No migration-specific event, staging table, cursor, or host-consumer path is added. Host-consumer parity remains tracked by [the follow-up](https://github.com/zama-ai/fhevm-internal/issues/1873).

The 0.15.1 adoption remains isolated in [the deferred rollout](https://github.com/zama-ai/fhevm/pull/3464).

## Validation

- repository coprocessor format, check, and Clippy pre-commit gates pass;
- `host-listener` compiles with all targets;
- the focused migration download test proves that request ID and existing key ID stay distinct and that digest mismatch fails;
- the database lifecycle test covers pending/finalized gating and verifies that only the compressed field changes;
- Helm lint passes;
- final GPU and stateful rollout evidence remains required on the integrated exact head.


Eikix:
## Automated review — PR #3448

Reviewed against head `5752da8f`, including cross-checks with the stacked #3447 activation lifecycle.

### Verdict

Solid design. The core invariants hold up under scrutiny:

- **No cross-path contamination**: migration rows can't be picked up by `activate_ready_key_activations` (requires `key_content_public`/`key_content_sks_key` both non-null) nor by the ordinary download path (now filtered with `key_material_id IS NULL`).
- **Double finality gating**: `host_chain_blocks_valid.block_status = 'finalized'` **and** `block_number <= rpc_finalized_block`; when the finalized RPC fails, `rpc_finalized_block = None` postpones both download and apply.
- **Integrity before apply**: digest verification plus `prepare_xof_key_set_for_db` deserialization check before a row becomes `ready`, so malformed material can never flip a Green worker ready.
- **Legacy material untouched** by the targeted `UPDATE keys SET compressed_xof_keyset = ...`, and orphaned-block cancellation covers migration rows.
- `.sqlx` offline cache updates are consistent with the new query shapes; CI is green (the only red check, `coprocessor-gpu-tests/setup-instance`, is L40 GPU stock exhaustion — infra, not code).

### Findings

#### Medium

**1. Missing `CompressedKeySet` digest hard-blocks block ingestion** — `host-listener/src/kms_generation/database.rs`, `insert_key_activation_event` (L65-93): for `existingKeyId != 0` events, a missing compressed digest `bail!`s, which propagates through `insert_kms_generation_events_tx` (mapped to `sqlx::Error::Protocol`) and fails ingestion of the entire block — permanently, on every retry, since the event content never changes. The ordinary path tolerates missing digests (stores `NULL`, skips download later). A misconfigured host/KMS event would stall the listener rather than just the migration. Consider logging + skipping (with a failure counter), or at minimum confirming the fail-stop is the intended operational choice. Relatedly, the `sqlx::Error::Protocol` mapping mislabels an application error as a wire-protocol error in logs.

**2. tfhe-worker health check opens a fresh DB pool on every probe** — `tfhe-worker/src/health_check.rs`: unlike the existing `database` check (which short-circuits on a heartbeat), the new `server_key_material` check unconditionally calls `connect_options_for_database_url` + `PgPoolOptions::connect_with` each probe, i.e. a new TCP connection + auth per readiness/liveness probe. sns-worker and zkproof-worker reuse their existing pools; tfhe-worker should do the same (or cache a lazy pool on `HealthCheck`). Under DB pressure this can also flap the probe (5s acquire timeout), which is precisely when you don't want probe-induced churn.

#### Low / worth a comment

**3. Latest-key gates are not chain-scoped** — both `all_pending_compressed_key_materials_to_download`/`apply_ready_compressed_key_materials` gate on `e.key_id = (SELECT key_id FROM keys ORDER BY sequence_number DESC LIMIT 1)` and apply with `MAX(sequence_number)` across all chains. On a DB shared by multiple chains, a newer key on chain B silently stalls chain A's migration forever — no error, no metric. This matches the stack's existing single-chain-per-DB assumption, but a code comment stating that would prevent future confusion.

**4. Catchup semantics changed for all chains, not just KMS ones** — `host-listener/src/cmd/mod.rs`, `consume_catchup_blocks`: the boundary test changed from `>=` to `>`, so the block exactly at `tip - catchup_finalization_in_blocks` is now ingested one step earlier than before. The `unwrap_or(u64::MAX)` startup case is behaviorally equivalent to the old `from_block + catchup_paging` (the paging loop caps the range anyway) — that refactor is fine. The `>=`→`>` widening is safe in practice (reorg/orphan handling plus the migration-specific gates still apply), but it's a silent behavior change to the shared catchup path worth an explicit note in the PR description.

**5. Coverage gap on the poller finality path** — `poller_catches_up_to_safe_tip` now sets `kms_generation_address: None` and drops `KMSGenerationTest`, so `HttpChainClient::finalized_block_number()` and the poller's gating are exercised only via DB-level tests with mocked `rpc_finalized_block` (which are good, e.g. the `None`/`Some(9)` negative cases). An anvil-based test asserting a pending block cannot start a download through the poller path would close the loop, matching the PR's own validation claim.

**6. New readiness dependency on key material** — all three workers now gate readiness on `server_key_material` (required check). On a fresh chain with no activated key yet, workers will report not-ready where they previously passed. That's presumably intentional for the deployment gate, but confirm it doesn't interfere with any bootstrap ordering where a rollout waits on worker readiness.

#### Nits

- `apply_ready_compressed_key_materials` projects `e.key_material_id` in its SELECT, but `PendingCompressedKeyMaterial` has no such field — harmless (FromRow ignores extras) but dead weight.
- `applied_compressed_key_materials()` re-selects all historical applied rows every poll cycle to reset the gauge; fine for now, grows unbounded with migrations.
- Helm exposes `forceLegacyServerKey` as a single global `commonConfig` value applied to all three worker deployments; since GPU builds bail on it, a global `true` would crash-loop GPU pods. The values.yaml comment documents this, but per-worker values (or a chart-side guard) would be more operator-friendly.
- `KMSGenerationTest.sol`: the new event declaration line is long enough to trip typical solhint line-length limits (not enforced here, but keeps style consistent).

### Positives worth calling out

- The DB lifecycle test `compressed_material_updates_only_the_compressed_representation` is well constructed: it asserts the ordinary path ignores migration rows, enforces both finality gates with negative cases (`None`, `Some(9)`), and verifies only `compressed_xof_keyset` changes while legacy `pks_key`/`sks_key` are preserved.
- The `FORCE_LEGACY_SERVER_KEY` test matrix (CPU honors it, GPU rejects it, default keeps compressed-first with legacy fallback) covers the matrix cleanly, and rejection happens at both env-parse and cache-construction time.
- Retaining the `keyId`/`existingKeyId` distinction as `key_material_id`/`key_id` keeps request and protocol identities auditable in the events table.

Eikix:
## Automated review report — `feat(coprocessor): prepare 0.15 key migration`

> Generated by an automated review agent (Claude Code). Findings are ranked by severity and each one cites the code it is derived from. Not a human review; treat every item as a claim to confirm, not a verdict.
>
> **Method and limits.** Full diff read, then claims verified against the branch source (migration columns, `KeyType` variants, helper signatures, all `DbKeyCache` call sites, `HealthStatus` semantics, the block-insert error path). No local build was performed — the tfhe workspace is too heavy for that — so compile/test confidence comes from CI on this head: `coprocessor-cargo-clippy` and every `coprocessor-cargo-test` crate pass (host-listener 33m, tfhe-worker 1h12m, sns-worker, zkproof-worker), but `coprocessor-gpu-tests/setup-instance` **failed**, so every `#[cfg(feature = "gpu")]` path added here is unverified.

The overall shape holds up: a staging row keyed by KMS request ID with `existing_key_id` pointing at the protocol key, a BYTEA→BYTEA copy inside Postgres (no 400 MB wire round-trip), digest verification before staging, legacy material retained, and finality gated on the canonical `finalized` RPC on top of the configured depth. `PendingCompressedKeyMaterial` deliberately excludes the blob, which is the right call.

Each PR-body claim was checked and none were contradicted — including that ordinary activation correctly ignores migration rows via the `key_content_public`/`key_content_sks_key IS NOT NULL` predicates, and that `cancel_orphaned_key_activations` covers migration rows and runs before the apply.

---

### 1. A malformed migration event silently drops an entire block of coprocessor events

`kms_generation/database.rs:66-75` hard-errors when `existingKeyId != 0` and no `CompressedKeySet` digest is present. That error is wrapped into `sqlx::Error::Protocol` at `kms_generation/mod.rs:131`, which fails `ingest_block_logs` → `db_insert_block` retries 10× with reconnects (`cmd/mod.rs:1003-1038`) → `bail!` → the caller runs `if status.is_err() { continue; }` (`cmd/mod.rs:1268-1270`).

The whole block is therefore dropped — ACL grants, FHE computations, CRS events, all of it — and because `last_valid_block` advances on the next successful block, the dropped block is recovered only by chance through `catchup_margin` on a restart.

The pre-existing ordinary path does the opposite: missing digests become `NULL` (`digest_server`/`digest_public` are `Option`), and the download query already filters `key_digest_server IS NOT NULL`, so such a row is simply never picked up.

**Suggested:** make the migration branch degrade the same way — insert with a NULL digest plus an error log and a counter — rather than escalating one bad key event into block-level data loss.

### 2. `bail!` in `apply_ready_compressed_key_materials` aborts the shared transaction

`database.rs:275-285` bails when `rows_affected() == 0`. That transaction (`mod.rs:165-171`) also carries `cancel_orphaned_key_activations`, `cancel_orphaned_crs_activations`, `activate_ready_key_activations`, and — never reached on the error path — `activate_ready_crs_activations`. A single inconsistent migration row thus rolls back every ordinary key and CRS activation for that cycle.

It is reachable: the SELECT and the UPDATE are separate statements, so under READ COMMITTED a concurrent replica committing a new key between them flips `MAX(sequence_number)` to a different `key_id`, yielding 0 rows — precisely the keygen-during-migration overlap this rollout creates.

**Suggested:** log-and-skip, or give the migration apply its own transaction, so the blast radius stays local.

### 3. The "must be the latest key" filter silently strands material

Both `database.rs:617-622` and `database.rs:265-273` require `existing_key_id = (SELECT key_id FROM keys ORDER BY sequence_number DESC LIMIT 1)`. If any key rotation lands after the migration event, the row sits at `pending`/`ready` forever — never downloaded, never applied, never cancelled — and it is invisible to `count_key_activation_remaining_pending`, which counts only `pending`. There is no terminal status and no metric for stranded material, so the deployment gate cannot distinguish "not yet applied" from "will never apply".

Related: the UPDATE additionally constrains `keys.sequence_number = (SELECT MAX(sequence_number) FROM keys)` on top of `keys.key_id = e.existing_key_id`. If `keys` ever holds more than one row for a key_id (a replayed activation in a different block — the conflict target is `(chain_id, block_hash, key_id_gw)`), only the newest row receives compressed material, and `query_db_keys` with `key_id = ANY(...)` can still return the un-patched one. Matching on `key_id` alone would be more obviously correct.

### 4. The readiness gate measures the database, not the worker

`is_server_key_material_available` (`fhevm-engine-common/src/db_keys.rs:36`) reads `keys`, and `set_custom_check(..., true)` feeds `is_healthy()` — i.e. the `/healthz` readiness probe. But both caches key on `(key_id, sequence_number)` (`db_keys.rs:121-141`, `sns-worker/src/keyset.rs:88-113`), and the migration UPDATEs `compressed_xof_keyset` in place, leaving key_id and sequence_number unchanged, with no NOTIFY. A running CPU worker therefore keeps serving the cached legacy keyset indefinitely while the probe reports `server_key_material: ok`.

For a check whose purpose is deciding when cutover is safe, that gap should be stated explicitly, and the "workers must be restarted to pick up applied material" requirement belongs in the PR description at minimum.

Two smaller consequences of the same check:

- on a fresh environment with an empty `keys` table, all three workers report NotReady until the first keygen (`helm --wait` would hang);
- the tfhe-worker now builds its pool and runs a query on every probe even when the heartbeat is fresh, where it previously skipped the DB entirely — one extra permanent connection per pod.

For GPU this is **not** a regression: `deserialize_db_key_row` already bails on GPU without compressed material, so the check merely surfaces a pre-existing hard requirement.

### 5. `.sqlx` offline data was not regenerated

Two orphans remain in `coprocessor/fhevm-engine/.sqlx/`:

- `query-6d9a239a…` — the pre-`existing_key_id IS NULL` version of `all_pending_key_activations_to_download`;
- `query-22dfa7103d…` — `SELECT compressed_xof_keyset, sns_pk FROM keys WHERE sequence_number = $1`, the query this PR deletes.

Harmless at runtime, and no CI job runs `sqlx prepare --check`, but it indicates the new JSON was added without a clean `cargo sqlx prepare`.

### 6. All the new SQL is runtime-checked while its neighbours are macro-checked

`is_server_key_material_available`, `read_server_key_by_sequence_number`, and every new query in `database.rs` use `query` / `query_as::<_, T>` / `query_scalar`, whereas the surrounding ordinary-path queries use `query!` / `query_as!`. The brand-new `existing_key_id` column, `key_content_compressed_xof_keyset`, and the `key_digest_server AS key_digest` → `FromRow` mapping are consequently validated only at runtime. A note on the rationale would help, given the file goes out of its way to keep "a single query shape … cacheable without a CUDA toolchain".

### 7. One extra `finalized` RPC per block, permanently, with no signal when it fails

`cmd/mod.rs:1278-1290` fetches `BlockId::finalized()` on every non-catchup block; `consume_catchup_blocks` (`cmd/mod.rs:515-526`) does it once per catchup page; the poller does the same per tick. That cost is paid forever for a one-shot migration — gating it on "a migration row exists" would confine it.

When the fetch fails, the only signal is a `warn!`. Since `None` postpones the entire migration indefinitely, a counter or gauge the rollout gate can alert on would be worth adding.

### 8. The chart flag is global, and GPU builds fail closed on it

`commonConfig.forceLegacyServerKey` is injected into sns-, tfhe- and zkproof-worker with no per-component override, and a GPU build turns it into a startup failure (`force_legacy_server_key_from_env`). In a mixed CPU/GPU deployment, flipping the single flag takes the GPU pods down. Either make it per-component or state the constraint in the chart.

Separately, `charts/coprocessor/values.yaml:20` bakes environment-specific rollout state ("production Blue") into a distributed chart.

### 9. Unrelated semantic changes bundled in

- `consume_catchup_blocks`: `from_block >= finalized_block` → `>`, plus `unwrap_or(u64::MAX)` for the no-tip case. The `>=`→`>` change makes the guard consistent with `paging_to_block.min(finalized_block)` and looks like a genuine off-by-one fix, but it changes catchup behaviour for every chain and appears in neither the description nor the validation list. The no-tip branch is equivalent for any `catchup_paging > 0`; it only loses the "Unknown top block" log.
- `poller/mod.rs:352`: `(latest, config.finality_lag)` → `(safe_tip, 0)` is a pure no-op, since `update_finalized_blocks_aux` computes `last_block_number - finality_lag` internally, and it degrades that function's `info!(last_block_number, finality_lag, …)` to always logging `0`. Reverting it is suggested.

### 10. Test coverage

The new DB lifecycle test is solid and does pin the K-vs-R distinction, the `pending`/`finalized` gating, and that only `compressed_xof_keyset` changes. Gaps:

- The poller test **dropped** KMS coverage (`kms_generation_address: None`, `sol!` removed) in the very PR that adds the `kms_generation_address.is_some()` branch. If the reason is that Anvil's `finalized` tag does not cooperate, that is worth stating — it is the same tag production depends on.
- Consequently the finalized-RPC gate has no test anywhere: `keygen_integration_tests` injects `Some(i64::MAX)` and the unit test injects `Some(10)`.
- Untested: the missing-digest insert path from §1 (cheap to add, and the highest-severity item), digest mismatch on migration material, and the stranded / `rows_affected == 0` paths.
- GPU CI never ran, leaving `test_force_legacy_is_rejected_by_gpu_workers` and `test_default_rejects_legacy_fallback_on_gpu` unverified.

### 11. Smaller items

- `mod.rs:173` vs `mod.rs:181`: local `applied_compressed_materials` sits next to the function `applied_compressed_key_materials` — near-identical names in one scope; `newly_applied` / `all_applied` would read better.
- The gauge loop full-table-scans activated rows on every block and uses a 64-hex-char `key_digest` label — bounded, but awkward on a dashboard; an info-style metric fits better.
- `retry_count` is incremented but never consulted. A persistently wrong digest re-downloads ~400 MB per block with no backoff. Pre-existing in shape, but the migration path is the one an operator error will actually exercise.
- `prepare_xof_key_set_for_db(&bytes)?` (`mod.rs:304`) decompresses the whole keyset purely to validate, discards the result, and does so synchronously on the async runtime thread inside an open transaction holding `FOR UPDATE` locks. No `spawn_blocking` exists anywhere in host-listener, so it matches local style — but here the work is pure overhead.
- With `FORCE_LEGACY_SERVER_KEY=true` on a key row whose `sks_key` is NULL, the `"server_key_blob!"` override produces an opaque `ColumnDecode` error, whereas `read_server_key_by_sequence_number` gained a clear message for the same condition. Not reachable today (activation requires `key_content_sks_key IS NOT NULL`) — just inconsistent.

### Small hardening suggestion

Adding an explicit `existing_key_id IS NULL` predicate to the two ordinary-path queries (`activate_ready_key_activations` and its `ready` selector) would make the separation between the ordinary and migration lifecycles intentional rather than incidental on the content-NULL predicates.

charts/coprocessor/values.yaml
```diff
@@ -17,6 +17,10 @@
 # -----------------------------------------------------------------------------
 # Shared configuration across all coprocessor components
 commonConfig:
+  # Set only on production Blue before compressed material is published.
+  # GPU workers reject this safeguard because they require compressed material.
+  forceLegacyServerKey: false
```
antoniupop: Not sure whta the best is - maybe set here to true as safety pin and swap in 0.15.1 for GCS. Either way is fine, but this is a dangerous config var to get wrong!

antoniupop (approved):

charts/coprocessor/values.yaml
Eikix: yes true, if a partner sets it wrong, it'll create drift
but I also want to make it simple for everyone not using the prod release to be ok (e.g. GPU devnet, etc.)

i can also reverse the default to true so that unless the tag is 0.15.1, it's impossible to have a footgun

antoniupop: that's what I'd do personally... I know it means more care is required in e2e etc. but it's such a massive footgun, I can see it happening

PR #3452, opened 2026-08-05
feat(host-contracts): add tasks to mirror canonical KMS context/epoch onto replicas

obatirou:
Adds build-calldata and broadcast Hardhat tasks that mirror the canonical ProtocolConfig's active KMS context and epoch onto non-canonical replicas, with tests.


tawadaa (approved):

isaacdecoded (approved):
LGTM! ✅

PR #3458, opened 2026-08-06
fix(coprocessor): replay accepted input proofs

Eikix:
## Summary

Forward-port #3433 so a coprocessor that locally rejects an input proof can replay it after the Gateway accepts the proof at quorum. The port preserves `main`'s blue/green write guards for final-event handling and rejected-proof retention.

## Validation

- `cargo check --locked -p gw-listener -p transaction-sender -p zkproof-worker`
- Gateway quorum tests: 2 passed
- Transaction-sender retention tests: 4 passed
- The zkproof replay tests compile, but local execution is blocked by the same XOF fixture deserialization failure reproduced by unchanged `tests::test_verify_proof` on `main`.

Forward-port of #3433. Refs zama-ai/fhevm-internal#1860.


antoniupop (approved):

Eikix:
@mergifyio queue

Eikix:
@mergify queue

PR #3459, opened 2026-08-06
chore(charts): set polygon mainnet contract addresses

piizama:
Sets the deployed Polygon mainnet contract addresses in the coprocessor and kms-connector chart presets.

- `coprocessor/configs/polygon-mainnet.yaml`: ACL, FHEVMExecutor, InputVerifier, KMSVerifier, HCULimit, ProtocolConfig
- `kms-connector/configs/contracts-mainnet.yaml`: `polygon.acl.address`

Chart versions bumped accordingly.


charts/coprocessor/configs/polygon-mainnet.yaml
```diff
@@ -1,7 +1,8 @@
-acl.address: ""
+acl.address: "0x6737F17e31cf26a1b62fb0362acC5a16CB156F49"
+contracts.version: v0.13.0
 decryption_oracle.address: ""
-fhevm_executor.address: ""
-hcu_limit.address: ""
-input_verifier.address: ""
-kms_verifier.address: ""
-pauser_set.address: ""
```
Eikix, line 7: no pauser set address?

piizama: The pauser_set is not needed in this chart as the coprocessor doesn't consume it.

Eikix (approved):

PR #3464, opened 2026-08-07
test(test-suite): validate 0.14 to 0.15 key migration

Eikix:
## What

Adds the production-shaped RFC 029 migration rollout:

1. boot the coherent stable `v0.14.1` Blue stack and generate a legacy Test key;
2. upgrade the changed contracts, Relayer, each KMS Core/Connector pair, and listener-core;
3. start locally built 0.15 Green with `FORCE_LEGACY_SERVER_KEY=true`;
4. delegate evaluation and cutover to the existing multi-chain Blue/Green e2e flow;
5. publish compressed material for the active key and prove `ServerKey/<keyId>` is unchanged;
6. restart active 0.15 workers and prove they still compute with the legacy representation.

The unchanged 0.15 substrate is frozen to reviewed PR-base SHA `6430345d834d78af47b7bc625d21cee2031b07f7`; feature components are built from this PR checkout. Green-only control-plane services are omitted from the 0.14 Blue lock.

## Why

This proves 0.15 can publish the compressed representation without adopting it. The 0.15.1 worker change and 0.15→0.15.1 Blue/Green rollout remain separate follow-ups.

## Validation

- `bun test` (346 passed)
- `bun run check`
- `git diff --check`
- [exact-head stateful rollout](https://github.com/zama-ai/fhevm/actions/runs/33866895468) passed at `d27b3248b7d7a28d006769a869bbc0284b07a034`, including rollback phases 2–4, retry/cutover, compressed-material agreement, unchanged `ServerKey`, legacy worker restart, and final computation/public/user decryption ([receipt](https://github.com/zama-ai/fhevm/actions/runs/33866895468/artifacts/9936791248))

Specification: https://github.com/zama-ai/tech-spec/pull/478

Tracking issue: https://github.com/zama-ai/fhevm-internal/issues/1621


test-suite/fhevm/rollouts/v0.14-to-v0.15-gpu-key-migration/run.ts
```diff
@@ -0,0 +1,534 @@
...
+          kmsPublicMaterialDigests("ServerKey", keyIdHex),
+          kmsPublicMaterialDigests("CompressedXofKeySet", keyIdHex),
+        ]);
+        return (
+          legacy.every((digest, index) => digest === legacyKmsDigests[index]) &&
+          new Set(compressed).size === 1
+        );
+      } catch {
+        return false;
+      }
```
Eikix: [AUTOMATED] ServerKey digest drift is a terminal RFC 029 violation, but this wait treats it as not-ready for up to 3600s.

Retrying HTTP failures from unpublished `CompressedXofKeySet` is correct. Returning false when `legacy` no longer matches `legacyKmsDigests`, or when every party already serves CompressedXofKeySet but the digests disagree, is not: those are protocol failures, not readiness. The empty `catch` then drops the only actionable messages (`does not serve ServerKey|CompressedXofKeySet/...`, later `assertOperatorMaterialAgreement` blocks).

The same empty-catch pattern is on the connector-boundary waits (~151 and ~448). Cursor-behind is retryable; image mismatch after upgrade is not, and schema-missing should not look like a listener-boundary timeout.

Suggested fix: throw immediately on ServerKey drift and post-publish CompressedXofKeySet disagreement; keep last error on timeout; retry only unpublished material / cursor-behind.

Confidence: 95/100

test-suite/fhevm/rollouts/v0.14-to-v0.15-gpu-key-migration/run.ts
```diff
@@ -0,0 +1,534 @@
...
+    const prefix = operator === 0 ? "coprocessor-" : `coprocessor${operator}-`;
+    for (const worker of KEY_WORKERS) {
+      const container = `${prefix}${role === "Green" ? "gcs-" : ""}${worker}`;
+      await waitUntil(`${container} loaded ${representation}`, async () => {
+        const result = await run(["docker", "logs", "--since", since, container], { allowFailure: true });
+        const logs = `${result.stdout}\n${result.stderr}`;
+        return logs.split("\n").some((line) =>
+          line.includes("server_key_representation") && line.includes(representation)
+        );
+      }, 300);
```
Eikix, line 334: [AUTOMATED] This gate cannot tell a worker that loaded `compressed-xof` from one that has not logged yet.

Retrying until `server_key_representation=legacy` appears after restart is fine. If the same `--since` window already contains `server_key_representation=compressed-xof`, that is a terminal force-legacy failure and should throw immediately. `allowFailure: true` also turns a failed `docker logs` / crashed container into the same 300s timeout as a missing log line.

Suggested fix: fail fast on the opposite representation (and on non-zero `docker logs`); keep retrying only when no `server_key_representation` line exists yet. Matching `server_key_representation=${representation}` is clearer than two independent `includes`.

Confidence: 88/100

Eikix (commented):
[AUTOMATED]

## Code Review

This is the 0.15 publication-stage proof for RFC 029: 0.14 → 0.15 Blue/Green with `FORCE_LEGACY_SERVER_KEY`, paired KMS upgrades, then same-key publication of `CompressedXofKeySet/K` while `ServerKey/K` and workers stay on legacy. The release split, K vs R identities, mixed-connector fail-closed, and post-cutover migration order look right.

Two fail-fast gaps will turn the invariants this rollout exists to catch into generic timeouts. Neither is a false-green on the happy path, but a mutated ServerKey or a worker that loaded compressed-xof will hang and then fail without naming the protocol violation.

0.15.1 adoption, GPU `FORCE_LEGACY` rejection, KMS copy atomicity, and pre-migration ciphertext reuse are correctly out of scope here.

antoniupop (approved):

PR #3474, opened 2026-08-10
chore(coprocessor): tonic 0.14 and yellowstone client v14.2.2

Eikix:
Stacked on #3466 — review that one first. Base retargets to `feature/solana` automatically once #3466 merges.

#3466 moved the geyser plugin the e2e stack runs to `v14.2.2+solana.4.1.0` and left the client in `host-listener` on `v5.0.1+solana.2.1.21`. Nine major versions apart, with a comment promising to fix it shortly. This is that fix.

### Why it is one PR and not two

The plan was to land the tonic upgrade alone, then bump the tag. That does not work. `yellowstone-grpc-proto` re-exports the tonic it was generated against, so:

- v5 tag + tonic 0.14 → our `Interceptor` and `yellowstone_grpc_proto::tonic::service::Interceptor` are different traits with the same name
- v14 tag + tonic 0.12 → same story for `Body`, and `GeyserClient` rejects our `Channel`

Either split leaves `host-listener --features solana-grpc` uncompilable. Verified both directions before combining them.

### What changed

**tonic 0.12.3 → 0.14** — 0.14 moved prost codegen out of `tonic-build` into `tonic-prost-build`, so `fhevm-engine-common/build.rs` calls that instead. `prost` 0.13.5 → 0.14. That is the whole upgrade; nothing else in the workspace touched tonic's API surface.

`tonic-build` ends up with no users at all. `stress-test-generator` listed it under `[dependencies]` rather than `[build-dependencies]` and has no build script, so it was already dead weight — removed along with the workspace entry.

**Client tag → v14.2.2+solana.4.1.0** — matches what `build-yellowstone-plugin.sh` fetches. v14 adds an optional cuckoo filter to the block and account subscriptions: a probabilistic membership set for callers whose account list is too long to send literally, trading false positives for size. We subscribe to one program and two sysvars, so both fields are `None` and the explicit list stays smaller and exact.

The crate also brings its own `solana-pubkey` 4.3.0 next to the 3.0.0 that `solana-sdk` pulls in. Harmless as written — the wire carries base58 strings and raw bytes, and we parse those into our own types — but the Cargo.toml now says so, because the error you get from handing one crate's `Pubkey` to the other names the type and not the version behind it.

### Clippy was not looking at any of this

`cargo clippy -p host-listener --all-targets` never compiled the Solana ingestion path, because it is behind `solana-grpc` / `solana-reconstruct`. The entire yellowstone client and its tests were unlinted, and had already accumulated a `cloned_ref_to_slice_refs` warning from #3200. Added the feature-gated invocation, fixed the warning.

The workflow's path filter also only watched `coprocessor/fhevm-engine/**`, so a change to `zama-host` — which `solana-reconstruct` compiles as a path dependency — could break this lane without triggering the job. Added the three crates it actually builds against.

### Verification

Local, against the v14 proto:

- `cargo check --workspace --all-targets` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo clippy -p host-listener --features solana-grpc,solana-reconstruct --all-targets -- -D warnings` clean
- `cargo test -p host-listener --features solana-grpc,solana-reconstruct --lib` — 123 passed, 0 failed

`solana-tests/host-listener-reconstruct` runs here and covers the decode path.

`solana-e2e` does **not** run on this PR — it only triggers on PRs targeting `feature/solana`, and this one targets #3466's branch. So the wire-level check of a v14 client against a v14 plugin happens after #3466 merges and this retargets. Worth noting that a base change alone may not re-fire the workflow, so that run probably needs a `synchronize` push or a manual dispatch rather than being assumed.

### DIFF
--- a/solana/scripts/sync-zama-host-idl.sh
+++ b/solana/scripts/sync-zama-host-idl.sh
@@ -13,6 +13,13 @@ set -euo pipefail
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 
 cd "$ROOT"
+
+# The goldens this writes are compared byte for byte against a fresh `anchor build` in CI, so
+# minting them under a different Anchor than CI runs produces a diff that looks like a real
+# IDL change and fails the next unrelated PR. Keep in lockstep with solana-tests.yml.
+EXPECTED_SOLANA="${EXPECTED_SOLANA:-4.1.2}"
+. "$(dirname "${BASH_SOURCE[0]}")/lib/require-pinned-toolchain.sh"
+
 NO_DNA=1 anchor build --ignore-keys
 # Writes all four vendored IDLs, including the two demo-only programs whose copies live
 # with the dapp that consumes their generated clients. The list they come from is
--- a/solana/scripts/sync-zama-host-idl.sh
+++ b/solana/scripts/sync-zama-host-idl.sh
@@ -18,7 +18,7 @@ cd "$ROOT"
 # minting them under a different Anchor than CI runs produces a diff that looks like a real
 # IDL change and fails the next unrelated PR. Keep in lockstep with solana-tests.yml.
 EXPECTED_SOLANA="${EXPECTED_SOLANA:-4.1.2}"
-. "$(dirname "${BASH_SOURCE[0]}")/lib/require-pinned-toolchain.sh"
+. "$ROOT/scripts/lib/require-pinned-toolchain.sh"
 
 NO_DNA=1 anchor build --ignore-keys
 # Writes all four vendored IDLs, including the two demo-only programs whose copies live
--- a/solana/scripts/update-cost-snapshots.sh
+++ b/solana/scripts/update-cost-snapshots.sh
@@ -45,43 +45,7 @@ done
 
 # Keep in lockstep with .github/workflows/solana-tests.yml SOLANA_VERSION.
 EXPECTED_SOLANA="${EXPECTED_SOLANA:-4.1.2}"
-EXPECTED_ANCHOR="$(awk -F'"' '/^anchor_version/ { print $2; exit }' Anchor.toml)"
-if [[ -z "$EXPECTED_ANCHOR" ]]; then
-  echo "error: could not read anchor_version from Anchor.toml" >&2
-  exit 1
-fi
-
-require_cmd() {
-  command -v "$1" >/dev/null || {
-    echo "error: missing required command: $1" >&2
-    exit 1
-  }
-}
-
-require_cmd solana
-require_cmd anchor
-require_cmd cargo
-
-solana_ver="$(solana --version)"
-anchor_ver="$(anchor --version)"
-
-# Require an exact version token after `solana-cli` (space or end), so e.g.
-# 2.1.05 cannot satisfy EXPECTED_SOLANA=2.1.0.
-case "$solana_ver" in
-  "solana-cli ${EXPECTED_SOLANA}"|"solana-cli ${EXPECTED_SOLANA} "*) ;;
-  *)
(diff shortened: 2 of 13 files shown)


Eikix (commented):
[AUTOMATED]

## Code Review

Solid dependency alignment: tonic 0.14 + yellowstone client v14 matching the geyser plugin tag is the right atomic move, and the proto delta is additive so the subscribe path with `cuckoo_*: None` stays wire-equivalent. CI clippy coverage for `solana-grpc,solana-reconstruct` and the IDL-spec pin are good hardening. `solana-e2e/full-vertical-reconstruct` already ran green on this head after the base retargeted to `feature/solana`.

One validated regression in the toolchain-helper extraction (inline comments below). Easy one-line fix per caller: source via `$ROOT/scripts/lib/...` after the `cd`.

No other high-signal bugs, guideline violations, or blocking protocol-compatibility issues found.

solana/scripts/sync-zama-host-idl.sh
```diff
@@ -13,6 +13,13 @@ set -euo pipefail
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 
 cd "$ROOT"
+
+# The goldens this writes are compared byte for byte against a fresh `anchor build` in CI, so
+# minting them under a different Anchor than CI runs produces a diff that looks like a real
+# IDL change and fails the next unrelated PR. Keep in lockstep with solana-tests.yml.
+EXPECTED_SOLANA="${EXPECTED_SOLANA:-4.1.2}"
+. "$(dirname "${BASH_SOURCE[0]}")/lib/require-pinned-toolchain.sh"
```
Eikix, line 21: [AUTOMATED] After `cd "$ROOT"`, this relative `dirname "${BASH_SOURCE[0]}"` is resolved against `solana/`, so any relative invocation from outside that directory fails immediately:

```text
$ bash solana/scripts/sync-zama-host-idl.sh   # from repo root
...: line 21: solana/scripts/lib/require-pinned-toolchain.sh: No such file or directory
```

That is exactly the remediation string `check_solana_abi.py` prints on IDL drift. Pre-PR both scripts were cwd-safe; this is a regression from the extraction. Absolute-path and `from solana/` invocations still work, and it fails loudly before any write — but please anchor on the already-absolute `ROOT`:

```bash
. "$ROOT/scripts/lib/require-pinned-toolchain.sh"
```

Same fix needed in `update-cost-snapshots.sh` (and the usage snippet in `lib/require-pinned-toolchain.sh`).

Confidence: 97/100

solana/scripts/update-cost-snapshots.sh
```diff
@@ -45,43 +45,7 @@ done
...
-    ;;
-esac
-if [[ "$anchor_ver" != "anchor-cli ${EXPECTED_ANCHOR}" ]]; then
-  echo "error: need Anchor ${EXPECTED_ANCHOR} (got: ${anchor_ver})" >&2
-  echo "       match CI / Anchor.toml anchor_version" >&2
-  exit 1
-fi
-
-echo "toolchain ok: ${solana_ver}; ${anchor_ver}"
+. "$(dirname "${BASH_SOURCE[0]}")/lib/require-pinned-toolchain.sh"
```
Eikix, line 48: [AUTOMATED] Same cwd-relative source bug as in `sync-zama-host-idl.sh`: this runs after `cd "$ROOT"`, so `bash solana/scripts/update-cost-snapshots.sh` from the repo root dies with `No such file or directory`. Pre-PR the guard was inline and cwd-safe. This file already absolutizes `SCRIPT_PATH` a few lines above — either reuse that or source via `$ROOT`:

```bash
. "$ROOT/scripts/lib/require-pinned-toolchain.sh"
```

Confidence: 97/100

PR #1539, opened 2025-12-10
chore(kms-connector): retry DB insertion

eudelins-zama:
Ref https://github.com/zama-ai/fhevm/pull/1531

PR #3477, opened 2026-08-10
refactor(solana): two-tier test framework — test-kit, specimen apps, bun:test e2e

Eikix:
Executes the plan in [fhevm-internal#1876](https://github.com/zama-ai/fhevm-internal/issues/1876): retire the duplicated Mollusk preambles, the Rust live-client, and the bash e2e layer in favor of a two-tier framework — a `zama-solana-test-kit` crate for in-process SVM tests and a TypeScript rim (SDK + bun:test scenarios) for the running stack.

Landing milestone by milestone on this branch; each milestone has its own gate (see the issue's retirement ledger).

## Milestones

- [x] **M1 — test-kit crate.** Extract the shared harness vocabulary (Mollusk boot, anchor builders, fixture accounts, cleartext oracle + `CleartextLedger`, KMS/coprocessor signing, cost snapshots, operator contract catalog) from `tests/support/` + per-suite preambles into `solana/test-kit`. Gate: cost snapshots byte-identical, kit smaller than the 2,843 lines it replaces (1,634 from the 8 suites + 1,209 from `tests/support/`); lands at 1,957, a ~31% net reduction.
- [x] **M2 — one `ExecutionDictionary`.** `encode` feature on `zama-host`; both callers compile against it before either copy is deleted.
- [x] **M3 — encrypted-counter specimen.** ~150-line program proving the kit onboards a new app in ~60 lines of test.
- [x] **M4 — TS rim.** `stack.ts` + bun:test scenarios replace live-client and the bash phases; `solana-e2e` goes green with zero grep assertions. Assertion inventory below.
- [x] **M5 — tail.** Dep-chain specimen, `solana_vertical.rs` wired into CI (then `solana_poc.rs` deleted per the ledger gate), TESTING.md rewrite, literal connector byte vectors, final whole-PR review. LiteSVM as a second backend is consciously blocked — see the M5 notes.

## M1 notes for review

- Test bodies are untouched — only preambles/imports changed. The suites keep thin wrappers (e.g. each suite's `evaluate_fhe_cpis` ext trait) so diffs stay reviewable.
- `snapshot.rs` switched from the `env!` macro to runtime `CARGO_MANIFEST_DIR` lookup: compiled into the kit, the macro would name the kit's own dir; the runtime variable names the crate whose test target runs, where the committed snapshots live.
- The operator catalog macros deliberately use `crate::run_*` paths that resolve in the invoking crate (`#[allow(clippy::crate_in_macro_def)]`): the catalog is shared, the runners are not.
- Snapshot JSON files are untouched; CI's snapshot job is the byte-identity gate.

## M4 notes for review

- The scenario suite (`test-suite/fhevm/e2e/scenarios/`) now carries the whole live vertical: `fhe-vertical` (trivial-encrypt → public + pure-SDK user decrypt, historical decrypt of an updated-away handle, the #1539 verified-input flow), `operators` (the 8-op live sweep), `confidential-transfer`, `token-vertical` (wrap → attested burn → seal → certified public decrypt with the leaf_count=2/leaf_index=0 lifecycle-batch proof → redeem with an SPL balance-delta assertion the bash never made → disclose), and the `adversarial-l4` context-mismatch tail.
- The side-stack setup is typed: `src/solana/{addresses,validator,deploy}.ts` absorb `setup-solana-side.sh` (gateway reads in-process instead of `cast`, the validator lifecycle with its agave-4.x/geyser operational knowledge moved along as comments, the zama-host bootstrap through the generated Codama client, host-chain registration, host-listener). `clean-e2e.sh` hands off to `bun run src/solana/deploy.ts`.
- Deleted: `full-vertical.sh`, `adversarial-l4.sh`, `check-e2e-greps.py` (nothing greps stdout labels any more), the 2,861-line live-client crate, `solana-input.ts`.
- **Status-check rename:** the solana-e2e job's check is now `solana-e2e/scenarios-reconstruct` (was `solana-e2e/full-vertical-reconstruct`). Anything requiring the old name (Mergify queue rules, branch protection) needs updating.
- L4-a (publicKey-substitution user-decrypt) is *not* ported: the attack lives in the kms repo's live harness (`cargo test -p kms --test solana_user_decrypt_live`, `SOLANA_UD_ATTACK=pubkey_substitution`), which signs a valid v3 request and swaps the ML-KEM key after signing. This repo only ever launched it; that stays its vehicle. L4-b (context-mismatch certificate) is in `token-vertical.scenario.test.ts`, in CI for the first time — and the rejection is pinned to `InvalidKmsContext` (anchor 6064) via the preflight logs, so an unrelated failure cannot impersonate the repelled attack.
- The milestone review (two adversarial passes: correctness/coverage + design/leanness) drove a hardening batch: the SNS-commit wait restored to the bash's most generous 240s budget, the proof-service restart gate re-homed (see the inventory row below), the job timeout back at 120, `test-suite/fhevm/src/*` excluded from the "rebuild everything" bucket of `select-overrides.sh`, the zama-host codegen drift guard wired into the `demo-dapp` paths filter, the side-stack setup failing fast on gateway reads (with the resolved signer sets logged), `SOLANA_DEPLOYER_KEYPAIR` unified with the demo deployer, the ledger ownership check on `lstat`, and a leanness pass (shared SDK-encrypt seam, one KMS-context-id constant, generated-PDA reuse, `certificate` vocabulary, `demo/tokenAccounts.ts` → `src/solana/spl.ts`).

## M5 notes for review

- **dep-chain specimen** (`solana/programs/dep-chain`): the load shape next to encrypted-counter's minimal shape — one `extend(links, amount)` packs a strictly dependent add chain into a single `fhe_execute` (each step reads the previous step's transient result; only the tail persists). Its depth is `MAX_CHAIN_LINKS = 16` = `zama_fhe::MAX_ON_CHAIN_EXECUTION_STEPS`, pinned by a compile-time assert: the first CI run failed with `TooManyStepsForDefaultHeap` at 32, which is the builder's documented point — a CPI-composing program on Anchor's default 32 KB heap cannot reach the host's 32-step ceiling; only off-chain builders can. The Mollusk twin replays full-depth (16), partial (4), and single-link chains through the kit's cleartext oracle and pins both out-of-range lengths (0, 17) to `InvalidChainLength`. The live half is `load-smoke.scenario.test.ts`: the host's full 32-step dependent chain through the raw typed client (off-chain builder, so the host ceiling applies — and no extra program deployment is needed), with an unrelated value released concurrently so a wedged slow lane fails the test.
- **`solana_poc.rs` → `solana_vertical.rs`**: re-homed and finally in CI — solana-e2e's new `worker-vertical` job anchor-builds the two programs and runs the previously manual-only test (LiteSVM transfer → off-chain reconstruction → real TFHE worker on a disposable Postgres → decrypt). Per the retirement ledger's gate, `solana_poc.rs` is deleted only after the job is green. Wiring it up immediately caught two cases of rot accumulated while the test was `#[ignore]`d: the transfer batch grew a sixth step (the sender re-encrypt from fhevm-internal#1853) that the old five-step assertion predated, and the plaintext seeding assumed Alice and Bob start on distinct handles when account init's `trivial_encrypt_u64(0)` gives both accounts the *same* deterministic handle — the second seed silently overwrote the first and the whole chain computed on the wrong balance. Both are pinned now (six-step assert; the three persistent steps pinned to their on-chain result handles and the transient select pinned by dataflow into the two steps that consume it; shared-handle assert + single seed) — exactly the drift the milestone exists to make visible.
- **Final whole-PR review batch** (two adversarial passes: correctness/coverage + design/leanness). The two real gaps found: the test-suite/fhevm typecheck + offline unit suite ran only behind the `e2e` PR label, and the SDK-side byte-vector consumer only ran on PRs against main/devex — both now have change-gated jobs in solana-tests.yml (`test-suite-fhevm`, `sdk-solana-vectors`). The leanness pass: `addresses.ts` reads the gateway through viem (the issue's stated shape) instead of hand-pinned selectors + a hand ABI decoder; the adversarial context-mismatch rejection is asserted as a typed kit instruction-error code (6064) instead of a log regex; the seven copied lazy vault/SDK loaders collapsed into `lazy-modules.ts`; `until()` moved to `src/utils` and now also backs the SNS-commit wait and the zkproof-worker restart gate (which previously fell through silently on timeout); `releaseAndExpect`/`publicDecryptExpect` no longer take an `expect:` parameter — the scenarios' own `expect()` calls are the single assertion (`releaseAndDecrypt`/`certifiedPublicDecrypt`); the historical user-decrypt now cross-checks the access-proof leaf count against on-chain state like its public sibling; dead surface dropped (`internScalar`, an unused committed `encrypted_counter` keypair + its Anchor.toml row, a one-field thresholds struct); TESTING.md/TESTING_STRATEGY.md/README corrected (stale `full-vertical.sh` rung, wrong `verticalSetup` memoization claim, the three unlisted Mollusk suites, the missing dep-chain row). Consciously declined with reasons: hand-built SPL instruction encoders stay (house convention — no `@solana-program/*` anywhere in the repo; layout-unit-tested), `zama_host::encode` stays a feature on the program crate (the encoder needs the crate's own types), the worker-vertical job stays ungated (solana-e2e has never been change-gated; the scenarios job costs more), and the byte vectors deliberately don't duplicate the v3 wire-key/client-id parity tests that already pin those.
- **LiteSVM as the kit's second backend: consciously blocked.** Two hard walls, both recorded in `solana/runtime-tests/Cargo.toml`: every published litesvm needs `solana-program-runtime >= 4.1`, which cannot coexist with mollusk 0.12.1-agave-4.0's `=4.0.0-rc.0` pin in one lockfile; and mollusk 0.15.0 (which resolves the wincode split) trips E0658 (`maybe_uninit_write_slice`) on the pinned rustc 1.91.1. No hollow one-variant `FhevmSvm` enum was added — that would be dead surface. Retry when upstream lands on one wincode line and the toolchain moves past 1.91.

## Live-client assertion inventory (M4 gate)

Sources: `live-client/main.rs` (2,835 lines; prints, the bash asserts), `full-vertical.sh` (543), `adversarial-l4.sh` (167). Every row is **re-homed** in a bun:test scenario, **re-homed** in the Rust tier, or **consciously dropped** with a reason.

Legend: [S] = re-homed to a bun:test scenario · [R] = covered in the Rust tier · [D] = dropped, with reason · [I] = infrastructure (not an assertion; absorbed by stack.ts / src/solana)

### Phase 0 — bootstrap + input proof
| Assertion | Verdict |
|---|---|
| BOOTSTRAP: HostConfig + live KMS context initialized from gateway env | [I] absorbed into `deploy.ts` typed bootstrap (Codama zama_host client) |
| input-proof submission returns a handle (via solana-input.ts) | [S] scenarios assert the typed handle from `@fhevm/sdk` submitInputProof |
| proof-service restart gate: readiness -> `docker restart` -> post-restart exact-inclusive replay (#1682/#3215) | [S] `stack.restartProofService()` at the top of `fhe-vertical.scenario.test.ts` — every proof the suite consumes afterwards is served by a restarted, ledger-replayed service (missed in the first inventory; restored after the milestone review) |

### Phase 1 — trivial encrypt via fhe_execute
| Assertion | Verdict |
|---|---|
| `result handle` / `output ACL record` / `encrypted value id` stdout extraction greps | [D] extraction, not assertion — typed client return values replace stdout scraping |
| `allow_for_decryption` grep | [S] typed: the scenario performs the public-decrypt release explicitly |
| SNS commit poll (ciphertext_digest row materializes) | [I] `stack.ts` `waitForSnsCommit(handle)` |
| public-decrypt cleartext == VALUE | [S] `fhe-vertical.scenario.test.ts` via SDK publicDecrypt |
| client-side `authorize_public` proof verification against live on-chain peaks | [S] `certifiedPublicDecrypt` calls the SDK's `verifyPublicDecryptProof` against `state.peaks`/`state.leafCount` before the KMS request (omitted from the first inventory; restored after review — leaf-count equality alone only proves the service read the same account) |
| pure-SDK user-decrypt == VALUE | [S] same scenario (was already SDK-driven inside the bash) |

### Phase 2 — historical update
| Assertion | Verdict |
|---|---|
| HIST H_old / mmrProofBytes extraction | [D] extraction — typed proof object from the proof-service client |
| old-handle SNS commit poll | [I] same stack helper |
| proof pins the ORIGINAL handle after the update (`H_OLD2 == H_OLD`) | [D] **no direct successor**: nothing compares two handles for equality. The substance survives elsewhere — the host's `previousState` validation fails the update on a wrong read, and the scenario adds a `.not.toBe()` the bash never had — but the handle-equality comparison itself is gone, and the earlier `[S]` row overstated it |
| `aclValueKey` unchanged across the update | [S] same scenario (encryptedValueId stability assertion) |
| client-side `historical_access_leaf_commitment` + `mmr_verify` against live peaks | [S] `historicalUserDecryptExpect` calls the SDK's `verifyHistoricalAccessProof` before the request (omitted from the first inventory; restored after review — without it a wrong-but-self-consistent leaf index surfaces only as a generic KMS rejection) |
| historical pure-SDK user-decrypt of the OLD handle == old cleartext | [S] same scenario — the semantic point of the phase |

### Phase 3 — verified-input add (#1539)
| Assertion | Verdict |
|---|---|
| result handle/ACL extraction | [D] extraction |
| decrypt(result) == V + TE_ADD (attested input + scalar in one fhe_execute) | [S] input-flow test in `fhe-vertical.scenario.test.ts` |

### Phase 4 — operator sweep (Sub, Mul·scalar, Cast, IfThenElse, RandBounded, Sum, IsIn, MulDiv)
| Assertion | Verdict |
|---|---|
| per-op `allow_for_decryption` grep + handle/ACL extraction | [D] extraction |
| per-op decrypted cleartext == expected (70, 42, 42, 42, <128, 30, 1, 14) | [S] `operators.scenario.test.ts` table-drives the same 8 ops — cleartext math is exhaustively covered by `operator_conformance`/`operator_mollusk_conformance` in the Rust tier; the live rows prove the coprocessor+SNS+decrypt pipeline computes them |
| in-binary expected-handle recomputation (keccak chain recomputed locally) | [R] `operator_mollusk_conformance` pins the exact handle math against a fixed previous bank hash; the live tier asserts decrypted cleartexts instead |

### Phase 5 — token vertical
| Assertion | Verdict |
|---|---|
| spl-token create/mint CLI greps | [I] `provision.ts` `createSplMint`/`mintSplTo` (hand-built SPL instructions, layout-unit-tested) |
| `confidential mint` + `compute_signer` stdout extraction | [D] extraction — `createConfidentialMint` returns typed addresses |
| `OK wrap_usdc` | [S] `token-vertical.scenario.test.ts` (wrap throws on failure; balance consumed downstream) |
| burn input-proof handle non-empty | [S] same scenario |
| `burned amount ACL` / `burned handle` extraction | [D] extraction — typed reads from the derived encrypted value account |
| burned-handle SNS commit poll | [I] stack helper |
| `OK make_handle_public` | [S] seal step throws on failure |
| public-decrypt leaf_count==2 / leaf_index==0 (lifecycle-batch double-append; earliest-leaf resolution) | [S] kept verbatim as typed `expectedLeafCount`/`expectedLeafIndex` assertions — the inventory's one subtle row, documented in the scenario |
| burned cleartext + KMS cert field extraction from decrypt JSON | [D] extraction — the typed claim carries them |
| `OK redeem_burned_amount` (act-once PendingBurn consumed, vault released) | [S] redeem, then SPL balance delta == certified cleartext (a stronger assertion than the bash made) |
| `OK disclose_secp` | [S] disclose step throws on failure |

### adversarial-l4.sh (manual-only before → in CI now)
| Assertion | Verdict |
|---|---|
| (a) publicKey-substitution user-decrypt rejected | [D→kms repo] the attack harness lives in the kms repo (`solana_user_decrypt_live`, ATTACK mode) and signs/swaps inside the SDK internals; this repo only launched it. Consciously left with its owner |
| (b) context-mismatched KMS cert must NOT disclose (InvalidKmsContext) | [S] `token-vertical.scenario.test.ts` adversarial tail: genuine cert, extra_data naming context 2 vs the live context-1 account → on-chain rejection |

### Client-internal checks worth naming
| Assertion | Verdict |
|---|---|
| `fetch_encrypted_value` (account fetch + deserialize) | [D] performs no comparisons — it did `get_account` + `try_deserialize` and returned. Nothing to re-home; the earlier row described comparisons this function never made. Address derivation is client-side in the typed client, and the host validates on-chain |
| `is_proof_service_lagging` unit tests | [S] re-homed as unit tests: `src/solana/fhe-vertical.test.ts` pins the retry classification directly (only 503+`lagging` retries; `corrupt_cache`, 404 and 500 are terminal and not retried), plus the sibling cap and the 32-byte sibling check. The live scenarios exercise the happy path on top |
| TOKEN_BALANCE_STATE JSON probe | [I]+[S] `provision.ts` `readTokenBalanceState`: token account program-owned; EncryptedValue matches canonical derivation (domain==mint, authority==tokenAccount, label re-derives the id); subjects == [owner, compute_signer]; handle nonzero version-0 euint64; HostConfig discriminator + chain-id checks. **Consciously simplified**: the Rust probe's `get_multiple_accounts` coherent re-read guarded a balance-pointer race the TS probe cannot have (every address derived client-side); exact-`SPACE` and bump-equality checks dropped (Anchor-internal invariants; structural MMR checks retained) |
| INITIALIZE_TOKEN_ACCOUNT setup seam | [I] `provision.ts` `initializeConfidentialTokenAccount` (get-or-create, mirroring the Rust seam) |

Counts, recounted from the table above after review: **20 [S]** scenario assertions (one of which is also [I]), **1 [R]** Rust-tier, **10 [D]** dropped with reasons (including the one handed to the kms repo), **7 [I]** infrastructure absorbed into stack.ts / src/solana. The first pass published 16/2/9/7: it under-counted [S], counted the `is_proof_service_lagging` row as Rust-tier when it is now a TS unit test, listed the `H_OLD2 == H_OLD` row as re-homed when it has no successor, and omitted the two client-side proof verifications entirely.


Eikix:
## Multi-agent review — four reviewers, eight lanes

Reviewed by four agents against a shared brief pinning the merge base (`3e3a1e77`), the eight load-bearing claims in the PR description, and the fifteen conscious decisions it records. Each reviewer was told the deleted code is the spec, and that the failure mode to hunt is **not "the new code is wrong" but "the new code is weaker and CI is still green."**

Lanes: (A) independently re-derive the assertion inventory from the 3,571 deleted lines of live-client + bash, then mutation-test every new scenario · (B) Rust refactor equivalence + production blast radius · (C) the bash→TS lifecycle port + docs-vs-tree · (D) CI wiring + generated-code reproducibility.

Generated Codama clients, IDL JSON and lockfiles (5,754 of the added lines) were deliberately not read line by line — they were verified reproducible instead.

**No BLOCKING findings.** The refactor conserves coverage: the assertion inventory is substantially accurate, test bodies really are untouched, cost snapshots really are byte-identical, and several ported probes came out *stricter* than the bash they replace. The findings below are worth-fixing items, not merge blockers.

---

### Verified and held

Checked mechanically, not taken on trust:

| Claim | Verdict | How |
|---|---|---|
| C1 "test bodies untouched" | **holds** | Every `#[test]` body extracted at both refs and text-diffed across 9 suites: 0 added/removed/renamed/`#[ignore]`d. Only mechanical helper renames, each verified behaviour-equivalent against its deleted original. |
| C2 cost snapshots byte-identical | **holds** | `git diff --stat 3e3a1e77..HEAD -- '*cost-snapshots*'` empty. |
| C4 zero grep assertions | **holds** | No executable reference to any deleted artifact anywhere; remaining mentions are historical prose. |
| C5 inventory complete | **holds in substance** | Independent re-derivation found no coverage hole. Four bookkeeping corrections below. |
| C6 `solana_vertical` in CI | **holds** | `worker-vertical` green on this PR at 16m57s, guarded by `grep "test result: ok\. 1 passed"` under `pipefail` — a rename yields `0 passed` and fails. |
| C7 error 6064 pinned | **holds** | Derived three independent ways (IDL index, enum variant count, Mollusk `host_error`). Guard removal → `InvalidKmsCertificate`, a *different* code → test fails. The bash it replaces was a non-gating `echo`. |
| C8 `encode` never in SBF | **holds** | Empirically, under both `-p zama-host` and anchor's `--manifest-path` mode: `feature="default"` only. |
| Assertion census | **holds** | Per-test counts identical across all 9 suites. The three file-level drops are exactly the helper assertions that moved into the kit. |
| `CleartextLedger` CPI invariants | **holds** | The likeliest silent-weakening site. Both call sites re-assert what their deleted private copies did. |
| Dedup is real | **holds** | 6 `mollusk()` → 1, 7 `anchor_ix` → 1, 2 attestation minters → 1. The kit is not copy #6. |
| Oracle not laundered | **holds** | `u64_at` reads the *on-chain* handle against independently-evaluated cleartext; a mismatch panics rather than passing. |
| M2 encoder equivalence | **holds** | Byte-identical `intern`/`intern_key`/`intern_subjects` against both deleted copies. No change in step encoding, operand ordering, or output layout. |
| Timeout-as-pass class | **closed** | All 6 `until()` sites + 3 hand-rolled loops throw on exhaustion. The zkproof-worker gate that previously fell through silently is genuinely fixed. |
| Operational knowledge landed | **holds** | agave-4.x `--bind-address`, the sbpf feature flag, the two-log death report and the surfpool/sysvar rejection are all attached to the code they protect, two of them additionally pinned by unit assertions. `isOwnedDirectory` uses `lstat` with an explicit no-follow note. |
| Generated tree unmodified | **holds** | `codegen:solana:check` exit 0 — a full recursive snapshot comparison, so an extra or missing file trips it too. |
| Byte vectors independent | **holds** | Consumed by *two* implementations (Rust connector + TS SDK). All three `extra_data_v1` blobs hand-verified byte-for-byte; three malformed records pin strict-parser rejection. Not self-fulfilling. |

---

### Findings, ranked by consequence

**1. The newly-CI-wired rot detector is blind in the dimension it exists to cover.**
`coprocessor/fhevm-engine/tfhe-worker/src/tests/solana_vertical.rs:70-78`

The shared-handle seeding fix is correct, but it leaves sender and recipient on an *identical* balance operand (same handle, same ciphertext, seeded once at 125). Mutation: in the transfer builder, swap the debit `sub`'s lhs from `from_balance_value` to `to_balance_value` — one operand slot. Reconstructed operands are byte-identical, every downstream handle is unchanged, the test still decrypts 25 / 225, CI stays green. The batch-shape assertions pin the *select* dataflow (`events[3].rhs`, `events[4].lhs`) but never the operands of steps 0 (`ge`) and 1 (`debit sub`).

The in-code comment — *"Every step's output differs from every input, so a misresolved operand anywhere in the batch shows up here"* — is not accurate as written: `transferred` = 125−100 = 100 also equals the `amount` input, and the two balance inputs are the same handle by construction. Fix: run one throwaway transfer, or seed a second diverged handle, so the two accounts hold different values before the asserted transfer. Cheap, and it restores exactly the property the comment claims.

**2. `bun test` exits 0 when every matched test is skipped — the `demo:smoke` gate is one rename from vacuous.**
`test-suite/fhevm/package.json:12`, `.github/workflows/solana-e2e.yml:228`

Measured on bun 1.3.6: `bun test e2e/scenarios/deposit-arc.scenario.test.ts` with no env prints `0 pass, 1 skip` and exits **0**. Rename `RUN_DEMO_SCENARIOS` in either the script or `deposit-arc.scenario.test.ts:188` and the entire deposit arc stops running with CI green. The glob-vacuity vector is closed by bun itself (a bad path exits 1) — the *skip* vector is not, and it applies to any future `describe.skipIf` on the five unconditional scenarios.

This lesson was already applied to `cargo test` (worker-vertical's trailing `grep "1 passed"`) but to no `bun test` invocation. Suggested shape: assert a nonzero pass count, or make the absence of `RUN_DEMO_SCENARIOS` throw rather than skip.

**3. The typecheck job does not fire on its own upstream inputs.** *(found independently by two reviewers)*
`.github/workflows/solana-tests.yml:104-106`

`test-suite-fhevm` is the PR's fix for "nothing ran `tsc` on an ordinary PR" — good catch, and it does close that hole. But it's gated on `test-suite/fhevm/**` only, while its `tsc` resolves through `@sdk-src/*` and `@demo-dapp/*` into two trees the filter doesn't name. The sibling `demo-dapp` filter *does* list `sdk/js-sdk/**` for exactly this reason; the asymmetry is what makes it look unintended.

Mutation: rename an exported type in `sdk/js-sdk/src/solana/`, in a PR touching only `sdk/js-sdk/`. `changes-test-suite-fhevm` is false, no typecheck runs, bun strips types — so it surfaces as a runtime failure ~40 min into the e2e job, and on a PR targeting `main` it doesn't surface at all.

**4. `stack.ts`'s readiness probes lost the per-attempt request timeout.**
`test-suite/fhevm/e2e/harness/solana/stack.ts:32,50,62` (also `src/solana/fhe-vertical.ts:219`)

Every bash probe used `curl -m2`/`-m3`. The TS ports pass no `signal`. A service that accepts the TCP connection but never responds hangs *inside* `fetch()`, so `until()`'s deadline is never re-evaluated and the poll cannot time out — the scenario dies on bun's per-test timeout with `test timed out` instead of `until(relayer liveness) timed out`. `validator.ts:188` kept `AbortSignal.timeout(2_000)`, so the pattern is already in the codebase and simply wasn't carried across.

**5. The historical access-proof client is a new untested fork.** *(qualifies decision D12)*
`test-suite/fhevm/src/solana/fhe-vertical.ts:209-248`

`fetchHistoricalAccessProof` forks the vault module's proof-service client (retry policy, transport-blob encoder, sibling validation) with no unit test, while the client it forks has deterministic tests pinning exactly the property the deleted `is_proof_service_lagging` unit tests pinned. Mutation: delete `&& body?.status === "lagging"` from line 242 — a `503` with `status: "corrupt_cache"` is then retried 10× before failing, i.e. an integrity failure is misclassified as transient. The stated successor ("exercised live by every scenario decrypt") only fires when the service actually lags, which never happens on a warm stack.

Secondary: `blob[0] = 0x01` is a bare literal where `MMR_PROOF_MODE_HISTORICAL` is exported; the fork omits the `MAX_MMR_SIBLINGS` cap the vault client enforces; and the retry budget silently shrank from the Rust client's 15×2s (~28s) to 10×1s (~10s) — deliberate convention alignment with the vault default, but it halves a budget the deleted code had tuned.

**6. `COPROCESSOR_DB_CONTAINER` is dead configuration.** *(found independently by two reviewers)*
`e2e/harness/loadEnv.ts:56,86` → `src/solana/sns.ts:24`

`resolveEnv` reads, validates and documents the override; `waitForSnsCommit` imports the hardcoded layout constant and ignores it. Setting it means every SNS wait `docker exec`s the wrong container, `allowFailure` swallows the non-zero exit, and every scenario dies after 240s with "ciphertext materialization timed out" rather than "wrong container".

**7. Two latent items in the Rust tier.**

- `solana/test-kit/src/lib.rs:385-396` — `HostConfigParams::new` defaults `coprocessor_signers` to the **live** key's EVM address, where `host_mollusk` and `operator_mollusk_conformance` previously pinned a dead `[0x11; 20]`. Inert today (neither suite exercises an external-input path), but it means a future `fromExternal` test would find the fixture already trusting the kit's signing key, where the dead signer would have forced a visible opt-in.
- `solana/Cargo.toml:2` + `runtime-tests/Cargo.toml:32` — `cargo build --workspace` *does* feature-unify `encode` onto zama-host (`--crate-name zama_host … feature="cpi" feature="encode" feature="no-entrypoint"`). Harmless under every invocation actually used, and pre-existing for `cpi`/`no-entrypoint` — but the comment "Never enabled for the SBF build" is true only because nobody runs a workspace-wide SBF build. If one is ever added it would ship a `.so` with `no-entrypoint`, which is the more serious half.

**8. PR-description accuracy** (bookkeeping — no coverage impact, but these are the numbers reviewers reconcile against):

- Test-kit sizing: measured **1,957** lines (claimed 1,928), replacing **2,843** (1,634 from the 8 suites + 1,209 from `tests/support/`), not 4,676. The real result is still a ~31% net reduction — the headline just doesn't match the diff.
- The inventory's `[S]` count is **17**, not 16 (`[D]`=9, `[I]`=7, `[R]`=2 all agree).
- Phase 2's `H_OLD2 == H_OLD` row claims a successor that doesn't exist — nothing compares two handles for equality. The substance survives (the host's `previousState` validation fails the update on a wrong read, and the scenario adds a `.not.toBe()` the bash never had), and the scenario's own header comment says so accurately. Only the table row is wrong.
- The `fetch_encrypted_value` `[D]` row (decision D10) names a function that performs no comparisons — it does `get_account` + `try_deserialize` and returns. Unverifiable as written.
- Two client-side *cryptographic proof verifications* in the deleted client (`authorize_public` against live on-chain peaks; `historical_access_leaf_commitment` + `mmr_verify`) appear nowhere in the table. The TS successor checks only `proof.leafCount === state.leafCount`. Coverage is preserved downstream — the gateway verifies against the peaks the request carries — but *diagnosis* is not: a proof service resolving a wrong-but-self-consistent leaf index now surfaces as a generic KMS decrypt rejection.

**9. Doc drift that survived the C9 sweep** — same classes the PR already fixed three times:

- `TESTING_STRATEGY.md:31` — "the maximum 16-record execution's exact 1,077-byte CPI envelope". The test asserts **32 records at 2,133 bytes** (`event_transport.rs:75-76`). Neither number appears anywhere in the tree. Pre-existing, but C9 claims this file is now clean, and it under-states the real envelope against DD-038's 10,240-byte limit by ~2×.
- `TESTING.md:41-45,83-95` — `disclose_packet_fit.rs` is an eleventh Mollusk suite, a test-kit consumer, runs under `cargo test --workspace`, and is listed in neither the coverage paragraph nor the command block.
- `TESTING.md:243-246` — `bun run test:e2e:harness` is described as "the harness unit tests (until / loadEnv)", but `until.test.ts` moved to `src/utils/` in this PR. Someone editing `until()` runs the command the doc names for it and exercises none of the timeout contract.
- `README.md:120-133` — the code map lists `test-suite/fhevm/demo` but neither `e2e/` (the scenario suite) nor `src/solana/` — i.e. it omits exactly the layer this PR made the live vertical, two lines below a row still pointing at `scripts/e2e`.

**10. Cost / maintenance, no correctness impact:**

- `worker-vertical`'s cargo cache can never restore across PRs — the key is job-unique and `solana-e2e` has no `push:` trigger, so nothing ever writes it at the base-branch cache scope. Every PR pays a cold `tfhe-worker` build plus two SBF builds; the 90-min timeout confirms the cold path *is* the normal path. Same class as the pre-existing `host-listener-reconstruct` and `native-coverage`, but the largest instance yet.
- `select-overrides.test.sh` runs in no workflow (pre-existing), and this PR adds a bucket rule it is now the only thing guarding. All 22 tests pass locally, but the two new specimen-program arms have no case, unlike `demo-vault` and `confidential-deposit-app`.
- `token-vertical.scenario.test.ts:194` hardcodes `6064`, a positional derivative of a Rust enum. It fails *closed*, so this is maintenance cost only — but deriving it from the committed IDL at test time would keep the pin and remove the manual step.

---

### Decision register — 15 recorded decisions, adjudicated

Twelve upheld on the evidence, and in three cases the evidence is stronger than the stated rationale:

- **D6** (byte vectors don't duplicate v3 parity) — better justified than claimed: the connector never re-derives the client id for the Solana path (`gw-listener/src/core/publish.rs:311` writes a placeholder, the relayer passes it through), so there is no second implementation to test parity *against*. The ML-KEM wire key is covered anyway, bound into the signing message the vectors pin.
- **D7** (dropped coherent re-read) — sound: every address is derived client-side, so there is no pointer indirection to race, and the one genuinely dropped assertion is enforced on-chain by `#[account(address = token_account.balance_encrypted_value)]`.
- **D3** (hand-built SPL encoders) — upheld: `spl.test.ts` pins the wire layout with literal discriminators and offsets, not by round-tripping the encoder against itself, and the one tautological derivation is backed by base58 goldens.

D4, D5, D8, D9, D13, D14, D15 all verified sound as stated. **D10** is a bookkeeping error (above). **D11** — three reviewers independently confirmed the status-check rename has no in-repo consumer: nothing greps the old name, `.mergify.yml` contains no Solana check and its four queue rules are all `base = main`, and `feature/solana` is not protected. **D12** is the one decision worth pushing back on (finding 5).

Two decisions are sound but live only in the PR description, which means the next person pays for them again:

- **D1** (LiteSVM blocked on the `solana-program-runtime >= 4.1` vs `=4.0.0-rc.0` lockfile conflict, and mollusk 0.15.0's E0658 on rustc 1.91.1) — `TESTING_STRATEGY.md:103` still lists #3045 as an ordinary deferred follow-up. The lockfile wall is documented in `runtime-tests/Cargo.toml` but not tied to the issue. Whoever picks up #3045 re-derives both walls from scratch.
- **D2** (L4-a left with the kms repo's live harness) — reasonable division, but nothing in this repo now points at where that attack is exercised. Cross-repo coverage without a backlink is how coverage quietly stops being run.

---

### Not verified — stated rather than guessed

- Classic branch protection on `main`: the API 404s at this token's permission level. `main` never ran `solana-e2e`, so a required-check entry is implausible, but it is the one thing not ruled out by observation. Likewise server-side Mergify queue conditions.
- `dep_chain`'s IDL by regeneration: no anchor/SBF toolchain available. Source↔IDL agreement was checked by hand (instructions, accounts, error codes 6000-6002 in order), and CI's `solana/**`-gated `build-and-test` does regenerate it. The other five vendored IDLs are byte-identical to `solana/target/idl/`.
- Bun's scenario file discovery order was reasoned about, not executed — relevant only to the observation that the proof-service restart gate now sits inside one scenario file rather than ahead of the whole run, so its blast radius (every proof in the run being post-restart) has shrunk. The gate itself still fires and still throws.

PR #3480, opened 2026-08-10
chore(sdk): isolate js-sdk from root npm workspace

geoxel:
Remove sdk/js-sdk and sdk/js-sdk/src from the root workspaces list so the SDK is no longer built as part of the monorepo's npm install/build graph.

test-suite/e2e now depends on a registry-pinned @fhevm/sdk version by default and installs it via a new scripts/install-sdk.sh helper, which supports two modes selected by the Dockerfile's SDK_SOURCE build arg:
- local (default): build and pack sdk/js-sdk, install the tarball, so e2e tests exercise the local, unreleased SDK.
- registry: install a fixed version from npm (defaults to the version pinned in package.json).

This decouples js-sdk's dependency resolution (and its prettier version pin) from the rest of the workspace, and lets e2e Docker builds choose between testing against local SDK changes or a released version.


geoxel:
Replaced by https://github.com/zama-ai/fhevm/pull/3652

PR #3481, opened 2026-08-11
feat(relayer): return all user-decrypt shares

manoranjith:
Implements [#1737][issue] on `release/0.13.x`.

- The relayer waits for up to `x` extra user-decryption shares, for at most `d` seconds after the threshold is reached, then returns every share it collected.
- That buys the client tolerance for `x` invalid KMS responses — enough to keep decryptions succeeding through a KMS migration window, where some parties answer with shares that no longer validate. Previously the relayer returned exactly the threshold, so one invalid share failed the decryption.
- v3 is untouched: there is no v3 implementation to change — its handler wraps v2's, and both read the same share query.

### Config

```yaml
gateway:
  contracts:
    user_decrypt_shares_threshold: 9                 # existing
    user_decrypt_additional_shares: 2                # x: extra shares to wait for
    user_decrypt_additional_shares_timeout_secs: 5   # d: max wait once the threshold is reached
```

- Either key at `0` disables the wait and restores the previous latency. Both are adjustable at runtime through the admin config API.
- Updated: `relayer/config/local.yaml.example`, `local.testnet.yaml.example`, `local.mainnet.yaml.example`, and the local stack's `test-suite/fhevm/templates/config/relayer.yaml` (wait left off — its KMS is centralized).

### Testing

**Relayer integration suite** — driven from an RPC mock that now emits a configurable number of shares instead of exactly the threshold.

- Covers the wait ending on `x` being reached, on a late share, and on `d`-second expiry — expiry returns every share collected, never an error or a truncation back to the threshold.
- Cases that end on the shares are told apart from cases that end on the clock by elapsed time, so they would fail if the wait were removed.
- Direct and delegated endpoints are both covered.
- Locally: 283 integration and 193 unit tests pass; `make lint` and `make openapi-check` clean.

**Manual, against a deployed threshold KMS** — the e2e suite cannot cover this: `release/0.13.x` has no multi-party KMS (`plan.kms`, `committeeSize` and `KmsMode` do not exist there), and a spare share needs a committee.

Read the share count with the `fhevm-sdk` CLI instead — it lives at `sdk/cli-js-sdk` on `main`, not on this branch, and pins `@fhevm/sdk` 0.13.2. Needs Node >= 22 and pnpm >= 10.

```bash
git clone --branch main --depth 1 https://github.com/zama-ai/fhevm.git /tmp/fhevm-main
cd /tmp/fhevm-main/sdk/cli-js-sdk
pnpm install
pnpm run build
cd packages/cli && pnpm add -g .   # or skip and use `pnpm --silent run cli` from the workspace root
fhevm-sdk --help
```

Configure — either export these or copy `sdk/cli-js-sdk/.env.example` to `.env`, which is loaded automatically:

```bash
export PRIVATE_KEY=0x...                          # funded on the target network; MNEMONIC also works
export SEPOLIA_RPC_URL=https://sepolia.drpc.org   # used by both `testnet` and `devnet`
export ZAMA_FHEVM_API_KEY=...                     # required on mainnet; devnet is keyless
```

Run a user decryption and count the shares it came back with:

```bash
fhevm-sdk -n devnet user-decrypt fresh --artifact /tmp/ud.json
fhevm-sdk -n devnet verify-user-decrypt --artifact /tmp/ud.json
```

- `fresh` encrypts a value, stores it in `FHETest` and decrypts it, so it needs gas; `user-decrypt stored` re-decrypts what the wallet already holds in its `FHETest` slots and needs no transaction.
- `verify-user-decrypt` prints `shareCount`, `kmsThreshold` and `kmsSignerCount`, taken from the KMS-verified reconstruction rather than from anything the relayer self-reports. The spare count is `shareCount - kmsThreshold`.
- `--relayer-url` targets a relayer built from this branch instead of the network's deployed one.

### Unrelated CI fixes

Two commits repair checks that were already red on `release/0.13.x`, independent of this change:

- `ci(relayer): pin sqlx-cli to the 0.8.x series` — the unpinned install resolved to sqlx-cli 0.9.0, which needs rustc 1.94 while the branch pins 1.91.1.
- `chore(relayer): bump audit-flagged dependencies` — crossbeam-epoch, quinn-proto and ruint, for RUSTSEC-2026-0204, -0185 and -0220.

[issue]: https://github.com/zama-ai/fhevm-internal/issues/1737


geoxel (approved):

manoranjith:
@mergify queue

zmalatrax (approved):
blind approval, checks are passing, haven't looked much into the actual design, its implementation and the corresponding test suite/docs
But at a high-level it should fix the encountered issue with decryption response returning invalid KMS shares when threshold could have been reached by using the additional shares that were not forwarded before

relayer/src/store/sql/repositories/user_decrypt_repo.rs
```diff
@@ -1077,22 +1075,12 @@ impl UserDecryptRepository {
...
-                WHERE gw_reference_id IN (
-                    SELECT gw_reference_id FROM user_decrypt_req WHERE ext_job_id = $1
-                )
-                ORDER BY created_at ASC, share_index ASC
-                LIMIT COALESCE(
-                    (SELECT resolved_threshold FROM user_decrypt_req WHERE ext_job_id = $1),
-                    $2
-                )
-            ) s ON r.gw_reference_id = s.gw_reference_id
+            LEFT JOIN user_decrypt_share s
```
nboisde, line 1078: user decrypt req table is the only one where we don't store res, so fix works there 👍

relayer/src/http/endpoints/v2/handlers/user_decrypt.rs
```diff
@@ -647,50 +658,122 @@ impl UserDecryptHandler {
...
-                                    (
-                                        StatusCode::INTERNAL_SERVER_ERROR,
-                                        Json(UserDecryptStatusResponseJson {
-                                            status: ApiResponseStatus::Failed,
-                                            request_id: request_id.to_string(),
-                                            result: None,
-                                            error: Some(
-                                                V2ErrorResponseBody::internal_server_error(
-                                                    "Internal server error",
+                            // Optimistic wait window: once reconstructable, keep reporting
```
nboisde, line 661: LGTM

nboisde (approved):
LGTM !

PR #3491, opened 2026-08-12
feat(sdk): support deterministic (seeded) encryption in the JS SDK

ppoliani:
Adds an optional `seed` to the encrypt pipeline so callers can produce a deterministic ciphertext/proof for a given value, instead of the client always sampling fresh randomness for it.

- `buildWithProofPacked` calls `build_with_proof_packed_seeded` on the TFHE WASM builder when `parameters.seed` is set, falling back to the existing `build_with_proof_packed` otherwise — same public key/CRS validation either way, just a different randomness source.
- `seed` is threaded end-to-end through the existing call chain rather than bolted on at the WASM boundary: `encryptValue` → `createZkProof` / `ZkProofBuilder.build` → coprocessor `encrypt` → `buildWithProofPacked`. It sits alongside `contractAddress`/`userAddress`/`options` on the public `encryptValue` params.
- Same value + same seed now produces the same `encryptedValue`/`inputProof` across calls. Omitting `seed`, or using a different one, still produces distinct output each call — unchanged from before.

### Constraints

- `build_with_proof_packed_seeded` only exists in TFHE `1.6.2`'s (and `1.6.0-dev`'s) WASM bindings — `1.5.3` doesn't implement it at all. Which TFHE version is active is resolved per-chain by `HyperWasmSolver`'s protocol-compatibility table, not chosen by the caller, so seeded encryption is only reachable on chains that resolve to `1.6.2`+.
- No runtime guard yet for calling with `seed` against a resolved version that lacks the seeded builder — it currently surfaces as a raw `TypeError: ... is not a function` from the WASM layer rather than a typed SDK error. Worth a follow-up.

### Testing

Added `should allow deterministic encryption` to the ethers encrypt suite (`clientEncrypt.encrypt.tests.ts`):

- Encrypts the same value twice with no seed → asserts `encryptedValue`/`inputProof` differ.
- Encrypts the same value twice with the same seed → asserts `encryptedValue`/`inputProof` are identical.

The test forces its client to TFHE `1.6.2` (`moduleVersions: { tfhe: '1.6.2', checkCompatibility: 'off' }`) since the suite's default chain resolves to `1.5.3`, which lacks the seeded builder. It also evicts the global FHE key cache (`globalFheEncryptionKeyCache`) before and after — that cache is keyed by relayer URL only, not by TFHE version, so forcing a version without evicting the entry populated by earlier tests in the file throws `TfheVersion mismatch`.

```bash
CHAIN=localstack npx vitest run --config test/fheTest/vitest.config.ts ethers/clientEncrypt.encrypt.test.ts

### CODE REVIEW

File: sdk/js-sdk/test/fheTest/ethers-common/clientEncrypt.encrypt.tests.ts [hunk 1]
```diff
@@ -108,5 +109,66 @@ export function defineClientEncryptEncryptTests(parameters: {
         );
       });
     }
+
+    it('should allow deterministic encryption', async () => {
```
geoxel, line 113: Could we also create a viem version of this test in `test/fheTest/viem-common/clientEncrypt.encrypt.tests.ts` ? (even if viem/ethers are not quite involved here, this is just for coherence)

geoxel:
Claude review below. Points 1 and 2 are the most important IMHO.

1. **Public API docs are now out of date** — `docs/api-reference.md` explicitly lists the `encryptValue`/`encryptValues` parameter shape (~L171, L182) and doesn't include the new `seed` field at all. `docs/encryption.md` has no mention of `seed` either. Anyone reading the docs won't discover this feature exists.

2. **No caveat, anywhere, about the privacy implication of seed reuse.** With a caller-supplied seed, the randomness for the packed ciphertext build is derived solely from `seed` (not from `contractAddress`/`userAddress`/metadata, which only get folded into the proof's associated data). That means encrypting the *same value* with the *same seed* under *different contracts* can produce the identical underlying ciphertext bytes, not just identical bytes within one context — i.e. a plaintext-equality oracle that spans contracts, visible to anyone observing on-chain input ciphertexts. That may well be an acceptable, intentional tradeoff for the feature's use case (reproducible tests, idempotent inputs, etc.), but nothing in the type (`EncryptValueParameters.seed`) or docs warns callers not to reuse a seed across genuinely distinct/sensitive values. Worth folding into the docs update in point 1.

3. **Minor style nit:** `EncryptValueParameters.seed?: Uint8Array` and `EncryptValuesParameters.seed?: Uint8Array` (in `encryptValue.ts` / `encryptValues.ts`) drop the `| undefined` suffix that the adjacent `options` field and every downstream `seed` type in the call chain use. No functional bug under `exactOptionalPropertyTypes` (destructuring still widens correctly), just a one-off inconsistency.

ppoliani:
> Claude review below. Points 1 and 2 are the most important IMHO.
> 
> 1. **Public API docs are now out of date** — `docs/api-reference.md` explicitly lists the `encryptValue`/`encryptValues` parameter shape (~L171, L182) and doesn't include the new `seed` field at all. `docs/encryption.md` has no mention of `seed` either. Anyone reading the docs won't discover this feature exists.
> 2. **No caveat, anywhere, about the privacy implication of seed reuse.** With a caller-supplied seed, the randomness for the packed ciphertext build is derived solely from `seed` (not from `contractAddress`/`userAddress`/metadata, which only get folded into the proof's associated data). That means encrypting the _same value_ with the _same seed_ under _different contracts_ can produce the identical underlying ciphertext bytes, not just identical bytes within one context — i.e. a plaintext-equality oracle that spans contracts, visible to anyone observing on-chain input ciphertexts. That may well be an acceptable, intentional tradeoff for the feature's use case (reproducible tests, idempotent inputs, etc.), but nothing in the type (`EncryptValueParameters.seed`) or docs warns callers not to reuse a seed across genuinely distinct/sensitive values. Worth folding into the docs update in point 1.
> 3. **Minor style nit:** `EncryptValueParameters.seed?: Uint8Array` and `EncryptValuesParameters.seed?: Uint8Array` (in `encryptValue.ts` / `encryptValues.ts`) drop the `| undefined` suffix that the adjacent `options` field and every downstream `seed` type in the call chain use. No functional bug under `exactOptionalPropertyTypes` (destructuring still widens correctly), just a one-off inconsistency.

@geoxel I pushed the proposed changes

PR #3493, opened 2026-08-12
fix(coprocessor): harden blue-green cutover txn

goshawk-3:
Makes `execute_cutover` produce the same state on every operator, and recover on its own when it fails.

Related: https://github.com/zama-ai/fhevm-internal/issues/1614

- Deletes blue's whole in-window output and lets the merge restore green's copy, in both the legacy and `*_branch` form of every table. Blue's rows surviving in the branch tables would reintroduce the per-operator divergence this prevents.
- Drops `dependence_chain` rows left with no computations, scoped to the chains of the deleted rows via `DELETE ... RETURNING` and checked across both schemas.
- Preserves `txn_is_sent` across the merge, so the tx-sender does not re-broadcast handles already committed on-chain.
- Refuses to cut over while blue's pre-window ciphertexts still owe S3 an upload. Only blue can finish those, so retiring it first would leave an on-chain digest pointing at an object that was never written.
- Retries cutover with bounded backoff (10 attempts, 0.5s to a 60s cap, cancellable) instead of a hot loop with no delay. Rows stay `UpgradeAuthorized`, so `reconcile` keeps retrying after the budget.
- Handles multi-chain proposals: every step walks each host chain's own window. The previous code read a single row and silently skipped the rest.
- Nothing before `pg_advisory_xact_lock` takes a row or table lock. Holding one there inverts the host-listener's order and deadlocks against activation.

### DIFF
--- a/coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
+++ b/coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
@@ -7,7 +7,7 @@
 
 use std::sync::atomic::{AtomicI64, Ordering};
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use fhevm_engine_common::database::GCS_SCHEMA_QUOTED;
 use fhevm_engine_common::gcs_activation::EVENT_DRY_RUN_ROLLED_BACK;
@@ -21,7 +21,10 @@ use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, warn, Level};
 
 mod coprocessor_tables;
-pub use coprocessor_tables::{CoprocessorTable, COPROCESSOR_TABLES};
+pub use coprocessor_tables::{
+    CoprocessorTable, CIPHERTEXT_TABLES, COMPUTATION_TABLES, COPROCESSOR_TABLES,
+    PBS_COMPUTATION_TABLES,
+};
 
 pub const UPGRADE_ACTIVATED_CHANNEL: &str =
     fhevm_engine_common::gcs_activation::EVENT_UPGRADE_ACTIVATED;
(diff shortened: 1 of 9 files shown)


coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -869,6 +910,303 @@ async fn transition_to_gw_dry_run_started(
...
+    );
+
+    // Last: the ciphertext deletes above reach their rows through these tables.
+    //
+    // `RETURNING` hands the deleted rows' dependence chains straight to
+    // `cutover_touched_chains`, which `cleanup_orphaned_dependence_chains` drains once
+    // every chain has been walked. `dependence_chain` has no chain or block column, so
+    // this is the only way to scope it to the upgrade window.
+    for comp_table in COMPUTATION_TABLES {
+        let sql = format!(
```
PanGan21, line 1041: I am thinking that we might need to take care of the `dependents[]` and `dependency_count` of the remaining rows that reference the deleted dependence chains.
It seems quite unlikely to happen but in theory a surviving chain that depended on a deleted one could keep `dependency_count` > 0 and never get scheduled  . What do you think?

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -1032,23 +1520,38 @@ pub async fn execute_cutover(pool: &Pool<Postgres>) -> Result<(), Error> {
         merge_gcs_table(&mut tx, table.name, table.conflict_cols).await?;
     }
 
-    // 5. Drop the gcs schema (and everything in it) now that its data has been
+    // This is to avoid CoprocessorAlreadyAdded (false negatives) for handles
+    // that were already committed by BCS on-chain before cutover.
+    // NB: This query must be deleted later on.
+    let resent_guard = sqlx::query(
+        "UPDATE public.ciphertext_digest d SET txn_is_sent = TRUE
```
PanGan21, line 1527: Should this restore also `tx_is_sent` for `ciphertext_digest_branch`? Or you skip this because it will be removed soon?

goshawk-3: Since we intend to remove the _branch tables, let's avoid that. 

I'm actually planning to push another PR that removes all references to _branch tables.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
goshawk-3: We preserve the dependence_chain table of the GCS so that scenario should not be possible but let me think about it.

goshawk-3: Issue filed to consider this once the stack is merged. 
ref: https://github.com/zama-ai/fhevm-internal/issues/1936

PanGan21 (approved):

PR #3494, opened 2026-08-12
test(relayer): use common setup in recovery tests

manoranjith:
Three things are broken on `main`, all in the relayer's test tooling.

**The recovery tests fail.** They build their own mock setup instead of using the shared one in `tests/common/utils.rs`, so they never registered a contract call the relayer makes while starting up. The relayer gives up, and the tests fail with `channel closed`.

**CI does not run them.** `make test-run` — all that `relayer-tests.yaml` invokes — listed test binaries by hand and left out five: `recovery_test`, `admin_endpoint_test`, `schema_isolation_test`, `threshold_resolver_test`, `keyurl_poller_test`.

**`make lint` fails.** #3049 changed the served `dataId` to hex and hand-edited `openapi.yml`, but the annotation that file is generated from still described a decimal string, so regenerating the spec reverted the edit.

What this changes:

- The recovery tests use the shared setup: host mock, gateway mock, relayer startup, and an isolated database schema like every other integration test. Adding the missing mock to the duplicated setup would have fixed the symptom and kept the duplication that caused it.
- `make test-build` and `make test-run` select `--test '*'`, so every file under `tests/` is covered, including ones added later.
- The `dataId` annotation says hex, and `openapi.yml` is regenerated from it.
- Both lockfiles are refreshed; the one under `test-support/ethereum_rpc_mock` was stale enough that any `cargo test` there rewrote it.

Test plan:

- [ ] CI: relayer integration tests. Locally the full suite passed before the lockfile refresh (`recovery_test` 6/6); since then only the build has been re-run.

### DIFF
--- a/relayer/Cargo.lock
+++ b/relayer/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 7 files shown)


geoxel (dismissed):

nboisde (dismissed):

manoranjith:
@mergify queue

nboisde (approved):
Since no modifications, and only a codeowners removal.

manoranjith:
@mergify queue

PR #3496, opened 2026-08-12
chore(kms-connector): chain-id in error logs

eudelins-zama:
### Motivation

When the KMS Connector is configured with multiple host chains, RPC call errors did not indicate which chain the failing node belongs to, making them hard to investigate.

### Changes

- Introduce a `HostRpcClient` wrapper (new `rpc.rs` module) around each host chain's `ACLInstance`, tagged with its chain ID. All host-chain RPC calls of the `DecryptionProcessor` (ACL checks and ERC-1271 signature verification) now go through it.
- Errors are wrapped with `on host chain {chain_id}` + ACL contract address when relevant.
- Add a generic `RequestCheckError::context` method to attach this context to the inner error.
- Rename the `DecryptionProcessor`'s `acl_contracts` field to `host_clients` accordingly.

### DIFF
--- a/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/decryption.rs
@@ -1,7 +1,7 @@
 use crate::core::{
     config::Config,
     event_processor::{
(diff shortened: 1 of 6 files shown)


maksymsur (approved):
LGTM!

PR #3502, opened 2026-08-12
refactor(solana): one step ceiling, a typed admission contract, and a measured frontier

Eikix:
# One step ceiling, a typed admission contract, and a measured frontier

Closes the heap-cap thread of fhevm-internal#1872.

## What this PR delivers

The branch started from three contradicting step caps — a guessed 16 blocking a real design, a measured-at-the-wrong-wall 20, and the host's 32. It ends with:

- **One step ceiling** — the host's `MAX_FHE_EXECUTION_STEPS = 32`; both copies deleted (`RELIABLE_STEPS`, `MAX_ON_CHAIN_EXECUTION_STEPS`, the `raised-heap` feature).
- **A builder that only returns executions a transaction can actually execute.** Every wall an app could previously discover as a runtime abort — heap exhaustion with no error, `MaxInstructionTraceLengthExceeded`, a rejected CPI — is now a typed error at `build()`, at the step that crosses it where possible. The heap budget covers the instruction's whole app-side lifecycle: build, serialize, resolve, invoke.
- **An exploration of the execution-shape frontier as regenerable data** — eleven host-side boundary sweeps and a 122-shape app-side grid (43 admitted), summarized below.
- **Real CU wins as a side effect** — batcher lifecycle −5–22% CU per instruction, token instructions −9–29%.

## The typed admission contract

| Axis | Ceiling | Where checked | Error |
|---|---|---|---|
| Steps | 32 (`MAX_FHE_EXECUTION_STEPS`) | each step | `TooManySteps` |
| Instruction trace | 64 per transaction — 3 system CPIs per created output + event CPIs + the app wrapper; at most **20 creates** per execution | each step | `ExceedsInstructionTraceLimit` |
| CPI packet | 10 KiB (`MAX_INSTRUCTION_DATA_LEN`, checked by the runtime on every invoke; an `fhe_execute` packet always travels by CPI — a transaction carries at most 1,232 bytes) | `finish`, exact byte count | `ExceedsCpiInstructionDataLimit` |
| Build heap | build + packet + **invoke-side account tables** ≤ 24 KiB of the fixed 32 KiB never-freeing region, leaving an 8 KiB reserve for what the builder genuinely cannot see: Anchor deserialization and the app's own allocations | build term: **pre-emptively, at every allocation site** — each allocation is admitted against the budget before the allocator serves it, so the tally never crosses the budget even transiently; packet + invoke terms: `finish`, exact byte count | `ExceedsBuildHeapBudget` |

Three properties keep this honest rather than guessed:

1. **The build-heap tally is proven, not modeled.** Every builder-owned table is a `TalliedVec` — a vector that records the bytes it requests from the allocator as it grows — so the tally is a property of the container, not a protocol callers must remember. `heap_budget.rs` asserts the total equals a counting global allocator **byte-for-byte** across every admitted shape — a new allocation that forgets to tally fails CI. This is the honest re-introduction of the reserve that produced the old guessed 16: the same 8 KiB idea, but applied per build to exact tallied bytes instead of converted into one step count for all shapes.
2. **The invoke side is priced the same way.** `finish` charges `resolve_accounts`'s three exact-sized vectors and the CPI account meta/info tables as an exact function of the account counts it already knows; a second counting-allocator test runs the real resolve + CPI-table code for all 43 admitted shapes and asserts the model **byte-exact**. To make that pricing exact, resolve dropped its per-candidate requirement clones and the invoke path reserves its table tails up front instead of doubling into them.
3. **The trace model is tied to measurement.** Mollusk probes the host top-level, one instruction short of the app wrapper the builder budgets, and the sweep asserts the measured wall sits exactly one create past the builder's cap (21 vs 20). The model and the measurement cannot drift apart silently.

`FheExecution::cost()` exposes the numbers — steps, creates, updates, events, exact packet bytes, tallied build and invoke heap bytes, the dynamic-account and authority counts, the guaranteed trace floor and the state-dependent worst case — so an app composing a transaction with more than the minimal wrapper can budget the rest.

### Host-side walls (Mollusk, top-level, regenerated snapshot)

| Shape | Runs up to | First fail | Wall |
|---|---|---|---|
| `dependent_chain` (transient chain, 1 create) | **32** | 33 | the step cap — no runtime wall below it |
| `mixed_chain_creates` (every 4th step creates, public) | **32** | 33 | the step cap |
| `all_private_creates` (a create per step) | 21 | 22 | instruction trace (64; 3 CPIs/create) |
| `subject_heavy_creates` (a create per step, 8 subjects each) | 21 | 22 | instruction trace |
| `all_created_public` (a public create per step) | 20 | 21 | host heap (trace one step away) |
| `subject_heavy_public_creates` (a public create per step, 8 subjects each) | **15** | 16 | **host heap — see the gap below** |
| `attestation_per_step` (max-size verified input per step) | 15 | 16 | host heap |
| `mature_updates_peaks_8` (updates of 8-peak values) | 11 | 12 | host heap |
| `mature_updates_peaks_32` | 6 | 7 | host heap |
| `mature_updates` (55 peaks, adversarial maturity) | 4 | 5 | host heap |
| `reduction_heavy` (60-operand sums) | 4 | 5 | host heap |

Readings:

- **Chain-shaped work — what the RFQ design needs — reaches the full 32 steps with no runtime wall in sight.**
- The **update-heap wall is a curve over on-chain state**, not a constant: MMR peak count (`popcount(leaf_count)`) drives the account-decode cost, and a value needs 2⁵⁵ leaves of history to force the 4-step extreme. The state is invisible at build time; the curve is the guidance.
- `reduction_heavy` was an unmeasured wall until this branch; the builder's budget stops the same shape earlier on the app side, so it cannot be reached through `build()`.

### One wall the builder cannot type: the host's own CPI frame

The `subject_heavy_public_creates` sweep found a wall the one-subject sweeps could not see. The host CPI runs in its own fresh 32 KiB heap frame, and what it allocates there grows with **creates × subjects-per-output** — each created account's subject table plus the public-outputs event payload. No app-side number can price this: with a shared audience the builder's packet interns the eight subjects **once** in the dictionary, while the host materializes them **per created account**. So the builder's typed ceilings admit the trace-capped 20 such creates, and the host survives 15.

This gap is now pinned **from both sides**: a builder test proves `build()` admits 15, 16, and 20 such creates (`the_builder_admits_what_the_host_heap_cannot_hold`), and the sweep proves the host wall is 15 and asserts the gap's direction so it cannot drift silently. It has its own [ANTI] invariant, #61 — `build()`'s contract is explicitly scoped to the *app-side* instruction, and what it deliberately does not promise is the host's side of the CPI. One consequence deserves the reviewer's eye: **deleting the old guessed 16-step cap newly admits the 16–20 band on-chain for this one shape** — those executions build cleanly and die on-chain with no error. Closing the gap is a design decision — a typed policy cap on the public-audience payload, a host-side heap model, or leaving it measured — proposed for follow-up in #1872 rather than guessed into this PR.

### App-side frontier (builder admission, 122 shapes, 43 admitted)

The grid (`print_build_frontier_grid`) sweeps persist-kind × output-count × subject-width, the attestation ladder, and the reduction/mixed-ops axes. Every admitted shape fits the program heap with the full 8 KiB reserve intact; every inadmissible shape is a typed rejection. Representative rows (totals are build + packet + invoke):

| Shape | Total heap | Verdict |
|---|---|---|
| 4-step typical (1 create) | 13,604 B | admitted |
| full 32-step chain (1 create) | 13,828 B | admitted |
| 20 creates, 1 subject | 21,595 B | admitted |
| 20 creates × 2 subjects | 22,295 B | admitted |
| 8 public creates × 8 subjects | 23,367 B | admitted (near budget) |
| 8 is_in steps × 8 operands | 23,715 B | admitted (nearest budget) |
| 4 max-size attestations | 21,730 B | admitted |
| one 60-operand sum | 21,278 B | admitted |
| 16+ creates × 4 subjects | — | `ExceedsBuildHeapBudget` |
| 21+ creates | — | `ExceedsInstructionTraceLimit` |
| 5–10 attestations | — | `ExceedsBuildHeapBudget` |
| 11+ attestations | — | `ExceedsCpiInstructionDataLimit` |
| 2+ steps of 60-operand sums | — | `ExceedsBuildHeapBudget` |

The ~10 KB floor on every row is the up-front table reservation — the price of never letting a `Vec` double on a heap that never frees. Charging the invoke tables shrank the admitted frontier from the interim 54 shapes to 43: the rows that left (20 creates × 4 subjects, 5–6 attestations, wide-subject updates) are exactly the shapes the adversarial review proved could build but not invoke.

### Regenerate either side

```
# host walls (Mollusk sweeps → cost-snapshots/fhe_execute_boundary.json)
bash scripts/update-cost-snapshots.sh
# app frontier
cargo test -p zama-fhe --lib --features cpi heap_budget::print_build_frontier_grid -- --ignored --nocapture
```

## Performance

Deleting the copies the tally exposed pays on-chain (all vs the branch base, regenerated snapshots):

- `EncryptedValueId` derives its PDA once at construction instead of per `address()` call, lowering moves subject lists instead of cloning them, invoke consumes the execution, and each `EncryptedValue` decodes once: batcher lifecycle **−5–22% CU** per instruction (`dispatch` 292,585 → 230,880; `redeem_claim` 316,133 → 246,334), token instructions **−9–29%** (`confidential_transfer/steady_state` 314,607 → 236,661).
- The exact-sized resolve path (first review round) trimmed another few hundred CU per batcher instruction.
- `fhe_execute/max_steps` itself: 151,018 → 79,569 CU (right-sized packet, no argument clone).
- The heap-safety rounds give a sliver back: the per-step gate cost +0.3–0.7% CU, and admitting every allocation before the allocator serves it costs another +0.5–1.5% on the builder-driven app instructions (net across both, e.g. `batcher/claim` 225,333 → 230,365, `confidential_transfer/steady_state` 236,661 → 240,443 — still far below the branch's starting numbers). That is the price of the guarantee that the builder's tally can never cross the budget even transiently, so the builder alone can never exhaust the heap region.

## What deliberately stays measured, not typed

- **Host heap for mature updates** — driven by MMR state the builder cannot see; covered by the per-maturity sweeps above.
- **Host heap for wide public audiences** — the creates × subjects driver described above; pinned at 15-of-20, design decision proposed in #1872.
- **State-dependent trace additions** — a rent top-up per growing update, the one-time lazy block-meter creation (3 CPIs); exposed through `cost().instruction_trace_worst_case()` and documented, never guessed into a cap.
- **Block-meter write contention** (every `fhe_execute` in a block serializes on the shared meter) — a throughput property, out of scope here; flagged in #1872 as follow-up.

## Verification

- Full workspace suites green against freshly built SBF programs (both counting-allocator proofs run in the workspace suite via feature unification); `cargo clippy --workspace --all-targets` clean; IDL/ABI goldens in sync.
- Confidential vault demo e2e (`deposit-arc` scenario — join → dispatch → settle → claim → user-decrypt, with a KMS user-decrypt roundtrip at the end) against the live docker stack built and deployed from this branch: **passed**; re-run by CI's solana-e2e on every push.

## Commit walk

Each commit is green on its own:

1. `test(solana)`: probe fhe_execute capacity as a shape matrix, each wall named
2. `refactor(solana)`: delete raised-heap — the heap is fixed, decided in DD-046
3. `perf(zama-fhe)`: invoke consumes the execution — no argument clone, one right-sized packet
4. `perf(zama-host)`: decode each EncryptedValue once per execution
5. `perf(solana)`: borrow signer seeds in place, reserve builder tables up front
6. `refactor(solana)`: one step ceiling — delete the on-chain cap
7. `fix(solana)`: address adversarial review of the series
8. `feat(zama-fhe)`: admit only executions that can land — typed ceilings for trace, packet, and build heap
9. `test(solana)`: sweep the frontier's missing axes, tie the trace model to the measured wall
10. `docs(solana)`: the four typed ceilings replace the measured-fit story
11. `fix(solana)`: satisfy CI — rustfmt, typo, and dead-surface declarations
12. `fix(ci)`: install the SDK dependency graph before the import check in solana-e2e (repairs a base breakage from #3492, which never triggered this workflow)
13. `feat(zama-fhe)`: charge the invoke-side account tables against the build heap budget
14. `test(solana)`: measure the invoke model byte-for-byte, pin the host-side public-audience wall
15. `docs(solana)`: the budget bounds the whole app-side lifecycle, and one wall stays measured
16. `refactor(zama-fhe)`: tables that tally their own growth, a per-step heap gate, ops split from admission
17. `test(solana)`: the boundary sweep becomes its own table-driven binary with its own snapshot
18. `docs(solana)`: scope `build()`'s contract to the app side; the host-side gap becomes invariant #61
19. `fix(zama-fhe)`: admit every allocation against the budget before the allocator serves it
20. `perf(zama-fhe)`: validate a persistent output without building the binding the caller discards
21. `test(solana)`: report every boundary sweep before failing; refresh moved-test pointers and snapshots

## Second review round: structure

Commits 16–18 answer the two architecture reviews on this PR. The shape of the code now matches the shape of the design:

- **`heap_tally.rs`** owns the allocator's growth model and `TalliedVec<T>` — a crate-private vector that records every byte it requests as it grows, so growth accounting is a property of the container instead of a `&mut usize` threaded through three files. `accounts.rs` no longer imports anything cost-related.
- **`builder.rs` (393 lines) is only the admission machine**: the tables, the per-step `commit_step` gate (the single home of the `TooManySteps` and per-step heap checks), and `finish`. The ~700 lines of named op constructors live in **`ops.rs`**; the lowering/rollback tally protocol lives in **`lower.rs`**. Gating the build-heap term per step (instead of only at `finish`) moved two shapes from a `finish`-time packet error to the earlier heap error; the packet ceiling stays pinned by a shared-audience shape that is packet-heavy but build-cheap.
- **`fhe_execute_boundary.rs`** is the capacity instrument extracted from `host_mollusk.rs` (now behavior-only, ~5,800 lines from 6,722): one shape table drives the sweeps, the per-shape wall pins (`WallPin`), the snapshot (`cost-snapshots/fhe_execute_boundary.json`), and the `#[ignore]`d matrix printer. Shared fixtures moved to `tests/host_fixtures/`. Every shape is swept before any failure is reported, so a moved boundary surfaces as one complete report — and an update run can never leave the snapshot partially rewritten.
- All 11 measured boundaries and both counting-allocator proofs came through the refactor **byte-identical** — the frontier is still 43 admitted of 122.

## Third round: the heap gate becomes pre-emptive

An adversarial self-review of the structural round found the gate's original placement unsound: it ran *after* a step's lowering had already allocated, so a near-budget build could request past the 32 KiB region itself before the gate could reject (eleven 8-operand set memberships followed by one 60-operand sum requested 33.5 KB, measured under the counting allocator) — an abort with no error, the exact failure mode `build()` promises away. Commits 19–21 close it:

- **Every allocation is now admitted before it happens.** The tally protocol in `lower.rs` headroom-checks each site — table growth, dictionary interning, purpose-table merges, attestation embeds — against the budget before the allocator serves the bytes, and the reduction ops clamp their caller-sized operand tables to the policy cap and admit the reservation up front (which also fixes an unvalidated `size_hint` sizing an allocation). The post-lowering gate remains as defense-in-depth. Net claim, now true and CI-enforced: **the tally never crosses the budget, even transiently, even if the app ignores rejections and keeps building** — proven adversarially by `the_tally_never_crosses_the_budget_even_transiently`, which drives the three heaviest per-step allocators into rejection with a probe after every call.
- **`PersistentOutput::validate()`** — the token and batcher programs called `binding()` purely as a validator and discarded the result, paying a subject-list clone on the never-freeing heap for nothing (a regression the structural round's `binding()` dedup introduced). Validation is now a shared allocation-free path; `binding()` validates before cloning, so its rejecting path allocates nothing.
- The boundary sweep reports all eleven shapes before failing, and the stale pointers the extraction left behind (module docs, the dead-surface root list, `docs/TESTING.md`) now name `fhe_execute_boundary.rs`.


Eikix:
[AUTOMATED]

## Thermo-nuclear code quality review

**Request changes.** The admission contract is the right design. The implementation still scatters it, checks heap too late, and grows two files that were already at their size limit. Behavior looking correct is not enough on this bar.

The branch does the thing it claims: one step ceiling (`MAX_FHE_EXECUTION_STEPS = 32`), typed refusals for trace / CPI packet / build heap, a byte-exact counting-allocator proof, host walls as snapshot data. That part should survive. What should not survive is how it was bolted on.

### Blockers

**1. `host_mollusk.rs` 5894 → 6722.** ~800 lines of a self-contained sweep (`ProbeCase`, `sweep_step_boundary`, eleven case builders, eleven near-identical `cost_snapshot_boundary_*` tests, plus `print_boundary_matrix` re-listing the same eleven shapes) landed in a file that was already past any healthy bound. Extract `runtime-tests/tests/fhe_execute_boundary.rs` (or similar). Drive tests and the printer from **one** shape table so those eleven functions disappear.

This is not a host behavior test. It is a capacity instrument. Dumping it into a 5.9k-line file guarantees the next axis lands in the same pile.

**2. `builder.rs` 940 → 1095.** It crossed 1k. The new work is admission (`commit_step` trace check, counters, tally, `finish` packet+invoke budget) sitting in a catalog of identical op wrappers. Keep `builder.rs` as the admission machine (`new` / `commit_step` / `finish`); move named methods to `ops.rs`. Adding more comments to a 1095-line file is not the fix.

**3. The heap tally is a protocol sprinkled across layers, not a type.** `cost::tally_push` is now called from `builder.rs`, `lower.rs`, and `ExecutionAccountMeta::promote`. `promote` grew a `tally: &mut usize` so account-metadata mutation knows about bump-allocator accounting. That is feature logic leaking into a shared path.

Own growth in a crate-private `TalliedVec<T>` (or solely in `StepTables`). `accounts.rs` should not import `cost`. Exact-size allocations (attestation embed, subject index list, finish bitmaps) can stay on `StepTables::tally_bytes`. A forgotten site then becomes "used the wrong vec" instead of "CI fails if we remembered to add a frontier row."

### Next

**4. Heap is the one ceiling checked after the allocations it claims to prevent.** The PR says each wall is typed "at the step that crosses it where possible." Steps and instruction-trace already do that in `commit_step`. Packet size needs `finish` (full serialization). Build-heap does not: `requested_heap_bytes` is live during the loop and is only tested in `finish`.

Checking only at finish means (a) over-budget shapes still allocate the whole execution before getting `ExceedsBuildHeapBudget`, and (b) a shape that exhausts the real 32 KiB *during* the loop still dies with no error — finish never runs. If `requested_heap_bytes > BUILD_HEAP_BUDGET_BYTES` after each step (or inside the tally helper), the fourth ceiling matches the other three. Packet + invoke remain a finish check; build-alone already exceeding 24 KiB implies the total will too.

**5. `cost.rs` mixed the public DTO with the Vec-growth model.** App programs need `FheExecutionCost`, the ceiling constants, and `instruction_trace_floor`. They do not need `tally_push`, `grown_capacity`, `minimum_nonzero_capacity`, or `pushes_from_empty`. Split: `cost.rs` stays the numbers an app budgets against; a crate-private `heap_tally.rs` owns growth + `invoke_table_heap_bytes`.

### Smaller, still real

- **Stale admission docs.** `builder.rs` still says the budget is "build + packet" and the 8 KiB reserve covers "account resolution." `heap_budget.rs` still says resolve vectors "sit on top" as slack. `finish` now charges invoke tables; the reserve is only Anchor + app. The comments describe the previous round.
- **`binding()` / `into_binding()` clone.** Two functions, one invariant. `binding(&self)` should clone then call `into_binding`, or both go through one helper.
- **`TooManySteps` still checked three times** — every op method, `commit_step`, and `finish`. `commit_step` is now the canonical gate. `finish`'s check is dead if `commit_step` holds.

### Keep

- One step ceiling instead of three copies. That deletion is the judo this branch was for.
- The counting-allocator keystone, plus the invoke measurement under `cpi`. Keep that proof; change how production pays the tally so the proof has fewer sites to police.
- Trace model tied to measurement (21 vs 20 creates). That coupling belongs next to the extracted sweep, not in `host_mollusk.rs`.
- Invoke consuming the execution, PDA derived once on `EncryptedValueId`, token signer seeds borrowed on the stack. Direct, boring, and they show up in the CU snapshots.

### Minimum to clear the bar

Extract the boundary sweep from `host_mollusk.rs`. Get `builder.rs` back under 1k by moving named ops out. Stop threading a tally into `promote` — own growth in one type. Check build-heap incrementally so the fourth ceiling matches the story. Split `cost.rs` so apps do not import a bump-allocator model.

Then this is an admission contract with a home, not a contract painted onto four files.

solana/crates/zama-fhe/src/builder.rs
```diff
@@ -6,15 +6,34 @@
...
-//! `MAX_FHE_EXECUTION_STEPS` has to be built off-chain or by a program bringing its own allocator.
-//! `heap_budget.rs` measures all of it, and [`MAX_ON_CHAIN_EXECUTION_STEPS`] enforces it: a program that
-//! keeps adding steps past the budget is told so, instead of being aborted by the allocator with no
-//! error of its own.
+//! step costs a few hundred heap bytes, and the step-bounded tables reserve their bound up front
+//! (a fixed ~10 KB every build pays) so growth never strands outgrown buffers on the
+//! never-freeing bump region (DD-046: the heap is fixed at 32 KB). The instruction pays out of
+//! that region twice — once building, once serializing the packet in `FheExecution::invoke`.
+//!
+//! An execution `build` returns is one the transaction can actually land. Four ceilings make
```
Eikix: [AUTOMATED] This lead sentence — and INVARIANT #54's `[HOLDS]` opener — still say an execution `build` returns is one the transaction can actually land. That is not true for the shared-audience host-sweep shape.

`dictionary_index` interns the same 8 subject pubkeys once, so 20 eight-subject `make_public` creates are expected to pass `FheExecution::build()` (trace floor 63; well under the app-heap budget that only rejects *distinct* 20×8). `subject_heavy_public_creates` is that shared-audience shape and heap-aborts at 16 (`max_ok=15`). Deleting `MAX_ON_CHAIN_EXECUTION_STEPS=16` newly admits 17–20 of those on-chain (`TooManyStepsForDefaultHeap` used to stop the 17th). Sixteen of this shape already aborted untyped before this PR; 17–20 is the new band.

The later #54 / `cost.rs` paragraph already records this as measured-not-typed. That does not make the headline true. Please qualify this lead and retag or split #54 so the public contract matches the 15-of-20 exception — `[GAP]`, or an explicit split between the four app-side ceilings and the host CPI frame. A typed host-frame cap is not required in this PR.

`max_creates_the_builder_admits` still only checks `instruction_trace_floor`, not `FheExecution::build()`, so it does not pin this.

Confidence: 90/100

solana/runtime-tests/tests/host_mollusk.rs
```diff
@@ -2548,6 +2571,811 @@ fn mollusk_fhe_execute_created_public_heap_boundary() {
...
+    let sweep = sweep_step_boundary(1, |steps| {
+        subject_heavy_public_creates_case(steps, Pubkey::new_from_array([0x3B; 32]))
+    });
+    assert_swept_boundary("fhe_execute_boundary/subject_heavy_public_creates", &sweep);
+    // On this axis pair the wall is the HOST's own CPI frame — the created accounts' subject
+    // tables and the public-outputs event payload grow with creates x subjects-per-output —
+    // and it sits well below the trace cap, which the one-subject sweeps could not see. The
+    // builder models only the app's CPI frame (build + packet + invoke tables), and no
+    // app-side number can see this driver: with a shared audience the dictionary interns
+    // eight subjects once, so the builder admits the trace-capped twenty such creates while
```
Eikix: [AUTOMATED] This comment asserts the builder admits the trace-capped twenty shared-audience creates, but the test only pins `limited_by == "heap"`. The sibling sweeps (`all_private_creates`, `all_created_public`) assert builder admission against the measured host wall; this one does not.

`max_creates_the_builder_admits` walks `instruction_trace_floor` only — it never calls `FheExecution::build()` — and the app-side frontier uses *distinct* per-output subjects, so nothing currently proves the shared-audience 8-subject `make_public` shape is `Ok` at 15 / 16 / 20. If the heap tally later admits 16+, this snapshot stays green while `build()` still returns a host-aborting execution.

Please add the same `builder_admits <= sweep.max_ok` (or an explicit `FheExecution::build()` specimen of this shared-audience shape) so the 15-of-20 hole cannot drift.

Confidence: 90/100

Eikix (commented):
[AUTOMATED]

## Code Review

Checked for runtime bugs, guideline compliance, error handling, and Solana protocol compatibility (host `fhe_execute` / `zama-fhe` admission / dep-chain).

No runtime bugs found in the changed code. Typed `finish()` ceilings fail closed; decode-once cache invalidates on write; invoke consumes the execution once; in-tree callers are lockstep with `invoke(self)` and the deleted 16-step cap.

One contract/docs issue below: the public landability claim is still categorical while this PR newly admits a host-heap-aborting shape. Host wire ABI is otherwise unchanged (step cap was already 32).

Eikix:
[AUTOMATED]

## Code Review

Checked for runtime bugs, guideline compliance, error handling, and Solana protocol compatibility (host `fhe_execute` / `zama-fhe` admission / dep-chain).

No runtime bugs found. Typed `finish()` ceilings fail closed; decode-once cache invalidates on write; invoke consumes the execution once; in-tree callers are lockstep with `invoke(self)` and the deleted 16-step cap. Host `fhe_execute` wire ABI is unchanged (the 32-step ceiling was already there).

### Issue: landability contract is still categorical

`builder.rs` and INVARIANT #54 still say an execution `build()` returns is one the transaction can land. That is not true for the shared-audience host-sweep shape.

- Shared-audience 8-subject `make_public` creates intern once, so 20 of them are expected to pass `FheExecution::build()`. The host sweep (`subject_heavy_public_creates`) dies at 16 (`max_ok=15`).
- Deleting `MAX_ON_CHAIN_EXECUTION_STEPS=16` newly admits **17–20** of that shape on-chain. Sixteen already aborted untyped; 17–20 is the new band.
- `cost.rs` and the later #54 paragraph already record this as measured-not-typed. That does not make the headline true.
- `max_creates_the_builder_admits` still only checks `instruction_trace_floor`, not `FheExecution::build()`, so it does not pin this. The sibling sweeps assert `builder_admits <= sweep.max_ok`; this one only pins `limited_by == "heap"`.

Please qualify the `builder.rs` lead and retag or split #54 so the public contract matches the 15-of-20 exception (`[GAP]`, or an explicit split between the four app-side ceilings and the host CPI frame). A typed host-frame cap is not required in this PR. Add a real `FheExecution::build()` specimen of the shared-audience shape (or the same builder-vs-host assertion the other create sweeps already have).

Inline notes: `solana/crates/zama-fhe/src/builder.rs:14`, `solana/runtime-tests/tests/host_mollusk.rs:3111`.

Eikix:
[AUTOMATED]

## Thermo-nuclear code quality review (round 3)

**Request changes.** The last round put the contract in the right files. Admission is still a remember-to-call protocol sitting on those files. Behavior looking correct is not the bar.

Commits 16–21 answered the previous review’s file-split demands. Do not reopen those:

- `builder.rs` 1095 → 408; named ops in `ops.rs`
- capacity sweep extracted to `fhe_execute_boundary.rs`, driven from one shape table
- `cost.rs` public numbers vs `heap_tally.rs` crate-private growth model
- invariant #61 + both-sides pin of the 15-of-20 host-frame gap

What should not survive is the claim that `TalliedVec` made forgotten sites unrepresentable. It made forgotten *accounting* unrepresentable for the intern tables. Admission is still a ritual.

### Blockers

**1. Heap admission is still a protocol. One `HeapBudget` so growth fails closed.**

`TalliedVec::push` always succeeds. It records growth. Every allocation site must still remember to admit first:

- intern tables: `ensure_headroom(vec.next_push_request())?` then `push` (`lower.rs:174-175`, `208-209`, `221-222`)
- exact-size: `tally_bytes` then `Vec::with_capacity` (`lower.rs:313-314`, `ops.rs:633-634`)
- nested purposes: `tally_push` on a raw `Vec` (`lower.rs:164-165`)
- the documented exception: `vec![purpose]` allocates *before* admission (`accounts.rs:79`), charged after the fact (`lower.rs:123-130`)

The budget inequality is implemented **five times**, with **two different totals**:

| Site | File | Total |
|---|---|---|
| `StepTables::ensure_headroom` | `lower.rs:93` | `stable_bytes + borrowed tables + upcoming` |
| `ensure_build_headroom` | `ops.rs:572` | `requested_heap_bytes() + pending` |
| `commit_step` belt | `builder.rs:255` | `stable_table_bytes + lowering.requested_bytes()` |
| `finish` bitmaps | `builder.rs:340` | `requested_heap_bytes() + bitmap_bytes` |
| `finish` packet+invoke | `builder.rs:383` | `build + packet + invoke` — the one that belongs |

The two pre-emptive formulas agree only because `ensure_build_headroom` is never called while `StepTables` is live. That is a coupling, not a model. The post-step belt’s own comment admits it cannot fire if every site is admitted — and if a site forgot, the never-freeing allocator already served. The belt does not protect the region. `heap_budget.rs` is what polices forgotten sites.

`ops.rs:8-11` says nothing in that file mutates builder tables except through `commit_step`. Then `verified_input` and `collect_reduction_operands` do. `lower.rs:9-11` says `accounts.rs` has no allocator accounting. Then it has the allocation and none of the accounting.

**Judo.** One `HeapBudget` owned by the builder:

```rust
impl<T> TalliedVec<T> {
    fn try_push(&mut self, budget: &mut HeapBudget, value: T) -> Result<()>;
}
impl HeapBudget {
    fn alloc_vec<T>(&mut self, len: usize) -> Result<Vec<T>>;
    fn admit(&mut self, upcoming: usize) -> Result<()>; // finish charges packet+invoke here too
}
```

Then `ensure_headroom`, `ensure_build_headroom`, `next_push_request` at call sites, and the `commit_step` heap `if` all disappear. Packet and invoke still land at `finish` — they are not knowable per step — but they go through the same `admit`. A forgotten site does not compile.

Same move collapses `purposes: Vec`. There are three `ExecutionAccountPurpose` variants. `[ExecutionAccountPurpose; 3]` (or a `u8` bitset) deletes the pre-admission `vec![purpose]`, the nested `tally_push`, and the last reason `tally_push` is `pub(crate)`.

**2. `heap_budget.rs` 208 → 1018.** This series crossed 1k in `2da0d66f7b` when the never-crosses test (a real proof, 121 lines) landed on an already-898-line kitchen sink. `#[cfg(test)]` is not a waiver.

The file currently owns four jobs: counting allocator, 122-cell specimen grid, proofs, and dead printers (`measured_shapes`, `print_measurement_table`, `debug_tally_breakdown`).

Delete the printers first (~90 lines; they are a subset of `frontier_shapes` plus an ignored debug leftover). Split the rest into an integration-test crate so the counting allocator does not wrap `tests.rs`: `harness.rs` / `shapes.rs` / `proofs.rs` / `frontier.rs`. Always-on keystone walks ~10 named specimens that hit each allocation site; the 122-cell grid stays `#[ignore]`. `the_shapes_past_each_ceiling_are_rejected` already pins the typed walls.

### High

**3. `host_mollusk.rs` leftover capacity instrument.** `MAX_CREATED_PUBLIC_CREATES_ON_DEFAULT_HEAP = 20`, `mollusk_fhe_execute_created_public_heap_boundary`, and a 20-line section header for a matrix that no longer lives here (`host_mollusk.rs:2369-2432`). The snapshot already pins `max_ok=20 / first_fail=21 / limited_by=heap`. Delete the header and the heap-boundary test. Keep the one-CPI lifecycle test; derive 20 from the snapshot or `max_creates_the_builder_admits`, not a parallel constant. Two sources of truth for the same wall is the failure mode the extraction was supposed to remove.

**4. `collect_reduction_operands` is a second admission machine in the op catalog** (`ops.rs:588-613`), with `explicit_heap_bytes +=` on three exits, then `sum`/`is_in` allocate a second table inside `commit_step`, copy-pasted (`ops.rs:632-637` vs `675-679`). `heap_budget.rs:708` still says those tallies live in `builder.rs`. Consume the iterator inside `commit_step` through `HeapBudget::alloc_vec` / `try_push`. One `StepLowering::operand_table` helper. Harvest-on-drop so the three `+=` arms die. The size_hint clamp stays — a hint is not a length.

**5. `cost_snapshot.rs` 244 → 470.** Boundary entries embed a full `Cost` (including CU) at the wall; ceilings snapshot asserts essay notes verbatim. A 200 CU jitter or a wording tweak fails the capacity golden. Boundary entry is `{max_ok, first_fail, limited_by}`. Ceiling entry is `{value, extendable}`. Notes belong in module docs.

### Do not reopen

`commit_step` as the only step-mutation path. In-place rollback (make it RAII on `StepTables::Drop` so the three call-sites stop being a protocol; do not delete it — it undoes interned tables, not heap). Builder `'id` identity. Four typed ceilings. Packet counted through a sink writer. Invoke charged from shape. `PersistentOutput::validate` without allocating. Consuming invoke. PDA derived once. Decode-once host cache. Named op methods stay named — marker traits are the API.

Invariant #14 still says a 32-step execution fits a 1,232-byte packet (`INVARIANTS.md:338`). #54 retired that. Docs pass.

### Minimum to clear the bar

One `HeapBudget` so growth cannot forget to admit — `try_push`, `alloc_vec`, purposes as an array. Split `heap_budget.rs` under 1k (delete the dead printers first). Remove the leftover capacity instrument from `host_mollusk.rs`. Then this is an admission type, not an admission protocol painted onto five comparison sites.

PR #3504, opened 2026-08-12
fix(coprocessor): sns-worker, improve S3 migration, metrics, logs, concurrency

rudy-6-4:


Eikix (approved):

Approving with the following advisory concern for the author to assess:

`SKIP LOCKED` can return an empty batch while another worker owns unfinished rows, but the startup loop treats empty batches as completion or exhausted retries.

This lets `before-and-quit` exit successfully before migration is complete. If the other worker crashes, unmigrated rows remain. Locked retry rows can also produce premature failure.

Please distinguish temporarily unavailable work from completed/exhausted work, and wait while pending or retry-eligible rows remain. Add a test through the real migration loop: hold a row lock, verify the runner does not finish, then release it and verify recovery.

Validated the SQL behavior with two PostgreSQL sessions and traced the startup decision; the full Rust/S3 scenario was not reproduced. The new two-worker test uses SQL helpers and does not exercise this completion path.

tawadaa (approved):

rudy-6-4:
> 
> Approving with the following advisory concern for the author to assess:
> 
> `SKIP LOCKED` can return an empty batch while another worker owns unfinished rows, but the startup loop treats empty batches as completion or exhausted retries.
> 
> This lets `before-and-quit` exit successfully before migration is complete. If the other worker crashes, unmigrated rows remain. Locked retry rows can also produce premature failure.
> 
> Please distinguish temporarily unavailable work from completed/exhausted work, and wait while pending or retry-eligible rows remain. Add a test through the real migration loop: hold a row lock, verify the runner does not finish, then release it and verify recovery.
> 
> Validated the SQL behavior with two PostgreSQL sessions and traced the startup decision; the full Rust/S3 scenario was not reproduced. The new two-worker test uses SQL helpers and does not exercise this completion path.

Yes, that can happen near the end of the migration. In that case it degenerates to 1 worker only, but the last worker can retry the rows that were unfinished, so  it is alone to finish doing retries on the last batch.

PR #3506, opened 2026-08-13
chore(sdk): dedupe base64 WASM payloads across cjs/esm builds and trim npm package weight

geoxel:
## Summary
- The multi-MB base64-encoded WASM payloads (`tfhe_bg.wasm.base64.js`, `kms_lib_bg.wasm.base64.js`) were being duplicated once per build target (`src/_cjs/wasm`, `src/_esm/wasm`) even though they're byte-identical regardless of module format. Both output trees now dynamic-import the single checked-in copy under `src/wasm/<lib>/v<version>/` instead of carrying their own copy — `codegen-loaders.mjs` gains a `wasmBase64DirPrefix` option so the generated loader can point at a shared relative path, and `wasm-build-common.mjs`/`build-cjs-wasm.mjs` exclude these files from the per-target copy/transpile accordingly.
- Sourcemaps are dropped for these payloads in the CJS build (they're data-only, so a sourcemap just doubles their published size) while linked sourcemaps are kept for actual code.
- `src/package.json`'s `files` glob now excludes `wasm/tfhe/v*/**` and `wasm/tkms/v*/**` wholesale, then re-includes just the shared base64 payload per version — and excludes it again for `*-dev` versions, since npm's `files` glob can't read `versionsManifest.js`'s `tags` at pack time to know which versions are dev-only.

### DIFF
--- a/sdk/js-sdk/scripts/build/build-cjs-wasm.mjs
+++ b/sdk/js-sdk/scripts/build/build-cjs-wasm.mjs
@@ -12,6 +12,7 @@ import {
   createWasmBuildContext,
   generatedWasmLoaderRels,
   generatedWasmRels,
+  isWasmBase64DataRel,
   logWasmBuildSummary,
   removeUnexpectedTscJsOutputs,
   writeGeneratedWasmArtifacts,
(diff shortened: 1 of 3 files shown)


nboisde (approved):
LGTM

geoxel:
@mergifyio queue

geoxel:
@mergifyio queue

geoxel:
@mergifyio queue

PR #3509, opened 2026-08-13
chore(library-solidity): cherry pick polygon support from v0.13.x version

nboisde:


library-solidity/config/ZamaConfig.sol
```diff
@@ -28,29 +35,47 @@ library ZamaConfig {
         }
     }
 
-    function getConfidentialProtocolId() internal view returns (uint256) {
+    function getEthereumCoprocessorConfig() internal view returns (CoprocessorConfig memory config) {
```
rudy-6-4, line 38: I would just filter on chain id and use the generic function for details (or vice-versa). Otherwise it double work to maintain both level ?
Eg
```
if ! (block.chainid == 1  || block.chainid == 11155111  || ... ) {
     revert ...
}
return getCoprocessorConfig
```

rudy-6-4 (approved):

nboisde:
@mergifyio queue

PR #3523, opened 2026-08-14
feat(relayer): host-generic Solana user-decrypt (sRFC-38), retiring the ed25519 form

KitHat:
Top of the host-generic user-decrypt stack: the relayer's sRFC-38 attestation envelope maps to the host-generic gateway call, retiring the ed25519 form.

The permit is pre-checked with the same crate the connector decodes with, so the relayer never accepts what the connector would refuse on form, and always declares the ACL-scope length honestly.

### Testing

- relayer: 542/547 (five spawn the full service and need the dev stand), incl. calldata slot-placement and byte-parity with the connector codec.

### Stack

| # | PR | scope |
|---|----|-------|
| 1 | #3524 | chore(kms-connector): proof carrier gets one owner |
| 2 | #3525 | permit container + gateway overload + connector wiring |
| 3 | 👉 this | relayer |
| 4 | #3549 | sdk client path + consumers |


Eikix:
## Automated review

This review covers [#3524](https://github.com/zama-ai/fhevm/pull/3524), [#3525](https://github.com/zama-ai/fhevm/pull/3525), and [#3523](https://github.com/zama-ai/fhevm/pull/3523) as one stack.

The review compares the stack with the EVM outcome, the Solana design discussions, the glossary, and KMS #751.

### Required outcome

The stack must provide these results:

- [ ] The user signs one reusable Solana permit.
- [ ] The permit contains no handles or MMR proofs.
- [ ] Each request selects its handles after permit creation.
- [ ] Each request supplies current account data or an MMR proof.
- [ ] The Connector verifies all request evidence against Solana state.
- [ ] A compatible second request reuses the permit without another wallet approval.
- [ ] The KMS response binds the exact request and recipient.
- [ ] The browser checks the configured KMS signer list.
- [ ] The current Gateway route remains usable.
- [ ] The design supports centralized and threshold KMS modes.

### Design alignment

The following choices match the agreed direction:

- The Solana permit uses a readable Ed25519 signing format.
- The permit separates reusable user consent from individual handles.
- The signed domain list matches the EVM `allowedContracts` outcome.
- The Connector checks Solana authorization.
- KMS Core does not parse Solana accounts or MMR proofs.
- Historical handles use MMR proofs.
- The 33-handle limit preserves one consistent account observation.
- The linker binds the program, chain, user, KMS context, handles, and transport key.
- The transport key uses the 869-byte safe-serialized form.
- Gateway remains in this stack. Gateway removal belongs to later work.

The later KMS decision replaces the gist’s older 800-byte transport-key text.

### 1. Blocker: the Gateway key can differ from the signed transport key

The Gateway stores `publicKey`. It later uses this key to check each KMS response.

The Connector uses the signed permit transport key. It does not compare the two keys.

A mismatch causes this sequence:

1. The Gateway accepts the request and collects the fee.
2. The Connector accepts the signed permit.
3. The KMS encrypts and signs for the permit transport key.
4. The Gateway checks the response against its different `publicKey`.
5. The Gateway rejects every valid response.

The current mismatch test accepts this case. The test protects confidentiality, but the request cannot complete.

Required changes:

- [ ] Reject a request when `publicKey` differs from the permit transport key.
- [ ] Perform this check before RPC reads and KMS work.
- [ ] Keep the permit transport key as the KMS seal target.
- [ ] Change the mismatch test to expect rejection.
- [ ] Add a test that submits the KMS response to the Gateway.

### 2. Blocker: the stack does not include the complete browser path

The JS SDK still sends `solana-ed25519-user-decrypt-v2`. The new Relayer accepts only `solana-srfc38-user-decrypt-v1`.

The Connector still uses a KMS version from before KMS #751. It does not send `solana_verifying_program_id`.

The browser WASM still uses the old linker and response checks.

The demo uses this JS SDK. Therefore, the demo cannot use the new permit path.

Required changes:

- [ ] Update the JS SDK to create the new Solana permit.
- [ ] Send the new request type and per-handle evidence.
- [ ] Support multiple handles.
- [ ] Cache and reuse a valid permit.
- [ ] Update the Connector to KMS #751.
- [ ] Send `solana_verifying_program_id` to KMS Core.
- [ ] Update the local KMS Core image.
- [ ] Rebuild the browser WASM from KMS #751.
- [ ] Load the KMS signer list from trusted configuration or `KmsContext`.
- [ ] Never load trusted KMS signers from a KMS response.
- [ ] Update the demo to use only the public JS SDK.

If another pull request owns these changes, link it before this stack merges.

### 3. Decision: the signed domain list is required, but its separate count is not

The permit contains a signed domain list. This list matches the EVM `allowedContracts` result.

The separate `allowedAclDomainKeyCount` adds no authorization.

The Gateway cannot read the domain list because `hostPayload` is opaque. The count lets the Gateway apply the maximum before its fee.

The Relayer and Connector already validate the signed list. The Connector also compares the separate count with the signed list length.

Recommended changes:

- [ ] Remove `allowedAclDomainKeyCount` from the Gateway call, event, and database.
- [ ] Let the Relayer reject an oversized list before submission.
- [ ] Let the Connector perform the final signed-list check.

If Gateway owners require the on-chain pre-fee check:

- [ ] Rename the field to `allowedDomainCount`.
- [ ] Document the field as temporary Gateway data.
- [ ] Keep the Connector length check.
- [ ] Remove the field with Gateway user decryption.

The separate count must not become part of the Solana permit.

### 4. Decision: the Gateway layer is more generic than the roadmap requires

Only Solana uses the new Gateway entry. Gateway removal should occur before another non-EVM host integration.

The generic host layer adds `hostKind`, database fields, dispatch code, and unclear names.

Recommended changes:

- [ ] Use `solanaRequest` or `solanaUserDecryptRequest` instead of `hostPayload`.
- [ ] Remove `hostKind` unless another host will use this Gateway entry.
- [ ] Keep the existing Gateway overload style.
- [ ] Do not rename the event only because generated code calls it `_4`.

The gist proposed a generic Gateway form. It also left the final Gateway form open for Gateway owners.

### 5. Required: use one Rust codec for the Solana request

The Relayer and Connector define the same Borsh layout separately.

Shared vectors detect current differences. They do not make future field changes compile together.

Required changes:

- [ ] Define one Solana request type and codec.
- [ ] Put the codec in shared Solana protocol code.
- [ ] Use the codec from both the Relayer and Connector.
- [ ] Keep the JS implementation independent.
- [ ] Keep fixed cross-language vectors.

Use `hostPayload` only for an external Gateway field if the ABI keeps that name.

### 6. Required: check every duplicated Gateway field

The Gateway call and the signed Solana request contain several copies of the same facts.

The Connector compares handles, validity, domain count, and KMS routing. It does not compare the transport key.

Required changes:

- [ ] List every duplicated field in one comparison function.
- [ ] Compare every duplicated field before authorization.
- [ ] Add one negative test for each mismatch.
- [ ] Delete any duplicate that the Gateway does not need.

### 7. Required: keep MMR proofs outside the permit

The user does not sign an MMR proof.

The user signs a reusable permit. Each request later supplies its current MMR proof as untrusted evidence.

The Connector checks the proof against the encrypted value account.

Required changes:

- [ ] Keep handles and MMR proofs outside the signed permit.
- [ ] Check each MMR proof against the current encrypted value account.
- [ ] Let a refreshed proof retry a request without another wallet approval.
- [ ] Do not discard a refreshed proof during request deduplication.

Including the proof in request deduplication does not make the proof part of the user signature.

### 8. Required: use the normative glossary

The new code calls the wallet-signed Solana permit an attestation. The glossary reserves attestation for a Coprocessor statement.

The code also calls the requester a subject. These names describe different roles.

Required changes:

- [ ] Use `permit` for the wallet-signed object.
- [ ] Use `requester` or `signer` for `user_pubkey`.
- [ ] Use `subject` for the person whose encrypted value is requested.
- [ ] Replace new uses of `owner` with `subject`.
- [ ] Use `encrypted value ID` instead of `value_key`.
- [ ] Keep `attestationType` only at an existing external JSON boundary.
- [ ] Update the OpenAPI request description.

### 9. Decision: the Relayer does not check the Solana signature before Gateway submission

The EVM Relayer rejects an invalid user signature before it submits a Gateway transaction.

The Solana Relayer checks request form, but the Connector checks the signature after Gateway submission.

The Connector remains the final authorization check. The difference only affects Relayer cost and early failure.

Recommended change:

- [ ] Add the local Ed25519 permit check if the shared permit crate makes it small.

If this change adds substantial work:

- [ ] Record the Relayer cost risk as an accepted POC limit.
- [ ] Keep the Connector signature check mandatory.

### 10. Blocker: the clean stack does not build

The Relayer Dockerfile does not mount the new Solana permit and ACL path dependencies.

The image build cannot resolve these crates.

Required changes:

- [ ] Mount both Solana crates in the Relayer Dockerfile.
- [ ] Build the Relayer image from a clean repository.
- [ ] Start the complete local stack from a clean state.

### 11. Required: clear the current CI failures

The stack has failures that come from these changes.

Required changes:

- [ ] Run the Rust formatter.
- [ ] Update the Rust SDK to use the correct legacy Gateway overload.
- [ ] Add a selector test for the Rust SDK call.
- [ ] Regenerate all Gateway bindings and selector files.
- [ ] Remove the retired Solana Gateway call from the upgrade task.
- [ ] Fix the two Relayer retry tests.
- [ ] Exclude the coverage artifact from the EIP-170 size assertion.
- [ ] Keep the normal optimized EIP-170 size test.
- [ ] Make every required check pass.

### End-to-end acceptance criteria

- [ ] A Phantom wallet signs one Solana permit.
- [ ] A current-handle request returns the correct plaintext.
- [ ] A second compatible request reuses the permit.
- [ ] The second request does not request another wallet approval.
- [ ] A multi-handle request succeeds.
- [ ] A historical request succeeds with an MMR proof.
- [ ] A refreshed MMR proof can retry an unavailable request.
- [ ] An invalid domain fails.
- [ ] An invalid subject fails.
- [ ] An expired or invalidated permit fails.
- [ ] A transport-key mismatch fails before KMS work.
- [ ] An unknown KMS signer fails.
- [ ] The local demo succeeds with centralized KMS.
- [ ] The threshold KMS tests use the same request and response rules.
- [ ] The Gateway accepts each valid KMS response.
- [ ] The browser checks each response before it shows plaintext.
- [ ] The complete flow passes on a clean local stack.

### Scope limits

- [ ] Do not add or harden delegated decryption in this stack.
- [ ] Record existing delegation issues as follow-up work.
- [ ] Do not design the future Gateway-free route in this stack.
- [ ] Do not add support for another host chain.
- [ ] Do not remove MMR proof support.
- [ ] Do not remove the atomic account-read requirement.
- [ ] Do not redesign Gateway event overloads without a functional need.
- [ ] Do not add production hardening that this POC does not require.

### Design references

- [Solana permit and linker gist](https://gist.github.com/KitHat/296fdc99aaf2a49d6703401c61f4731b)
- [Solana Guild transcript, July 23](https://docs.google.com/document/d/1yOxqMaIgQR6mPE_gmZ-Phf7lTkELTjnwTwqIgaWqdUw/edit)
- [Solana Guild notes, July 30](https://docs.google.com/document/d/1SFWPuHpbaS_d3P8vBCKbA07500mcKSG6Jat5HPeoD6g/edit)
- [Solana Guild transcript, August 6](https://docs.google.com/document/d/1OdAm9K3SjrhPrffbSgg8YOR8hr1sTJ--IUQHoR2amW8/edit)
- [Solana Guild transcript, August 13](https://docs.google.com/document/d/1s2O5n5PZVRl09Dd-ynia45oahAWFAx5qlrCKNClqFCE/edit)

KitHat:
All eleven addressed. CI green on `e48952dc2`.

**1. Transport key** — fixed as part of item 6; it was one of the duplicated fields, not a
separate case. Two corrections: nothing leaked (the seal target was always the signed key), the
break was completability — valid responses refused with the fee spent. And it cannot run before
the fee: the blob is opaque to the contract. It now runs before any RPC or KMS work.

**2. Browser path** — out of this stack, ships as a separate SDK PR. The connector→#751 half
cannot land before kms#751 reaches us. See Q1.

**3. `allowedAclDomainKeyCount`** — removed, not renamed. Sharper than "redundant": the bound
protected nobody. A lying relayer declares `0` and burns the fee anyway; an honest one cannot
overflow, because `AclDomainKeys::new` inside `PermitFields::decode` caps at ten and both sides
call that decode. The relayer-side rejection you ask for is already there by construction.

**4. Genericity** — agreed on the premise. `hostKind` gone, `hostPayload` → `solanaRequest`, cap
unconditional, DB discriminator now `solana_request IS NOT NULL`. Dropping the two admission
checks returned 417 bytes of EIP-170 headroom, so the optimizer went back to `runs = 200`.
On the event: nothing was renamed anywhere in the stack — the new one is a Solidity overload
sharing the name.

**5. One codec** — `solana/crates/zama-solana-request`, both sides call it. Your point held
exactly: a field added to the permit failed on both sides, a field added to a handle entry did
not, because the entry types were distinct. Two more duplicated constants folded in (33-handle
cap, 64-sibling proof bound). Vectors kept — they now pin one Rust impl against borsh-js.

**6. Duplicated fields** — one `check_event_permit_parity` before authorization; four fields
survive items 3 and 4. Only `publicKey` lacked a negative test. The gap existed because the
checks lived in three modules in three styles, so nothing answered "are they all checked".

**7. MMR proofs** — already held, no change. Outside the signed permit and documented as
unsigned evidence; verified against live snapshot peaks (`handle_binding.rs:80-128`); a refreshed
proof survives dedup because the key hashes the whole blob (`user_decrypt_request.rs:136-145`).
Stricter than asked: inclusion is verified before age, so a valid older proof is not rejected
after every append.

**8. Glossary** — applied. `owner` → `subject` only where it means the pubkey whose encrypted
value is requested; Solana's account `owner` and EVM's `ownerAddress` keep their names.
`user_pubkey` documented as the requester, explicitly not the subject on a delegated entry.
`attestation` gone from Rust identifiers; `attestationType`/`attestedPayload` stay — the first is
the dispatch discriminator shared with the EVM envelope. Also: the Solana envelope had no OpenAPI
schema at all, now registered.

**9. Relayer signature check** — added, but the premise needs correcting, and the correct one is
stronger. The EVM relayer does not verify either (`validations.rs:353-354`). What protects EVM is
the contract's own check inside the transaction: a bad signature reverts and the revert takes the
fee with it. No contract can do that for a permit in an opaque field, so here the request was paid
for and then died in the connector — same class as item 1, not relayer cost. The connector's check
stays authorizing.

**10. Clean build** — fixed, and the same gap was in the connector's four Dockerfiles, which
copied two of the three shared crates.

**11. CI** — green.

PR #1782, opened 2026-01-14
feat(host-contracts): add rust bindings

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/371

PR #3527, opened 2026-08-14
refactor(solana): read host-listener contract bindings from the committed rust_bindings crate

Eikix:
## Why

Bringing the demo up from a clean tree ended with the dApp dead: Vite failed to resolve `@noble/curves/ed25519.js` from the symlinked SDK source and served nothing but 500s.

**Root cause:** the `solana` scenario's host-process step compiles the host-listener natively on the host, and `host-listener/build.rs` manufactured the hardhat artifact JSONs its `sol!` macros read by running `npm ci` + hardhat **inside `host-contracts/`**. Run from a workspace directory, npm reifies the shared root `node_modules` to the host-contracts graph alone — erasing the SDK's ~450 runtime dependencies that `clean-e2e.sh` installed, which every `@fhevm/sdk` symlink consumer resolves through the SDK's real path. (npm's hidden lockfile after a broken bring-up: 958 of 962 entries under `host-contracts/…`, no `@noble/*` anywhere; npm debug logs match the up's host-listener build to the second.)

**Why this crate was the odd one out:** `fhevm_host_bindings` (#1782, Jan 2026) landed two months *after* host-listener — every crate written since consumes the committed bindings (gw-listener, transaction-sender, consensus-detector, relayer, kms-connector), and nobody retrofitted host-listener. The clobber was invisible until #3492 made the root `node_modules` load-bearing at runtime.

## What

Finish adopting the #1782 pattern — no shims, direct imports of the generated names like every other consumer:

- **host-listener** (+ tfhe-worker tests/benches, stress-test-generator) imports `fhevm_host_bindings` with the real contract names: `TfheContract` → `FHEVMExecutor` (which it always was), `AclContract` → `ACL`, `BridgeContract` → `BridgeEvents`. `src/contracts/mod.rs` (the five inline `sol!` blocks) is deleted.
- **build.rs** loses `build_contracts()` entirely: `cargo build -p host-listener` no longer needs node/npm/hardhat on any machine and never mutates the checkout. Only the Solana IDL codegen and the crate-local test-contract compile remain. Bindings freshness stays guarded by the existing `make check-bindings` / `ci/contracts_bindings_update.py` drift check.
- **Dockerfiles**: the `contract_builder` npm stages in `host-listener/Dockerfile`, `stress-test-generator/Dockerfile`, and the host half of `Dockerfile.workspace` are gone; every fhevm-engine image gains a `COPY host-contracts/rust_bindings` line next to its existing `gateway-contracts/rust_bindings` one (cargo loads every workspace member's path-dep manifests).
- **Obsolete workarounds removed**: solana-e2e.yml's post-up `npm ci` re-install (the CI-side patch for this exact clobber) and its stale comment; `clean-e2e.sh` now just re-proves both runtimes import `@fhevm/sdk/solana` through the symlink after the up, as a loud regression tripwire.
- Two `warn!` sites that Debug-formatted the events enums log `topic0`/the raw log instead — the generated enums carry an empty extra-derives slot (no `Debug`). Adding `Debug` to the bindings codegen would be a nice follow-up but regenerates 140 files, so not here.

Net: **44 files, +469/−625**.

## Verified

- **Hermeticity**: with `host-contracts/artifacts` removed from the tree and no special env, the full `solana_host_listener` (grpc+reconstruct) build completes with **zero npm invocations** (`~/.npm/_logs` count and the hidden-lockfile mtime unchanged).
- `cargo check --workspace --all-targets` green across fhevm-engine; pre-commit fmt + clippy pass; `tsc --noEmit` clean in test-suite; `shellcheck` clean on clean-e2e.sh.
- Root cause confirmed live: on the demo boot that exhibited the bug, re-installing the SDK graph restored the dApp with no restart, after which every demo flow passed end-to-end with the burner wallet (deposit → fast-forward → redeem → cUSDC re-deposit, exact expected arithmetic).

## Follow-ups (not here)

- `NODE_PATH` override for the seed in `demo-up.sh` (papered over the same hole; harmless now).
- `--extra-derives Debug` in the bindings codegen.
- `Dockerfile.workspace`'s remaining gateway-contracts artifact compile looks unconsumed by any .rs — worth a look by the coprocessor team.


Eikix:
Replacement for this closed PR: #3554.

That restores the SDK workspace graph after `fhevm-cli up` in `clean-e2e.sh` instead of migrating host-listener onto rust_bindings. Bindings adoption can stay a separate coprocessor change (#3550).

PR #3528, opened 2026-08-14
feat(coprocessor): inject synthetic FHE ops in GCS host-listener

goshawk-3:
fixes https://github.com/zama-ai/fhevm-internal/issues/1914


coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -259,7 +262,43 @@ pub async fn ingest_block_logs(
     let mut allow_event_count: i32 = 0;
     let mut fhe_event_count: i32 = 0;
 
-    for log in &block_logs.logs {
+    // GCS only: on one designated block of each chain's dry-run window, append a
+    // deterministic synthetic transaction so a quiet chain can still anchor consensus. The
+    // logs go through the normal decode below, so `fhe_event_count`, `is_allowed`, the
+    // dependence chain and `schedule_order` are all derived by the production path. Blue
+    // never injects, so `public` stays untouched. See `database::synthetic_ops`.
+    let synthetic = if db.gcs_mode() {
```
PanGan21, line 368: Are you sure this is correct? Is `db.gcs_mode()`? This is not updated live. I think in the gw FHE ops pr you handle it better by using `AND version = STACK_VERSION`

coprocessor/fhevm-engine/host-listener/src/database/synthetic_ops.rs
```diff
@@ -0,0 +1,606 @@
...
+            + target_version.len()
+            + 16
+            + B256::len_bytes(),
+    );
+    hasher_input.extend_from_slice(SYNTHETIC_TXN_DOMAIN);
+    hasher_input.extend_from_slice(proposal_id);
+    hasher_input.extend_from_slice(target_version.as_bytes());
+    hasher_input.extend_from_slice(&chain_id.to_be_bytes());
+    hasher_input.extend_from_slice(&block_number.to_be_bytes());
+    hasher_input.extend_from_slice(block_hash.as_slice());
```
PanGan21, line 159: Why to use block hash? Similar [with this comment](https://github.com/zama-ai/fhevm/pull/3546#discussion_r3842318573) we might have an issue on reorgs

goshawk-3: Otherwise we could reach quorum while the operators are running non-canonical chains which sounds doubtful.

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
goshawk-3: `gcs_mode` isn't introduced in this PR and additionally the synthetic path is also gated on more than that flag.

The plan query requires an active GCS row for that specific host_chain_id, with status = 'in_progress' and state IN ('UpgradeActivated', 'DryRunStarted'), and injection fires only on the exact block start_block + 1. So a stale gcs_mode by itself could not cause an injection.

If you still think gcs_mode needs hardening, let's track it in its own issue with the details. It would apply to every existing caller, not just this change, so it's better scoped separately.

PanGan21: Agreed that `gcs_mode` predates this PR, but the public getter and both its uses are added here.

On the gate you are describing: I don't think those conditions exclude it. What I am thinking isn't a stale flag but a stale flag during the next upgrade. A green turned live listener isn't restarted at cutover, so `Database.gcs_mode` stays true while StackMode correctly flips to false. At the next proposal there is an active GCS row for that chain `in_progress` and this listener will ingest that chain's `start_block + 1` because so all three conditions are met and it injects, as the live stack.

I think adding `AND version = $STACK_VERSION` to the query also closes it and is a one liner if you'd rather keep the change minimal.

goshawk-3: If you agree, we can close this by addressing it correctly in: https://github.com/zama-ai/fhevm-internal/issues/1942

PanGan21 (dismissed):

goshawk-3:
@mergifyio queue

goshawk-3:
@mergifyio queue

goshawk-3:
@mergifyio queue

goshawk-3:
@Mergifyio unqueue

PanGan21 (approved):

goshawk-3:
@mergifyio queue

PR #3538, opened 2026-08-17
feat(coprocessor): inject synthetic GW input for GCS dry-run consensus

goshawk-3:
A quiet Gateway leaves the input-verification consensus track with nothing to anchor on, so the dry-run times out. The GCS gw-listener now builds one deterministic input proof itself at gw_start_block + 1.

Determinism comes from tfhe's build_with_proof_packed_seeded: one seed drives both the encryption noise and the ZK prover, so the blob is a pure function of (pks, crs, aux data, plaintext, seed). The seed is keccak over on-chain values only, so every operator produces byte-identical bytes, hence the same handles and state hash. The host chain is the lowest chain id in the proposal that is also in host_chains, so multi-chain proposals pick the same one everywhere.

Everything after the insert is the production path. Two exceptions:
- transaction-sender skips zk_proof_id >= SYNTHETIC_ZK_PROOF_ID_BASE, since no contract requested this input and publishing a response would revert.
- cutover deletes the row and its derived ciphertexts/input_handles before the merge, so synthetic data never becomes live.

Also fixes a GCS wake-up gap: CREATE TABLE ... LIKE INCLUDING ALL does not copy triggers, so inserts into gcs.computations and gcs.pbs_computations notified nobody and the tfhe-/sns-workers only woke on their polling interval (observed as a 60s stall). create_gcs_tables now recreates both notify triggers, and the DryRunStarted transition also fires work_available for work queued while the fleet was parked.

Moves the ZK aux-data layout into fhevm-engine-common so prover and verifier share one implementation.

fixes: https://github.com/zama-ai/fhevm-internal/issues/1522?issue=zama-ai%7Cfhevm-internal%7C1915


PanGan21 (dismissed):
LGMT!

PanGan21 (approved):

goshawk-3:
@mergifyio queue

PR #3542, opened 2026-08-18
fix(coprocessor): retire wave1 branch state

antoniupop:
This PR removes the first part of block-scope re-randomisation and ct materialisation that is no longer needed. We only remove all *_branch references from the code, but don't drop the tables until v0.16 to avoid having an upgrade scheduling constraint in v0.15.


coprocessor/fhevm-engine/.cargo/audit.toml
```diff
@@ -3,7 +3,9 @@
 
 [advisories]
 # The ignored vulnerability RUSTSEC-2024-0388 is due to sqlx-mysql which is not used
-ignore = ["RUSTSEC-2023-0071", "RUSTSEC-2025-0111"]
+# RUSTSEC-2026-0258: only reachable via h2 0.3 pulled by aws-sdk-s3
+# "test-util" (legacy hyper-0.14 test connector); not in production paths.
+ignore = ["RUSTSEC-2023-0071", "RUSTSEC-2025-0111", "RUSTSEC-2026-0258"]
```
antoniupop, line 8: Temporary to get CI working, this is tracked to be fixed up before v0.15 here https://github.com/zama-ai/fhevm-internal/issues/1921

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -1056,542 +761,6 @@ impl Database {
...
-        // `ciphertext_digest` row cancels a not-yet-sent
-        // `addCiphertextMaterial`; an already-sent one cannot be recalled and
-        // re-association after canonical re-inclusion re-sends it, which the
-        // contract's `CoprocessorAlreadyAdded` path treats as benign. The
-        // single DELETE .. RETURNING both removes the orphaned observations
-        // and captures the flag under the row lock, so an association racing
-        // finalization is either retracted or never happens.
-        let retracted_bridged = sqlx::query!(
-            r#"
-            DELETE FROM handle_bridged_events
```
PanGan21, line 1479: This is not branch table cleanup and i understand that it removes a bridge association when its destination block becomes orphaned. Since bridge association can happen before finality, removing this code can leave the copied ciphertext and digest behind.
Should we keep this bridge cleanup since it still seems useful?

antoniupop: You're right, this and also the delegate_user_decrypt / fallback_granted_events / bridge_handle_events deletes is legacy event-table retraction, not branch cleanup; it gotmixed up because this was fixed in the wave1 PR initially. I'll restore these four deletes together with their tests.

antoniupop: Fixed.

PanGan21:
Running e2e test suite from this branch: https://github.com/zama-ai/fhevm/actions/runs/32235461380
It will give more confidence on approving

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -1065,409 +780,14 @@ impl Database {
...
         // Confidential bridge: destination-side reorg retraction. The bridge
         // worker sets `is_associated` in the same transaction as the
         // ciphertext copy, so a flagged observation in an orphaned block is
         // the provenance of a materialization that must be retracted — or
         // re-attributed to a surviving sibling observation, see below (a
-        // fallback-materialized handle is never flagged and its state is
-        // compute-pipeline state, cleaned above). Deleting the copied
-        // `ciphertext_digest` row cancels a not-yet-sent
+        // fallback-materialized handle is never flagged and its synthetic
+        // compute rows are retained like all compute state). Deleting the
```
PanGan21, line 789: I think we should wait for the fallback event to be finalized before materializing it. 
If the event is later orphaned, this cleanup removes only the fallback but keeps its ciphertext and digest. I understand that this existing ciphertext will prevent the worker from copying the new ciphertext for the same handle. Maybe we should gate fallback materialization on finality/

antoniupop: Yes indeed, this is a consensus hazard, not just a stale row: fork visibility differs per copro, so whether the fallback trivial-encrypt or the real bridged copy wins, the ON CONFLICT DO NOTHING race is order-dependent per node. Retraction can't safely undo fallback materialisation, so gating synthesis on finality sounds good. Fallback is a recovery path, so the finality lag is acceptable. FIxing this, thanks!

antoniupop: Done.

PanGan21 (approved):

antoniupop:
@mergifyio queue

PR #3543, opened 2026-08-18
feat(host-contracts): discriminate handles by operand origin

antoniupop:
**Problem.** Handles are content-derived, so a replacement block sharing a parent and timestamp (L1 equivocation; routine unsafe-head reorgs on fixed-block-time chains like Base) can mint the *same* handle from
transactions that source its operands differently - persisted boundaries on one side, raw local intermediates on the other. The two sourcings legitimately produce different bytes (`decompress(compress(x))` is bit-inexact.

**Proposal.** The executor records every handle it derives in its own transient storage (ops, trivial encrypts, randoms - not verified inputs) and folds one *boundary bit* per operand into the minted handle's preimage:
1 iff the operand was NOT minted in this transaction, i.e. its only consumable representation is the canonical persisted form.

A bit-0 collision means both blocks produced the operand inside the consuming transaction - same operand handle, so by induction (bottoming out at ZK-pinned inputs and tx-scoped rand seeds) the same raw bytes. A bit-1 collision means both consumed the canonical form of the same persisted bytes. 

Measured cost (forge gas-report, versus the same contracts with the discriminant neutralised): +135 gas per mint and +108 per encrypted operand consumed, i.e. +351 on a binary op, +472 on a ternary, +1207 on a five-element `fheSum`, and nothing at all on `allow()`. The slot is the handle itself; see `_markMinted` for why namespacing is unnecessary in transient storage.


Eikix (commented):
Overall LGTM
We'd ideally need a review from protocol-contracts

rudy-6-4 (commented):
How can we ensure no minting is forgotten and no call to _oneOperandBoundaryBit is forgotten ? (utils  function or tests or analysis tool)
Could utils  function around abi.encodePacked forcing the call on _oneOperandBoundaryBit per input ct and doin gthe minting help ?

antoniupop:
> How can we ensure no minting is forgotten and no call to _oneOperandBoundaryBit is forgotten ? (utils function or tests or analysis tool) Could utils function around abi.encodePacked forcing the call on _oneOperandBoundaryBit per input ct and doin gthe minting help ?

That's a good point - I'd focus on tests, abi.encodePacked is variadic over mixed types, so a wrapper would need one overload per op shape... so we'd be back at square one. 
Fixing.

isaacdecoded (approved):
LGTM! 👍

antoniupop:
@mergifyio queue

PR #3546, opened 2026-08-18
fix(coprocessor): delete synthetic host ops at cutover

goshawk-3:
Synthetic host ops injected for dry-run consensus were never deleted, so at cutover they merged into public as live data and the green transaction-sender tried to publish digests for handles that exist on no chain.

The injector now stores its transaction hash in upgrade_state.synthetic_txn_hash. It cannot be recomputed at cutover, because the hash covers block_hash and that is not persisted unambiguously.

Cutover resolves the handles from computations.transaction_id, then deletes the ciphertexts, digests, pbs rows, ACL entries, computations and orphaned dependence chains in both legacy and branch forms before the merge, logging each statement with its row count.


coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -698,14 +699,46 @@ async fn synthetic_logs_for_block(
...
+            AND host_chain_id = $2
+            -- Pin to the proposal the hash was derived from. Without it, a proposal activated
+            -- between the SELECT above and this UPDATE would be stamped with a hash computed
+            -- from the previous one, and cutover would then delete nothing.
+            AND proposal_id = $3",
+    )
+    .bind(synthetic_txn_hash.as_slice())
+    .bind(chain_id_i64)
+    .bind(&proposal_id)
+    .execute(tx.as_mut())
```
PanGan21, line 753: We could check the rows affected here and warn or error if no rows were affected

coprocessor/fhevm-engine/db-migration/migrations/20260818120000_add_synthetic_txn_hash_to_upgrade_state.sql
```diff
@@ -0,0 +1,20 @@
...
+-- Used at cutover to delete the synthetic computations, their ciphertexts and their dependence
+-- chains before `gcs.*` merges into `public`. Without it, synthetic handles become live
+-- production rows and the newly-live green transaction-sender tries to publish ciphertext
+-- digests for handles that exist on no chain.
+--
+-- Nullable: BCS rows never carry one, and a GCS row has none until its window reaches
+-- `start_block + 1`. Cleared on rollback, because a re-attempt lands on a different block hash
+-- and therefore derives a different transaction hash.
+ALTER TABLE upgrade_state
+    ADD COLUMN IF NOT EXISTS synthetic_txn_hash BYTEA NULL;
```
PanGan21: Issue I see and a cheap way to fix it:

- GCS injects synthetic FHE work on one block of the dry run window
- The id of that work is the block hash
- That injection happens at start_block + 1, with no wait for finality.
- When the block is replaced by reorg, the listener injects again on the replacement which means different block hash  and a different id in the db, separate set of synthetic rows in the gcs schema.
- `synthetic_txn_hash` holds one value, so the second write overwrites the first. 
- Nothing else cleans it up. `retract_orphaned_event_state` doesn't delete orphaned fork
- At cutover we delete only the id we still remember. The forgotten set merges into public as live data

Quite edge case, but instead we could do the following:
- `ADD COLUMN IF NOT EXISTS synthetic_txn_hashes BYTEA[] NOT NULL DEFAULT '{}'`
- append instead of overwrite:
  `SET synthetic_txn_hashes = CASE WHEN $1 = ANY(synthetic_txn_hashes)
       THEN synthetic_txn_hashes ELSE synthetic_txn_hashes || $1 END`
  (the `ANY` guard keeps re ingests idempotent)
- at cutover, flatten every chain's array into `hashes`
- clear to `'{}'` instead of `NULL` in `delete_gcs_synthetic_ops` and `rollback_dry_run`

goshawk-3: Indeed. We should support an array of `synthetic_txn_hashes` per chain.

goshawk-3: Done in https://github.com/zama-ai/fhevm/pull/3595

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
goshawk-3: Done in https://github.com/zama-ai/fhevm/pull/3595

PanGan21 (approved):

goshawk-3:
@mergifyio queue

PR #3548, opened 2026-08-18
fix(library-solidity): add type and index checking in fromExternal

0xalexbel:


library-solidity/codegen/src/templates/FHE.sol-template
```diff
@@ -79,6 +79,12 @@ library FHE {
     /// @notice Returned if the sender is not allowed to use the handle.
     error SenderNotAllowedToUseHandle(bytes32 handle, address sender);
 
+    /// @notice Returned if the handle cannot be used as an external handle.
+    /// @dev Reverted when the handle's type byte does not match the expected `FheType`, or when the handle's
+    /// input index is `0xff`, a reserved value that never identifies a user input.
```
dartdart26: I would remove these such that we don't maintain them. In the future, we might change that.

library-solidity/package.json
```diff
@@ -45,7 +45,7 @@
     "@openzeppelin/contracts": "^5.0.1",
     "@openzeppelin/contracts-upgradeable": "5.0.2",
     "@openzeppelin/hardhat-upgrades": "3.3.0",
-    "@trivago/prettier-plugin-sort-imports": "^4.0.0",
+    "@trivago/prettier-plugin-sort-imports": "^5.2.2",
```
dartdart26: Is this a mandatory change in this PR?

library-solidity/codegen/src/templates/FHE.sol-template
```diff
@@ -90,6 +96,20 @@ library FHE {
...
+    /**
+     * @dev Checks if `handle` is of type `expectedType` and if its internal input index is valid
+     */
+    function _checkExternalHandle(bytes32 handle, FheType expectedType) private pure {
+        // Check handle type using uint8 comparison to avoid type cast panic and
+        // revert with a proper error instead
+        if (uint8(handle[30]) != uint8(expectedType)) revert InvalidExternalHandle(handle);
+
+        // Check handle index
+        uint256 indexHandle = (uint256(handle) & 0x0000000000000000000000000000000000000000ff00000000000000000000) >>
```
dartdart26: Would it be possible to factor out this masking such that we don't repeat it everywhere in code? Into a function `getHandleIndex()` or similar.

library-solidity/codegen/src/templateFHEDotSol.ts
```diff
@@ -595,6 +595,7 @@ function handleSolidityTFHEConvertPlaintextAndEinputToRespectiveType(fheType: Ad
                   : 0
             });
           }
+          _checkExternalHandle(inputBytes32, FheType.${fheType.isAlias ? fheType.aliasType : fheType.type});
```
dartdart26: Do we want to do that before anything else in `fromExternal()`?

PR #3550, opened 2026-08-19
refactor(coprocessor): retire host-listener Anvil mock contracts

Eikix:
## Summary
- Retire host-listener Anvil mock Solidity and the foundry/solc `build.rs` path. Production decode uses `fhevm_host_bindings`.
- Keep historical names (`AclContract`, `TfheContract`, `BridgeContract`) as aliases in `host_listener::contracts` so call sites do not rename event enums.
- Anvil process tests (`main`, `run_poller`, reorg, catchup) encode events from rust_bindings and emit them through a test-only `RawLog` fixture. Not a second ACL/FHE ABI.

Implements zama-ai/fhevm-internal#1922

## Test plan
- [ ] `cargo test -p host-listener` with the coprocessor Postgres harness (Anvil + `test_health` SIGSTOP)
- [ ] Confirm `host-listener/contracts/` is gone and image builds no longer compile test solc
- [ ] `poller_catches_up_to_safe_tip` still calls `run_poller` against Anvil
- [ ] tfhe-worker / stress-test-generator still import `host_listener::contracts`


Eikix:
**Automated cross-branch impact review** (agent-generated, posted at Elias's request)

The retirement itself looks sound on `main`. Three things this PR will run into, flagged ahead of time:

### 1. Dead CI steps left behind on `main`

With `host-listener/build.rs` gone, nothing consumes the hardhat artifacts anymore, but two workflows still produce them on every run:

- `.github/workflows/coprocessor-gpu-tests.yml` lines ~156–173 — `Use Node.js` + `Install safe-chain` + `npm --workspace=host-contracts ci` + `deploy:emptyProxies && npx hardhat compile`
- `.github/workflows/coprocessor-benchmark-gpu.yml` lines ~227–246 — the same block

These existed to pre-populate what `build_contracts()` read. After this PR no `.rs` reads `host-contracts/artifacts` (gw-listener / transaction-sender `build.rs` compile their own crate-local contracts via `foundry_compilers`, no npm), so the steps add minutes of npm+hardhat per GPU run and keep the npm-install-on-build hazard class alive — the exact class this PR retires. Worth deleting in this PR. The equivalent cleanup already landed on `feature/solana`.

### 2. Head-on fold conflicts with `feature/solana` (#3527)

#3527 already moved host-listener onto `fhevm_host_bindings` on the feature branch, with different choices on the same files. The next fold of `main` into `feature/solana` will hit delete/modify conflicts across this PR's whole surface. Divergence map, with a suggested resolution rule — **structure from this PR, names from feature/solana**:

| Surface | This PR (`main`) | `feature/solana` (#3527) | At fold time |
|---|---|---|---|
| `src/contracts/mod.rs` | kept as alias layer (`AclContract` → `ACL`, …) | deleted; call sites renamed to real names | drop the alias layer, keep the renames |
| `build.rs` | deleted | kept — it also hosts the Solana IDL codegen (`generate_solana_abi_schema_hashes`) | the deletion cannot be taken as-is; keep the solana `build.rs` minus `compile_test_contracts()` |
| Anvil mocks | retired; tests emit via the `RawLog` fixture | kept (symlinks + `HOST_LISTENER_SKIP_TEST_CONTRACTS` opt-out) | take the retirement; the skip env in `build.rs` and three Dockerfiles becomes dead and should be removed with it |
| integration tests | rewritten around `RawLog` | independently reworked (incl. `bridge_handler_tests`, the Solana adapter) | manual port, not a merge-take |
| docker-build path filter, `COPY host-contracts/rust_bindings`, Cargo.lock entry | identical fixes on both branches | identical | merges clean |

### 3. Nit: fixture provenance

`tests/fixtures/RawLog.json` is a committed compiled artifact (abi + bytecode) with no regeneration guard. Fine while the fixture contract stays frozen — but a one-line note in `RawLog.sol` recording how the JSON was produced (solc version + command) would keep it reproducible if it ever needs to change.

rudy-6-4 (dismissed):

rudy-6-4:
can close https://github.com/zama-ai/fhevm-internal/issues/152

rudy-6-4 (dismissed):

Eikix:
@mergifyio queue

.github/workflows/coprocessor-benchmark-gpu.yml
```diff
@@ -175,10 +175,8 @@ jobs:
             echo "COMMIT_HASH=$(git rev-parse HEAD)";
           } >> "${GITHUB_ENV}"
 
-      - name: Install rust
-        uses: dtolnay/rust-toolchain@888c2e1ea69ab0d4330cbf0af1ecc7b68f368cc1
-        with:
-          toolchain: nightly
+      - name: Install Rust nightly
```
Eikix: Replacing `dtolnay/rust-toolchain` with a bare `rustup toolchain install nightly` assumes rustup and a default stable toolchain are already present on the Hyperstack GPU runner — the removed action used to guarantee that (bootstrap rustup, add `~/.cargo/bin` to PATH, set a default).

Since this workflow is `workflow_dispatch`-only, PR CI never exercised it: on a runner image without rustup preinstalled (or without `~/.cargo/bin` on PATH before the `cargo +stable install sqlx-cli` / `make init_db` steps), this step fails with `rustup: command not found`, or the later plain-`cargo` steps fail with no default toolchain — discovered only at the next manual benchmark dispatch. The sibling `coprocessor-gpu-tests.yml` still uses `dtolnay/rust-toolchain@stable` on the same runner family, so the two GPU workflows now make different bootstrap assumptions.

Eikix (commented):
Automated review pass (Claude Code) over the mock-contract retirement. Core migration checks out: KeyType/digest ordering matches the retired mock, all workspace Dockerfiles copy `host-contracts/rust_bindings`, dropping foundry/Node from the GPU workflows is safe with build.rs gone, and coprocessor CI is green on this head. Six notes inline — the first three are CI-trigger/bootstrap gaps, the rest test hygiene.

.github/workflows/coprocessor-docker-build.yml
```diff
@@ -162,8 +162,7 @@ jobs:
           - .github/workflows/coprocessor-docker-build.yml
           - coprocessor/fhevm-engine/host-listener/**
           - coprocessor/fhevm-engine/Cargo.*
-          - host-contracts/contracts/*Events.sol
-          - host-contracts/contracts/shared/**
+          - host-contracts/rust_bindings/**
```
Eikix, line 167: The docker-build filter for host-listener now watches `host-contracts/rust_bindings/**` (👍), but the check-changes filters of `coprocessor-cargo-tests.yml` and `coprocessor-cargo-clippy.yml` were not extended the same way — they still only watch the workflow file, `coprocessor/fhevm-engine/**`, and `coprocessor/proto/**`.

Result: a PR that changes a host contract and regenerates `rust_bindings` (e.g. renames or re-types an event field) skips the entire coprocessor test and clippy suites, yet this filter matches — so a host-listener image whose event decoders were never compiled against by tests/clippy in that PR gets rebuilt and shipped. A decoder break would surface only on the next unrelated coprocessor PR or in a deployed environment.

.github/workflows/coprocessor-docker-build.yml
```diff
@@ -197,6 +196,7 @@ jobs:
           - coprocessor/fhevm-engine/tfhe-worker/**
```
Eikix, line 199: tfhe-worker (and the stress-test-generator image) statically link the host-listener crate, whose ABI source this PR moves into `host-contracts/rust_bindings` and copies into their Dockerfiles — but this check-changes filter still only watches `coprocessor/fhevm-engine/tfhe-worker/**` and `Cargo.*`, not `host-contracts/rust_bindings/**` (nor `host-listener/**`).

A push that only regenerates `rust_bindings` changes the bytes compiled into the `tfhe_worker` binary (via its host-listener dependency) but does not trigger a tfhe-worker image rebuild; the registry keeps serving an image with stale event-decoding code until some `Cargo.*` or `tfhe-worker/**` change happens to land. The PR fixed exactly this class of trigger for host-listener but left its dependents' filters untouched.

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
```diff
@@ -496,29 +471,38 @@ async fn emit_dependent_burst_seeded(
         .unwrap_or_else(|| trivial_encrypt_handle(U256::from(seed), 4_u8));
```
Eikix: The local `trivial_encrypt_handle`/`fhe_add_handle` helpers in this file are byte-for-byte duplicates of the new `common::mock_trivial_encrypt_result`/`common::mock_fhe_add_result` that build the emitted events.

`emit_dependent_burst_seeded` computes expected handles with the local pair (here and line 491) while the events it emits carry results from the common pair (`fhe_add_request`/`trivial_encrypt_request`). If either copy of the `keccak("fheAdd" || lhs || rhs || scalar)` recipe is ever edited alone, expected handles and emitted event handles silently diverge and the dependence-tracking assertions fail in a way that looks like a listener bug. Suggest importing the common helpers (or defining the handle fns in `common`) instead of duplicating.

coprocessor/fhevm-engine/host-listener/tests/keygen_integration_tests.rs
```diff
@@ -34,11 +33,8 @@ use tokio_util::bytes::{self, Bytes};
 use tracing::{info, Level};
 use tracing_subscriber::fmt::{writer::MakeWriterExt, MakeWriter};
 
-sol!(
-    #[sol(rpc)]
-    KMSGenerationMock,
-    "artifacts/KMSGenerationTest.sol/KMSGenerationTest.json"
-);
+mod common;
+use common::{activate_crs_request, activate_key_request, RawLog};
```
Eikix: This file keeps local `const TEST_KEY_ID: u64 = 16` and `MATERIALIZED_KEY_BYTES: &[u8] = b"key_bytes"` that duplicate `common::TEST_KEY_ID` and `common::KEY_BYTES`, which now drive the emitted ActivateKey/ActivateCrs events.

The mocked S3 client serves `MATERIALIZED_KEY_BYTES` and assertions use the local `TEST_KEY_ID`, while the on-chain events carry key id and digest from common's constants (`activate_key_request` uses `common::TEST_KEY_ID` and a digest of `common::KEY_BYTES`). Changing one copy (e.g. bumping the test key id in common) makes the listener look up key 16 in the DB while the event announced a different id — tests fail with a confusing "key not found" rather than a compile error. Suggest reusing common's constants here.

coprocessor/fhevm-engine/host-listener/tests/fixtures/RawLog.json
```diff
@@ -0,0 +1,28 @@
+{
```
Eikix, line 1: `RawLog.json` carries hand-compiled solc bytecode that must stay in sync with `tests/fixtures/RawLog.sol`, but unlike `host-contracts/rust_bindings` (guarded by `make check-bindings` in host-contracts-integrity-checks) nothing in CI verifies the artifact matches the source.

If `RawLog.sol` is later edited (e.g. raising the topic limit or changing `emitLog`'s signature) without re-running the documented solc command, tests keep deploying the old bytecode and the change silently does nothing — or the `.sol` and `.json` drift permanently. A tiny check (recompile-and-diff, or a build-time assert on the bytecode hash) would pin them together.

.github/workflows/coprocessor-benchmark-gpu.yml
Eikix: Holding this until we see whether Hyperstack already has rustup. Dispatched the GPU benchmark on this branch (manual `workflow_dispatch`, l40 / erc20): https://github.com/zama-ai/fhevm/actions/runs/33478145033

- If `rustup toolchain install nightly` / later `cargo +stable` succeed, we'll leave this as-is and note that the runner image already provides rustup.
- If it fails at rustup, we'll restore `dtolnay/rust-toolchain` (matching `coprocessor-gpu-tests.yml`).

.github/workflows/coprocessor-docker-build.yml
Eikix: Fixed in 9844d42804: `host-contracts/rust_bindings/**` is now in the check-changes filters of both `coprocessor-cargo-tests.yml` and `coprocessor-cargo-clippy.yml`.

.github/workflows/coprocessor-docker-build.yml
Eikix: Fixed in 9844d42804: `host-contracts/rust_bindings/**` is now on every docker-build image filter that `COPY`s those bindings (db-migration, gw-listener, host-listener, sns-worker, tfhe-worker, tx-sender, zkproof-worker, consensus-detector, upgrade-controller), plus `coprocessor-stress-test-tool-docker-build.yml`.

coprocessor/fhevm-engine/host-listener/tests/host_listener_integration_tests.rs
Eikix: Fixed in 9844d42804: dropped the local `trivial_encrypt_handle` / `fhe_add_handle` helpers and switched this file (including `emit_dependent_burst_seeded`) to `common::mock_trivial_encrypt_result` / `common::mock_fhe_add_result`.

coprocessor/fhevm-engine/host-listener/tests/keygen_integration_tests.rs
Eikix: Fixed in 9844d42804: this file now imports `TEST_KEY_ID` and `KEY_BYTES` from `common` instead of keeping local duplicates.

.github/workflows/coprocessor-benchmark-gpu.yml
Eikix: Confirmed: Hyperstack does **not** have rustup.

Run https://github.com/zama-ai/fhevm/actions/runs/33478145033 failed at **Install Rust nightly** with `rustup: command not found` (exit 127).

Restored `dtolnay/rust-toolchain` in d7c5cc51f0, same pin and nightly toolchain as `coprocessor-benchmark-cpu.yml` / `coprocessor-gpu-tests.yml`. Re-dispatching the same l40/erc20 bench on this commit to verify the install step.

.github/workflows/coprocessor-benchmark-gpu.yml
```diff
@@ -176,7 +176,8 @@
           } >> "${GITHUB_ENV}"
 
       - name: Install rust
-        uses: dtolnay/rust-toolchain@888c2e1ea69ab0d4330cbf0af1ecc7b68f368cc1
+        # Hyperstack GPU runners do not ship rustup (run 33478145033: rustup: command not found).
+        uses: dtolnay/rust-toolchain@a54c7afa936fefeb4456b2dd8068152669aa8203 # zizmor: ignore[superfluous-actions] Hyperstack GPU runners do not ship rustup
```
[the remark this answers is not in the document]
Eikix: The previous pin (`a54c7afa`) is not a tag, so `stale-action-refs` is expected.

Repinned in d869f34c31 to `6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772`, which is the current `dtolnay/rust-toolchain@v1` tag (same commit as `master` HEAD). Kept the `superfluous-actions` ignore: Hyperstack still does not ship rustup.

coprocessor/fhevm-engine/host-listener/tests/fixtures/RawLog.json
Eikix: Leaving this out of this pass: no CI check that `RawLog.json` matches `tests/fixtures/RawLog.sol`. Happy to do it as a follow-up rather than block this retirement.

Eikix:
Ready for re-review on `d869f34c31`.

Review notes from 2026-08-31:

| # | Topic | Status |
|---|---|---|
| 1 | GPU bench rustup | Fixed. Hyperstack has no rustup (`rustup: command not found` on [33478145033](https://github.com/zama-ai/fhevm/actions/runs/33478145033)). Restored `dtolnay/rust-toolchain` pinned to current `v1` (`6c977a6`). Verified: [33479515991](https://github.com/zama-ai/fhevm/actions/runs/33479515991) completed the erc20/l40 bench. |
| 2 | cargo-test / clippy `rust_bindings` filters | Fixed |
| 3 | docker-build / stress-test-generator `rust_bindings` filters | Fixed |
| 4 | Duplicate handle helpers in host-listener tests | Fixed |
| 5 | Duplicate `TEST_KEY_ID` / `KEY_BYTES` | Fixed |
| 6 | `RawLog.json` vs `RawLog.sol` CI sync | **Deferred** (follow-up) |

zizmor is green on this HEAD. Coprocessor cargo-test / GPU tests were green on the previous HEAD; this commit is workflow-pin only.

Please re-approve:
- `@zama-ai/fhevm-coprocessor` (previous approve was dismissed by later pushes)
- `@zama-ai/fhevm-devops` (Dockerfiles)

.github/workflows/coprocessor-benchmark-gpu.yml
```diff
@@ -224,27 +222,6 @@ jobs:
...
-        with:
-          node-version: 20.x
-
-      - name: Install safe-chain
-        shell: bash
-        run: |
-          # safe-chain v1.5.3 is distributed as an immutable GitHub release asset —
-          # its content is permanently fixed and cannot be modified after the tag is published.
-          # Verify immutability: curl -fsSL https://api.github.com/repos/AikidoSec/safe-chain/releases/tags/1.5.3 | grep immutable
-          curl -fsSL https://github.com/AikidoSec/safe-chain/releases/download/1.5.3/install-safe-chain.sh | sh -s -- --ci
```
fegmorte, line 238: I'm not sure about removing this part as it was asked by @zama-ai/security-team to add this action

.github/workflows/coprocessor-gpu-tests.yml
```diff
@@ -153,25 +150,6 @@ jobs:
...
-        with:
-          node-version: 20.x
-
-      - name: Install safe-chain
-        shell: bash
-        run: |
-          # safe-chain v1.5.3 is distributed as an immutable GitHub release asset —
-          # its content is permanently fixed and cannot be modified after the tag is published.
-          # Verify immutability: curl -fsSL https://api.github.com/repos/AikidoSec/safe-chain/releases/tags/1.5.3 | grep immutable
-          curl -fsSL https://github.com/AikidoSec/safe-chain/releases/download/1.5.3/install-safe-chain.sh | sh -s -- --ci
```
fegmorte, line 167: Same comment here

.github/workflows/coprocessor-benchmark-gpu.yml
Eikix: Restored in e1418885e6: `Use Node.js` + `Install safe-chain` (same pin/script as `main`).

Left out the host-contract hardhat/foundry compile that used to follow it — that step existed for host-listener `build.rs`, which this PR deletes. npm is no longer invoked in this job; safe-chain is back so the security wrapper stays in place.

.github/workflows/coprocessor-gpu-tests.yml
Eikix: Same restore here in e1418885e6 (`Use Node.js` + `Install safe-chain`). Host-contract compile stays dropped; this job no longer runs npm, but the security-team action is back.

coprocessor/fhevm-engine/host-listener/build.rs
rudy-6-4, line 1: Great to have that removed. It was blocking concurrent compilation (IDE + manual)

rudy-6-4 (approved):

fegmorte (approved):
LGTM for devops part

Eikix:
@mergifyio queue

PR #3557, opened 2026-08-19
feat(coprocessor): publish generation-scoped block manifests

rudy-6-4:


coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/lineage.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub(crate) fn as_mut_vec(&mut self) -> &mut Vec<RangeNode> {
+        &mut self.ranges
+    }
+}
+
+/// One immutable aligned range root on a specific host-chain lineage.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct RangeNode {
+    /// Generation that produced the newest block covered by this range.
+    pub end_block_generation: i64,
```
rudy-6-4, line 31: No longer usefull as history are scoped by generation. This can be removed from all range and signatures.

coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/lineage.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    pub(crate) fn as_mut_vec(&mut self) -> &mut Vec<RangeNode> {
+        &mut self.ranges
+    }
+}
+
+/// One immutable aligned range root on a specific host-chain lineage.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct RangeNode {
```
rudy-6-4, line 29: comment field
rename start with start_block_number etc

coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/manifest_archive.rs
```diff
@@ -0,0 +1,630 @@
...
+pub(crate) enum StoreOutcome {
+    Inserted,
+    AlreadyPresent,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum ManifestSource {
+    Local,
+    #[allow(dead_code)]
+    Peer,
```
rudy-6-4: Check where it is used

coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/publication/block_discovery.rs
```diff
@@ -0,0 +1,479 @@
...
+pub(crate) async fn discover_completed_sns_blocks(pool: &PgPool) -> Result<u64, ExecutionError> {
+    let generation = crate::manifest_consensus::storage::active::load_generation(pool).await?;
+    discover_completed_sns_blocks_for_generation(pool, generation).await
+}
+
+pub(crate) async fn discover_completed_sns_blocks_for_generation(
+    pool: &PgPool,
+    generation: i64,
+) -> Result<u64, ExecutionError> {
+    let inserted = sqlx::query_scalar!(
```
rudy-6-4: not readable. split in several query.

rudy-6-4:
Automated fix: addressed CI failure in coprocessor-cargo-test/tfhe-worker (and sns-worker / GPU bridge tests).

Bridge fixtures used 4-byte `ciphertext` digests as `s3_publication_verified_digest`, which violates `ciphertext_digest_s3_publication_verified_digest_check` (must be NULL or 32 bytes). `test_garbage_collect` also omitted the new S3 verification witness, so GC could not delete uploaded `ciphertexts128` rows.

`coprocessor-cargo-test/gw-listener` has no available logs (runner died mid-job; blob 404) and looks infra-only; not changed in code.

coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/lineage.rs
rudy-6-4: Addressed in e36ef70496694991b89cfc8944cc359f00ff7aee: renamed `RangeNode.start`/`end` to `start_block_number`/`end_block_number` and documented the remaining fields (`scale`, hashes, digest). Call sites in `manifest_frontier.rs` and `manifest_history.rs` were updated with the same names.

coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/publication/block_discovery.rs
rudy-6-4: Addressed in d9a4eeda8c6dc878aa87650a0f5214d0e4bcfdef: the single CTE is now four queries in one transaction — `load_discovery_bounds` (frontier + latest valid + generation window), `load_valid_host_block` + `insert_manifest_state` for a missing generation anchor, then `insert_producer_blocks` for `handle_producer_block` rows at/after `first_block`. Chain cadence is a Rust helper. Semantics are unchanged: gen-0 still anchors at the latest valid block, later generations still refuse to cross `generation_block_window.start_block`, and producer inserts still require a present anchor when the frontier is empty.

coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/lineage.rs
rudy-6-4: `end_block_generation` is still required on `RangeNode` even though a live frontier is generation-scoped.

- The signed `HistoricalRange` wire format still carries `end_block_generation`, and `ManifestPayload::validate` rejects a range whose generation does not equal `payload.generation`.
- `block_range_commitment` is keyed on `(host_chain_id, end_block_generation, range_start, range_end, ...)`, so hydration of a previous manifest still looks the range up by that column.
- `append_leaf` uses the field as the generation-boundary check (`range.end_block_generation != leaf.end_block_generation`) so a reconstructed frontier cannot silently merge ranges from two generations.

Dropping it from `RangeNode` would only re-thread the same value from the publishing generation into every insert/hydrate path, including the durable commitment row. I would keep the field until a later change removes it from the signed schema and the commitment primary key together.

coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/manifest_archive.rs
rudy-6-4: `ManifestSource::Peer` is used.

- Production publication (`publisher.rs`) always stores with `ManifestSource::Local`.
- `store_authenticated_manifest` takes `source` and writes `manifest_source`; if a later local store races a peer row it upgrades `peer` → `local`.
- Tests construct `Peer` directly: `manifest_archive_tests.rs` (duplicate/identity cases) and `manifest_soak_tests.rs` (peer ingest during soak).

The `#[allow(dead_code)]` is only because the non-test binary never constructs `Peer`; the variant is not unused. It stays as the publication-side hook for the follow-up verification PR that ingests peer manifests.

Eikix (commented):

Reviewed `cd901d6`. The generation-scoped publication design looks sound; no additional core publication defect was confirmed.

- **Fix local startup:** [run_with_localnet.sh](https://github.com/zama-ai/fhevm/blob/cd901d682129cc83cace56b653387b82c07d2610/coprocessor/fhevm-engine/consensus-detector/run_with_localnet.sh#L36) supplies `--my-bucket`, enabling publication, but neither `--private-key` nor AWS signer mode. The default signer therefore fails before startup. Pass the configured signer/key explicitly.
- **Exercise the wire contract in CI:** add an explicit test invocation and path trigger for `shared/block-manifest`; compiling it as a service dependency does not run its unit tests.

Validation: all 21 standalone wire tests passed locally using offline-resolved dependencies. Full service/E2E tests were not run locally; current-head CI evidence is still needed after resolving the merge conflicts.

PR #3560, opened 2026-08-20
feat(coprocessor): transaction-scope materialisation

antoniupop:
This PR updates ciphertext materialisation to happen at transaction boundaries instead of after every operation. Re-randomisation remains at the op level.

The PR also:
  - Persists operand-origin masks so the worker can distinguish transaction-local operands from canonical ciphertexts. Also see #3543 for context on how output handles computed onchain also discriminate based on operand origin.
  - Defers and retries work that cannot yet be interpreted.
  - Quarantines repeatedly deferred work so it cannot starve younger dependence chains.
  - Adds unit, integration, and consensus E2E coverage for materialisation and aliasing behaviour.


coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -351,36 +371,100 @@ fn execute_partition(
...
+                            // allow always lands in a transaction that
+                            // minted the handle — the original mint, a
+                            // re-mint of the same spelling (which inserts
+                            // and stamps its own row), or a bridge delivery
+                            // (no listener-stamped computation rows; the
+                            // ingest allow set covers bridge dst handles) —
+                            // and an allow can never arrive later for a
+                            // handle that was not already persisted.
+                            let forwarded = if is_allowed {
+                                match compress_output(&working, &tid, opcode) {
```
PanGan21, line 426: Compression used to run on main inside `catch_unwind` which would catch any compression underlying panics. Shall we do the same here to avoid panics on the scheduler?

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -624,64 +1755,424 @@ WHERE c.transaction_id IN (
...
+    // outputs are equally unobtainable, so their transaction-local
+    // consumers drain with them.
+    uncomputable.extend(dead_input_rows);
+    loop {
+        let mut progressed = false;
+        let mut index = 0;
+        while index < ops.len() {
+            let blocked = ops[index].inputs.iter().any(|input| match input {
+                DFGTaskInput::LocalDependence(dh) => uncomputable.contains(dh),
+                _ => false,
```
PanGan21: Shall we enumerate the other variants instead of using `_ => false`? Not important for this pr but can become footgun

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -351,36 +371,100 @@ fn execute_partition(
...
+                            if let Some(forwarded) = forwarded {
+                                for edge in edges.edges_directed(nidx, Direction::Outgoing) {
+                                    let child_index = edge.target();
+                                    let Some(child_node) = dfg.graph.node_weight_mut(child_index)
+                                    else {
+                                        error!(target: "scheduler", {index = ?child_index.index() }, "Wrong dataflow graph index");
+                                        continue;
+                                    };
+                                    child_node.inputs[*edge.weight() as usize] =
+                                        DFGTaskInput::Value(forwarded.clone());
```
PanGan21, line 460: It seems that this is one of the most important changes, a same-transaction op used to be handed to its consumer compressed, now it's handed over raw. We need to make sure the deployment works correctly with B/G otherwise this could potentially create issues. (No action on the pr just wanted to flag it)

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
```diff
@@ -1016,11 +1103,25 @@ pub async fn synthesize_finalized_fallback_grants(
...
     if logs.is_empty() {
         return Ok(());
     }
+    // LISTENER-side computation inserts are fail-closed on a missing mask
+    // (insert_computation rejects None; the worker's derive-on-read fallback
+    // exists only for rows written by pre-mask binaries), so derive it
+    // here too. Synthesized `TrivialEncrypt` rows carry no encrypted
+    // operands, so every derived mask is empty; deriving rather than
+    // hardcoding keeps this path honest if the shape ever changes.
+    populate_operand_boundary_masks(&mut logs)?;
```
PanGan21: AI flagging this, I think it is correct when validating it:

This call can stall the listener on an entire block, in order to compute a value that is always zero here.

What this path does: it takes finalized bridge fallback grants and synthesizes them as TrivialEncrypt computation rows.

Why it can fail: populate_operand_boundary_masks rejects any computation event
whose transaction_hash is None — "refusing computation event without transaction
hash for operand-origin derivation". It needs the hash because it groups the minted-handle set per transaction, and the hash is the map key.

Why that's reachable here: fallback_granted_events.transaction_id is a NULLABLE
column, and this path is written to tolerate that everywhere else — it reads it
as an Option, fallback_grant_conflicts takes an Option, insert_pbs_computations
does .map() over it. So None is a normal value here, not a corruption signal.

What happens on a NULL row:
  1. populate_operand_boundary_masks returns Err
  2. the `?` propagates it out of the synthesis function
  3. the block's ingest transaction rolls back
  4. the block is retried, and fails the same way — forever

So one row with no transaction id stops the listener advancing, and it won't
recover on its own.

The part that makes this feel wrong rather than merely strict: the value being computed can't depend on the missing input. TrivialEncrypt has no encrypted
operands (tfhe_encrypted_operand_positions returns vec![] for it), so the mask is
always 32 zero bytes no matter which transaction it belongs to. The comment
directly above this line already says so: "Synthesized TrivialEncrypt rows carry
no encrypted operands, so every derived mask is empty."

Suggestion, any of these would do:
  - skip and warn on the individual row instead of failing the whole block
  - only require a transaction hash for events that actually have encrypted
    operands
  - pass an explicit empty mask on this path (the existing comment argues against
    hardcoding, which is fair, so one of the first two is probably nicer)

Same class of thing at line 80: the missing / duplicate log_index checks are also
whole-block failures now, where ensure_logs_order used to warn and carry on. A
one-off provider quirk that used to be a log line is now a permanent stall.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -327,7 +1209,100 @@ async fn tfhe_worker_cycle(
...
+                // the rotation goes through the manager's own pool connection.
+                // Rotation is bounded: execution-class errors can also be
+                // GLOBAL (a lost GPU device panics outside the per-op
+                // catch_unwind and surfaces as Other), and rotating chain
+                // after chain would silently rewrite the whole backlog's
+                // FIFO order — after consecutive failures across different
+                // chains the guard above stops matching and the error
+                // propagates to the run loop's log-and-retry instead.
+                WORKER_ERRORS_COUNTER.inc();
+                error!(target: "tfhe_worker", { error = %cycle_error },
```
PanGan21: Let's include chain id and transaction id. It could be useful info for debugging

test-suite/e2e/test/consensus/README.md
```diff
@@ -0,0 +1,72 @@
...
+GATEWAY_RPC_URL=... \
+GATEWAY_CONFIG_ADDRESS=... \
+CIPHERTEXT_COMMITS_ADDRESS=... \
+DATABASE_URL_0=... DATABASE_URL_1=... DATABASE_URL_2=... \
+npx hardhat test test/consensus/materializationConsensus.ts --network localCoprocessor
+```
+
+The gate also runs an **aliased-handle** scenario (`AliasFixture.sol`).
+Under the minted-in-transaction handle discriminant (see
+`docs/minted-discriminant-note.md`), sourcing is part of the handle: two
```
PanGan21: Is `docs/minted-discriminant-note.md` uncommitted maybe?

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
antoniupop: Good point indeed. Fixing.

test-suite/fhevm/docker-compose/coprocessor-docker-compose.yml
```diff
@@ -201,6 +201,13 @@ services:
       - --private-key=${TX_SENDER_PRIVATE_KEY}
       - --ciphertext-commits-address=${CIPHERTEXT_COMMITS_ADDRESS}
       - --input-verification-address=${INPUT_VERIFICATION_ADDRESS}
+      # The consensus-completing addCiphertextMaterial (the threshold-th add)
+      # finalizes quorum in the same transaction and costs well over the
+      # estimate taken against the pre-quorum state; 120% overprovision
+      # starves it (ReentrancySentryOOG) whenever this sender loses the race
+      # to be an early adder. Every handle has a threshold-th adder in
+      # byte-consensus topologies.
+      - --gas-limit-overprovision-percent=300
```
PanGan21: We change this in e2e, but by reading the comment, could it be that it needs to change in prod values as well?

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
antoniupop: Yes indeed, we have multiple breaking changes in v0.15.

coprocessor/fhevm-engine/host-listener/src/database/ingest.rs
antoniupop: Good catch, this is worse than a per-block stall, it runs inside the finalization loop, which is deliberately batch-stop, so one NULL-transaction_id grant halts finalization advancement permanently.

Fixing this by deriving the mask per event at construction with `operand_boundary_mask_from_minted(&event, |_| false)` and dropping the block-level pass from this path. The empty-minted-set closure is correct by construction asthe executor never ran, so nothing synthesized can be a transaction-local operand.

On the line-80 checks, we'd keep those fail-closed. Under the old warn-and-continue, log order would depend on RPC delivery order, so two nodes behind different providers could silently diverge in ciphertext bytes. A stall is visible, retried against a fresh fetch and operator-fixable. Adding a metric.

test-suite/e2e/test/consensus/README.md
antoniupop: This was an older note, shouldn't have been linked - removing references.

test-suite/fhevm/docker-compose/coprocessor-docker-compose.yml
antoniupop: Yes, although by hte time this gets in prod, gw commitments will no longer be material. Fixing this here anyway, but could be worth backporting really.

coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
```diff
@@ -492,7 +572,59 @@ fn try_execute_node(
...
+    let ct_bytes = match compressed {
+        Ok(compress_result) => compress_result.inspect_err(|error| {
+            telemetry::set_current_span_error(error);
+        })?,
+        Err(e) => {
+            let msg = panic_message(e);
+            error!(target: "scheduler", { txn_id = %telemetry::short_hex_id(transaction_id), msg },
+                "Panic while compressing operation output");
+            telemetry::set_current_span_error(&msg);
+            return Err(SchedulerError::ExecutionPanic(msg).into());
```
PanGan21, line 621: We could potentially take advantage of the `RETRYABLE_STAMP_MARKER` that you added and retry the compress panics by [extending the tfhe-worker query](https://github.com/zama-ai/fhevm/blob/ddd26cff2bd3811858de875a9018dee579d4989d/coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs#L1708) to fetch those rows as well.
I understand that compress can fail for example due to GPU memory which seems like a retryable error. What do you think?

antoniupop: Yes, makes sense. fixing

PanGan21 (dismissed):
Nice work!

antoniupop:
> Nice work!

Thanks a lot for the review - you found a lot of important improvements! Cheers!

PanGan21 (approved):

fegmorte (approved):

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

PR #3562, opened 2026-08-20
chore(library-solidity): adding polygon mainnet to zama config

nboisde:


0xalexbel (approved):
LGTM!

0xalexbel (approved):

library-solidity/config/ZamaConfig.sol
```diff
@@ -26,6 +26,8 @@ library ZamaConfig {
             config = _getEthereumConfig();
         } else if (block.chainid == 11155111) {
             config = _getSepoliaConfig();
+        } else if (block.chainid == 137) {
```
melanciani, line 27: very nit: maybe best to keep mainnet chains on top, and testnet under (ie move the else if one level up)

library-solidity/config/ZamaConfig.sol
melanciani: comment to update (+ a few others, see https://github.com/zama-ai/protocol-apps/pull/246 as a comparison)

melanciani: could we take the opportunity to remove this comment ?

nboisde: This comment has been removed on main branch already (only fixes here), but overall agreed.

library-solidity/config/ZamaConfig.sol
nboisde: Sorted this by chains instead of Zama envs, but could be managed by zama envs indeed.

melanciani (approved):
great thanks !

PanGan21 (approved):

PR #3563, opened 2026-08-20
fix(sdk): mainnet protocolConfig address and browser-smoke test path

geoxel:
## Summary
- Fill in the previously-unset `protocolConfig` address for the mainnet chain definition, pointing it at the deployed ProtocolConfig contract.
- Fix `test:browser-smoke`'s webServer command to pass an absolute path to `vite.config.ts`, so it's immune to Vite's root-relative re-resolution of relative `--config` paths (the vite config sets `root` two directories up, which broke the previous relative-path fix).

## Test plan
- [x] `npm run test:browser-smoke` passes locally in `sdk/js-sdk` (all 15 tests, chromium/firefox/webkit)
- [x] `js-sdk-dod` CI workflow passes


0xalexbel (approved):

geoxel:
@mergifyio queue

PR #1916, opened 2026-02-05
ci(common): e2e tests for pr on release branch

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/1015

[PR](https://github.com/zama-ai/fhevm/pull/1915) on `release/v0.11.x` showing the e2e tests CI run as expected.

PR #3580, opened 2026-08-24
feat(coprocessor): decide upgrades by the consensus protocol version

PanGan21:
## Summary

- A new `consensus_version` column on `versioning` holds the value that decides roles. It is compiled into the binary and moves only when a release changes what operators must agree on.
- Services resolve their role at startup from the compiled consensus version, and a retired binary refuses to start. The existing write guard now refuses a service retired while running by comparing consensus versions rather than releases.
- A fresh database seeds both versions from the compiled binary, guarded by a one row setup marker so it refuses to run against a database that already has versions or upgrade history.
- `versioning.stack_version` keeps its meaning: written at cutover, and what a pre-0.15 binary compares against to retire itself. 0.15 reads it at cutover, to check the proposal's release. No service branches on it to pick a role anymore.
- The e2e reads the green release from the binary (`--stack-version`) instead of a configured value. The Cargo override features stay for local runs only (`run_with_localnet.sh`).
- The blue-green **preview env** needs the stacked PR that follows this one: with this change alone its migrator ordering seeds green's versions as live. The compose e2e is unaffected.

| | `CONSENSUS_PROTOCOL_VERSION` (u32) | `STACK_VERSION` (string) |
|---|---|---|
| Bumped | only when a release changes what operators must agree on and requires synchronization | every release |
| Decides blue/green | **yes** | no |
| Decides retirement | **yes** | only for pre-0.15 binaries |
| Names the GCS schema | no | **yes** — `gcs-<STACK_VERSION>` |
| Identifies a proposal | no | **yes** |
| Column written at cutover | `versioning.consensus_version` | `versioning.stack_version` |

## Who decides green or blue

`resolve_gcs_mode`, once per service at startup, from the compiled consensus version against `versioning.consensus_version`:

- newer than live -> **green** (writes to the `gcs-<STACK_VERSION>` schema)
- equal to live -> **blue**, the live stack
- older than live -> treated as blue here; the write guard stops it (below)
- row missing or unseeded -> blue

While running, `reconcile_stack_mode` reacts to the `stack-version-upgraded` notification: a green service whose consensus version now equals live becomes the live stack, and a live service whose version no longer matches pauses into no-op mode. No service reads `stack_version` to pick a role.

## Who retires a binary

The write guard. Every write opens its transaction through `begin_guarded_pool` or `begin_guarded_conn`, and each one checks the live consensus version first. A binary older than live gets `RetiredStack` and its write is refused. 
Pre-0.15 binaries have no consensus version, so they compare `stack_version` to decide they are retired. That is why raising the consensus version also means raising `STACK_VERSION`, otherwise the old binaries never step down.

## What `STACK_VERSION` is still responsible for

- The green schema name, `gcs-<STACK_VERSION>`.
- Proposal identity: `validate_cutover_release` requires the proposal to name this exact release, and to be above the live one.
- The value written to `versioning.stack_version` at cutover.

## Who writes the columns

- `bootstrap_versioning`, on a fresh install only, seeds both from the compiled binary. It is gated by a one-row setup marker, so it refuses a database that already carries versions or migration history, and services refuse to start
  while that marker is present.
- `execute_cutover` promotes both in one transaction and emits `stack-version-upgraded`.

Related:
- https://github.com/zama-ai/fhevm-internal/issues/1916
- https://github.com/zama-ai/fhevm-internal/issues/1620


PanGan21:
Mult copro multi chain E2E test run for blue green: https://github.com/zama-ai/fhevm/actions/runs/32717359024/job/97401348173

coprocessor/fhevm-engine/fhevm-engine-common/src/lib.rs
```diff
@@ -22,56 +22,35 @@ pub mod common {
...
+// operators must agree on:
+//   - new key parameters
+//   - the GPU feature is turned on
+//   - randomization changes
+//   - the scheduling logic changes
+// Leave it as is for every other release, which then rolls out without a cutover.
+// A macro because the schema name needs it in `concat!`.
+macro_rules! consensus_protocol_version {
+    () => {
+        1
```
goshawk-3: Should it be a SemVer style (v1.0.0 or simply 1.0.0)?

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
```diff
@@ -16,169 +11,295 @@ use sqlx::{Connection, PgConnection, Pool, Postgres, Transaction};
...
-            gcs_mode,
-            "stack-version-upgraded received; no mode change"
-        );
+    match classify_consensus_version(CONSENSUS_PROTOCOL_VERSION, live)? {
+        RuntimeStackMode::Candidate => {
+            mode.set(RuntimeStackMode::Candidate);
+            info!(
+                binary_consensus_version = CONSENSUS_PROTOCOL_VERSION,
+                live_consensus_version = live,
+                "service is green"
```
goshawk-3: Why do we introduce a new naming?  Here it says `service is green` but in the code we call it `Candidate`? 

Isn't this better aligned with the BlueGreen deployment terminology?

```rust
RuntimeStackMode {
      Blue = 0,
      Green = 1,
      Retired = 2,
  }
```

coprocessor/fhevm-engine/tfhe-worker/src/tests/utils.rs
```diff
@@ -148,6 +148,14 @@ async fn setup_test_app_custom_docker() -> Result<TestInstance, Box<dyn std::err
 
     println!("Running migrations...");
     sqlx::migrate!("./migrations").run(&pool).await?;
+    // Match the database to this build so the worker runs live, not green.
```
goshawk-3: `Live` here means blue?

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
```diff
@@ -16,169 +11,295 @@ use sqlx::{Connection, PgConnection, Pool, Postgres, Transaction};
...
+    Ok(consensus)
 }
 
-/// True iff this binary's [`STACK_VERSION`] equals `live` (same major.minor.patch).
-pub fn binary_matches(live: &str) -> bool {
-    parse_version(STACK_VERSION) == parse_version(live)
+/// Current role of a running service.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[repr(u8)]
+pub enum RuntimeStackMode {
```
goshawk-3: What's the difference with "StackMode" struct, we already use?

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
```diff
@@ -190,65 +201,76 @@ pub async fn resolve_gcs_mode(database_url: &str) -> anyhow::Result<bool> {
     )
     .await?;
     let mut conn = PgConnection::connect_with(&options).await?;
-    let live = live_stack_version(&mut conn).await?;
+    // The versions are not final until setup finishes.
+    let setup_in_progress: bool =
+        sqlx::query_scalar("SELECT to_regclass('public._fhevm_versioning_bootstrap') IS NOT NULL")
```
PanGan21, line 206: Startup guard so a service never picks its mode from a database that is still being migrated.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -1474,19 +1474,41 @@ pub async fn execute_cutover(pool: &Pool<Postgres>) -> Result<(), Error> {
         warn!("cutover: GCS proposal is not UpgradeAuthorized — skipping (already cut over)");
         return Ok(());
     };
+    let live_stack_version: String =
```
PanGan21, line 1477: A refused cutover marks the attempt failed rather than erroring after this change, because rows left at UpgradeAuthorized can't be replaced by a later proposal.

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
```diff
@@ -70,34 +68,14 @@ pub async fn run_confidential_bridge(
...
         apply_gcs_mode_search_path(gcs_mode),
     )
     .await?;
-    {
-        let listener_pool = pool.clone();
-        let listener_mode = stack_mode.clone();
-        let listener_cancel = cancel_token.child_token();
-        tokio::spawn(async move {
-            if let Err(err) =
-                run_stack_version_listener(listener_pool, listener_mode, listener_cancel).await
```
goshawk-3: Why do we need to drop the version_listener ?

coprocessor/fhevm-engine/fhevm-engine-common/src/bin/bootstrap_versioning.rs
```diff
@@ -8,5 +8,5 @@ async fn main() -> anyhow::Result<()> {
         connect_pool_with_options(&database_url, PgPoolOptions::new().max_connections(1), None)
             .await?;
 
-    fhevm_engine_common::versioning::bootstrap_versioning(&pool).await
+    fhevm_engine_common::bootstrap_versioning::bootstrap_versioning(&pool).await
```
goshawk-3, line 11: Is this binary needed in production or only in e2e setup?

coprocessor/fhevm-engine/db-migration/migrations/20260819120000_add_consensus_version_to_versioning.sql
```diff
@@ -0,0 +1,4 @@
+-- Active consensus version. Existing databases start at version 0. New
+-- databases set this value during setup.
+ALTER TABLE versioning
+    ADD COLUMN IF NOT EXISTS consensus_version BIGINT NOT NULL DEFAULT 0;
```
goshawk-3, line 4: This default 0 would mean that the live mainnet is running with consensus version 0.0.0

IMO, consensus version 1.0.0 corresponds to the currently deployed mainnet -  v0.14.0, CPU keys, and the current parameters.

We can keep the 0.0.0 for current live but it might sound awkward when we create the public registry with mappings (consensus version -> stack_version, keys, params ...)  WDYT?

coprocessor/fhevm-engine/db-migration/initialize_db.sh
```diff
@@ -203,7 +203,60 @@ repair_bridge_tables_migration_checksum() {
     \$\$;" || { log "Failed to repair bridge_tables migration checksum."; exit 1; }
 }
 
+# Sets DEPLOYMENT_MODE from the database: a setup marker means an interrupted
+# setup to finish, migration history means an upgrade, an empty database means a
+# new one.
+resolve_deployment_mode() {
```
goshawk-3, line 209: What's the meaning of deployment mode? Does it run in production?

coprocessor/fhevm-engine/fhevm-engine-common/src/lib.rs
```diff
@@ -26,41 +27,37 @@ pub mod common {
...
 pub const HANDLE_VERSION: i16 = 0;
 
+// Decides blue/green mode. Raise it by one when a release changes the results
+// operators must agree on:
+//   - new key parameters
+//   - the GPU feature is turned on
+//   - randomization changes
+//   - the scheduling logic changes
+// Leave it as is for every other release, which then rolls out without a cutover.
+pub const CONSENSUS_PROTOCOL_VERSION: u32 = 1;
```
goshawk-3, line 59: As stated above, we probably should designate this as consensus version 2.0.0

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
goshawk-3: That's a good comment to be added cause the "until setup finishes" sounded unclear to me.

coprocessor/fhevm-engine/fhevm-engine-common/src/lib.rs
goshawk-3, line 59: In localnet testing, it's convenient to modify the version with ENV

#[cfg(feature = "consensus-version-override")]
macro_rules! stack_version {
    () => {
        env!("BUILD_CONSENSUS_VERSION")
    };
}

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
goshawk-3: What does it mean a "refused cutover"? We need to have a retryable cutover transaction since some errors in cutover  could be transient.

host-contracts/COPROCESSOR_UPGRADE_RUNBOOK.md
```diff
@@ -67,7 +68,7 @@ npx hardhat task:prepareCoprocessorUpgrade \
   --duration 30m \
   --buffer 1h \
   --proposal-id 1 \
-  --software-version v0.14.0
+  --software-version v0.15.0
```
goshawk-3, line 71: In transition v0.14 -> v0.15
In the actual upgrade proposal, what software-version we should pass - v0.15.0 or v1.0.0?

In transition v0.15 -> v0.15.1
We should pass "1.0.0" ?

goshawk-3 (changes requested):

coprocessor/fhevm-engine/tfhe-worker/src/bridge.rs
goshawk-3: That's outdated actually

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
PanGan21: Comment added

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
PanGan21: The cutover stays retryable here for transient error, this didn't change.

It is only reached when the proposal names a different release than the binary which is deterministic, so retrying can never help. It marks the attempt failed because rows left at `UpgradeAuthorized` cannot be replaced by a later proposal, which would block every future upgrade. I rewrote the comment to explain this better.

coprocessor/fhevm-engine/fhevm-engine-common/src/bin/bootstrap_versioning.rs
PanGan21: It can run in production yes, but only against a fresh database. `initialize_db.sh` runs it after the migrations when the mode is bootstrap. So in every local stack and e2e run (the databases are always fresh), a new operator or a new environment.

The version was seeded by `20260521160000_create_versioning_table.sql`, which hard codes 'v0.14' and is already delivered. 
This is solving [#1916](https://github.com/zama-ai/fhevm-internal/issues/1916) cleanly for tests, e2e tests or starting a new operator.

coprocessor/fhevm-engine/db-migration/migrations/20260819120000_add_consensus_version_to_versioning.sql
PanGan21: Agreed migration defaults now to 1, constant is 2 in commit: https://github.com/zama-ai/fhevm/pull/3580/changes/8f6acfca1585e33f4b0af31fea52f837ce12862b

On 1.0.0 vs 1: today it is a u32 counter and not semver format since I couldn't think of a use case. 
It is only ever compared three ways against the live value: newer means green, equal means live, older means retired. 
Where would you use the semver parts? A consensus change is breaking by definition, which is what triggers a cutover, so I cannot see what a minor or patch bump would be.

coprocessor/fhevm-engine/db-migration/initialize_db.sh
PanGan21: Also production as mentioned [here](https://github.com/zama-ai/fhevm/pull/3580#discussion_r3947363608). Existing when `_sqlx_migrations` is present and the setup marker is not, `bootstrap` otherwise, that is what stops a redeploy reseeding versions on a live database. Documented at `initialize_db.sh`.

coprocessor/fhevm-engine/fhevm-engine-common/src/lib.rs
PanGan21: That is changed as 2 for now.

PanGan21: I would rather not. Removing the build time override is the point of this PR. We wanted the compiled code as the only source of truth, which is why BUILD_STACK_VERSION is gone. An env override brings the drift back: binary says one thing, env another, database seeded from whichever ran.

The e2e no longer needs it either, it reads the release from the binary via `gcsBinaryRelease()` and on localnet changing the constant costs the same rebuild. If you have a workflow that cannot rebuild, tell me and I will look again.

host-contracts/COPROCESSOR_UPGRADE_RUNBOOK.md
PanGan21: `--software-version` takes the release tag as agreed offline. If we want to change to consensus version this will require follow up refactor post v0.15 or we can also change it to take consensus version now.
As you prefer but software-version = stack version to me makes sense.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
goshawk-3: > proposal names a different release than the binary which is deterministic

If such proposaCoproUpgrade event occurs, we should reject the proposal immediately without activating either the upgrade or the dry-run. Cutover can be reached only after such validation has passed.

coprocessor/fhevm-engine/db-migration/migrations/20260819120000_add_consensus_version_to_versioning.sql
goshawk-3: I don't think we'll ever need the minor or patch components, as you mentioned. 

I mainly use `1.0.0` because incrementing the major version (X.0.0) explicitly signals a breaking change, which also matches the SemVer semantics.

coprocessor/fhevm-engine/fhevm-engine-common/src/lib.rs
goshawk-3: It's needed for convenience when testing locally and it can be behind a `consensus-version-override` that is never used in production.

When I want to test quickly a fix or feature against a running BCS (from e2e setup), I run a `run_fleet.sh`

```
run_fleet.sh exports BUILD_STACK_VERSION when you pass --stack-version, then cds into each service and runs its

- Each run_with_localnet.sh passes --features fhevm-engine-common/stack-version-override to cargo run. That is the only place the override appears in them.

```

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
PanGan21: Yes it makes sense. I changed the guard to be before cutover is reached in this commit: https://github.com/zama-ai/fhevm/pull/3580/changes/334e09600041198c9822495076d70b3e68999369

coprocessor/fhevm-engine/fhevm-engine-common/src/lib.rs
PanGan21: So the two things that made the override worth having are both gone because of the bootstrap versioning usage.

On this branch the default build is already the bumped one, 0.15.0 with consensus version 2, against 0.14.0 on main. So running the fleet against a BCS from the e2e setup is the default case now, not something you need a flag to reach. If you want a different version than the branch ships, it is one line in lib.rs plus the rebuild you were doing anyway. For example in your script now you can just remove the BUILD_STACK_VERSION override and it should just work.

In that sense, in which case you would still need to override the version? I am trying to understand the usage so I develop something that is actually useful.

coprocessor/fhevm-engine/db-migration/migrations/20260819120000_add_consensus_version_to_versioning.sql
PanGan21: I understand. But do you recommend using `1.0.0` string instead of a monotonic u32 `1` for the actual consensus version?

coprocessor/fhevm-engine/fhevm-engine-common/src/lib.rs
goshawk-3: I simply need to run this scenario (let know if it's already doable)

1. ./fhevm-cli deploy ( A low-spec, single-operator Copro that is considered BCS consensus-version 2.0.0 )
2.  ./run_fleet.sh - compiles from local src a copro (GCS) and bumps consensus  version 3.0.0

(Changing manually the compiled-in version is not convinient as it might be committed by mistake)

coprocessor/fhevm-engine/db-migration/migrations/20260819120000_add_consensus_version_to_versioning.sql
goshawk-3: For the sake of semantics and schema consistency, yes, but I don’t think we need to change anything at this point. That's why I resolved this conv

coprocessor/fhevm-engine/fhevm-engine-common/src/lib.rs
PanGan21: I will work on adding back the override to allow your setup. For now only for consensus version and if needed we can do it for stack version as well.

One thing worth flagging about that setup regardless of the flag: with an override in your setup, BCS is HEAD with a different version number, not an older release.
The test-suite setup pins BCS to `v0.14.0`, so it exercises the older binary and the migration is pinned in HEAD. An override built pair can't reach this setup. Fine for quick iterating on green side code, but it misses the realistic upgrade flow that will happen in prod.

PanGan21: Added both overrides, consensus alone is not enough, because a cutover also requires the stack version to be higher than BGS stack version. Commit: https://github.com/zama-ai/fhevm/pull/3580/commits/815c3a0d46697421a9339d777304cefee8567f6e

Your local setup should work like this:
```
# BCS — built from current branch
./fhevm-cli deploy --build

# GCS — the fleet, one version above
BUILD_STACK_VERSION=0.16.0 BUILD_CONSENSUS_VERSION=3 \
  cargo build --release --workspace \
  --features fhevm-engine-common/stack-version-override,fhevm-engine-common/consensus-version-override
./run_fleet.sh
```
Before, a fresh database got the v0.14 hard coded in the shipped migration, whatever binary you built. Now `bootstrap_versioning` writes the versions of the binary that ran the migration, so `deploy --build` gives you BCS at 0.15.0 / consensus 2 with nothing to edit.

Let me know if you agree with that flow, or if you see any potential gap/issue.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
PanGan21: Btw a follow up AI finding on that:
```
[P1] Revalidate the proposal at cutover, including after a restart.
[upgrade-controller/src/lib.rs:1845](https://github.com/zama-ai/fhevm/pull/3580/files)
The new release check runs only when entering DryRunStarted. A replacement controller finding DryRunStarted or UpgradeAuthorized resumes directly without that check. If it is a different build, cutover writes the proposal’s release alongside the replacement binary’s consensus version, and merges the replacement’s release-specific schema. This can promote a consensus version that never passed the proposal’s dry run. Validate the proposal against the executing binary again inside the cutover transaction, before changing versions or merging data.
```

Would you say it is possible for the upgrade-controller to restart for some reason in between and the new service being a different image version? Looks like a rare case but not sure if I should add the check defensively

PanGan21: I added a defensive check during cutover in this commit: https://github.com/zama-ai/fhevm/pull/3580/commits/03e0d40a922da1e9a64924f9f707e954b4bded1c

goshawk-3 (dismissed):
LGTM

PanGan21:
@Mergifyio queue

PanGan21:
@Mergifyio dequeue

goshawk-3 (dismissed):

fegmorte (dismissed):

PanGan21:
@Mergifyio queue

PanGan21:
@Mergifyio dequeue

goshawk-3 (approved):

fegmorte (approved):

PR #3591, opened 2026-08-25
refactor(host-contracts): optimize gas from byte-by-byte operations

isaacdecoded:
E2E tests passing [here](https://github.com/zama-ai/fhevm/actions/runs/34376304651)

Closes https://github.com/zama-ai/fhevm-internal/issues/1213


host-contracts/contracts/InputVerifier.sol
obatirou, line 266: Shouldn't this be in the natspec of the function?
There is no explicit definition of the inputProof layout and it was a challenge to verify the changes made here and understand what's going on in memory.
Link to SDK could also be interesting maybe ([here](https://github.com/zama-ai/fhevm/blob/6607ed115380b24d02c129d2bf4ed97d7942b996/sdk/js-sdk/src/core/coprocessor/InputProof-p.ts#L162-L245))
This could also help auditors when reviewing those changes

isaacdecoded: Good idea. I placed it as a NatSpec [here](https://github.com/zama-ai/fhevm/pull/3591/changes/f38152460e09549afb2753d57d997280eda6af1f#diff-cfdd06859b5fbcd907a11252dc94ef7ffb38c243eeecfcf094deab1541a50b92R233-R242). About adding a link to the SDK, I'm not sure, since the SDK code could change and it's hard to keep the link accurate.

host-contracts/docs/contract_selectors.txt
```diff
@@ -210,6 +210,13 @@ BridgeEvents
 | Event | LzReceivePerPayloadByteGasSet(uint32,uint64)   | 0x9365fb160bf0e4372e7c1b99983ee71b41cb5d94189d58d4097c5c5db0dc10eb |
 ╰-------+------------------------------------------------+--------------------------------------------------------------------╯
 
+BytesOps
```
obatirou, line 213: I am surprised to see it in selectors and rust bindings but I am no expert at it 🤔 
It is an internal library so should not be exposed no?

host-contracts/contracts/InputVerifier.sol
obatirou: Nice! 
Thanks a lot, it is way better for understanding what's going on.
I agree with your point with SDK, let's not add it.

host-contracts/docs/contract_selectors.txt
isaacdecoded: I got surprised too indeed. On the Gateway contracts we have the [HandleOps library also in the selectors](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/selectors.txt#L1421-L1427). Not sure if this is something we can configure to skip from the selectors/rust bindings though

host-contracts/contracts/shared/BytesOps.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.24;
+
+/**
+ * @title   BytesOps library
+ * @notice  Library that provides utility functions for bytes arrays.
+ */
+library BytesOps {
```
obatirou, line 10: I like this! 
But this goes against what pattern we currently have. 
This should be shared more broadly with the team so everyone work on this front when needed, what do you think?

host-contracts/contracts/InputVerifier.sol
```diff
@@ -236,6 +231,16 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
+     * @dev                 The inputProof is the concatenation of, in this order: numHandles (1 byte),
+     *                      numSigners (1 byte), the handles (32 bytes each), the coprocessor signatures
+     *                      (65 bytes each) and extraData (all the remaining bytes). Counting from the first
+     *                      byte of the inputProof, handle i starts at 2 + 32 * i, signature i starts at
+     *                      2 + 32 * numHandles + 65 * i and extraData starts at 2 + 32 * numHandles + 65 * numSigners.
+     *                      In memory a bytes value starts with its 32-byte length, which is why the assembly
+     *                      blocks below read handle i at inputProof + 34 + 32 * i.
+     *                      Each handle is keccak256(keccak256(bundleCiphertext) || index)[0:20] || index[21] ||
+     *                      chainId[22:29] || type[30] || version[31], where bundleCiphertext is
+     *                      compressedPackedCT + ZKPOK.
```
obatirou, line 241:
```suggestion
     *                      Each handle is keccak256(keccak256(bundleCiphertext) || index)[0:20] || index[21] ||
     *                      chainId[22:29] || type[30] || version[31]), where bundleCiphertext is
     *                      compressedPackedCT + ZKPOK.
```

host-contracts/contracts/InputVerifier.sol
```diff
@@ -153,17 +154,11 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
...
      */
     /// @custom:oz-upgrades-unsafe-allow missing-initializer-call
     /// @custom:oz-upgrades-validate-as-initializer
-    function reinitializeV2(
-        address[] memory newSignersSet,
-        uint256 threshold
-    ) public virtual reinitializer(REINITIALIZER_VERSION) {
-        defineNewContext(newSignersSet, threshold);
-    }
+    function reinitializeV3() public virtual reinitializer(REINITIALIZER_VERSION) {}
```
obatirou, line 161: Tasks should be updated also no?
As now the signature changed

host-contracts/contracts/InputVerifier.sol
```diff
@@ -322,9 +316,11 @@ contract InputVerifier is UUPSUpgradeableEmptyProxy, EIP712UpgradeableCrossChain
             uint8 numHandles = uint8(inputProof[0]);
             /// @dev We know inputProof is non-empty since it has been previously cached.
             if (numHandles <= indexHandle || indexHandle > 254) revert InvalidIndex();
+            /// @dev The proof cache is keyed on the full inputProof bytes, so a cache hit means these
+            ///      bytes already passed the length checks and the handle read below is in bounds.
             uint256 element;
-            for (uint256 j = 0; j < 32; j++) {
-                element |= uint256(uint8(inputProof[2 + indexHandle * 32 + j])) << (8 * (31 - j));
+            assembly {
+                element := mload(add(inputProof, add(34, mul(indexHandle, 32))))
```
obatirou: this could be `BytesOps.readBytes32(data, offset)` ?
this would deduplicate this call and the one above
at `Deserialize handle and check that they are from the correct version.`

But maybe in a following up issue

host-contracts/contracts/shared/BytesOps.sol
```diff
@@ -0,0 +1,23 @@
...
+     * @notice          Returns a copy of the size bytes of data starting at offset.
+     * @dev             The caller must have checked that the range lies within data.
+     * @param data      The bytes array to copy from.
+     * @param offset    The offset of the first byte to copy, counted from the first byte of data.
+     * @param size      The number of bytes to copy.
+     * @return result   The copied bytes.
+     */
+    function slice(bytes memory data, uint256 offset, uint256 size) internal pure returns (bytes memory result) {
+        result = new bytes(size);
+        assembly {
```
obatirou: you should be able to add "memory-safe" as it writes in a new memory

host-contracts/contracts/InputVerifier.sol
isaacdecoded: Really nice suggestion, implemented [here](https://github.com/zama-ai/fhevm/pull/3591/changes/0f2d55446a37a5ae4104013e9949dea8b5346b23#diff-1d2050d5a891019c27982c5fb17cde8a4dbc194cd5273b362280bafef01024e6R33-R37).

host-contracts/contracts/shared/BytesOps.sol
isaacdecoded: Nice catch, added [here](https://github.com/zama-ai/fhevm/pull/3591/changes/0f2d55446a37a5ae4104013e9949dea8b5346b23#diff-1d2050d5a891019c27982c5fb17cde8a4dbc194cd5273b362280bafef01024e6R21). Although, it seems it's not that impactful if viaIR is not used/activated.

host-contracts/contracts/InputVerifier.sol
isaacdecoded: [Indeed](https://github.com/zama-ai/fhevm/pull/3591/changes/0f2d55446a37a5ae4104013e9949dea8b5346b23#diff-2d3b1642261372dd8e4536c2c4014b88917ad2ec8db6705115dc5260f322115e), missed it during a main branch rebase. Thanks.

host-contracts/contracts/InputVerifier.sol
isaacdecoded: Not sure this suggestion is valid: the parenthesis is correctly closed already.

obatirou: Oh you are right
Sorry for the noise

jatZama (dismissed):
LGTM, good gas savings!

obatirou (approved):

isaacdecoded:
@mergifyio queue

PR #3593, opened 2026-08-25
chore(sdk): isolate js-sdk from root npm workspace

geoxel:
### What
Removes `sdk/js-sdk` and `sdk/js-sdk/src` from the root `package.json`
workspaces list, so the SDK is no longer built as part of the monorepo's
npm install/build graph. `sdk/js-sdk` now has its own standalone
`package-lock.json` and resolves its dependencies (including its own
`prettier` version pin) independently of the rest of the workspace.

`test-suite/e2e` no longer declares `@fhevm/sdk` in `package.json`.
Instead, a new `scripts/install-sdk.sh` helper installs it in place
(without touching `package.json`/`package-lock.json`), in one of two
modes selected by the Dockerfile's `SDK_SOURCE` build arg:

- **`local`** (default): build and pack `sdk/js-sdk`, install the
  tarball, so e2e tests exercise the local, unreleased SDK.
- **`registry`**: install a version from npm, passed explicitly via
  `SDK_VERSION` — there's no pinned default to fall back to.

`ethers` is added as a direct dependency of `test-suite/e2e` (it's
imported directly by e2e test code and by `@fhevm/sdk`'s `ethers`
adapter) and consistently installed with `--no-workspaces`, so both the
SDK and `ethers` land nested under `test-suite/e2e/node_modules` where
they can resolve each other, instead of `@fhevm/sdk` getting hoisted to
the repo root where it can't see `ethers`.

Also included: an `npm` `overrides` entry in `sdk/js-sdk/package.json`
forcing `nanoid` to `^5.1.16` wherever `estimo` (a transitive dependency
via `@size-limit/preset-big-lib`) requests it, fixing
`GHSA-28wg-ghj8-5hjv` / `GHSA-2v37-7h3g-55p8` (infinite loop on
negative/zero size) — no upstream `estimo` release fixes this yet. This
resolves the last `npm audit` finding in `js-sdk`.

### Why
Decouples js-sdk's dependency resolution from the rest of the monorepo
workspace and lets e2e Docker builds choose between testing against
local SDK changes or a released registry version, without either one
polluting the other's lockfile or install graph.

### Testing
- `cd sdk/js-sdk && npm ci` installs cleanly standalone.
- `test-suite/e2e`'s Docker build with `SDK_SOURCE=local` (default) packs
  and installs the local SDK; verified `ethers`/`@noble/*` resolve
  correctly at runtime (previously failed with `MODULE_NOT_FOUND`).
- `./scripts/install-sdk.sh registry` (no version) now fails fast with a
  clear error instead of silently resolving to `undefined`.


0xalexbel (approved):
LGTM!

tawadaa (approved):

PR #3595, opened 2026-08-26
fix(coprocessor): store all synthetic txn hashes so a reorg cannot le…

goshawk-3:
…ak synthetic rows

Injection at `start_block + 1` does not wait for finality, so a reorg makes the GCS host-listener re-inject on the replacement block: different block hash, different transaction hash, a second set of synthetic rows. The single-value `upgrade_state.synthetic_txn_hash` forgot the first set, and cutover then merged it into `public` as live data.

- Migration renamed to `synthetic_txn_hashes`: `BYTEA NOT NULL DEFAULT ''`, N 32-byte digests back to back, with a CHECK enforcing 32-byte alignment.
- host-listener appends instead of overwriting, using an offset-aligned membership test (not `position()`, which could match a needle straddling two hashes). Re-ingesting the same block stays idempotent.
- If the UPDATE matches no row (a concurrent FSM write moved it under READ COMMITTED), the block is skipped and no synthetic logs are returned - work without a marker would survive cutover.
- upgrade-controller unpacks the column with `generate_series(..., 32)` and deletes every set. A defensive length check refuses the delete if any stored hash is not 32 bytes, because `computations.transaction_id` carries short sentinels (`'\x00'`, `'\x01'`) that a misaligned predicate would match.
- Rollback and cutover reset the column to empty.


PanGan21 (approved):

goshawk-3:
@mergifyio queue

goshawk-3:
@mergifyio queue

PR #3597, opened 2026-08-26
feat(relayer): support pre-v0.14 protocol deployments

manoranjith:
Adds config-driven `/v2/keyurl` for the upcoming v0.15 release, so the relayer can run against protocol v0.13. The design is described in [fhevm-internal#1937][issue].

### How it works

The change is small. `/v2/keyurl` already serves whatever sits in a watch channel, so `config` mode seeds that channel at startup and starts no poller. The variant holds `KeyData`, the response type, so the config and the response cannot drift apart.

### Configuration from env vars

Most of the diff is that idea meeting env-var configuration, which is how these deployments pass keyurl values. Config keys are snake_case while the response is camelCase, so `data_id` needs an alias. config-rs also reads `..__URLS__0` as a map rather than a list, so `urls` needs the `deserialize_vec_from_map_or_seq` helper.

### Validation and tests

Static values reach clients untouched, so config load now rejects a malformed `data_id`, empty `urls`, or a URL that does not parse — none of which the issue asks for. Protocol v0.13 is simply two missing `ProtocolConfig` getters, so the tests revert both before the relayer boots and a stray poller fails the startup gate. Chain mode, the response shape and `openapi.yml` are unchanged.

Existing deployments need `keyurl.source` added, and static URLs are now a list, so the env key is `APP_KEYURL__FHE_PUBLIC_KEY__URLS__0` rather than `…__URL`.

[issue]: https://github.com/zama-ai/fhevm-internal/issues/1937

### DIFF
--- a/ci/preview-env/relayer/values-relayer-e2e.yaml
+++ b/ci/preview-env/relayer/values-relayer-e2e.yaml
@@ -210,14 +210,17 @@ env:
     valueFrom: {configMapKeyRef: {name: gw-sc-addresses, key: input_verification.address}}
 
   # --- keyurl (KMSGeneration poller) ---
-  # HEAD relayer replaced the old static fhe_public_key/crs storage-URL config
-  # with a poller that reads the active key/CRS/KMS context straight from the
-  # KMSGeneration contract on the host chain (settings.rs: KeyUrlConfig). It
-  # resolves the active key/CRS itself once the real DKG ceremony reaches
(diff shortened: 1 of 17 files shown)


nboisde (approved):
LGTM

manoranjith:
@mergify queue

manoranjith:
@mergifyio dequeue

manoranjith:
@mergifyio requeue

PR #3598, opened 2026-08-26
feat(coprocessor): runtime optimization

antoniupop:
Runtime optimizations to the worker scheduling path and the listener's dependence-chain
identity assignment, the benchmark and multi-worker harnesses used to measure them, and the
chart/config propagation for every flag involved.

No direct implementation of RFCs 19/20 which should already be covered, but they provide context and justify some of the design decisions.

## Summary

Four things, in dependency order:

1. **Listener** — a boundary handle consumed across blocks no longer collapses
   independent branches onto one DCID, and a cross-block parent that is still outstanding
   now arms the child's gate.
2. **Worker** — the worker acquires and releases a bounded *set* of
   dependence chains per cycle instead of one; GPU execution gains the accounting and
   backpressure that batching makes necessary; error classification becomes a rule with a
   reversible default and a demote-to-slow-lane path instead of condemnation.
3. **Harnesses** — a reportable one-shot benchmark driver over seven
   canonical scenarios, and a three-coprocessor consensus topology with GPU worker
   substitution.
4. **Configuration surface** — the full `tfhe_worker` CLI is now recorded with defaults in
   the Helm chart, guarded by a test.

---

## 1. Listener: cross-block chain identity

Both changes are on the ingest path and both are unconditional.

**Cross-batch fork detection.** A boundary handle produced in one block and consumed in a
later ingest batch used to make *every* consumer a linear continuation of the producer's
chain, serializing independent branches onto one DCID. A second LRU now records which
transaction claimed each boundary handle: the first cross-batch consumer still extends the
producer's chain, and every *different* later consumer forks into its own chain, gated on
the producer through `split_dependencies`.

**Cross-block parents arm the gate.** Previously only in-block dependencies contributed to
`dependency_count`, so a chain could be acquired while a cross-block producer was still
outstanding. Cross-block parents that are not yet `processed` are now locked, adopt the
child into their `dependents`, and each one actually updated adds to the child's gate.
Counting only the rows updated keeps this in the safe direction — a parent already
`processed` delivers no future decrement and must not be counted — and park-retry remains
the correctness backstop for every race.

The parent lock is `FOR UPDATE SKIP LOCKED`, not a plain `FOR UPDATE`. A parent row locked
by another session is precisely one a worker is releasing right now; skipping it
under-counts, which the gate tolerates by design. Waiting instead would hold the ingest
transaction behind an FHE batch release and close a deadlock cycle against the worker.

### Scheduling

`LockMngr` holds `locks: Vec<_>` instead of a single `Option`. A worker acquires up to
`--dependence-chains-per-batch` (20) chains per cycle under `--dcid-batch-execution` (on by
default); completed chains are released independently by `release_completed_locks`, so a
freed slot refills on the next cycle without waiting for the rest of the batch to drain.
The dependents decrement is multiplicity-aware — a child gated on two parents released
together must observe both decrements.

Four properties this has to get right, each of which cost a defect:

- **A RETRYABLE stamp is not terminal.** The completion test mirrors the work window
  exactly; treating an errored row as done would retire the chain as `processed` with
  `lock_expires_at = NULL`, a state no acquisition predicate can match — losing the
  computation outright.
- **Release + dependents' decrement are two statements in one transaction.** A child row
  the listener `INSERT`s while the release is in flight is invisible to a statement that
  started before it, so the single-statement form reads the parent's updated `dependents`
  array and then matches no row.
- **Every chain-row lock is taken up front, sorted, `FOR NO KEY UPDATE`,** over owned ids
  and their dependents. With a batch, two workers deadlock with no cycle in the DCID graph
  — just two cross-block edges pointing opposite ways. Measured 8/8 runs before, 0/8 after.
  Scoped to worker-vs-worker deliberately: the listener does not take the same order, so a
  worker/listener interleaving can still report `40P01`. That window is self-healing (one
  side aborts and retries a block or a release, neither partially applied), so it is a
  retry cost rather than a wedge. Closing it properly needs skip-or-defer semantics on the
  listener's upsert — follow-up work on the ingest path.
- **Discharge is scoped to the status flip.** Discharging on every drained release
  double-decrements a child when a listener refresh releases the chain as `updated` and it
  retires on a later cycle.

The timeslice is per lock and rotates what it evicts. Measured across the batch, one
never-finishing chain could evict 19 healthy ones; released with an unchanged
`last_updated_at`, the evicted chains returned to the FIFO front and were re-acquired
immediately.

### Terminal is now a rule, not a list

**Terminal iff the verdict is a pure function of immutable on-chain data** — the opcode,
the operand types, the graph shape — plus the cascade from a genuinely terminal producer.
The default arm is RETRYABLE, so a variant added to either error enum later fails towards
the reversible outcome. That is the opposite of what it replaces: `FhevmError` was
classified terminal wholesale, so a non-panicking `tfhe::Error` out of `compress()` got a
permanent verdict while the identical device fault arriving as a panic was retried.

**The rule has to cover the verdicts raised at EXECUTION time, not just the ones
`check_fhe_operand_types` catches before the graph is built** (`36caab9`).
`UnsupportedFheTypes` is the engine's most common type verdict and every one of its raise
sites is a match on the operation id and the operand type names, so it belongs on the
terminal side by the rule above. Leaving it to the retryable default arm did more than
retry it until demotion: the dead-producer predicate reads the stamp, so the errored
producer never counted as dead and its cross-transaction consumers deferred as
missing-inputs forever. The cast, rand and trivial-encryption type verdicts are classified
with it, and a unit test pins both polarities.

Other consequences: re-randomization failures (a property of the *key*) and all four GPU
reservation failures leave no stamp at all. Decompression failures are stamped retryable
and bounded. A panic describing resource pressure never spends retry budget.

The retryable marker is applied at the stamping decision rather than baked into an error's
`Display`, so the class lives in one place. That changes the marker text from
`ExecutionPanic` to `RETRYABLE`, and four predicates key on it. The migration carries no
backfill because that population is empty on every deployable baseline (re-selectable error
stamps arrived in `6e479ee2c` / #3560, not an ancestor of `release/0.14.x`); the migration
comment records the evidence and the scoping a backfill would need.

### Limbo instead of condemnation

Nothing is terminalised for running out of attempts. Exhausting
`--computation-retry-demote-threshold` (3, was a 20-attempt cap) **demotes** the row: the
work window stops selecting it and the chain's completion test stops counting it, so the
chain retires and its dependents discharge, while the row keeps its retryable stamp and
stays pending. An hourly sweep re-arms those chains at `SchedulePriority::Slow` with counts
reset, so a transient failure heals on a later pass and a permanent one costs a bounded
trickle instead of a verdict.

- The cone unblocks itself: `DEAD_PRODUCER_PREDICATE` treats a retryable stamp as not-dead
  and is recomputed every cycle, so consumers of a demoted row defer rather than drain.
- The slow lane already exists (the listener uses it for oversized dependency closures);
  its upsert raises priority with `GREATEST` and never lowers it, so a demotion survives a
  refresh.
- `is_resource_pressure` and its five hardcoded OOM substrings are **deleted** — they
  existed only to keep memory pressure away from a promotion that no longer exists.
- Retention will not delete a chain that still owns non-completed *pending or demoted*
  work. Terminal verdicts do not block deletion (they never heal, so they would accumulate
  an unbounded residue at the head of the `last_updated_at` scan).

New `computations.error_retry_count` bounds the one nondeterministic error class. The stamp
write is not the progress signal — Postgres matches a row on an identical-value `UPDATE`,
which would reset `no_progress_cycles` every cycle and disable the parking that paces a
failing panic — so the pre-image is read through a self-join and a genuine state transition
is reported instead.

**Dependents are pruned on retirement.** The array was append-only, so it grew without
bound; and because the decrement is keyed on retirement rather than on the arming, a chain
re-armed with new work and retired again decremented its historical children a second time,
opening a gate while a sibling producer was still unrun.

### GPU

- Memory reservations become an RAII guard with a timeout and typed errors, closing a
  permanent reservation leak on every caught panic, and reserve only the operation's own
  scratch since inputs are already resident.
- A process-wide per-device semaphore bounds concurrent partitions at
  `--gpu-streams-per-device` (16); dispatch was previously unbounded. The limiter is a
  `OnceLock` rather than a `Mutex<HashMap>` keyed by process-lifetime constants, which makes
  double admission over one device structurally impossible.
- `gpu_idx` is propagated truthfully — partitions previously ran under one device's key
  while telling the memory accounting they were device 0.
- **The reservation wait is bounded by the lease it runs under.** The wait is a blocking
  loop inside batch execution and the DCID lease is only renewed between worker cycles, so
  a 300 s timeout against a 30 s lease *guaranteed* the lease lapsed mid-batch and another
  worker recomputed the chain. The effective timeout is capped at
  `0.8 × --dcid-ttl-sec`, and the flag's default is 20 000 ms so it sits **under** that cap
  at the shipped TTL — see "Configuration" below.

### Boundary operands

Both backends put the canonical COMPRESSED form in the graph, and the executor memoizes
`Decompress(cmp(h))` per partition. RFC-020 permits exactly this; without it a boundary
handle consumed by K ops was decompressed K times from identical bytes (the CPU path did).

Memoizing in the executor rather than materializing up front buys three things beyond the
saving: `DFGTxInput::Value` keeps zero constructors, so the `dfg.rs` consensus tripwire
stays a live assertion on GPU as well as CPU; each value is decompressed under the
partition's own key, landing on the device that consumes it; and a ciphertext that fails to
decompress drains its own op through the terminal path instead of aborting the whole graph
with nothing stamped. Trade-off recorded: the memo is partition-scoped (a handle used in
two partitions is decompressed twice, in exchange for device affinity), decompression is
inline on the partition thread, and the memo is unbounded today.

### Acquisition hygiene and fairness

- `--dcid-adaptive-batch-execution` defaults to **true**, paired with
  `--dcid-batch-execution`; a test asserts they stay paired. **This changes the unit of
  `--work-items-batch-size`**: the adaptive window counts `(chain, transaction)` groups
  where the non-adaptive one counts computation rows, with no residual row cap. Deployments
  that set the flag explicitly need retuning.
- The adaptive window collapses to one row per transaction before its `LIMIT` — the ranking
  is per `(chain, transaction)`, so a transaction whose rows a fork-exposed database
  retained under a second chain was ranked once per chain and spent the window on duplicates
  of itself.
- A worker no longer re-acquires a chain it already holds. The work-stealing branch matches
  on expiry alone, so a lease that lapsed while still held in memory was stolen back by its
  own owner into a *second* lock-set entry; the renewal then returned one row for two ids
  and reported a permanent, false "Not all locks extended" for the life of the process.
- `release_locks` discharges dependents on the status flip that actually happened, not on
  its `mark_as_processed` parameter.
- Locks are handed back when the worker exits at cutover instead of waiting out
  `--dcid-ttl-sec` (the startup release filters on `worker_id`, a fresh UUID per process).
- `set_processing_error` writes to the chain that owns the failing computation instead of
  every chain the worker holds — with batching, one failure was spreading its message
  across ~20 unrelated chains' diagnostics.
- Range validators on `--work-items-batch-size` and `--dependence-chains-per-batch`
  (previously accepted zero and negatives); saturating multiply at the no-progress
  escalation threshold.
- A gauge beside the deferred-transactions counter, set every cycle including zero.

### Index

The acquisition index omits `is_error` from its partial predicate: the work window selects
`(is_error = false OR error_message LIKE '%RETRYABLE%')`, which does not imply
`is_error = false`, so an index carrying it can never be proven applicable. Verified on 200k
rows: Parallel Seq Scan before, Bitmap Index Scan after. It is a plain in-transaction
`CREATE INDEX` because the migration images and CI pin sqlx-cli 0.7.2 (no `no_tx` support);
the concurrent build happens ahead of the migration via `precreate_index` in
`initialize_db.sh`.

## 3. Benchmark and multi-worker harnesses

The reportable "Main one-shot" benchmark driver defines **seven canonical scenarios**
covering the shapes the scheduler actually sees — independent work, dependent chains,
cross-transaction dependence, an auction, two lagged traffic mixes, and
`skewed_chains_250_1x200_50x1`. The traffic ones are staged through the real host-listener
ingest path rather than hand-written `dependence_chain` rows, so DCID construction is
exercised rather than assumed.

The skewed scenario exists because every other scenario is uniform, and a uniform shape
cannot show either side of the adaptive work window. That window gives each acquired chain
`ceil(work_items_batch_size / acquired_chains)` transactions and does not redistribute what
a chain cannot use, so with 50 of 51 chains supplying exactly one transaction the window
fills to roughly `share + 50` — at bench defaults, ~80 transactions where 1500 were
budgeted. The same fixture measures what the window is *for*: with it off, the long chain
owns the oldest `schedule_order` rows and fills the window while the short chains wait out
the timeslice.

A reportable run refuses to start without a clean worktree and an explicitly named
dedicated database, and records its provenance alongside the results. Bench targets are
gated behind the `bench` feature and excluded from autobenches, so nothing here reaches the
shipped binaries. One piece of production surface is added deliberately:
`start_runtime_with_readiness` lets the harness start a real worker in-process and wait for
a deterministic ready signal, keeping startup outside the measured interval. Production
passes `None` and the channel never fires.

The multi-worker side adds a 3-of-3 unanimous topology and a launcher that substitutes
host-built GPU binaries for the Docker CPU workers on a pinned device, plus the e2e
public-runtime build path that makes a local stack reproducible from public base images.
The two new Docker stages default to the existing certified bases, so image builds are
unchanged unless the new target is selected explicitly.

## Reviewer notes

- **Four test-suite tooling changes are unconditional** and affect every stack run, not only
  consensus runs: a docker socket bind-mount, a server-key blob added to the bootstrap
  readiness probe, a kms-connector `BUILD_ID` freshness check, and a hard error for any
  multi-coprocessor scenario without a local build override.
- **`--work-items-batch-size` changes meaning** under the new adaptive default. An A/B
  against the non-adaptive mode needs retuning first — the two modes count different units,
  so the same number is not the same window. `BENCH_DCID_ADAPTIVE_BATCH_EXECUTION=false`
  measures the non-adaptive arm.
- **Migrations:** `20260814120000_pending_computations_dcid_acquisition_index.sql`,
  `20260828120000_computations_error_retry_count.sql`. No backfill, by the argument recorded
  in the migration comment.
- **New flags:** `--dcid-batch-execution` (on), `--dcid-adaptive-batch-execution` (on),
  `--computation-retry-demote-threshold` (3), `--gpu-memory-reservation-timeout-ms` (20000),
  `--gpu-streams-per-device` (16). `--dependence-chains-per-batch` (20) is not new but is
  newly load-bearing.
- **Deadlock scope:** the sorted pre-lock closes worker-vs-worker only. Worker-vs-listener
  `40P01` remains possible and self-healing; the ingest-side fix is follow-up.

## Verification

```
cargo test --benches --features bench            # keeps the bench targets compiled in test mode
cargo test -p tfhe-worker --lib daemon_cli       # flag defaults + chart drift guard
make print_bench_one_shot_scenarios
make benchmark_erc20_main_block_one_shot_suite_cpu
make benchmark_erc20_main_block_one_shot_suite_gpu
```

`cargo bench` never sets `cfg(test)` and clippy builds bench targets in bench mode, so
`cargo test --benches --features bench` is the command that catches a bench target that
compiles for benching but not for testing; it is part of the recipe for that reason.


coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -1900,6 +1917,68 @@ impl Database {
...
+                .collect::<Vec<_>>();
+            let gated_outer_count = if chain.new_chain
+                && !outer_dependencies.is_empty()
+            {
+                sqlx::query_scalar!(
+                    r#"
+                    WITH parents AS (
+                        SELECT dependence_chain_id
+                        FROM dependence_chain
+                        WHERE dependence_chain_id = ANY($1::bytea[])
```
Eikix: [AUTOMATED] [P1] Gate this fork on the required producer, not the whole DCID

`dependency_count` is incremented whenever the parent DCID is not `processed`. That DCID is reused by later linear continuations, so new tail transactions can keep setting it to `updated` after this child's specific boundary handle has already been persisted. The child is discharged only when the entire parent later changes from `processing` to `processed`; a continuously extended parent can therefore leave this independent fork unacquirable indefinitely. Readiness needs to be recorded or discharged for the specific boundary producer rather than for retirement of the whole mutable chain.

Confidence: 95/100

coprocessor/fhevm-engine/tfhe-worker/src/dependence_chain.rs
```diff
@@ -668,6 +1141,16 @@ impl LockMngr {
 
         if should_run_cleanup {
             self.last_cleanup_at = Some(SystemTime::now());
+            // Before deleting: give demoted rows their next pass. Running the
+            // sweep first means a chain re-armed here is no longer 'processed'
+            // when the delete runs, so the two cannot race over it.
+            let rearmed = rearm_demoted_chains(&self.pool, self.demote_threshold).await?;
```
Eikix: [AUTOMATED] [P1] Keep demoted work recoverable in lockless mode

This rearm is unreachable when `--disable-dcid-locking` is enabled because `do_cleanup` returns at the start of the function. The lockless work query still excludes a retryable row once `error_retry_count` reaches the threshold, and no other failure path resets that count. A transient failure that reaches the threshold in the documented fallback mode is therefore never selected again, even after the underlying fault clears. The rearm must remain active without locking, or demotion must be bypassed in that mode.

Confidence: 99/100

coprocessor/fhevm-engine/tfhe-worker/src/daemon_cli.rs
```diff
@@ -31,13 +31,62 @@ pub struct Args {
...
+    /// Small on purpose: the old value of 20 existed because exhausting the
+    /// budget used to promote the row to TERMINAL, which condemns the whole
+    /// downstream cone. Nothing is terminalised for running out of attempts
+    /// any more, so there is no reason to be generous.
+    #[arg(
+        long,
+        env = "FHEVM_COMPUTATION_RETRY_DEMOTE_THRESHOLD",
+        default_value_t = 3
+    )]
+    pub computation_retry_demote_threshold: i16,
```
Eikix, line 89: [AUTOMATED] [P2] Reject non-positive demotion thresholds

This option accepts zero and negative values, while selection requires `error_retry_count < threshold` and rearming resets the count to zero. With a non-positive value, retryable work can never be selected after being stamped, including after a rearm. A positive-range parser should reject this configuration at startup.

Confidence: 100/100

Eikix (changes requested):
[AUTOMATED]

## Automated code review

This review was generated automatically against commit `52c8931`. All published PR conversation comments, submitted reviews, and inline review comments were checked immediately before preparation; no existing human findings or duplicates were found.

Two blocking liveness findings and one non-blocking configuration finding are attached.

coprocessor/fhevm-engine/tfhe-worker/src/daemon_cli.rs
antoniupop: Yes - and it's not only selection: at 0, error_retry_count >= 0 also matches every row. Fixing

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
antoniupop: Correct - fixing this so the gate only arms on the boundary handle if it's not materialised yet and sealing the producer chain so it can't extend after a dependent is identified. This bounds the wait time and the only risk is a delay.

rudy-6-4:
Just  to confirm. One copro (in multi copro) can choose any dcid working policy. There is no need to have the same concurency tree between copros. So the "first" on 1 side don't need to be the first on another copro.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -72,13 +72,39 @@ pub struct Chain {
...
+/// materialized. Without this the producer keeps absorbing continuations of
+/// its own unrelated traffic, and the fork — gated on the chain, not on the
+/// handle — waits behind all of it. Sealing stops the growth, so the fork
+/// waits only for the producer's undrained tail at seal time; the next
+/// continuation starts its own chain gated on the sealed one.
+///
+/// In memory and LRU-bounded on purpose: losing an entry re-enables extension,
+/// which is exactly the pre-existing behaviour, so a restart or an eviction
+/// degrades to "slower to discharge", never to "wrong".
+pub type SealedChainGuard = Arc<RwLock<lru::LruCache<ChainHash, ()>>>;
```
Eikix, line 115: [AUTOMATED] [P1] Rebuild the producer seal during restart catchup

`sealed_chains` is process-local and starts empty. On a normal restart, the default catchup replays the last five blocks and calls `dependence_chains()` before duplicate computation inserts are rejected. That replay rebuilds the producer-handle mappings and consumed-boundary state, but the duplicate inserts leave `at_least_one_insertion` false, so `update_dependence_chain()`—the only place that restores this seal—is skipped. A later consumer of the rebuilt tail mapping can therefore extend the parent again; sustained tail traffic can keep the already-gated child waiting indefinitely for that parent to retire. The seal needs to be persisted or reconstructed during duplicate catchup, with a restart regression test covering this sequence.

Confidence: 94/100

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -1627,32 +2285,88 @@ async fn query_for_work<'a>(
...
+    // and `rearm_demoted_chains` is what gives it another pass. That sweep
+    // runs from `do_cleanup`, which returns immediately when locking is
+    // disabled, and it re-arms through the chain lifecycle (`status =
+    // 'processed' AND worker_id IS NULL`) that lockless mode does not drive.
+    // So with `--disable-dcid-locking` there is nothing to reset the count,
+    // and a bounded pause would become a permanent one: a transient failure
+    // that spent its attempts would never be selected again, even after the
+    // fault cleared. The documented fallback mode must not be able to lose
+    // work, so demotion does not apply there.
+    let demote_threshold = if args.disable_dcid_locking {
```
Eikix: [AUTOMATED] [P2] Do not encode disabled demotion as `i16::MAX`

The retry counter saturates at `32767`, which is the same value used here as the lockless threshold, while both work selectors require `error_retry_count < demote_threshold`. After 32,768 failed executions the count reaches `32767`, the row becomes permanently ineligible, and lockless mode still has no rearm sweep to restore it after the underlying fault clears. The retry-count predicate should be bypassed explicitly when locking is disabled instead of representing “disabled” with a finite sentinel.

Confidence: 100/100

Eikix (changes requested):
[AUTOMATED]

## Automated follow-up review

Head `c3eee6d` was re-reviewed after the previous threads were resolved. All published comments and submitted reviews were checked again before posting. These comments cover residual failure paths in the new fixes and do not duplicate the earlier review.

One blocking restart liveness issue and one lockless-mode availability issue remain.

coprocessor/fhevm-engine/host-listener/.sqlx/query-e71590a0aad41b62e79acbe36b530aa0929f532beda19b43fbe4d94b53fd5381.json
```diff
@@ -0,0 +1,24 @@
+{
```
rudy-6-4, line 1: what is the policy regarding .sqlx. We have a centralized one and some crate have their own. Should we split the centralized one ?

antoniupop: For a while I thought we moved to purely centralised, but can't say we ever were strict on that. Does it make any real difference?

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -234,18 +262,105 @@ impl Database {
...
+        .await;
+        let rows = match rows {
+            Ok(rows) => rows,
+            Err(error) => {
+                warn!(
+                    %error,
+                    "could not rebuild the sealed-chain set; producers may be \
+                     extended while a fork waits on them"
+                );
+                return;
```
Eikix: [AUTOMATED] [P1] Require seal reconstruction before startup continues

Any SQL error here is only logged, after which initialization succeeds with an empty `sealed_chains`; no later retry is scheduled. Catchup can still rebuild producer routing from duplicate blocks, so after a transient startup query failure a continuation can be folded back into an unsealed producer while its existing child remains gated. Under sustained continuation traffic, that child can again wait indefinitely. The reload should be retried or its error propagated so ingestion cannot start without the required seal state.

Confidence: 96/100

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -234,18 +262,105 @@ impl Database {
...
+            }
+        };
+        let mut sealed = self.sealed_chains.write().await;
+        let mut restored = 0_usize;
+        for row in rows {
```
Eikix: [AUTOMATED] [P2] Preserve the reconstructed seal recency order

The query returns the newest producers first, but each `LruCache::put` makes the inserted entry most-recently used. Iterating in this order therefore leaves the first—and newest—recovered producer as the least-recently used entry. When the restored cache is full, the next distinct seal evicts that newest producer first; a later continuation can then extend it while its child remains gated. The returned rows should be inserted oldest-to-newest, for example by reversing the iteration.

Confidence: 94/100

Eikix (changes requested):
[AUTOMATED]

## Automated follow-up review

Head `f9b28a2` was reviewed after the latest fixes. All published conversation comments, submitted reviews, and inline review threads were checked immediately before posting. These findings cover distinct residual paths in the new seal-reconstruction implementation and do not duplicate existing review threads.

One blocking startup-failure path and one narrower cache-ordering issue remain.

coprocessor/fhevm-engine/host-listener/.sqlx/query-e71590a0aad41b62e79acbe36b530aa0929f532beda19b43fbe4d94b53fd5381.json
rudy-6-4: Not important, just less clear now which command to run to generate them.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -1695,9 +2503,98 @@ async fn query_for_work<'a>(
...
+  -- Like the non-adaptive query below: load ALL rows of the selected
+  -- transactions, fork-retained siblings included; ownership is re-applied
+  -- in code, which turns rows of other chains into recompute-only
+  -- producers excluded from results, persistence and completion.
+            ",
+            dcid_filter.as_deref(),
+            dcid_transaction_share(transaction_batch_size, dependence_chain_ids.len(), true,)
+                as i64,
+            transaction_batch_size as i64,
+            &cooled_transactions,
```
PanGan21: Suppose operation B depends on operation A in the same transaction. After A reaches its retry limit, B is still pending, so the whole transaction is loaded again including A. A is retried even though it should have been moved to the slow retry queue, while B remains pending with `MissingInputs`. I understand that this can repeat forever.
Would it make sense to move B, and any other operations in the same transaction that depend on A, to the slow retry queue as well?

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -2302,7 +3242,14 @@ async fn build_transaction_graph_and_execute<'a>(
             keys.pks.clone(),
             #[cfg(feature = "gpu")]
             keys.gpu_sks.clone(),
+            #[cfg(feature = "gpu")]
+            gpu_execution_limiter,
             health_check.activity_heartbeat.clone(),
+            // The worker has no per-batch cancellation; the token exists so a
+            // GPU memory reservation wait can be interrupted when one is
+            // introduced, and reservation waits stay bounded by the timeout.
+            tokio_util::sync::CancellationToken::new(),
```
PanGan21: So this token will not be cancelled now right? And will be used in the future only?

antoniupop: Yes, correct. This was introduced for lease renewal, but now we're not going to have that in this PR (too many things) so I'm removing it and will come in with the use itself, makes more sense.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
antoniupop: Good catch - this is a regression of no longer making terminal all dependents trnsitively. Fixing, but I'm more in favour of moving full transactions. That was the intent initially, but the base object is the computation and that's why it broke.

coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -60,27 +80,80 @@ struct WorkItem {
...
+/// terminal. A retryable stamp is a statement about the failure, not about
+/// the operands, so no amount of patience running out turns it into a
+/// verdict.
+///
+/// Substring matching is admittedly coarse, and it is now load-bearing in
+/// five places (this constant, the two work-window queries,
+/// `release_completed_locks` and [`DEAD_PRODUCER_PREDICATE`]). A dedicated
+/// error-class column is the follow-up; until then, every one of those five
+/// sites must be changed together.
+pub(crate) const RETRYABLE_STAMP_MARKER: &str = "RETRYABLE";
```
Eikix: [AUTOMATED] this is a substring protocol with no type, and LockMngr now imports it from the cycle file.

Retryability is load-bearing in at least five SQL/code sites (this constant, both work-window queries, `release_completed_locks`, `DEAD_PRODUCER_PREDICATE`). Then `is_terminal_verdict` became a second classifier, then GPU reservation became a third path that skips stamping. A later reader changing one site silently desyncs the others. That is how this file keeps growing.

i think there's a code-judo move here that makes this much simpler. one module next to `types.rs`:

```rust
enum StampDisposition { LeavePending, Retryable, Terminal }
fn disposition(err: &CoprocessorError) -> StampDisposition { ... }
```

`upload_transaction_graph_results` becomes `match disposition(...)`. GPU no-stamp is `LeavePending`. Delete `is_retryable_gpu_reservation_error`, `is_terminal_verdict`, and the upload `continue` chain. Work-window SQL, completion, rearm, and dead-producer import one SQL fragment from that module — not from `tfhe_worker.rs`.

A dedicated error-class column can come later. Shipping five `LIKE` sites while calling the column a follow-up is the wrong order.

coprocessor/fhevm-engine/tfhe-worker/src/daemon_cli.rs
```diff
@@ -31,13 +31,63 @@ pub struct Args {
...
+    /// non-adaptive one counts computation rows, with no residual row cap. A
+    /// value tuned for rows therefore admits roughly
+    /// `work_items_batch_size * average rows per transaction` rows here, so
+    /// the two flags must be retuned together.
+    #[arg(
+        long,
+        env = "FHEVM_DCID_ADAPTIVE_BATCH_EXECUTION",
+        default_value_t = true
+    )]
+    pub dcid_adaptive_batch_execution: bool,
```
Eikix: [AUTOMATED] two bools that the comments already admit are one mode.

They default ON, "disable BOTH or NEITHER," and changing one retunes `--work-items-batch-size` because they count different units. `batching_and_its_fairness_mitigation_ship_together` only asserts the *defaults* match. Runtime still accepts mixed configs.

They are also clap presence-only switches that default ON, so Helm `extraArgs` cannot turn them off. Tests already construct the forbidden pairing (`dcid_adaptive_batch_execution: false` with batching on).

One enum, clap `value_enum`:

```text
DcidExecution { Sequential, BatchedFair }
```

Default `BatchedFair`. Reject `dependence_chains_per_batch > work_items_batch_size` at parse time, not with a startup warn. That also lets you delete the non-adaptive work-window query in `query_for_work`.

`Args` is now a 40-field clap dump (161 → 512). Nested structs (`WorkerLoop`, `DcidLocking`, `GpuRuntime`) behind one `Args` would stop benches/tests reconstructing the struct by hand and bypassing clap — they currently set `dcid_cleanup_interval_sec: 0`, which `parse_positive_u32` rejects on the real CLI.

Eikix (dismissed):
[AUTOMATED]

## Thermo-nuclear code quality review

Behavior here is unusually well-documented. Structure is not. The new scheduler, stamp class, and chain-identity protocols were inlined into already-god files instead of becoming types. That is a maintainability regression even though the races this PR caught are real.

**Not mergeable at this quality bar** until the incidental control-plane is deleted, not rearranged.

### Blockers

1. **Stamp class is a substring protocol.** `LIKE '%RETRYABLE%'` is load-bearing in five-plus sites; `is_terminal_verdict` and GPU no-stamp are two more taxonomies. `LockMngr` importing the marker from `tfhe_worker.rs` inverts the layer. One `StampDisposition { LeavePending, Retryable, Terminal }` deletes the upload `continue` chain.
2. **`Vec` of locks bolted onto a still-singular API.** `get_current_lock` returns `.first()`, `park_current_lock` / defer-rotate operate on the whole set, and the cycle reimplements refill beside `LockMngr`. Extract `HeldLocks` + `ensure_held`; park only the stuck chain; move slow-lane rearm out of this type. That split also brings `dependence_chain.rs` back under 1k (761 → 1384).
3. **Four listener caches answer one identity question.** `past_chains` / `consumed_boundaries` / `sealed_chains` / `outer_boundary_handles`, plus a forever `tokio::spawn` from `Database::new`. One `ChainIdentity` + `Gate` + `Extension`. Construction does not spawn.
4. **Two bools, two work-window SQLs, two schedulers.** Defaults already ship as one mode. `enum DcidExecution { Sequential, BatchedFair }` and one SELECT. Copy the `Ciphertext128Format` pattern this same PR already uses.
5. **File-size bar.** Four files crossed 1k (`dependence_chain.rs`, `tests/dependence_chain.rs`, `benches/erc20.rs`, `benches/utils.rs`). `tfhe_worker.rs` grew 2529 → 3763. Split before adding more protocol to them.

### Keep

GPU reservation RAII and `GpuExecutionLimiter` are the right shape. The two migrations are tight. The comments are the spec — types should make the invariants they document unrepresentable.

A targeted extraction that preserves SQL and scheduling behavior is enough. Keeping two work-window queries, four identity caches, and five `LIKE '%RETRYABLE%'` sites is not.

Eikix (dismissed):

antoniupop:
@mergifyio queue

Eikix:
[AUTOMATED]

## Non-blocking follow-up review

This automated review was run at head `cc710a0a55`. Existing comments and reviews were checked again. These points were not already posted.

This is not a request to change this PR. The approval stays, and the PR can merge. These cases could be handled in follow-up PRs in the stack before 0.15 goes live.

1. **Missing an earlier event can store the wrong operand mask.**

   The new two-listener test removes the last operation. The unsafe direction is the opposite: transaction `T` has `A` then `B(A)`, but the first listener sees only `B`. It stores `B` with boundary bits `0b11`. The full catch-up view gets the correct local bits `0b00`, but `ON CONFLICT DO NOTHING` keeps the first value. The worker then follows the wrong source rule. This was reproduced by reversing the missed event in the new test: the stored value was `3` instead of `0`.

   A follow-up could test this direction and add a safe way to check or repair the mask when the full transaction is later seen. A plain last-write-wins update would only reverse the race.

   Confidence: 98/100

2. **An internal operation failure can miss the retry path.**

   Errors returned before `OpResult` is built are saved only when the node is allowed. If an internal operation fails while reading an input or panics, its error is dropped. Its allowed consumer then reports `MissingInputs`, which upload ignores. A follow-up could make sure the pending allowed work enters backoff or demotion instead of staying in the normal retry path.

   Confidence: 96/100

3. **Raising the retry limit can strand work from the old limit.**

   For example, a row can be moved to the slow lane at limit `3`. If the worker restarts with limit `5` before the sweep, its stored count no longer matches the sweep. Its chain is already `processed`, so it cannot be picked up normally, and cleanup now keeps it. A follow-up could re-arm any retryable row on an unowned processed chain without comparing it with the new limit, or store the slow-lane state directly.

   Confidence: 98/100

4. **GPU reservation failures skip the lockless cooldown.**

   The upload path returns early for these failures before the transaction is added to the cooldown list. With DCID locking disabled, the same oldest transaction can be selected again after every timeout and keep younger work out. A follow-up could send these transaction IDs through the same cooldown path.

   Confidence: 99/100

Would it make sense to track these in the next PRs in the stack?

antoniupop:
> ## Non-blocking follow-up review
> [...]
> 
> Would it make sense to track these in the next PRs in the stack?

I'll fix these in a follow up PR and add that directly to the stack. Good finds, thanks!

Eikix (commented):
[AUTOMATED]

## Automated head re-review

Head `9e3db6e1c7` was reviewed after the latest push. Existing comments and reviews were checked again. This updates the earlier automated follow-up; resolved points are not repeated. This is a comment review, not a request for changes.

The retry-limit change is fixed, and same-handle results are now assigned to the correct transaction. The earlier missing-event operand-mask issue is unchanged.

Three follow-ups remain:

1. **An internal operation error is now kept by the scheduler, but it still cannot be saved to its database row.**

   The listener stores a non-allowed internal row with `is_completed = true`. `set_computation_error` only updates rows with `is_completed = false`, so the update affects no row. The allowed consumer then returns `MissingInputs`, which is ignored, and no retry count or demotion is reached.

   This was reproduced with the same row state produced by the listener. The new demotion test uses `is_completed = false`, so it does not cover the production state.

2. **The 60-second lockless cooldown can expire before the next failing GPU batch finishes.**

   With the default batch size of 100, 16 streams, and a 20-second reservation timeout, one GPU can take about 140 seconds to finish a full failing batch. Two old groups can then alternate after each other's cooldown expires, while younger work that would fit is never selected. This only affects `--disable-dcid-locking`; locking is enabled by default.

   A queue or cursor would give the needed fairness. Increasing the timeout alone would still depend on batch duration.

3. **A seal refresh can remove a seal created by an ingest transaction that has not committed yet.**

   Ingest adds the seal to memory before commit. A refresh on another database connection cannot see that uncommitted gate, treats the seal as discharged, and removes it. After commit, this listener stays unsealed until the next refresh, so unrelated work can extend the parent and delay the gated child.

   This exact two-connection order was reproduced. Publishing the seal after commit, or protecting in-flight seals from refresh removal, would close the race.

Would it make sense to finish these here, or place them in the next PR in the stack before 0.15 is deployed?

antoniupop:
First point yes it's a bug introduced in the last round of fixes, so fixing. Other two are borderline, one is probably inaccurate the other an edge case with no correctness impact, just a mild perf degradation in very improbable conditions so I'll look a bit more into those before pushing a fix.

Confirmed, first item fixed, other two are overstated and not necessary.

Eikix (commented):
[AUTOMATED]

## [P1] Scalar bytes are treated as dependency handles

The earlier internal-producer case is handled by this head, but one unsafe edge remains in the new recursive fallback.

`computations.dependencies` stores both encrypted handles and plain scalar bytes. `propagate_error_to_dependents` follows every byte value equal to the failed handle.

For example:

- internal operation `P` produces handle `H`;
- a real consumer depends on encrypted `H`;
- an independent scalar operation uses `H` as its plain scalar value.

If `P` fails, both operations are found by the recursive query. The independent scalar operation can then be marked with a permanent error and skipped forever, even though it does not depend on `P`.

This was reproduced at the current head with a negative database test. The new positive test passes, but the scalar-alias test fails.

Could the error be propagated using the typed worker graph, or could the SQL follow only non-scalar operands whose boundary bit is `0`? A negative test where scalar bytes equal the failed handle would cover this case.

Confidence: 99/100

Eikix:
[AUTOMATED]

## Automated head review

Head `3fe6e0e` was reviewed. Existing comments and reviews were checked before posting, and these points were not found there. This is a follow-up comment, not a request for changes. The latest commit fixes the earlier case where an operation itself returns an error.

Three follow-ups remain:

1. **Errors raised before a foreign producer returns a result are still dropped.**

   The new snapshot now covers every producer. However, if a foreign producer fails while loading or decompressing an input, the scheduler reports the error only when that producer is owned. The error is dropped before the snapshot can be used. Its owned consumer keeps returning `MissingInputs`, and the chain can repeat forever.

   A foreign error should be passed only to the owned allowed consumers from the snapshot. Its foreign database row should not be updated directly.

   Confidence: 97/100

2. **The adaptive row cap can hide newer valid transactions forever.**

   The cap is applied to the oldest raw rows before distinct transactions are ranked. One large transaction that always waits for an input can fill the whole scan. A newer valid transaction in the same chain is then never seen. Releasing and taking the chain again starts from the same rows.

   This was reproduced across two work cycles. The cap could be applied after enough distinct transactions are found, or scan progress could be saved.

   Confidence: 99/100

3. **Old 0.14 cycle errors are not repaired for the 0.15 upgrade.**

   The new code makes future cycle errors retryable, but 0.14 stored some of these errors as final. Those rows survive the blue-green switch. They can block the dry run or cause a later GCS consumer to be marked dead.

   An upgrade migration could clear these old errors and re-arm or recreate their dependence-chain work before the dry run.

   Confidence: 98/100

Would it make sense to cover these in the next PR in the stack before 0.15 is deployed, so this merge does not need to wait?

antoniupop:
I've fixed the first two, last one is wrong and dangerous- that fix would likely block the B/G upgrade.

Eikix:
[AUTOMATED]

## Full review at `d3d92be`

The entire PR was reviewed again from scratch, including the listener, scheduler, worker, migrations, GPU code, benchmarks and deployment tests. The 0.15-only rollout and separate blue-green schemas were taken into account. Existing comments, reviews and inline feedback were checked; the five findings below were not already posted.

This is a follow-up comment, not a request for changes. No approval is changed. Could these be covered in follow-up PRs in the stack before 0.15 is deployed?

### Correction to the previous review

The foreign-producer error and adaptive row-cap findings are fixed at this head.

The earlier recommendation to migrate historical 0.14 cycle errors is withdrawn. A regression requiring that repair was not established, and replaying historical work during blue-green was not a justified recommendation. That suggested migration should not be applied based on the earlier automated review.

### New findings

1. **[P1] GPU memory checks can admit an operation that crashes the worker.**

   [tfhe_ops.rs:908–915](https://github.com/zama-ai/fhevm/blob/d3d92beaeee7abbb75322f8c1a8a2453809bff14/coprocessor/fhevm-engine/fhevm-engine-common/src/tfhe_ops.rs#L908-L915)

   The extra ciphertext-sized reservation is removed on the assumption that TFHE's estimate includes the output. For `FheNot`, pinned TFHE 1.6.3 estimates scratch memory but separately allocates a full ciphertext copy. With enough free memory for scratch but not the copy, the check passes and allocation fails. The CUDA handler aborts the entire worker, bypassing Rust error recovery.

   Output memory needs to be included in the reservation. This was independently verified against the pinned dependency source, not reproduced on GPU hardware: [output copy](https://github.com/zama-ai/tfhe-rs/blob/1a747bef9decd86d105a7661d6f6b0a9f8a93d87/tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs#L1205-L1209), [estimate](https://github.com/zama-ai/tfhe-rs/blob/1a747bef9decd86d105a7661d6f6b0a9f8a93d87/tfhe/src/integer/gpu/server_key/radix/bitwise_op.rs#L1326-L1380), [abort handler](https://github.com/zama-ai/tfhe-rs/blob/1a747bef9decd86d105a7661d6f6b0a9f8a93d87/backends/tfhe-cuda-backend/cuda/include/device.h#L13-L18).

2. **[P2] Existing benchmarks stop making progress.**

   [benches/utils.rs:189](https://github.com/zama-ai/fhevm/blob/d3d92beaeee7abbb75322f8c1a8a2453809bff14/coprocessor/fhevm-engine/tfhe-worker/benches/utils.rs#L189)

   Dependence-chain locking is now enabled in shared setup. Existing ERC20, DEX and synthetic fixtures insert computations without creating the required chain rows. Nothing is acquired, and the benchmarks keep waiting. Only the new main-block paths create those rows.

   Either the missing chain rows need to be added to those fixtures, or their previous locking mode needs to be preserved. This was reproduced using the real acquisition and cleanup code against an isolated database.

3. **[P2] Benchmark reset can delete the wrong database.**

   [Makefile:97–99](https://github.com/zama-ai/fhevm/blob/d3d92beaeee7abbb75322f8c1a8a2453809bff14/coprocessor/fhevm-engine/tfhe-worker/Makefile#L97-L99)

   Migration uses configurable `BENCH_DB_URL`, but reset always drops `coprocessor`. With a dedicated benchmark database, its old data is left intact while the default database is deleted. Reset, migration and execution need to use the same explicit target.

   This was verified with a dry run using a different database name. No database was deleted.

4. **[P2] The KMS public-runtime option is ignored.**

   [compose.ts:173–179](https://github.com/zama-ai/fhevm/blob/d3d92beaeee7abbb75322f8c1a8a2453809bff14/test-suite/fhevm/src/generate/compose.ts#L173-L179)

   `KMS_CONNECTOR_RUNTIME_BASE_IMAGE` is passed by Compose, but it is never declared or used in `kms-connector/Dockerfile.workspace`. All three runtime stages still use the private Chainguard image. The public-runtime build therefore still requires private-registry access.

   The option needs to be implemented in the Dockerfile. Current tests only check the generated Compose arguments.

5. **[P2] The stream-count test does not compare different stream counts.**

   [materialization.rs:283–287](https://github.com/zama-ai/fhevm/blob/d3d92beaeee7abbb75322f8c1a8a2453809bff14/coprocessor/fhevm-engine/tfhe-worker/src/tests/materialization.rs#L283-L287)

   The test runs with one stream, then changes the setting to sixteen in the same process. However, the worker's first concurrency limit is stored in a process-wide `OnceLock`. Both runs retain the first limit, so the test can pass without exercising the promised concurrency difference.

   Separate processes would allow both settings to be tested.

### Known follow-ups, not new findings

- The previously discussed incomplete-log case remains: an incorrect operand mask can be kept by the first successful insert.
- A child can still wait after its input is ready because unrelated work keeps the producer chain pending. Sealing prevents further growth but does not remove an existing pending tail. The delay depends on the workload; this is not evidence of incorrect ciphertexts.

### Checks

Passed: 24 worker tests, 8 scheduler tests, 73 common-library tests, 365 harness tests, TypeScript checks and the additional isolated-database regression. GPU execution and full end-to-end deployment were not run.

Eikix (approved):
LGTM

antoniupop:
@mergifyio queue

PR #3601, opened 2026-08-26
fix(relayer): stop expiry deleting unresolved requests

manoranjith:
Closes zama-ai/fhevm-internal#1938.

The request delete and the share delete are one statement rather than two, so a request can never lose its shares or outlive them. `ON DELETE CASCADE` would be the natural way to express that, but `user_decrypt_req.gw_reference_id` is nullable and not unique, so it cannot be the target of a foreign key without changing that first.

This leaves a gap rather than closing it: a request stuck in `queued`, `processing` or `tx_in_flight` now survives instead of being deleted, but nothing yet reports it. Only `receipt_received` has a staleness detector.

This path has no test coverage, before or after. The correlated delete and the orphan sweep were verified by hand against a live database inside a rolled-back transaction — a resolved request and its share go together, a live request keeps both, an orphaned share is collected.


nboisde (approved):

manoranjith:
@mergify queue

manoranjith:
@mergifyio dequeue

manoranjith:
@mergifyio requeue

PR #3603, opened 2026-08-26
chore(kms-connector): kms context in-memory caching

eudelins-zama:
### Summary

The kms-worker validated the KMS context of every request with a DB query on the hot path. This PR replaces it with an in-memory cache of the `kms_context` and `kms_epoch` tables, periodically refreshed from the DB.

Closes zama-ai/fhevm-internal#1090

### Changes

- New `KmsContextCache`: a TTL'd in-memory snapshot of the context/epoch tables, following the same pattern as the `CoprocessorRegistry`. Reads are lock-free after a snapshot clone; a background task reloads it on a configurable interval.
- `DbContextManager` now checks requests against the cache instead of querying the DB. Unknown pairs still fall back to the on-chain validation, and a pair confirmed on-chain is inserted in the cache (and the DB) right away, so following requests hit the cache without waiting for the next refresh.
- Validities are never overridden in-memory: a destruction event processed concurrently is not reverted, only the epoch's context association is repaired (mirroring the DB upsert).
- New config field `kms_context_cache_refresh` (env: `KMS_CONNECTOR_KMS_CONTEXT_CACHE_REFRESH`, defaults to 60s).

Note: invalidations written by the gw-listener are picked up at the next refresh, so a destroyed context/epoch can still validate for up to one refresh interval.


kms-connector/crates/kms-worker/src/core/event_processor/context/cache.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    /// Registers a pair confirmed valid on-chain.
+    ///
+    /// This way, following requests conclude from the cache instead of re-validating on-chain
+    /// until the next cache refresh.
+    pub fn insert_valid_pair(&self, context_id: U256, epoch_id: U256) {
+        let Ok(mut guard) = self.snapshot.write() else {
+            // Let the background task handle the lock poisoning, only warning here.
+            return warn!("KMS context cache lock poisoned, not caching pair");
```
dvdplm, line 157: This reads weird to me. It looks like we're returning the output of `warn!` (which I guess is `()` so it's fine); in reality we want to 1. log a warning and *then* 2. return. So I'd split into two lines.

kms-connector/crates/kms-worker/src/core/event_processor/context/cache.rs
```diff
@@ -0,0 +1,342 @@
...
+        };
+        let mut snapshot = (**guard).clone();
+
+        // We don't override the context and epoch `is_valid` flags, in case a destruction event
+        // is processed between the on-chain read and this write.
```
dvdplm, line 162: I *think* I understand this but not 100% sure. Can you ELI5 what this protects against?

kms-connector/crates/kms-worker/src/core/event_processor/context/cache.rs
```diff
@@ -0,0 +1,342 @@
...
+                })
+                .collect(),
+        }
+    }
+
+    #[test]
+    fn check_concludes_from_cached_pairs() {
+        let snapshot = snapshot(
+            &[(1, true), (2, false)],
+            &[(10, Some(1), true), (11, Some(1), false), (12, None, false)],
```
dvdplm, line 243:
```suggestion
            // Contexts: list of `(context_id, is_valid)`-tuples
            &[(1, true), (2, false)],
            // Epochs: list of `(epoch_id, context_id, is_valid)`-tuples
            &[(10, Some(1), true), (11, Some(1), false), (12, None, false)],
```

kms-connector/crates/kms-worker/src/core/event_processor/context/manager.rs
```diff
@@ -175,7 +138,12 @@ impl<P: Provider> DbContextManager<P> {
     }
 
     /// Caches a pair confirmed valid on-chain, so subsequent requests skip the RPC calls.
+    ///
+    /// The pair is registered in memory (effective immediately) and persisted in the DB (to
+    /// survive restarts, and as the source the in-memory cache refreshes from).
     async fn cache_valid_pair(&self, context_id: U256, epoch_id: U256) {
+        self.context_cache.insert_valid_pair(context_id, epoch_id);
+
```
dvdplm, line 146: Shouldn't this be moved *after* the data is written to the DB? To like line 175, once we know `query_result` is `Ok`?

kms-connector/crates/kms-worker/src/core/config.rs
```diff
@@ -393,6 +405,7 @@ mod tests {
             env::set_var("KMS_CONNECTOR_S3_MAX_CONCURRENT_GETS", "8");
             env::set_var("KMS_CONNECTOR_S3_MAX_CIPHERTEXT_SIZE", "1048576");
             env::set_var("KMS_CONNECTOR_COPRO_REGISTRY_REFRESH", "90s");
+            env::set_var("KMS_CONNECTOR_KMS_CONTEXT_CACHE_REFRESH", "30s");
```
dvdplm, line 408: Seems a bit odd to not use the default 60s here?

kms-connector/crates/kms-worker/src/core/event_processor/context/cache.rs
```diff
@@ -0,0 +1,342 @@
...
+            .read()
+            .expect("KMS context cache lock poisoned")
+            .clone()
+    }
+
+    /// Registers a pair confirmed valid on-chain.
+    ///
+    /// This way, following requests conclude from the cache instead of re-validating on-chain
+    /// until the next cache refresh.
+    pub fn insert_valid_pair(&self, context_id: U256, epoch_id: U256) {
```
dvdplm, line 154: How do we know these passed values are valid? I don't think we do; it seems like it's the callers responsibility to check this, but it's a `pub` fn so that is not enforced anywhere?

I'd rename to `insert_pair` and document that it is assumed that the values passed are valid.

dvdplm (commented):
The code here is fine, and contexts/epochs change rarely enough to make a cache like this viable.

That said I wouldn't be honest if I didn't raise a few objections. 

- More code, more things could go wrong. Caching means introducing a new layer where things could go wrong. This PR is not presenting a "transparent cache", but it is explicitly allowed to go stale for 1 refresh cycle. 
- Values read from the cache before an update will remain live in the system and used until their operation is concluded. This isn't necessarily a big deal but it is something to keep in mind; the day we run 5 connectors all using one postgres DB it might be problematic when their caches are out of sync.
- No cache invalidation on epoch/context destruction I think? Ideally all mutations should also refresh the cache (either in toto or only relevant entries).
- Did you consider any off-the-shelf solutions, e.g. the `sqlx-cache` crate?
- The cache here is impl'd with 2 HashMaps but they are both behind the same lock, which means they can't be queried independently. If they are never used that way, I wonder if there isn't some nifty way to combine them into a single HashMap (dunno, group all `(epoch, is_valid)` tuples under their context?).
- Did you try a concurrent hash map like the `papaya` crate, to get rid of the lock?
- Isn't there a non-poisoning lock we can use?

Finally, I'd be curious to see what performance would be like if you just cached the current context and epoch in a thread-local variable and didn't bother with the full table-cache shenanigans.

kms-connector/crates/kms-worker/src/core/event_processor/context/cache.rs
eudelins-zama: It's on purpose, to save one line of code (and the style is OK for me). We use this pattern all over the place in the connector :sweat_smile: .

But if really an anti-pattern to you, I can make the change, but might worth doing it everywhere in a follow-up PR then?

kms-connector/crates/kms-worker/src/core/event_processor/context/cache.rs
eudelins-zama: 1. We receive a request with a given context/epoch `(c, e)`, not registered in cache
2. We check on-chain, `(c, e)` is valid. Let's say the call to `insert_valid_pair` is not done immediatly (tokio scheduling)
3. Concurrently, `(c, e)` is destroyed, thus in DB, `(c, e)` is not valid anymore
4. `KmsContextCache` is loaded with `(c, e)` not being valid
5. We finally perform `insert_valid_pair`

In this edge case (that would probably never happen lol), we don't want to override the `is_valid` flags for `(c, e)`.

kms-connector/crates/kms-worker/src/core/event_processor/context/manager.rs
eudelins-zama: I don't have a strong opinion:
- the "real" source of truth for context is Ethereum. So makes sense to do the cheap, infallible operation of updating the cache as soon as we receive a context/epoch switch event, then update the DB. This way, even if the DB insert fails, our cache is correctly aligned with Ethereum (temporarily at least, before it gets refreshed)
- on the other hand, the cache is supposed to be sourced from DB, so we could argue that if the DB insert fails, we don't update the cache to keep it aligned

kms-connector/crates/kms-worker/src/core/config.rs
eudelins-zama: This is for a test where we modify the value via its env var, to see if it correctly override the default value, so it's on purpose here :sweat_smile:

kms-connector/crates/kms-worker/src/core/event_processor/context/cache.rs
dvdplm: Got it. You decide if it's worth adding this detail to the comment!

kms-connector/crates/kms-worker/src/core/event_processor/context/manager.rs
dvdplm: Yeah, I guess it doesn't matter. If we have this clear hierarchy though (eth –> DB –> cache) then perhaps it's more consistent to stick to that order as much as possible?

kms-connector/crates/kms-worker/src/core/event_processor/context/cache.rs
dvdplm: You're the boss in this code base, so your preference is what matters. Ignore me. :)

eudelins-zama:
Converting to draft for now. We don't need it right now, I'll keep it to run some bench and maybe use it later

PR #3639, opened 2026-08-27
feat(kms-connector): kms-connector api

eudelins-zama:
First draft for the Relayer <-> Connector interface

Closes https://github.com/zama-ai/fhevm-internal/issues/1950


kms-connector/crates/api/src/error.rs
```diff
@@ -0,0 +1,99 @@
...
+            | Self::CiphertextNotFound
+            | Self::UpstreamTransient => true,
+        }
+    }
+}
+
+/// The `v1` error body.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ErrorResponse {
```
nboisde, line 68: ErrorCode forward is great here thanks ! 
But thinking out loud, could be great additionnaly this error responses comes from the KMS connector and not the proxy, by adding a origin field (We could in the relayer trigger different behaviour if we detect 401). Maybe ErrorCode will be sufficient here, wdyt ?

kms-connector/crates/api/tests/vectors.json
```diff
@@ -0,0 +1,89 @@
+{
```
nboisde, line 1: Thanks for the useful examples 🙏

nboisde (dismissed):
LGTM @eudelins-zama, nice one, thanks ☝️🙏

kms-connector/crates/api/src/error.rs
eudelins-zama: I like the idea :slightly_smiling_face: but just want to make sure we really want it:

In theory, the error code alone is enough (we don't even need the `retryable` in theory for ex).
So right now, it's more for human readability, but also for the edge case where relayer api version is different from connector one, in that case the retryable field could be the source of truth for the retry logic.

For the origin, I guess it would be purely human readability, as indeed relayer could infer from the error code.

TL;DR: no strong opinion, as you wish haha

eudelins-zama: I'll say for now YAGNI.

We can add it later if we see this is needed, this would not break backward compat at all so we should be good.

nboisde: Totally agree, lets keep it minimal for now, and we will be able to add fields before making it permissionless I'd say !

nboisde (approved):
LGTM, thanks simon for implementing this.
Will be able to start relayer-http out of this 🙏

kms-connector/crates/api/src/error.rs
```diff
@@ -0,0 +1,128 @@
...
+impl ErrorCode {
+    /// HTTP status associated with this code.
+    pub fn http_status(self) -> u16 {
+        match self {
+            Self::Malformed => 400,
+            Self::SenderAuthenticationFailed => 401,
+            Self::AclDenied | Self::UserSignatureRejected => 403,
+            Self::CiphertextNotFound => 404,
+            Self::KmsContextDestroyed => 410,
+            Self::KmsContextInvalid => 412,
```
jot2re, line 52: Does this mean that the KMS Context was not found or the data in it is not valid? Since you return 412 it seems to indicate that it does not exist (also makes sense since it is retryable). Maybe the name is a bit misleading then? Should it maybe be `KmsContextNotFound`, in that case a 404 might also make more sense

kms-connector/crates/api/src/types.rs
```diff
@@ -0,0 +1,221 @@
...
+    sol_types::{Eip712Domain, SolStruct},
+};
+use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
+
+/// EIP-712 domain for the `v1` decryption interface.
+pub const DECRYPTION_EIP712_DOMAIN: Eip712Domain = Eip712Domain {
+    name: Some(Cow::Borrowed("kms-connector-decryption")),
+    version: Some(Cow::Borrowed("1")),
+    chain_id: None,
```
jot2re, line 20: Should these values not be set when computing the request ID?

kms-connector/crates/api/src/types.rs
```diff
@@ -0,0 +1,221 @@
...
+    struct HandleEntry {
+        bytes32 handle;
+        address contractAddress;
+        address ownerAddress;
+    }
+
+    /// The user-signed validity window of a user-decryption request. Timestamps are Unix seconds.
+    #[derive(Debug, PartialEq, Serialize, Deserialize)]
+    #[serde(deny_unknown_fields)]
+    struct RequestValidity {
```
jot2re, line 46: Nitpick: According to claude `RequestValidity` is used on-chain to express days, so it might safer to call this `RequestValiditySeconds`.

jot2re (approved):
Overall looks great to me! Just a couple of tiny comments

kms-connector/crates/api/src/error.rs
eudelins-zama: This would be detailed in the `message` field of the `ErrorResponse` struct.

Did not wanted to bloat too many error code for now. And we can still add new code later, this is not a breaking change for the API :slightly_smiling_face:

kms-connector/crates/api/src/types.rs
eudelins-zama: These values are used for the request ID computation.

But they should be hard-coded, every request ID computation should use the same EIP-712 domain.

kms-connector/crates/api/src/types.rs
eudelins-zama: `RequestValiditySeconds` exists because Solidity does not allow two types with the same name to coexist in a scope, so we needed to keep `RequestValidity` in days for backward compat with the old relayer-sdk.
If it was up to us, we would have used `RequestValidity` (in seconds) everywhere.

Here we do not have this constraint, so I think this makes sense to keep it like this.

eudelins-zama:
@mergifyio queue

PR #3640, opened 2026-08-27
fix(sdk): harden FHE key handling

jfrery:
## Summary

Hardens FHE encryption-key handling by preventing relayer-supplied public-key and CRS bytes from being cached, deserialized, or used before authentication.

Production encryption now fails closed unless the client provides either:

- `fheEncryptionKeyTrust` containing trusted public-key and CRS digests, or
- caller-pinned `fheEncryptionKey` bytes.

Pinned keys are checked against any explicit trust configuration before cache admission. Cleartext clients retain their explicit unauthenticated path.

The cache now stores authenticated bytes independently of the TFHE version, while deserialized WASM entries remain isolated by exact TFHE version. Key material is also cloned before caching or returning it to callers to prevent mutation of trusted entries.

## Compatibility

Applications performing encryption must configure a trust anchor independent of the relayer response. Digest resolvers can support authenticated key rotation without recreating the client.

## Verification

- Focused FHE key-authentication tests: 17 passed
- Full unit suite: 926 passed
- ESLint and TypeScript checks passed
- Formatting and declaration builds passed

ref https://github.com/zama-ai/sdk-internal/issues/1


jfrery:
@claude review the latest commit for security issues. Focus on whether FHE encryption-key trust remains independent of the relayer, whether localstack trust discovery fails closed, and whether any authentication/cache bypass or regression remains. Report only concrete, actionable findings.

jfrery:
@claude independently review the latest commit now as an SDK maintainer. Inspect the actual diff for correctness, minimality, repository conventions, regression risk, test quality, and whether the LocalStack v11/v12/v13/latest CI fix is precise. Do not rely on or refer to an earlier review; report only concrete, actionable findings, or explicitly say none.

sdk/js-sdk/src/core/types/coreFhevmClient.ts
```diff
@@ -15,19 +15,27 @@ export type FhevmBaseOptions = {
...
+  FhevmEncryptionKeyOptions & {
+    readonly moduleVersions?: FhevmEncryptModuleVersions | undefined;
+  };
 
-export type FhevmOptions = FhevmBaseOptions & {
-  readonly fheEncryptionKey?: FheEncryptionKeyBytes | undefined;
-  readonly moduleVersions?: FhevmModuleVersions | undefined;
-};
+export type FhevmDecryptOptions = FhevmBaseOptions &
+  FhevmEncryptionKeyOptions & {
```
geoxel, line 31: I guess we should not have any FhevmEncryptionKeyOptions for decryption only?

sdk/js-sdk/examples/node-ethers/env.local.example
```diff
@@ -0,0 +1,7 @@
+# Optional: omit to use a random wallet.
+WALLET_PRIVATE_KEY=
+
+# Required for encryption. Obtain these pins from finalized KMSGeneration state
+# or another authenticated chain-operator source, never from the relayer response.
+FHEVM_PUBLIC_KEY_DIGEST=<0x-prefixed 32-byte digest>
+FHEVM_CRS_DIGEST=<0x-prefixed 32-byte digest>
```
geoxel: The fhevm/sdk should query the two digests directly from the KMSGeneration contract, through the `getKeyMaterials` and `getCrsMaterials` view functions. See previously proposed fix for the ZamaSDK at https://github.com/zama-ai/sdk/pull/569/changes#diff-c369b9d1a4c8e54a720c31f041d7cea02ff1678c136b2893281d5cf1157740feR79

To be checked: can the KMSGeneration contract address be read elsewhere on chain? Like the FHEVMExecutor contract address that can be read from `getFHEVMExecutorAddress` from the `ACL.sol` contract.
If not, then it has to be added to the fhevm/sdk configs, in `sdk/js-sdk/src/core/chains/definitions` and  `sdk/js-sdk/test/chains`, the same way `protocolConfig` address has been added recently.

Open question: if the KMSGeneration contract address is not specified in the config, should the digests check be skipped? With a warning in the diagnostic logs? (I guess yes)

geoxel (changes requested):

PR #3644, opened 2026-08-27
feat(sdk): add polygon mainnet CLI network

fegmorte:
## Summary

Add `--network polygon` to `fhevm-sdk` so the CLI can target Polygon PoS mainnet (chain id 137) with the shared mainnet relayer and the deployed FHETest contract.

## Changes

- Add the `polygon` network preset (`POLYGON_RPC_URL`, host chain 137, FHETest `0xFb10eda9e9b4f3f7dd928B6F32fBB94E2a20451d`)
- Wire completion and load-test network lists
- Record the address in `sdk/js-sdk/test/chains/chain-defaults.json`

## Test plan

- [ ] `fhevm-sdk -n polygon fhe-test info` resolves chain id 137 and the FHETest address above
- [ ] `POLYGON_RPC_URL` overrides the default RPC
- [ ] Tab completion lists `polygon`

### DIFF
--- a/sdk/js-sdk/test/chains/chain-defaults.json
+++ b/sdk/js-sdk/test/chains/chain-defaults.json
@@ -83,6 +83,11 @@
     "fheTestAddress": "0xa66bCEd74D1Df0736d0eb8E52371b1b1AAA1F0F0",
     "fheTestVersion": "v2"
   },
+  "polygon_mainnet": {
+    "rpcUrl": "https://polygon-rpc.com",
+    "fheTestAddress": "0xFb10eda9e9b4f3f7dd928B6F32fBB94E2a20451d",
+    "fheTestVersion": "v2"
+  },
   "ingen_trex_cleartext": {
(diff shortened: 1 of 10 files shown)


0xalexbel (approved):
Appart from the polygon addresses that  I did not double check
LGTM!

tawadaa (approved):

sdk/js-sdk/test/chains/chain-defaults.json
```diff
@@ -83,6 +83,11 @@
     "fheTestAddress": "0xa66bCEd74D1Df0736d0eb8E52371b1b1AAA1F0F0",
     "fheTestVersion": "v2"
   },
+  "polygon_mainnet": {
+    "rpcUrl": "https://polygon-rpc.com",
```
zmalatrax, line 87: I suppose it's correct but wanted to confirm it's the expected rpc url

fegmorte: Yep it's the good default one

PR #3646, opened 2026-08-27
fix(coprocessor): wait for the GCS schema before opening a green pool

PanGan21:
## Summary

- The host-listener and gw-listener pin `search_path` to the versioned GCS schema, outside the activation gate that guards the workers. Postgres ignores a missing schema in `search_path` rather than erroring, so a green service that starts before the upgrade-controller has created that schema resolves its writes to `public` instead, which is the live blue tables.
- Both now wait for the schema before opening their pool, and the pool manager runs the same check for services that build their pool through it.
- The rows written on that path today are decoded from host-chain events, every insert is `ON CONFLICT` guarded, and blue’s listener writes the same rows concurrently, so the present effect is duplicate ingestion rather than divergence.

fixes https://github.com/zama-ai/fhevm-internal/issues/1943

### DIFF
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
@@ -224,6 +224,68 @@ pub struct StaleStackError {
     pub live: String,
 }
 
+/// Fail unless the GCS schema exists.
+///
+/// Green services pin `search_path` to that schema, and Postgres ignores it when
+/// missing, so writes would go to `public`. The upgrade-controller creates it at
+/// startup; a green service that starts first must wait. Call this before
+/// opening a pool.
+pub async fn assert_gcs_schema_exists(database_url: &str) -> anyhow::Result<()> {
+    let options = crate::database::connect_options_for_database_url(
+        &crate::utils::DatabaseURL::from(database_url),
+    )
+    .await?;
+    let mut conn = PgConnection::connect_with(&options).await?;
+    let exists: bool =
+        sqlx::query_scalar("SELECT EXISTS(SELECT 1 FROM pg_namespace WHERE nspname = $1)")
+            .bind(crate::database::GCS_SCHEMA)
+            .fetch_one(&mut conn)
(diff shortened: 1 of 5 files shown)


coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
```diff
@@ -224,6 +224,68 @@ pub struct StaleStackError {
...
+/// Every green service starts alongside the upgrade-controller that creates the
+/// schema, so losing that race is normal and must not end the process. Give up
+/// only after `timeout`, which means the controller never got there.
+pub async fn wait_for_gcs_schema(database_url: &str, timeout: Duration) -> anyhow::Result<()> {
+    const POLL: Duration = Duration::from_secs(2);
+    let mut waited = Duration::ZERO;
+    loop {
+        match assert_gcs_schema_exists(database_url).await {
+            Ok(()) => return Ok(()),
+            Err(err) if waited >= timeout => {
```
goshawk-3, line 282: I'd rather make this never times out or at least increase the default timeout significantly.

If an operator has a misconfigured upgrade-controller, some of the GCS services could silently exit before the issue is fixed.

goshawk-3 (changes requested):

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
PanGan21: I raised the timeout to an hour like you suggested here: https://github.com/zama-ai/fhevm/pull/3646/commits/a1cce08ab3b4a5e7bfe090e01d772ccbe3f416ac

I looked at dropping the timeout entirely, but gw-listener and host-listener wait before their health port opens, so they would sit `NotReady` forever instead of crash-looping. Keeping a timeout means a broken deploy still ends in `CrashLoopBackOff`, which is harder to miss. 
Let me know if you agree with this rationale

goshawk-3 (approved):

PanGan21:
@mergify queue

PR #3647, opened 2026-08-27
fix(relayer): stop losing gateway events on restart

manoranjith:
Closes zama-ai/fhevm-internal#1944.

First of three stacked PRs. #3648 adds SIGTERM handling on top of this, and #3649 adds single-dispatcher election on top of that.

Three commits, best read in order. The first splits log delivery so each log reaches one handler; the second is the substance, holding the recorded chain position back until handlers complete; the third is independent of both and can be reviewed on its own.

## Highlights for review

- **A re-observed event is dispatched a second time rather than awaited, because response handling is idempotent.** All five gateway events are responses, and each handler's status guard makes a repeat a no-op — the second dispatch finds the row already past that status and does nothing. Awaiting the first task instead would hold that listener's read loop behind a single racing event, which is the additional reason not to.
- **The claim now happens before transaction preparation.** Previously a caller that lost the claim and then failed gas estimation could write `failure` to the row while the winner's transaction was live on chain.

## Test plan

Local counts were measured at the top of the three-PR stack, not at this PR's tip in isolation.

- [ ] CI: clippy `-D warnings` and `cargo fmt -p fhevm-relayer --check`, both clean locally.
- [ ] CI: `sqlx-check` — this PR adds a migration.


relayer/src/gateway/arbitrum/listener.rs
```diff
@@ -414,16 +415,27 @@ impl ArbitrumListener {
                                 "Event received"
                             );
 
+                            let Some(gateway_event) =
+                                gateway_chain_event_for_log(event_log.clone(), tx_hash)
+                            else {
+                                debug!(
```
nboisde: Isn't supposed to never happens and hence be an error there, even if we keep continue ? 
Maybe not.

relayer/src/gateway/arbitrum/polling_listener.rs
```diff
@@ -394,16 +395,25 @@ impl PollingListener {
             block_number, block_hash, log_index, topic0, topic1, tx_hash
         );
 
+        let Some(gateway_event) = gateway_chain_event_for_log(event_log.clone(), tx_hash) else {
+            debug!(
```
nboisde: Same remark here, maybe not relevant.
Maybe since its only a skip its ok to keep debug.

relayer/relayer-migrate/migrations/20260824120000_create_gateway_chain_cursor.sql
```diff
@@ -0,0 +1,36 @@
+-- One cursor for the relayer: every gateway chain event up to this block has been handled.
+--
+-- It replaces the per-listener rows of `gateway_block_number_store`, whose key was a
+-- listener's position in the config list, so reordering that list handed one listener
+-- another's position. Listeners are redundant readers of the same chain and deduplicate
+-- against each other, so one row says all there is to say.
+--
+-- `gateway_block_number_store` is left in place, untouched: a rolled-back relayer still
+-- reads and writes it.
+CREATE TABLE gateway_chain_cursor (
```
nboisde, line 10: Question comes to my mind regarding one edge case, but maybe its taken care of here: 
How this is behaving, if two relayer hasn't the same configuration files, and for instance relayer 1 is having only two listeners, and relayer 2 have the full listener suite ?

relayer/src/config/settings.rs
```diff
@@ -183,20 +184,20 @@ pub struct ListenerPoolConfig {
...
     /// Polling interval in milliseconds (for polling type listeners)
     pub poll_interval_ms: u64,
-    /// TTL for event deduplication cache in seconds (1-10)
-    pub dedup_ttl_seconds: u64,
-    /// Maximum capacity for deduplication cache
+    /// How long the event registry remembers an event, in seconds (1-10).
     ///
-    /// **Sizing guidance:**
-    /// The cache should accommodate all events received during the TTL window with a safety buffer.
+    /// It bounds two things: how long the same log observed by two listener instances is
```
nboisde, line 194: 👍

relayer/src/config/settings.rs
```diff
@@ -841,6 +842,21 @@ impl Settings {
             )));
         }
 
+        // Only the polling listener replays blocks missed while the relayer was not
```
nboisde, line 858: Good to know there.

relayer/src/gateway/arbitrum/listener.rs
```diff
@@ -491,16 +329,17 @@ impl ArbitrumListener {
         }
     }
 
-    /// Creates a log subscription for the given provider and starting block.
+    /// Creates a log subscription for the given provider.
```
nboisde, line 332: Hence the poller to fill gaps.
Makes more sense now. 🙏

relayer/src/gateway/arbitrum/polling_listener.rs
```diff
@@ -23,162 +21,78 @@ use std::{str::FromStr, sync::Arc, time::Duration};
...
-                    .insert_initial_block_info(
-                        current_block_from_rpc,
-                        block_hash_from_rpc,
-                        self.instance_id,
-                    )
-                    .await?;
-                current_block_from_rpc
-            }
-        };
+        if let Some(recorded) = self.chain_cursor_repo.get().await? {
```
nboisde, line 79: Question here, would it be better to start from DB registered tip, and fall back to config if there is nothing present inside ? 
Not sure what is the behaviour here, but I think it could loop over and over on the same block if we don't update the config each time.

relayer/src/gateway/handled_events.rs
```diff
@@ -0,0 +1,488 @@
...
+    /// entry's TTL, so the window in which a duplicate is recognized runs from completion.
+    async fn complete(&self, key: EventKey) {
+        self.entries.insert(key, Phase::Done).await;
+    }
+}
+
+/// One attested range, and the dispatches that must finish before the cursor may pass it.
+struct PendingRange {
+    to_block: u64,
+    outstanding: Arc<AtomicUsize>,
```
nboisde, line 129: Don't get the rationale behind this variable.

relayer/src/gateway/arbitrum/transaction/helper.rs
```diff
@@ -51,9 +51,24 @@ impl fmt::Display for TransactionType {
     }
```
nboisde, line 51: Seems we are going back regarding what "in-flight" means, since we are now updating even before pure broadcasting, fine on my side as you already know.

nboisde: Fits perfectly with Claimed by the engine, and keep the change minimal. 
Thanks.

nboisde (commented):
Some little concerns here and there to clarify. 
Thanks mano for this work regarding HA 🙌

manoranjith:
Bringing a point from our discussion onto the PR so the answer lives with the code:

> TTL versus handler duration. moka's `time_to_live` runs from the insert, and `or_insert` in `claim` is that insert. With `dedup_ttl_seconds` capped at 10, a handler slower than the TTL has its `InFlight` entry evicted, so the next poll sees a fresh entry and dispatches a third copy. The config comment says the window "restarts when they finish", which is true of `complete`'s insert but not of survival while running.

Thanks for flagging this — good thing to pin down.

The cache holds a dedup marker only; the dispatch itself runs as its own detached task (`handled_events.rs:269`), and a skip needs a *present* entry, so an eviction can at most cause an extra dispatch. The cursor is separate: it advances only once `outstanding` reaches zero (`:319`), so anything unfinished leaves it where it is and gets replayed after a restart.

Two improvements came out of your comment (91d542416): the abandon-on-panic path now carries `alert = true`, and `PhaseCache` documents what an eviction costs — `InFlight`/`Done` reads like a work lifecycle, so it's better stated than left to be inferred.

Leaving the TTL at 5s; it's configurable up to 10s if you'd prefer more margin.

relayer/src/gateway/arbitrum/listener.rs
manoranjith: Thanks for spotting this — it's a fine point. Two conditions were sharing that one `debug!`, but they should have been handled differently, so I applied the following change:

- **A log we can't decode** → `warn!`
- **A response we can't match** → `info!` plus a counter

The decode branch should never happen. I also removed a cause of it: the five signatures were listed twice, in the polling filter and in the resolver. Both now read one table, which let the WebSocket subscription narrow to those five as well.

An unmatched response is different — it can simply belong to another relayer's request, e.g. a local instance on the same gateway. Within our HA pair it's a non-issue, since both pods share the database. It should stay rare today, so `relayer_gateway_unmatched_events_total` carries the rate and the log line gives the detail.

(9d835188b)

relayer/src/gateway/arbitrum/polling_listener.rs
manoranjith: Answered on the same remark in `listener.rs` — both branches changed together: decode failures now `warn!`, and both filters read one shared signature table. (9d835188b)

relayer/relayer-migrate/migrations/20260824120000_create_gateway_chain_cursor.sql
manoranjith: Thanks for raising this — it should be fine, here's how it works.

A listener is a transport, not a slice of the chain: `ListenerInstanceConfig` is just `{type, url}`, and all listeners watch the same five signatures. Both relayers see the same events; relayer 2 just has more connections. That's why one shared row works.

If the two drift apart:

- The slower one can't pull the cursor back — the upsert only writes a higher block (`chain_cursor_repo.rs:56-63`).
- The faster one can't skip the other's work — a block is passed only once every event in it was handled into this same DB, and handling is idempotent.

One config worth noting: a relayer with only WebSocket listeners never touches the cursor, since only polling can attest to a full range. There's a startup warning for it (`settings.rs:848`), and no environment runs that way today.

relayer/src/gateway/handled_events.rs
manoranjith: Good question — it wasn't clear. Documented on the field in a79911397.

relayer/src/gateway/arbitrum/polling_listener.rs
manoranjith: Thanks for flagging this.

`last_block_number` isn't set in any of our environments, so the relayer starts from the saved cursor. Setting it is a manual step to force a replay, and it then applies on every restart.

Separately, after a restart with a lot of blocks to catch up, the first query asked for all of them at once. That now happens in batches of a configurable size (41e1c7116).

manoranjith:
Bringing the second point from our discussion onto the PR:

> `CursorState.pending` has no cap. A permanently hung handler grows one `VecDeque` per poll indefinitely. `relayer_gateway_pending_ranges` exposes it, but nothing bounds it — that's an alert, not a guard.

Thanks for flagging this — it is unbounded.

Memory stays small: ~48 bytes per entry, one every 2s.

What suffers is the saved block number. Say the listener reads 100, 101, 102, and a handler for 101 never returns. 101 and 102 wait behind it, so the number stays at 100 while the listener carries on to 103, 104. After a restart we come back to 100 and redo the rest.

Capping the queue would mean dropping entries, and saving a number whose blocks were never handled — exactly what the queue prevents.

Redoing the work is safe, and two changes came out of this:

- 41e1c7116 — the restart goes through the backlog in batches rather than one large query.
- b11b23f7c — the cursor write held a lock the listener needs, so a slow write could pause it.

`relayer_gateway_pending_ranges` is the signal if a range does get stuck.

relayer/src/gateway/handled_events.rs
```diff
@@ -296,6 +300,7 @@ impl HandledEvents {
                 "Gateway event handled"
             ),
             Err(e) => error!(
+                alert = true,
```
nboisde, line 303: 👍

nboisde (approved):
Latests changes looks good to me ! Thanks @manoranjith for addressing this.

manoranjith:
@mergify queue

PR #3648, opened 2026-08-27
fix(relayer): handle SIGTERM instead of being killed

manoranjith:
Closes zama-ai/fhevm-internal#1945.

Second of three stacked PRs, on top of #3647 — that one first. The shutdown policy here relies on its guarantee that an unfinished event leaves the recorded chain position untouched.

One commit. The shutdown sequence is in `startup.rs`; the readiness-worker changes are the part that is easiest to miss on a first read.

## Highlights for review

**In-flight readiness checks are cancelled on shutdown; in-flight sends are not.**

Cancelling a check is safe. It only reads an external endpoint, and the semaphore permit it holds is released when its future is dropped, so nothing is left behind.

A send is different. The nonce is taken and returned by explicit calls rather than by a guard that releases on drop, so cancelling after the RPC has already gone out would leave the nonce held with nothing left running to release it. Sends are therefore neither cancelled nor awaited — they are abandoned, and the next boot re-reads nonce state from the chain.

**The connection pools are no longer closed before exit.** Closing them waits on the connections the abandoned tasks still hold.

## Test plan

Local counts were measured at the top of the three-PR stack, not at this PR's tip in isolation.

- [ ] CI: clippy `-D warnings` and `cargo fmt -p fhevm-relayer --check`, both clean locally.

### DIFF
--- a/relayer/src/bin/fhevm-relayer.rs
+++ b/relayer/src/bin/fhevm-relayer.rs
@@ -6,6 +6,8 @@ use anyhow::Context;
 use clap::Parser;
 use fhevm_relayer::config::settings::Settings;
 use fhevm_relayer::tracing::init_tracing;
+use tokio::signal::ctrl_c;
+use tokio::signal::unix::{signal, SignalKind};
 use tokio_util::sync::CancellationToken;
 use tracing::info;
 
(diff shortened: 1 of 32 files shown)


relayer/src/bin/fhevm-relayer.rs
```diff
@@ -34,14 +36,19 @@ async fn main() -> anyhow::Result<()> {
     // We need to keep the guard to force-flush on SIGINT
     let chrome_tracing_guard = init_tracing(&settings.log)?;
 
-    // Create cancellation token and handle Ctrl+C
+    // Create cancellation token and handle shutdown signals (SIGINT, SIGTERM).
```
nboisde, line 39: Great to handle this case, even for non HA relayer it could have been useful

relayer/src/startup.rs
```diff
@@ -245,14 +277,43 @@ pub async fn run_fhevm_relayer(
     }
 
     // === Runtime Phase ===
-    // Wait for shutdown signal and shutdown all tasks via orchestrator
-    orchestrator
-        .run_until_shutdown(shutdown_token)
-        .await
-        .context("Failed during shutdown")?;
+    // Wait for the shutdown signal (SIGTERM/SIGINT, see `fhevm-relayer.rs`).
```
nboisde, line 271: lgtm !

nboisde (dismissed):
Basically a task tracker with cancellation token propagation here, with 503 forward at the right time !
LGTM, thanks Mano !

nboisde (dismissed):

nboisde (approved):

manoranjith:
@mergify queue

eudelins-zama:
@mergifyio queue

PR #3649, opened 2026-08-27
feat(relayer): enable ha by running multiple pods with one active

manoranjith:
Closes zama-ai/fhevm-internal#1946.

Third of three stacked PRs, on top of #3648. Review #3647 and #3648 first; this one only makes sense once an unfinished event is durable and a pod stops cleanly.

Two commits: the dispatcher lock with its schema and sweep, then the two-relayer test. The test is separate because it references only what the first commit already provides.

## Why the first commit is not split further

It contains five pieces — the advisory lock, the epoch schema, the sweep, the write fence and the dispatch gate. Split into five commits, the first four have no consumer until the gate arrives:

- the lock has no reader
- the columns have no reader until the sweep and the fence exist
- the sweep cannot be ownership-led before there is an owner

In that sequence the gate also removes about 2200 lines the earlier commits add, including a time-based `claim_after` staleness window and its rationale. This repo squash-merges, so that churn would stay in `main`'s history rather than being a review-time cost only.

## Highlights for review

**Losing the lock exits the process rather than going passive.**

All replicas share one signer key, and nonce state is an in-memory map seeded once from the chain. A passive pod would keep a stale high-water mark while its successor sends from the same wallet, which `owner_epoch` does not cover. Exiting discards that state, and the next boot re-reads it. `heartbeat_failures_before_exit` provides the tolerance window instead.

**`attempts` is a lifetime budget across owners, not per owner.**

A successor can therefore fail a row it did not itself attempt. A row that has spent its budget is exhausted regardless of which pod spent it. Resetting on takeover would give a crash-looping pod a fresh budget on every restart, since a restart mints a new epoch, and `max_attempts` would then never bind.

**The claim compares `<` while every fenced write compares `<=`.**

Two different questions about one column:

- may a write of mine land here — yes at or above the row's epoch, hence `<=`
- is anybody already driving this row — nobody, strictly below, hence `<`

A row reaches its attempt bound as soon as its own claim increments the counter, so `<=` in the exhaustion check would fail a live request mid-send.

## Startup ordering, and what it costs at handover

The dispatcher-lock task is spawned after `initialize_gateway`, so a pod reports Ready at HTTP bind. A rolling update can therefore retire the lock holder while its replacement is still initializing, and the interval with no dispatcher covers that initialization tail.

Requests are not lost: intake persists the row as `queued`, and the new holder's first sweep tick claims it. Handover takes longer than it did before. This follows from the ordering rather than being a regression against it.

## Measured on `zws-dev`, two replicas

- [ ] TODO: attach the measurement report and the steps used to produce these numbers. The values below are recorded from a sandbox run; the run itself is not yet reproducible from anything in this PR.

| Observation | Result |
| --- | --- |
| Dispatchers at any moment | exactly one |
| Shutdown duration, three terminations | 5.003–5.009s, of which 3–7ms real drain |
| Reached by SIGKILL | no |
| Lock released explicitly | yes |
| Restarts, all pods, whole test | zero |
| Ready pods during a rolling update | never fewer than two |

The handover figures were taken before the ordering change above, so they no longer describe post-merge behaviour. What they did show: the lock changed hands in 0.45–0.67s of a roughly 6s gap, the remainder being the configured load-balancer propagation wait. That wait remains the main lever on handover latency, now alongside the initialization tail.

## Test plan

- [ ] CI: clippy `-D warnings` — clean on each of this PR's commits individually and at the tip — and `cargo fmt -p fhevm-relayer --check`.
- [ ] CI: `sqlx-check` — this PR adds two migrations.


relayer/src/config/settings.rs
```diff
@@ -425,6 +425,183 @@ impl Default for ShutdownConfig {
...
+    }
+}
+
+/// The dispatch sweep: while this pod holds the dispatcher lock, it periodically claims and
+/// re-dispatches requests nobody is driving - one a non-holder accepted and left for the
+/// dispatcher, or one orphaned by a former holder that died. It is also how a
+/// restart recovers its own previous incarnation's work. Every field has a serde default so
+/// existing config files keep deserializing unchanged.
+#[derive(Debug, Deserialize, Clone)]
+pub struct SweepConfig {
```
nboisde, line 474: What is max attempts you have in mind there ? Isn't it overlap with the timeout configuration ?
I would say I understand about the interval setting, but not especially about the max attempts one.

relayer/relayer-migrate/migrations/20260825120000_add_owner_epoch_and_attempts.sql
```diff
@@ -0,0 +1,63 @@
...
+-- same type and rationale; ChainCursorRepository::advance fences on it alongside its
+-- monotonic block-number check.
+--
+-- `attempts` counts sweep re-dispatches of a row: the claim increments it, and a row
+-- reaching max_attempts is failed out instead of claimed again. NOT NULL DEFAULT 0 keeps
+-- pre-migration writers correct. The cursor gets no `attempts`: it is never re-dispatched.
+--
+-- All columns are additive and unconstrained; pre-migration INSERTs/UPDATEs keep working.
+-- ADD COLUMN with a constant default is metadata-only on Postgres 11+ (deployed: 17).
+ALTER TABLE user_decrypt_req
```
nboisde: I would add proper indexes, especially for claim sweep queries which could perform full table scans.

relayer/src/gateway/arbitrum/transaction/helper.rs
```diff
@@ -65,6 +65,18 @@ pub enum TxClaimOutcome {
     ClaimLost,
 }
 
+/// Outcome of recording a receipt this pod's own send obtained. The send succeeded either
+/// way - the chain has a real transaction - so this never decides a retry; it says whether
+/// the DB write recording the receipt landed, which the epoch fence refuses once a successor
+/// has claimed the row. Callers must not treat `Refused` as `Recorded`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ReceiptRecordOutcome {
```
nboisde, line 75: I understand this as we discussed it, but as a note we should be revisited to handle refused state after second pod becomes active.

relayer/src/orchestrator/dispatcher_lock.rs
```diff
@@ -0,0 +1,793 @@
...
+                self.exit_if_past_failure_bound(failures);
+            }
+        }
+    }
+
+    /// Shared bounded-retry accounting for a failed query on the dedicated connection. The
+    /// connection is dead either way and a live PID cannot be reconnected into, so past the
+    /// threshold the only move is exit. Takes the count, not the guard, so callers keep
+    /// their disjoint field borrows.
+    fn exit_if_past_failure_bound(&self, consecutive_failures: u32) {
```
nboisde, line 714: What are the exact triggers for getting an hearbeat failure ? 
From my understanding we should be careful combining this with potential long running queries or unindexed queries, a query that overlaps the heartbeat_timeout ? 

Normally we should not have issues regarding this, but prefer to point it here.

relayer/src/gateway/arbitrum/listener.rs
```diff
@@ -26,6 +26,8 @@ enum RecycleReason {
     StreamEnded,
     /// Planned connection recycle timer triggered
     RecycleTimer,
+    /// This pod stopped being the confirmed dispatcher, so the subscription is dropped
```
nboisde, line 29: 👍

nboisde (commented):
First round of review, waiting for changes after analysis.

relayer/relayer-migrate/migrations/20260825120000_add_owner_epoch_and_attempts.sql
manoranjith: Fixed in 9f71f0b53: one partial index per request table, over the three live statuses, leading on `owner_epoch` 

0d510fda5 widened the key to `(owner_epoch, id)`, so the claim's `ORDER BY owner_epoch, id` is the index's own order and Postgres stops at the batch bound with no sort.

The second unindexed query on this path was the retry-budget fail-out, and it is gone with the budget itself (3ec2a6a58) — see the `max_attempts` thread.

relayer/src/config/settings.rs
manoranjith: The setting is gone from `SweepConfig`, along with the `attempts` column it read — 3ec2a6a58.

relayer/src/orchestrator/dispatcher_lock.rs
manoranjith: - A slow query cannot fail hearbeat, because heartbeat and lock queries use a dedicated `PgConnection` separate from cron and app pools.

- The heartbeat query itself cannot fail either. It's a `SELECT pg_backend_pid()` (`dispatcher_lock.rs:491`) — no user table, no locks, no planner work.

So, the failures could only be  connection-level: a closed socket, a terminated backend, or saturation of the whole instance — not query duration. An error or a timeout demotes the pod to `Unconfirmed`, which closes its gates while keeping the session, and `exit_after_consecutive_failures` consecutive failures exit the process. A backend pid that changed under an active holder exits immediately instead of counting.

On the name: `heartbeat_timeout` is now `query_timeout` (`b65828669`), since it bounds four operations on that connection rather than the heartbeat alone.

The query that could have run long is the claim after a failover, which took the whole inherited backlog in one statement; `5483a8d37` bounds it per tick. That is a separate improvement — the heartbeat was never exposed to that path.

relayer/src/gateway/arbitrum/transaction/helper.rs
manoranjith: Agreed, and better tracked as its own issue than grown into this PR.

For the record of what is there today: `Refused` means the receipt write lost the epoch fence, so a successor had already claimed the row and reset it from `tx_in_flight` to `processing`. The send itself succeeded, so the outcome never decides a retry — the successor's sweep drives the row from its own claim and records its own receipt. The predecessor logs it, leaves `gw_reference_id` unstored (the gateway-event listener keys off it, so this must not read as success) and increments `transaction_receipt_refused`. What needs revisiting is whether it should do anything beyond that.

One neighbouring case is fenced now: `0845795c5` stops a send abandoned mid-handover from marking the row `failure`, so its state stays put for the successor's sweep.

relayer/src/gateway/arbitrum/transaction/engine.rs
```diff
@@ -474,6 +497,14 @@ impl
                     self.nonce_manager
                         .confirm_nonce(self.sender_address(), nonce)
                         .await;
+                    if nonces_walked > 0 {
```
nboisde: Revise this, not needed.

relayer/src/gateway/arbitrum/transaction/engine.rs
```diff
@@ -490,13 +521,37 @@ impl
...
+                                    step = %TxEngineStep::TxRetrying,
+                                    nonce = tx.nonce,
+                                    error = %err_msg,
+                                    "Nonce already mined"
+                                );
+                                self.nonce_manager
+                                    .confirm_nonce(self.sender_address(), nonce)
+                                    .await;
+                            } else if response_error_string
+                                .contains("replacement transaction underpriced")
```
nboisde: Lets remove "replacement ", the match will carry more rpc implementation versions.

relayer/src/gateway/arbitrum/transaction/engine.rs
```diff
@@ -490,13 +521,37 @@ impl
...
+                                nonces_walked += 1;
+                                debug!(
                                     int_job_id = %job_id,
                                     step = %TxEngineStep::TxRetrying,
                                     nonce = tx.nonce,
                                     error = %err_msg,
-                                    "Nonce too low"
+                                    "Nonce held by a pending transaction"
                                 );
                                 self.nonce_manager
```
nboisde: Release the nonce instead of confirming, and we can use the arming else if branch below: 

Rationnale behind this: 
with transaction underpriced, we cannot assume that a transaction using the same nonce will have a receipt and not be reverted, so its basically similar to a nonce too high issue. 
In this case, we make the nonce available again: 
1. if the older transaction is included, we will get for this exact same nonce a nonce too low error, and this transaction will be retried. 
2. if the older transaction is reverted, the transaction that will now consume this nonce, will pass.

relayer/src/gateway/arbitrum/transaction/nonce_manager.rs
```diff
@@ -177,6 +186,59 @@ impl NonceManager for NonceManagerNonOptimistic {
...
+                next_nonce,
+                ..Default::default()
+            })),
+        );
+        manager
+    }
+
+    /// The contrast the send triage relies on: a confirmed (retired) nonce is never drawn
+    /// again, a released one is drawn next.
+    #[tokio::test]
```
nboisde, line 213: update this test accordingly

relayer/src/metrics/transaction.rs
```diff
@@ -116,6 +116,9 @@ impl TransactionStatus {
 pub enum TransactionErrorType {
     InvalidAddress,
     Nonce,
+    /// A nonce a peer still holds. Separate from `Nonce` because a burst of these is a
+    /// handover, not a fault, and alerting cannot tell them apart under one label.
```
nboisde: Not sure this error is relevant with fixes.

nboisde (commented):
First round review for tx manager changes and sweep.

relayer/relayer-migrate/migrations/20260904120000_add_active_status_indexes.sql
```diff
@@ -0,0 +1,62 @@
...
+-- (readiness queue) and moves to `processing` (TX queue), so its ETA reads a count from each.
+CREATE INDEX IF NOT EXISTS idx_user_decrypt_req_active
+    ON user_decrypt_req (req_status, id)
+    WHERE req_status IN ('queued'::req_status, 'processing'::req_status, 'tx_in_flight'::req_status);
+
+CREATE INDEX IF NOT EXISTS idx_public_decrypt_req_active
+    ON public_decrypt_req (req_status, id)
+    WHERE req_status IN ('queued'::req_status, 'processing'::req_status, 'tx_in_flight'::req_status);
+
+-- Autovacuum tuning ships with the indexes because they only pay off while the visibility map is
```
nboisde, line 33: Lets double check this, I don't understand the rationale here..

relayer/src/store/sql/repositories/input_proof_repo.rs
```diff
@@ -148,24 +159,43 @@ impl InputProofRepository {
...
+                ins.is_inserted AS "is_inserted!",
+                ins.req_status AS "req_status!: ReqStatus",
+                ins.accepted,
+                ins.res,
+                (
+                    SELECT COUNT(*)
+                    FROM (
+                        SELECT 1
+                        FROM input_proof_req q
+                        WHERE q.req_status = 'processing'::req_status
```
nboisde, line 188: Please check if there is no change with the original algo behavior !

relayer/src/http/retry_after/state.rs
```diff
@@ -156,102 +150,6 @@ impl RetryAfterState {
 
     // ========== ETA Computation ==========
 
-    /// Compute ETA for a queued request based on queue info.
```
nboisde, line 159: Only old helpers i'd say.

nboisde (commented):
Lets have a look on ETA older compat, but could be good.

relayer/src/gateway/arbitrum/transaction/engine.rs
manoranjith: Done in f4a93fb53. The underpriced arm now releases the nonce and is folded into the nonce too high branch below, so an undecided nonce goes back in the pool either way.

relayer/src/store/sql/repositories/input_proof_repo.rs
manoranjith: Yes, i checked. The SQL based algo follows the same one as in-memory.

However, the algo itself had some small inaccuracies, which was too smalll compared to the floor of 2s. So its fine  as it is.

relayer/src/gateway/arbitrum/transaction/engine.rs
manoranjith: Done in f4a93fb53. The match is now on transaction underpriced, which also covers the case the TODO below flagged as untriaged.

relayer/src/gateway/arbitrum/transaction/engine.rs
manoranjith: Removed in f4a93fb53. With the underpriced arm releasing, nothing walks forward any more, so the nonces_walked counter went with the warning.

relayer/src/metrics/transaction.rs
manoranjith: Removed in f4a93fb53. Both former call sites now report nonce_error. The variant was never added to the documented label set in transactions.md either.

relayer/src/gateway/arbitrum/transaction/nonce_manager.rs
manoranjith: Updated in f4a93fb53. The assertions still hold, so only the doc comment changed: it now contrasts a mined nonce with an undecided one.

relayer/relayer-migrate/migrations/20260904120000_add_active_status_indexes.sql
manoranjith: Dropped manual autovacuum settings in 1bed27a86.

It is not relevant for our case. If needed, we can set it later with a plain ALTER TABLE on the live database.

relayer/src/http/retry_after/state.rs
manoranjith: Yes, only the old helpers. compute_queued_eta and compute_for_get were never on the production path.

relayer/src/gateway/arbitrum/transaction/helper.rs
manoranjith: Duplicate is absorbed by design, so the row still completes once, through the successor.

The orphaned event is counted in relayer_gateway_unmatched_events_total with nothing marking it as expected. Will be followed up in a logging + metrics PR.

nboisde (approved):
LGTM

manoranjith:
@mergify queue

PR #3650, opened 2026-08-28
feat(relayer): validate ciphertext commit readiness from s3

manoranjith:
Implements RFC-023 Part 2 on the Relayer — [fhevm-internal#1622][issue].

**Supersedes [#3394][old]**, which should be closed rather than updated. That PR targeted `main` and carried the RFC-023 work on an older relayer base; this one is rebased onto the relayer v1 with significant changes and targets `mano/relayer/ha-dispatcher-lock`, so the diff here is this feature's own commits rather than fifteen. The consensus code was also substantially reworked since — the verdict types, the error surface and the crate layout all changed — so #3394's diff no longer describes the proposal.

Before submitting a decryption to the Gateway, the relayer now evaluates Coprocessor attestation consensus per handle and fails closed when consensus is absent. `gw_ciphertext_check.source` selects between the on-chain Gateway check and the off-chain attestation check; both fail closed, and neither is a default.

**How #3394's review threads were addressed**

- *Split the client into its own crate* — done, though the opposite way round: the separate `ciphertext-attestation-client` crate is folded into `shared/ciphertext-attestation` behind a `client` feature, removing the duplicate `Cargo.lock` that the thread named as the concrete cost. A consumer needing only the wire types takes the default build; the relayer opts in.
- *Gate 1 / Gate 2 comment renders as one paragraph* — rewritten as a real docstring with a summary line, a `# Gates` list and an `# Errors` section. The cross-serving rationale is kept and now spells out the attack it prevents.
- *Unify the error types* — partially. `ValidationError` is gone, folded into `AttestationError`. The rest stay: `ciphertext_checker.rs` decides retry-versus-abandon by matching the variant, so merging them would either lose that discrimination or reproduce it with unreachable arms.

**Design notes**

- The tracker keeps one slot per registered Coprocessor and derives every count from that board, rather than storing counts alongside the record they summarise. A verdict states what the round proves: whether the threshold is still reachable this round, and whether any round with the current signer set could reach it at all.
- `Round` has two renderings for two audiences. `Display` is redacted — counts and absent signers, no digest values — and `Debug` is the full board with digest prefixes. `Display` is the rendering that fires implicitly (`format!`, `%` in tracing, every `#[error(...)]` attribute), and one such attribute reaches a database column and an HTTP response body, so the redacted form is the one that fires by default and digests stay in operator logs. Signer addresses do appear in user-facing strings: they are public on-chain state, and they tell a caller whether the fault is theirs.
- `ThresholdStatus` has four variants rather than three plus `Option`. The waiting state and the "this round reached nothing" state lead to different caller behaviour, and `Option::None` can carry only one of them.
- `ThresholdStatus::Reached` carries inline fields rather than reusing `consensus::Consensus`, leaving `consensus::evaluate` as that type's only owner. The trade-off is that inline fields get harder to change once kms-worker adopts the crate.
- `EventProcessingError::AttestationsNotReady` carries the round as a `String` rather than a `Round`, because that enum derives `Serialize`, `Deserialize` and `Clone`, which `Round` does not implement. Rendering therefore happens at that boundary, which also keeps the digest-bearing type out of anything serialized.

**Behaviour changes a reviewer will not see in the changed lines**

- A readiness check that exhausts its retries now stores its round alongside the standard phrase, so the 503 explains itself. Label selection was converted from whole-string equality to prefix matching first, in its own commit, which was behaviour-preserving at the time and is what makes the appended detail safe.
- A single Coprocessor answering twice with different material used to occupy two groups at once and could produce a terminal disagreement with no second signer in existence. One slot per signer makes that unrepresentable.
- A transient RPC error early in a readiness poll used to make the exhausted-retries verdict a terminal 500 instead of a retryable 503.

**Compatibility**

`consensus::evaluate` is byte-for-byte unchanged. `kms-connector`, `sns-worker` and `ethereum_rpc_mock` need no edits and were verified to build against the reworked crate. The `client` feature is opt-in, so consumers that only need the wire types do not pull in `alloy`, `tokio`, or the rest.

**Test plan**

- [ ] CI: `ciphertext-attestation` default features — locally 42 passed.
- [ ] CI: `ciphertext-attestation --all-features` — locally 59 passed. The 42/59 split is the proof the `client` gate is real; a misplaced `cfg` shows up as a count, not a compile error.
- [ ] CI: relayer `cargo test --lib` — locally 245/245.
- [ ] CI: relayer integration — locally `public_decrypt_v2_test` 53, `user_decrypt_v2_test` 78, `delegated_user_decrypt_v2_test` 80, `recovery_test` 6.
- [ ] CI: clippy `-D warnings` on both crates, fmt clean.

**Follow-ups, deliberately not here**

- `READINESS_CHECK_TIMEOUT_MSG` says "on the gateway chain", which is wrong under `source: coprocessor_attestations`. One source-neutral message is preferred over two constants.
- No `Retry-After` on the readiness 503 though the machinery exists.
- `extract_revert_selector` scans for any 8-character hex window, so a message that falls through the prefix ladder while carrying a handle could collide with a known revert selector. Not reachable from this change — reader and writer share the same prefix constant — but worth hardening separately.

[issue]: https://github.com/zama-ai/fhevm-internal/issues/1622
[old]: https://github.com/zama-ai/fhevm/pull/3394


shared/ciphertext-attestation/src/lib.rs
```diff
@@ -27,6 +27,22 @@ use serde::{Deserialize, Serialize};
 
 pub mod consensus;
 pub mod sign;
+pub mod tracker;
+
+/// Networked half of off-chain ciphertext attestation consensus: on-chain registry mirroring, S3
+/// `HEAD` attestation fetch, and the fan-out that evaluates consensus over the results. Behind
+/// the `client` feature so a consumer that only needs the wire types above is not made to pull in
+/// `alloy`, `tokio`, and the rest of this module's networked dependencies. See the module's own
+/// doc comment for the consumer flow.
```
eudelins-zama, line 36: There's docstring here + on top of the client/mod.rs module.

We should keep only the client/mod.rs one IMO

shared/ciphertext-attestation/src/lib.rs
```diff
@@ -186,6 +207,15 @@ pub enum AttestationError {
     Serde(#[from] serde_json::Error),
     #[error("signer error: {0}")]
     Signer(#[from] alloy_signer::Error),
+    /// The signature is genuine, but for a different bucket's key — cross-serving. See Gate 2 of
+    /// [`ValidAttestation::validate`](crate::tracker::ValidAttestation::validate): without this
+    /// check a bucket could serve another bucket's (validly signed) attestation and have it
+    /// counted as its own.
```
eudelins-zama, line 213:
```suggestion
    /// The signature is genuine, but for a different bucket's key.
```

shared/ciphertext-attestation/src/consensus.rs
```diff
@@ -34,6 +35,22 @@ pub struct ConsensusError {
...
+    /// audience as [`crate::tracker::Round`]'s hand-written `Debug`.
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "ct:{:02x}..,sns:{:02x}..",
+            self.ciphertext_digest.0[0], self.sns_ciphertext_digest.0[0]
+        )
+    }
+}
+
```
eudelins-zama, line 53: I don't get this.
We lose some information here (that could be useful?), and don't see a clear need for this compact form.

shared/ciphertext-attestation/src/client/s3.rs
```diff
@@ -0,0 +1,162 @@
+//! Interactions with the Coprocessors' S3 buckets: attestation fetching via `HEAD` requests.
```
eudelins-zama, line 1: Why not simply copy-pasting the entire `BoundedClient` from the `kms-connector`? (even if it means the `get` part of the client would not be used by the relayer, I think it's fine)

AFAIU, the effort of bringing the `BoundedClient` here was to be used by both kms-connector and relayer, but in current state, kms-connector can't use it

shared/ciphertext-attestation/src/client/registry.rs
```diff
@@ -0,0 +1,430 @@
...
+//!
+//! A consensus consumer needs the on-chain signer↔bucket binding for every registered
+//! Coprocessor (which signer a bucket is allowed to speak for, and the S3 bucket URL to fan
+//! attestation HEAD requests at) plus the majority threshold. Querying these on every request
+//! would trigger N+1 RPC calls, so the [`CoprocessorRegistry`] holds a whole snapshot behind a
+//! short TTL and tolerates registration changes within one refresh window.
+//!
+//! Keeping the triple together (rather than a flat signer set plus a separate bucket map) is
+//! what lets a caller validate each reply locally: a bucket's reply only counts if it recovers to
+//! *that bucket's* registered signer, not merely to some registered signer.
```
eudelins-zama, line 11:
```suggestion
//! TTL'd snapshot of the on-chain Coprocessor registry.
//!
//! A consensus consumer needs the on-chain signer↔bucket binding for every registered
//! Coprocessor (which signer a bucket is allowed to speak for, and the S3 bucket URL to fan
//! attestation HEAD requests at) plus the majority threshold. Querying these on every request
//! would trigger N+1 RPC calls, so the [`CoprocessorRegistry`] holds a whole snapshot behind a
//! short TTL and tolerates registration changes within one refresh window.
```

shared/ciphertext-attestation/src/client/registry.rs
```diff
@@ -0,0 +1,430 @@
...
+    time::Duration,
+};
+use tokio_util::sync::CancellationToken;
+use tracing::{error, warn};
+
+/// A periodically-synced mirror of the on-chain Coprocessor registry.
+///
+/// `N` defaults to [`Ethereum`] but any [`Network`] the caller's provider is built for (e.g.
+/// `alloy::network::AnyNetwork`) works: it is threaded through so this type fits whichever
+/// network the consumer's provider stack already uses.
```
eudelins-zama, line 33:
```suggestion
/// A periodically-synced mirror of the on-chain Coprocessor registry.
```

shared/ciphertext-attestation/src/client/registry.rs
```diff
@@ -0,0 +1,430 @@
...
+                    entry.tx_sender
+                )));
+            }
+            if !seen_signers.insert(entry.signer) {
+                return Err(RegistryError::Critical(format!(
+                    "duplicate Coprocessor signer in registry: {}",
+                    entry.signer
+                )));
+            }
+        }
```
eudelins-zama, line 153: I would remove this, this seems overengineered.

shared/ciphertext-attestation/src/client/registry.rs
```diff
@@ -0,0 +1,430 @@
...
+
+        // Registered Coprocessors without a bucket URL were dropped above, but the threshold comes
+        // from chain and is validated there only against the number of *registered* Coprocessors.
+        // If more were dropped than the margin allows, no round can ever reach threshold: every
+        // request is rejected until the registration is completed. That fails closed, which is
+        // right, but the retriable verdict it produces is indistinguishable from "attestations not
+        // uploaded yet" — so say plainly what is wrong, on every load, rather than leaving an
+        // operator to infer it from a timeout. Not `Critical`: crash-looping the caller over
+        // persistent on-chain state would be worse than serving a loud, retriable failure.
```
eudelins-zama, line 162:
```suggestion
```

The code speaks for itself, no need for 8 lines of comment...

Btw, I think we should raise a `RegistryError::Critical` error in such case despite Claude's sloppy explanation, there's no point in keeping the loop running here 😅

shared/ciphertext-attestation/src/client/registry.rs
```diff
@@ -0,0 +1,430 @@
...
+/// Gateway RPC endpoints can carry embedded API keys, so the URL must never reach a log line or an
+/// error surfaced to a caller. Applied at every point an RPC error becomes a [`RegistryError`], so
+/// consumers get a redacted message without having to remember to redact it themselves.
+fn redact_rpc_url(err: impl std::fmt::Display) -> String {
+    let msg = err.to_string();
+    match msg.find(" for url (") {
+        Some(at) => msg[..at].to_owned(),
+        None => msg,
+    }
+}
```
eudelins-zama, line 189: I would remove this.

It is only used below:

```
    let copro = contract
        .getCoprocessor(copro_tx_sender_addr)
        .call()
        .await
        .map_err(|e| anyhow::anyhow!("{}", redact_rpc_url(e)))?;
```

I'm not aware of any `alloy` error that would display this url. So this is useless code AFAIK

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,923 @@
...
+//!   [`MissedThisRound`](ThresholdStatus::MissedThisRound) (retriable) from
+//!   [`Unreachable`](ThresholdStatus::Unreachable) (terminal, but only for the current round's
+//!   registered signer set — a later on-chain change to that set could still succeed).
+//!
+//! A Coprocessor that failed to answer (timeout, HTTP error, a missing or malformed header) is
+//! dead for *this* round but alive for the *next* one — "has not published yet" is not "voted
+//! against" — and that asymmetry between outstanding replies and non-attesting ones is the
+//! entire reason [`MissedThisRound`](ThresholdStatus::MissedThisRound) and
+//! [`Unreachable`](ThresholdStatus::Unreachable) are distinct. See
+//! [`ConsensusTracker::status`] for the exact arithmetic.
```
eudelins-zama, line 29: Pretty sure this can be summarized in few lines.

I can't read all of this 🫠

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,923 @@
...
+    ///   ([`CiphertextAttestation::verify`]).
+    /// - **Gate 2** — that embedded signer equals `registered_signer`, the signer this bucket is
+    ///   registered to on-chain. Without this gate, an attacker controlling one Coprocessor's
+    ///   bucket could copy a *different* Coprocessor's genuine attestation into their own
+    ///   bucket's S3 object: Gate 1 alone would pass it, since the signature really is valid,
+    ///   and it would be counted as though that bucket had attested to it itself — cross-serving.
+    ///
+    /// # Errors
+    /// Returns [`AttestationError`]: a signature-recovery variant if Gate 1 fails, or
+    /// [`AttestationError::SignerNotRegisteredForBucket`] if Gate 2 fails.
```
eudelins-zama, line 57:
```suggestion
```

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,923 @@
...
+        }
+        ThresholdStatus::MissedThisRound(round.clone())
+    }
+}
+
+impl std::fmt::Display for Round {
+    /// Redacted: counts plus who was absent, no digest values. This is the rendering that fires
+    /// implicitly — `format!("{round}")`, `%round` in tracing, `.to_string()`, and every
+    /// `#[error("...{round}")]` — so it is the only one allowed to reach a stored error string
+    /// or an API response. See [`std::fmt::Debug`] for the full board.
```
eudelins-zama, line 310:
```suggestion
```

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,923 @@
...
+            }
+        }
+        Ok(())
+    }
+}
+
+/// Short form for the operator-log board: `0x` plus the address's leading byte.
+fn short_addr(addr: &Address) -> String {
+    format!("0x{:02x}..", addr.0[0])
+}
```
eudelins-zama, line 360: Why this?

shared/ciphertext-attestation/src/client/fetch.rs
```diff
@@ -0,0 +1,325 @@
...
+    .await
+}
+
+/// Drains `tasks`, feeding each reply to `tracker`, and returns the round's verdict.
+///
+/// Split out from [`fetch_attestations_and_check_consensus`] so the post-drain sweep (see below)
+/// can be unit-tested against a `JoinSet` built directly, including one task that genuinely
+/// panics — something no HTTP mock can trigger, since `reqwest`/`hyper` guarantee a `Result` for
+/// every malformed or hostile response, never a panic.
+async fn drain(
```
eudelins-zama, line 97: `drain` is super vague name for what seems to be a crucial piece of the attestation consensus check work...

shared/ciphertext-attestation/src/lib.rs
manoranjith: Addressed in 9f3d78e81 — the `lib.rs` copy is gone, `client/mod.rs` keeps it.

shared/ciphertext-attestation/src/lib.rs
manoranjith: Addressed in 9f3d78e81.

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: Addressed in 9f3d78e81.

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: Addressed in 9f3d78e81.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Addressed in 9f3d78e81.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Addressed in 9f3d78e81.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Addressed in 9f3d78e81.

shared/ciphertext-attestation/src/client/fetch.rs
manoranjith: Addressed in ea122b889 — `drain` is now `resolve_round`.

shared/ciphertext-attestation/src/consensus.rs
manoranjith: Addressed in 76c5fd74a . Having shortened digest could loose info. So replaced this with `Debug` macro derived impl.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Addressed in 76c5fd74a by removing it.

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: Addressed in f5ad9ae14 — agreed. `GatewayConfig.sol` already reverts both duplicates against persistent storage.  Should be safe to drop the check for duplicates.

shared/ciphertext-attestation/src/client/s3.rs
manoranjith: Done. BoundedClient is copied from kms-connector verbatim in 4942a5fac (that commit does not compile), then stripped of the kms-connector-specific logic in 952c1cd9d. a74072549 replaces the two hard-coded fan-out constants with one config key.

```
diff <(git show 4942a5fac:kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/s3.rs) \
     <(git show 4942a5fac:shared/ciphertext-attestation/src/client/s3.rs)
```

#3678 describes how the shared BoundedClient can be used in kms-connector.

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: The URL comes from `reqwest` rather than from alloy. On a connect or timeout failure:

- The HTTP transport wraps the `reqwest::Error` in [`TransportErrorKind::Custom`](https://docs.rs/alloy-transport-http/2.4.1/src/alloy_transport_http/reqwest_transport.rs.html#47).
- Each layer above displays its inner error unchanged: [`Custom`](https://docs.rs/alloy-transport/2.4.1/src/alloy_transport/error.rs.html#54-55) is `#[error("{0}")]`, [`RpcError::Transport`](https://docs.rs/alloy-json-rpc/2.4.1/src/alloy_json_rpc/error.rs.html#48-49) and [`alloy_contract::Error::TransportError`](https://docs.rs/alloy-contract/2.4.1/src/alloy_contract/error.rs.html#35-36) are `#[error(transparent)]`.
- [`reqwest::Error`'s `Display`](https://docs.rs/reqwest/0.13.4/src/reqwest/error.rs.html#280) ends with ` for url ({url})`.

So the `{}` prints the endpoint. Harmless with eRPC, but not if we point this at a direct RPC provider URL that carries a secret.

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: Thanks for spotting this — the comment was doing too much work, so I dropped it and reworked the code around it. Everything below is in 4ea7a4c38.

Two conditions now fail the request immediately, without entering the retry budget, both answering **503 `gateway_not_reachable`** (retryable):
- **Too few registered bucket URLs to ever reach the threshold** — `snapshot.consensus_reachable()` is false, so we return right there. Previously we probed and retried until the retries ran out, reporting `readiness_check_timed_out`, which looks identical to attestations merely being late.
- **The last refresh failed critically** — `last_refresh_failed_critically()` exposes it, so a stale snapshot is refused rather than quietly served from.

On your suggestion:
- I did not raise `RegistryError::Critical` for the sub-threshold case. `load` returns either an error or a snapshot, never both, and here the read succeeded: the snapshot is accurate, it just says consensus is out of reach. Returning `Err` would throw it away and leave the older snapshot in place.
- The loop was a real find, though — it should keep running, but the cancel should not have been there. `spawn_refresh_task` was cancelling the `CancellationToken` its *consumer* handed in, stopping the dequeue workers while the HTTP server kept queueing into a channel with no reader. Removed in the same commit.

Let me know your comments.

.github/workflows/shared-crates-tests.yml
```diff
@@ -77,7 +80,10 @@ jobs:
 
       - name: Run Tests
         run: |
-          RUST_BACKTRACE=full cargo test
+          # --all-features: ciphertext-attestation's `client` feature (and any future feature on
+          # a crate sharing this template) is off by default, so a bare `cargo test` would run
+          # clean while silently skipping its tests forever.
+          RUST_BACKTRACE=full cargo test --all-features
```
eudelins-zama:
```suggestion
          RUST_BACKTRACE=full cargo test --all-features
```

shared/ciphertext-attestation/src/consensus.rs
```diff
@@ -17,6 +17,9 @@ pub struct Consensus {
 }
 
 /// The ciphertext material a consensus group agreed on.
+///
+/// Digest values are operator-only. There is deliberately no `Display` impl; if one is ever
+/// needed, it must not carry digest values.
```
eudelins-zama:
```suggestion
```

shared/ciphertext-attestation/src/consensus.rs
```diff
@@ -47,14 +50,10 @@ impl From<&CiphertextAttestation> for ConsensusMaterial {
...
 /// An attestation counts only if its signature recovers to its embedded `signer` and that signer
-/// is in the `allowed_signers` set.
-/// Survivors are grouped by material tuple; the largest group wins if it gathers at least
-/// `threshold` distinct signers.
-///
-/// `threshold` is a [`NonZeroUsize`]: a zero threshold would let any single attestation win,
-/// so it is unrepresentable here and rejected where the threshold is loaded.
+/// is in the `allowed_signers` set. Survivors are grouped by material tuple; the largest group
+/// wins if it gathers at least `threshold` distinct signers. `threshold` is a [`NonZeroUsize`]:
+/// a zero threshold would let any single attestation win.
```
eudelins-zama:
```suggestion
/// is in the `allowed_signers` set. Survivors are grouped by material tuple; the largest group
/// wins if it gathers at least `threshold` distinct signers.
```

shared/ciphertext-attestation/src/client/s3.rs
```diff
@@ -0,0 +1,490 @@
...
+    sync::{Arc, Mutex},
+    time::Duration,
+};
+use tokio::sync::Semaphore;
+
+/// An HTTP client with a ceiling on the `HEAD` probes it may have in flight against any one
+/// Coprocessor bucket at once.
+///
+/// Every bound is required at construction and this crate offers no default for any of them: each
+/// consumer fans out differently, so a value that is right for one is wrong for the other.
```
eudelins-zama:
```suggestion
/// An HTTP client with a ceiling on the requests it may have in flight.
```

shared/ciphertext-attestation/src/client/fetch.rs
```diff
@@ -0,0 +1,249 @@
...
+/// A reached consensus together with the buckets to fetch the ciphertext from.
+///
+/// `winning_buckets` are the buckets whose registered signer is in the winning group; a consumer
+/// that only needs the consensus verdict can ignore it.
+#[derive(Debug)]
+pub struct ResolvedConsensus {
+    pub material: ConsensusMaterial,
+    pub signers: Vec<Address>,
+    pub winning_buckets: Vec<String>,
+}
```
eudelins-zama, line 38:
```suggestion
/// A reached consensus together with the buckets to fetch the ciphertext from.
#[derive(Debug)]
pub struct ResolvedConsensus {
    pub material: ConsensusMaterial,
    pub signers: Vec<Address>,
    /// The buckets whose registered signer is in the winning group.
    pub winning_buckets: Vec<String>,
}
```

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+        });
+    }
+    Ok(ConsensusMaterial::from(attestation))
+}
+
+/// One registered Coprocessor's on-chain identity triple: which tx sender it is, which signer it
+/// is bound to, and which bucket serves its attestations. (`getCoprocessor(txSender)`)
+///
+/// Lives here rather than in `client::registry` because [`Round`] holds one per slot, and
+/// `tracker` is not behind the `client` feature. `registry` re-exports it.
```
eudelins-zama:
```suggestion
/// The on-chain identity of a registered Coprocessor.
```

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+    NoReply,
+    /// It answered, but the attestation failed validation.
+    Rejected,
+    /// It has not answered yet.
+    Outstanding,
+}
+
+/// One round of asking every registered Coprocessor for an attestation.
+///
+/// `Display` is redacted (no digest values); `Debug` is the full board, operator diagnostics only.
```
eudelins-zama:
```suggestion
/// One round of asking every registered Coprocessor for an attestation.
```

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+    pub threshold: NonZeroUsize,
+    /// One slot per registered Coprocessor, in roster order. Filled in place as replies arrive.
+    pub replies: Vec<(CoprocessorEntry, Reply)>,
+}
+
+impl Round {
+    /// The roster size, not the number of replies received.
+    pub fn asked(&self) -> usize {
+        self.replies.len()
+    }
```
eudelins-zama: Do we need this?

Can't we simply use `self.replies.len()` where we need it?

If we need it, the docstring + function name are not easy to understand...

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+        groups.sort_by(
+            |(left_material, left_entries), (right_material, right_entries)| {
+                right_entries
+                    .len()
+                    .cmp(&left_entries.len())
+                    .then_with(|| left_material.cmp(right_material))
+            },
+        );
+        groups
+    }
```
eudelins-zama: Maybe we should not sort groups by size here, and let `agreeing` look for the max?

It's a bit unintuitive to me to rely on `groups` being sorted so `groups.first()` works in `agreeing`

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+}
+
+/// Where a round stands, recomputed after every reply.
+#[derive(Clone, Debug)]
+pub enum ThresholdStatus {
+    /// Not everyone has answered and the threshold is still reachable this round.
+    AwaitingReplies,
+    /// A group of at least `threshold` distinct signers agreed on the same material. `winners`
+    /// are their roster entries, so the caller reads both the signers and their buckets off it.
```
eudelins-zama:
```suggestion
    /// A group of at least `threshold` distinct signers agreed on the same material.
```

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+    Reached {
+        material: ConsensusMaterial,
+        winners: Vec<CoprocessorEntry>,
+    },
+    /// Every Coprocessor answered or failed; no group reached the threshold. Retriable.
+    MissedThisRound(Round),
```
eudelins-zama:
```suggestion
    /// Every Coprocessor answered or failed; no group reached the threshold. Retriable.
    NotReached(Round),
```

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+    /// The Coprocessors that answered disagree, and even the best possible outcome would still
+    /// fall short of threshold. Terminal for this round's registered signer set: cast
+    /// attestations are immutable.
```
eudelins-zama: Attestations are not immutable currently on copro side AFAIK. So I'm not sure we can really distinguish `MissedThisRound` from `Unreachable` unless I'm missing something?

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+
+/// Incremental consensus over one fan-out round. No network, no time, no I/O: the caller feeds
+/// it exactly one reply per registered Coprocessor and reads a freshly recomputed verdict back
+/// after each one.
+pub struct ConsensusTracker {
+    round: Round,
+}
+
+impl ConsensusTracker {
+    /// Builds the board from the roster: one slot per registered signer, all
```
eudelins-zama: I don't like `roster` being used everywhere like it was something the reader is supposed to know what it means.
I don't get why weird term like `board` are used instead of `tracker` as well.

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+            .replies
+            .iter_mut()
+            .find_map(|(entry, slot)| (entry.signer == signer).then_some(slot))
+        {
+            Some(slot @ Reply::Outstanding) => *slot = reply,
+            Some(_) => {}
+            None => debug_assert!(
+                false,
+                "record() called for signer {signer}, not in the roster"
+            ),
```
eudelins-zama: Why `debug_assert!` instead of `warn!`?

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+                    .into_iter()
+                    .map(|entry| (entry, Reply::Outstanding))
+                    .collect(),
+            },
+        }
+    }
+
+    /// Fills `signer`'s slot and returns the freshly recomputed verdict. First write wins: a
+    /// later reply for the same signer is dropped, which makes a single Coprocessor structurally
+    /// unable to occupy two groups at once. An unknown signer is ignored (`debug_assert!`).
```
eudelins-zama: > First write wins: a later reply for the same signer is dropped, which makes a single Coprocessor structurally unable to occupy two groups at once.

I think this assume the copro attestation are immutable, which is not the case AFAIK

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+        }
+        self.verdict()
+    }
+
+    /// Reads the board without changing it.
+    pub fn verdict(&self) -> ThresholdStatus {
+        let round = &self.round;
+        let threshold = round.threshold.get();
+        let groups = round.groups();
+        let largest = groups.first().map_or(0, |(_, entries)| entries.len());
```
eudelins-zama: We should use `agreeing` here, right?

shared/ciphertext-attestation/src/tracker.rs
eudelins-zama: EDIT: by reading at the `verdict` function, I understand the nuance between the two, ignore this :)

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+        .await
+        .unwrap();
+        let material = validate(&att, HANDLE, COPROCESSOR_CONTEXT_ID, signer.address()).unwrap();
+        (signer.address(), Reply::Attested(material))
+    }
+
+    /// A filler roster slot that never replies this round.
+    fn filler() -> Address {
+        PrivateKeySigner::random().address()
+    }
```
eudelins-zama: This makes test harder to understand.

I think we should just use `PrivateKeySigner::random().address()` inline, way more explicit

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+        let round = disagreeing_round().await;
+        let rendered = round.to_string();
+
+        for leaked in [CT_DIGEST, SNS_DIGEST, DISSENTING_SNS_DIGEST] {
+            assert!(
+                !rendered.contains(&format!("{leaked:x}")),
+                "leaked digest {leaked}: {rendered}"
+            );
+        }
+    }
```
eudelins-zama: Do we really need a test for this?...

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+        let round = disagreeing_round().await;
+        let rendered = format!("{round:?}");
+
+        for expected in [CT_DIGEST, SNS_DIGEST, DISSENTING_SNS_DIGEST] {
+            assert!(
+                rendered.contains(&format!("{expected:x}")),
+                "missing digest {expected}: {rendered}"
+            );
+        }
+    }
```
eudelins-zama: Same

shared/ciphertext-attestation/src/tracker.rs
```diff
@@ -0,0 +1,806 @@
...
+        } else {
+            (&material2, &material1)
+        };
+        assert_eq!(
+            &groups[0].0, smaller,
+            "groups()[0] must be the smaller material on a tie, matching consensus::evaluate"
+        );
+        assert_eq!(&groups[1].0, larger);
+    }
+}
```
eudelins-zama, line 784: Do we need this?

.gitattributes
```diff
@@ -0,0 +1 @@
+Cargo.lock linguist-generated=true
```
eudelins-zama: What is this?

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Dropped in 78d1854ac.

Same for three more `Round` accessors with no caller outside the crate — `silent` and `rejected` now private, `outstanding` `pub(crate)`. `attested` and `agreeing` stay `pub` for the integration tests.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Dropped in 78d1854ac.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Dropped in 78d1854ac.

Digest values are not confidential, so the two renderings differ in shape alone and the test pinned formatting rather than behaviour.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Dropped in 78d1854ac.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Updated in ab76cd73b. `groups()` is now a private `winner()` using `max_by`, so nothing depends on sort order.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Updated in ab76cd73b. `verdict` and `agreeing` both read one `winner()` call now.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Updated in addf4b8bf. `roster` is now "registered Coprocessors", `board` is "the replies".

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Suggestion accepted in d913dc996, with `NotReachedThisRound` as the name: "this round" makes explicit that the threshold can still be reached later, unlike the terminal `Unreachable`. Same rename in `ConsensusCheckError`.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Fixed in a7a817e8e. `Unreachable` is terminal for the round, not because attestations are immutable.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: `record` writes only into an `Outstanding` slot, so a later reply is dropped whether or not the attestation changed. `second_different_vote_from_same_signer_does_not_create_second_group` covers that case. Wording clarified in a7a817e8e.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Updated in 3d15b7600. `warn!` instead of `debug_assert!`, which was silent in release.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Suggestion accepted in e038a7a85. Added a line on why the type is not in `client::registry`: `tracker` is not behind the `client` feature.

shared/ciphertext-attestation/src/client/s3.rs
manoranjith: Suggestion accepted in e038a7a85. Added a line on why every bound is required with no default.

shared/ciphertext-attestation/src/consensus.rs
manoranjith: Suggestion accepted in e038a7a85.

.gitattributes
manoranjith: Marks every `Cargo.lock` as generated for GitHub Linguist. Added in 63cad6cc2, when two lockfiles here were burying the diff. Pattern-based, so it covers all ten `Cargo.lock` files rather than the two in this PR.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Yes, this test ensures the tie-break stays deterministic.

It mirrors the test of the same name in `consensus.rs` that pins it for `evaluate`

Renamed and comment trimmed in ab76cd73b.

.github/workflows/shared-crates-tests.yml
manoranjith: Accepted suggestion in 267264496.

shared/ciphertext-attestation/src/consensus.rs
manoranjith: Accepted suggestion in 267264496.

shared/ciphertext-attestation/src/client/fetch.rs
manoranjith: Accepted suggestion in 267264496.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Accepted suggestion in 267264496.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Accepted suggestion in 267264496.

shared/ciphertext-attestation/src/client/registry.rs
eudelins-zama: Haven't dig in the code as registry is now out of my review scope. But I find the reply very suspicious. I maintain what I said above, I think it should have raised a `RegistryError::Critical` if we're not able to reach consensus via the config that lives onchain.
There's no sense in trying to run anything in such case

shared/ciphertext-attestation/src/client/registry.rs
eudelins-zama: I think this is out of our scope, and must rely on `alloy` here (we could check there codebase, or submit issues to their repo if needed).

We would bloat our entire codebase if this was important.

shared/ciphertext-attestation/src/tracker.rs
eudelins-zama: > record writes only into an Outstanding slot, so a later reply is dropped whether or not the attestation changed

That is the problem I was raising. If a corpo upload a wrong attestation, and wants to override it (in case of reorgs, or whatever), this could be a legit case.
So I'm not sure we should drop the later reply.

.gitattributes
eudelins-zama: Seems like already supported by default? https://github.com/github-linguist/linguist/issues/3988

I don't think we need this tbh

shared/ciphertext-attestation/src/client/registry.rs
eudelins-zama: Or at least we should explore cleaner solution IMO

.gitattributes
manoranjith: Thanks for pointing this out — removed in acd6a4bca.

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Within a round — one fan-out, one HEAD per registered bucket — a Coprocessor replies once, so there is no later reply to drop. If its attestation changes, the next round re-probes every bucket and picks up the new one.

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: The URL is appended by `reqwest`, not by alloy: `reqwest::Error::Display` ends with ` for url (…)`, and alloy forwards it unchanged under `#[error(transparent)]`.

The upstream discussion on that message asks for more context in it, not less ([reqwest#2342](https://github.com/seanmonstar/reqwest/discussions/2342)), so the URL is unlikely to go away.

Relying on alloy would therefore keep the endpoint in the message, API key included.

A cleaner solution should be possible. I will do it in a follow-up PR, since the relayer uses the same pattern and both could be updated in one go.

shared/ciphertext-attestation/src/tracker.rs
eudelins-zama: Ok I understand thanks!

But then saying "First write wins: later reply for the same signer is dropped" is quite misleading, as in theory, we would query attestation only once per bucket per round, so talking about a later reply is a bit confusing to me.

shared/ciphertext-attestation/src/client/registry.rs
eudelins-zama: FYI, there is this [without_url](https://docs.rs/reqwest/latest/reqwest/struct.Error.html#method.without_url) function, that we may try to leverage for this.

Don't know how hard it would be to implem a tiny tower middleware or whatever that we would set up on `alloy` Provider to use this for every transport error that requires it.
Would avoid manual handling like this everywhere

shared/ciphertext-attestation/src/tracker.rs
manoranjith: Updated the docs in ee7263dca to make it clear.

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: Yeah, a tower middleware is what i had in mind when i posted the previous message. 

But will do it in a separate PR, since other part in relayer also uses this pattern and i would fix both in one go.

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: Following our discussion, registry validation failures are now `Critical` at the first attempt — too few Coprocessors for the majority threshold, or none with a usable bucket. `ConsensusUnreachable` reported the same thing and is folded into it. Only a read that never arrived stays `Transient`.

A `Critical` is detected on the first check after the registry reaches that state:

- In the relayer, the service keeps running and the caller gets a 503 with the registered and required counts.
- In kms-connector, it could crash on a `Critical`. The refresh loop needs no link to the consumer.

The shutdown token stays what the application passes in to stop the refresh loop.

Changed in b5a3cb6ea.

eudelins-zama: I find it quite weird to store the critical error within the registry struct, but anyway, I keep that part of the code outside of my code review, as not used by kms-connector for now

eudelins-zama (dismissed):
LGTM (review focused on `shared/ciphertext-attestation`, but not `shared/ciphertext-attestation/src/client`)

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: The rationale here, in relayer, it will continue to serve HTTP and by looking at this error, return an error to the user without trying to do the fan out.

In kms-connector side (if this is adopted), it could crash itself on first encounter of critical error.

eudelins-zama: I get the rationale, I'm just surprised that this would be the cleaner approach

relayer/config/local.mainnet.yaml.example
```diff
@@ -122,9 +122,27 @@ gateway:
         max_attempts: 3
         retry_interval_ms: 1000
     gw_ciphertext_check:
+      # Off-chain Coprocessor attestation consensus (RFC 023). The alternative is the on-chain
+      # Gateway check, via source: gateway_chain, which takes only `retry`:
```
nboisde, line 126: Perfect for retro compatible relayers 👍

relayer/src/core/errors.rs
```diff
@@ -14,6 +14,18 @@ use thiserror::Error;
 pub const READINESS_CHECK_TIMEOUT_MSG: &str =
     "Ciphertext not ready for decryption on the gateway chain";
 
+/// Terminal counterpart to [`READINESS_CHECK_TIMEOUT_MSG`]: the Coprocessors were reachable and
```
nboisde, line 17: 👍

relayer/src/core/errors.rs
```diff
@@ -14,6 +14,18 @@ use thiserror::Error;
...
+/// did not agree, so waiting cannot make the request ready. A prefix rather than a whole message,
+/// following the host ACL errors, so the per-handle detail survives into the stored reason.
+pub const NO_ATTESTATION_CONSENSUS_PREFIX: &str =
+    "Coprocessors did not reach consensus on the ciphertext material:";
+
+/// The registry could not support consensus, so nothing was probed. Retryable — the fault is on
```
nboisde, line 23: Do you mean here for consensus = 1, or working for consensus equal 1, and something else related here ?

relayer/src/gateway/ciphertext_checker.rs
```diff
@@ -1,84 +1,163 @@
...
         let decryption_address = Address::from_str(&gateway_config.contracts.decryption_address)
             .map_err(|_| {
                 EventProcessingError::ConfigError(AppConfigError::InvalidAddress(
                     "contracts.decryption_address".to_owned(),
                 ))
             })?;
 
-        let url = Url::parse(&gateway_config.blockchain_rpc.read_http_url).map_err(|e| {
-            EventProcessingError::ValidationFailed {
+        let provider = build_gateway_provider(&gateway_config.blockchain_rpc.read_http_url)
```
nboisde, line 146: Replacing Provider struct using builder from host looks good to me !

relayer/src/gateway/ciphertext_checker.rs
```diff
@@ -232,3 +312,343 @@ impl CiphertextChecker {
...
+    request_timeout: Duration,
+    max_concurrent_handles: NonZeroUsize,
+}
+
+impl CoprocessorAttestationCheck {
+    /// Mirrors the Coprocessor registry from `GatewayConfig` and starts its refresh task.
+    ///
+    /// `cancel_token` stops the refresh task and nothing else — a critically failed refresh is
+    /// answered per request, see [`Self::check_handles_with_retry`].
+    #[allow(clippy::too_many_arguments)]
```
nboisde, line 331: Nit: optimise this later.

relayer/src/gateway/ciphertext_checker.rs
```diff
@@ -232,3 +312,343 @@ impl CiphertextChecker {
...
+                    "Ciphertext attestation check exceeded its overall request budget"
+                );
+                Err(ReadinessCheckError::GwTimeout)
+            }
+        }
+    }
+
+    /// Re-probes the handle set until it succeeds, becomes unreachable, or runs out of attempts.
+    /// Bounded in time only by its caller — see [`Self::check_handles_with_retry`].
+    async fn retry_while_missing(
```
nboisde, line 492: 👍

relayer/src/gateway/ciphertext_checker.rs
```diff
@@ -232,3 +312,343 @@ impl CiphertextChecker {
...
+        let snapshot = self.registry.snapshot();
+
+        // Handles are copied out of the slice rather than borrowed: a borrow taken from the
+        // iterator would tie the closure to one anonymous lifetime, and the resulting future is
+        // not general enough for the `run_consumer` callers to hold across an await.
+        let mut verdicts = stream::iter(handles.iter().copied().enumerate())
+            .map(|(index, handle)| {
+                let snapshot = snapshot.clone();
+                async move {
+                    let outcome = fetch_attestations_and_check_consensus(
```
nboisde, line 568: Looks good to me.

relayer/src/gateway/ciphertext_checker.rs
```diff
@@ -232,3 +312,343 @@ impl CiphertextChecker {
...
+                Err(e @ ConsensusCheckError::NotReachedThisRound { .. }) => {
+                    if first_missed.as_ref().is_none_or(|(at, _)| index < *at) {
+                        first_missed = Some((index, e));
+                    }
+                }
+            }
+        }
+
+        match first_missed {
+            Some((_, e)) => Err(e),
```
nboisde, line 594: As a nit for later: would log error or warn the first missed here, even if its propagated above.

nboisde (dismissed):
Small nits on some points that could be addressed further, nothing blocking there. Looks great overall on the relayer side !
Lets check on relayer sdk about new labels.
Thanks for this integration 🙏

shared/ciphertext-attestation/src/client/registry.rs
manoranjith: I think this is good enough to cover both relayer (detect and keep forwarding the error to user) and connector (crash). So, would leave as-is for the scope of this PR.

If we find a cleaner approach later, we could sure revise. Because it would be a opaque and not breaking change to the end user (HTTP client / Operators of relayer, kms)

relayer/src/gateway/ciphertext_checker.rs
manoranjith: I agree :+1:  Not much i favor too many arguments. We could optimise it later though.

relayer/src/core/errors.rs
manoranjith: A threshold of 1 works normally. This error covers the registry falling short of the threshold, for example 4 required with only 3 Coprocessors carrying a registered bucket URL.

nboisde (approved):
Approving for rebase.

eudelins-zama (approved):

manoranjith:
@mergify queue

PR #3585, opened 2026-08-24
docs(sdk): rewrite release notes for v0.13.3, v0.14.1-0, and v0.15.0

geoxel:
Replace the abandoned v1.2.1/v1.2.0-alpha.1 entries with notes for the actual protocol-aligned release lineage that followed v1.1.0-alpha.7.

PR #3657, opened 2026-08-28
fix(sdk): forward delegatorAddress through V2 permit parse/serialize

eitjuh:
parseSignedDecryptionPermitV2 never read delegatorAddress off the input permit, silently dropping delegation for a manually-reconstructed V2 permit (the offline-signing shape). serializeSignedDecryptionPermitToJSON has the same gap on the other side: a delegated permit signed via signUnifiedDecryptionPermit loses its delegation on the sign -> serialize -> parse round trip.

V1 doesn't hit this because delegatorAddress lives inside eip712.message; V2 keeps it as post-sign metadata, so it has to be carried explicitly at every boundary. Adds two tests mirroring the existing manual-sign and serialize/parse patterns, covering the delegated case for both.

### DIFF
--- a/sdk/js-sdk/src/core/actions/chain/serializeSignedDecryptionPermit.ts
+++ b/sdk/js-sdk/src/core/actions/chain/serializeSignedDecryptionPermit.ts
@@ -17,7 +17,6 @@ export type SerializeSignedDecryptionPermitReturnType = {
   readonly eip712: Eip712Like;
   readonly signature: string;
   readonly signerAddress: string;
-  /** V2 (unified) permits only — delegation is metadata alongside the signed message, not part of it. */
   readonly delegatorAddress?: string | undefined;
 };
 
--- a/sdk/js-sdk/src/core/actions/chain/serializeSignedDecryptionPermit.ts
+++ b/sdk/js-sdk/src/core/actions/chain/serializeSignedDecryptionPermit.ts
@@ -17,6 +17,8 @@ export type SerializeSignedDecryptionPermitReturnType = {
   readonly eip712: Eip712Like;
   readonly signature: string;
   readonly signerAddress: string;
+  /** V2 (unified) permits only — delegation is metadata alongside the signed message, not part of it. */
+  readonly delegatorAddress?: string | undefined;
 };
 
 /**
--- a/sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
+++ b/sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
@@ -94,6 +94,22 @@ export function isSignedDecryptionPermitV2(value: unknown): value is SignedDecry
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+// _assertSignerIsNotDelegator
+////////////////////////////////////////////////////////////////////////////////
+
+function _assertSignerIsNotDelegator(
+  signerAddress: ChecksummedAddress,
+  delegatorAddress: ChecksummedAddress | undefined,
+): void {
+  if (delegatorAddress !== undefined && signerAddress.toLowerCase() === delegatorAddress.toLowerCase()) {
+    throw new Error(
+      'signerAddress and delegatorAddress must be different. ' +
+        'Use a non-delegated permit to decrypt your own values.',
+    );
+  }
+}
+
+////////////////////////////////////////////////////////////////////////////////
 // _createSignedDecryptionPermitV2
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -107,6 +123,8 @@ async function _createSignedDecryptionPermitV2(
   },
 ): Promise<SignedDecryptionPermitV2> {
   const { eip712, signature } = parameters;
+
+  _assertSignerIsNotDelegator(parameters.signerAddress, parameters.delegatorAddress);
 
   // Enforced here (the choke point shared by signDecryptionPermitV2 and
   // parseSignedDecryptionPermitV2). The message field is a validated uint256
@@ -150,6 +168,7 @@ export async function signDecryptionPermitV2(
   if (delegatorAddressArg !== undefined) {
     assertIsAddress(delegatorAddressArg, {});
     delegatorAddress = addressToChecksummedAddress(delegatorAddressArg);
+    _assertSignerIsNotDelegator(signerAddress, delegatorAddress);
   }
 
   // All message construction (KMS context read + extraData version assert, duration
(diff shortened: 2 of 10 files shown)


sdk/js-sdk/test/fheTest/viem-common/clientDecrypt.unifiedPermit.tests.ts
```diff
@@ -176,6 +176,74 @@ export function defineClientDecryptUnifiedPermitTests(parameters: {
...
+      const parsed = await client.parseSignedDecryptionPermit({
+        serializedPermit: serialized,
+        transportKeyPair,
+      });
+      expect(parsed.version).toBe(2);
+      expect(parsed.isDelegated).toBe(true);
+      expect(parsed.signerAddress.toLowerCase()).toBe(config.bob.account.address.toLowerCase());
+      expect(parsed.encryptedDataOwnerAddress.toLowerCase()).toBe(config.alice.account.address.toLowerCase());
+    });
+
```
geoxel, line 246: Could we have the same test for ethers lib (for consistency) : `sdk/js-sdk/test/fheTest/ethers-common/clientDecrypt.unifiedPermit.tests.ts` ?

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
geoxel, line 172: `SignedDecryptionPermitV1-p.ts:126` checks for `signerAddress` and `delegatorAddress` equality and throws like:

```
    if (signerAddress.toLowerCase() === delegatorAddress.toLowerCase()) {
      throw new Error(
        'signerAddress and delegatorAddress must be different. ' +
          'Use a non-delegated permit to decrypt your own values.',
      );
    }
```
Maybe worth adding?

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
```diff
@@ -196,6 +196,14 @@ export async function parseSignedDecryptionPermitV2(
   assertRecordBytesHexProperty(permit, 'signature', permitName, options);
   assertRecordAddressProperty(permit, 'signerAddress', permitName, options);
 
+  // Delegation is post-sign metadata, not part of the signed eip712 message,
+  // so it's read from the permit object itself (mirrors signDecryptionPermitV2).
+  let delegatorAddress: ChecksummedAddress | undefined;
+  if (isRecordNonNullableProperty(permit, 'delegatorAddress')) {
+    assertRecordAddressProperty(permit, 'delegatorAddress', permitName, options);
+    delegatorAddress = addressToChecksummedAddress(permit.delegatorAddress);
+  }
```
geoxel, line 226: Same thing here.

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
```diff
@@ -219,6 +227,7 @@ export async function parseSignedDecryptionPermitV2(
     signature: permit.signature,
     eip712,
     signerAddress,
+    delegatorAddress,
   });
 }
```
geoxel, line 251: In `sdk/js-sdk/src/core/kms/SignedDecryptionPermitV1-p.ts:220` we could add a check to prevent the user from specifying a `delegatorAddress` in a V1 permit:

```
  // V1 delegation lives inside the signed eip712 message (primaryType
  // 'DelegatedUserDecryptRequestVerification'), unlike V2 where it's post-sign
  // metadata alongside the message. A top-level delegatorAddress here almost
  // certainly means the permit was built for V2 and mistakenly parsed as V1 —
  // reject it instead of silently ignoring it.
  if (isRecordNonNullableProperty(permit, 'delegatorAddress')) {
    throw new Error(
      `Unexpected 'delegatorAddress' field on a V1 permit. Delegation for V1 permits is encoded in ` +
        `the signed eip712 message (primaryType 'DelegatedUserDecryptRequestVerification'), not as a ` +
        `top-level field. Did you mean to parse this as a V2 (unified) permit?`,
    );
  }
```

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
eitjuh: Fixed

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
eitjuh: Fixed

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
eitjuh: Fixed

sdk/js-sdk/test/fheTest/viem-common/clientDecrypt.unifiedPermit.tests.ts
eitjuh: Fixed

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
```diff
@@ -219,6 +227,7 @@ export async function parseSignedDecryptionPermitV2(
...
```
geoxel, line 252: In `src/core/types/signedDecryptionPermit.ts:56` we could change the comment to :
```
 * Uses the unified `KmsUserDecryptEip712V2` EIP-712 shape — there is no longer
 * a separate delegated variant. `eip712.message.userAddress` always equals
 * `signerAddress`; delegation is post-sign metadata carried alongside the
 * signed message, not part of it, so `encryptedDataOwnerAddress` may differ
 * from `signerAddress` even though the eip712 message never does.
```

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
geoxel, line 173: We could also add a test for that. Same for the parsing below.

sdk/js-sdk/src/core/actions/chain/serializeSignedDecryptionPermit.ts
```diff
@@ -17,6 +17,8 @@ export type SerializeSignedDecryptionPermitReturnType = {
   readonly eip712: Eip712Like;
   readonly signature: string;
   readonly signerAddress: string;
+  /** V2 (unified) permits only — delegation is metadata alongside the signed message, not part of it. */
+  readonly delegatorAddress?: string | undefined;
```
geoxel, line 20: We can update the doc here `docs/api-reference.md:284` and here `docs/api-reference.md:292`.

Don't update the full doc, because a PR is coming for that 🙂 : https://github.com/zama-ai/fhevm/pull/3585

eitjuh: Fixed

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
eitjuh: Fixed

sdk/js-sdk/src/core/kms/SignedDecryptionPermitV2-p.ts
eitjuh: Fixed

geoxel (approved):
LGTM !

PR #3660, opened 2026-08-28
feat(solana): delegated user-decrypt — connector, relayer, SDK, Squads e2e (fhevm-internal#1690)

KitHat:
Delegated user decryption for the Solana host (fhevm-internal#1690). A delegator —
a wallet or a program-controlled PDA (e.g. a Squads vault signing via CPI) — grants
a delegate the right to request user decryption of its encrypted values. Grants are
scoped per encrypted value account authority or via the wildcard row (`[0xff;32]`),
expire by slot, and are revocable. On-chain instructions landed earlier on this
branch; this PR wires every consumer. EVM RFC-017 parity verified against `ACL.sol`
(lower-bound-only expiry, pause-gated grant/revoke, same wildcard/self refusals).

### Shared crate (`zama-solana-acl`)
- One delegation-record decoder + liveness rule (`is_live_at`) for every off-chain
  reader; both the connector and the relayer route liveness through it.
- The host program imports `DELEGATION_SEED` and the wildcard sentinel from the
  crate; the layout is pinned by a host-side serialize→decode round-trip test.

### KMS Connector (authoritative check)
- Delegated entries (subject ≠ permit signer) authorize against the same atomic
  `confirmed` snapshot as the rest of the request: the authority-specific row or
  the delegator's wildcard row; the audit record names which row carried the grant.
- Sentinel-authority guard: an encrypted value account naming `[0xff;32]` as its
  authority is refused at resolution — otherwise the authority-specific row
  derivation lands on the wildcard row itself. Covered by a cross-implementation
  auth vector whose world holds exactly the row an unguarded check would accept.

### Relayer (advisory pre-check)
- Negative-only: refuses only what the connector could never authorize — both rows
  definitively dead (absent / revoked / expired) at one `getMultipleAccounts` slot;
  every data ambiguity passes. Transport failure refuses, same policy as the EVM
  pre-check. Motive: the Decryption contract has no rejection channel, so a doomed
  request otherwise costs a gateway transaction and dies by timeout.
- The chain is picked from the permit's signed `chain_id`, never from unsigned
  handle bytes; PDA derivations are pinned against the host program's fixtures;
  the plan/judge core is pure (row pairing by construction, no index arithmetic —
  a pairing defect passes with a warning instead of misattributing rows).

### SDK
- Instruction builders (`delegateForUserDecryption`, `revokeDelegationForUserDecryption`,
  `revokePermits`), a both-rows record reader, the connector's liveness boundary,
  and a wildcard-grant warning. Readers are owner-checked (a dusted PDA reads as
  absent, not as layout drift) and byte-strict against the Rust twin decoders.
- All addresses, discriminators and instruction bytes cross-pinned by shared
  fixture literals (TS ↔ Mollusk ↔ relayer).

  `confirmed` snapshot as the rest of the request: the authority-specific row or
  the delegator's wildcard row; the audit record names which row carried the grant.
- Sentinel-authority guard: an encrypted value account naming `[0xff;32]` as its
  authority is refused at resolution — otherwise the authority-specific row
  derivation lands on the wildcard row itself. Covered by a cross-implementation
  auth vector whose world holds exactly the row an unguarded check would accept.

### Relayer (advisory pre-check)
- Negative-only: refuses only what the connector could never authorize — both rows
  definitively dead (absent / revoked / expired) at one `getMultipleAccounts` slot;
  every data ambiguity passes. Transport failure refuses, same policy as the EVM
  pre-check. Motive: the Decryption contract has no rejection channel, so a doomed
  request otherwise costs a gateway transaction and dies by timeout.
- The chain is picked from the permit's signed `chain_id`, never from unsigned
  handle bytes; PDA derivations are pinned against the host program's fixtures;
  the plan/judge core is pure (row pairing by construction, no index arithmetic —
  a pairing defect passes with a warning instead of misattributing rows).

### SDK
- Instruction builders (`delegateForUserDecryption`, `revokeDelegationForUserDecryption`,
  `revokePermits`), a both-rows record reader, the connector's liveness boundary,
  and a wildcard-grant warning. Readers are owner-checked (a dusted PDA reads as
  absent, not as layout drift) and byte-strict against the Rust twin decoders.
- All addresses, discriminators and instruction bytes cross-pinned by shared
  fixture literals (TS ↔ Mollusk ↔ relayer).

### e2e
- Two live arcs: headless wallet (grant → delegate decrypts → identical repeat is
  one job → revoke → refused with `not_allowed_on_host_acl`) and a real Squads v4
  2-of-3 multisig granting through its vault (below-threshold execution refused by
  the Squads program itself).
- Squads fixtures are fetched from mainnet and sha256-pinned: verify-then-rename
  (nothing unverified on disk), PIN MISMATCH fatal vs offline soft-skip; the arc
  probes the running validator so stale-boot fails legibly.

### Tests
- Mollusk 39 (CPI-signed vault delegator, self-paid grant, prefunded and
  foreign-squatted record address, cross-position discriminator substitution);
  connector auth vectors + routing/delegation suites; relayer 254; SDK 409.


relayer/src/host/acl_checker.rs
```diff
@@ -593,6 +796,87 @@ fn group_handle_entries_by_chain(handles: &[HandleEntry]) -> HashMap<u64, Vec<Ha
...
+        "params": params,
+    });
+
+    let response = chain
+        .http
+        .post(chain.rpc_url.clone())
+        .json(&body)
+        .send()
+        .await
+        .map_err(|e| format!("transport: {e}"))?;
```
Eikix: [AUTOMATED] [P1] Redact the Solana RPC URL before this error is stored.

Reqwest can include the full request URL in a transport error. Provider API keys are commonly stored in the URL path or query. This string is logged, stored as `HostAclFailed`, and returned by the status API.

The existing `redact_error(&e)` helper should be used for both Reqwest error mappings. A regression test should verify that fake path and query secrets do not appear in the returned error.

relayer/src/host/acl_checker.rs
```diff
@@ -518,6 +545,182 @@ impl HostAclChecker {
...
+            })
+            .collect();
+        if delegated.is_empty() {
+            return Ok(());
+        }
+
+        // Round 1: the encrypted value accounts, to learn each entry's authority.
+        let encrypted_value_addresses =
+            encrypted_value_read_addresses(&delegated, chain.program_id);
+        let (_, encrypted_value_accounts) = self
```
Eikix: [AUTOMATED] [P1] Preserve the first read slot when the delegation rows are read.

The first response slot is discarded here. The second RPC call has no `minContextSlot` requirement. A load-balanced RPC can return the value account at slot 500 and the delegation rows at slot 480. A grant confirmed between those slots then appears absent, and a valid request is rejected before the Connector can decide.

The first slot should be passed as `minContextSlot` to the second read. The returned slot should also be checked. A backward read must be treated as indeterminate and passed to the Connector.

relayer/src/host/acl_checker.rs
```diff
@@ -77,37 +81,60 @@ struct HostChainAcl {
...
+                        })?;
+                let rpc_url = Url::parse(&hc.url).map_err(|e| {
+                    anyhow::anyhow!("Invalid host chain URL for chain {}: {}", hc.chain_id, e)
+                })?;
+                solana_chains.insert(
+                    hc.chain_id,
+                    SolanaHostChain {
+                        rpc_url,
+                        program_id,
+                        http: reqwest::Client::new(),
```
Eikix: [AUTOMATED] [P2] Add a deadline to each Solana RPC attempt.

`reqwest::Client::new()` has no request timeout. If the configured RPC accepts a connection but never finishes the response, the retry loop is never reached. One readiness permit is held forever. Enough stalled calls can consume all readiness permits.

A per-attempt timeout should be configured under `host_acl_check`. A test server that accepts a connection but does not reply should be used to verify the timeout and final `CallFailed` result.

test-suite/fhevm/e2e/scenarios/delegated-user-decrypt.scenario.test.ts
```diff
@@ -0,0 +1,346 @@
...
+      });
+      expect(revokedRows.exact?.revoked).toBe(true);
+      await expect(runSolanaCurrentUserDecrypt(environment)).rejects.toThrow(
+        /not_allowed_on_host_acl/,
+      );
+    },
+    SCENARIO_TIMEOUT_MS,
+  );
+
+  test.skipIf(!squadsAvailable)(
```
Eikix: [AUTOMATED] [P2] Require this Squads scenario in the GitHub E2E lane.

`clean-e2e.sh` treats a Squads fixture download failure as a warning. This `skipIf` then lets CI pass without the only real 2-of-3 Squads delegation flow.

Offline local runs can keep the optional behavior. CI should set an explicit requirement flag and fail when the fixtures are unavailable.

Eikix:
[AUTOMATED]

## Non-blocking delegation-scope note

An exact Solana delegation is scoped to the encrypted value account authority. It is not scoped to one encrypted value ID.

The grant therefore covers each value for which both conditions hold:

- The encrypted value uses the delegated authority.
- The delegator is a subject for that value.

For the confidential-token program, this scope is safe. The token account authority PDA includes the mint, which is also the domain. A grant cannot cross into another mint.

The same grant can cover the balance, transferred amount, burned amount, and their historical handles. This is coherent when access is granted to the confidential token account.

A generic application can reuse one wallet or PDA authority across several domains. Such a grant would cover those domains. This is not a defect if authority-wide access is intended.

No code change is required for this PR. The authority-wide property should be documented. Applications that need domain isolation should derive the authority from the domain.

A focused test with one authority under two domains could also pin this intended behavior.

Eikix (commented):
[AUTOMATED] [P1] Retry a missing delegation account

A confirmed delegation can be visible to the Relayer RPC and absent from the Connector RPC. The two components read independently and can observe different confirmed slots.

A missing delegation is currently classified as terminal. The request is then marked as failed. It is not retried after the Connector RPC catches up. A valid delegated decryption can therefore fail permanently during normal RPC lag.

`DelegationFailure::Absent` should be classified as transient. Revoked, expired, malformed, foreign-owner, and tuple-mismatch results should remain terminal.

Eikix (approved):
LGTM!

PR #3665, opened 2026-08-29
chore(listener): final events feature

nboisde:
# Listener v0.2.0 — Finality feature

This PR adds end-to-end finality support to the listener and its consumer library: consumers can register FINAL watchers and receive events only once their block is finalized (never reorged), request historical replays of finalized ranges, and run live and final pipelines side by side. No breaking changes — everything final is additive.

## Design choices

**Separated flows.** The finality pipeline is a full parallel of the live flow (own queue, own advisory lock, own `final_blocks` tip table, own delivery/cleaner/catchup queues) instead of an extension of it. The point is isolation: a finality stall never impacts the real-time algorithm, and vice versa. Since finalized blocks cannot reorg, the final pipeline also drops all reorg machinery while keeping the same fetcher, error classification, and publish-before-commit (at-least-once, zero loss) guarantees.

**Two finality strategies (`finality_tag`).** Not every chain/RPC exposes a `finalized` tag, so one switch supports both: `true` asks the node for the finalized block (consensus truth), `false` (default) approximates it as `head - finality_depth` (portable heuristic).

**Disclaimer on the tag strategy:** with `finality_tag: true` the finalized head advances by epochs — the lag moves in "stairs", often sitting more than the 64-block figure behind on Ethereum, and observed finality can exceed ~12 minutes. This is normal consensus behavior, not a stall — which is also why finality stall is deliberately not an alert signal (the `listener_finality_*` metrics keep it observable).

**Other choices**

- LIVE/FINAL watcher types live in the same `filters` table (`filter_type` enum): uniqueness is per (chain, consumer, addresses, type), and each flow queries only its own type through dedicated partial indexes — the same address can be watched live and final simultaneously.
- Wire compatibility: `FilterCommand.filter_type` is optional and omitted when unset, so legacy payloads and consumers keep the exact byte format; a missing type defaults to a LIVE watcher.
- Every self-perpetuating flow (fetch, finality, both cleaners) holds its own salted advisory lock: HPA-safe deduplication where duplicate broker deliveries are acked and skipped, and no two flows can starve each other.
- `finality_active` (default `true`) gates the whole final side — flow, cleaner, catchup — so a chain can run live-only with zero final-side queries; requests arriving while inactive are dropped with identifying logs.
- Config is reused rather than duplicated: the final cleaner shares `cleaner.*`, the final catchup shares `catchup.*`, and the consumer library's final flows share the live/catchup tuning options.

## Changes

**Database / migration**
- `filter_type` SQL enum (`LIVE`/`FINAL`) + column on `filters`, uniqueness extended per type, partial indexes per type; old chain-only index dropped.
- New `final_blocks` tip table (no status column — finalized blocks never reorg), unique per (chain_id, block_number) and (chain_id, block_hash).
- Single re-versioned migration for the whole PR; sqlx offline cache regenerated.

**Watchers / control plane**
- `FilterCommand.filter_type: Option<FilterType>` (serde-defaulted, wire-compatible); watch/unwatch handlers thread the type to the DB.
- FINAL filter queries (`get_final_filters_by_chain_id` / `get_final_filters_by_consumer_id`).

**RPC provider**
- `finality_tag` / `finality_depth` on `SemEvmRpcProvider`; `get_final_block_number()` with the two strategies (`get_finalized_block_number_from_tag` / `_from_depth`), shared `hex_to_u64` helper.

**Finality flow (listener core)**
- `fetch-final-block` queue + `FinalityHandler` (own FlowLock salt, 4-step lock/process/release/republish contract, inactive early-out).
- `EvmListener::fetch_final_blocks` (tip from `final_blocks`, head from `get_final_block_number`, parallel fetch + in-order `final_processing`), `validate_and_init_final_block` boot anchor (crash-loop pattern, anchored at the finalized head).
- `FinalBlockRepository` (insert, latest, min, retention delete — all chain-scoped).

**Publisher**
- `BlockFlow::Final` and `BlockFlow::FinalCatchup` wire tags; `FINAL_EVENT` / `FINAL_CATCHUP_EVENT` routing keys and helpers.
- `publish_final_block_events` / `publish_final_catchup_block_events`, mirroring the live paths through the shared queue-existence/stale-publish helper; publish-before-commit wired into `final_processing`.

**Cleaners**
- Broker-dedup FlowLock added to the existing cleaner flow (pre-existing gap: duplicated deliveries could multiply the clean loop).
- `Cleaner::run_final` + `clean-final-blocks` queue + `FinalCleanerHandler` (own lock salt), gated on `cleaner.active && finality_active`; `blocks_to_keep` validation decoupled from `finality_depth` (static floor >= 999).

**Final catchup**
- `final-catchup` (orchestrator, clamps to the finalized head, splits sub-ranges) → `range-final-catchup` (workers) → `{consumer_id}.final-catchup-event`; dead-letter on malformed payloads, finality-inactive drops with identifying logs; reuses `catchup.*` config.

**Consumer library**
- Registration: `create_final_filter_on_log_address`, `create_final_full_block_filter`, `register_final_contracts` / `unregister_final_contracts`, `register_final_full_block` / `unregister_final_full_block`.
- Consumption: `consume_final`, `consume_final_catchup`, `request_final_catchup`, `ensure_final_consumer`, `ensure_final_catchup_consumer`, topics accessors, `cancel_final` / `cancel_final_catchup` (cancellation tree extended to four child tokens); `BlockFlow` re-exported.

**Example crate**
- Reworked into a thin `main.rs` wiring `live_events.rs` (low-level FilterCommand style, live + catchup) and `final_events.rs` (high-level contract API, final + final catchup) on the same token; `full_block.rs` shows both wildcard subscriptions; flow-tagged logging throughout.

**Config**
- `finality_tag` (default false), `finality_active` (default true); pool floor raised to 8 and default max connections to 12 (four advisory-lock holders); yaml examples updated.

**Metrics / dashboards**
- Nine new chain-labeled metrics: finality iterations/tip/height/range-duration/active + de-conflated `listener_final_catchup_*` set (the final catchup no longer pollutes live catchup series); queue-depth poller now covers all finality and catchup queues.
- Grafana: "Finality" and "Final Catchup" rows on the per-chain dashboard, "Finality — Fleet" row on the fleet dashboard; templating untouched.

**Docs & tests**
- `docs/error_flows.md` (full failure-behavior matrix per flow: Postgres/broker/RPC outages, missing consumer queues), metrics catalog and dashboard docs updated, CLAUDE.md routing/error notes.
- Serde wire-format tests (filter type, flow tags), routing non-collision tests, consumer e2e round-trips (register FINAL, final-catchup request wire, both delivery queues) — all green alongside the untouched existing suites.

**Versioning**
- Changed crates bumped to 0.2.0 (`listener_core`, `consumer`, `primitives`, `example`); untouched crates stay 0.1.0 (`broker`, `telemetry`, `test-support`); Helm chart `0.2.0` / appVersion `0.2.0`; both `Cargo.lock` files (listener + coprocessor engine) refreshed.

## Deploy notes

- Deploy the listener before consumers start using the FINAL registration API: an old listener ignores the unknown `filter_type` field and would silently create a LIVE watcher from a FINAL request.
- The migration was re-versioned into a single file for this PR: local/dev databases that applied the intermediate version need a reset; released environments are unaffected.
- The chart's default image tag now resolves to `0.2.0` (appVersion): publish the corresponding release/image tag before deploying the chart with an empty `image.tag`, or override the tag explicitly.


listener/crates/listener_core/src/config/config.rs
```diff
@@ -436,18 +454,14 @@ impl BlockchainConfig {
                 "chain_id must be strictly positive".to_string(),
             ));
         }
-        if self.cleaner.blocks_to_keep < 2 {
+        // Static floor, deliberately decoupled from finality_depth: the finality
+        // boundary can move at runtime, so retention must not depend on it.
+        if self.cleaner.blocks_to_keep < 999 {
```
rudy-6-4, line 459: Maybe 999 should be a named const. Does it assume finality is never bigger than 999 blocks late ?.

listener/crates/listener_core/src/core/cleaner.rs
```diff
@@ -84,16 +101,69 @@ impl Cleaner {
...
+    /// another chain's final blocks.
+    pub async fn run_final(&self) -> bool {
+        if !self.final_active {
+            info!("FinalCleaner: inactive — skipping cleanup and not re-triggering");
+            return false;
+        }
+
+        match self
+            .final_blocks
+            .delete_blocks_keeping_latest(saturating_u64_to_i64(self.blocks_to_keep))
```
rudy-6-4, line 127: can't it be done from the last final ?

nboisde: What do you mean ? Its done from the last final block inserted into final_blocks table already.

listener/crates/listener_core/src/config/config.rs
nboisde: True It could be for const.
I assume there is no such chain where the finality is above 1k blocks indeed. 
The goal there is to keep minimal blocks into the db, to be able to sustain underlying logic.
But could be settled above if we integrate a new chain which is above in term of number of blocks for finality.

listener/crates/listener_core/src/core/evm_listener.rs
```diff
@@ -1123,6 +1130,522 @@ impl EvmListener {
...
+        // authority on what actually made it into the DB. Same matrix as the
+        // cursor flow, minus the reorg arm (finalized blocks never reorg).
+        match (final_outcome, fetcher_outcome) {
+            // === SUCCESS PATH ===
+            // All blocks processed and inserted. Sleep before returning to avoid hammering RPC.
+            (Ok(()), Ok(())) => {
+                info!(
+                    range_start = range_start,
+                    range_end = range_end,
+                    "Finality iteration complete — all final blocks processed and inserted"
```
rudy-6-4, line 1612: isn't it just for a page (range_size) ?

listener/crates/listener_core/src/core/evm_listener.rs
```diff
@@ -1123,6 +1130,522 @@ impl EvmListener {
...
+    /// 3. Calculates the range of blocks to fetch (capped by `range_size`)
+    /// 4. Spawns two concurrent tasks via `tokio::spawn`:
+    ///    - **Producer** (`fetch_blocks_in_parallel`): fetches blocks via
+    ///      parallel RPC calls, fills an `AsyncSlotBuffer` out-of-order
+    ///    - **Consumer** (`final_processing`): reads slots sequentially in
+    ///      order and inserts them into `final_blocks`
+    /// 5. Awaits both tasks and analyzes outcomes
+    ///
+    /// # Returns
+    /// - `Ok(())` — up-to-date (nothing to fetch) or all blocks processed;
```
rudy-6-4, line 1464: isn't it just for a page (range_size) ?

listener/crates/listener_core/src/store/flow_lock.rs
```diff
@@ -48,23 +48,80 @@ impl Drop for FlowLockGuard {
     }
 }
 
+/// Salt XORed into the chain_id to derive the cleaner lock key.
+/// Keeps cleaner mutual exclusion independent from the fetch/reorg lock
+/// (raw chain_id): the cleaner holds its lock across the cron sleep and
+/// must never block or be blocked by the cursor flows.
+const CLEANER_LOCK_SALT: i64 = 0x434C_4541_4E45_5221; // "CLEANER!"
```
rudy-6-4, line 55: You need them to be integer ? How are you sure there is no collision ?

rudy-6-4 (approved):

listener/crates/listener_core/src/core/evm_listener.rs
nboisde: Yes, its only for a range size and new blocks will be proceed at the next loop iteration if there is some to add.

listener/crates/listener_core/src/store/flow_lock.rs
nboisde: As a quick comment, there is a collision possible but above chain id 2^50, we could improve this on a next iteration partitionning the bits instead of XORing: key = (flow_id << 60) | chain_id, with a config validation that rejects chain_id >= 2^60.

listener/crates/listener_core/src/core/evm_listener.rs
nboisde: Same comment than below.

nboisde:
@mergifyio queue

PR #3671, opened 2026-08-31
fix(sdk): use parentPort in TFHE workers

Eikix:
Bun 1.4 aligned `worker_threads` with Node ([oven-sh/bun#37075](https://github.com/oven-sh/bun/pull/37075)): parent messages now go only to `parentPort`, not `self` — bun 1.3 had been double-delivering them, which hid that our TFHE workers were listening on the wrong channel.

## Summary
- TFHE rayon workers now use the same Node-like check as `environment.ts` (`process.versions.node`). `addEventListener` is treated as a browser only when that check is false, so bun/Node workers listen on `parentPort` instead of `self`.
- That matches Node (and bun 1.4+) message routing. On bun 1.4, parent messages no longer reach `self`, so `initThreadPool` previously hung forever waiting for `wasm_bindgen_worker_ready`.
- Regenerated embedded workers/hashes for TFHE 1.5.3, 1.6.0-dev, and 1.6.2.

This is the SDK-side fix for the solana-e2e encrypt-path hang after unpinned `setup-bun` started installing bun 1.4.0. `feature/solana` still needs this commit (merge or cherry-pick) before those jobs go green.

## Test plan
- [x] `tfheWorkerChannel.test.ts`: generated `tfhe-worker.mjs` files start with the template bootstrap
- [x] bun 1.3.14 (local): `initAsync` + `initThreadPool(2)` still completes
- [x] bun 1.4.0 (local): same path now reaches worker `ready` (previously hung)
- [ ] `npx vitest run --config src/vitest.config.ts src/wasm/tfhe/tfheWorkerChannel.test.ts` in `sdk/js-sdk`


Eikix:
Hi @0xalexbel ! I had a bug in the Solana PoC regarding bun in the worker. 
TLDR is that Bun is Nodelike but still exposes browser function in its workers like addEventListener which makes the tfhe worker think it is a browser and not a node env!

let me know if this fix is ok?

Eikix:
This PR is not up to date anymore.
But if the problem was solved in another way we can close it now. Any opinion @0xalexbel?

PR #3672, opened 2026-08-31
chore(kms-connector): decryption source in db

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/1951
Closes https://github.com/zama-ai/fhevm-internal/issues/1962

### Why

With the decryption-via-HTTP process, the kms-connector now needs to return the error code and context to the caller of the decryption request (i.e. the relayer), so our error handling must be tweaked a bit. Today the worker only classifies errors internally (recoverable/irrecoverable) and persists a generic event `status`. This PR introduces granularity at error-construction site so the response tables can carry it.

### What

Attaches a caller-facing `ErrorCode` (from the `kms-connector-api` crate) to every processing error produced by the `kms-worker`, and restructures error handling so the retry/terminal decision lives in one place.

> **Note:** the new `code` field is intentionally write-only for now. It is groundwork for upcoming work that will store the code in the `error_code` column of the response tables, so the HTTP decryption API can return precise error responses (`ErrorResponse { code, message, retryable }`) instead of a generic status. No consumer reads it in this PR.

#### `ProcessingError`: enum → struct

The 3-variant enum (`Recoverable` / `Irrecoverable` / `Aborted`) becomes a struct with three orthogonal pieces:

- `class: ProcessingErrorClass` — the connector-internal recoverability classification (drives retry vs. terminal handling, unchanged semantics),
- `code: ErrorCode` — the caller-facing error code (new, unused for non-decryption events and gateway-sourced decryption),
- `source: anyhow::Error` — the cause chain.

#### Error dispatch moved from the processor to the worker

`EventProcessor::process` now returns `Result<Option<KmsResponseKind>, ProcessingError>` instead of swallowing errors and returning `Option`. The error handling match moves from `DbEventProcessor::process` into `KmsWorker::handle_processing_error`:

This puts the error policy next to the response-publishing policy and leaves the processor with pure processing logic. `error_counter` / `already_sent` bookkeeping is unchanged.

### Misc

- `getKmsContextAnchor` is now read at the finalized block (`.block(BlockId::finalized())`), aligning it with the other finalized reads in `context.rs` and protecting the previous-context lookup against reorgs. If the anchor is not yet finalized the lookup fails as `transient` and the event is retried until finality.


kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
```diff
@@ -213,17 +215,23 @@ fn aggregate_resolved_handles(
     resolved_handles: Vec<ResolvedHandle>,
 ) -> Result<VerifiedCiphertexts, ProcessingError> {
     let Some(key_id) = resolved_handles.first().map(|r| r.key_id) else {
-        return Err(ProcessingError::Recoverable(anyhow!("no handles resolved")));
+        return Err(ProcessingError::recoverable(
+            ErrorCode::CoproConsensusFailed,
```
dvdplm, line 219: nitpick: do you think naming it `ErrorCode::CoproConsensus` would work? Something clearly failed given we're returning an error so `Failed` is kind of implied?

dvdplm: …and I guess we could take it one step further and change `ErrorCode` to just `Code`: it is pretty clear from the context that it's an error?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -1,30 +1,101 @@
...
-#[derive(Debug, Error)]
-pub enum ProcessingError {
-    #[error("Processing failed with irrecoverable error: {0:#}")]
-    Irrecoverable(anyhow::Error),
-    #[error("Processing failed: {0:#}")]
-    Recoverable(anyhow::Error),
-    #[error("Processing stopped: the operation was aborted on the KMS Core")]
+#[derive(Debug)]
+pub struct ProcessingError {
+    pub class: ProcessingErrorClass,
```
dvdplm: nitpick: I don't love the name here. Did  you consider  using `kind`, to stay aligned with https://doc.rust-lang.org/std/io/enum.ErrorKind.html ?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -1,30 +1,101 @@
...
+    pub source: anyhow::Error,
+}
+
+/// Recoverability classification of a [`ProcessingError`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProcessingErrorClass {
+    Recoverable,
+    Irrecoverable,
     Aborted,
 }
```
dvdplm, line 23: The PR dewscription describes this as the internal decision maker on whether we can retry this or should give up.

- Can it be `pub(crate)`?
- If not, should it be `non_exhaustive` to signal callers that there may be new variants here some day?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -1,30 +1,101 @@
...
+
+    /// Wraps the inner error with additional context.
+    pub fn context(mut self, ctx: String) -> Self {
+        self.source = self.source.context(ctx);
+        self
+    }
+
+    pub fn details(&self) -> String {
+        format!("{:#}", self.source)
+    }
```
dvdplm: not sure I agree that we need this one-liner

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
```diff
@@ -59,56 +57,58 @@ where
...
+
+            let (error_count, result) = self.kms_client.send_request(&request).await;
+            event.error_counter += error_count;
+            let send_response = result?;
+            event.already_sent = true;
+
+            // Invalidate epochs associated to destroyed context returned by the KMS Core.
+            // A failure must not prevent the event from completing, as retrying would result in an
+            // `NotFound` from KMS Core with no epoch IDs.
+            if let KmsSendResponse::DestroyedEpochs(epoch_ids) = send_response {
```
dvdplm, line 86: Not new code, I know, but is it expected that the KMS has a destroyed epoch that the connector does not know about? Or is this check here to guard against an anomaly? If it's the latter, should we log a warning here? Like `warn!("KMS epoch IDs {ids} are destroyed; we are out of sync?");`

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
```diff
@@ -162,7 +165,7 @@ impl<P: Provider> ProtocolConfigProcessor<P> {
             .call()
             .await
             .map_err(|e| {
-                ProcessingError::Recoverable(anyhow!("getCompletedKeyIds call failed: {e}"))
+                ProcessingError::transient(anyhow!("getCompletedKeyIds call failed: {e}"))
```
dvdplm, line 168: Maybe the `anyhow!`  call can move into  `transient()` so we'd have `ProcessingError::transient("bla bla bla", e)`?

dvdplm (dismissed):
lgtm, comments are mostly naming-nitpicks.

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: Yes I considered it, argued a bit with the bot who convinced me to keep class.

Main reason is that in the code base, we use `kind` for enum that really represent different kind of stuff.
Like we have `ProtocolEventKind` to represent both a decryption and a keygen event under the same enum, and they are completely different.
This is not really the case here, we have different class of errors, but they still are processing error regardless...

But no strong opinion, I could go for the kind approach tbh

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: - No as it is used in integration test
- Not sure it matters tbh, this enum is internal, just `pub` for integration test

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: Fair

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
eudelins-zama: I'd rather keep it simple like this. It allows for `.map_err(ProcessingError::transient)` notation for ex, instead of `.map_err(|e| ProcessingError::transient("", e))`  with your suggestion :slightly_smiling_face:

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
eudelins-zama: Not sure I understand your Q.

When kms-connector send a context destruction request, the KMS replies with the list of epoch that were tied to that context and thus destroyed along the way.
This code is simply to handle the invalidation of these epoch.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: `ErrorCode`: I wouldn't do it, this type is a shared one with the Relayer (from `kms-connector-api` crate), and I think it's clearer tbh.

`CoproConsensusFailed`: I understand from the typing perspective and I could do it. But if we do, this means the Relayer would receive as `ErrorResponse`:

```json
{
    "code": "copro_consensus",
    ...
}
```

instead of:

```json

```json
{
    "code": "copro_consensus_failed",
    ...
}
```

I personally prefer the later response, but that is a detail.

Ofc we could use annotation to have `CoproConsensus` being ser/deser as "copro_consensus_failed", but why bother...

dvdplm: Hmm, I somehow thought `ErrorCode` would be a number, or something like `E308`, not a string intended for human readers. If `ErrorCode` stays as it is, wouldn't it be serialized as `error_code`? and so callers would see 

```json
{ error_code: "copro_consensus" }
```

…in which case it might work out?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
dvdplm: Fair. In another point in the PR there's a match on a tuple of (error.class, event.kind) which reads a bit funny, hence this comment. You decide!

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
dvdplm: So the connector initiates the destruction but awaits confirmation from KMS before invalidating? I thought the  connector would invalidate its own storage, then send the destruction call. I guess it's possible that some other actor calls the KMS context destruction and the  connector only learns about it here?

Either way, the code is fine as-is.

kms-connector/crates/kms-worker/src/core/event_processor/protocol_config.rs
dvdplm: ah, indeed, good point.

kms-connector/crates/kms-worker/src/core/event_processor/ciphertext/manager.rs
eudelins-zama: > I somehow thought ErrorCode would be a number

We could discuss on this, but in current form, it would be a string because of how `ErrorResponse` and `serde` is implemented.

And it would be `{ "code": "copro_consensus"  }`

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
eudelins-zama: > I guess it's possible that some other actor calls the KMS context destruction and the connector only learns about it here?

The connector is the only entity allowed to talk with the KMS, not sure what you mean by that

> I thought the connector would invalidate its own storage, then send the destruction call.

This is halfway true. When connector receive a context destruction event, it invalidate this context in storage, but not the tied epochs, then forward to the kms-core. But I guess we could invalidate tied epochs as well indeed! I think at some point it was two distinct tables so it was not possible, but now we have a `context_id` in the epoch table

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: I'll do the renaming. Was just tricked by the bot into convincing me it was a good idea, but I don't even like it myself lol

eudelins-zama: Fixed b642b2809eda08ab7e0e034e1e029f44c4e7fde3

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
eudelins-zama: Fixed d771796aa5fdbc66c045a66252ff4b381e458d29

kms-connector/crates/kms-worker/src/core/event_processor/processor.rs
dvdplm: > The connector is the only entity allowed to talk with the KMS, not sure what you mean by that

This is not enforced by the code on either side. We can imagine a CLI tool used to administer a KMS node (or cluster) that can perform the same operations.

dvdplm: > But I guess we could invalidate tied epochs as well indeed! I think at some point it was two distinct tables so it was not possible, but now we have a context_id in the epoch table

I think that would be good! (not in this PR obviously)

dvdplm (dismissed):
still lgtm

eudelins-zama:
@mergifyio queue

dvdplm (dismissed):

eudelins-zama:
@mergifyio queue

eudelins-zama:
@mergifyio queue

dvdplm (approved):

eudelins-zama:
@mergifyio queue

eudelins-zama:
@mergifyio queue

eudelins-zama:
@mergifyio queue

PR #3684, opened 2026-09-01
feat(coprocessor): bind the output handle in the rerand transcript

antoniupop:
Per RFC 019, the per-operation seed transcript's function description is Encode(h_out, op): the output handle's preimage commits to the opcode, every operand handle and each operand's origin, making it the collision-resistant commitment to the evaluated function, while the opcode stays alongside it so the binding remains visible.

Everything else the RFC specifies is already the implemented behaviour: every encrypted operand of every operation is re-randomized, seeds are counter-indexed per encrypted-operand position, and RFC 020 selects each operand's base representation (raw for transaction-local, canonical decompressed for boundary).

This changes computed ciphertext bytes, so it requires a coordinated activation and fresh known-answer vectors.

### DIFF
--- a/coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
+++ b/coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs
@@ -261,14 +261,28 @@ impl<'a> Scheduler<'a> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


PanGan21 (approved):

Eikix (approved):
Lgtm

antoniupop:
@mergifyio queue

PR #3686, opened 2026-09-01
test(coprocessor): add benchmarks and test suite updates for GPU

antoniupop:
Adds CPU/GPU CI reporting for the seven one-shot benchmark scenarios, including Slab result parsing, partial-result reporting, and failure propagation. Updates GPU provisioning with Scaleway profiles, bump to CUDA 12.8/GCC 12, and resource logging.

Also seeds benchmark dependence-chain rows, aligns benchmark database configuration, adds consensus revision checks, waits for server-key material during bootstrap, and conditionally mounts the host Docker socket for E2E fault tests.

### DIFF
--- a/.github/actions/gpu_setup/action.yml
+++ b/.github/actions/gpu_setup/action.yml
@@ -1,13 +1,16 @@
-name: Setup Cuda
-description: Setup Cuda on Hyperstack or GitHub instance
+name: Setup CUDA
+description: Set up CUDA and its host compiler on a GPU runner
 
 inputs:
   cuda-version:
-    description: Version of Cuda to use
+    description: Space-separated acceptable CUDA versions; installs the first if none are present
     required: true
-  github-instance:
-    description: Instance is hosted on GitHub
-    default: 'false'
+  gcc-version:
+    description: Version of GCC to use as the CUDA host compiler
+    required: true
+  cloud-provider:
+    description: Cloud provider hosting the GPU runner
+    required: true
 
 runs:
   using: "composite"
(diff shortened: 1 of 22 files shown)


Eikix (dismissed):
[AUTOMATED]

## Full review at `0342810`

All 21 changed files were reviewed against the merged #3598 baseline. Existing PR comments, reviews and inline feedback were checked; the two findings below were not already posted.

Approval is being submitted with these non-blocking follow-ups. Could they be covered in the next PRs in the stack before the affected test paths are relied on?

1. **[P2] CUDA installation fails on Ubuntu 24.04.**

   [gpu_setup/action.yml:77–105](https://github.com/zama-ai/fhevm/blob/03428101b60bd1e108dd32ef41966810afddf384/.github/actions/gpu_setup/action.yml#L77-L105)

   The download URL is selected from the host's Ubuntu version, but the checksum is pinned to the Ubuntu 22.04 package. NVIDIA's Ubuntu 24.04 package has a different checksum. An Ubuntu 24.04 runner without the required CUDA toolkit therefore fails checksum validation before installation.

   Both public packages were downloaded and hashed independently:

   - Ubuntu 22.04: `d93190d50b98ad4699ff40f4f7af50f16a76dac3bb8da1eaaf366d47898ff8df`
   - Ubuntu 24.04: `d2a6b11c096396d868758b86dab1823b25e14d70333f1dfa74da5ddaf6a06dba`

   The checksum needs to be selected alongside the repository URL. Hosts that already have the required toolkit skip this installation path.

2. **[P2] The consensus revision check accepts mixed worker builds.**

   [compat.ts:559–588](https://github.com/zama-ai/fhevm/blob/03428101b60bd1e108dd32ef41966810afddf384/test-suite/fhevm/src/compat/compat.ts#L559-L588)

   The smallest reproduction is:

   - Node 0 uses `source.mode: local` with every service built locally.
   - Node 1 uses `source.mode: local` with `localServices: [host-listener]`.
   - The combined override is treated as fully local, so both nodes pass the revision check.
   - Generated Compose runs a local TFHE worker on node 0 and a published TFHE worker on node 1.

   This allows the consensus test to compare different worker builds despite the new same-revision check. Each node's service selection needs to be checked before the selections are combined. Guard acceptance and the generated image choices were both reproduced.

The earlier missing-chain benchmark finding is addressed by the shared staging helper. The reset now also respects the configured database name.

**Verification:** 376 CLI tests passed, along with TypeScript, workflow and shell checks. No GPU execution, cloud deployment or full Rust benchmarks were run as part of this review.

Eikix (approved):

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

antoniupop:
@mergifyio queue

PR #3697, opened 2026-09-02
chore(kms-connector): decryption source in db

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/1962

### Context

Decryption requests will soon be submitted directly to the Connector over HTTP, in addition to the onchain path through the Gateway. The `kms-worker` and the database need to know where a request came from, because HTTP-sourced decryptions must be ignored by the `tx-sender`. Also, any failure must be returned to the caller as a response instead of being silently retried.

### Database migration

A new `request_source` enum (`onchain` | `http`) is added as a `source` column on the four decryption request and response tables, defaulting to `onchain` so existing rows and an old `gw-listener` keep working. The response tables also gain nullable `error_code` and `error_details` columns, and their payload columns become nullable, with a CHECK constraint enforcing that a response row is either a payload or an error but never both.

The existing `tx-sender` notify triggers are restricted to `onchain` responses, and new row-level triggers emit `pg_notify` on HTTP-sourced response inserts and updates, carrying the hex-encoded `decryption_id` so the future Connector endpoint can wake up the right waiting connection.

### `kms-worker`

`KmsResponse` and `ProtocolEvent` now carry a `RequestSource`, propagated from the picked event to the published response. The response publisher writes decryption responses and completes the associated request row in a single statement, with the response status set to `pending` for onchain requests (until the `tx-sender` publishes them) and `completed` for HTTP requests (terminal on insert). Upserts follow one rule: a retry may override a previous error row but never an existing successful response.

In the worker, a failure on an HTTP-sourced decryption short-circuits the usual pending/failed/aborted retry logic and is stored as an error response instead.

### `gw-listener` and `tx-sender`

The `gw-listener` explicitly tags every stored decryption request as `onchain`. The `tx-sender` response picker only selects `pending` responses whose `source` is `onchain`, so HTTP-sourced responses never reach the Gateway.


kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
```diff
@@ -75,47 +79,92 @@ impl KmsResponsePublisher for DbKmsResponsePublisher {
...
+                    status = EXCLUDED.status,
+                    error_code = NULL,
+                    error_details = NULL
+                WHERE public_decryption_responses.decrypted_result IS NULL
+                RETURNING decryption_id
             )
-            VALUES ($1, $2, $3, $4, $5, $6) ON CONFLICT DO NOTHING",
+            UPDATE public_decryption_requests SET status = 'completed'
+            WHERE decryption_id IN (SELECT decryption_id FROM response)",
             response.decryption_id.as_le_slice(),
```
dvdplm: This query is a bit too clever for my taste. Had to ask the bots to even understand what it does!

Would you consider splitting it into two queries, i.e. first do:

```sql
INSERT INTO public_decryption_responses AS existing (
    decryption_id,
    decrypted_result,
    signature,
    extra_data,
    created_at,
    otlp_context,
    source,
    status
)
VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
ON CONFLICT (decryption_id) DO UPDATE
SET
    decrypted_result = EXCLUDED.decrypted_result,
    signature        = EXCLUDED.signature,
    extra_data       = EXCLUDED.extra_data,
    created_at       = EXCLUDED.created_at,
    otlp_context     = EXCLUDED.otlp_context,
    source           = EXCLUDED.source,
    status           = EXCLUDED.status,
    error_code       = NULL,
    error_details    = NULL
WHERE existing.decrypted_result IS NULL
RETURNING decryption_id
```

and then, in Rust, check if it returned a row, do:

```sql
UPDATE public_decryption_requests
SET status = 'completed'
WHERE decryption_id = $1
```

(wrap both in a transaction ofc)

If this is a hotpath and we want to cram out the best performance, I think it deserves a fair amount of comments for the not-so-good-SQL-reader. Sol suggested this single-query version (still pretty opaque to my eyes):

```sql
WITH written_response AS (
    INSERT INTO public_decryption_responses AS existing (
        decryption_id,
        decrypted_result,
        signature,
        extra_data,
        created_at,
        otlp_context,
        source,
        status
    )
    VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
    ON CONFLICT (decryption_id) DO UPDATE
    SET
        decrypted_result = EXCLUDED.decrypted_result,
        signature        = EXCLUDED.signature,
        extra_data       = EXCLUDED.extra_data,
        created_at       = EXCLUDED.created_at,
        otlp_context     = EXCLUDED.otlp_context,
        source           = EXCLUDED.source,
        status           = EXCLUDED.status,
        error_code       = NULL,
        error_details    = NULL
    WHERE existing.decrypted_result IS NULL
    RETURNING decryption_id
)
UPDATE public_decryption_requests AS request
SET status = 'completed'
FROM written_response
WHERE request.decryption_id = written_response.decryption_id
```

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
```diff
@@ -75,47 +79,92 @@ impl KmsResponsePublisher for DbKmsResponsePublisher {
...
         response: UserDecryptionResponse,
         created_at: DateTime<Utc>,
         otlp_ctx: PropagationContext,
+        source: RequestSource,
     ) -> anyhow::Result<PgQueryResult> {
         sqlx::query!(
-            "INSERT INTO user_decryption_responses(
-                decryption_id, user_decrypted_shares, signature, extra_data, created_at,
-                otlp_context
+            "WITH response AS (
```
dvdplm: Same comment as above: I'm not smart enough to read/debug this.

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
```diff
@@ -223,17 +272,128 @@ impl DbKmsResponsePublisher {
     }
 
     /// Sets the `status` field of the event to `pending` in the database.
-    pub async fn mark_event_as_pending(&self, event: ProtocolEvent) -> anyhow::Result<()> {
+    pub async fn mark_event_as_pending(&self, event: &ProtocolEvent) -> anyhow::Result<()> {
```
dvdplm, line 287: Hmm, this mutates the `event` in the DB, but the borrow here is immutable. Is that ok? Would `&mut` work here?

kms-connector/crates/kms-worker/tests/response_publisher.rs
```diff
@@ -291,3 +300,480 @@ async fn test_publish_epoch_result_response() -> anyhow::Result<()> {
...
+
+    info!("Checking the request row was marked as failed by the error publication...");
+    let request_status = sqlx::query_scalar::<_, OperationStatus>(
+        "SELECT status FROM public_decryption_requests WHERE decryption_id = $1",
+    )
+    .bind(rand_decryption_id.as_le_slice())
+    .fetch_one(test_instance.db())
+    .await?;
+    assert_eq!(request_status, OperationStatus::Failed);
+    info!("Error response successfully stored!");
```
dvdplm, line 406: nit: I'd make all these `debug!`

kms-connector/crates/kms-worker/tests/response_publisher.rs
```diff
@@ -291,3 +300,480 @@ async fn test_publish_epoch_result_response() -> anyhow::Result<()> {
...
+            .is_some()
+    );
+    assert!(row.try_get::<Option<Vec<u8>>, _>("signature")?.is_some());
+    assert_eq!(row.try_get::<Option<String>, _>("error_code")?, None);
+    assert_eq!(row.try_get::<Option<String>, _>("error_details")?, None);
+    assert_eq!(
+        row.try_get::<RequestSource, _>("source")?,
+        RequestSource::Http
+    );
+    // Terminal on insert: the tx-sender's `status = 'pending'` picking queries must never see it.
```
dvdplm: What does "terminal on insert" mean? I think this comment can be made clearer.

kms-connector/crates/tx-sender/tests/response_picker/source.rs
```diff
@@ -0,0 +1,80 @@
...
+    let responses = response_picker.pick_responses().await?;
+
+    info!("Checking only the onchain-sourced {response_type} was picked...");
+    assert_eq!(responses.len(), 1);
+    assert_eq!(responses[0].kind, onchain_response);
+    assert!(response_is_pending(db, response_type, http_response_id).await?);
+    info!("Data OK!");
+    Ok(())
+}
+
```
dvdplm, line 58: Do you think we should test scenarios where the same decryption request is submitted through both the GW and http? Or is that impossible?

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
eudelins-zama: Yep it's in the hotpath, will be done for every single decryption.

I agree the queries are non-trivial. I'll try to document them, as clearly as possible, but don't want to split in two queries to avoid the DB contention issue we saw in our bench (that lead to the context caching PR)

kms-connector/crates/kms-worker/tests/response_publisher.rs
eudelins-zama: This means we do not retry. Agreed

kms-connector/crates/tx-sender/tests/response_picker/source.rs
eudelins-zama: Definitely not possible in term of `decryption_id` (the content of the request could be the same though), otherwise we have a collision, which we assume is not possible in term of probability.

But I guess more tests would appear after the `endpoint` service is implemented (both integration and e2e tests)

kms-connector/crates/kms-worker/tests/response_publisher.rs
eudelins-zama: Test code, keeping the `info!` :slightly_smiling_face:

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
eudelins-zama: The mutation only happens in the DB row, the in-memory `ProtocolEvent` has no `status` field so there is nothing to mutate on the Rust side

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
eudelins-zama: Clarification attempt 7711016c0d690a8cfde2b13bbc3fa42599dc32ce

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
eudelins-zama: Clarification attempt 7711016c0d690a8cfde2b13bbc3fa42599dc32ce

kms-connector/crates/kms-worker/tests/response_publisher.rs
eudelins-zama: Is this better 405257d59017dbce3380f8dc6c0ddebeeb824e6e ?

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
dvdplm: oh, I see. But two borrows of the same `ProtocolEvent` could both call this method on different threads yeah? I guess they could do that before as well, by just cloning so not asking for a change here, it just sort of tripped up my mind to realize that the DB != in-memory.

eudelins-zama: Yes indeed

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
```diff
@@ -75,47 +79,104 @@ impl KmsResponsePublisher for DbKmsResponsePublisher {
...
         &self,
         response: PublicDecryptionResponse,
         created_at: DateTime<Utc>,
         otlp_ctx: PropagationContext,
+        source: RequestSource,
     ) -> anyhow::Result<PgQueryResult> {
+        // Single statement so the response row and the request status move atomically:
+        //   1. upsert the response, but only if no successful payload exists yet (a retry may
+        //      override a previous error row, but never a payload);
+        //   2. mark the request `completed` only if the response row actually written.
```
dvdplm:
```suggestion
        //   2. mark the request `completed` only if the response row is actually written.
```

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
```diff
@@ -75,47 +79,104 @@ impl KmsResponsePublisher for DbKmsResponsePublisher {
...
         &self,
         response: UserDecryptionResponse,
         created_at: DateTime<Utc>,
         otlp_ctx: PropagationContext,
+        source: RequestSource,
     ) -> anyhow::Result<PgQueryResult> {
+        // Single statement so the response row and the request status move atomically:
+        //   1. upsert the response, but only if no successful payload exists yet (a retry may
+        //      override a previous error row, never a payload);
+        //   2. mark the request `completed` only for the response row actually written.
```
dvdplm:
```suggestion
        //   2. mark the request `completed` only for the response row is actually written.
```

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
```diff
@@ -223,17 +284,135 @@ impl DbKmsResponsePublisher {
...
+        Ok(())
+    }
+}
+
+fn decryption_insert_status(source: RequestSource) -> OperationStatus {
+    match source {
+        // Onchain-sourced decryption are pending until the tx-sender publishes them on-chain.
+        RequestSource::OnChain => OperationStatus::Pending,
+        // HTTP-sourced decryption have no on-chain step so they are marked as `completed` on
+        // immediately on insert.
```
dvdplm:
```suggestion
        // HTTP-sourced decryption have no on-chain step so they are marked as `completed`
        // immediately on insert.
```

kms-connector/crates/kms-worker/src/core/kms_response_publisher.rs
dvdplm: Bah, better, I guess. Maybe readers just have to have some strong coffee and make an effort! Not sure what else you can do there, while keeping the single query.

PR #3704, opened 2026-09-02
fix(coprocessor): keep base history for diff coverage

Eikix:
## What

Fetch full history during the coverage checkout and remove the unauthenticated follow-up fetches.

## Why

`persist-credentials: false` makes the current `git fetch` calls fail before `diff-cover` runs.


Eikix:
@mergifyio queue

antoniupop (approved):

PanGan21 (approved):

PR #3731, opened 2026-09-02
feat(coprocessor): add GPU image builds

antoniupop:
Add flows to build images for coprocessor workers targeting GPU

### DIFF
--- a/.github/actions/gpu_setup/action.yml
+++ b/.github/actions/gpu_setup/action.yml
@@ -1,13 +1,16 @@
-name: Setup Cuda
-description: Setup Cuda on Hyperstack or GitHub instance
+name: Setup CUDA
+description: Set up CUDA and its host compiler on a GPU runner
 
 inputs:
   cuda-version:
-    description: Version of Cuda to use
+    description: Space-separated acceptable CUDA versions; installs the first if none are present
     required: true
-  github-instance:
-    description: Instance is hosted on GitHub
-    default: 'false'
+  gcc-version:
+    description: Version of GCC to use as the CUDA host compiler
+    required: true
+  cloud-provider:
+    description: Cloud provider hosting the GPU runner
+    required: true
 
 runs:
   using: "composite"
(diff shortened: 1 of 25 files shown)


Eikix (dismissed):
[AUTOMATED]

Reviewed all five changed files at 8499065521d2c6dc86babae2777114c2bdce12d0, against #3686. Approving with the following follow-up feedback. Existing comments and reviews were checked; these points are not duplicates.

### P2 — Explicit GPU selection is ignored

[compose.ts:331](https://github.com/zama-ai/fhevm/blob/8499065521d2c6dc86babae2777114c2bdce12d0/test-suite/fhevm/src/generate/compose.ts#L331) always adds a GPU device request with `count: all`.

- Worker A is pinned with `NVIDIA_VISIBLE_DEVICES=0`.
- Worker B is pinned with `NVIDIA_VISIBLE_DEVICES=1`.
- Both receive `count: all`, which causes Docker to override those pins.
- Both workers see all GPUs and can compete for memory instead of using separate cards.

A renderer reproduction confirmed that the per-instance environment values are retained but both workers receive the all-device request. The runtime consequence is confirmed in [Docker's device handling](https://github.com/moby/moby/blob/v28.0.0/daemon/nvidia_linux.go) and [NVIDIA's environment parser](https://github.com/NVIDIA/nvidia-container-toolkit/blob/v1.17.8/internal/config/image/builder.go); it was not reproduced on GPU hardware.

Could the device request use `device_ids` for an explicit card selection, omit `count` in that case, and include a test for that override?

### Nit — README still says there is no default command

[README:349](https://github.com/zama-ai/fhevm/blob/8499065521d2c6dc86babae2777114c2bdce12d0/coprocessor/README.md#L349) says the images set no `CMD`, but Dockerfile.gpu now provides one through the staged `worker` symlink. Please update that paragraph.

Validation: 379 CLI tests, TypeScript checks, and workflow lint passed. The extra two-instance renderer reproduction passed. Earlier H100 and L40 image builds succeeded, but neither was at this head; no GPU execution was performed at this head. No other confirmed new issues were found.

antoniupop:
@mergifyio queue

Eikix (approved):
[AUTOMATED]

Renewing approval after a focused re-review of 76c5296baf765338a4fe3dd328447f7cbadf14fd and its integration with current main (4f66757b).

Both previous findings are addressed:
- Explicit GPU pins now produce device_ids without count: all. The regression test checks the exact request for index and UUID/multiple-device selections.
- The README now correctly describes the default CMD.

Validation:
- PR head: 380 CLI tests passed; TypeScript and GPU workflow lint passed.
- GitHub merge preview fc1421b28: 394 CLI tests passed; TypeScript and GPU workflow lint passed.
- Independent focused review found no new actionable issue in the fix. No GPU hardware execution was performed.

Nit for follow-up: the current merge preview contains two identical 109-line "test-suite docker socket runtime" describe blocks in test-suite/fhevm/src/render-compose.test.ts. This appears when combining the branch with the squashed parent PR; keeping just one avoids running the same tests twice. This does not block approval.

antoniupop:
@mergifyio queue

PR #3800, opened 2026-09-03
fix(library-solidity): add decryption signature invalidation to IACL co…

eitjuh:
## Summary
- `IACL` in `host-contracts/lib/Impl.sol` is a generated file (from
  `library-solidity/codegen/src/templates/Impl.sol-template`), but the
  template was never updated when `invalidateDecryptionSignaturesBefore`
  and `decryptionSignatureInvalidatedBefore` were added to `ACL.sol`.
- Without this fix, the next `npm run codegen` run would regenerate
  `Impl.sol` and silently drop these two declarations from the `IACL`
  interface.
- Adds the matching declarations to the template and regenerates
  `lib/Impl.sol` to confirm it's byte-identical to the checked-in version.

## Test plan
- [x] `forge build` compiles cleanly
- [x] `forge test --match-path test/acl/acl.t.sol` — 58/58 pass
- [x] `npm run codegen` regenerates `lib/Impl.sol` unchanged, confirming template/output parity
- [x] `prettier --check` passes on the modified file


isaacdecoded (approved):

host-contracts/lib/Impl.sol
```diff
@@ -469,6 +469,19 @@ interface IACL {
         address contractAddress,
```
tawadaa, line 469: @eitjuh Could we please rebase so the E2E are on the latest commit hash?
Could we also bump `CORE_VERSION` to a stable version v0.14.1 this commit is hard to follow since kms is in a different repo

tawadaa: Based on the [test suite E2E](https://github.com/zama-ai/fhevm/actions/runs/33851337865/job/100959972014#step:20:51):

commit is `0c3dd7b` for host contract is correct but for other components it's still on an old commit `e7e7fec`, it should be on [the latest commit](https://github.com/zama-ai/fhevm/commits/release/0.14.x)

tawadaa (dismissed):

host-contracts/lib/Impl.sol
eitjuh: Rebase: done, branch is now on the latest release/0.14.x, so E2E resolves the other components against the current tip instead of e7e7fec.

CORE_VERSION v0.14.1: tried it, had to back it out. v0.14.1 rewrote the kms-gen-keys CLI, the old flags are gone:
```
error: unexpected argument '--public-storage' found
Usage: kms-gen-keys --config-file <CONFIG_FILE>
```

Our compose and generate/kms-core.ts still use the old form, so the stack never boots, all e2e jobs failed.

main fixed this in #3341 by probing --help and picking the right form. Backporting it touches ~5 files plus tests, which is a lot to bolt onto a 3-line codegen fix. I'd rather do it as its own PR, then the bump is a one-liner. OK?

eitjuh: See fix: https://github.com/zama-ai/fhevm/pull/3953

tawadaa: @dd23 which kms-core version we should use this branch to make sure we are on the latest stable?
We can go with v0.14.1, but I know that there were other fixes, so HEAD might have diverged already

dd23: kms-core v0.14.1 is the latest stable and is fine to use for everything that does not run KMS context operations like resharing. Anything around decryptions (that I think we're targeting here) is working fine in that version.
KMS context and resharing will be fixed and ready in v0.14.2.

All versions v0.14.x and newer have made changes to `kms-gen-keys` as mentioned in the [release notes](https://github.com/zama-ai/kms/releases/tag/v0.14.1), so we should adapt the test suite to this.

jatZama:
Not sure why would it need to merge this PR? this method is not used in FHE.sol, it seems it just makes the library size bigger to (almost) no benefit? Not a strong opinion, maybe we could add it if i missed something, in which case don't forget to update main branch as well and update the version in the library package.json. Also the scope of the pr is misleading, it should be `library-solidity`, not `host-contracts`.  I don't think we need to push this, especially not on the 0.14 release branch, unless we also port it to main, but i dont see why we need it even on main, to begin with?

eitjuh:
> Not sure why would it need to merge this PR? this method is not used in FHE.sol, it seems it just makes the library size bigger to (almost) no benefit? Not a strong opinion, maybe we could add it if i missed something, in which case don't forget to update main branch as well and update the version in the library package.json. Also the scope of the pr is misleading, it should be `library-solidity`, not `host-contracts`. I don't think we need to push this, especially not on the 0.14 release branch, unless we also port it to main, but i dont see why we need it even on main, to begin with?

@jatZama it's necessary because we need to be able to use this here: https://github.com/zama-ai/sdk/pull/727

I basically implemented this stopgap because this interface is not updated: https://github.com/zama-ai/sdk/pull/727/changes#diff-0c1a438e8469942414fc5b4f9a460f0dde137d7e2f5d2f910692cae0d06751d3

jatZama:
> > Not sure why would it need to merge this PR? this method is not used in FHE.sol, it seems it just makes the library size bigger to (almost) no benefit? Not a strong opinion, maybe we could add it if i missed something, in which case don't forget to update main branch as well and update the version in the library package.json. Also the scope of the pr is misleading, it should be `library-solidity`, not `host-contracts`. I don't think we need to push this, especially not on the 0.14 release branch, unless we also port it to main, but i dont see why we need it even on main, to begin with?
> 
> @jatZama it's necessary because we need to be able to use this here: [zama-ai/sdk#727](https://github.com/zama-ai/sdk/pull/727)
> 
> I basically implemented this stopgap because this interface is not updated: https://github.com/zama-ai/sdk/pull/727/changes#diff-0c1a438e8469942414fc5b4f9a460f0dde137d7e2f5d2f910692cae0d06751d3

ok if it simplifies stuff for the SDK, then I agree we could merge your PR, thanks for adding this context.

jatZama (approved):

eitjuh:
@mergifyio queue

eitjuh:
@Mergifyio dequeue

0xalexbel (approved):

tawadaa (approved):

PR #3802, opened 2026-09-03
feat(coprocessor): e2e consensus coverage

antoniupop:


Eikix (commented):
## Automated review (Codex)

Reviewed head `fb8ed74752a096adb11d48381413d0c3bf926a54`. Checked existing comments and reviews before posting; these findings are not duplicates.

1. **[P1] Cutover monitoring can stop permanently after a database outage.**
   The shared listener now exits after 31 consecutive receive errors. Host-listener restarts it, but SNS, gateway-listener and transaction-sender still spawn it once and only log its error. After an outage exhausts this limit, those services can keep running without observing a later cutover—for example, the new stack's SNS uploader can remain parked. This concerns the new exit path introduced here, not all pre-existing listener recovery issues. Could transient errors keep retrying, or could every caller handle termination by restarting the listener or stopping the service? [Changed code](https://github.com/zama-ai/fhevm/blob/fb8ed74752a096adb11d48381413d0c3bf926a54/coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs#L214-L218).

2. **[P2] The worker guard rejects legitimate blue-green setups.**
   On a Linux host with local Docker, the process scan sees both blue and green workers, but the allowed PID list includes only blue containers. Active green workers are therefore reported as stray processes and startup fails. Could the guard include all containers belonging to the scenario? A matching process name alone also does not establish that a worker uses the same queue. [Changed code](https://github.com/zama-ai/fhevm/blob/fb8ed74752a096adb11d48381413d0c3bf926a54/test-suite/fhevm/src/flow/readiness.ts#L290-L304).

3. **[P2] The consensus probe can pass without comparing SNS digests.**
   Missing digests are filtered out, and fewer than two means the comparison is skipped. The crash-retry test accepts this through `snsDigestsChecked || true`. The fallback assumes submission removes the digest, but the sender deletes ciphertext payloads from `ciphertexts128`, not the durable `ciphertext_digest.ciphertext128` values this query reads. Could the probe require a digest from every compared operator and fail if one never arrives, instead of treating incomplete evidence as agreement? [Changed code](https://github.com/zama-ai/fhevm/blob/fb8ed74752a096adb11d48381413d0c3bf926a54/test-suite/e2e/test/consensus/probe.ts#L201-L206).

4. **[P2] A failed database query counts as “keys ready.”**
   The query allows command failure. Empty output parses as `NaN`; `NaN < 1` is false, so the operator is not marked pending and readiness succeeds. Could this require a successful command and a valid positive count, retrying otherwise? [Changed code](https://github.com/zama-ai/fhevm/blob/fb8ed74752a096adb11d48381413d0c3bf926a54/test-suite/fhevm/src/flow/readiness.ts#L750-L758).

Validation: 416 CLI tests, TypeScript check and workflow lint passed. Targeted mocked reproductions confirmed findings 2–4; they do not replace live stack testing. No local Rust, full-stack or GPU tests were run. The inspected CI worker failure reports runner disk exhaustion, which leaves that validation incomplete rather than establishing a code defect.

Eikix (commented):
## Automated review (Codex)

Reviewed `a91e4e90937999e2806fbb3f94f0dc1d2e351457`, with separate correctness and maintainability passes. Checked existing comments and reviews: the four findings in the earlier automated review are fixed; the points below are not duplicates.

The failure-testing work belongs in this repository. The size alone is not a reason to reject it. The main concerns are two CI integration failures, gaps in what the assertions prove, and repeated execution/cleanup logic.

### Confirmed findings

1. **[P2] The new CLI test breaks the existing CLI-only install/check path.** `multichain-deployer.test.ts` imports an E2E helper that requires `ethers`, which is not a CLI dependency. The existing E2E workflow installs only CLI dependencies before running its TypeScript check. Reproduced with that dependency layout: `TS2307: Cannot find module 'ethers'`. Installing both packages hides the failure. Could this test live beside its helper in the E2E package? [Test](https://github.com/zama-ai/fhevm/blob/a91e4e90937999e2806fbb3f94f0dc1d2e351457/test-suite/fhevm/src/consensus/multichain-deployer.test.ts#L3-L5).

2. **[P2] The stackless oracle command fails on Node 22.23.2 before running tests.** Reproduced the committed Mocha/ts-node command: `ReferenceError: __dirname is not defined in ES module scope` in `abortRecovery.test.ts`. Disabling native TypeScript stripping makes it run. Could the runner explicitly select a working TypeScript loading mode for the Node 22 environment used by CI? [Runner](https://github.com/zama-ai/fhevm/blob/a91e4e90937999e2806fbb3f94f0dc1d2e351457/test-suite/fhevm/scripts/run-stackless-cases.sh#L45-L46).

3. **[P2] Quorum evidence is not tied to the database bytes being compared.** `assertQuorumOutcome` checks the handle, authorized senders and threshold, but not the event's key ID or ciphertext/SNS digests against the checked database evidence. An event for the right handle but wrong digests can satisfy this assertion. Could this reuse the stronger event-binding checks already used by the materialization tests? This is a test coverage gap, not evidence that production publishes incorrect bytes. [Assertion](https://github.com/zama-ai/fhevm/blob/a91e4e90937999e2806fbb3f94f0dc1d2e351457/test-suite/e2e/test/consensus/probe.ts#L246-L301).

4. **[P2] The comparator checks only the first alias row's operation.** Reproduced with operation lists `[8, 8]` and `[8, 9]` across two operators: comparison succeeds when the other evidence matches, despite reporting operation as compared. Could every alias row's operation be validated? [Comparator](https://github.com/zama-ai/fhevm/blob/a91e4e90937999e2806fbb3f94f0dc1d2e351457/test-suite/e2e/test/consensus/comparator.ts#L180-L200).

5. **[P2] Reported GPU scheduling flags can differ from the worker's actual flags.** The reporting helper gives an explicit GPU override priority over the environment file. systemd gives `EnvironmentFile` priority over the `Environment` values supplied through `--setenv`. With conflicting boolean settings, the report therefore describes a different configuration from the running worker. The current workflow's numeric overrides do not themselves trigger this conflict. Could startup and reporting use one resolved configuration? [Helper](https://github.com/zama-ai/fhevm/blob/a91e4e90937999e2806fbb3f94f0dc1d2e351457/test-suite/fhevm/scripts/gpu-consensus-workers.sh#L1176-L1184).

### Focused cleanup suggestions

- **Define case execution once.** The inventory already supplies runners, but the workflow separately dispatches by case ID, and the planner separately assigns shards by case ID. Current routing is consistent; the concern is keeping these copies aligned. Could the existing manifest hold structured arguments, setup and shard information, with CI executing the resolved plan?
- **Keep common recovery ordering in the existing lifecycle helpers.** Several runners repeat cancellation, mutation recovery, service restoration and result finalization. Could they share this sequence while retaining fault-specific cleanup where ordering matters?
- **Test callable helpers instead of slicing source text.** Some tests extract functions using comment/function-name strings. Moving a comment can then break a test without changing behavior. Exporting or sourcing the existing helpers would make these checks less brittle.

These suggestions do not call for a new framework or a size-only rewrite.

### Validation and limits

Linux CLI: 723 passed, 13 skipped; TypeScript passed with both packages installed. The CLI-only dependency failure was reproduced separately. All 27 shell scripts passed Linux Bash syntax checks. With the Node loading workaround, the oracle tests passed (139 passed, 8 SQL tests pending), but the full stackless run still had one image-provenance test timeout. That test passed separately; the timeout cause remains unresolved.

At this head, the inspected Rust and GPU CI checks passed; lint and typo checks failed. A complete consensus CPU/GPU campaign was not verified. No production corruption is claimed by this review.

Eikix:
Automated review of head `f98c825564`, covering the full PR.

The overall strategy looks sound: check operator agreement, exercise forks/reorgs and different scheduling, and require valid evidence and successful cleanup before reporting a pass.

Four points remain:

1. **Node 22 launcher:** the standalone comparator command fails with `__dirname is not defined in ES module scope`. Reproduced on Node 22.23.2; disabling native TypeScript stripping lets all 116 contracts pass.
2. **Fork/reorg quorum checks:** these verify database agreement and enough authorized Gateway votes separately, but do not check that the event’s key and digests match the database evidence. Please reuse the binding check already used by materialization tests. This is a test coverage gap, not a demonstrated production bug.
3. **Alias comparison:** the new comparator reads the operation from only the first computation row. A different operation on a later alias row is accepted. The existing materialization helper rejects this inconsistency.
4. **GPU settings:** the report treats explicit overrides as stronger than the environment file, but systemd applies the opposite precedence. Conflicting values can therefore make the report describe settings the worker did not receive.

Also worth adjusting: make expensive CPU/GPU jobs wait for the cheap harness checks, so a launcher failure does not waste provisioned resources.

These should be focused fixes, not a redesign. Reuse existing checks where possible; additional regression tests only where they add useful coverage. Then validate the final head, including the cold-start build/receipt path.

Review limitation: local checks and focused negative tests were run, but no live CPU/GPU consensus campaign was launched.

antoniupop:
> Automated review of head `f98c825564`, covering the full PR [...]

1/3/4 yes makes sense to address.
2 is overstated - shouldn't be in e2e, already covered in unit tests.

Eikix (approved):
LGTM

antoniupop:
@mergifyio queue

PR #3803, opened 2026-09-03
feat(coprocessor): add GPU e2e CI flows

antoniupop:


Eikix (commented):
[AUTOMATED]

## Automated review (Codex)

Reviewed head `b27ab28caf9373051632530c768d192274098a1d`. Checked existing comments and reviews; this finding has not already been posted.

**[P1] The two new inputs make the shared E2E workflow invalid.**

Adding `runner` and `gpu-host` to `shared_inputs` raises the `workflow_dispatch` input count from 25 to 27. GitHub allows at most 25. This affects the shared E2E workflow, not just the new GPU flow. Both local `actionlint` and this head's lint CI report the same error.

Could these two caller-only inputs be declared under `workflow_call` without adding them to `workflow_dispatch`, or could the manual input set otherwise stay within the limit?

[Changed lines](https://github.com/zama-ai/fhevm/blob/b27ab28caf9373051632530c768d192274098a1d/.github/workflows/test-suite-e2e-tests.yml#L16-L29) · [CI evidence](https://github.com/zama-ai/fhevm/actions/runs/34115155679) · [GitHub input limit](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs)

**Stack integration:** this branch still carries an older version of its parent. Comparing the current heads of #3802 and #3803 produces 46 conflicted files. The four GPU CI commits after `206876ca7` account for six files (+699/−1); the larger displayed diff also includes inherited work. Could the branch be updated onto the current parent while preserving the fixes already reviewed there? The combined result needs checking after that update; the conflicts themselves do not establish that code will be lost.

**Validation:** the 17 new build-filter tests pass. The input-limit finding was independently confirmed. No live GPU E2E campaign was run, so this review does not establish that the full orchestration succeeds. The unchanged findings on #3802 remain tracked in its existing review and are not repeated here.

Eikix (commented):
[AUTOMATED]

## Automated review (Codex)

Reviewed `4e68d475adb0ed5abfe4d99c206ad7149625b6f0`. This updates the existing review with a correctness, maintainability and lean-design pass, including two independent reviewers. The earlier 27-input limit and stack conflicts are resolved. Finding 2 was previously reported on #3802 and moved here after the split.

### Confirmed findings

1. **[P1] The GPU workflow passes four inputs the shared E2E workflow no longer accepts.** It passes separate connector image versions, but the callee now accepts `connector-versions` JSON. Local `actionlint` and this head's lint CI reject the call. Could this use the same JSON input as the CPU orchestrator? [Changed lines](https://github.com/zama-ai/fhevm/blob/4e68d475adb0ed5abfe4d99c206ad7149625b6f0/.github/workflows/test-suite-orchestrate-e2e-gpu-tests.yml#L361-L364) · [CI evidence](https://github.com/zama-ai/fhevm/actions/runs/35006984952).

2. **[P2] The new multichain test breaks the CLI-only install used by CI.** The CLI test imports an E2E helper that needs `ethers`, which the CLI package does not install. After `bun install --frozen-lockfile` in `test-suite/fhevm`, typechecking fails with `TS2307: Cannot find module 'ethers'`; the test also fails to load. Could this test live in the E2E package and run through its test flow? Installing both packages locally hides the failure. [Test imports](https://github.com/zama-ai/fhevm/blob/4e68d475adb0ed5abfe4d99c206ad7149625b6f0/test-suite/fhevm/src/consensus/multichain-deployer.test.ts#L3-L5).

3. **[P2] The GPU image-selection path omits the connector endpoint.** Unlike the CPU orchestrator, it does not check `endpoint_build_result` or forward an endpoint version. Running the actual selection script with the listed builds successful and the endpoint build failed still succeeds, without an endpoint tag. Once finding 1 is fixed, this can test the baseline endpoint instead of the branch's endpoint, or omit it when the baseline has none. Could the endpoint be included in build-result validation and the connector JSON? Converting only the existing four inputs to JSON leaves this gap. [Image mapping](https://github.com/zama-ai/fhevm/blob/4e68d475adb0ed5abfe4d99c206ad7149625b6f0/.github/workflows/test-suite-orchestrate-e2e-gpu-tests.yml#L269-L273).

### Design and maintainability

The overall approach fits the repository: reuse the existing E2E workflow, add small `runner` and `gpu-host` inputs, return the exact tag from the GPU build, and keep GPU container preparation separate. The size alone is not a reason to reject the change, and it does not add a new public CLI command.

The main simplification worth making is **one shared CPU/GPU image-selection and validation implementation**, with GPU worker tags supplied as overrides. The duplicated mapping has already drifted, as findings 1 and 3 show. Could both workflows share this logic and the existing image-tag validation helpers? This can stay a narrow change; it does not need a general CI framework or a rewrite of provisioning.

The filter test's hand-written Cargo/YAML parsing is also a robustness concern. No additional current failure was established from it, so this is a non-blocking improvement, not a fourth correctness finding.

### Verification

22 focused tests passed. CLI typechecking and multichain test loading failed as described; `actionlint` reported the four invalid inputs. The endpoint omission was reproduced by executing the selection script with synthetic build results. `git diff --check` passed. No live GPU E2E campaign or H100 kernel validation was performed.

The direction is sound, but the three confirmed integration issues remain at this head.

Eikix (approved):
LGTM

antoniupop:
@mergifyio queue

PR #3804, opened 2026-09-04
chore(test-suite): hardening e2e erc-1271

maksymsur:
A revert carrying no reason was treated as a definitive signature rejection - 400 on the first attempt, no retry. Only a reasoned revert (GS0xx, "Hash not approved") proves the wallet rejected anything; an empty one is also what an under-gassed call produces, and what the nested-Safe e2e positives hit on zws-dev before passing on an unchanged re-run.

- `Erc1271Error::EmptyRevert` splits a reasonless revert from a reasoned one. The relayer retries it alongside transport failures and reports a rejection only if it survives every attempt; the connector treats it as recoverable.
- Gas cap 100_000 -> 250_000 in the relayer configs, the KMS connector default and the js-sdk, which cap the same call: a nested Safe costs ~91k, too little headroom for deeper nesting. The verifier tries 100_000 first and logs when it has to fall back, so the raise stays observable.
- e2e: nested positives cross-check the signature on-chain before submitting, and assertion messages carry the requestId that keys into the relayer log.

### DIFF
--- a/kms-connector/crates/kms-worker/src/core/config.rs
+++ b/kms-connector/crates/kms-worker/src/core/config.rs
@@ -216,7 +216,10 @@ fn default_s3_connect_timeout() -> Duration {
 }
 
 fn default_erc1271_gas_limit() -> u64 {
-    100_000
+    // Bounds a read-only STATICCALL. A nested Safe (a Safe owning a Safe) costs ~91k gas;
+    // the rest is headroom for deeper nesting, and `ERC1271_GAS_WARN_THRESHOLD` logs a call
+    // that needs more than 100k. Kept equal to the relayer config and the js-sdk.
+    250_000
 }
 
 fn default_ct_attestation_verifier_enabled() -> bool {
--- a/kms-connector/crates/kms-worker/src/core/config.rs
+++ b/kms-connector/crates/kms-worker/src/core/config.rs
@@ -216,9 +216,6 @@ fn default_s3_connect_timeout() -> Duration {
 }
 
 fn default_erc1271_gas_limit() -> u64 {
-    // Bounds a read-only STATICCALL. A nested Safe (a Safe owning a Safe) costs ~91k gas;
-    // the rest is headroom for deeper nesting, and `ERC1271_GAS_WARN_THRESHOLD` logs a call
-    // that needs more than 100k. Kept equal to the relayer config and the js-sdk.
     250_000
 }
 
--- a/kms-connector/crates/kms-worker/src/core/event_processor/error.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/error.rs
@@ -38,6 +38,7 @@ impl From<Erc1271Error> for ProcessingError {
                 Self::Irrecoverable(anyhow::Error::new(err))
             }
             Erc1271Error::Transport(_)
+            | Erc1271Error::EmptyRevert(_)
             | Erc1271Error::WrongMagic(..)
             | Erc1271Error::Rejected(..) => Self::Recoverable(anyhow::Error::new(err)),
         }
@@ -136,7 +137,9 @@ impl RequestCheckError {
 impl From<Erc1271Error> for RequestCheckError {
     fn from(err: Erc1271Error) -> Self {
         let kind = match &err {
-            Erc1271Error::Transport(_) => RequestCheckKind::Network,
+            // A reasonless revert does not prove the signature is bad — an under-gassed
+            // call looks identical — so report it as an environment problem.
+            Erc1271Error::Transport(_) | Erc1271Error::EmptyRevert(_) => RequestCheckKind::Network,
             Erc1271Error::EmptySigOnEoa(_)
             | Erc1271Error::EoaMismatchNoCode(_)
             | Erc1271Error::Rejected(..)
--- a/kms-connector/crates/kms-worker/src/core/event_processor/error.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/error.rs
@@ -137,8 +137,6 @@ impl RequestCheckError {
 impl From<Erc1271Error> for RequestCheckError {
     fn from(err: Erc1271Error) -> Self {
         let kind = match &err {
-            // A reasonless revert does not prove the signature is bad — an under-gassed
-            // call looks identical — so report it as an environment problem.
             Erc1271Error::Transport(_) | Erc1271Error::EmptyRevert(_) => RequestCheckKind::Network,
             Erc1271Error::EmptySigOnEoa(_)
             | Erc1271Error::EoaMismatchNoCode(_)
--- a/kms-connector/crates/kms-worker/src/core/event_processor/error.rs
+++ b/kms-connector/crates/kms-worker/src/core/event_processor/error.rs
@@ -137,8 +137,9 @@ impl RequestCheckError {
 impl From<Erc1271Error> for RequestCheckError {
     fn from(err: Erc1271Error) -> Self {
         let kind = match &err {
-            Erc1271Error::Transport(_) | Erc1271Error::EmptyRevert(_) => RequestCheckKind::Network,
-            Erc1271Error::EmptySigOnEoa(_)
+            Erc1271Error::Transport(_) => RequestCheckKind::Network,
+            Erc1271Error::EmptyRevert(_)
+            | Erc1271Error::EmptySigOnEoa(_)
             | Erc1271Error::EoaMismatchNoCode(_)
             | Erc1271Error::Rejected(..)
             | Erc1271Error::WrongMagic(..) => RequestCheckKind::Signature,
--- a/relayer/src/host/signature_prechecker.rs
+++ b/relayer/src/host/signature_prechecker.rs
@@ -231,6 +231,14 @@ mod tests {
         U256::from_be_bytes(bytes)
     }
 
+    /// A node's reasonless-revert response: "execution reverted" with empty `data`.
+    fn empty_revert() -> alloy::rpc::json_rpc::ErrorPayload {
+        alloy::rpc::json_rpc::ErrorPayload::internal_error_with_message_and_obj(
+            std::borrow::Cow::Borrowed("execution reverted"),
+            serde_json::value::RawValue::from_string("\"0x\"".to_string()).unwrap(),
+        )
+    }
+
     fn checker(asserter: Asserter) -> UserDecryptSignaturePreChecker {
         let mut providers = HashMap::new();
         providers.insert(
@@ -242,7 +250,7 @@ mod tests {
         UserDecryptSignaturePreChecker {
             providers,
             decryption_contract: Address::from([0xCA; 20]),
-            erc1271_gas_limit: 100_000,
+            erc1271_gas_limit: ERC1271_GAS_WARN_THRESHOLD,
             retry: RetrySettings {
                 max_attempts: 3,
                 retry_interval_ms: 0,
@@ -282,6 +290,46 @@ mod tests {
             .await
             .unwrap_err();
         assert!(matches!(err, SigPreCheckError::HostCallFailed(_)));
+    }
+
+    /// Cap above the warn threshold, so a first attempt at the threshold is followed by one
+    /// at the full budget.
+    fn checker_with_headroom(asserter: Asserter) -> UserDecryptSignaturePreChecker {
+        UserDecryptSignaturePreChecker {
+            erc1271_gas_limit: 250_000,
+            ..checker(asserter)
+        }
+    }
+
+    #[tokio::test]
+    async fn intermittent_empty_revert_is_retried() {
+        let asserter = Asserter::new();
+        // Attempt 1 exhausts both budgets; attempt 2 succeeds at the threshold.
+        asserter.push_failure(empty_revert());
+        asserter.push_failure(empty_revert());
+        let mut returndata = [0u8; 32];
+        returndata[..4].copy_from_slice(&[0x16, 0x26, 0xba, 0x7e]);
+        asserter.push_success(&returndata);
+
+        checker_with_headroom(asserter)
+            .verify(&unified_request(Bytes::from(vec![0x11; 65])))
+            .await
+            .unwrap();
+    }
+
+    #[tokio::test]
+    async fn persistent_empty_revert_is_invalid_not_server_error() {
+        let asserter = Asserter::new();
+        // Three attempts, each spending both budgets: no longer plausibly transient.
+        for _ in 0..6 {
+            asserter.push_failure(empty_revert());
+        }
+
+        let err = checker_with_headroom(asserter)
+            .verify(&unified_request(Bytes::from(vec![0x11; 65])))
+            .await
+            .unwrap_err();
+        assert!(matches!(err, SigPreCheckError::Invalid { .. }), "{err:?}");
     }
 
     #[tokio::test]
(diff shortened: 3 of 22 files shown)


relayer/src/host/signature_prechecker.rs
```diff
@@ -115,36 +117,50 @@ impl UserDecryptSignaturePreChecker {
...
+                        warn_above = ERC1271_GAS_WARN_THRESHOLD,
+                        gas_limit = self.erc1271_gas_limit,
+                        "ERC-1271 verification needed more gas than expected"
+                    );
+                    return Ok(());
+                }
+                Ok(_) => return Ok(()),
+                // Retry what is not proof of a bad signature: transport failures, and
+                // reasonless reverts, which an under-gassed call produces too.
+                Err(e @ (Erc1271Error::Transport(_) | Erc1271Error::EmptyRevert(_))) => {
```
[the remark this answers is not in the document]
maksymsur: Good catch, addressed in https://github.com/zama-ai/fhevm/pull/3804/commits/14e42e799973cb797347e543add9f6064cae6dfd

shared/user-decryption-signature/src/lib.rs
```diff
@@ -183,21 +221,37 @@ async fn check_erc1271_signature<P: Provider>(
...
+    let warn_above = gas.warn_above.min(gas.limit);
+    match check_erc1271_signature(provider, claimed_signer, digest, signature, warn_above).await {
+        Ok(()) => Ok(Erc1271Accepted::WithinWarnThreshold),
+        // Exhausting the budget is indistinguishable from a wallet that reverts without a
+        // reason, so the only way to tell them apart is to offer the full budget and see.
+        Err(Erc1271Error::EmptyRevert(_)) if warn_above < gas.limit => {
+            check_erc1271_signature(provider, claimed_signer, digest, signature, gas.limit)
+                .await
+                .map(|()| Erc1271Accepted::AboveWarnThreshold)
+        }
```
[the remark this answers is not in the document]
maksymsur: Good catch, addressed in https://github.com/zama-ai/fhevm/pull/3804/commits/14e42e799973cb797347e543add9f6064cae6dfd

shared/user-decryption-signature/src/lib.rs
```diff
@@ -183,21 +221,44 @@ async fn check_erc1271_signature<P: Provider>(
...
+    let warn_above = gas.warn_above.min(gas.limit);
+    match check_erc1271_signature(provider, claimed_signer, digest, signature, warn_above).await {
+        Ok(()) => Ok(Erc1271Accepted::WithinWarnThreshold),
+        // Neither outcome proves the signature is bad, and an exhausted budget looks like
+        // either one depending on the node: a reasonless revert, or an RPC error carrying no
+        // revert data. Offer the full budget before deciding.
+        Err(first @ (Erc1271Error::EmptyRevert(_) | Erc1271Error::Transport(_)))
+            if warn_above < gas.limit =>
+        {
+            check_erc1271_signature(provider, claimed_signer, digest, signature, gas.limit)
```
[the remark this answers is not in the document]
maksymsur: addressed in https://github.com/zama-ai/fhevm/pull/3804/commits/31d76d022d9db5171b26c4a352778297f71a6397

shared/user-decryption-signature/src/lib.rs
```diff
@@ -183,21 +221,44 @@ async fn check_erc1271_signature<P: Provider>(
...
 
-    check_erc1271_signature(provider, claimed_signer, digest, signature, gas_limit).await
+    let warn_above = gas.warn_above.min(gas.limit);
+    match check_erc1271_signature(provider, claimed_signer, digest, signature, warn_above).await {
+        Ok(()) => Ok(Erc1271Accepted::WithinWarnThreshold),
+        // Neither outcome proves the signature is bad, and an exhausted budget looks like
+        // either one depending on the node: a reasonless revert, or an RPC error carrying no
+        // revert data. Offer the full budget before deciding.
+        Err(first @ (Erc1271Error::EmptyRevert(_) | Erc1271Error::Transport(_)))
+            if warn_above < gas.limit =>
```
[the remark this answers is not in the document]
maksymsur: fixed with design change https://github.com/zama-ai/fhevm/pull/3804/commits/627772366cc4eec65596d0da6c613869fa93ae82

shared/user-decryption-signature/src/lib.rs
```diff
@@ -183,21 +221,44 @@ async fn check_erc1271_signature<P: Provider>(
...
+        {
+            check_erc1271_signature(provider, claimed_signer, digest, signature, gas.limit)
+                .await
+                .map(|()| match first {
+                    // Only a revert is gas-shaped; a transport error says nothing about cost.
+                    Erc1271Error::EmptyRevert(_) => Erc1271Accepted::AboveWarnThreshold,
+                    _ => Erc1271Accepted::WithinWarnThreshold,
```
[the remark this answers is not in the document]
maksymsur: fixed with design change https://github.com/zama-ai/fhevm/pull/3804/commits/627772366cc4eec65596d0da6c613869fa93ae82

sdk/js-sdk/src/core/utils-p/decrypt/verifyErc1271UserDecrypt-p.ts
```diff
@@ -21,10 +21,11 @@ export const ERC1271_MAGIC_VALUE = '0x1626ba7e' as const;
...
 /**
  * Gas cap for the `isValidSignature` STATICCALL. Matches the relayer / KMS
- * default (`erc1271_gas_limit = 100000`) so all three layers bound the wallet's
- * verification the same way.
+ * default (`erc1271_gas_limit = 250000`) so all three layers bound the wallet's
+ * verification identically. A nested Safe (a Safe owning a Safe) costs ~91k
+ * gas; the rest is headroom for deeper nesting.
  */
-export const ERC1271_GAS_LIMIT = 100_000n;
+export const ERC1271_GAS_LIMIT = 250_000n;
```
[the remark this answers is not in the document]
maksymsur: indeed... accepting

manoranjith (approved):
Relayer part looks good to me.

maksymsur:
> Relayer part looks good to me.

Thnx @manoranjith ! Let's wait for @eudelins-zama review

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -136,7 +137,9 @@ impl RequestCheckError {
 impl From<Erc1271Error> for RequestCheckError {
     fn from(err: Erc1271Error) -> Self {
         let kind = match &err {
-            Erc1271Error::Transport(_) => RequestCheckKind::Network,
+            // A reasonless revert does not prove the signature is bad — an under-gassed
+            // call looks identical — so report it as an environment problem.
+            Erc1271Error::Transport(_) | Erc1271Error::EmptyRevert(_) => RequestCheckKind::Network,
```
eudelins-zama:
```suggestion
            Erc1271Error::Transport(_) | Erc1271Error::EmptyRevert(_) => RequestCheckKind::Network,
```

kms-connector/crates/kms-worker/src/core/config.rs
```diff
@@ -216,7 +216,10 @@ fn default_s3_connect_timeout() -> Duration {
 }
 
 fn default_erc1271_gas_limit() -> u64 {
-    100_000
+    // Bounds a read-only STATICCALL. A nested Safe (a Safe owning a Safe) costs ~91k gas;
+    // the rest is headroom for deeper nesting, and `ERC1271_GAS_WARN_THRESHOLD` logs a call
+    // that needs more than 100k. Kept equal to the relayer config and the js-sdk.
+    250_000
```
eudelins-zama:
```suggestion
    250_000
```

This should be explained in the `kms-worker.toml` (already the case AFAICT) and not duplicated here

shared/user-decryption-signature/src/lib.rs
```diff
@@ -62,6 +101,11 @@ pub enum Erc1271Error {
         "ERC-1271 isValidSignature reverted or returned malformed data for userAddress {0}: {1}"
     )]
     Rejected(Address, String),
+    #[error(
+        "ERC-1271 isValidSignature reverted with no reason for userAddress {0}; \
+         typically an under-gassed call, or a wallet that reverts without a reason"
+    )]
+    EmptyRevert(Address),
```
eudelins-zama: Why do we need to distinguish `EmptyRevert` from `Rejected`?

Pure logging info or functionally we need to handle them differently? AFAIU, an empty revert would be considered as rejected.

shared/user-decryption-signature/src/lib.rs
```diff
@@ -136,11 +181,17 @@ async fn check_erc1271_signature<P: Provider>(
...
         RpcError::ErrorResp(e) => {
             if let Some(revert_data) = e.as_revert_data() {
-                // A revert is interpreted as an invalid signature
-                Erc1271Error::Rejected(
-                    addr,
-                    format!("isValidSignature call reverted: {revert_data}"),
-                )
+                // Nodes answer an under-gassed call with `execution reverted` and
+                // `data: "0x"`, which `as_revert_data` still reports as revert data. Only a
+                // reason makes the rejection deterministic.
```
eudelins-zama: "0x" could mean the contract does not exist at all, not necessarily an under-gassed call...

eudelins-zama: The only thing we could be sure here, is that it is an `EmptyRevert`, so the error classification is right, but the comment can be misleading AFAIU

shared/user-decryption-signature/src/lib.rs
```diff
@@ -183,21 +234,39 @@ async fn check_erc1271_signature<P: Provider>(
...
+/// retried with the full `limit`, whose result is authoritative. The configured cap is
+/// therefore always offered before a rejection, and a wallet that only succeeds on the retry
+/// is reported as [`Erc1271Accepted::AboveWarnThreshold`]. The probe costs one extra call on
+/// every rejection; set `warn_above >= limit` to disable it.
 pub async fn verify_signature<P: Provider>(
     provider: &P,
     claimed_signer: Address,
     digest: B256,
     signature: &[u8],
-    gas_limit: u64,
```
eudelins-zama, line 191: Since for now, we're always using the hardcoded `ERC1271_GAS_WARN_THRESHOLD` (via `capped_at`), I would suggest we keep the code as it was before here and remove this `Erc1271GasBudget` struct.

shared/user-decryption-signature/src/lib.rs
```diff
@@ -183,21 +234,39 @@ async fn check_erc1271_signature<P: Provider>(
...
-) -> Result<(), Erc1271Error> {
+    gas: Erc1271GasBudget,
+) -> Result<Erc1271Accepted, Erc1271Error> {
```
eudelins-zama: I would also remove this `Erc1271Accepted` struct and revert to how it was.

I would simply `warn!` directly in the `Err(_) if warn_above < gas.limit` arm, so the caller has nothing to do on its own.

shared/user-decryption-signature/src/lib.rs
```diff
@@ -183,21 +234,39 @@ async fn check_erc1271_signature<P: Provider>(
...
+        // reasonless one — so no failure at the probe budget is evidence about the signature.
+        // Offer the full budget and let its result decide.
+        Err(_) if warn_above < gas.limit => {
+            check_erc1271_signature(provider, claimed_signer, digest, signature, gas.limit)
+                .await
+                .map(|()| Erc1271Accepted::AboveWarnThreshold)
+        }
+        Err(e) => Err(e),
+    }
 }
```
eudelins-zama: A bit concerned that `verify_signature` now calls `check_erc1271_signature` at 100k first and on any failure calls again at the configured cap. That doubles the `eth_call` on every rejection, including `EoaMismatchNoCode` where extra gas can't change the answer (hence the mock in decryption.rs being pushed twice). In the relayer pre-checker this is user-facing: a bad signature costs 2x host RPC, and an empty revert up to 6 calls with sleeps.

What it buys is one `warn!` line saying a wallet needed more than 100k. `Erc1271GasBudget` and `Erc1271Accepted` exist only to carry that...

Would keeping `EmptyRevert`, and the 250k default be enough, without duplicating this call?

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
maksymsur: done

kms-connector/crates/kms-worker/src/core/config.rs
maksymsur: Agreed, removed - kms-worker.toml already carries the explanation.

shared/user-decryption-signature/src/lib.rs
maksymsur: Functional, not logging: the relayer retries EmptyRevert alongside transport errors and only rejects if it survives every attempt, whereas Rejected is final on the first response.

Actually - that's the fix for this PR's failure - an empty-data revert on zws-dev that passed on an unchanged re-run.

shared/user-decryption-signature/src/lib.rs
maksymsur: Hm.. you're right, reworded the comment and the error string for clarity.

shared/user-decryption-signature/src/lib.rs
maksymsur: Agree - back to `gas_limit: u64`.

shared/user-decryption-signature/src/lib.rs
maksymsur: Agree - `verify_signature` returns `Result<(), Erc1271Error>` again, and with the probe now gone (below) there's no warning left to carry.

shared/user-decryption-signature/src/lib.rs
maksymsur: Agreed, removed the probe .. `verify_signature` makes a single call at the new configured cap. My initial motivation was somehow to track pricy transactions... but that can be achieved by other means + EmptyRevert and the 250k default carry the fix on their own; gas visibility stays in the e2e cross-check, which prices the call on-chain. So Ok... let's back to the initial sate. Thanks for bringing that up!

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
```diff
@@ -136,7 +137,7 @@ impl RequestCheckError {
 impl From<Erc1271Error> for RequestCheckError {
     fn from(err: Erc1271Error) -> Self {
         let kind = match &err {
-            Erc1271Error::Transport(_) => RequestCheckKind::Network,
+            Erc1271Error::Transport(_) | Erc1271Error::EmptyRevert(_) => RequestCheckKind::Network,
```
eudelins-zama: I would classify `Erc1271Error::EmptyRevert(_)` as a `RequestCheckKind::Signature` error, not `RequestCheckKind::Network`.

This is purely for metrics, fyi, not functional. `Erc1271Error::EmptyRevert(_)` is rightly classified as a `Recoverable` error above, which is good 🙂.

eudelins-zama (commented):
LGTM, thanks!

One last tiny comment 😄

kms-connector/crates/kms-worker/src/core/event_processor/error.rs
maksymsur: Addressed in https://github.com/zama-ai/fhevm/pull/3804/commits/937b83a3a1f141b970b0b36d5c6cb431248fe6a9 @eudelins-zama

eudelins-zama (approved):
Thanks!

PR #3812, opened 2026-09-04
feat(kms-connector): decryption endpoints

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/1953

### Summary

Implements the decryption routes of the KMS Connector `endpoint` service. The service re-uses the shared connector database: a request is validated, given a content-derived id, stored in the existing decryption request tables with `source = 'http'`, picked up by the unchanged `kms-worker`, and answered as soon as the worker writes the response row.

### How it works

Each route follows the same flow, shared through the `DecryptionRoute` trait. The body is deserialized, cheaply validated (handle format, chain id...) and hashed into its EIP-712 `decryption_id`, so the id is never accepted from the wire and public and user requests are domain-separated.

The handler then takes an in-flight permit or answers `503 overloaded` immediately, without touching the DB. It registers a waiter for the id. In one transaction it either serves an existing payload or non-retryable error row, or upserts the request: a new row is inserted as `pending`, a `failed` HTTP row is re-armed to `pending`, and any other conflict means the worker is already on it and the connection simply attaches.

A dedicated `ResponseListener` holds a `LISTEN` connection on the `pg_notify` channels fired by the response triggers and wakes every waiter of the notified id, which then reads the response row once and maps it to the `200` body or the `ErrorResponse` of the `kms-connector-api` crate.

### Robustness

Every wait is bounded by `decryption_timeout` (default 30s) and answers `504 timeout`, a new retryable `ErrorCode`, so an orphaned request can never hold an in-flight permit forever.

A wake-up caused by the notification of a stale retryable error row, still in flight when the waiter registered, is detected through the row's `created_at` (overwritten by the worker on every write) and waited through instead of being served.

The `WaiterGuard` releases the permit and the registry entry on drop, including on client disconnect thanks to `h1_allow_half_closed(false)`.

### Out of scope, follow-ups

The `endpoint health` subcommand and the monitoring server are not wired yet, and there is no Dockerfile, CI build job, Helm template or docker-compose entry for the new service. The per-route serve window on stored results and the `Retry-After` header on `503` from RFC 033 are not implemented in this PR. These will be handled in follow-up issues.


kms-connector/config/endpoint.toml
```diff
@@ -0,0 +1,57 @@
...
+# database_pool_size = 16
+
+# Public bind address of the `v1` HTTP decryption interface (optional, defaults to "0.0.0.0:8080")
+# ENV: KMS_CONNECTOR_HTTP_ENDPOINT
+# http_endpoint = "0.0.0.0:8080"
+
+# Maximum number of decryption requests held in-flight by this replica. Beyond this limit, requests
+# are rejected with `503 overloaded` (optional, defaults to 1000)
+# ENV: KMS_CONNECTOR_MAX_IN_FLIGHT_DECRYPTIONS
+# max_in_flight_decryptions = 1000
```
dvdplm, line 23: Just curious: do we have a well-informed intuition on what this number should be in prod? Or could be? The KMS default for a similar setting to this is 10k and I have bumped it to 20k for some benchmarking runs.

kms-connector/config/endpoint.toml
```diff
@@ -0,0 +1,57 @@
...
+# decryption_timeout = "30s"
+
+# Maximum size of a JSON request body, in bytes (optional, defaults to 1 MiB)
+# ENV: KMS_CONNECTOR_MAX_BODY_BYTES
+# max_body_bytes = 1048576
+
+# Maximum total bit size of the handles of a decryption request (optional, defaults to 2048).
+# Mirrors `MAX_DECRYPTION_REQUEST_BITS` of the `Decryption` gateway contract.
+# ENV: KMS_CONNECTOR_MAX_DECRYPTION_REQUEST_BITS
+# max_decryption_request_bits = 2048
```
dvdplm: If this ever goes out of sync with the `Decryption` contract (or the KMS/tfhe-rs) I think a lot of things would break. Should it really be configurable? It feels like it's not really the kms-connector's concern to configure this? Is there an easy way to share this value among projects (other than monorepo)?

kms-connector/config/endpoint.toml
```diff
@@ -0,0 +1,57 @@
...
+# ENV: KMS_CONNECTOR_MONITORING_ENDPOINT
+# monitoring_endpoint = "0.0.0.0:9100"
+
+# The timeout to perform each external service connection healthcheck (optional, defaults to 3s)
+# ENV: KMS_CONNECTOR_HEALTHCHECK_TIMEOUT (format: https://docs.rs/humantime/latest/humantime/)
+# healthcheck_timeout = "3s"
+
+# The maximum number of tasks to process requests concurrently (optional, defaults to 1000)
+# ENV: KMS_CONNECTOR_TASK_LIMIT
+# task_limit = 1000
```
dvdplm: Hmm, is  this a good thing to leave as user-configurable? On the KMS we set tokio threads but leave the task count unbounded I think. I would try to avoid making things configurable unless there's a clear need for it (changing options is a breaking change if one is strictly adhering to semver, which we're not I guess but still...).

kms-connector/crates/api/src/error.rs
```diff
@@ -16,7 +17,7 @@ pub enum ErrorCode {
     SenderAuthenticationFailed,
     /// Per-sender rate limit exceeded (connector proxy).
     RateLimited,
-    /// Endpoint replica at its in-flight cap; comes with `Retry-After` (connector endpoint).
+    /// Endpoint replica at its in-flight cap (connector endpoint).
```
dvdplm:
```suggestion
    /// Endpoint is overloaded.
```

kms-connector/crates/utils/src/tests/db/responses.rs
```diff
@@ -292,3 +292,63 @@ pub async fn insert_rand_epoch_result_response(
...
+    let decryption_id = id.unwrap_or_else(rand_u256);
+    sqlx::query!(
+        "INSERT INTO public_decryption_responses(
+            decryption_id, error_code, error_details, extra_data, created_at, otlp_context,
+            status, source
+        )
+        VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
+        ON CONFLICT (decryption_id) DO UPDATE SET
+            error_code = EXCLUDED.error_code, error_details = EXCLUDED.error_details,
+            created_at = EXCLUDED.created_at",
```
dvdplm, line 311: Do you really want these to be upserts? Isn't it a bug if two tests insert the same decryption_id? And isn't it harder to fix that bug if an upsert happened than if one of the tests blow up?

kms-connector/crates/utils/src/tests/db/responses.rs
```diff
@@ -292,3 +292,63 @@ pub async fn insert_rand_epoch_result_response(
...
+    error_details: &str,
+) -> anyhow::Result<U256> {
+    let decryption_id = id.unwrap_or_else(rand_u256);
+    sqlx::query!(
+        "INSERT INTO user_decryption_responses(
+            decryption_id, error_code, error_details, extra_data, created_at, otlp_context,
+            status, source
+        )
+        VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
+        ON CONFLICT (decryption_id) DO UPDATE SET
```
dvdplm, line 339: same q here

kms-connector/crates/endpoint/tests/integration_tests.rs
```diff
@@ -0,0 +1,914 @@
...
+use tokio_util::sync::CancellationToken;
+use tracing::info;
+
+const CHAIN_ID: u64 = 31337;
+const EUINT64: u8 = FheTypes::Uint64 as u8;
+const PUBLIC_REQUESTS: &str = "public_decryption_requests";
+const USER_REQUESTS: &str = "user_decryption_requests";
+
+#[rstest]
+#[timeout(Duration::from_secs(60))]
```
dvdplm, line 45: Hmm. I'd avoid setting timeouts outside the test itself. If the test itself wants to check the runtime, then put the timeout into the test. If it's not part of the  test, I would avoid it and let CI inform you how long it takes to run and only once it's clear that we need a timeout, add it here.

kms-connector/crates/endpoint/tests/integration_tests.rs
```diff
@@ -0,0 +1,914 @@
...
+            .post(self.url(route))
+            .header("content-type", "application/json")
+            .body(body)
+            .send()
+            .await
+            .expect("raw request failed")
+    }
+
+    /// Fires a public-decrypt request in a background task.
+    fn spawn_post_public(&self, request: &PublicDecryptionRequest) -> JoinHandle<Response> {
```
dvdplm: I don't love this name. "post": does it mean "after"? or "send"? "public": does it mean "out in the open"? or "public decryption"?

Maybe `spawn_pdec()`? I think `pdec` and `udec` are by now established terms and removes the ambiguity?

kms-connector/crates/endpoint/tests/integration_tests.rs
```diff
@@ -0,0 +1,914 @@
...
+                .post(url)
+                .json(&request)
+                .send()
+                .await
+                .expect("request failed")
+        })
+    }
+
+    /// Fires a user-decrypt request in a background task.
+    fn spawn_post_user(&self, request: &UserDecryptionRequest) -> JoinHandle<Response> {
```
dvdplm: Same nitpick as above. How about `spawn_udec`?

kms-connector/crates/endpoint/tests/integration_tests.rs
```diff
@@ -0,0 +1,914 @@
...
+        db,
+        Some(db_id(id)),
+        Some(OperationStatus::Completed),
+        RequestSource::Http,
+    )
+    .await
+}
+
+/// Plays the kms-worker: stores a completed user decryption for `id`.
+async fn complete_user(
```
dvdplm: Here too I'd disambiguate the name. Perhaps `complete_udec`?

kms-connector/crates/endpoint/tests/integration_tests.rs
```diff
@@ -0,0 +1,914 @@
...
+async fn count_rows(db: &Pool<Postgres>, table: &str) -> i64 {
+    sqlx::query(&format!("SELECT COUNT(*) AS n FROM {table}"))
+        .fetch_one(db)
+        .await
+        .unwrap()
+        .get::<i64, _>("n")
+}
+
+async fn wait_for_request_row(db: &Pool<Postgres>, table: &str, id: B256) -> sqlx::postgres::PgRow {
+    let deadline = Instant::now() + Duration::from_secs(10);
```
dvdplm, line 855: 10s is pretty generous... If a DB write didn't happen within 2s I think something is very wrong.

kms-connector/crates/endpoint/src/core/db.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub extra_data: Vec<u8>,
+    pub error_code: Option<String>,
+    pub error_details: Option<String>,
+    pub created_at: DateTime<Utc>,
+}
+
+/// A `user_decryption_responses` row, either a payload or a worker-side error.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct UserDecryptionResponseRow {
+    pub user_decrypted_shares: Option<Vec<u8>>,
```
dvdplm, line 30: naming nitpick: these are the shares that the user will use to reconstruct the final decryption locally I think; the name here sounds like these are decrypted shares (decrypted by the user). I suggest naming this `udec_shares`.

kms-connector/crates/endpoint/src/core/db.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub error_code: Option<String>,
+    pub error_details: Option<String>,
+    pub created_at: DateTime<Utc>,
+}
+
+/// Reads the HTTP-sourced public decryption response of `id`, if any.
+pub async fn read_public_decryption_response<'e>(
+    executor: impl PgExecutor<'e>,
+    id: B256,
+) -> sqlx::Result<Option<PublicDecryptionResponseRow>> {
```
dvdplm, line 42: So `Ok(None)` here means "query went fine, but no pdec result found"?

kms-connector/crates/endpoint/src/core/db.rs
```diff
@@ -0,0 +1,229 @@
...
+        UserDecryptionResponseRow,
+        "SELECT user_decrypted_shares, signature, extra_data, error_code, error_details, created_at
+        FROM user_decryption_responses WHERE decryption_id = $1 AND source = 'http'",
+        db_id.as_slice(),
+    )
+    .fetch_optional(executor)
+    .await
+}
+
+/// Upserts an HTTP-sourced public decryption request, or re-arms it if it previously `failed`.
```
dvdplm:
```suggestion
/// Upserts an HTTP-sourced public decryption request, or sets its status to `pending` if it previously `failed`.
```

kms-connector/crates/endpoint/src/core/db.rs
```diff
@@ -0,0 +1,229 @@
...
+///
+/// Only HTTP-sourced rows are ever re-armed: a Gateway row can never be touched by the endpoint.
```
dvdplm:
```suggestion
/// Only HTTP-sourced rows are ever written to.
```

kms-connector/crates/endpoint/src/core/db.rs
```diff
@@ -0,0 +1,229 @@
...
+        request.extraData.as_ref(),
+        Utc::now(),
+        bc2wrap::serialize(otlp_ctx)?,
+    )
+    .execute(executor)
+    .await
+    .map_err(anyhow::Error::from)
+}
+
+/// Upserts an HTTP-sourced user decryption request, or re-arms it if it previously `failed`.
```
dvdplm: nit: I'd reword this without using "re-arm".

kms-connector/crates/endpoint/src/core/endpoint.rs
```diff
@@ -0,0 +1,70 @@
+use crate::core::{Config, ResponseListener, Waiters, http};
+use anyhow::anyhow;
+use connector_utils::conn::connect_to_db;
+use sqlx::{Pool, Postgres};
+use std::sync::Arc;
+use tokio_util::sync::CancellationToken;
+use tracing::{error, info};
+
+/// The `Endpoint` service: the `v1` HTTP decryption interface of the KMS Connector.
+pub struct Endpoint {
```
dvdplm, line 10: Do you think it's a good idea to add a `version: usize` member already to make it easier to add a v2 of the endpoint? I think it's always a good idea to include a version in all http APIs, so `https://somehost/someapi/v1/someendpoint`. Saves so much hassle down the line.

kms-connector/crates/endpoint/src/core/response_listener.rs
```diff
@@ -0,0 +1,88 @@
...
+            match self.db_listener.try_recv().await {
+                Ok(Some(notification)) => self.handle(notification),
+                // `sqlx` already reconnected and re-issued the `LISTEN`s. Notifications emitted
+                // during the gap are lost, so the in-flight waiters are failed fast.
+                Ok(None) => {
+                    warn!("Postgres listener connection was lost and re-established");
+                    self.fail_in_flight_waiters();
+                }
+                Err(e) => {
+                    break error!("Error while listening for Postgres notifications: {e}");
```
dvdplm: So `Ok(Some())` is the happy path, `Ok(None)` means we have missed some messages but otherwise we're good so we stay in the loop? And `Err` breaks, fails the in-flight-ers and end? I think it would be slightly more readable if the `fail_in_flight_waiters()` call was here in the error path directly, along with a `return`.

kms-connector/crates/endpoint/src/core/response_listener.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    /// Drops the registered waiters, failing their requests with a retryable error.
+    fn fail_in_flight_waiters(&self) {
+        let in_flight = self.waiters.len();
+        if in_flight > 0 {
+            warn!(
+                "Failing the waiters of {in_flight} in-flight decryption(s): their response may have been missed"
+            );
+        }
+        self.waiters.clear();
```
dvdplm: COuld be inside the `> 0` branch I think?

kms-connector/crates/endpoint/src/core/response_listener.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    fn handle(&self, notification: PgNotification) {
```
dvdplm: "handle" is slightly ambiguous in this codebase, perhaps clarify with a rename to `handle_notif`?

kms-connector/crates/endpoint/src/core/validation.rs
```diff
@@ -0,0 +1,356 @@
...
+) -> Result<u64, ValidationError> {
+    let (mut chain_id, mut total_bits) = (None, 0u64);
+    for handle in handles {
+        let fhe_type = extract_fhe_type_from_handle(handle.as_slice()).map_err(|e| {
+            ValidationError::InvalidHandle {
+                handle: *handle,
+                reason: e.to_string(),
+            }
+        })?;
+        let bits = fhe_type_size(*handle, fhe_type)?;
```
dvdplm: Can't `fhe_type_size` take a reference?

kms-connector/crates/endpoint/src/core/validation.rs
```diff
@@ -0,0 +1,356 @@
...
+
+/// Checks the handle list: non-empty, well-formed handles of a decryptable FHE type, on a single
+/// supported chain, within the total bit size budget.
+fn validate_handles<'a>(
+    handles: impl IntoIterator<Item = &'a B256>,
+    config: &Config,
+) -> Result<u64, ValidationError> {
+    let (mut chain_id, mut total_bits) = (None, 0u64);
+    for handle in handles {
+        let fhe_type = extract_fhe_type_from_handle(handle.as_slice()).map_err(|e| {
```
dvdplm: `extract_fhe_type_from_handle` could take a handle ref and then take a slice internally perhaps? Accepting a `&[u8]` seems needlessly flexible.

kms-connector/crates/endpoint/src/core/validation.rs
```diff
@@ -0,0 +1,356 @@
...
+            ValidationError::InvalidHandle {
+                handle: *handle,
+                reason: e.to_string(),
+            }
+        })?;
+        let bits = fhe_type_size(*handle, fhe_type)?;
+        total_bits += u64::from(bits);
+
+        let handle_chain_id =
+            extract_chain_id_from_handle(*handle).map_err(|e| ValidationError::InvalidHandle {
```
dvdplm: Same: why does it need to own the handle?

Maybe we could have a `let (kind, bitsize, chain_id) = handle.metadata()` method to do all three "extractions"?

kms-connector/crates/endpoint/src/core/waiters.rs
```diff
@@ -0,0 +1,237 @@
...
+            let _ = waiter.wake.send(());
+        }
+        true
+    }
+
+    /// Drains the whole registry. Dropping the senders fails every waiter with `RecvError`.
+    pub fn clear(&self) {
+        let drained: Vec<Waiter> = self.lock().drain().flat_map(|(_, v)| v).collect();
+        drop(drained);
+    }
```
dvdplm, line 59: This looks odd to me. What's the point of locking + draining and then allocate each waiter. And then just dropping them?

Can't you `std::mem::replace` the old `HashMap` with a new, empty one and then `drop` the old one?

kms-connector/crates/endpoint/src/core/mod.rs
```diff
@@ -1,5 +1,12 @@
...
+pub mod http;
+pub mod response_listener;
+pub mod validation;
+pub mod waiters;
 
 pub use config::Config;
-pub use entrypoint::Entrypoint;
+pub use endpoint::Endpoint;
+pub use response_listener::ResponseListener;
+pub use waiters::{WaiterGuard, Waiters};
```
dvdplm: Does all of this need to be `pub`? Probably yes, but just want to make sure we don't end up exposing more than we need.

kms-connector/crates/endpoint/src/core/http/version.rs
```diff
@@ -0,0 +1,7 @@
+use actix_web::HttpResponse;
+use kms_connector_api::VersionResponse;
+
+/// `GET /v1/version`
```
dvdplm, line 4: ah nice, you're way ahead of me... :)

kms-connector/config/endpoint.toml
eudelins-zama: Nope, would need benchmark for this.

But from what we see on current connector limitation (CT in-memory consumption etc.), I'd say it should be lower than KMS, maybe even lower that this.

But I'd say it is a good starting point

kms-connector/config/endpoint.toml
eudelins-zama: > If this ever goes out of sync with the Decryption contract (or the KMS/tfhe-rs) I think a lot of things would break.

I don't think this is true for that setting. 

We could hardcode it though, no strong opinion, it just means any update on this value would require a software upgrade.
And no there's no way to share this value easily AFAIK (and don't seem important to me, connector endpoint could have a different value than Decryption contract as it does not have the same limitation EVM has).

kms-connector/config/endpoint.toml
eudelins-zama: All config with default are not meant to be overriden. There are there "just in case" we need to change something without having to make a software upgrade.

This specific config is shared across all connector services. If we want to make a pass to remove it (as well as other config fields), why not, but I'll suggest we do so in a dedicated follow-up :slightly_smiling_face:

kms-connector/crates/utils/src/tests/db/responses.rs
eudelins-zama: Yes it's on purpose, this reproduce the kms-worker behavior. Here's the scenario:

1. HTTP decryption request (endpoint)
2. Retryable processing error, ex network error during a RPC call (kms-worker)
3. This error response is stored in DB (plain insert) to be returned by the endpoint (kms-worker)
4. Endpoint serve the error (endpoint)
5. HTTP decryption request retry (endpoint)
6. Processing is successful (kms-worker)
7. We must upsert the result for this decryption (kms-worker)

I would say debugging is quite easy with the bots these days...

kms-connector/crates/endpoint/tests/integration_tests.rs
eudelins-zama: We use this pattern for almost all integration tests of the connector. Should be part of a dedicated re-work if it's a concern IMO.

That being said, I tends to disagree. Often these timeouts are large enough so we can consider something's wrong. For ex, often the test would be:
1. Simulate an incoming decryption
2. Wait for the decryption to be processed by the system under tests
3. Check the result

Step 2. requires a timeout. Including it in the test itself is not what I would recommend, it would be more verbose, and if many wait operations are done within the test, you would need multiple manual timeouts implem.

kms-connector/crates/endpoint/tests/integration_tests.rs
eudelins-zama: Post is the HTTP verb.

But fair, I'll rename :slightly_smiling_face:

kms-connector/crates/endpoint/tests/integration_tests.rs
eudelins-zama: With the amount of concurrent tests (> 300 at most), all sharing the same DB server, better to be generous.

Losing 8s for a case where something's really wrong is not too much to pay IMO.

kms-connector/crates/endpoint/src/core/db.rs
eudelins-zama: Yes exactly.

This is because before forwarding the request to the kms-core, we first look if there's already a response tied to the decryption_id or not.

kms-connector/crates/endpoint/src/core/db.rs
eudelins-zama: The naming is aligned with what we used to send to the [Decryption.sol](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L757) contract.

This renaming would mean a DB migration to change the column's name and so on, not worth it IMO.

kms-connector/crates/endpoint/src/core/endpoint.rs
eudelins-zama: Idk what we would need in the future tbh. Better to implement this when we would need it? (the version is already part of the route FYI)

kms-connector/crates/endpoint/src/core/response_listener.rs
eudelins-zama: - `Ok(Some())` is the happy path
- `Ok(None)` means we were disconnected, and now we're successfully reconnected, so we **might** have missed some messages during that time window.
- `Err` breaks, fails the in-flight-ers and end

Fair for the readability suggestion!

kms-connector/crates/endpoint/src/core/response_listener.rs
eudelins-zama: Good catch

kms-connector/crates/endpoint/src/core/response_listener.rs
eudelins-zama: Yep

kms-connector/crates/endpoint/src/core/validation.rs
eudelins-zama: It does not, but as the type implements `Copy`, I guess this is not such a big deal? I'll try to use a ref

But as `validate_handles` is still quite short, I would keep the different steps inlined in the loop, without nesting the code within a `metadata` method that the reader would have to dig to understand what it does.

kms-connector/crates/endpoint/src/core/mod.rs
eudelins-zama: For a binary crate, I think we don't care tbh. I'll reduce the scope anyway as it's cheap to do

kms-connector/crates/endpoint/src/core/waiters.rs
eudelins-zama: Good catch :slightly_smiling_face:

kms-connector/config/endpoint.toml
dvdplm: I think this is tied to the largest data type supported by tfhe-rs, Euint2048 and that if it changes, then we are making a major change across the stack(s). I'd hardcode it but perhaps a second opinion is in order?

kms-connector/config/endpoint.toml
dvdplm: Changing the config later is a breaking change, and as soon as it's deployed it kind of matters and is a hassle to change. I think the config should be minimal, OR clearly documented as "you prolly shouldn't touch this".

kms-connector/crates/api/src/error.rs
eudelins-zama: Fixed b66a7046921a2b0ed6caa7fcd26836049a1fc710

kms-connector/crates/endpoint/src/core/db.rs
eudelins-zama: Fixed b66a7046921a2b0ed6caa7fcd26836049a1fc710

kms-connector/crates/endpoint/src/core/db.rs
eudelins-zama: Fixed b66a7046921a2b0ed6caa7fcd26836049a1fc710

kms-connector/crates/endpoint/src/core/db.rs
eudelins-zama: Fixed b66a7046921a2b0ed6caa7fcd26836049a1fc710

kms-connector/crates/endpoint/src/core/response_listener.rs
eudelins-zama: Fixed c5e4005a93157e6565fa51707c9593c620ddad69

kms-connector/crates/endpoint/src/core/response_listener.rs
eudelins-zama: Fixed c5e4005a93157e6565fa51707c9593c620ddad69

kms-connector/crates/endpoint/src/core/validation.rs
eudelins-zama: Fixed c5e4005a93157e6565fa51707c9593c620ddad69

kms-connector/crates/utils/src/tests/db/responses.rs
dvdplm: > I would say debugging is quite easy with the bots these days...

Last famous words! ;)

> 7. We must upsert the result for this decryption (kms-worker)

Here you lost me: why must we upsert? As opposed to simply writing a new row.

kms-connector/crates/utils/src/tests/db/responses.rs
eudelins-zama: [Same](https://github.com/zama-ai/fhevm/pull/3812#discussion_r3949380133)

kms-connector/crates/endpoint/src/core/validation.rs
eudelins-zama: Fixed b068d0132d0154243fa1f58a9f4ccff3d81cdbce

kms-connector/crates/endpoint/src/core/validation.rs
eudelins-zama: Fixed b068d0132d0154243fa1f58a9f4ccff3d81cdbce

kms-connector/crates/endpoint/src/core/waiters.rs
eudelins-zama: Fixed 29b7e06be2e2bc8aef84b742e381b5cfa9608e51

kms-connector/crates/endpoint/src/core/mod.rs
eudelins-zama: Fixed c088491c6a9ca2fac0914a5060e387e41406d1e2

kms-connector/crates/endpoint/src/core/response_listener.rs
eudelins-zama: Fixed ec3394be5a383c27f1d5b8ef426aa78707d3ad85

kms-connector/crates/endpoint/tests/integration_tests.rs
eudelins-zama: Fixed ec3394be5a383c27f1d5b8ef426aa78707d3ad85

kms-connector/crates/endpoint/tests/integration_tests.rs
eudelins-zama: Fixed ec3394be5a383c27f1d5b8ef426aa78707d3ad85

kms-connector/crates/endpoint/tests/integration_tests.rs
eudelins-zama: Fixed ec3394be5a383c27f1d5b8ef426aa78707d3ad85

kms-connector/crates/endpoint/tests/integration_tests.rs
dvdplm: Why does step 2 need a timeout?

If your code is doing something that needs to be done within a deadline, then the deadline is part of the business logic of your code and needs tests to check that we stay within bounds. 

If the deadline is not part of the business logic, then having a `#[timeout]` is an "extra", something added to the harness and the "context" of the running test but ultimately irrelevant for the logic we're testing.

For the first case I'd argue the deadline logic should be inside the test itself, and if it's more verbose, so be it.
For the second case I'd argue we don't need it. Or if we need it we should know precisely why.


---

Imagine you have some code that must be serialized. Dunno, some write to disk that must not happen concurrently. To test that your locking works you would *not* slap a `[#serial]` attribute on the test and then trust `cargo nextest` to exercise the lock. That'd be all kinds of wrong. Instead you'd likely spin up ten threads inside the test, tell them to all write the thing on disk and then check that all writes happened in sequence.

Why is a timeout different?

kms-connector/crates/endpoint/src/core/db.rs
dvdplm: That's a bit sad.

kms-connector/crates/endpoint/src/core/endpoint.rs
dvdplm: I saw later that you're already doing this. 👍

kms-connector/crates/endpoint/src/core/http/decrypt.rs
```diff
@@ -0,0 +1,212 @@
...
+                    format!("no response within {timeout:?}"),
+                    Some(id),
+                )
+            })?;
+        wake_signal.map_err(|_| {
+            warn!(decryption_id = %id, "Response listener gone while waiting");
+            upstream_transient("connection to the response stream was lost", id)
+        })?;
+
+        let row = R::read_response(&state.db_pool, id)
```
dvdplm, line 176: Do you think we should rearm the `receiver` here, before we read the row? Rather than wait until line 196? If the worker writes an update while we're processing it, we would miss a notification I think. Not sure though, double check this!

Before this line, we'd do `let next = guard.rearm()` and then iff we need to, set `receiver = next` on line 196, that way we'd still get the notification if a write happens between here and line 196.

kms-connector/crates/endpoint/src/core/http/decrypt.rs
```diff
@@ -0,0 +1,212 @@
...
+            _ => {
+                tx.commit().await?;
+                debug!(decryption_id = %id, "Serving existing response row");
+                return Ok(LookupOutcome::Serve(row));
+            }
+        }
+    }
+
+    let otlp_ctx = PropagationContext::inject(&tracing::Span::current().context());
+    let result = R::upsert_request(&mut *tx, id, request, &otlp_ctx).await?;
```
dvdplm, line 131: The kms-worker could potentially write to the DB in the time between our reading at line 118 and upserting at line 133. I think the worst consequence is an extra round of work, so perhaps it's fine. Fixing would require a lock on the row I think, which is nasty af.

kms-connector/crates/endpoint/src/core/http/decrypt.rs
eudelins-zama: Indeed, using https://docs.rs/tokio/latest/tokio/sync/watch/index.html might be a better approach, to be confirmed after our sync

kms-connector/crates/endpoint/src/core/http/decrypt.rs
eudelins-zama: > I think the worst consequence is an extra round of work, so perhaps it's fine.

That is my understanding as well

kms-connector/crates/endpoint/src/core/http/decrypt.rs
eudelins-zama: Fixed a716b60232501c0e190796bc62f40c51c4ebce4e

kms-connector/crates/endpoint/src/core/waiters.rs
```diff
@@ -0,0 +1,227 @@
...
+    receiver: Option<watch::Receiver<()>>,
+    _permit: OwnedSemaphorePermit,
+}
+
+impl Waiter {
+    /// Resolves on the next wake-up not observed yet.
+    pub async fn wait(&mut self) -> Result<(), watch::error::RecvError> {
+        self.receiver
+            .as_mut()
+            .expect("receiver is only taken on drop")
```
dvdplm, line 101: Hmm. Not sure I understand why this is guaranteed to be ok. Wouldn't two calls to `wait()` from different threads blow up here?

kms-connector/crates/endpoint/src/core/waiters.rs
```diff
@@ -0,0 +1,227 @@
...
+pub struct Waiter {
+    waiters: Arc<WaiterRegistry>,
+    id: B256,
+    // Only `None` while dropping.
+    receiver: Option<watch::Receiver<()>>,
+    _permit: OwnedSemaphorePermit,
+}
+
+impl Waiter {
+    /// Resolves on the next wake-up not observed yet.
```
dvdplm: Worth mentioning that this blocks until a value is received? Or is that obvious?

kms-connector/crates/endpoint/src/core/waiters.rs
```diff
@@ -0,0 +1,227 @@
...
+            .as_mut()
+            .expect("receiver is only taken on drop")
+            .changed()
+            .await
+    }
+}
+
+impl Drop for Waiter {
+    fn drop(&mut self) {
+        let mut map = self.waiters.lock();
```
dvdplm, line 109: Is it possible that taking the lock panics? It's not right? Because of the `unwrap_or_else()` call? I recall there are gotchas with panicking `Drop` impls...

kms-connector/crates/endpoint/src/core/waiters.rs
```diff
@@ -0,0 +1,227 @@
...
+        // Unsubscribe under the lock so a concurrent `register` cannot race the cleanup below.
+        drop(self.receiver.take());
+        if map
+            .get(&self.id)
+            .is_some_and(|sender| sender.receiver_count() == 0)
+        {
+            map.remove(&self.id);
+        }
```
dvdplm, line 117: Is this correct: if `&self.id` is not in the map, then something went wrong somewhere, but we don't need to do anything here because the `Waiter` is dropped and the `map` is fine anyway? In other words: we do not need an else branch here.

kms-connector/crates/endpoint/src/core/waiters.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        let mut w1 = waiters.register(id, permit(&sem).await);
+        let mut w2 = waiters.register(id, permit(&sem).await);
+        assert!(waiters.contains(&id));
+        assert_eq!(sem.available_permits(), 0);
+
+        assert!(waiters.wake(&id));
+        w1.wait().await.unwrap();
+        w2.wait().await.unwrap();
+        // WaiterRegistry stay registered after a wake-up.
```
dvdplm:
```suggestion
        // WaiterRegistry stays registered after a wake-up.
```

kms-connector/crates/endpoint/src/core/waiters.rs
```diff
@@ -0,0 +1,227 @@
...
+        let sem = Arc::new(Semaphore::new(1));
+        let id = B256::repeat_byte(2);
+        let mut w = waiters.register(id, permit(&sem).await);
+
+        assert!(waiters.wake(&id));
+        w.wait().await.unwrap();
+
+        // `w` is not awaiting `wait()` while this wake-up happens.
+        assert!(waiters.wake(&id));
+        // Its next `wait()` must observe it instead of hanging.
```
dvdplm, line 162: nice

kms-connector/crates/endpoint/src/core/endpoint.rs
```diff
@@ -0,0 +1,72 @@
...
+    waiter_registry: Arc<WaiterRegistry>,
+    response_listener: ResponseListener,
+}
+
+impl Endpoint {
+    pub async fn from_config(config: Config) -> anyhow::Result<Self> {
+        let db_pool = connect_to_db(&config.database_url, config.database_pool_size).await?;
+        let waiter_registry = Arc::new(WaiterRegistry::new());
+        let response_listener =
+            ResponseListener::connect(&db_pool, waiter_registry.clone()).await?;
```
dvdplm: Maybe `Arc::clone(waiter_registry)` here? ;)

kms-connector/crates/utils/src/tests/db/responses.rs
eudelins-zama: > Here you lost me: why must we upsert? As opposed to simply writing a new row.

Because same request = same decryption_id . And the decryption_id is the primary key of the table

kms-connector/crates/endpoint/src/core/waiters.rs
eudelins-zama: `&mut self` can't be use as is by different threads right?
Not sure to understand your concerns here tbh, we just wait for the receiver to receive something from its sender

kms-connector/crates/endpoint/tests/integration_tests.rs
eudelins-zama: Fair.

For the first case, deadlines that are part of the behavior are already asserted inside the tests (`wait_for_request_row` / `wait_for_request_status`) I guess.

So the annotation is more a guard against hanging test. We could remove it, but then let's consider a follow-up to align all tests behavior.

kms-connector/crates/endpoint/src/core/waiters.rs
eudelins-zama: `unwrap_or_else(|e| e.into_inner())` is what makes that lock non-panicking

eudelins-zama: I'll add a comment for this

kms-connector/crates/endpoint/src/core/waiters.rs
eudelins-zama: >  if &self.id is not in the map, then something went wrong somewhere

Not necessarily, we could have simply `WaiterRegistry::clear()` then drop the waiter.

Not sure to understand if there's something to change here, lgtm

kms-connector/config/endpoint.toml
eudelins-zama: Done 3d8e7c31fed9c807c0f6e712eac60d473ac83d12

kms-connector/config/endpoint.toml
eudelins-zama: Removed 3d8e7c31fed9c807c0f6e712eac60d473ac83d12 (was not even really used under the hood)

kms-connector/crates/endpoint/src/core/waiters.rs
eudelins-zama: bd50f0993d78cee87edbe1c364fe99c05972d4f2

kms-connector/crates/endpoint/src/core/waiters.rs
eudelins-zama: Done bd50f0993d78cee87edbe1c364fe99c05972d4f2

kms-connector/crates/endpoint/src/core/endpoint.rs
eudelins-zama: bd50f0993d78cee87edbe1c364fe99c05972d4f2

kms-connector/crates/endpoint/src/core/waiters.rs
eudelins-zama: bd50f0993d78cee87edbe1c364fe99c05972d4f2

dvdplm (approved):
Lgtm, and thank you for humoring me on the config stuff!

Issue #1909, opened 2026-02-05
Remove typos ignore for .sol and fix typos


Eikix:
WHAT
Remove typos ignore for `.sol`, then fix all typos in `.sol` files flagged by `typos`.

WHY
We want typos checks enabled for Solidity, but only after the current backlog is cleaned so CI stays green.

HOW
1) Remove `.sol` from `typos.toml` `extend-exclude`.
2) Run `typos` and fix all `.sol` typos it reports.
3) Re-run `typos` to confirm clean for `.sol`.

Eikix:
Closing this for now due to deployment safety.

Even typo-only edits in Solidity files can affect compiled artifacts and interact with upgrade/deploy flows wired in infra. We don’t want typo cleanups to accidentally trigger contract rollout behavior.

If infra confirms a safe rollout path for Solidity typo checks, we can reopen and do this.

PR #3817, opened 2026-09-04
fix(solana): land zama-host review pass

Eikix:
## Summary

Land Nico's zama-host review as one host PR: cheap PDA creates, EVM-parity HCU through euint128, opaque 32-byte KMS context ids, and the admin/hygiene pass (upgrade-authority init, atomic `set_admin` / EIP-712, no taint bit).

## Changes

- Fast-path `create_account` for empty PDAs; keep the 3-CPI squat fallback; builder floor is 1 CPI per persistent create
- Hardcoded EVM HCU tables through euint128; comparisons meter operand width; drop euint256/euint160
- Init admin from BPF upgrade authority; one `set_admin` with new-admin co-sign (skip if PDA); one EIP-712 domain setter
- KMS context id is opaque `[u8; 32]`; public-decrypt return is 96 bytes; v1 extra_data is exact 33
- Hygiene: Sum/IsIn precheck, destroy-already-destroyed, coprocessor no-op, remaining-account asserts, setter tests, `user_address` docs

Closes zama-ai/fhevm-internal#1982

## Second review round (7b1488eaa3)

Two more reviews were run at 0da75e0aff, one for correctness and one for structure. Both lists are closed in 7b1488eaa3:

- HCULimit.sol parity test moved to its own `hcu/tests/evm_parity.rs`; the `checked > 250` floor is replaced by exact assertions (every pair agrees, every parsed cell reached).
- `initialize_host_config` tests moved to `host_admin_mollusk.rs` and table-driven; `host_mollusk.rs` shrinks by 366 lines.
- `ExceedsInstructionTraceLimit` deleted as unreachable; `MAX_PERSISTENT_CREATES` documented as a policy cap; INVARIANTS #54, DESIGN_DECISIONS and the boundary ceiling note updated.
- Relayer `extra_data` collapsed to one length rule per version and one error variant.
- Nine oracle `*_is_unsupported` tests collapsed to one table; `type_bits` stays a pure lookup.
- `program_data_account` built from `UpgradeableLoaderState`; redundant type-gate checks and the test-only bounded-rand helper removed.
- `set_admin` gains the co-signing keypair happy path and the unsigned keypair rejection.

## Review close-out (0da75e0aff, rebased onto feature/solana@439b734b54)

Two review passes are on this PR. Everything blocking and every should-fix is closed. The last commit fixes what the first CI failure was hiding: `cargo test` stops at the first failing binary, so host_mollusk's stale 900-HCU trivial-encrypt expectation (EVM: 32) kept operator_conformance from running, and that suite had lost nine euint160/euint256 rejection tests when the oracle stopped mapping those types. The oracle now asks the host type gate first. The four-literal HCU spot check is now a test that parses `HCULimit.sol` and compares every shared cell both ways; RandBounded has its own row (no ebool cost in the EVM). One doc comment that Markdown read as a list fixed the workspace clippy step.

Left open on purpose, all nits: metering runs before the type gate so a non-scalar `Div` reports `HcuUnknownCost` rather than `DivisorMustBeScalar`; `set_admin` reuses `HostConfigAdminMismatch` for three conditions (a new error variant changes the IDL and the generated clients); `assert_reduction_count` runs in the walk precheck and again in the shared type predicate the builder also uses; no boundary profile pins the instruction-trace wall since the create fast path moved the binding wall to heap.

## Rebase onto #3831, #3832, #3830, #3834 (d2c0c97cc2)

- Invariant 36 keeps #3834's text plus this branch's `remove_subject` sentence.
- #3834's shared `HostConfig` decoder in `zama-solana-acl` walked past an 8-byte KMS context id; this branch widens it to 32 bytes, so the decoder's `PAUSED_OFFSET` and documented account size (301 -> 325) move with it. The host-side pin test `shared_crate_decoder_reads_what_the_program_serializes` is what caught the shift.
- Batcher cost snapshot regenerated on top of #3831/#3832.


Eikix (commented):
# Review: fix(solana): land zama-host review pass (#3817)

Where it lands, it lands cleanly: the HCU tables match `HCULimit.sol` cell-for-cell (284 shipped cells diffed, 0 mismatches), and the `[u8;32]` context id moved consistently through state, PDA seeds, events, the pinned IDL, the golden manifest and the generated TS client (`HostConfig` 301 → 325 bytes). Two things block: the bring-up's public-decrypt context id still depends on the truncation this PR removes, and no cost snapshot was regenerated after the create fast path changed the CPI count.

## Findings

### Blocker

- **The E2E public-decrypt vertical can no longer resolve its context** — `test-suite/fhevm/src/layout.ts:152` still sets `SOLANA_DEFAULT_PUBLIC_DECRYPT_CONTEXT = 0x07…0001`, which reaches the certificate's `extra_data[1..33]` via `e2e/harness/solana/vertical.ts:60` and `public-decrypt.ts:101`, while the bring-up defines the context at `canonicalTestContextId(1)` = `0x00…0001` (`deploy.ts:118`) and `verify_public_decrypt.rs:137-141` now compares all 32 bytes plus the canonical PDA. Define the bring-up context at the tagged id the gateway actually mints (ZH-12).
- **Cost snapshots not regenerated** — `runtime-tests/cost-snapshots/host_mollusk.json:12` still records `cpi_instructions: 3` and `fhe_execute_boundary.json:80-101` still pins `all_created_public` at 61 CPIs with a sibling `limited_by: "instruction_trace"`, but the fixture's output PDA is unseeded (`host_mollusk.rs:3688`) so it takes the new one-CPI path (`common.rs:394`), and `assert_cost_snapshot` fails on drift in either direction. Run `scripts/update-cost-snapshots.sh`; those `limited_by` fields are where ZH-01's "correct the created-public stop" belongs.

### Should-fix

- **Worst-case trace budget no longer bounds the squat branch** — `crates/zama-fhe/src/cost.rs:168` charges one CPI per create plus one 3-CPI meter allowance, but a pre-funded output PDA costs 3 each, so 20 creates with two event kinds and an armed block cap is `2+60+2+3 = 67` against a 64-entry trace. Charge `CPIS_PER_SQUAT_CREATE * persistent_creates`, or drop the claim that a budget against it "lands regardless of on-chain state".
- **The new init authority gate has no negative test** — `initialize_host_config.rs:19-23` is the PR's most security-relevant new check, yet every init test fabricates `program_data_account(admin)` (`host_mollusk.rs:3242`, …). Add ZH-06's negatives: a signer that is not the upgrade authority, a non-signer admin, a wrong `program_data` address.
- **`set_admin`'s off-curve escape defeats the guard it replaces** — `set_admin.rs:34-39` skips co-sign for any key not on the curve, and roughly half of all mistyped 32-byte values are off-curve, so a typo becomes admin unsigned with no on-chain repair. Make the caller prove the PDA (owner program plus seeds), or keep the two-step propose/accept shape; the only test is the PDA happy path (`host_admin_mollusk.rs:131`).
- **No test pins any HCU value to its EVM literal** — `hcu/tests.rs:339` asserts relationships only, "so calibration can change numbers freely", and each magnitude appears exactly once in the tree. Generate the table from `HCULimit.sol`, or pin it in a fixture.
- **A finalized or non-upgradeable deployment can never be initialized** — `initialize_host_config.rs:21` requires `upgrade_authority_address == Some(admin)`, which is `None` after `set-upgrade-authority --final`. Add ZH-06's pinned bootstrap-key fallback, or record it as an invariant.
- **The double-destroy reject this PR adds has no test** — `destroy_kms_context.rs:41-44`'s `require!(!destroyed)` is asserted nowhere; only `mollusk_destroy_kms_context_rejects_current` was added.
- **`remove_subject` still accepts trailing account metas** — ZH-18f named three handlers; `encrypted_value.rs:177` and `verify_public_decrypt.rs:118` got the assert, `remove_subject.rs` did not. Add the line plus one negative.
- **Docs still describe the old behaviour** — `constants.rs:46` ("~3 CPIs" per created output), `EVM_PARITY.md:82` ("Uint8..Uint256"), `:194` ("uncalibrated placeholders"), `:67` (single admin, no rotation), `DESIGN_DECISIONS.md:1757` (`SPACE 151 -> 293`), `INVARIANTS.md:244` (no rotation). No design decision records the upgrade-authority init gate, the PDA-skip-co-sign rule, or the EIP-712 domain setter.

### Nit

- `relayer/src/host/extra_data.rs:31-38` still accepts v1 `extra_data` at `len() >= 33` where the host now demands exactly 33, and `validations.rs:205-207` still says the v3 context id "is not 0x07-tagged"; ZH-12 asked for both sides of the wire.
- Metering runs before the type gate (`fhe_execute.rs:101` vs `:128`), so a non-scalar `Div` dies with `HcuUnknownCost` (`hcu/mod.rs:94`) instead of `DivisorMustBeScalar`; error codes are client-facing.
- `set_admin.rs:32/37` reuses `HostConfigAdminMismatch` for three distinct conditions, and `test-kit/src/oracle.rs:423-424` still maps `7 => 160`, `8 => 256`.
- `RandBounded` meters through `RAND` (`hcu/mod.rs:385`), which carries an `ebool` row `checkHCUForFheRandBounded` lacks (unreachable); `assert_reduction_count` now runs twice for `Sum`/`IsIn`.
- The `user_address` fix (`DESIGN_DECISIONS.md:231`) states the divergence but drops the consequence the issue asked for: an observer can replay another user's input.
- `meter_comparison_prices_operand_width_not_ebool` covers only the `EarlierStep` pricing path; the dictionary and `VerifiedInput` paths (`hcu/mod.rs:308-316`) are untested.

## CI on the head (64a0ea75f2)

Three Solana jobs fail before any test runs, so the mollusk and snapshot suites have not executed on this head; the two blockers above come from reading the code and fixtures.

- `solana-tests/build-and-test` and `solana-proof-service-tests/fmt-clippy`: rustfmt drift in `zama-host/src/eip712.rs:122`, `instructions/common.rs:579`, `runtime-tests/tests/token_mollusk.rs:47,5010`, `test-kit/src/lib.rs:19,25,449` (import ordering, e.g. `canonical_test_context_id` in the `test-kit` export list).
- `solana-tests/test-suite-fhevm`: `tsc` fails at `test-suite/fhevm/src/solana/deploy.test.ts:89` with TS1308, an `await` outside an async function.

## Coverage against Nico's review

| ZH | Status | Where |
|---|---|---|
| 01 | Done; snapshots/docs stale | `common.rs:394`, `cost.rs:81` |
| 02 | Done, values verified | `hcu/mod.rs:80-184` |
| 03 | Done (1-operand sum = 90,900) | `hcu/mod.rs:173` |
| 04 | Done, prices on lhs as EVM does | `hcu/mod.rs:298-348` |
| 05 | Done (3-knob search) | `common.rs:554-594` |
| 06 | Done; negatives missing | `initialize_host_config.rs:19` |
| 07 | One-step + co-sign per the issue, not two-step | `set_admin.rs` |
| 08 | Done | `set_eip712_domain.rs` |
| 09 | Done (init event dropped) | `events.rs`, `initialize_host_config.rs:70` |
| 10 | Partial: 3 setters + depth + destroy-current; both delegation instructions still untested, unmentioned | `host_admin_mollusk.rs` |
| 11 | Done (v1 exactly 33) | `eip712.rs:132` |
| 12 | Done host-side; relayer comment unchanged | `eip712.rs:126-148` |
| 13 | Deferred (issue: not in scope) | — |
| 14 | Partial: Sum/IsIn hoisted; subject cap deferred (#1964) | `walk.rs:265` |
| 15 | Deferred to #1964 / RFC 035 | — |
| 16 | Deferred to #1907 / DD-044 | — |
| 17 | Done at every call site | `fhe_execute.rs`, `walk.rs` |
| 18 | a–e done; f partial | `common.rs:320`, `walk.rs:174` |
| 19 | a, c, d, f done; e done by dropping contiguity; b deferred (#1909) | `destroy_kms_context.rs:41` |
| Docs | Partial; v0-extra-data qualifier still missing from `INVARIANTS.md:174` | — |

## Not checked

- Only `cargo test -p zama-host --lib` (117 pass) and `cargo check --workspace --all-targets` ran; the mollusk, snapshot and E2E suites need `anchor build`, so both blockers come from the code and committed fixtures, not a failing run.
- Cross-instruction trace exhaustion (ZH-01 item 2).
- `solana-mcp`'s `program_autofixer` was unavailable.

Eikix (commented):
# Re-review at 0ec2cdc5c5

Both blockers are fixed and verified: `BRINGUP_KMS_CONTEXT_ID` is now the tagged `0x07…0001` the certificate constant already carried (`addresses.ts:48`), and all four cost snapshots are regenerated — `host_mollusk.json:12` reads `cpi_instructions: 1` and `all_created_public` moved to `max_ok: 22, limited_by: "heap"`, exactly the number Nico predicted for the decoupled heap wall. Six of eight should-fix items are fully closed, two are partial, and three nits remain open; nothing in the follow-up looks wrong.

| Item | Status | Where |
|---|---|---|
| Blocker: E2E public-decrypt context | Fixed | `addresses.ts:33-48`, `deploy.test.ts:102-104` |
| Blocker: cost snapshots | Fixed | `host_mollusk.json:12`, `fhe_execute_boundary.json:80-83` |
| Worst-case trace budget vs squat | Fixed | `cost.rs:163-176`, test `:230-238` |
| Init authority negative tests | Fixed | four new tests, `host_mollusk.rs` (`rejects_wrong_upgrade_authority`, `rejects_unsigned_admin`, `rejects_wrong_program_data_address`, `rejects_finalized_program`) |
| `set_admin` off-curve escape | Fixed | `set_admin.rs:37-39` adds `owner != system_program::ID`; negative at `host_admin_mollusk.rs:168` |
| HCU values pinned to EVM literals | Partial | `hcu/tests.rs:357-369` pins 4 cells; ~280 of the 284 shipped cells are still unpinned |
| Finalized / non-upgradeable init | Partial | documented (`initialize_host_config.rs:19`) and the reject is now tested; no bootstrap-key fallback |
| Double-destroy test | Fixed | `host_mollusk.rs` `mollusk_destroy_kms_context_rejects_already_destroyed` |
| `remove_subject` remaining accounts | Fixed | `remove_subject.rs:24`; negative `mollusk_remove_subject_rejects_remaining_accounts` |
| Stale docs (6 sites) | Fixed | `constants.rs:46`, `EVM_PARITY.md:67/82/194`, `DESIGN_DECISIONS.md:1757`, `INVARIANTS.md:244` |
| Nit: relayer v1 length + v3 comment | Fixed | `relayer/src/host/extra_data.rs` now rejects non-33-byte v1 with `WrongLength` + test; `validations.rs:205-207` comment corrected |
| Nit: metering before the type gate | Not addressed | `fhe_execute.rs:101` vs `:128` |
| Nit: `set_admin` error reuse / `oracle.rs` 160-256 | Partial | `oracle.rs:420-424` cleaned; `set_admin.rs:31/38` still one error for three conditions |
| Nit: `RandBounded` ebool row / double `assert_reduction_count` | Not addressed | `hcu/mod.rs:385`, `walk.rs:265` |
| Nit: `user_address` consequence | Partial | added at `README.md:114`; `DESIGN_DECISIONS.md:231-235` still omits it |
| Nit: comparison-width test paths | Not addressed | `hcu/mod.rs:308-316` dictionary and `VerifiedInput` paths still untested |
| CI: rustfmt drift (4 files) | Fixed | `cargo fmt --all -- --check` clean |
| CI: TS1308 `deploy.test.ts:89` | Fixed | `await` hoisted to `deploy.test.ts:89-90` |

Coverage rows re-checked: **ZH-10** still partial — `delegate_for_user_decryption` and `revoke_delegation_for_user_decryption` are constructed by no Rust test. **ZH-14** unchanged (output-subject cap still deferred to #1964). **ZH-18f** now complete, all three handlers assert. **ZH-19b** unchanged — `define_kms_context` still never stamps `updated_slot`, so `INVARIANTS.md:229` ("every change stamps `updated_slot`") stays false as written; deferred to #1909. **Docs** still partial: `INVARIANTS.md:174-176` has no v0/v1/v3 qualifier on the extra-data binding chain.

## New findings

- **Nit** — `canonicalTestContextId` (`test-suite/fhevm/src/solana/addresses.ts:27`) is now an unused export; its only remaining reference is the doc comment at `:45`. Delete it or mark it test-only.
- **Nit** — deleting `WallPin::TraceOneUnderBuilderCap` (`fhe_execute_boundary.rs`) leaves no boundary profile with `limited_by: "instruction_trace"`, so an `instruction_trace_floor` regression is now pinned only by the `cost.rs` unit test, not by a measured sweep.

## CI on the head (0ec2cdc5c5)

The rustfmt and `tsc` failures from the previous head are gone. One Solana job fails now: `solana-tests/demo-dapp`, at the step "Check the dapp's generated program clients still match the committed IDLs". The IDL under `solana/idl` changed in this PR (32-byte context id, `HostConfig` size, dropped types) and the dapp's generated clients were not regenerated to match. Everything else Solana-side is still running at the time of writing.

## Not checked

- `cargo test -p zama-host --lib` (118 pass) and `cargo fmt --all -- --check` ran. The mollusk, cost-snapshot and E2E suites still need `anchor build` and a validator, so the new negative tests and the regenerated snapshot numbers were read, not executed.
- Cross-instruction trace exhaustion (ZH-01 item 2).

PR #3823, opened 2026-09-04
fix(coprocessor): fix blue-green consensus divergence

goshawk-3:
fixes: https://github.com/zama-ai/fhevm-internal/issues/1998


coprocessor/fhevm-engine/consensus-detector/src/state_hash.rs
```diff
@@ -513,6 +546,19 @@ async fn compute_and_upload_state_hashes(
...
+        // each host chain against the retained gcs schema. compute_and_insert_gcs
+        // skips already-hashed blocks (NOT EXISTS) and is bounded by batch_limit,
+        // so a full [0, i64::MAX] range is safe and idempotent.
+        let chain_rows = sqlx::query("SELECT chain_id FROM public.host_chains ORDER BY chain_id")
+            .fetch_all(&mut *tx)
+            .await?;
+        for row in chain_rows {
+            let chain_id: i64 = row.try_get("chain_id")?;
+            compute_and_insert_gcs(&mut tx, chain_id, 0, i64::MAX, batch_limit, "public").await?;
+        }
```
PanGan21, line 561: Is the "LOCAL CHANGE" needed to be committed? Should be removed i guess

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -522,46 +534,94 @@ async fn prune_gcs_computations_before_start(
...
+          --    below vacuously true (nothing to find yet != nothing wrong).
           COALESCE(
             (SELECT MAX(block_number) FROM public.host_chain_blocks_valid WHERE chain_id = $1),
             -1
           ) >= $3
+          -- 2. start_block itself is finalized — anchors the dry-run window on a
+          --    settled block that cannot be reorged out.
+          AND EXISTS (
+              SELECT 1 FROM public.host_chain_blocks_valid
+              WHERE chain_id = $1 AND block_number = $3 AND block_status = 'finalized'
```
PanGan21, line 596: This needs the `start_block` row to still exist in `host_chain_blocks_valid` while the host-listener prunes finalized rows older than 10.000 blocks.

If the dry run stays in UpgradeActivated longer, the row is pruned and readiness stays false forever. We should either protect this row from the prune, or not require it once the watermark is past `start_block`.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -1441,6 +1619,51 @@ async fn authorized_stack_version(
...
+        let mut pbs_total: i64 = 0;
+        for (chain_id, pending) in &pbs_pending {
+            warn!(
+                host_chain_id = chain_id,
+                pending,
+                "cutover blocked: BCS pre-window PBS computations not yet completed — ct128/ciphertext_digest not yet produced"
+            );
+            pbs_total += pending;
+        }
+        return Err(Error::PendingBcsUploads { pending: pbs_total });
```
PanGan21, line 1664: Can a `pbs_computations` row stay `is_completed = false` forever? The host-listener creates one for every allowed handle, and the sns-worker only completes it once the ciphertext exists. If the computation errored, the ciphertext never comes and the row never completes.

Isn't it the case that a block like this blocks every cutover attempt. Should we skip handles whose computation is terminally errored, like the readiness check now does?

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -1303,6 +1476,11 @@ async fn delete_bcs_gw_leftovers(tx: &mut Transaction<'_, Postgres>) -> Result<(
         "delete_bcs_gw_leftovers: deleted BCS-leftover ciphertexts and input_handles"
     );
 
+    // verify_proofs
+    sqlx::query("TRUNCATE TABLE public.verify_proofs")
+        .execute(&mut **tx)
+        .await?;
```
PanGan21, line 1482: This also deletes pre dry run proofs that blue verified right? 
Isn't it more correct to delete only `block_number >= gw_start_block`, like the `input_handles` deletes?

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -1168,15 +1206,150 @@ async fn delete_bcs_chain_leftovers(
...
+        );
+        return Ok(());
+    }
+    for (chain_id, start_block) in rows {
+        // By-handle tables first: reached through the computation rows deleted below.
+        for handle_table in CIPHERTEXT_TABLES.iter().chain(["ciphertext_digest"].iter()) {
+            for comp_table in COMPUTATION_TABLES {
+                let sql = format!(
+                    "DELETE FROM {GCS_SCHEMA_QUOTED}.{handle_table} h
+                      USING {GCS_SCHEMA_QUOTED}.{comp_table} comp
```
PanGan21, line 1246: This delete only finds a ciphertext if its computation row is still in `gcs.computations`. However the prune at DryRunStarted already deleted the pre-start computation rows. So a pre-start ciphertext whose computation row is gone is not deleted here, and it still gets merged into public.

`gcs.pbs_computations` is not pruned and also has block_number. So we could use that to find the handles instead.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
```diff
@@ -62,9 +62,21 @@ pub use fhevm_engine_common::versioning::EVENT_STACK_VERSION_UPGRADED;
 /// Number of host-chain blocks below `start_block` whose computations must
 /// also be fully settled before GCS can leave `UpgradeActivated`. Hard-coded
 /// for now; expected to become configurable.
-const READINESS_CONFIRMATIONS: i64 = 100;
+// Set equal to the host-listener's `BLOCKS_VALID_RETENTION` (10_000). The readiness
+// window `[start_block - READINESS_CONFIRMATIONS, start_block]` then spans the full
+// retained block history; its low end sits at the prune horizon, so the oldest
+// confirmation blocks may already have been pruned and pass the check vacuously.
+// Keep the two constants in lockstep: if `BLOCKS_VALID_RETENTION` changes, change this.
+const READINESS_CONFIRMATIONS: i64 = 10_000;
```
PanGan21, line 70: Going from 100 to 10_000 blocks makes this query read a lot more rows, and it re-runs on every `ciphertext-computed` notification against the live DB. Did you check how long it takes on a chain like Sepolia? It seems like a performance hit

goshawk-3: It took at most several milliseconds on the testbox as we have DB indexes in place.  We also log the elapsed time so we can monitor it.

The increase is intended to reduce the risk of false-positive readiness checks.

PanGan21: Ok fair! Thanks for checking

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
goshawk-3: If it has already been pruned, that means the active tip BCS is following is 10K blocks ahead of our `start_block`, which would imply enormous downtime. In such case, we should drop the proposal.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
goshawk-3: Shouldn't be needed since the verify_proofs rows are temporary but as defensive check, this can be added too.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
goshawk-3: That's  a good catch. We should skip handles whose computation is terminally errored. Ideally, a pbs_computations must be marked as failed if the corresponing fhe op is terminally errored.

coprocessor/fhevm-engine/upgrade-controller/src/lib.rs
goshawk-3: The GCS host-listener does not insert computations < start_block per chain. The deletion above seems like a leftover.

goshawk-3:
@mergifyio queue

PanGan21 (approved):

PR #3825, opened 2026-09-04
fix(relayer): fix parsing error when replaying req from v0.13 after restart

manoranjith:
Implements [fhevm-internal#1975][issue]. Cross-repo reference, so merging will not close it — needs closing by hand.

[issue]: https://github.com/zama-ai/fhevm-internal/issues/1975

Rows already stored in the tagged shape become unreadable, so **dev's database must be cleared before this ships**. Mainnet and testnet hold untagged rows only and are unaffected. Per the issue's sequencing this lands here first, ahead of promotion, then on `main` where the recovery call site differs.

Deviations from the issue's proposal:

- The "either way" clause — write `failure` with an `err_reason` on an unreadable payload — is not implemented. One stored shape, selected by a `NOT NULL` column the writer derives from the request itself, leaves the parse no failing input to handle.
- `UserDecryptReq::parse_req_data` is deleted rather than rewired: it has no callers on this branch, and removing `Deserialize` from the enum stops it compiling. The `main` change keeps it, where it is live.

The test fixtures cite the `v0.13.0` structs and write path they were derived from. Without that the shape claim rests on the author's word, and a wrong fixture would still pass.

The first commit is unrelated to the fix: the committed offline query data still described the `user_decrypt_req_type` enum as it stood before migration `20260519152353`, so `sqlx prepare --check` failed on a clean tree.

Test plan

- [ ] CI: relayer `cargo clippy --all-targets --features integration-tests` (locally clean).
- [ ] CI: relayer `cargo test --lib` (locally 204 passed, 1 ignored, including 4 new).
- [ ] CI: `sqlx prepare --check`.

Follow-ups

- The issue's recovery regression — seed a v0.13-shape row, restart, assert recovery — is not included. The unit tests cover the parse, not `startup_recovery` end to end.


nboisde (approved):
lgtm !

tawadaa (approved):

PR #3826, opened 2026-09-04
chore(kms-connector): endpoint healthcheck

eudelins-zama:
Healthcheck implementation for the new `kms-connector/crates/endpoint` service

### DIFF
--- a/kms-connector/crates/endpoint/src/core/endpoint.rs
+++ b/kms-connector/crates/endpoint/src/core/endpoint.rs
@@ -10,7 +10,6 @@ use tracing::{error, info};
 
 /// The `Endpoint` service: the `v1` HTTP decryption interface of the KMS Connector.
 pub struct Endpoint {
-    config: Arc<Config>,
     app_state: http::AppState,
     response_listener: ResponseListener,
 }
@@ -34,7 +33,6 @@ impl Endpoint {
         );
 
         let endpoint = Self {
-            config,
             app_state,
             response_listener,
         };
@@ -45,8 +43,9 @@ impl Endpoint {
     pub async fn start(self, cancel_token: CancellationToken) -> anyhow::Result<()> {
         info!("Starting Endpoint");
 
-        let server = http::run_server(self.app_state, self.config.http_endpoint)?;
-        info!("HTTP server listening at: {}", self.config.http_endpoint);
+        let http_endpoint = self.app_state.config.http_endpoint;
+        let server = http::run_server(self.app_state, http_endpoint)?;
+        info!("HTTP server listening at: {http_endpoint}");
         let server_handle = server.handle();
 
         let mut listener_task = tokio::spawn(self.response_listener.start());
--- a/kms-connector/crates/endpoint/src/core/endpoint.rs
+++ b/kms-connector/crates/endpoint/src/core/endpoint.rs
@@ -1,7 +1,9 @@
-use crate::core::{Config, ResponseListener, WaiterRegistry, http};
+use crate::{
+    core::{Config, ResponseListener, WaiterRegistry, http},
+    monitoring::health::State,
+};
 use anyhow::anyhow;
 use connector_utils::conn::connect_to_db;
-use sqlx::{Pool, Postgres};
 use std::sync::Arc;
 use tokio_util::sync::CancellationToken;
 use tracing::{error, info};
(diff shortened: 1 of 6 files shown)


kms-connector/crates/endpoint/src/core/endpoint.rs
```diff
@@ -1,41 +1,51 @@
...
+            config.http_endpoint,
+            Arc::clone(&app_state.in_flight_limiter),
+            config.max_in_flight_decryptions,
+            config.healthcheck_timeout,
+        );
+
+        let endpoint = Self {
+            config,
+            app_state,
             response_listener,
```
dvdplm, line 37: Not sure what I'd do differently tbh, but the `Config` is added kind of all over the place here:
- inside the `AppState`
- parts of it in the `State`
- then again in the `Endpoint` (with an extra copy inside the `app_state` member)

Is there something we can do about that that I'm not seeing?

kms-connector/crates/endpoint/src/monitoring/health.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    fn service_name() -> &'static str {
+        "kms-connector-endpoint"
+    }
+}
+
+/// Serializable representation of `Endpoint`'s health status.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct HealthStatus {
```
dvdplm, line 84: Do you think adding a timestamp here is useful?

kms-connector/crates/endpoint/src/monitoring/health.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Returns the number of decryption requests currently held in-flight by this replica.
+    fn in_flight_decryptions(&self) -> usize {
+        self.max_in_flight_decryptions
+            .saturating_sub(self.in_flight_limiter.available_permits())
+    }
+}
+
+impl Healthcheck for State {
+    async fn healthcheck(&self) -> actix_web::HttpResponse {
+        let (db_res, http_res) = tokio::join!(
```
dvdplm, line 50: Would it be a good idea to wrap this in a `tokio::time::timeout` so we have a guarantee that we don't hang here? Like, if e.g. the DB is slooooooow to connect we could end up waiting for a long time here.

```rust
let (db_res, http_res) = timeout(Duration::from_secs(3), async { tokio::join!( … …) }).await;
```

or something along those lines?

kms-connector/crates/endpoint/src/monitoring/health.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Database connection status.
+    pub database_connected: bool,
+    /// Whether the public `v1` HTTP server answers requests.
+    pub http_server_reachable: bool,
+    /// Number of decryption requests currently held in-flight by this replica (informational).
+    pub in_flight_decryptions: usize,
+    /// Maximum number of in-flight decryption requests of this replica (informational).
+    pub max_in_flight_decryptions: usize,
+    /// Details about any issues encountered during healthcheck.
+    pub details: String,
```
dvdplm, line 96: A `Vec<dyn std::error::Error>` would be nicer here?

dvdplm (dismissed):
LGTM

Q: do you think it should ping the KMS as well, sort of how this PR checks both the DB and itself, it could also check that the KMS is reachable.

kms-connector/crates/endpoint/src/monitoring/health.rs
eudelins-zama: There is a `timeout` in the `database_healthcheck` implem

kms-connector/crates/endpoint/src/monitoring/health.rs
eudelins-zama: This is the type for what we return to the HTTP client.

What we want to return to the client is a plain `Vec<String>`, not a `Vec<dyn std::error::Error>` I would say.
And AFAIU, this would not compile as this would not implement `Deserialize`/`Serialize`, which is needed for this feat.

kms-connector/crates/endpoint/src/core/endpoint.rs
eudelins-zama: Small improvement: 290ca22302654d026e4d330339939ceb54168281

eudelins-zama:
> Q: do you think it should ping the KMS as well, sort of how this PR checks both the DB and itself, it could also check that the KMS is reachable.

Nope, this is the healthcheck implem of the `endpoint` service alone.

The ping to the KMS is done by the healthcheck implem of the `kms-worker` already :slightly_smiling_face:

kms-connector/crates/endpoint/src/monitoring/health.rs
eudelins-zama: Could be!

TL;DR from our discussion would be to generalized this to other services as part of a follow-up work

kms-connector/crates/endpoint/src/monitoring/health.rs
dvdplm: Hmm. Not sure I follow, but the conversion from error to "error string" should be the responsibility of the serialization layer, instead this PR is casting to String and concatenate. Maybe that's what you're saying too?

dvdplm (approved):

kms-connector/crates/endpoint/src/monitoring/health.rs
eudelins-zama: > Maybe that's what you're saying too?

I think yes, basically the `HealthStatus` is used as the serialization layer, via the derived trait. This would require a bit of (overengineered?) work to accept this `Vec<dyn std::error::Error>`, probably a manual implementation of the trait (not sure it would even compiled, compiler was complaining about size of the data not being known at compile time with such thing).

eudelins-zama:
@mergifyio queue

PR #3835, opened 2026-09-06
chore(coprocessor): run `upgrade-controller` and `consensus-detector` tests in ci

PanGan21:
### Summary

- Adds `upgrade-controller` and `consensus-detector` to the `coprocessor-cargo-test` matrix.
- Switches the `upgrade-controller` tests from `ImportMode::WithKeysNoSns` to `ImportMode::None`, since they only cover the upgrade state machine and GCS schema. Importing the keyset per test OOM killed the runner.

Related: https://github.com/zama-ai/fhevm-internal/issues/1988


goshawk-3 (approved):

PanGan21:
@Mergifyio queue

PR #1985, opened 2026-02-13
feat(test-suite): rewrite fhevm-cli as Bun/TypeScript CLI

Eikix:
## Summary

- Rewrites the 340-line bash `fhevm-cli` as a typed Bun/TypeScript CLI using Commander.js
- Same Docker Compose orchestration, same arguments — now with typed modules for env management, deploy DAG, test dispatch, and docker operations
- Original `fhevm-cli` becomes a 3-line bash passthrough (`exec bun run src/index.ts "$@"`)

### New modules
| Module | Purpose |
|--------|---------|
| `dag.ts` | 13-step deploy DAG with explicit dependency edges |
| `env.ts` | Version defaults + env file preparation (replaces `cp`+`sed`) |
| `docker.ts` | `composeUp`/`Down`, `waitForService`, `getContainerIp` |
| `executor.ts` | `Bun.spawn` wrapper with dry-run recording for testing |
| `cache.ts` | BuildKit local cache env var computation |
| `test-suites.ts` | Suite name → grep pattern mapping |
| 10 commands | `up`, `test`, `clean`, `logs`, `restart`, `upgrade`, `pause`, `unpause`, `status`, `doctor` |

### Test coverage (100 tests, 856 assertions)
- **Unit tests**: DAG invariants, env patching, cache vars, test-suite mappings
- **Migration tests**: structural diff against bash deploy script + argument parity against original bash CLI (pinned to `a4a9aa47`)
- **Live e2e runner**: compares both CLIs against a Docker stack (requires running services)

## Test plan

- [ ] `bun test` passes all 100 tests (no Docker required)
- [ ] `bunx tsc --noEmit` type-checks cleanly
- [ ] `bun run src/index.ts --help` lists all 10 commands
- [ ] `bun run src/index.ts up --dry-run` traces the full 13-step deploy sequence
- [ ] Live: `./fhevm-cli up --build` deploys the full stack identically to the old bash script
- [ ] Live: `./fhevm-cli test input-proof` runs tests identically
- [ ] Live: `bun run src/e2e-runner.ts test --all` compares both CLIs

Issue #1973, opened 2026-02-12
Solana PoC: canonical HCU tracking (tx + global window)


Eikix:
Last synced: 2026-02-12

## Context
Canonical tracker for HCU behavior in the Solana host-listener PoC.

Current model is intentionally 1:1-ish with EVM tx semantics, extended with a global window guard to protect Zama-side throughput on fast chains.

## Done
- Tx-scoped HCU meter is active (`begin_hcu_meter` / `close_hcu_meter`).
- Global HCU usage window is active at meter close (`last_seen_window_id`, `used_hcu_in_window`, `limit_hcu_per_window`).
- Revert paths covered in host-program tests (invalid authority, tx cap, depth cap, tracked-handle cap, global cap).
- Solana exploration docs refreshed to current state (full op-surface + HCU model).

## Remaining
- Define hardening strategy beyond PoC global cap (fee-market and/or per-signer quotas).
- Document/lock the SDK-level queue/throttle contract for dapps to reduce user-facing reverts.
- Expand diff-test matrix for HCU revert equivalence vs known EVM behavior where intended.
- Validate parameterization of window size/cap under realistic throughput assumptions.

## Non-goals (for this PoC)
- Perfect anti-DoS economics.
- Production-grade fairness scheduling.
- Full policy layer for all downstream app patterns.

Status: Active

PR #3836, opened 2026-09-06
feat(solana): rfc 035 one-path authorization

Eikix:
Implements fhevm-internal RFC 035 (one-path authorization) across the Solana host, the shared ACL crate, the SDK builder, the reference apps, the coprocessor host listener, the KMS connector, the relayer, the JS SDK, the demo dapp, the test suite and the docs.

## What changes

**Identity.** An encrypted value account lives at `["encrypted-value", program, authority, scope, label]`. On create, the output authority is proven to be a PDA of `program` using the declared seeds. Updates reuse that stored identity, require the authority signature and expected previous handle, and do not re-prove seeds. `scope` is what that program declares (the mint for the token program). The pair `(program, scope)` is the application: it is what the HCU block meter charges, the deny list names, the permit scopes to, and the rand seed binds. `compute_subject` is gone; a stored value is read into a computation under its authority's signature and nothing else. (DD-047)

**Allows.** The account keeps no list of who may decrypt. A write declares the keys allowed on the handle it installs; the host seals one keccak `HistoricalAccessLeaf` per key, then the `PublicDecryptLeaf` when the output is `make_public`. `allow_subjects`, `remove_subject`, `authorize_current`, `previous_subjects` and the subject cap are deleted. A viewer decrypts and nothing else. Account size is `181 + 32·peaks`, at most 2229 bytes. (DD-048)

**Deny list.** `set_deny_scope` / `DenyScopeRecord` at `["deny-scope", program, scope]` gates every allow the host would seal, including `make_handle_public`. An `fhe_execute` passes the record of every application whose value it touches, including a foreign stored amount and one written under an additional signing authority (the transfer receipt), so a denied application cannot be written into from another program's execution; the meter and the rand seed still key on the default authority's application. Token transfer and burn forward these records in first-touch order, omitting duplicate applications. A denied key is no longer a concept.

**Rand.** Every execution with a rand step passes and advances the host's `RandNonce` singleton; the seed is `H(nonce, op_index, program, scope, host id, chain id, bank hash, timestamp)`. The persistent-write anchor is gone. (DD-043 amended)

**Leaf record.** The standalone `solana-proof-service` is deleted. The host listener recomputes every leaf from the confirmed instruction stream and stores it in the same database transaction as the compute rows; it serves proofs over `POST /v1/solana/leaf-proofs` behind an API key (committed OpenAPI, RFC 033 errors).

**Connector.** One pipeline: strict decode → signature → window → deployment → KMS pair → confirmed snapshot (two reads when an entry is delegated) → pause → watermark → per-entry scope against the permit's `allowedScopes` → proof requests fanned out to every configured coprocessor → verify all peer candidates against the deciding snapshot's peaks → retry only unresolved retryable queries once → delegation freshness. A peer's leaf count cannot displace a valid proof; an ahead path is shortened to the snapshot's peak height before verification, with strict leaf-index bounds. Failed refreshes preserve successful queries. RPC and proof HTTP requests share `host_rpc_call_timeout`, including body reads. Clients supply no proof; a request carrying one is rejected. Delegation records are consumed (INVARIANTS #27 closed).

**Bootstrap / restart.** On an empty database, optional `--start-slot` anchors inclusive Yellowstone replay to an existing confirmed block. A saved committed checkpoint wins on restart. `Tip`, `Resume`, and `ReplayFrom` carry the apply/skip distinction through subscription and validation. RPC supplies the anchor hash; Yellowstone must supply blocks and historical sysvar context. Local retention is 256 slots; this adds no archival RPC recovery or historical on-chain peaks.

**Relayer / SDK / apps.** The relayer drops the proof passthrough and keeps an advisory dead-delegation pre-check. Permits carry `allowedScopes` (≤ 7 pairs, empty = permissive); a user-decrypt entry is `{handle, encryptedValueAccount, allowedKey?}`; public-decrypt `extraData` v3 is exactly `0x03 ‖ ctx ‖ account` (65 bytes). The RPC evidence source and proof-service clients are deleted. Confidential-token gains `allow_balance_viewers` / `allow_total_supply_viewers` (a re-write by the authority) and `make_*_handle_public`.

**Tests and docs.** Leaf vectors (`solana/test-fixtures/leaves`), connector auth vectors and permit vectors regenerated with pinning tests; cost snapshots regenerated (`allow_heavy_*` profiles replace `subject_heavy_*`); the dead-surface sweep rejects retired ACL collection/mutation names, `compute subject`, ACL-sense `domain`, and `value account`, while allowing unrelated English or local variables named `subject`. The full guard and its self-test pass locally. GLOSSARY, INVARIANTS (retired in place), MMR_ACL_MVP, EVM_PARITY, TESTING* and the READMEs rewritten; DD-032/035/037/039/043 amended; DD-047 and DD-048 added.

## Coverage changes to be aware of

- Wallets cannot own values any more (a value's authority is a program PDA), so the wallet-signed `fhe_execute` e2e driver is gone. The live e2e now drives two specimen programs (`encrypted-counter`, `dep-chain`). The **live operator matrix** (every binary/unary/comparison op through the coprocessor) and the **input-flow row** (VerifiedInput + scalar add → public decrypt) no longer run end to end; the operators are covered in-process by the runtime tests and the attested-input path is exercised live only by the token burn and the deposit-arc join.
- `PublicOutputsProducedEvent` is still emitted and now has no consumer in this repository (the proof service was the last). Retiring it is a host/IDL change left for fhevm-internal#1665.
- Deployed Yellowstone replay of historical sysvars still needs operational validation; listener replay-state tests alone do not establish provider retention behavior.

## Verified locally

Per step (see commit messages): `cargo test` for `zama-solana-acl`, `zama-host`, `zama-fhe`, the runtime tests incl. Mollusk and cost snapshots, host-listener (`solana_reconstruct::`, `solana_leaves_tests` against Postgres), `kms-worker solana`, connector-utils vectors, relayer; SDK vitest, demo-dapp vitest + tsc, test-suite bun tests + tsc, Codama `codegen:solana --check`, `select-overrides.test.sh`; `dead-surface-check.sh` and its self-test; Solana workspace `cargo fmt --check` + `clippy -D warnings`. Live scenarios are CI's job (`solana-e2e`).

## Latest fix validation and review

Current documented head: [`55c952851`](https://github.com/zama-ai/fhevm/commit/55c9528513f5ad60072181bb73b5b2518d4731e5).

The proof/deadline, token deny-record, bootstrap and vocabulary fixes passed 188 focused local tests: 58 Connector, 96 token runtime, 33 listener, and one real-Postgres proof-route test. Host/token SBF were rebuilt with pinned Solana/Anchor tools; token cost snapshots were regenerated with the pinned Rust toolchain. Formatting, Clippy, and the full dead-surface self-test passed. The subsequent replay-state simplification passed all 33 listener tests and feature-enabled all-target Clippy, plus pre-commit checks.

Two independent reviews of the fixes found no introduced correctness regression and one structural issue: duplicate enum/boolean replay state. The final cleanup carries the enum end to end and removes the boolean/tuple conversion. CI, including Solana E2E, is still pending for this head; see live checks for its current status.

The reports attached below are historical snapshots against `c297709e5e`, not a current list of unresolved findings. The broader application naming cleanup, `AppScope` module extraction, retirement of `PublicOutputsProducedEvent`, and the wider live operator matrix remain separate work. The connector's DB-backed `acl` suite was not rerun locally in this fix pass.

Reviewer guide: https://gist.github.com/Eikix/0fd6f1e313a7072cffd471874ae5669d

## Open on the spec side

RFC 035 amendments this implementation implies: `scope` back as a declared field with the verified-program argument; account growth "bounded" not "reserved"; no compute identity; the rand nonce. RFC 034 should be reconciled with the implemented `allowed_key` / `encrypted_value_account` request fields; there is no remaining `subject` wire field in this implementation.

---

<details>
<summary>Correctness review (Fable 5.1) — written against c297709e5e, before the fix commit</summary>

# RFC 035 correctness review — branch `elias/solana-rfc-035` (b8b572d23a..c297709e5e)

**Verdict: not mergeable as is. One Major security finding in the host program (deny list bypass through the "foreign signing authority" rule introduced in Step 4), plus doc/guardrail drift that follows from the same change. Everything else reviewed (ACL crate, connector pipeline, leaf record, HTTP route, relayer pre-check, extraData v3, SDK permit) is sound.**

Scope: read `git diff b8b572d23a...HEAD` for zama-host, zama-solana-acl, kms-worker `core/solana`, host-listener `solana_leaves.rs` / `http_server.rs` / `solana_grpc_listener.rs` / `solana_reconstruct.rs`, relayer `solana_delegation_precheck.rs` / `extra_data.rs`, SDK `publicDecryptCertificate.ts`, docs. Ran `cargo test -p zama-solana-acl` (28 pass) and `cargo test -p zama-host` (112 pass). Did not run Mollusk/SBF suites or e2e.

---

### Major 1 — A denied application can still compute and seal allows on its own values

**Where:** `solana/programs/zama-host/src/instructions/fhe_execute/preflight.rs:88-97` (`fold_app` skips any value whose authority is not the default signer), `fhe_execute.rs:97-101` (deny check runs only `if let Some(app)`), `fhe_execute/block_cap.rs:44` (`app == None` is refused only under a finite block cap; ship default is `u64::MAX`).

**What is wrong:** Step 4 of the worklog changed the record's rule "all persistent values of one execution share one `(program, scope)`" to "only values under the *default* authority fold into the application". The deny record is then located and checked for the folded application only. A persistent output written under an *additional signing* authority (`output_authority_index: Some(i)`) is never deny-checked against its own `(program, scope)`, and an execution with **no** default-authority value has `app == None` and skips the deny check entirely.

**Failing scenario (no accomplice needed):** admin denies `(P, S)`. Program `P` CPIs `fhe_execute` with `encrypted_value_account_authority` = any throwaway signer it controls (a second PDA of `P`, or the user's wallet it forwards), and lists its real value authority `A = PDA(P, ...)` as a signing remaining account, `output_authority_index = Some(idx_A)`, `output_allow_indexes = [viewer]`, `previous_handle_index` = the current handle. Preflight: `fold_app(A, (P,S))` returns early (A != default), `app` stays `None`, `mark_deny_record` is skipped, `assert_all_used` passes (A is marked by `mark_output_authority`). `fhe_execute`: no deny check; `block_cap::charge` returns immediately under the default unrestricted cap. Walk: `persistent_output_authority` accepts A (signer, equals declared), `bind_execution_output` updates the value and `seal_allow_leaves` seals the viewer's leaf. The connector will later authorize that viewer from the sealed leaf. Under a *finite* block cap the same program anchors on a clean sibling scope `(P, S2)` value under the default authority and gets the same result (`values_of_another_signing_authority_keep_their_own_application` pins this behavior as intended). The worklog (Step 4) only names the "friend program" variant; the self-service variant is worse.

**Consequence:** INVARIANTS #10, DD-048 §4, EVM_PARITY row 66 and the record's guardrail "Denied scope cannot compute or allow" are all false for this path. `make_handle_public` is not affected (it reads the value's own `(program, scope)`).

**Suggested fix:** deny-check every distinct `(program, scope)` the execution *writes* (and, to match the record, reads), not just the folded one. Concretely: in preflight collect a small `Vec<AppScope>` of every persistent output's declared app and every operand's stored app (bounded by `MAX_FHE_EXECUTION_STEPS`), require the canonical deny record for each to be present among remaining accounts when the list is enabled, and run `check_scope_not_denied_info` on each. Keep the metering identity as the default authority's app if the receipt/batcher composition must stay. Add a Mollusk negative: denied `(P,S)` value written under `output_authority_index: Some`, with and without a default-authority anchor, both → `ScopeDenied`. Update INVARIANTS #10 / DD-047 / MMR_ACL_MVP to say what the code does.

### Minor 2 — HCU meter can be charged to a different application than the one whose values are written

**Where:** same mechanism as Major 1; `block_cap.rs` meters `app` = the default authority's app.

**What is wrong:** Under a finite block cap, a program can route its heavy executions onto a cooperating program's meter (or, with an unrestricted cap, be entirely unmetered) by writing its values under an additional signing authority. INVARIANTS #40/#41 say "the application is `(program, scope)` with program verified from the output authority ... nobody mints applications under a program they do not control" — true, but the meter can still be someone else's with their consent. Low impact (needs the other program's `invoke_signed`), worth a sentence in INVARIANTS #41 or a per-written-app charge. No code change required if documented.

### Minor 3 — Docs state the one-application rule the code no longer enforces

**Where:** `solana/docs/DESIGN_DECISIONS.md:2256-2257` (DD-047: "every stored operand and output it touches must carry the same pair (`FheExecuteMixedScopes`)"), `solana/docs/MMR_ACL_MVP.md:25-27` (same), `solana/docs/INVARIANTS.md:71-79` (#10 "every allow the host seals passes the deny list ... the value's application record").

**What is wrong:** `FheExecuteMixedScopes` only fires for values under the default authority (`preflight.rs:88-97`); values under additional signing authorities are exempt by design (Step 4). DD-047 should carry the amendment the worklog records, and #10 must be rewritten once Major 1 is fixed (or downgraded to GAP now).

### Minor 4 — INVARIANTS #54 quotes a boundary profile that is not in the snapshot

**Where:** `solana/docs/INVARIANTS.md:492-493` ("`mature_updates_peaks_8` 19 steps, `_32` 7, the 55-peak extreme 4"); `solana/runtime-tests/cost-snapshots/fhe_execute_boundary.json` has `mature_updates`, `mature_updates_peaks_8` (max_ok 19) and `mature_updates_peaks_32` (max_ok 7) — no 55-peak profile.

**Fix:** drop the 55-peak clause or add the profile. The other quoted numbers (`allow_heavy_public_creates` max_ok 17 / first_fail 18, the 18–20 band, `181 + 32·peaks ≤ 2229`, `MAX_FHE_EXECUTION_STEPS = 32`, `MAX_MMR_PEAKS = 64`) match code and snapshot.

### Minor 5 — Leaf-proof route does one account load per query

**Where:** `coprocessor/fhevm-engine/host-listener/src/http_server.rs`, `leaf_proofs` loop → `prove` → `load_recorded_leaves` (two SELECTs, all leaves of the account) per query.

**What is wrong:** correctness is fine (each query gets a self-consistent state row + leaves cut at `leaf_count`), but a 64-query batch naming the same account 64 times does 128 queries and re-hashes all leaves 64 times (`mmr_peaks_from_leaves`). Since the connector dedups by `(account, handle, kind)` and `MAX_REQUEST_HANDLES` is 33 this is bounded, but an unauthenticated... no, it is API-keyed, so this is a cost nit only. Group queries by account before loading.

### Nit 6 — `expect` in program code

`solana/programs/zama-host/src/instructions/fhe_execute/walk.rs:64-70` (`rand_seed` `expect`). Unreachable today because `consume_rand_nonce` runs first; the worklog already flags it. Returning `Err(FheExecuteRandNonceMissing)` costs nothing.

### Nit 7 — `solana-e2e.yml` still triggers on a retired branch name

`.github/workflows/solana-e2e.yml:32` lists `elias/solana-proof-service-recovery`. Worklog defers it "for the PR discussion"; noting so it does not get lost.

### Nit 8 — `mmr_append` capacity guard is one off from its doc

`solana/crates/zama-solana-acl/src/mmr.rs:52` checks `peaks.len() > MAX_MMR_PEAKS`; the 64-peak case is caught by the `checked_add` on `leaf_count` instead. Behavior is correct (test `append_at_peak_cap_fails_without_mutating` pins it); the guard could be `>=` for clarity. No action required.

---

## Verified sound

**zama-host**
- `assert_authority_is_program_pda` (`fhe_execute.rs`): `create_program_address` over the declared seeds (bump included) must equal the declared authority, on every create; a wallet or another program's PDA cannot pass, and the authority must also sign. Unit test covers wrong program, wallet, out-of-range seed.
- Update path (`bind_execution_output`): declared `(program, authority, scope, label)` must derive `output_info.key()` and `read_canonical_encrypted_value` re-derives the stored fields to the same key with bump check, so declared == stored; seeds are refused on update; `previous_handle` must equal the stored handle (create refuses a previous handle); the signer must equal the declared authority (`persistent_output_authority`). No way found to write a value one does not control.
- Leaf order: `seal_allow_leaves` in list order on the NEW handle, then `append_public_decrypt_leaf`; keys non-zero and distinct (`assert_allow_keys`). `sealing_matches_shared_value_account_reconstruction` pins parity with the crate's `reconstruct`.
- `make_handle_public`: pause, authority signature, current-handle equality, deny record for the value's own `(program, scope)` by canonical address, realloc.
- Deny record location: by canonical derived address among remaining accounts (never by index), uninitialized = not denied, initialized must be program-owned, right size, matching fields, `denied == false`; disabled list refuses a passed record. Foreign record → `DenyRecordMissing` (Mollusk test).
- RandNonce: Anchor `seeds`/`bump` constraint on the Option account; required iff a rand step exists; `checked_add`; seed = `H(nonce, op_index, program, scope, host id, chain, bank hash, ts)`; nonce consumed once per execution, `op_index` separates steps. Mollusk `rand_consumes_the_nonce_and_never_repeats_a_seed` exists.
- Account size: `8 + 5·32 + 8 + (4 + 32·peaks) + 1 = 181 + 32·peaks`, max 2229; `grow_account_if_needed` never shrinks, tops up rent from payer, requires writable; `create_pda_strict` refuses non-empty/non-system/executable targets and tolerates lamport squats.
- Pause gating on `fhe_execute` and `make_handle_public`; `verify_public_decrypt` is read-only and unaffected (acceptable).
- VerifiedInput: `attestation.contract_address` must equal the execution's program; refused with no application.

**zama-solana-acl**
- Seeds: constant tag + four fixed-width fields, no hashing, unambiguous. keccak256 for leaf commitments and MMR nodes with distinct prefixes; leaf index bound into the commitment; account key bound in. `mmr_verify` checks `leaf_index < leaf_count`, `peaks.len() == popcount`, sibling count == mountain height, exact peak match. `authorize_historical`/`authorize_public` bind handle and key/exactness. `reconstruct` uses a single running counter; tests cover swap, drop, multi-mountain, cross-account non-authorization.

**kms-connector**
- One observation point: all rules after the reads use `observation` (the second read for delegated requests, ordered by `deciding_after`). Accounts are resolved from that snapshot; proofs are verified against `encrypted_value_account.encrypted_value().peaks` of that snapshot and nothing else (`check_leaf`). No client-supplied proof exists in the wire type (`request.rs`); the coprocessor's answer is a path plus counts, and only the path enters verification.
- Resolution: presence, owner == host program, discriminator, borsh, `create_program_address(seeds + stored bump) == key`, sentinel authority refused. Scope check per entry on the account's `(program, scope)`; empty list permissive (EVM parity, documented).
- Delegation: `allowed_key != signer` selects the branch; the leaf proven is the delegator's; rows read from the deciding snapshot; live = not revoked, `expiration_slot >= observed`, `last_update_slot <= observed`; wildcard row derived from the reserved sentinel, which resolution refuses as a real authority.
- Proof reader: batch dedup, fan-out, `merge_outcomes` monotone, one retry only when behind; failure classes: `NoLeaf`/`HistoryIncomplete`/`MmrStateInconsistent` terminal, others retryable/transient. Public decrypt: one handle, `0x03 ‖ ctx ‖ account` parsed, same resolution + `check_public_binding`.

**host-listener**
- Failed transactions dropped before decoding; instructions resolved in execution order (top-level then its inner CPIs) so leaf sources and the seeds event pair correctly; `apply_block` writes compute rows, leaves and checkpoint in one transaction; inclusive replay of the applied checkpoint block is `SealDecision::Replay` (skipped), so leaves are not double-applied; divergence (`CreateOnRecordedAccount`, `PreviousHandleMismatch`, `PublicHandleMismatch`) is fatal; incomplete histories tracked without leaves and answered `HistoryIncomplete`.
- HTTP: bearer key compared before body parse, constant-time; `MAX_LEAVES_PER_REQUEST = 64` matches the connector's cap; allowed needs key / public refuses key; the route recomputes peaks from stored leaves and refuses (502) on mismatch; OpenAPI pinned by test.

**relayer / SDK**
- Pre-check refuses only when both rows are dead (absent/revoked/expired) at the read slot; every unreadable/misshapen state passes; the connector's stricter refusals (`NewerThanObservation`, `TupleMismatch`, `ForeignOwner`) are `Indeterminate` here, so nothing the connector could authorize is refused (modulo the documented lagging-node absent window). `extra_data.rs`: v3 is exactly 65 bytes, trailing bytes and truncation rejected. SDK `buildSolanaPublicDecryptExtraData` asserts 32+32 and emits 65 bytes; permit `PERMIT_MAX_SCOPES = 7` = `MAX_ALLOWED_SCOPES`, 64-byte scopes.

---

## Guardrail table (record §2) vs. tree

| Guardrail | Status |
|---|---|
| Leaves hash identically in program, indexer, connector (`leaves_v1.json`) | **Partial.** Replayed by `zama-solana-acl/tests/leaf_vectors.rs`, `host-listener/database/solana_leaves.rs`, `sdk/js-sdk/src/solana/proof.test.ts`. No kms-connector test replays the vectors (it reuses the crate, so the risk is low, but the table says one per implementation). |
| `shared_crate_decoder_reads_what_the_program_serializes` + new `shared_crate_address_matches_program_seeds` | **Partial.** The second test does not exist by that name; the first now asserts `find_program_address(decoded.seeds())` equals the program's `canonical_address()`, which covers the intent. |
| `FORBIDDEN` tokens in the three teardown twins | **Weaker than stated.** Connector list has `AccessEvidence`, `authorize_current`, `access_proof`, `proof_leaf_count`, `encrypted_value_id`; relayer has the last three; SDK has camelCase equivalents. None of the three pins `subjects`, `compute_subject` or `domain`; those are covered only by `dead-surface-check.sh` aliases. |
| Connector reads state once/twice + one proof batch, counted | **Present** (`solana_snapshot_architecture.rs::the_leaf_record_is_read_once_and_once_more_only_when_behind`). |
| `connector_auth_v1.json` regenerated | **Present.** Not re-derived here. |
| Rand seeds never repeat within a slot; nonce required | **Present** (`host_mollusk.rs:3647-3717`). |
| Compute needs the authority: wallet / wrong-program PDA / unsigned rejected | **Present** (`host_mollusk.rs:480`, `:527`, `:577`, `:2997`). |
| Denied scope cannot compute or allow | **Weaker than stated.** `mollusk_denied_application_cannot_write_or_seal_but_its_sibling_scope_can` covers only the default-authority path; the foreign-authority path (Major 1) is not tested and is a bypass. |
| Cost in budget; INV #36 numbers | **Mostly.** Snapshots present and match INVARIANTS except the "55-peak extreme 4" clause (Minor 4). |
| IDL / Codama in sync | Not verified in this review (script-driven). |
| Dead surface / vocabulary | **Present** (`dead-surface-check.sh` rules for subject, domain, compute subject; `value_key` exception removed). |
| End to end, "proof service restart gate re-pointed to the coprocessor" | **Changed, not re-pointed.** No restart gate exists; scenarios wait on `waitForSnsCommit` (test-suite `stack.ts`). Scenarios for current, historical, delegated and public decrypt exist as described. Not run here. |
| Reviews | This report. |

</details>

<details>
<summary>Code-quality review (Opus) — written against c297709e5e, before the fix commit</summary>

# RFC 035 code-quality review — `elias/solana-rfc-035` (b8b572d23a..HEAD, 11 commits, 485 files)

**Verdict: approve with changes.** The retirement itself is clean: no proof-service surface survives
in config, env templates, compose, Dockerfiles, CI jobs, commitlint scopes or module references
outside the three teardown gates and the deliberate historical records; `dead-surface-check.sh` is
clean; no test was skipped, weakened or deleted to get green (zero added `#[ignore]`, `.skip(`,
`todo!()`, `assert!(true)`, or `#[allow(dead_code)]` on a helper anywhere in the diff), and every
committed fixture has a pinning test. The findings are where a grep sweep cannot look: one new wire
contract spelled twice with nothing testing the seam, a response field and an on-chain event with no
consumer, three behaviours whose only tests died with the deleted service, and an invariant marked
`[HOLDS]` that the host does not enforce.

> **Caveat on freshness.** The worktree was clean when this review started and is now dirty: 24 files
> are modified against HEAD (`c297709e5e`), including a `deny_scope_record` → `deny_scope_records`
> multi-application refactor across the token and batcher programs, edits to `fhe_execute.rs` /
> `preflight.rs` / `heap_budget/proofs.rs` / `docs/TESTING.md`, and three deleted cost snapshots
> (`host_mollusk.json`, `token_mollusk.json`, `vault_mollusk.json`). Every finding below was
> re-verified against `git show HEAD:<path>`, so all line numbers are HEAD's. Findings on those 24
> files may already be overtaken by the work in progress.

---

## Major

**M1. The leaf-proof wire is spelled twice with nothing testing the seam.**
`coprocessor/fhevm-engine/host-listener/src/http_server.rs:118-175` declares `LeafQueryKind` /
`LeafQuery` / `LeafProofRequest` / `LeafProof`;
`kms-connector/crates/kms-worker/src/core/solana/proof.rs:249-283` re-declares `LeafQueryWire` /
`LeafKindWire` / `LeafProofRequestWire` / `LeafProofWire` independently. Nothing checks them against
each other: the connector's two mocks build the expected request with the connector's own serializer
(`tests/solana_public_decrypt_carrier.rs:154`, `src/core/event_processor/decryption.rs:1967`), so the
request assertion is the serializer against itself; and the committed OpenAPI
(`host-listener/openapi/solana_leaf_proofs.json`) is read by exactly one test — the listener's own
generated-vs-committed staleness pin (`http_server.rs:466-478`). No connector test reads it. This
seam already produced a live bug during the build (tag-only `rename_all`, `leaf_index` vs
`leafIndex`), caught by chance rather than by a gate.
*Fix:* a committed golden pair under `solana/test-fixtures/leaf-proofs/`, following the convention
this repo already uses for `leaves_v1.json`, `permit_v1.json` and `connector_auth_v1.json`: one
request body and one response body per outcome variant. The listener test deserializes the request
and serializes the response against it; the connector test does the mirror. That is the only thing
that makes two spellings one contract.

**M2. `LeafProof::Found.peaks` is a response field with no reader.**
`http_server.rs:161` declares it, `:411` fills it from `recorded.state.peaks`, and the OpenAPI
documents it. The connector's `LeafProofWire::Found` (`proof.rs:270-276`) does not declare the field
at all — serde drops it on the floor. It can never be used: `proof.rs:1-10` states the record is "a
source of proofs, never of decisions" and the connector verifies against the peaks it read on chain.
Its only reader is the test written because the field exists
(`tests/solana_leaves_tests.rs:207-218`). It is also the largest thing in the response — up to 64 hex
peaks per `Found`, up to 64 `Found`s per batch.
*Fix:* delete `peaks` from `LeafProof::Found`, from the fill at `:411` and from the OpenAPI; drop the
assertion at `solana_leaves_tests.rs:212-218`.

**M3. `merge_outcomes` — the rule that decides which coprocessor's answer authorizes a decrypt — has
no test, and a doc added by this change claims it does.**
`kms-connector/.../solana/proof.rs:110-135` (`rank`, `merge_outcomes`), fan-out at `:186-205` and the
partial-failure path at `:430-453`. Every connector test uses `ScriptedProofReader` (a single record,
`tests/solana_support/mod.rs:928`); the one test using the real `HttpHostProofReader` configures
exactly one endpoint (`tests/solana_public_decrypt_carrier.rs:109-113`). So "a proof beats no proof,
more history beats less, one unreachable coprocessor cannot sink a request" is asserted nowhere, and
`is_behind` (`:102`) is only reachable through that untested path. Meanwhile
`solana/docs/TESTING_STRATEGY.md:58`, new in this change, claims the suite covers "the leaf-proof
read with its fan-out merge and one retry". Only the retry half is true
(`tests/solana_snapshot_architecture.rs:604`).
*Fix:* a table test over `merge_outcomes` covering every ordered pair of the four outcomes plus the
`NotFound` leaf-count tiebreak, and a two-mock-server case where one coprocessor is behind and the
other 500s. Until then, correct `TESTING_STRATEGY.md:58`.

**M4. `PublicOutputsProducedEvent` is emitted on chain and consumed by nothing; the docs that
justify emitting it are now false.**
`solana/programs/zama-host/src/events.rs:33-40`, emitted at
`src/instructions/fhe_execute/event_transport.rs:13,48`, sized by the transport guard there, carried
in the IDL (`host-listener/idl/zama_host.json:1718,3778`) and pinned by
`solana/scripts/check_solana_abi.py:122`. Its only in-repo consumer was the deleted proof store; the
listener resolves created-public handles from instruction data plus streamed entropy
(`solana_reconstruct.rs`), and only `FheExecuteRandomSeedsEvent` is actually decoded
(`solana_grpc_listener.rs:1083-1095`). `DESIGN_DECISIONS.md:1525-1528` records this and defers the
deletion to #1665, which is a decision — but three live texts still state the opposite as current
fact: `events.rs:5-9` ("carry the only data an indexer cannot recompute from instruction data at
all", now true of only one of the two), `.github/workflows/solana-e2e.yml:11-13`, and
`solana/programs/zama-host/README.md:78-80`.
*Fix:* at minimum correct all three to say the created-public event has no consumer in this repo and
name the issue that owns it. Better: delete the event, its emission, its transport sizing and its ABI
pin — nothing off chain reads it.

**M5. INVARIANTS #22 is marked `[HOLDS]` but the host does not hold it.**
`solana/docs/INVARIANTS.md:191-192` — "version 3 is exactly 65 bytes: the id then the 32-byte key of
the account whose public leaf is proven". `solana/programs/zama-host/src/eip712.rs:134-137` accepts
**any** v3 payload of 33 bytes or more (`extra_data.get(1..33)?`), while the version-1 arm two lines
above does enforce `!= 33`. The connector enforces 65 strictly
(`kms-connector/crates/utils/src/types/solana_extra_data.rs:21,38`) and so does the relayer; the
host, which is where the invariant's binding chain is anchored, does not. Its own doc comment at
`eip712.rs:120` still says v3 "may append an MMR-proof tail" — the exact shape this change removed.
*Fix:* add `if extra_data.len() != 65 { return None; }` to the `Some(3)` arm, matching the v1 arm, and
rewrite the `:120` doc comment.

**M6. The corrupt-leaf-record branch lost its only tests with the deleted service.**
`http_server.rs:392-407` — the `inconsistent()` arm, taken when the stored leaf rows disagree with
`leaf_count` or the proof cannot be built, answering 502. Nothing tests it. The deleted service pinned
the same property four ways: `solana-proof-service/crates/solana-proof-service/src/proof.rs`
(`returns_corrupt_when_persisted_peaks_do_not_match_leaves`,
`returns_corrupt_when_equal_count_peaks_diverge`) and `src/http.rs`
(`handler_corrupt_cache_json_body`, `corrupt_store_maps_to_500_corrupt_cache_json_body`).
*Fix:* in `tests/solana_leaves_tests.rs`, delete one leaf row (or bump `leaf_count`) and assert the
`upstream_transient` 502.

**M7. The Yellowstone token-redaction guard died with the deleted service.**
`coprocessor/fhevm-engine/host-listener/src/solana_grpc_listener.rs:137-143` —
`SolanaGrpcListenerConfig` derives `Debug` over `x_token: Option<String>`. The deleted
`YellowstoneSourceConfig` had a hand-written redacting `Debug` pinned by
`debug_redacts_yellowstone_x_token`. The derive predates this change; what this change removes is the
only test of the property, and the listener is now the sole Yellowstone client, so the guard is
load-bearing and absent.
*Fix:* hand-write `Debug` emitting `[REDACTED]` for `x_token` and port the deleted test.

**M8. The vocabulary gates are blind to two trees the change now depends on, and a survivor proves
it.**
Three separate holes with one shape:
- `relayer/src` is in none of `dead-surface-check.sh`'s roots (`RUST_ROOTS`, `TS_ROOTS`,
  `SCRIPT_ROOTS`, `CORE_ROOTS`, `KMS_EXTRA_ROOTS`, lines 76-150), and the relayer's own teardown gate
  (`relayer/tests/solana_v0_teardown.rs:17-85`) lists no `subject`, `compute_subject`,
  `AccessEvidence` or `authorize_current`. Consequence: `relayer/src/host/acl_checker.rs:99-101`
  still documents live behaviour as "the request carries the encrypted value id and the subject …
  their authorization is membership in the encrypted value account" — the retired model, described as
  current, in a file this change edited. `relayer/src/readiness/checker.rs:139-140` repeats the
  membership claim. A second survivor sits in another unswept tree,
  `kms-connector/connector-db/migrations/20260812120000_user_decrypt_solana_request.sql:11`.
- `dead-surface-check.sh:521-524`: the comment says "No exception any more: the connector's
  `acl_value_key` went with the client-supplied proof it keyed", but line 524 still passes
  `--exclude-dir=utils` — a directory blanket, not a named exception. The symbol is alive at
  `kms-connector/crates/utils/src/types/extra_data.rs:229`, inside a test still named
  `solana_v3_mmr_proof_blob_returns_context_only`.
- `GLOSSARY.md:94` bans "proof service" and no `check_alias` entry implements it. The term survives at
  `kms-connector/crates/kms-worker/src/core/config.rs:174,180` ("the coprocessors' leaf-proof
  service").
*Fix:* add `relayer/src` (and `kms-connector/connector-db`) to the alias roots or record why not; add
the subject/evidence tokens to the relayer gate, split-spelled the way the kms gate does at
`solana_v0_teardown.rs:58,62`; replace `--exclude-dir=utils` with a path-anchored exception and rename
that test to `solana_v3_extra_data_returns_context_only` (its `acl_value_key` local →
`encrypted_value_account`); add a `proof service` alias entry and reword `config.rs:174,180` to
"leaf-proof endpoint"; correct the two relayer comments to "an allow leaf sealed on the write" and
"the request names the allowed key and the encrypted value account".

---

## Minor

**m1. Guardrail-table names that do not resolve, and one replay that asserts less than its siblings.**
- `shared_crate_address_matches_program_seeds` (record §2) exists nowhere. The substance is real but
  folded into `solana/programs/zama-host/src/state/encrypted_value.rs:109`, which asserts both the
  round-trip and `find_program_address(&decoded.seeds())` at `:129`. Its two sibling decoder tests
  (`host_config.rs:176`, `user_decryption_delegation.rs:100`) assert no address at all. *Fix:* split
  the address assertion into a test of that name, or drop the claim.
- `coprocessor/.../database/solana_leaves.rs:829` `replays_the_normative_leaf_vectors` checks leaves,
  `leaf_count` and peaks but never the vector's `proof` field — which both other implementations do
  verify (`zama-solana-acl/tests/leaf_vectors.rs:171`, `sdk/js-sdk/src/solana/proof.test.ts:110`).
  *Fix:* verify the committed proof against the reduced peaks.
- `solana/docs/TESTING_STRATEGY.md:63-64` claims "every handwritten TypeScript PDA seed matches its
  Rust counterpart"; `solana/scripts/check-pda-seeds.py:26-32` pins exactly one pair
  (`PENDING_BURN_SEED`). Uncovered: `sdk/js-sdk/src/solana/encryptedValueAccount.ts:66`,
  `actions/userDecryptionDelegation.ts:83`, `actions/revokePermits.ts:20`,
  `solana/demo-dapp/src/vault/joinBatch.ts:103`, `settleBatch.ts:203`. *Fix:* add the pairs, or narrow
  the sentence to the one pair.

**m2. Doc numbers that no longer match the code or the regenerated snapshots.**
Each checked against `git show HEAD:` of the source named:
- `DESIGN_DECISIONS.md:2215` — DD-046 (adopted) says the all-created-public heap wall is "21 steps";
  `cost-snapshots/fhe_execute_boundary.json` → `all_created_public.max_ok = 22`.
- `DESIGN_DECISIONS.md:2221` — DD-046 says `mature_updates_peaks_8/32/55` are "11, 6, 4"; the snapshot
  says 19 / 7 / 4, which INVARIANTS #54 already quotes correctly. The third profile is also named
  `mature_updates`, not `mature_updates_peaks_55`. DD-046 was not updated when this change regenerated
  the snapshot.
- `INVARIANTS.md:494` — `reduction_heavy` "host heap wall at 5 steps"; snapshot `max_ok = 4`,
  `first_fail = 5`, and the same sentence quotes `max_ok` for the three profiles before it.
- `zama-host/README.md:80` — "The bounded 16-output maximum fits one CPI"; the maximum is
  `MAX_FHE_EXECUTION_STEPS` = 32, asserted at `event_transport.rs:55-76` (2,133 bytes).
  `TESTING_STRATEGY.md:43` already says 32.
- `TESTING.md:13` — "379 named, filterable cases" for `operator_conformance`; the suite runs 328.
- `solana/scripts/README.md:13` — `sync-zama-host-idl.sh` writes "all four committed IDLs";
  `check_solana_abi.py` owns six. The two specimens were added by this change; the count was not.

**m3. Doc statements that contradict current behaviour or name things that do not exist.**
- `DESIGN_DECISIONS.md:927-928` — DD-025, `Status: resolved`, still describes "the proof-service
  confirmed RPC / Yellowstone path (… bounded RPC recovery)" as a current commitment consumer. The
  service is deleted and the listener deliberately has no RPC recovery.
- `EVM_PARITY.md:91`, `:140` and `FUTURE_DESIGN.md:56-57` describe user-decrypt as routed through a
  typed `userDecryptionRequestSolana` entrypoint. It does not exist: `Decryption.sol:765` is the
  host-generic `userDecryptionRequest(...)` overload, and `relayer/tests/solana_v0_teardown.rs:26`
  fails the build if the typed name reappears.
- `EVM_PARITY.md:98-99` — both rows claim the Solana entrypoint runs
  `_extractCtHandlesCheckConformanceHandleEntry`; it runs `_checkCtHandlesConformanceHostChain`
  (`Decryption.sol:813`) into `_checkCtHandlesBitBudget`.
- `EVM_PARITY.md:101`, `:211` — "Residual `native-v0` connector library/store code exists" / "dead
  scaffolding to retire". No such code exists anywhere. *Fix:* delete both rows.
- `EVM_PARITY.md:43` — `confidential-token/src/fhe.rs` is a directory, `.../src/fhe/`.
- `DESIGN_DECISIONS.md:356` — DD-010's RFC-035 note binds the disclosed account "to the mint's ACL
  domain"; `GLOSSARY.md:92` bans that sense and `EVM_PARITY.md:172` says "mint scope".
- `DESIGN_DECISIONS.md:602` cites `acl_storage_rationale.md`, in no tree;
  `kms-connector/config/kms-worker.toml:184` cites `acl_storage_spec.md` (also absent) and "the
  native-v0 Solana request format", directly above the Solana request block.
- `FUTURE_DESIGN.md:21` ("a deep-encrypted value account public-decrypt proof") and
  `TESTING_STRATEGY.md:17` ("foreign-encrypted value account proof") are broken sentences left by a
  global `lineage` → `encrypted value account` rename.
- `zama-host/README.md:78,80` — "one versioned Anchor CPI lifecycle **execution**": same rename
  artifact; the thing emitted is an *event*.
- `TESTING.md:41-43` — the Mollusk coverage list omits `host_admin_mollusk` and
  `user_decryption_delegation_mollusk`, both present.
- `DESIGN_DECISIONS.md:1513` — DD-037's consequences cite `event_budget.rs`, `MAX_CPI_EVAL_EVENTS`,
  `eval_event_capacity`, `should_emit_eval_events_as_cpi`; none survive. DD-037 is
  `Status: replaced by DD-038`, so this is historical narrative — lowest priority here, but every file
  and symbol name in it resolves to nothing.
- `DESIGN_DECISIONS.md:1778, :1829, :1952` — DD-041, DD-042, DD-043 carry no `Status:` line; every
  other entry DD-001..DD-048 does, and the header at `:7` defines the vocabulary.

**m4. Stale comments in live code describing the deleted model.**
- `solana/programs/zama-host/src/state/host_config.rs:41`,
  `src/instructions/set_grant_deny_list_enabled.rs:1` and `:8` — all three describe the flag as
  "deny-list checks/witnesses for persistent grant authorities". Per-grantee deny records and witness
  accounts are gone; the flag now gates the application deny record on `fhe_execute` (compute *and*
  every inline allow) and on `make_handle_public`. Keeping the field name to avoid the IDL/SDK cascade
  is on the record and fine; a fail-closed switch described at the wrong scope is not. *Fix:* "Enables
  the deny list: a denied application `(program, scope)` cannot compute, allow, or make a handle
  public."
- `kms-connector/crates/utils/src/types/extra_data.rs:46-49` — "Format (v3, Solana MMR proof) … Bytes
  33..: Solana MMR-proof blob (ignored)". v3 is now exactly `0x03 ‖ ctx ‖ encrypted value account`.
- `solana/programs/zama-host/src/state/user_decryption_delegation.rs:24` — "every value of that
  authority in every domain: the domain is not one of the PDA's seeds". The alias rule
  (`dead-surface-check.sh:544`) matches `acl domain` / `mint domain` / `.domain`, not bare "in every
  domain". *Fix:* "in every scope: the scope is not one of the PDA's seeds."
- `solana/scripts/dead-surface-check.sh:645` — the script's own prose still describes an encrypted
  value account as holding "handle, subject list and MMR". The file is excluded from every alias sweep
  by `:470`, so nothing catches it. *Fix:* "handle and MMR peaks".

**m5. `.github/workflows/solana-e2e.yml` header is stale in three places.**
`:29-32` still triggers on branch `elias/solana-proof-service-recovery`, a proof-service cutover base
this change makes moot. `:5-6` advertises "the fhe_execute operator wiring" among the scenarios —
`operators.scenario.test.ts` is deleted and the live operator matrix is gone, so the workflow should
not still claim it. `:11-13` states the created-public CPI event as load-bearing (M4).
*Fix:* delete the second `branches:` entry and the comment above it; rewrite the scenario list to the
six files in `test-suite/fhevm/e2e/scenarios/`; drop the event paragraph.

**m6. The two viewer handlers are one handler written twice.**
`solana/programs/confidential-token/src/instructions/allow_balance_viewers.rs:72-146` and
`total_supply_viewers.rs:64-140`: structurally identical ~75-line bodies —
`assert_token_value` → `deny_scope_record` → `uint64_operand` → `PersistentOutput::new` → one-step
`add(x, 0)` → `ExecutionAccountSet::for_execution` → `fhe::Execute` → `emit_cpi!` — of which a 25-line
run (the nine-field `fhe::Execute` literal, including the identical `hcu_block_meter` /
`hcu_trusted_app_record` `.as_ref().map(...)` pair) is byte-for-byte identical. Everything else
differs only by identifier.
*Fix:* one helper in `instructions/common.rs` — `rewrite_allowing(ctx_parts, value, id, authority,
allows) -> Result<[u8; 32]>` returning the new handle — leaving each handler its `assert_token_value`
and its `emit_cpi!`. (Note: both files are among the 24 currently being edited.)

**m7. The public-leaf proof build is written twice in TypeScript.**
`solana/demo-dapp/src/vault/settleBatch.ts:99-121` (`buildBurnedAmountPublicProof`) and
`test-suite/fhevm/src/solana/fhe-vertical.ts:133-156` (`publicLeafProof`) are the same algorithm —
`reconstructSolanaEncryptedValueAccount` over an expected history, element-wise peaks/leafCount
comparison against the live account, `mmrBuildProof` — with two different error strings.
*Fix:* move the generic form to the SDK beside `reconstructSolanaEncryptedValueAccount` /
`mmrBuildProof` (`sdk/js-sdk/src/solana/proof.ts`), and let `buildBurnedAmountPublicProof` become
`publicLeafProof(account, live, burnedAmountLeafHistory(...))`.

**m8. The canonical seed order is open-coded once.**
`solana/programs/zama-host/src/instructions/fhe_execute.rs:524-537` spells the five seeds by hand for
`create_pda_strict`, while `zama_solana_acl::encrypted_value_seeds`
(`solana/crates/zama-solana-acl/src/lib.rs:120-133`) exists precisely so "there is exactly one way to
build it". Every other site goes through it.
*Fix:* build from `encrypted_value_seeds(&declared.app.program.to_bytes(), …)` and push
`&[output_pda.bump]`.

**m9. `EncryptedValue::space` is a subtract-then-re-add round trip.**
`solana/programs/zama-host/src/state/encrypted_value.rs:43-45` returns
`zama_solana_acl::EncryptedValue::account_size(n) - 8`, and both call sites
(`instructions/encrypted_value.rs:108`, `instructions/fhe_execute.rs:515`) write `8 + …::space(n)`.
It is not used in an Anchor `space =` attribute.
*Fix:* delete `space`; call `account_size(n)` at both sites.

**m10. New `pub` items with no caller outside their own file, in the two trees the check does not
sweep.** `kms-connector/.../solana/proof.rs:40` `MAX_LEAVES_PER_READ`, `:102` `is_behind`, `:127`
`merge_outcomes`; `host-listener/src/http_server.rs:34` `MAX_LEAVES_PER_REQUEST`. The exclusion is a
recorded decision (`dead-surface-check.sh:782-787`), but this change adds ~1,000 lines of new export
surface into exactly those trees. *Fix:* drop `pub` on the four.

**m11. `.expect` in on-chain program code.**
`solana/programs/zama-host/src/instructions/fhe_execute/walk.rs:64-69` —
`.expect("rand steps consume the nonce before the walk")`. Unreachable, because preflight rejects a
rand step without the nonce, but it degrades a would-be program error into a panic abort in a program
that has a named variant for everything else.
*Fix:* have `rand_seed` take `&RandAnchor`; both call sites (`:237`, `:259`) are already inside the
rand arms, so the `Option` is destructured once where it is known present.

**m12. `BatchLifecycle`'s `'proving'` variant no longer means proving.**
`solana/demo-dapp/src/batchTypes.ts:27` kept the name after `proofReady` was dropped; it now only
means "dispatched, not yet settled", and the UI already renders it "Settlement in progress"
(`DepositJourney.tsx:24-25`). It also collides with `DepositStage.proving` (`deposit.ts:44`), a
genuine ZK-proof stage.
*Fix:* rename to `'dispatched'` at `batchTypes.ts:27`, `settlement.ts:84`, `DepositJourney.tsx:24,73`,
`JourneyPrimitives.tsx:40`, `RedeemJourney.tsx:125`, `useDemoController.ts:295,349`.

**m13. Proof-service DB-transactionality concerns with no counterpart.**
`store_postgres.rs` pinned `rollback_before_commit_then_successful_replay`,
`empty_block_atomic_checkpoint`, `proof_snapshot_isolation_sees_complete_old_or_new`,
`cross_slot_signature_reuse_halts_as_constraint_conflict` and
`same_slot_signature_index_conflict_halts`; `tests/solana_leaves_tests.rs` commits only happy paths.
Related and worth one test: `store_block_leaves` (`solana_leaves.rs:396-417`) inserts leaves with **no
`ON CONFLICT`**, which is safe only because `solana_grpc_source.rs:194-215` skips the already-applied
checkpoint block — tested at the source level, never against the DB. (The genuinely architectural
deletions — all of `rpc_recovery.rs`, `runner.rs`, `readiness.rs` — are documented decisions, not
silent losses; per-account `history_complete` is covered at `solana_leaves_tests.rs:279` and "listener
behind" moved to `solana_snapshot_architecture.rs:604`.)

**m14. Lost test cases with equivalents elsewhere** — recorded so the PR can say so rather than have
it found later. `zama-solana-acl/src/encrypted_value_account.rs` 12→5 (small-shape cases subsumed by
`realistic_sequence_indices_and_proofs` + the leaf vectors); `demo-dapp/src/vault/reads.test.ts` 8→2
(decoder negatives moved to `sdk/js-sdk/src/solana/encryptedValueAccount.test.ts:92-111` when the
hand-written decoder was replaced by the SDK's); `sdk/js-sdk/src/solana/proof.test.ts` 25→14 (four
negatives merged into one at `:191` — "rejects a tampered commitment (substitution)" has no direct
successor); `kms-connector/.../solana_extra_data.rs` 4→2 (the fixture gained `truncated-account` /
`unknown-version`, pinned at `solana_extra_data_byte_vectors.rs:95`). Every other lost name maps to
code the RFC deleted. Separately, `test-suite/fhevm/e2e/scenarios/operators.scenario.test.ts` took 8
live operator cases with it: the in-process replacements are real and documented
(`TESTING.md:13,15,21`), but nothing now runs operator semantics through the deployed Solana stack,
and the only real-TFHE-across-Solana test is `#[ignore]`d
(`coprocessor/fhevm-engine/tfhe-worker/src/tests/solana_vertical.rs:98`, pre-existing).

---

## Nit

**n1.** `solana/crates/zama-solana-permit/tests/signed_text_is_never_parsed.rs:51` uses
`"\nACL domains: ALL (permissive)"` as its tamper suffix; the line it mimics is now
`"Scopes: ALL (permissive)"` (`src/render.rs:28`).

**n2.** `confidential-token/src/instructions/allow_balance_viewers.rs` vs `total_supply_viewers.rs` —
both hold an `allow_*_viewers` handler plus a make-public handler; only one carries the verb. Rename
the second `allow_total_supply_viewers.rs`.

**n3.** The application pair is named by its scope half throughout: `AppScope`
(`zama-host/src/state/mod.rs:522`), `DenyScopeRecord`, `set_deny_scope`, `ScopeDenied`,
`FheExecuteMixedScopes`. `GLOSSARY.md:15` had to add an "`AppScope` in code" gloss to bridge the
mismatch with its own normative term. If the IDL churn is acceptable: `Application`,
`ApplicationDenyRecord`, `ApplicationDenied`, `FheExecuteMixedApplications`.

**n4.** `AppScope` and the five `*_address` helpers sit in a 955-line `state/mod.rs` while every
smaller account type has its own file. Move them to `state/app_scope.rs`.

**n5.** `test-suite/fhevm/src/solana/fhe-vertical.ts:133` `publicLeafProof` / `:159`
`buildPublicLeafProof` — a two-function split where each has one caller.

**n6.** One wire cap, two names: `MAX_LEAVES_PER_READ` (connector) and `MAX_LEAVES_PER_REQUEST`
(listener), both 64. If M1's golden fixture lands, pin the number there and reference it from both.

---

## Verified sound

- `bash solana/scripts/dead-surface-check.sh` — clean, all seven checks reporting.
- **No test skipped, weakened or deleted to get green.** Zero added `#[ignore]`, `.skip(`, `it.skip`,
  `describe.skip`, `xit(`, `todo!()`, `assert!(true)`, or `#[allow(dead_code)]` on a test helper
  anywhere in the diff; assertions were broadly strengthened (`expect_err` with stated reasons across
  the connector suite). No `TODO`/`FIXME`/`HACK`/"for now" added. The four suppressions added — two
  `#[allow(clippy::too_many_arguments)]` on test helpers, one `eslint-disable` for a fixture's
  snake_case fields, one `cfg_attr(not(feature = "solana-grpc"), allow(dead_code))` — each carry a
  reason.
- **Every committed fixture has a pinning test**, each behind an explicit update env var:
  `leaves_v1.json` → `committed_vectors_match_the_generator` (`ZAMA_UPDATE_LEAF_VECTORS`) +
  `scripts/update-leaf-vectors.sh`; `permit_v1.json` → `zama-solana-permit/tests/vectors.rs:767`;
  `connector_auth_v1.json` → `solana_connector_auth_vectors.rs:1856`;
  `permit_invalidation_account_v1.json` → `permit_invalidation_mollusk.rs:605`;
  `openapi/solana_leaf_proofs.json` → `http_server.rs:466`; cost snapshots →
  `test-kit/src/cost_snapshot.rs:59`; `host-listener/idl/*.json` → `check_solana_abi.py` via
  `check-zama-host-idl.sh`, run in CI at `solana-tests.yml:292`, plus `execution_contracts.rs:409`;
  `extra_data_v1.json` → `solana_extra_data_byte_vectors.rs:62`; `relayer_envelope_v1.json` → both
  sides of the seam (`relayer/tests/solana_envelope_fixture.rs:152`, `sdk/.../request.test.ts:63`);
  `transaction_decoding_v1.json` → `solana_grpc_listener.rs:1210`.
- **Leaf vectors are genuinely replayed by three independent implementations**: the ACL crate
  (`zama-solana-acl/tests/leaf_vectors.rs`), the listener's reduce rule (`solana_leaves.rs:829`,
  modulo m1) and the SDK (`proof.test.ts:98-131`), all through one shared `leaf_vectors.rs` schema.
- **The record's Mollusk guardrails all exist and assert what is claimed**: rand — two identical
  batches in one slot yield different seeds and the nonce moves 7→9 (`host_mollusk.rs:3689`), rand
  without the nonce account is rejected with `FheExecuteRandNonceMissing` (`:3647`); compute
  negatives — wallet authority (`:480`), wrong-program PDA (`:533`), unsigned authority (`:2997`);
  deny list — a denied `(program, scope)` fails both `fhe_execute` and `make_handle_public` with
  `ScopeDenied` while a sibling scope of the same program passes both (`:857`), and a missing record
  fails closed (`:948`); snapshot architecture — the proof read is a counted, planned read, pinned at
  1 in step, 2 when behind, never 3 (`solana_snapshot_architecture.rs:604`).
- No proof-service surface outside deliberate records: no env template, compose file, override group,
  Dockerfile, CI job, commitlint scope, `PROOF_SERVICE*` / `proofServiceUrl` / port-8088 reference, or
  `solana-proof-service/` tree survives; the only live hit is the `solana-e2e.yml` branch trigger
  (m5). References to every deleted module (`solana_v2_fetcher`, `SolanaAclVerifier`,
  `solana_encrypted_value_acl`, `rpcEvidence`, `proofService.ts`, `check_proof_store_idl.py`,
  `operators.scenario`, `settlement.test`) exist only inside the three teardown gates.
- `zama-solana-request` is renamed end to end (`allowed_key`, `encrypted_value_account`); no `subject`
  wire field survives, contrary to the worklog's Step 5 deferral note.
- Docs cross-references: DD-001..DD-048 all defined and every `DD-0xx` citation resolves;
  INVARIANTS #1..#61 all defined and every `#N` / `INV #N` citation resolves — nothing dangling in
  either direction. Every other backticked path and quoted symbol/test name in the changed docs
  resolves, including all six scenario files and every `fhe_execute_boundary/*` profile except the
  misnamed `mature_updates_peaks_55`. Paths written as retired correctly do not exist.
- Docs numbers that do check out: `MAX_ALLOWED_SCOPES = 7` / `PERMIT_MAX_SCOPES = 7`,
  `MAX_REQUEST_HANDLES = 33` / `MAX_SOLANA_USER_DECRYPT_HANDLES = 33`, `MAX_MMR_PEAKS = 64`,
  `MAX_FHE_EXECUTION_STEPS = 32`, `MAX_PERSISTENT_CREATES = 20`, `HostConfig::SPACE = 317`,
  `account_size = 181 + 32·peaks` capped at 2229 (asserted at `zama-solana-acl/src/lib.rs:276-280`),
  extra_data v1 = 33 bytes, the 2,133-byte CPI event, the 513-slot ALT cooldown. The heap-frontier
  claims reproduce exactly: the grid is 122 cells, `ADMITTED_FRONTIER_SHAPES = 57`, and all five
  totals INV #54 quotes come back identical from `print_build_frontier_grid`. INV #61's
  `allow_heavy_public_creates` wall 17 / first-fail 18 / band 18-20 matches the snapshot, as does
  INV #54's 19 / 7 / 4. `MAX_REQUEST_HANDLES = 33` does bound the proof batch below the 64 cap, as
  `proof.rs:37-39` claims.
- Client-side MMR proof building has exactly one implementation (`sdk/js-sdk/src/solana/proof.ts`);
  demo-dapp and test-suite both call into it.
- New connector error enums (`ProofReadError`, `SnapshotError`, `RequestFormError`,
  `DeploymentIdentityError`) — every variant is constructed and matched in production code.
- CI runs the new work: `coprocessor-cargo-tests.yml:85-89` builds `host-listener` with
  `solana-grpc,solana-reconstruct`, so the new Postgres and HTTP suites execute; `solana-tests.yml`'s
  `dead-surface` path filter tracks the script's roots (check 7 enforces the correspondence and
  passes) and its `--self-test` fixture-parity gate is wired at `solana-tests.yml:131`;
  `cargo fmt -p host-listener --check` is clean.
- `.sqlx` regenerated; no orphaned Solana query files remain in `coprocessor/fhevm-engine/.sqlx`.
- Token error ordinals shifted by the deleted variants: nothing off chain pins a numeric token error
  code (checked SDK, demo-dapp, test-suite).
- Comment quality in the rewritten host paths is high — `fhe_execute/preflight.rs`,
  `handle_binding.rs`, `proof.rs` and `solana_leaves.rs` module docs explain the rule and the reason
  rather than restating the code, and the cross-program-composition rule Step 4 introduced is written
  out where it is enforced.

</details>


Eikix (commented):
[AUTOMATED]

Reviewed `056619a0d8de768036b6d0b898febc21f53900d8` against `feature/solana` at `b8b572d23a8ced7ac7c8e0ed5f1b780544149754`, read-only. **Changes recommended before merge:** the direction is coherent, but proof availability and cross-application amount handling have concrete defects. Four inline findings describe the failure cases and fixes.

The goal is to separate authority-signed compute from leaf-authorized decrypt, remove the persistent subject list/current-decrypt branch, and make independently operated coprocessors supply proofs to the Connector. The coordinated account, permit, SDK and service changes mostly achieve that. In particular, the final-head host checks every touched application's deny record, including additional signing authorities; the earlier reported host bypass is fixed. The remaining token-side witness omission is a different integration defect.

### Correctness findings

1. **P1 — One stalled proof endpoint blocks the whole fanout.** `join_all` waits for every endpoint, while the production client has no request deadline. An endpoint that accepts a connection but never completes its response prevents a healthy endpoint's proof from being used.
2. **P1 — A replacement indexer has no supported historical bootstrap.** An empty DB starts at the live tip. Accounts first seen through updates become `history_complete=false` and never accumulate usable proofs. The deleted service's bounded RPC bootstrap/gap recovery was not carried over. Yellowstone replay alone also cannot recover an outage beyond the provider's retained history.
3. **P2 — Proof ranking can discard the only valid proof.** The reader selects the largest reported `leaf_count` before verification. A faulty response, or an honest response ahead of the deciding snapshot, can shadow another coprocessor's valid proof on every attempt. Retain candidates until verification; counts should classify failures after no candidate verifies. A failed refresh also currently discards an earlier potentially valid proof.
4. **P2 — Foreign-app amounts cannot be transferred or burned with the deny list enabled.** The token helpers request the token application's deny record but omit the stored amount's application. Supplying the missing record is rejected as an unexpected remaining account. This affects a legitimate application PDA signing a from-value transfer/burn, not merely a synthetic wallet-authority fixture.

### Quality and verification gaps

- **Retain the distinct live operator shapes.** Deleting `operators.scenario.test.ts` removes the deliberately small integration matrix retained by [#1696](https://github.com/zama-ai/fhevm-internal/issues/1696). The specimen/token scenarios cover some shapes, but cast, bounded randomness, sum and `isIn` no longer traverse the live validator → listener → worker → decrypt path. In-process semantic tests do not replace that wiring evidence, especially while reconstruction and random-seed plumbing change. Port those shapes behind a PDA-owned specimen rather than restoring wallet-owned values. The PR body discloses the loss accurately.
- **The proof-cost issue is larger than duplicate account loading.** Each query loads the account's entire leaf history, searches it, rebuilds all peaks and builds a proof. Multiple queries repeat that work: `O(queries × account_history)`. Persisted MMR nodes/checkpoints and immutable proof caching remain explicitly required by [#1971](https://github.com/zama-ai/fhevm-internal/issues/1971). This cost was already present in the old service; it is unfinished retained work, not a newly discovered algorithmic regression. API-key authentication does not bound the history of a legitimate busy account.
- **Use reachable authorization fixtures.** Some from-value tests still inject wallet-authority encrypted accounts. Such accounts can no longer be created through the host. Keep decoder/adversarial fixtures where useful, but prove the valid cross-program path with a real program PDA and deny-list-enabled execution.

### RFC 034, 035 and 036 review

All three are **open proposals**, not specifications merged on `tech-spec/main`: [034 / #506](https://github.com/zama-ai/tech-spec/pull/506) at `7bda2d47`, [035 / #448](https://github.com/zama-ai/tech-spec/pull/448) at `e092c8fa`, and [036 / #511](https://github.com/zama-ai/tech-spec/pull/511) at `e1a7a715`. These comments concern the RFCs; they do not request EVM or KMS redesign inside this PR.

**RFC 034:** Owner-keyed e-slots plus transient forwarding give a clear on-chain compute boundary, and host-written peaks preserve proof verifiability. The rollout still needs an explicit historical cutover rule: indexing existing events before the host starts writing peaks does not put those past allows into the host's new tree. When the Connector becomes proof-only, what authorizes pre-activation historical handles? Specify a legacy fallback/cutoff or a verifiable migration. Also reconcile keeping unmigrated deployed contracts on the legacy compute map with the later step that stops `allow` from writing that map. The retirement condition and actual supported old/new combinations must be explicit. The drafts share an MMR model, but their leaf fields differ; “same leaf encoding” needs an exact tagged byte layout and cross-chain vectors before being treated as byte-level compatibility.

**RFC 035:** It describes the central model well, but still specifies three identity fields with no `scope`, program-only permit filtering, a never-growing account with all peak space reserved, and program-PDA seeds supplied on every write. This PR uses four fields, `(program, scope)` permits, bounded realloc, and immutable create-time program verification followed by canonical-account/authority checks on update. The [September 7 Slack decision](https://zama-ai.slack.com/archives/C0AUWMS51EX/p1788781434894489?thread_ts=1788436784.351159) explicitly brings scope back for metering and the deny list, so this is intentional evolution that needs a coordinated RFC amendment. Record the nonce singleton and its writable-account contention, and distinguish the declared scope from the proven program. The account-ownership check for scope is explicitly deferred to [#1994](https://github.com/zama-ai/fhevm-internal/issues/1994) before the feature branch merges to main; it is not a newly raised blocker here.

**RFC 036:** The separation of reusable permit, unsigned entries, and response linker is sound, as is proving the delegator's leaf instead of the delegate's. Its normative bytes are stale: v1 text, `allowed_programs` (10), entry `key`, request byte `0x01`, and scope-less PDA derivation disagree with this PR's v2 text, `allowed_scopes` (7 pairs), `allowed_key`, request byte `0x02`, and scoped PDA. Pin the amended canonical text and vectors together. Its one-observation explanation also omits the implementation's deliberate first-read pause exception; including HostConfig in the second read changes the worst-case budget from `3N+1` to `3N+2`. State the selected behavior explicitly. Genesis-derived cluster IDs remain proposed work under [#1880](https://github.com/zama-ai/fhevm-internal/issues/1880), not an implemented guarantee of this PR. Finally, the threat model calls RPC nodes untrusted, but the Connector trusts `getMultipleAccounts` owner/data/slot without a chain-state authenticity proof. This is a security-model gap that must be resolved before accepting the RFC. Either state the RPC trust assumption or specify the validation mechanism: an MMR proof checked against RPC-fabricated peaks does not authenticate chain state.

### Accordance with the issue history

Searched **all open and closed Eikix-authored `fhevm-internal` issues mentioning Solana**: 185 results, cross-checked against the paginated author inventory. Screened all for relevance and followed current requirements and supersession links; this is not a claim that all 185 historical proposals should be implemented by this PR.

- **#1963–#1969:** the central one-path authorization, shared leaf hashing, raw PDA derivation and compute ownership direction is implemented. Reconcile the scope/create-only-verification/batched-allow amendments with the issue bodies and RFCs. The proof-selection finding prevents treating #1967's race/availability acceptance as complete.
- **#1971:** folding the service is decided, but bootstrap/gap recovery, persisted deployment identity checks, and node-backed proof construction have not all survived. Keep these obligations visible; do not close the issue based on service deletion.
- **#1891:** owner-signed compute and the one-execution join receipt are retained; subject-list preservation is superseded by the later leaf-only decision, not a reason to restore subjects.
- **#1908, #1904:** pause/revocation plumbing and canonical quit destination protections are present. Their old issue descriptions should not be repeated as fresh findings here.
- **#1970, #1972, #1973, #1949:** latency measurement, remaining listener identity/reconstruction work, document reconciliation and production-readiness acceptance need their own evidence. A green localnet run does not establish the public-devnet/governance outcomes in [KR7.2](https://github.com/zama-ai/planning-blockchain/issues/1252), nor the RFC-merge outcome in [KR7.3](https://github.com/zama-ai/planning-blockchain/issues/1253).
- **#1985, #1981, #1746, #1692, #1710, #1634, #1994:** deployment, freeze policy, threshold KMS, delegated spend/holds, governance and scope ownership remain separate work. Older relayer-owned/client-carried proofs, permanent subjects and earlier compute-identity proposals are superseded, not compatibility requirements to reintroduce.

### Validation

Locally ran shared ACL tests and vectors (**30 passed**), host/permit tests and doctests (**229 passed**), and the host-listener reconstruction filter (**13 passed**). GitHub reports successful Solana build/tests, Solana e2e scenarios, host-listener, Connector and JS SDK checks for the exact reviewed SHA. The review did not rerun a live stack or database-backed integration suite. The new failure cases above need regression tests; the existing green jobs do not exercise them.

kms-connector/crates/kms-worker/src/core/solana/proof.rs
```diff
@@ -0,0 +1,560 @@
...
+        queries: &[LeafQuery],
+    ) -> Result<Vec<LeafProofOutcome>, ProofReadError> {
+        if queries.len() > MAX_LEAVES_PER_READ {
+            return Err(ProofReadError::TooManyQueries {
+                count: queries.len(),
+            });
+        }
+        let body = serde_json::to_vec(&leaf_proof_request_body(queries))
+            .map_err(|error| unavailable(format!("request body does not serialize: {error}")))?;
+        let answers = join_all(
```
Eikix, line 350: [AUTOMATED]

[P1] Bound every coprocessor request before joining the fanout

One configured endpoint can prevent all healthy endpoints from serving proofs: `join_all` waits for every `read_from`, which awaits both response headers and the body without a deadline. Production constructs `reqwest::Client::new()` (`kms_worker.rs:334-336`); the pinned reqwest 0.12.24 defaults both total and read timeouts to `None`, and event processing supplies no outer deadline. An endpoint that accepts a connection but never finishes its response therefore hangs authorization indefinitely even when another returned a valid proof. Add a bounded per-endpoint deadline that covers headers/body and retain successful peer responses; test a stalled endpoint alongside a healthy one.

coprocessor/fhevm-engine/host-listener/src/bin/solana_host_listener.rs
```diff
@@ -102,6 +116,23 @@ async fn main() -> Result<()> {
...
+            info!(
+                slot = checkpoint.slot,
+                "resuming after the recorded checkpoint"
+            );
+            StartPosition::Resume(BlockCheckpoint {
+                slot: checkpoint.slot,
+                block_hash: checkpoint.block_hash,
+            })
+        }
+        None => StartPosition::Tip,
```
Eikix: [AUTOMATED]

[P1] Restore a historical bootstrap and gap-recovery path

A fresh/replacement database always starts at the live tip, so it cannot reconstruct permissions for existing encrypted values. When an account is first observed through an update, `database/solana_leaves.rs` records `history_complete=false` and subsequent updates skip its leaves; the proof API then returns `HistoryIncomplete`. There is no supported rebuild/bootstrap path here, and resuming via Yellowstone alone fails when the checkpoint is outside provider retention. This replaces the deleted service's bounded RPC bootstrap/recovery rather than preserving it, although #1971 explicitly retains that obligation. Restore ordered historical catch-up from a configured activation point, with gap recovery and readiness withheld until history is complete; cover starting against an already-active host.

kms-connector/crates/kms-worker/src/core/solana/proof.rs
```diff
@@ -0,0 +1,560 @@
...
+            Self::HistoryIncomplete => 0,
+        };
+        (tier, self.leaf_count().unwrap_or(0))
+    }
+}
+
+/// Picks the better of two answers to one query. Stated once and used for both merges — across
+/// coprocessors and across the retry — so the two cannot prefer differently.
+fn merge_outcomes(current: LeafProofOutcome, other: LeafProofOutcome) -> LeafProofOutcome {
+    if other.rank() > current.rank() {
```
Eikix: [AUTOMATED]

[P2] Verify proof candidates before discarding peer responses

`rank()` prefers `Found` with the largest responder-supplied `leaf_count`, and this merge discards the other candidate before handle binding verifies it. A faulty peer can return an invalid proof with `leaf_count = u64::MAX` and repeatedly shadow a healthy peer's valid proof; an honest peer ahead of the deciding snapshot can also replace a snapshot-compatible proof. Retrying the same ranking does not recover availability. Preserve candidates until a proof verifies against the deciding peaks, then use counts only to classify failure when none verifies. Also retain the first usable response if the retry read fails instead of discarding it through `?`. Add mixed valid/invalid and ahead/compatible peer tests.

solana/programs/confidential-token/src/instructions/common.rs
```diff
@@ -273,47 +333,67 @@ fn compute_transfer_handles<'info>(
...
             payer: accounts.payer,
             event_authority: accounts.zama_event_authority,
             zama_program: accounts.zama_program,
             host_config: accounts.host_config,
-            deny_subject_records: accounts.deny_subject_records,
-            compute_authority,
+            deny_scope_records: fhe::deny_scope_records(
+                accounts.host_config,
+                accounts.remaining_accounts,
+                std::iter::once(token_app(mint_key))
```
Eikix, line 395: [AUTOMATED]

[P2] Include the stored amount application in deny-record plumbing

For a from-value transfer, the encrypted amount can belong to a different application whose PDA legitimately signs the token CPI. The execution then touches that stored amount's application, but this list includes only the token application and optional receipt application. With the deny list enabled, passing just these records reaches host preflight and fails `DenyRecordMissing`; supplying the foreign application's record is rejected by `deny_scope_records` as an unexpected remaining account. Burn has the same omission at `confidential_burn.rs:508-512`. Include and deduplicate the stored amount's application in both paths, and cover an actual program-PDA-owned foreign amount with allowed and denied records.

KitHat:
Two small things on the host / connector side, neither about correctness today.

1. Rand nonce per application

["rand-nonce"] is one writable account for every Rand / RandBounded execution of every program, so all rand executions in a slot serialize on it — the argument RFC 035 makes against a global peaks account. The seed already binds (program, scope); a nonce at ["rand-nonce", program, scope], lazily created like the HCU meter, keeps (program, scope, nonce) unique and restores parallelism between applications. Nothing in the repo uses rand yet, so the change is cheap now and expensive later. Deliberate choice, or the smallest step from the persistent-write anchor?

2. Two .expect in the connector's authorization path (automated finding)

- kms-worker/src/core/solana/pipeline.rs:312 — batch.position(&query).expect("every entry's query was planned into the batch")
- kms-worker/src/core/solana/proof.rs:309 — serde_json::to_value(LeafProofRequestWire { leaves }).expect("plain strings serialize")

Both invariants hold today by construction — the batch is built from the same entries two loops earlier, and the wire struct is strings only. Neither is enforced by a type, so a later refactor (dedup rule change, an extra field on the wire struct) can break them silently and turn what should be a rejected request into a panic in the worker. Same reasoning the fix commit applied to walk.rs; both callers already return AuthorizationFailure / ProofReadError, so an ok_or into a terminal "planning defect" variant costs one line each.

Eikix:
> Two small things on the host / connector side, neither about correctness today.
> 
> 1. Rand nonce per application
> 
> ["rand-nonce"] is one writable account for every Rand / RandBounded execution of every program, so all rand executions in a slot serialize on it — the argument RFC 035 makes against a global peaks account. The seed already binds (program, scope); a nonce at ["rand-nonce", program, scope], lazily created like the HCU meter, keeps (program, scope, nonce) unique and restores parallelism between applications. Nothing in the repo uses rand yet, so the change is cheap now and expensive later. Deliberate choice, or the smallest step from the persistent-write anchor?
> 
> 2. Two .expect in the connector's authorization path (automated finding)
> 
> * kms-worker/src/core/solana/pipeline.rs:312 — batch.position(&query).expect("every entry's query was planned into the batch")
> * kms-worker/src/core/solana/proof.rs:309 — serde_json::to_value(LeafProofRequestWire { leaves }).expect("plain strings serialize")
> 
> Both invariants hold today by construction — the batch is built from the same entries two loops earlier, and the wire struct is strings only. Neither is enforced by a type, so a later refactor (dedup rule change, an extra field on the wire struct) can break them silently and turn what should be a rejected request into a panic in the worker. Same reasoning the fix commit applied to walk.rs; both callers already return AuthorizationFailure / ProofReadError, so an ok_or into a terminal "planning defect" variant costs one line each.

1/ I'd rather do a global rand account in the sense that rand calls are very rare and as such they're ok to be on a hot account.

2/ Interesting let me check

PR #3837, opened 2026-09-06
feat(relayer): return every user-decrypt share

manoranjith:
Implements [#1737][issue].

Port of [#3481][src], which landed the same feature on `release/0.13.x`. Re-implemented rather than cherry-picked: 18 of its 33 files conflict here, mainly because the share query has since grown the `queue_position` and `tx_queue_size` subqueries from #3649.

`user_decrypt_additional_shares` and `user_decrypt_additional_shares_timeout_secs` are required config with no default, so gitops values files need both before this deploys.

Based on `mano/relayer/ct-commits-on-ha-v2` (#3650), not `main`. Locally: 268 unit, 429 integration, clippy clean.

[issue]: https://github.com/zama-ai/fhevm-internal/issues/1737
[src]: https://github.com/zama-ai/fhevm/pull/3481


nboisde (dismissed):

nboisde (approved):

nboisde:
@mergifyio queue

PR #3844, opened 2026-09-07
feat(host-contracts): improve scalar operand handling

jatZama:
Refs https://github.com/zama-ai/fhevm-internal/issues/1947


host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -823,6 +825,16 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
         emit VerifyInput(msg.sender, inputHandle, userAddress, inputProof, inputType, result);
     }
 
+    /**
+     * @notice Asserts the type of an initialized ciphertext handle.
+     * @dev This checks metadata only; it does not verify a proof or grant ACL permissions.
+     * @param handle Ciphertext handle. Zero is an uninitialized sentinel, not a ciphertext.
+     * @param expectedType Expected FHE type.
+     */
+    function checkType(bytes32 handle, FheType expectedType) external view virtual {
```
isaacdecoded: nit: make this method naming a bit more self-explanatory, e.g., `checkHandleType`, `checkCtHandleType`.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -823,6 +825,16 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
+        if (handle == bytes32(0) || _typeOf(handle) != expectedType) revert InvalidType();
```
isaacdecoded, line 835: nit: unrelated to this PR, but maybe also an opportunity to make the error more self-explanatory, e.g., `InvalidHandleType`, `InvalidCtHandleType`

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -906,16 +918,19 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
+        uint256 maxValue;
+        if (scalarType == FheType.Bool) maxValue = 1;
+        else if (scalarType == FheType.Uint8) maxValue = type(uint8).max;
+        else if (scalarType == FheType.Uint16) maxValue = type(uint16).max;
+        else if (scalarType == FheType.Uint32) maxValue = type(uint32).max;
+        else if (scalarType == FheType.Uint64) maxValue = type(uint64).max;
+        else if (scalarType == FheType.Uint128) maxValue = type(uint128).max;
+        else if (scalarType == FheType.Uint160) maxValue = type(uint160).max;
+        else if (scalarType == FheType.Uint256) return;
+        else revert UnsupportedType();
```
isaacdecoded, line 931: nit: also unrelated to this PR - we can try to make the error clearer like `UnsupportedFheType`

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -906,16 +918,19 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
+    function _checkScalarRange(uint256 scalar, FheType scalarType) internal pure virtual {
+        uint256 maxValue;
+        if (scalarType == FheType.Bool) maxValue = 1;
+        else if (scalarType == FheType.Uint8) maxValue = type(uint8).max;
+        else if (scalarType == FheType.Uint16) maxValue = type(uint16).max;
+        else if (scalarType == FheType.Uint32) maxValue = type(uint32).max;
+        else if (scalarType == FheType.Uint64) maxValue = type(uint64).max;
+        else if (scalarType == FheType.Uint128) maxValue = type(uint128).max;
+        else if (scalarType == FheType.Uint160) maxValue = type(uint160).max;
+        else if (scalarType == FheType.Uint256) return;
```
isaacdecoded, line 930: Question: why for `FheType.Uint256` do we just return here?

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -953,6 +968,10 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
             FheType rhsType = _typeOf(rhs);
             FheType lhsType = _typeOf(lhs);
             if (lhsType != rhsType) revert IncompatibleTypes();
+        } else {
+            // Comparisons return Bool, but the scalar belongs to the lhs domain.
+            _checkScalarRange(uint256(rhs), _typeOf(lhs));
+            if ((op == Operators.fheDiv || op == Operators.fheRem) && rhs == bytes32(0)) revert DivisionByZero();
```
isaacdecoded: nit: maybe for better readability we can break it down like:

```
bool isFheDivOrRemOperator = op == Operators.fheDiv || op == Operators.fheRem;

if (isFheDivOrRemOperator && rhs == bytes32(0) {
   revert DivisionByZero();
}
```

library-solidity/lib/FHE.sol
```diff
@@ -91,5538 +91,3787 @@ library FHE {
         return ebool.unwrap(v) != 0;
     }
 
+    /**
+     * @dev Validates the FHE type, or replaces an uninitialized handle with a correctly typed encrypted zero.
+     */
+    function _validate(ebool value) private returns (ebool) {
```
isaacdecoded: nit: same suggestion for the method naming i.e., consider self-explanatory name like `_getValidatedFheType` since this method returns the value, not only validates it.

isaacdecoded (approved):
LGTM! 👍 I just realized that some NatSpec docs from the FHEVMExecutor contract may be improved, e.g., what [LHS and RHS](https://github.com/zama-ai/fhevm/blob/a4e730546de2f5af9971f607722b69e346261296/host-contracts/contracts/FHEVMExecutor.sol#L186-L187) mean exactly. Maybe we can create an issue to review all host contract NatSpecs and improve them if possible.

library-solidity/codegen/src/templateFHEDotSol.ts
```diff
@@ -452,15 +441,10 @@ function handleSolidityTFHEShiftOperator(fheType: AdjustedFheType, operator: Ope
 }
 
 function checkInitialized(varname: string, type: string) {
```
obatirou: nit: could be renamed, because now the type is validated not only a check of initialization

obatirou (approved):

host-contracts/contracts/FHEVMExecutor.sol
jatZama: Because there is nothing to check here, hence Uint256 is allowed to take any value a bytes32 could (the whole range is accepted).

host-contracts/contracts/FHEVMExecutor.sol
jatZama: The InvalidType name already existed previously, are we sure your renaming suggestion wont break other components? I think the hotfix should not include this kind of breaking change to void unforeseen consequences, wdyt?

host-contracts/contracts/FHEVMExecutor.sol
jatZama: ok will do

host-contracts/contracts/FHEVMExecutor.sol
jatZama: Same as previously, wont do since this is potentially a breaking change, to avoid disruption.

host-contracts/contracts/FHEVMExecutor.sol
jatZama: done in https://github.com/zama-ai/fhevm/pull/3844/changes/7a1c0c5819bb10c3b0a02d46058ed65a5676828c

host-contracts/contracts/FHEVMExecutor.sol
jatZama: done in e9c516313ae0ca3e4610834010219dbb32e9fa8a

library-solidity/codegen/src/templateFHEDotSol.ts
jatZama: fixed in 29e8ffef4cd507cae5d1e4065e2d6720b058c4d9

library-solidity/lib/FHE.sol
jatZama: fixed in 29e8ffef4cd507cae5d1e4065e2d6720b058c4d9

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -906,16 +918,19 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
...
-        if (scalarType == FheType.Uint128) return uint128(scalarUint) == 0;
-
-        revert UnsupportedType();
+    function _checkScalarRange(uint256 scalar, FheType scalarType) internal pure virtual {
+        uint256 maxValue;
+        if (scalarType == FheType.Bool) maxValue = 1;
+        else if (scalarType == FheType.Uint8) maxValue = type(uint8).max;
+        else if (scalarType == FheType.Uint16) maxValue = type(uint16).max;
+        else if (scalarType == FheType.Uint32) maxValue = type(uint32).max;
+        else if (scalarType == FheType.Uint64) maxValue = type(uint64).max;
```
0xalexbel, line 927: Wouldn't it make sense to check `FheType.Uint64` first to minimize a little bit of gas (not much) in the preferred scenarios. FheType.Uint64 is the most popular type with FheType.Bool ? It's probably not very significant.

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -953,6 +968,14 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
             FheType rhsType = _typeOf(rhs);
             FheType lhsType = _typeOf(lhs);
```
0xalexbel: Maybe put  `FheType lhsType = _typeOf(lhs);` at the top as it is used in the 2 branches. Maybe it can minimize a little bit of byte code size ... Probably not very significant

0xalexbel (approved):
I left a few remarks. Nothing very significant.
LGTM! on my side 👑 !

host-contracts/contracts/FHEVMExecutor.sol
jatZama: Not sure it is worth the decreased readability. Each additional EQ opcode is just 3 gas anyways.

host-contracts/contracts/FHEVMExecutor.sol
jatZama: I am not sure how this would decrease bytecode size? Also solc optimizer might already do some reordering automatically here if needed?

host-contracts/contracts/FHEVMExecutor.sol
0xalexbel: sure! not very significant

host-contracts/contracts/FHEVMExecutor.sol
0xalexbel: 1 variable + 1 function call vs 1 variable + 1 function call + one inlined function call.

> Also solc optimizer might already do some reordering automatically here if needed?

To be perfectly honest this would require a forge compilation test to measure the difference in bytecode size. Again, as I mentioned in my remark, it's probably minimal - plus the compiler can optimize of course.

jatZama: You are right good catch, I fixed it https://github.com/zama-ai/fhevm/pull/3844/changes/e8a91fcc590ddcc8336ae0e26e4171a39662f43e and actually solc optimizer is so poor (this is known fact to reduce attack surface) that it did not automatically optimize its placement: according to codex `With the current deployment settings, runtime bytecode dropped 21,827 → 21,670 bytes, saving 157 bytes. Remaining headroom: 2,906 bytes (11.8%).`

obatirou (approved):

library-solidity/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@fhevm/solidity",
   "description": "A Solidity library for interacting with fhevm protocol",
-  "version": "0.13.3",
+  "version": "0.13.4",
```
tawadaa, line 4: Should we release this one manually after merge?

nboisde: Yes !

tawadaa (approved):

tawadaa (approved):

PR #3878, opened 2026-09-09
feat(coprocessor): host-consumer, recover autonomously after drift

rudy-6-4:


rudy-6-4:
E2E test https://github.com/zama-ai/fhevm/actions/runs/34351034275/job/102464081916#step:22:328

Eikix (commented):

Non-blocking: could we preserve ordinary restart continuity?

If the consumer processes through block 100, crashes, and restarts at 111, the new UUID creates a fresh queue. Blocks 101–110 remain in the old queue, and without a drift signal or manual catchup flags, no replay is requested. Previously, the stable consumer ID resumed that backlog.

Could we retain a resumable subscription or automatically replay the missing interval? This affects normal operation even if the old drift-recovery mechanism is retired soon. This scenario was traced through the code, not reproduced at runtime.

Also, were the consumer-only recovery tests run elsewhere? The inspected CI run skipped the broker-backed recovery tests and E2E. Given the planned retirement, additional validation should match whether this PR will actually be deployed.

rudy-6-4:
> 
> Non-blocking: could we preserve ordinary restart continuity?
> 
> If the consumer processes through block 100, crashes, and restarts at 111, the new UUID creates a fresh queue. Blocks 101–110 remain in the old queue, and without a drift signal or manual catchup flags, no replay is requested. Previously, the stable consumer ID resumed that backlog.
> 
> Could we retain a resumable subscription or automatically replay the missing interval? This affects normal operation even if the old drift-recovery mechanism is retired soon. This scenario was traced through the code, not reproduced at runtime.
> 
> Also, were the consumer-only recovery tests run elsewhere? The inspected CI run skipped the broker-backed recovery tests and E2E. Given the planned retirement, additional validation should match whether this PR will actually be deployed.

Ho yes the UUID was for being sure of not reusing a past queue for drift but it breaks the intent of the persistent queue on restart. I am looking at if it's still correct (depend on the auto catchup conf) and  how  it can be improved.

coprocessor/fhevm-engine/host-listener/src/consumer/drift_recovery.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub fn resolve(
+        &mut self,
+        args: &ManualCatchupArgs,
+        live: u64,
+    ) -> Result<Option<(u64, u64)>> {
+        if self.manual_range.is_none() && args.enabled() {
+            self.manual_range = Some(args.resolve(live)?);
+        }
+        match (self.manual_range, self.drift_start) {
+            (manual, None) => Ok(manual),
```
rudy-6-4, line 36: Here it should take the last live block know on db, to  catchup from it all queues are fresh.

coprocessor/fhevm-engine/host-listener/src/consumer/runner.rs
```diff
@@ -0,0 +1,306 @@
...
+            db.tick = self.db.tick.clone();
+            super::SLOW_LANE_PROMOTION_ATTEMPTED
+                .store(false, std::sync::atomic::Ordering::Release);
+            let id = format!("{}.s{subscription_id}", self.identity);
+            let client = ListenerConsumer::new(
+                &self.broker,
+                self.db.chain_id.as_u64(),
+                &id,
+            );
+            info!(subscription_id, consumer_id = %id, "Starting consumer subscription");
```
rudy-6-4, line 171: Read top of chain block number in db to add to recovery plan.

PR #3883, opened 2026-09-09
feat(solana): compose encrypted state with transient grants

Eikix:
Replace per-value application storage with host-owned `EncryptedState` slots and shared MMR history. Transaction-local scratch grants let programs compose encrypted results: confidential-token returns the transferred handle, then the batcher updates the contribution slot controlled by each user's JoinRecord. This removes `TransferReceipt`, the permanent transfer-result register, and the retired host/builder account paths without compatibility aliases.

The migration connects host/builder, token/batcher, SDK, listener/proof indexing, connector, relayer, Gateway admission and the vault demo. Burn retains its result slot and PendingBurn for settlement across transactions. Execution-level `returned_results` explicitly selects up to 32 handles; return bytes grant no permission. New decrypt grants/publication for history-only handles are deferred to zama-ai/fhevm-internal#2007. Fresh output permissions, historical decryption and current-slot publication remain supported. Generic disclosure authenticates the exact handle and cleartext through a KMS certificate and its public permission through the State's MMR proof, without certifying a token-kind label.

Validation: 1,028 Solana tests passed (2 explicitly ignored), 377 SDK Solana tests, 242 demo unit tests and 75 demo lifecycle tests passed. Strict Solana lint, the full SDK production build, generated-interface checks and independent correctness/maintainability reviews passed. Runtime coverage includes nested scratch-close rollback, foreign burn State and PendingBurn identity substitution, stale-proof rejection followed by refresh, and dust settlement cancellation/refund.

The fresh localstack smoke passed with real input proofs, join, dispatch, MMR/KMS settlement, claim and private decryption. A computer-use pass with the burner wallet exercised the demo's public and private deposits, partial and full redemption, yield advancement, permit-backed balance reveal/hide, disconnect/reconnect, validation guards, activity/developer details and architecture page. All ten Prometheus targets were up; KMS decryption traces and nine backend service logs were inspected. No service errors were observed; two SNS slow-query warnings were recorded. This verifies the local burner flow, not Phantom's approval UI or a multi-party deployment.

CI at `0ff95ee7d` passed every non-GPU check, including full Solana E2E scenarios, fresh deposit smoke, browser CORS/faucet checks and all localstack variants. The GPU job failed during provider setup, before tests ran: `slab-ci.zama.ai` returned HTTP 502 across three attempts and still returned 502 on 2026-09-10. Commit `8ded2ee8e` corrects documentation and generated API comments only; its new CI run is pending. Local validation of that cleanup passed: pinned SBF/IDL rebuild, ABI and codegen checks, five heap-budget tests, two builder doctests, the resource-ceiling snapshot test, all seven dead-surface checks and formatting. No production execution or ABI change is included in that correction.

Cost tradeoff versus `feature/solana`: join 338,790 CU (+29.3%), claim 249,811 (+9.8%), dispatch 203,040 (+2.5%), open batch 212,764 (+24.0%), settlement 274,179 (-2.5%). See the committed runtime snapshots for packet/account measurements and shape limits. Builder admission does not guarantee that the host's heap or the full transaction fits; INVARIANTS #14/#54/#61 distinguish these limits and cite current measurements.

Targets `feature/solana`. Design context: zama-ai/fhevm-internal#1997 and RFC35, following #3836. RFC35 is maintained separately; its `a858c10` revision was checked against the implemented interfaces. The invariant register restores the account-count, delegation-snapshot and HCU guarantees, and clarifies v4 certificate routing and when transient-grant consumers are deny-checked.


Eikix:
[AUTOMATED]

## Return-data channel: make returning explicit, decouple it from step count

**What the branch does today.** `fhe_execute` computes exactly one handle per step and `return_execution_handles` writes every one of them to return data, `steps × 32` bytes, no length prefix, whatever the step's output kind (slot, allow, grant, or purely transient). `ReturningFheExecution::invoke` then slices the wanted handle by **step index**. Three things are coupled by this: the one-result-per-step format, the return layout, and the builder's selection rule. The 32-step cap and Solana's 1024-byte return-data cap coincide only because of this layout.

**Two problems.**

1. Intermediates are returned although nobody can use them: a transient result has no slot, no leaf and no grant. The caller pays return-data bytes proportional to the step count for one handle it wants.
2. Multi-output operators do not fit. `main` already stores N outputs of one op as N rows sharing a `group_id` with `output_index 0..N-1` (#2514, arity up to 255; no operator wired yet). When shuffle or a similar op lands, one step yields N handles, a step-indexed slice points at the wrong bytes, and a single 8-way shuffle plus a few surrounding steps can exceed 32 returned handles.

**Proposal.** Treat "return" like `allow`, `make_public` and `allow_transient`: an output policy the step declares.

- The step output carries a `return` marker. The host returns only marked results, in declaration order.
- The builder selects by **return index**, not step index; `build_returning` keeps its signature and marks the selected result itself.
- A returned entry is addressed as `(step, output_index)` so a future multi-output step can mark any subset of its outputs.
- Return data shrinks to what the caller asked for, typically 32 bytes; the step count is no longer bounded by the return channel.
- Token and batcher call sites do not change; only `zama-fhe` and the host's execution loop do.

Not a blocker for this PR since no operator is multi-output yet, but it is cheaper to fold in now than after the step format is frozen in the IDL, listener and SDK. RFC 035 will state the intended semantics in the "Sharing a result inside one transaction" section.

PR #3889, opened 2026-09-09
feat(test-suite): http connector e2e

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/2006
Test suite [run](https://github.com/zama-ai/fhevm/actions/runs/34371078646)

### Summary

This PR adds end-to-end test coverage for the kms-connector's new HTTP decryption endpoint. It lets us deploy the `kms-connector[-i]-endpoint` container per KMS party and exercise `POST /v1/public-decrypt` and `POST /v1/user-decrypt` directly against every party's endpoint instead of only going through the relayer (for now).

### What changed

**fhevm-cli test profiles.** `commands/test.ts` adds four new profiles — `connector-http-public-decrypt`, `connector-http-user-decrypt`, `connector-http-negative`, and the umbrella `connector-http` (part of `standard`) — each guarded by a `connectorEndpointSkipReason()` precondition so they're skipped with a clear log line on bundles without the endpoint container.

**New e2e suites** under `test-suite/e2e/test/connectorHttp/`:
- `connectorHttpPublicDecrypt.ts` — recovers each KMS party's signature over `/v1/public-decrypt`, checks them against `KMSVerifier`, verifies 2t+1 quorum, cache/idempotency behavior, and agreement with the relayer path.
- `connectorHttpUserDecrypt.ts` — covers the unified EIP-712 permit against `/v1/user-decrypt` for every decryptable type, permissive mode, and idempotency (structural checks on shares; full reconstruction is left as a follow-up for the SDK).
- `connectorHttpNegative.ts` — exercises rejections the kms-worker decides against the real ACL (`403 acl_denied`, `403 user_signature_rejected`, `422 unprocessable`) and confirms a retryable error re-arms once the ACL changes. Endpoint-only rejections (`400 malformed`, `503`, `504`) stay covered by the Rust crate's own tests.
- `setup.ts` — shared fixture/party setup for the three suites above.

A thin client lives in `test/sdk/connector/connectorHttp.ts`: it fans out the request body (play Relayer's role), and re-submits transient worker outcomes.

**Compat and topology plumbing.** `supportsConnectorEndpoint` in `test-suite/fhevm/src/compat/compat.ts` detects whether the resolved bundle actually includes the endpoint image (either because `CONNECTOR_ENDPOINT_VERSION` is set or because an override targets it), so every downstream consumer degrades gracefully on stacks built before this container existed. `kms-party.ts`, `generate/compose.ts` and `generate/env.ts` were updated to append `-endpoint` alongside the existing `-db-migration`/`-gw-listener`/`-kms-worker`/`-tx-sender` suffixes per party, build the `kms-connector-endpoint` compose service conditionally, and populate a new `KMS_CONNECTOR_ENDPOINT_URLS` (comma-separated, one per party) plus `KMS_THRESHOLD` env var for the test-suite container.


.github/workflows/test-suite-e2e-tests.yml
```diff
@@ -65,21 +65,11 @@ on:
...
-        default: ""
-        type: string
-      connector-tx-sender-version:
-        description: "KMS Connector Transaction Sender Image Version"
-        default: ""
+      connector-versions:
+        description:
+          'KMS Connector image versions as a JSON object. Keys: db-migration, gw-listener, kms-worker, tx-sender,
+          endpoint. Example: {"gw-listener":"01cf0ff","endpoint":"01cf0ff"}'
+        default: "{}"
```
eudelins-zama, line 72: Had to do this because we had too many inputs (limited to 25 by Github) otherwise...

test-suite/e2e/test/connectorHttp/connectorHttpNegative.ts
```diff
@@ -0,0 +1,191 @@
...
+      expectErrorOnEveryParty(out.results, { status: 422, code: 'unprocessable' });
+
+      // Non-retryable: the stored error row is served as is on re-submission.
+      const again = await fanOut<UserDecryptionResponse>(USER_DECRYPT_ROUTE, body);
+      expectErrorOnEveryParty(again.results, { status: 422, code: 'unprocessable' });
+    });
+  });
+
+  describe('re-arm', function () {
+    it('test connector endpoint negative a retryable acl_denied becomes 200 once the ACL allows the handle', async function () {
```
nboisde, line 171: Test is neat here. 
So I assume, but hope not, that on eth, propag time could be above 12 seconds (for 1 block delay).
Make me think if we don't test ACL in relayer what would you consider as a good timeout and retry window (in relayer) then ?

test-suite/e2e/test/connectorHttp/connectorHttpPublicDecrypt.ts
```diff
@@ -0,0 +1,124 @@
...
+    ({ urls, kms } = await connectorSetup(this));
+    await initSigners(2);
+    signers = await getSigners();
+    instances = await createInstances(signers);
+    const factory = await ethers.getContractFactory('HTTPPublicDecrypt');
+    contract = await factory.connect(signers.alice).deploy();
+    await contract.waitForDeployment();
+  });
+
+  it('test connector endpoint version route answers on every party', async function () {
```
nboisde, line 42: Is this test related purely to public decrypt or version endpoint only ?

test-suite/e2e/test/sdk/connector/verify.ts
```diff
@@ -0,0 +1,159 @@
...
+  const first = ok[0].body as PublicDecryptionResponse;
+  expect(first.decryptionId).to.match(/^0x[0-9a-f]{64}$/i);
+  const signers = new Set<string>();
+  for (const r of ok) {
+    const body = r.body as PublicDecryptionResponse;
+    expect(body.decryptionId, `decryptionId differs at ${r.url}`).to.equal(first.decryptionId);
+    expect(body.decryptedResult, `decryptedResult differs at ${r.url}`).to.equal(first.decryptedResult);
+    const signer = recoverPublicDecryptSigner(ctHandles, body);
+    expect(kms.signers.has(signer), `${signer} (from ${r.url}) is not a registered KMS signer`).to.equal(true);
+    expect(signers.has(signer), `${signer} signed twice (${r.url})`).to.equal(false);
```
nboisde, line 128: This is what I were wondering, if there was a way to check each responses against a set of referred signers. 
Will add this as optional check on the relayer.

test-suite/e2e/test/sdk/connector/verify.ts
```diff
@@ -0,0 +1,159 @@
...
+export function verifyUserDecryptResponses(results: PostResult<UserDecryptionResponse>[]): { decryptionId: string } {
+  expect(results.length).to.be.greaterThan(0);
+  for (const r of results) {
+    expect(r.httpStatus, describeResult(r)).to.equal(200);
+  }
+  const first = results[0].body as UserDecryptionResponse;
+  expect(first.decryptionId).to.match(/^0x[0-9a-f]{64}$/i);
+  for (const r of results) {
+    const body = r.body as UserDecryptionResponse;
+    expect(body.decryptionId, `decryptionId differs at ${r.url}`).to.equal(first.decryptionId);
```
nboisde, line 154: Coherent for testing alignements, but on the relayer can I deal with the majority, to handle corrupted shares or a misbehaving kms connector ?

nboisde (approved):
LGTM ! 
Thanks for this huge work @eudelins-zama , pretty neat and helpful, shows a lot of things that could be checked in the relayer fanout module as well, and will be included.

test-suite/e2e/test/sdk/connector/verify.ts
```diff
@@ -0,0 +1,159 @@
...
+/** Field order is authoritative — it determines the EIP-712 type hash (mirror of the Gateway `Decryption` struct). */
+const PUBLIC_DECRYPT_TYPES = {
+  PublicDecryptVerification: [
+    { name: 'ctHandles', type: 'bytes32[]' },
+    { name: 'decryptedResult', type: 'bytes' },
+    { name: 'extraData', type: 'bytes' },
+  ],
+};
+
+/** The KMS signs `PublicDecryptVerification` under the GATEWAY chain id and the gateway `Decryption` contract. */
```
nboisde, line 52: Reading once again after reflection, are we sure we want to keep the checks on the gateway chain, and not leverage on what lives on host chain instead to check signatures ?

nboisde (approved):

test-suite/e2e/test/connectorHttp/connectorHttpNegative.ts
eudelins-zama: > Make me think if we don't test ACL in relayer what would you consider as a good timeout and retry window (in relayer) then ?

I thought ACL was tested on relayer's side :thinking: .
I would say it seems surprising to me to compute a handle, and immediately ask for its decryption, thus entering this race condition.
But to answer, I would say timeout ~30s, retry window ~3s (or exponential backoff starting ~1s, don't know Relayer's strategy on this)

test-suite/e2e/test/connectorHttp/connectorHttpPublicDecrypt.ts
eudelins-zama: It's the version endpoint only, did not wanted to create a dedicated file "just for this" for now, but we can if you prefer!

test-suite/e2e/test/sdk/connector/verify.ts
eudelins-zama: Yes ofc!

eudelins-zama: I'd say we should try to mimic this in our test-suite once Relayer is integrated to the e2e tests :slightly_smiling_face:

test-suite/e2e/test/sdk/connector/verify.ts
eudelins-zama: Yes! For now, decryption coming from both the Gateway/HTTP are handled the same way by the `kms-worker`. (and I think that's what we want for now)

So they use the same EIP-712 domain for public decryption, i.e. using the gateway chain id. For unified user decryption, we use the chain id of the host chain id though (but the decryption contract address on Gateway for the `verifying_contract` field).

nboisde: Yes ! I would say for stage 1 , lets not change what is behind endpoints indeed. But i think now we will be able to leverage on host chain for signatures if I understand correctly.

test-suite/e2e/test/connectorHttp/connectorHttpPublicDecrypt.ts
nboisde: Fine for me to not create more and more files, lets see further if we need to create new endpoints for health, liveness etc, at this point we will separate 👍

test-suite/e2e/test/connectorHttp/connectorHttpNegative.ts
nboisde: We will probably add optional modules to check ACL and ct-commits, but for the POC, I focused on the aggregator purely and check how it behaves if checks are not performed.
I get your point regarding directly asking for a handle plus a race condition, but i'd say if tomorrow relayer is permissionless, user can reimplement it and totally bypass thoses checks. 
At the end, I will probably get "acl_denied" from the kms connector, and the aggregator should react properly to this.

eudelins-zama: >  i'd say if tomorrow relayer is permissionless, user can reimplement it and totally bypass thoses checks.

Definetely! But in such case, priority is security (no issue here, we would simply reject the request).

eudelins-zama: And for UX, we still say that the error is retryable, so up to this permissionless relayer to take that into account and impl proper retry logic I'd say :slightly_smiling_face:

eudelins-zama:
@mergifyio queue

PR #3893, opened 2026-09-09
ci(sdk): add npm publish workflow for cleartext sdk packages

geoxel:
## Summary

Adds CI workflows to publish the cleartext SDK npm packages.

### Shared
- `sdk-cleartext-publish-package.yml` — shared reusable workflow that builds, packs, validates against npmjs.com, and publishes a package with npm provenance. Supports `dry_run` to run every step short of the actual publish.

### Packages
- `sdk-cleartext-publish-host-contracts-cleartext.yml` — publishes the `host-contracts-cleartext` package.
- `sdk-cleartext-publish-hardhat-plugin-v2.yml` — publishes the Hardhat v2 plugin package.
- `sdk-cleartext-publish-hardhat-plugin-v3.yml` — publishes the Hardhat v3 plugin package.

Each package has its own `workflow_dispatch` entry point that calls the shared workflow, so the packages can be published independently. The shared workflow has no trigger of its own — it only runs when called from one of these entry points.

### DIFF
--- a/.github/workflows/sdk-cleartext-publish-hardhat-plugin-v3.yml
+++ b/.github/workflows/sdk-cleartext-publish-hardhat-plugin-v3.yml
@@ -1,0 +1,28 @@
+name: sdk-cleartext-publish-hardhat-plugin-v3
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Build and check without publishing'
+        required: false
+        default: true
+        type: boolean
+
+# No permissions at the workflow level — the calling job declares only what it needs.
+permissions: {}
+
+jobs:
+  publish:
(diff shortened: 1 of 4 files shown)


.github/workflows/sdk-cleartext-publish-hardhat-plugin-v3.yml
```diff
@@ -0,0 +1,28 @@
+name: sdk-cleartext-publish-hardhat-plugin-v3
```
tawadaa, line 1: @geoxel why do we have 2 workflows for the same plugin?
Can't we have one workflow and manage the versioning?

geoxel: Thank you for your feedback. I am closing this PR since HH plugin won't be in fhevm repo for the moment. But we will take your remark into account for the next version of the scripts.

PR #3899, opened 2026-09-10
fix(coprocessor): guard a green deployment against a database in the wrong state

PanGan21:
## Summary

Two defensive fixes for a new release (green) deployment that meets a database in the wrong state.

- A service started before its migration no longer reads the missing `versioning.consensus_version` column as "unseeded -> blue". It fails, restarts, and resolves its role once the migration has run. A missing `versioning` table stays permissive.
- A migrator only creates an empty database when told to: `ALLOW_DB_BOOTSTRAP=true`, exposed as `dbMigration.allowBootstrap` (off by default). `initialize_db.sh` exits before creating anything otherwise; `bootstrap_versioning` checks it again.
- Chart comment and runbook: run the migration before starting green; a green release never creates a database.

Chart default change: fresh installs must set `dbMigration.allowBootstrap: true`.

Related: 
- zama-ai/fhevm-internal#2009
- zama-ai/fhevm-internal#2010

### DIFF
--- a/charts/coprocessor/templates/coprocessor-db-migration.yaml
+++ b/charts/coprocessor/templates/coprocessor-db-migration.yaml
@@ -53,6 +53,8 @@ spec:
         command: ["/initialize_db.sh"]
         {{- end }}
         env:
+          - name: ALLOW_DB_BOOTSTRAP
+            value: {{ if .Values.dbMigration.allowBootstrap }}"true"{{ else }}"false"{{ end }}
           {{- with .Values.commonConfig.databaseUser }}
           - name: DATABASE_USER
             {{- if .value }}
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
@@ -327,10 +327,11 @@ pub struct StaleStackError {
     pub live: i64,
 }
 
-/// True if `err` is Postgres `undefined_table` (SQLSTATE 42P01) or `undefined_column`
-/// (SQLSTATE 42703) - i.e. the `versioning` row is not readable yet (migrations not applied).
+/// True if `err` is Postgres `undefined_table` (SQLSTATE 42P01) - i.e. the `versioning`
+/// table does not exist yet (migrations not applied). A missing `consensus_version` column
+/// is the previous release's database before this release's migration; that stays an error.
 fn is_unmigrated(err: &sqlx::Error) -> bool {
-    matches!(err, sqlx::Error::Database(db) if matches!(db.code().as_deref(), Some("42P01" | "42703")))
+    matches!(err, sqlx::Error::Database(db) if db.code().as_deref() == Some("42P01"))
 }
 
 /// Fetch the live consensus version singleton, or `None` if the `versioning` row is
--- a/coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
+++ b/coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
@@ -330,7 +330,7 @@ pub struct StaleStackError {
 /// True if `err` is Postgres `undefined_table` (SQLSTATE 42P01) - i.e. the `versioning`
 /// table does not exist yet (migrations not applied). A missing `consensus_version` column
 /// is the previous release's database before this release's migration; that stays an error.
-fn is_unmigrated(err: &sqlx::Error) -> bool {
+fn is_undefined_table(err: &sqlx::Error) -> bool {
     matches!(err, sqlx::Error::Database(db) if db.code().as_deref() == Some("42P01"))
 }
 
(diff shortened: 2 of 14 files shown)


coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
```diff
@@ -327,10 +327,11 @@ pub struct StaleStackError {
...
 }
 
-/// True if `err` is Postgres `undefined_table` (SQLSTATE 42P01) or `undefined_column`
-/// (SQLSTATE 42703) - i.e. the `versioning` row is not readable yet (migrations not applied).
+/// True if `err` is Postgres `undefined_table` (SQLSTATE 42P01) - i.e. the `versioning`
+/// table does not exist yet (migrations not applied). A missing `consensus_version` column
+/// is the previous release's database before this release's migration; that stays an error.
 fn is_unmigrated(err: &sqlx::Error) -> bool {
-    matches!(err, sqlx::Error::Database(db) if matches!(db.code().as_deref(), Some("42P01" | "42703")))
+    matches!(err, sqlx::Error::Database(db) if db.code().as_deref() == Some("42P01"))
```
goshawk-3, line 334: It's unclear why you removed the "42703" err code?

charts/coprocessor/templates/coprocessor-db-migration.yaml
```diff
@@ -53,6 +53,8 @@ spec:
         command: ["/initialize_db.sh"]
         {{- end }}
         env:
+          - name: ALLOW_DB_BOOTSTRAP
```
goshawk-3, line 56: Can you elaborate on the rationale of `feat(coprocessor): let only the first release create a database` ?

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
```diff
@@ -327,10 +327,11 @@ pub struct StaleStackError {
     pub live: i64,
 }
 
-/// True if `err` is Postgres `undefined_table` (SQLSTATE 42P01) or `undefined_column`
-/// (SQLSTATE 42703) - i.e. the `versioning` row is not readable yet (migrations not applied).
+/// True if `err` is Postgres `undefined_table` (SQLSTATE 42P01) - i.e. the `versioning`
+/// table does not exist yet (migrations not applied). A missing `consensus_version` column
+/// is the previous release's database before this release's migration; that stays an error.
 fn is_unmigrated(err: &sqlx::Error) -> bool {
```
goshawk-3: The name is a bit misleading. It should be something like is_undefined_table or is_undefined_column.

A table/column may be missing for reasons other than an unapplied migration

coprocessor/fhevm-engine/tfhe-worker/src/tests/versioning.rs
```diff
@@ -55,6 +55,31 @@ async fn gcs_mode_tracks_consensus_version_not_release() {
...
+        .expect("connect");
+
+    // The previous release's database: versioning exists, the column does not.
+    sqlx::query("ALTER TABLE versioning DROP COLUMN consensus_version")
+        .execute(&pool)
+        .await
+        .expect("drop column");
+    let error = resolve_gcs_mode(db.db_url())
+        .await
+        .expect_err("must not pick a mode from an unmigrated database");
```
goshawk-3, line 76: If resolve_gcs_mode fails, how will each service handle the error? Will it keep restarting and retrying? Have you tested the Copro stack in this edge case?

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
PanGan21: `42703` means table exists, no `consensus_version` column. This can happen if GCS runs before the migrations.
I removed it because reading that as unseeded makes a 0.15 service resolve as blue. now the read fails the service exits and resolves correctly once it restarts after the migration is applied.

charts/coprocessor/templates/coprocessor-db-migration.yaml
PanGan21: On an empty database, whichever migration job runs first makes its release the live one. 
In B/G the green migration can win that race, so blue starts retired and green starts blue, this broke the preview env recently. The migrator now needs `dbMigration.allowBootstrap` to create a database (off by default): the first install sets it, a green release never does.
This is mostly a safeguard and make sure that creating a new db is intentional

coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs
PanGan21: Agreed, i renamed it here: https://github.com/zama-ai/fhevm/pull/3899/changes/fb28acf3cee91b661b45ffed67fe9900231e2e39

coprocessor/fhevm-engine/tfhe-worker/src/tests/versioning.rs
PanGan21: If resolve_gcs_mode fails it is treated as fatal error the service exits and k8s should restart it. The first restart after the migration resolves this normally.
The best way to test this is on k8s environment, but in reality it is one more defensive check in case migration hasn't run before the GCS.

charts/coprocessor/templates/coprocessor-db-migration.yaml
PanGan21: Some more details about this on this ticket: https://github.com/zama-ai/fhevm-internal/issues/2010

goshawk-3 (approved):

PR #3900, opened 2026-09-10
fix(coprocessor): pipeline ordered broadcasts and bound retry preparation in tx-sender

antoniupop:


coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
```diff
@@ -135,10 +362,31 @@ where
...
+        // Estimation has its own deadline. The binary leaves gas unset on all
+        // attempts so terminal revert handling can recognize completed work.
+        // Pre-set gas remains supported for existing library callers.
+        let overprovisioned_txn = tokio::time::timeout(
+            gas_estimation_timeout,
+            self.overprovision_gas_limit(txn_request, percent),
+        )
+        .await
+        .map_err(|_| TransportErrorKind::custom_str("gas estimation timeout"))
+        .flatten()?;
```
Eikix, line 393: [AUTOMATED]

**[P1] Bound retained WebSocket requests across estimation timeouts**

This timeout releases the admission permit, but Alloy 1.1.2's WebSocket request manager retains unanswered requests until a response arrives and replays them on reconnect. Since the binary uses `connect_ws`, repeated timeout/retry cycles can accumulate abandoned estimates despite `max_inflight_sends`.

Reproduced with this provider, one permit, and a WebSocket server that never answers estimates: four sequential 150 ms timeouts left four unanswered requests, and reconnect replayed all four.

Please make retirement of locally retained requests agree with admission, or use a request path with bounded retention. Add a regression covering unanswered estimates followed by reconnect.

Confidence: 98/100

coprocessor/fhevm-engine/transaction-sender/src/bin/transaction_sender.rs
```diff
@@ -97,9 +97,34 @@ struct Conf {
...
+    /// (`eth_sendRawTransaction`). Inclusion is bounded separately by
+    /// `--txn-receipt-timeout-secs`, which used to be an alias of this flag and
+    /// is now a distinct phase.
+    #[arg(long, default_value_t = 4)]
     send_txn_sync_timeout_secs: u16,
 
+    /// Deadline for waiting for a transaction to be mined, awaited outside the
+    /// nonce mutex so it does not block other sends. A timeout here does not
+    /// reset the nonce sequence: the transaction is already acknowledged.
+    #[arg(long, default_value_t = 30)]
```
Eikix, line 109: [AUTOMATED]

**[P1] Preserve the existing timeout flag contract or migrate deployment arguments**

`--txn-receipt-timeout-secs` previously aliased the sync submission timeout; it now sets only the receipt timeout. The declared [testnet](https://github.com/zama-zws/gitops/blob/5e5680739deb1f6866b30097d10560923d5e4dde/values/gw-blockchain/gw-coprocessor/values-zws-testnet.yaml#L146) and [mainnet](https://github.com/zama-zws/gitops/blob/5e5680739deb1f6866b30097d10560923d5e4dde/values/gw-blockchain/gw-coprocessor/values-coproc-mainnet.yaml#L162) configurations still pass `4`, overriding the new 30-second receipt default. Existing users passing a larger value also silently fall back to a four-second submission timeout.

An image-only upgrade can therefore retain unintended receipt deadlines or shorten configured submission budgets, causing premature retries. Preserve the alias and introduce a distinct receipt flag, or coordinate the affected deployment arguments before rollout. These are declared configurations; running pod arguments were not checked.

Confidence: 98/100

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
```diff
@@ -50,43 +155,165 @@ where
...
+    ) -> TransportResult<TransactionReceipt> {
+        let tx_hash = *pending.tx_hash();
+        // Reserve a slice of the deadline for the final lookup below.
+        let final_budget = (receipt_timeout / 10).clamp(
+            Duration::from_millis(250).min(receipt_timeout),
+            Duration::from_secs(2),
+        );
+        let watch_budget = receipt_timeout.saturating_sub(final_budget);
+
+        match tokio::time::timeout(watch_budget, pending.get_receipt()).await {
```
Eikix, line 267: [AUTOMATED]

**[P2] Give the underlying receipt watcher a finite timeout**

`pending.get_receipt()` registers an Alloy watcher with no internal timeout. The outer Tokio timeout drops the waiting receiver but does not unregister that watcher. If the transaction is subsequently dropped or replaced and its hash never mines, the watcher remains retained and keeps block polling active.

A local Anvil reproduction using this provider continued polling after the receipt deadline and mempool removal; a comparison with a finite builder timeout stopped polling.

Set `pending.with_timeout(Some(receipt_timeout))` before `get_receipt()`, while retaining the outer deadline and bounded final lookup.

Confidence: 98/100

Eikix (changes requested):
[AUTOMATED]

Reviewed `42f6443f1a6bbe7b13fd92e7f8ee338a84da6233` against `release/0.13.x`.

Seeding from `pending`, preserving the nonce after receipt timeouts, and draining started tasks are useful reliability fixes. Requesting changes for the WebSocket request retention, timeout flag compatibility, and receipt watcher expiration issues detailed inline.

Validation: 15 focused tests passed locally; transaction-sender CI reported 67 passed and one ignored load test. Two additional review-only reproductions confirmed the request retention and watcher lifetime issues.

Performance validation before rollout: the release base already ran concurrent sync submissions. [Upstream Nitro waits for sequencing before returning ordinary submission success](https://github.com/OffchainLabs/nitro/blob/a618155919315241665356fe60f3cd00d66d5e46/execution/gethexec/sequencer.go#L548-L585); Conduit's deployed behavior was not verified. Please compare completion throughput and submission/receipt latency against the release base on the actual Gateway WebSocket endpoint, and qualify the claim that the nonce lock is necessarily released before inclusion. The local tests do not establish that performance benefit.

coprocessor/fhevm-engine/transaction-sender/src/nonce_managed_provider.rs
antoniupop: Good point - fixing.

PR #3902, opened 2026-09-10
chore(library-solidity): remove redundant checks to save gas

jatZama:
<html><head></head><body><h2>Remove redundant type checks</h2><p>Remove 457 redundant type assertions across 449 overloads in <code dir="ltr">FHE.sol</code>, reducing external <code dir="ltr">checkHandleType</code> calls while preserving type safety and zero-handle initialization.</p><p>The changes rely on checks already performed by <code dir="ltr">FHEVMExecutor</code>:</p><ul><li><strong>Same-type binary operations:</strong> validate one operand’s expected type; the executor requires the other operand to match.</li><li><strong>Mixed-width binary operations:</strong> retain source validation in the widening cast. The executor requires the uncast operand to match the cast’s output type.</li><li><strong>Encrypted shifts and rotations:</strong> retain validation of the shift amount before widening; the executor checks that the other operand matches. For two <code dir="ltr">euint8</code> operands, retain one explicit type assertion.</li><li><strong>Scalar-left subtraction:</strong> trivial encryption establishes the left operand’s type, which the executor requires the encrypted right operand to match.</li><li><strong><code dir="ltr">select</code>:</strong> retain one branch’s expected-type assertion. The executor already requires a boolean control and matching branch types.</li></ul><p>Update the generator and regenerate both library copies. Regression tests cover incorrect underlying types, matching-but-wrong type pairs, and uninitialized operands. Invalid inputs still revert, although some failures now use the executor’s <code dir="ltr">IncompatibleTypes</code> or <code dir="ltr">UnsupportedType</code> errors instead of <code dir="ltr">InvalidType</code>.</p><p>Validation: 73 library tests and 100 executor tests passed, plus 18 local gas benchmarks.</p><h2>Gas comparison</h2><p>Measured before and after with Solidity 0.8.24, optimizer enabled with 800 runs, Cancun, and <code dir="ltr">viaIR=false</code>. Measurements use the same adapter and real executor, ACL, and HCU proxies, with cold account/storage access. Values represent adapter execution gas, excluding setup and transaction intrinsic gas.</p><div><div><div>

| Operation with initialized handles | Before | After | Gas saved | Reduction |
| --- | ---: | ---: | ---: | ---: |
| `and(ebool, ebool)` | 53,111 | 51,437 | 1,674 | 3.15% |
| `add(euint32, euint32)` | 53,058 | 51,383 | 1,675 | 3.16% |
| `eq(eaddress, eaddress)` | 53,549 | 51,874 | 1,675 | 3.13% |
| `add(euint16, euint32)` | 67,982 | 66,307 | 1,675 | 2.46% |
| `add(euint32, euint16)` | 67,991 | 66,316 | 1,675 | 2.46% |
| `shl(euint8, euint8)` | 53,027 | 51,352 | 1,675 | 3.16% |
| `shl(euint32, euint8)` | 68,078 | 66,403 | 1,675 | 2.46% |
| `sub(uint32, euint32)` | 61,610 | 59,935 | 1,675 | 2.72% |
| `select(ebool, ebool, ebool)` | 56,770 | 53,422 | 3,348 | 5.90% |
| `select(ebool, euint32, euint32)` | 58,896 | 55,547 | 3,349 | 5.69% |

</div></div></div><p>Uninitialized-operand cases save 38–76 gas: those paths already bypassed type assertions and performed trivial encryption. Unchanged scalar, unary, and cast operations have identical gas costs before and after.</p><p>The benchmark adapter’s deployed bytecode increases by 71 bytes, from 6,990 to 7,061 bytes, due to inline initialization.</p></body></html>

### DIFF
--- a/library-solidity/codegen/src/templateFHEDotSol.ts
+++ b/library-solidity/codegen/src/templateFHEDotSol.ts
@@ -188,7 +188,8 @@ function handleSolidityTFHEEncryptedOperatorForTwoEncryptedTypes(
   }
 
   if (lhsFheType.type.startsWith('Uint') && rhsFheType.type.startsWith('Uint')) {
-    // Widening casts validate and initialize their source; avoid checking it twice.
+    // A widening cast validates its source and fixes the result type. Otherwise validate the left operand.
+    // The executor requires the other operand to match that type, so it only needs initialization here.
     // Determine the maximum number of bits between lhsBits and rhsBits
     const outputBits = Math.max(lhsFheType.bitLength, rhsFheType.bitLength);
     const castLeftToRight = lhsFheType.bitLength < rhsFheType.bitLength;
@@ -215,8 +216,14 @@ function handleSolidityTFHEEncryptedOperatorForTwoEncryptedTypes(
     function ${
       operator.name
     }(e${lhsFheType.type.toLowerCase()} a, e${rhsFheType.type.toLowerCase()} b) internal returns (${returnType}) {
-        ${castLeftToRight ? '' : assignValidatedFheValue('a', lhsFheType.type)}
-        ${castRightToLeft ? '' : assignValidatedFheValue('b', rhsFheType.type)}
+        ${
+          castLeftToRight
+            ? ''
+            : castRightToLeft
+              ? initializeFheValue('a', lhsFheType.type)
+              : assignValidatedFheValue('a', lhsFheType.type)
+        }
+        ${castRightToLeft ? '' : initializeFheValue('b', rhsFheType.type)}
         return ${returnType}.wrap(${implExpression});
     }
 `);
@@ -227,7 +234,7 @@ function handleSolidityTFHEEncryptedOperatorForTwoEncryptedTypes(
     */
     function ${operator.name}(ebool a, ebool b) internal returns (ebool) {
         ${assignValidatedFheValue('a', 'Bool')}
-        ${assignValidatedFheValue('b', 'Bool')}
+        ${initializeFheValue('b', 'Bool')}
         return ebool.wrap(Impl.${operator.name}(ebool.unwrap(a), ebool.unwrap(b), false));
     }
 `);
@@ -238,7 +245,7 @@ function handleSolidityTFHEEncryptedOperatorForTwoEncryptedTypes(
       */
       function ${operator.name}(eaddress a, eaddress b) internal returns (ebool) {
           ${assignValidatedFheValue('a', 'Address')}
-          ${assignValidatedFheValue('b', 'Address')}
+          ${initializeFheValue('b', 'Address')}
           return ebool.wrap(Impl.${operator.name}(eaddress.unwrap(a), eaddress.unwrap(b), false));
       }
   `);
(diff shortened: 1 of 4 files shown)


library-solidity/codegen/src/templateFHEDotSol.ts
```diff
@@ -227,7 +234,7 @@ function handleSolidityTFHEEncryptedOperatorForTwoEncryptedTypes(
     */
     function ${operator.name}(ebool a, ebool b) internal returns (ebool) {
         ${assignValidatedFheValue('a', 'Bool')}
-        ${assignValidatedFheValue('b', 'Bool')}
+        ${initializeFheValue('b', 'Bool')}
```
0xalexbel, line 237: minus 1 func call 🚀

library-solidity/test/FHETypeSafety.t.sol
```diff
@@ -230,14 +250,117 @@ contract FHETypeSafetyTest is HostContractsDeployerTestUtils {
...
+        executor.checkHandleType(adapter.add32(u32, 0), FheType.Uint32);
+        executor.checkHandleType(adapter.mixedAdd(0, u32), FheType.Uint32);
+        executor.checkHandleType(adapter.mixedAdd(u16, 0), FheType.Uint32);
+        executor.checkHandleType(adapter.mixedAddReverse(0, u16), FheType.Uint32);
+        executor.checkHandleType(adapter.mixedAddReverse(u32, 0), FheType.Uint32);
+        executor.checkHandleType(adapter.shift8(0, u8), FheType.Uint8);
+        executor.checkHandleType(adapter.shift8(u8, 0), FheType.Uint8);
+        executor.checkHandleType(adapter.shift32(0, u8), FheType.Uint32);
+        executor.checkHandleType(adapter.shift32(u32, 0), FheType.Uint32);
+        executor.checkHandleType(adapter.subScalar32(1, 0), FheType.Uint32);
```
0xalexbel, line 356: what is this individual test purpose ? 
looks like a different kind of test ? I'm probably missing something here.
`addScalar32` is also available -> expecting symmetry

jatZama: These assertions check zero-handle initialization and the result’s FHE type.
The purpose of subScalar32(1, 0) is to verify that scalar-left subtraction still initializes its encrypted operand after removing the redundant type assertion. Remember that fheSub is a special case, handled differently than other operators, since it is not symmetric (unlike fheAdd or fheMul, fheEq, etc). I included subtraction specifically to cover the initialization code introduced by this PR. However, if you want I could add another assertion for addition.

jatZama: Added in this commit https://github.com/zama-ai/fhevm/pull/3902/changes/65004ef0e27cb76452efb8e72e0213d39613c8cf

0xalexbel: Then a better naming would be cool because subScalar32 inverts its args on purpose I guess. Not urgent at all

0xalexbel (approved):

isaacdecoded (approved):

obatirou (approved):

PR #3904, opened 2026-09-10
feat(coprocessor): process staged KMS activations in host-consumer

rudy-6-4:


coprocessor/fhevm-engine/host-listener/src/consumer/ingestion.rs
```diff
@@ -118,14 +167,22 @@ impl BlockIngestor {
...
-            catchup,
-        )
-        .await;
+        let finalized =
+            matches!(payload.flow, BlockFlow::Final | BlockFlow::FinalCatchup);
+        let catchup = matches!(
+            payload.flow,
+            BlockFlow::Catchup | BlockFlow::FinalCatchup
+        );
+        // Final delivery is expected to repeat live ingestion; it is not a live duplicate.
```
rudy-6-4: only for live

coprocessor/fhevm-engine/host-listener/src/consumer/ingestion.rs
```diff
@@ -177,6 +125,52 @@ impl BlockIngestor {
...
+
+    async fn ready(&self) -> Result<bool, HandlerError> {
+        if self.mode.is_paused() {
+            return Ok(false);
+        }
+        if !self.last_known_drift.read().await.is_finished {
+            return Ok(false);
+        }
+        Ok(true)
+    }
```
PanGan21, line 137: For better coverage of edge cases we can also call `latest_signal_for_chain()` here since `last_known_drift` is cached and not updated in memory in between blocks.

coprocessor/fhevm-engine/host-listener/src/consumer/ingestion.rs
```diff
@@ -177,6 +125,52 @@ impl BlockIngestor {
...
+    ) {
+        let mut interval = tokio::time::interval(Duration::from_secs(5));
+        interval
+            .set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        loop {
+            tokio::select! {
+                _ = cancel.cancelled() => return,
+                _ = interval.tick() => {},
+            }
+            if let Err(error) = self.process_kms_once(s3.clone()).await {
```
PanGan21, line 155: `process_kms_once` can be called inside a `tokio::select!` with `cancel.cancelled()` so shutdown cancels the pass and releases its transaction

PR #3905, opened 2026-09-10
feat(solana): share transient storage and computation ingestion

Eikix:
Current validation: CI is green at `55c779725ad604dd49a28e50b7bfed28e3e05eba`: 81 successful check runs plus the successful CLA status, no failed or pending checks (28 skipped and one neutral excluded). This includes CPU worker/listener tests, all SDK DoD variants, [Solana E2E](https://github.com/zama-ai/fhevm/actions/runs/34580209760/job/103201765652), and [GPU tests and teardown](https://github.com/zama-ai/fhevm/actions/runs/34580209467). Two CPU setup jobs and SDK v11 initially failed downloading dependencies; each passed one retry without code changes. The latest commit only corrects the glossary. The preceding finishing commit `c6c75341da0254b7252555d8d7893e054a1999e0` passed 161 listener tests (one existing ignore), five SQL integration tests, 14 focused demo tests, demo typecheck/lint, strict Solana-feature Clippy and repository precommit formatting/workspace checks/Clippy. Earlier full-stack and browser evidence below retains its tested-commit attribution; no new browser session was run for the glossary correction. Coprocessor-owner review is with Elias.

Implements zama-ai/fhevm-internal#2003 on the merged encrypted-storage design. One `TransientStore` records produced handles, explicit Store grants, operand origins and HCU. `fhe_execute` evaluates ordered arithmetic, then applies Store effects and returns selected handles. Rust applications and JavaScript transaction builders share the transient store across CPIs and top-level instructions; a batcher consumes a token result through a Store grant and updates its own contribution.

Persistent storage is named `EncryptedStore`, paired with `TransientStore`. The Rust SDK uses `Store`, `StoreId`, `StoreOutput` and `StoreSlot`; SDK, proof-service, relayer and connector fields use `encryptedStore`. Anchor account/create/public-seal discriminators and serialized account fixtures are regenerated together; deploy and reseed the PoC components together because old account discriminators are no longer accepted. The PDA seed remains `encrypted-state`; account layouts, request/permit bytes, MMR commitments and SQL schema/migration checksums are unchanged.

The listener decodes native EVM/Solana inputs into one computation representation for dependency analysis, database insertion and the existing slow-lane policy. Its fields are private; checked constructors validate operand kinds, arity, byte widths and output-group bounds before ingestion. Malformed computations return an error rather than being skipped, with expected/received operand shape and EVM block/transaction/log or Solana slot/mapped-transaction-ID/record context. Operand contents are not logged. Ciphertext-type and execution semantics remain owned by the executor/worker. Worker production interfaces and schema stay unchanged. Multi-output groups count newly inserted rows, including partial replay. EVM `Initialized`, `Upgraded` and `VerifyInput` events still count toward event reconciliation, but no longer inflate computation-chain size/depth: they previously produced no computation rows. Disabling slow-lane scheduling resets all Slow rows, including temporarily locked rows; the existing database pool bounds each batch/lock wait to 120 seconds.

The fixed 10,168-byte transient store holds 112 result occurrences and 32 explicit grants, is created by `open_transient_store` and closed last by `close_transient_store` with a rent refund. The JS `createSolanaFheTransaction` wrapper inserts both; callers forward `accounts.transientStore`. The payer supplies funding, while Store authority remains the credential. Return data confers no permission. HCU totals and critical-path depth span the transaction; application block accounting remains attributed to `(program, scope)`. Packet, heap and account limits can bind before the journal caps. Settlement serialization fits seven certificate signatures and five MMR siblings using the provisioned 27-entry ALT; six siblings reject before sending.

Independent correctness and maintainability reviews were folded in after each implementation milestone. The final follow-up pins maximum repeated-handle depth, both Store operand signature paths, all 256 host/listener boundary-bit positions, EVM non-computation events and bounded SQL reset waits. Settlement pins the final compiled close instruction and shared transient-store/Instructions-sysvar forwarding through the generated ALT-decoded instruction; the existing two-transaction real-worker vertical checks the second transaction’s boundary mask in both its handle preimage and stored row. Local Solana runtime/SBF/ABI, SDK and demo checks passed. The final constructor follow-up passed 161 listener tests (one existing ignore), four SQL integration tests, strict Clippy/workspace checks and the explicitly enabled real-TFHE two-transaction vertical. A fresh renamed-API demo passed the clean deposit smoke, all seven live scenarios including Squads, and browser deposit/yield/private redeposit/full redemption with real permits and decryption. Final balances were 900 USDC, 107 cUSDC and zero cShares. Same-page outage recovery, reconnect, Explorer, metrics and traces also passed. That live boot used c527 host/SDK/listener; the subsequent listener constructor change was verified separately through the production reconstruction/SQL/real-worker path. The subsequent persistent-store rename passed the rebuilt Solana runtime/SBF/ABI checks, SDK379, demo250, harness466, listener161, SQL5 and relayer49+5 tests. All 191 Solana connector integration tests and the shared extraData byte-vector test also passed. CI passed at `860739da6351b062533480beeba52e19486dd3e4`: 82 successful checks, no failures or pending checks (28 skipped and one neutral are excluded). This includes all SDK DoD variants, host-listener, Solana runtime, KMS/relayer, [Solana live scenarios and the full deposit arc](https://github.com/zama-ai/fhevm/actions/runs/34504290314/job/102962816229) (8 passing tests, 45 assertions across the two suites), and [GPU worker tests with teardown](https://github.com/zama-ai/fhevm/actions/runs/34504290392). The browser session above predates this mechanical rename; it is not claimed as a new browser run.

Composition is documented in `solana/README.md` and DD-050 in `solana/docs/DESIGN_DECISIONS.md`. Runtime snapshots include the transient-store lifecycle; baseline and current fixtures/toolchain metadata differ, so their deltas are observations rather than controlled performance comparisons. Local demo transaction evidence now survives RPC pruning without inventing unavailable details, and the validator retains up to 1,000,000 shreds for longer QA sessions.


coprocessor/fhevm-engine/host-listener/src/database/computation.rs
```diff
@@ -0,0 +1,595 @@
...
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Computation {
+    pub operation: SupportedFheOperations,
+    pub operands: Vec<Operand>,
+    /// Ordered outputs; their first handle identifies a multi-output group.
+    pub outputs: Vec<Handle>,
```
Eikix: [AUTOMATED]

[P2] Enforce the agreed computation invariants at construction

Issue zama-ai/fhevm-internal#2003 asks to keep the fields private and validate arity, width and encoding through constructors. These public fields and the unchecked `single(operation, operands, result)` leave that requirement unenforced: callers can construct, for example, a zero-operand `FheAdd` or a `FheCast` with a wrongly sized clear type operand. `insert_tfhe_event` validates only output count before serializing operands, so the common dependency/SQL path still relies on every producer maintaining the contract separately.

Please make the fields private and route production construction through checked constructors, with focused invalid-arity/width cases. Reuse existing operation metadata where applicable; this does not need a new framework. If validation is intentionally owned elsewhere, identify and enforce that boundary and reconcile the issue's acceptance requirement. No current decoder emitting these malformed examples was established; this is a confirmed design-agreement gap.

Confidence: 95/100

Eikix (commented):
[AUTOMATED]

Reviewed `c52703c84ef37885d875977ed875875b30be51e0` against `feature/solana` at `83544b3b3cdb6a7d345f1353a99028d6f431b64f`, including the design constraints in zama-ai/fhevm-internal#2003.

The shared transaction store, separate arithmetic/effects, and native decoding into common listener data improve the architecture. The migrated token composition and rewritten host README address the corresponding problems in the base branch. Separate on-chain and listener passes did not confirm an introduced runtime defect in the reviewed paths.

One design-contract gap remains: the new computation representation does not enforce the construction invariants requested in #2003. The inline comment describes the missing boundary and a scoped remedy; it is not a claim that a current on-chain decoder already produces malformed computations.

Validation: the two focused `fheTransaction.test.ts` tests passed against this head using existing local dependencies (@solana/kit 6.10.0 and Vitest 4.1.9); `git diff --check` passed. Source review covered transaction lifetime/grants/HCU, listener masks/dependency grouping/SQL replay, and SDK/demo composition. GitHub reports the Solana build/runtime, demo and test-suite jobs successful; host-listener CI was still running at the last check. No fresh full Cargo/SBF build, live validator/demo flow or end-to-end decryption was run for this review.

PR #3907, opened 2026-09-10
feat(coprocessor): verify peer block manifests

rudy-6-4:
Download and archive authenticated peer revisions, and then localize handle differences. 
Local difference are recorded in a table to later contain drift propagation and heal.


Eikix (commented):

Reviewed `c0611a9` against #3557. The observation-only approach looks sound. Three fixes are recommended before approval:

- **False drift findings:** [`drift_findings.rs:613`](https://github.com/zama-ai/fhevm/blob/c0611a9fabfad9dfa6e8f0a99075de990eeafd47/coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/verification/drift_findings.rs#L613) compares entire descriptors, including error wording and Gateway provenance excluded from consensus digests. A block with one real mismatch and another diagnostic-only difference records both handles as drifted. Reproduced with the extracted comparison function. Compare consensus fields and add this two-handle regression test.
- **Lost retry for a newer revision:** [`peer_downloader.rs:752`](https://github.com/zama-ai/fhevm/blob/c0611a9fabfad9dfa6e8f0a99075de990eeafd47/coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/verification/peer_downloader.rs#L752) can use an older cached manifest after downloading a discovered newer revision times out. If the cached observations reach consensus, the task becomes terminal and never retries that newer revision. Preserve the incomplete-download state; test a timeout with an older revision already archived. Code-path finding, not a runtime reproduction.
- **Dotted bucket names:** [`peer_manifest_source.rs:276`](https://github.com/zama-ai/fhevm/blob/c0611a9fabfad9dfa6e8f0a99075de990eeafd47/coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/verification/peer_manifest_source.rs#L276) parses `s3://operator.manifests/prefix` as bucket `operator`. Reproduced with the extracted parser. Preserve the full bucket name and cover dotted names in parser tests.

Validation: current-head consensus-detector CI passed 127 tests; full E2E was skipped. Focused local tests reproduced the first and third issues. Formatting and chart-version checks also need fixing.

Before rollout, please verify cross-operator LIST + GET with the deployed AWS role: existing peer GET access alone does not establish ListBucket permission. This is a validation question, not a proven deployment failure.

rudy-6-4:
> 
> Reviewed `c0611a9` against #3557. The observation-only approach looks sound. Three fixes are recommended before approval:
> 
>     * **False drift findings:** [`drift_findings.rs:613`](https://github.com/zama-ai/fhevm/blob/c0611a9fabfad9dfa6e8f0a99075de990eeafd47/coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/verification/drift_findings.rs#L613) compares entire descriptors, including error wording and Gateway provenance excluded from consensus digests. A block with one real mismatch and another diagnostic-only difference records both handles as drifted. Reproduced with the extracted comparison function. Compare consensus fields and add this two-handle regression test.
> 
It's right. Error wording and gateway id are excluded from digest  but hashing is not used everywhere . Fixing.

rudy-6-4:
> * **Lost retry for a newer revision:** [`peer_downloader.rs:752`](https://github.com/zama-ai/fhevm/blob/c0611a9fabfad9dfa6e8f0a99075de990eeafd47/coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/verification/peer_downloader.rs#L752) can use an older cached manifest after downloading a discovered newer revision times out. If the cached observations reach consensus, the task becomes terminal and never retries that newer revision. Preserve the incomplete-download state; test a timeout with an older revision already archived. Code-path finding, not a runtime reproduction.

That's ok, we only need the quorum to make all copro progress. So if one copro manifest is stalled (only happens if healed)/missing/corrupt , but it's not breaking quorum , it's ok. Only drifted/healing copro could be interested by new revision, but they can only heal if a quorum exists, so nobody will really care about the revised manifest, it's more for observability.

rudy-6-4:
> * **Dotted bucket names:** [`peer_manifest_source.rs:276`](https://github.com/zama-ai/fhevm/blob/c0611a9fabfad9dfa6e8f0a99075de990eeafd47/coprocessor/fhevm-engine/consensus-detector/src/manifest_consensus/verification/peer_manifest_source.rs#L276) parses `s3://operator.manifests/prefix` as bucket `operator`. Reproduced with the extracted parser. Preserve the full bucket name and cover dotted names in parser tests.

Fixing.

Eikix (commented):

Follow-up at `fe6d7e7` against #3557. The canonical-descriptor comparison and ordinary dotted-bucket fixes are present. The older-revision concern is withdrawn as a blocker under the clarified contract: older independently signed observations can still establish quorum.

No additional production blocker was confirmed. Before approval, fix the stale `finding.version` query in `concurrent_one_drifter_replay_resolves_exact_handle_findings` ([CI: 134 passed, 1 failed](https://github.com/zama-ai/fhevm/actions/runs/35664868972/job/106548301250)) and the missing chart version bump, then rerun the checks.

Downstream contract worth preserving: a drift finding records disagreement with a peer; it does not necessarily mean the local result differs from the quorum. Containment must account for that distinction.

Validation: source review and CI-log inspection; no full E2E run.

PR #3911, opened 2026-09-10
feat(kms-connector): proxy implementation

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/1955

### Summary

This PR implements the KMS Connector `proxy` binary, a TLS-terminating reverse proxy built on Cloudflare's Pingora that sits between the client (i.e. Relayer for now) and the Connector endpoint replicas. The proxy authenticates the client, restricts traffic to the three public `v1` routes and load-balances requests round-robin across the configured endpoints.

### What it does

TLS is enforced via a PEM certificate and key validated at startup (using a file path).

The proxy rejects any request that does not carry a valid `Authorization: Bearer <api-key>` header. Only the SHA-256 digest of the key is configured, and the comparison is constant-time.

Routing: `POST` on the public and user decryption routes and `GET` on the version route are forwarded, anything else gets a `404` or a `405` with an `Allow` header.

Request bodies are capped by `max_body_bytes` and rejected with `413` when exceeded.

Before forwarding, hop-by-hop headers, `Connection`-nominated headers, `x-forwarded-*` headers and the client's `Authorization` are stripped. Endpoint responses are passed through untouched.


kms-connector/config/proxy.toml
```diff
@@ -0,0 +1,57 @@
...
+
+# Bind address of the proxy (optional, defaults to "0.0.0.0:8443")
+# ENV: KMS_CONNECTOR_PROXY_BIND_ADDRESS
+# bind_address = "0.0.0.0:8443"
+
+# SHA-256 digest of the API key the relayer sends as `Authorization: Bearer <key>` (required).
+# The plaintext key lives in a secret shared with the relayer; only its digest is configured here.
+# Compute it with: printf '%s' "$API_KEY" | sha256sum
+# ENV: KMS_CONNECTOR_API_KEY_DIGEST
+api_key_digest = "0x9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08"
```
dvdplm: Should we really provide this in the example?

kms-connector/config/proxy.toml
```diff
@@ -0,0 +1,57 @@
...
+
+# `host:port` addresses of the Connector endpoint replicas (required).
+# Can be a Kubernetes ClusterIP Service (kube-proxy balances), a docker-compose service name, or
+# plain IPs. Hostnames are resolved once at startup.
+# ENV: KMS_CONNECTOR_ENDPOINT_ADDRESSES (comma-separated list, e.g. "endpoint-1:8080,endpoint-2:8080")
+endpoint_addresses = ["kms-connector-endpoint:8080"]
```
dvdplm: Is there a way to let the service pick the port itself? Like, what if I don't care, or what if the open ports are dynamic on the host? Maybe not a concern at all, dunno.

dvdplm: Maybe we should show an example with two hosts, `["kms-connector-endpoint:8080", "kms-connector-edpt-2:9090"]`?

kms-connector/config/proxy.toml
```diff
@@ -0,0 +1,57 @@
...
+# shutdown_grace_period = "60s"
+
+# TLS configuration of the proxy (required)
+[tls_config]
+# Path of the PEM certificate chain
+# ENV: KMS_CONNECTOR_TLS_CONFIG__CERT_PATH
+cert_path = "/etc/kms-connector/proxy/tls.crt"
+# Path of the PEM private key
+# ENV: KMS_CONNECTOR_TLS_CONFIG__KEY_PATH
+key_path = "/etc/kms-connector/proxy/tls.key"
```
dvdplm, line 58: Honest q: is this "proper"? I think example config files should also be "teaching tools" for operators, nudging unfamiliar people in the right direction and helping them avoid silly mistakes. Is it a good practice to keep a private key in `/etc/my-app/`? Hardly right? Or is it ok? not sure tbh.

kms-connector/crates/proxy/src/bin/proxy.rs
```diff
@@ -33,14 +28,15 @@ async fn run() -> anyhow::Result<()> {
...
             debug!("{config:?}");
-            init_otlp_setup(config.service_name.clone())?;
 
-            let cancel_token = CancellationToken::new();
-            set_task_limit(config.task_limit);
-            install_signal_handlers(cancel_token.clone())?;
+            // The tonic OTLP exporter spawns its channel worker with `tokio::spawn` at build time,
+            // so a runtime must be entered here. The worker then runs on this runtime's threads,
+            // so it must stay alive for the whole process lifetime.
+            let runtime = tokio::runtime::Runtime::new()?;
```
dvdplm: How do the two runtimes avoid hogging CPU cores from each other? Do we need to adjust the tokio thread pool size for the OTLP exporter so that pingora gets enough of the machine's resources? Is it not possible to spawn tasks on pingora's runtime instead?

kms-connector/crates/proxy/src/core/auth.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Extracts the token of a `Bearer <token>` credential.
+fn bearer_token(header: &HeaderValue) -> Option<&str> {
+    let value = header.to_str().ok()?;
+    let (scheme, token) = value.trim().split_once(char::is_whitespace)?;
+    if !scheme.eq_ignore_ascii_case("bearer") {
+        return None;
+    }
+    let token = token.trim();
+    (!token.is_empty()).then_some(token)
+}
```
dvdplm, line 44: This is not constant time. I think that's ok though?

dvdplm: Random thought: I wonder if the `HeaderValue` is capped in length and if that is checked upstream? If it's not, maybe we should truncate `token` to, dunno, 4kb?, to stop a sender from sending in a 1Gb long token and force us to `Sha256::digest` a big blob of bytes.

kms-connector/crates/proxy/src/core/auth.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    fn header(value: &str) -> HeaderValue {
+        HeaderValue::from_str(value).unwrap()
+    }
+
+    #[test]
+    fn accepts_matching_bearer_token() {
+        assert!(verifier().verify(Some(&header(&format!("Bearer {KEY}")))));
+        assert!(verifier().verify(Some(&header(&format!("bearer {KEY}")))));
+        assert!(verifier().verify(Some(&header(&format!("Bearer   {KEY} ")))));
```
dvdplm, line 66: toss in some funny unicode whitepace here?

kms-connector/crates/proxy/src/core/config.rs
```diff
@@ -1,30 +1,88 @@
...
+}
+
+/// The TLS material used by the proxy to terminate TLS.
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[cfg_attr(test, derive(Serialize))]
+pub struct TlsConfig {
+    /// The path of the PEM certificate chain.
+    pub cert_path: PathBuf,
+    /// The path of the PEM private key.
+    pub key_path: PathBuf,
```
dvdplm, line 65: Does this have to be `pub`?

kms-connector/crates/proxy/src/core/config.rs
```diff
@@ -33,14 +91,173 @@ fn default_service_name() -> String {
...
     fn default() -> Self {
         Self {
+            bind_address: default_bind_address(),
+            tls_config: TlsConfig {
+                cert_path: PathBuf::from("/etc/kms-connector/proxy/tls.crt"),
+                key_path: PathBuf::from("/etc/kms-connector/proxy/tls.key"),
+            },
+            api_key_digest: "0x9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08"
+                .parse()
+                .unwrap(),
```
dvdplm: Hmm. Do we have to provide this in the `Default` impl. Convenient but easy to misuse?

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+        error: ErrorResponse,
+        extra_headers: &[(HeaderName, String)],
+    ) -> pingora::Result<()> {
+        let body = serde_json::to_vec(&error).unwrap_or_default();
+        let fixed_headers = [
+            (CONTENT_TYPE, "application/json".to_string()),
+            (CONTENT_LENGTH, body.len().to_string()),
+        ];
+        let mut header =
+            ResponseHeader::build(status, Some(fixed_headers.len() + extra_headers.len()))?;
```
dvdplm, line 58: Is this the right size hint? It's the number of headers, but shouldn't it be the number of bytes? So `extra_headers.iter().sum(|(n,s)| n.len() + s.len())` or something?

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+        for (name, value) in fixed_headers.iter().chain(extra_headers) {
+            header.insert_header(name.clone(), value.as_str())?;
+        }
+        session
+            .write_response_header(Box::new(header), false)
+            .await?;
+        session
+            .write_response_body(Some(Bytes::from(body)), true)
+            .await
```
dvdplm, line 67: Do we have to await these? It feels like it should be possible to "fire and forget" these (like: if we can't write the header for  some reason, do we really care?).

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+    fn new_ctx(&self) -> Self::CTX {
+        RequestContext {
+            started_at: Instant::now(),
+            route: None,
+            endpoint: None,
+            body_bytes: 0,
+        }
+    }
+
+    // No downstream module: in particular no response compression, bodies must stay untouched.
```
dvdplm: curious: why do we need uncompressed responses?

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+            .headers
+            .get(CONTENT_LENGTH)
+            .and_then(|v| v.to_str().ok())
+            .and_then(|v| v.parse::<usize>().ok());
+        if declared_length.is_some_and(|len| len > self.config.max_body_bytes) {
+            let error = ErrorResponse::new(
+                ErrorCode::Malformed,
+                format!(
+                    "request body exceeds the {} bytes limit",
+                    self.config.max_body_bytes
```
dvdplm: Not sure we should reveal our configuration to randos on the internet; just say "body too big". If you remove that you can use a `const` here and save on an allocation too.

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+            self.respond_with_error(session, StatusCode::BAD_REQUEST, error, &[])
+                .await?;
+            return Ok(true);
+        }
+
+        Ok(false)
+    }
+
+    // Checks the body cap on chunked requests.
+    async fn request_body_filter(
```
dvdplm, line 160: Does this run automatically? At which point? Above we're checking the declared length by parsing the header, but if we're going to count the body bytes here anyways, maybe we can check that the declared length matches here instead?

Like, instead of:
1. check declared len against `self.config.max_body_bytes`
2. bail if too big
3. read the whole body
4. check actual bytes against `self.config.max_body_bytes`
5. bail if too big

…we only do this:

1. read the whole body
2. check actual bytes against `self.config.max_body_bytes`
3. bail if too big

Wouldn't that be ok? Like, if I'm malicious I'm surely going to lie in the `CONTENT_LENGTH` header anyway and force us to read the whole body yeah?

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+    ) -> pingora::Result<()> {
+        if let Some(chunk) = body {
+            ctx.body_bytes += chunk.len();
+            if ctx.body_bytes > self.config.max_body_bytes {
+                return abort_with(ErrorResponse::new(
+                    ErrorCode::Malformed,
+                    format!(
+                        "request body exceeds the {} bytes limit",
+                        self.config.max_body_bytes
+                    ),
```
dvdplm: Same  as above: don't reveal our config and save an alloc

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+    ) -> pingora::Result<Box<HttpPeer>> {
+        let Some(backend) = self
+            .endpoint_balancer
+            .select(SELECT_KEY, MAX_SELECT_ITERATIONS)
+        else {
+            return abort_with(ErrorResponse::new(
+                ErrorCode::Overloaded,
+                "no endpoint available",
+                None,
+            ));
```
dvdplm, line 193: This is harsh, but maybe fair. Should we include a `Retry-After` header though?

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+        };
+        ctx.endpoint = Some(backend.addr.to_string());
+
+        let mut peer = HttpPeer::new(backend.addr, false, String::new());
+        peer.options.connection_timeout = Some(self.config.endpoint_connect_timeout);
+        peer.options.total_connection_timeout = Some(self.config.endpoint_connect_timeout);
+        peer.options.read_timeout = Some(self.config.endpoint_response_timeout);
+        peer.options.write_timeout = Some(self.config.endpoint_response_timeout);
+        peer.options.idle_timeout = Some(self.config.endpoint_idle_timeout);
+        Ok(Box::new(peer))
```
dvdplm, line 203: This surprised me a bit, the amount of setup. But perhaps it only runs once per session? Or we do this for every request? `HttpPeer::new` is cheap but not free it seems.

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+/// forwarding information and the HTTP/1.1 hop-by-hop headers (RFC 9110 §7.6.1).
+pub fn sanitize_headers(req: &mut RequestHeader) {
+    // Headers listed in `Connection` are hop-by-hop too.
+    let connection_headers = req
+        .headers
+        .get_all(CONNECTION)
+        .iter()
+        .filter_map(|v| v.to_str().ok())
+        .flat_map(|v| v.split(','))
+        .filter_map(|name| HeaderName::from_bytes(name.trim().as_bytes()).ok())
```
dvdplm: This allocates. No way around that?

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -0,0 +1,417 @@
...
+        ) => 0,
+        (_, ErrorSource::Downstream) => StatusCode::BAD_REQUEST.as_u16(),
+        (_, ErrorSource::Internal | ErrorSource::Unset) => {
+            StatusCode::INTERNAL_SERVER_ERROR.as_u16()
+        }
+    }
+}
+
+/// Removes the headers that must never reach the endpoint: the proxy credentials, client-supplied
+/// forwarding information and the HTTP/1.1 hop-by-hop headers (RFC 9110 §7.6.1).
```
dvdplm: I think this is a pretty expensive call, multiple allocations, clones and iterations. What is it actually buying us? "must never reach the endpoint" – why not?

dvdplm (commented):
I have no experience with `pingora` but I was  surprised by the amount of code required and a bit worried about how low-level it is. Not gonna lie, I'm a bit concerned about us taking on maintenance of a full-fat https proxy but I think the RFC outlines a design that requires a fair bit of flexibility in how we manage requests. I assume that we already looked at existing solutions (nginx?, envoy?).

kms-connector/crates/proxy/src/core/config.rs
eudelins-zama: I guess yes, or `pub(crate)` at least, which to me nuancing is pointless for binary crate but that is opinionated.

Fields of the config are all supposed to be used somewhere else in the code, otherwise we would not need them.

kms-connector/crates/proxy/src/bin/proxy.rs
eudelins-zama: Pingora only creates its runtimes inside `Server::run` and doesn't expose handles before that. The one hook would be a `background_service`, but that gets its own dedicated runtime anyway under the hood, and initialising the tracing subscriber from there we'd lose early logs and the error path in `main`.

But indeed `Runtime::new()` is oversized (spawns as many worker threads as machine's core), so a dedicated one-thread runtime seemed like the simpler option.

And for CPU usage, idle tokio workers cost nothing in CPU + AFAIU, the only work on that "tracing" runtime is the tonic gRPC channel for span export, which is I/O bound and periodic.

kms-connector/config/proxy.toml
eudelins-zama: > Is there a way to let the service pick the port itself?

These are the addresses of the `endpoints` behind the `proxy`, not the bind address of the proxy itself, so no there's no way to dynamically choose the port

kms-connector/config/proxy.toml
eudelins-zama: No strong opinion, bot claims `/etc/kms-connector/proxy/tls.key` is the idiomatic location for a containerized service

kms-connector/crates/proxy/src/core/auth.rs
eudelins-zama: > This is not constant time. I think that's ok though?

Yes, I don't think we care, constant time is when we want to hide the equality being checked at runtime. Here it's plain "bearer" equality check, that you know just by reading the code / disassembling the binary.

> Random thought: I wonder if the HeaderValue is capped in length and if that is checked upstream?

Pingora [checks](https://github.com/cloudflare/pingora/blob/0.9.0/pingora-core/src/protocols/http/v1/server.rs#L308) header size is below ~1MiB.

kms-connector/crates/proxy/src/core/auth.rs
eudelins-zama: `HeaderValue::to_str` enforces visible ASCII, so non-ASCII bytes never reach the trim/split logic at all

kms-connector/crates/proxy/src/core/config.rs
eudelins-zama: Nothing in production ever builds `Config::default()`, it is only for testing purpose as commented on top of the trait impl.
Unfortunately, I can't hide it behind `#[cfg(test)]` for integration test, would have to rely on feature flags if really want something like this

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: I'd say yes we care, we don't want to return an incomplete HTTP response (without headers) to the client, it would not make sense to me.
Headers contains the status code for ex, I don't think a random http client would know what to do just with the request body, and not the headers

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: Yep: [code](https://github.com/cloudflare/pingora/blob/0.9.0/pingora-http/src/lib.rs#L508-L514)

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: Bot overstated that this is a "must". I'll remove the override and simply rely on Pingora's default.

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: We can't read the whole body, we're just seeing "chunks" of the body at each `request_body_filter`.
See [docs](https://github.com/cloudflare/pingora/blob/0.9.0/pingora-proxy/src/proxy_trait.rs#L140-L148).

So we can't check equality. We could bail as soon as length is higher than what was announced in the `CONTENT_LENGTH` though. But that means we need to store that data + reject anyone who does not set this header...

> if I'm malicious I'm surely going to lie in the CONTENT_LENGTH header anyway and force us to read the whole body yeah?

Yep, that is why we check the limit of the body size itself, so we're good in such case anyway

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: The proxy has no idea when an endpoint comes back here, `select()` just returned nothing... 
Any `Retry-After` would be an arbitrary constant

kms-connector/crates/proxy/src/bin/proxy.rs
dvdplm: Maybe it's fine as it stands then. Ty for cehcking.

kms-connector/config/proxy.toml
dvdplm: Ofc, dumb q.

I guess the follow-up observation is that the configs must be kept aligned manually between `endpoint` instances and the proxy. It'd be neat to have some well-defined and controllable auto-discovery mechanism but that seems like a project on its own.

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: From [docs](https://github.com/cloudflare/pingora/blob/0.9.0/pingora-proxy/src/proxy_trait.rs#L55-L59) it seems it runs once per request to me.
But I think the cost should be negligible tbh...

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: Yes, it allocates.
The `collect()` is needed because `connection_headers`/`forwarded_headers` borrow `req.headers` immutably, while `remove_header` needs `&mut`, so we can't remove while iterating.

kms-connector/config/proxy.toml
dvdplm: Hmm. Not sure I buy the idea of using containerization as a security barrier. Like, if I install this on a bare metal machine, I'd expect the defaults to be reasonably secure; if I use it in a docker container I think it's fair to have the same expectation.

That said, I don't know if putting private keys in `/etc` is frowned upon or not. Shouldn't we think about this key as valuable and treat it sort of like we do with keys in `~/.ssh/`?

eudelins-zama: On bare metal server, I'd say `/etc/ssl`/`etc/<app>/ssl` is the kind of location I would expect.

And to be fair, I think `/etc/nginx` is the place you would expect config things for `nginx`, so why not `/etc/kms-connector-proxy`?

> I'd expect the defaults to be reasonably secure

I think wherever you put certs, there's the same level of security, what matters is file permissions and ownership.

kms-connector/crates/proxy/src/core/config.rs
dvdplm: What I was getting at was: do we really need to use the `PathBuf` in other places, or can we rewrite this so that using this path becomes internal.

kms-connector/crates/proxy/src/core/config.rs
dvdplm: Yeah, in the `kms` we have the `testing` feature for that purpose. Not very eleg nt but it does work. You could then perhaps have a `test_config` method that returns a fully configured struct, and avoid making it `Default`at all?

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: I agree it is complex. I think the goal was to be fully "compliant" with what is expected from a proxy in terms of HTTP RFC. Maybe we don't care much in our case, just strip the API key and main headers without the complex parsing and alloc is fine (will do).

On the cost though, I think it is negligible, one `Vec`, a few `HeaderName` clones and a couple of string splits are probably nanoseconds against the TLS handshake for ex...

kms-connector/crates/proxy/src/core/http_proxy.rs
dvdplm: So it's not setting the `CONTENT-LENGTH` header then? Because that's measured in bytes!

kms-connector/crates/proxy/src/core/http_proxy.rs
dvdplm: Fair enough. I find it a bit odd that writing headers and writing the body are two different calls. Sure, for a header-only response it's useful, but odd that there isn't a "flush this whole bag of bytes"-method that does it all.

kms-connector/crates/proxy/src/core/config.rs
eudelins-zama: > You could then perhaps have a test_config method that returns a fully configured struct, and avoid making it Defaultat all?

This does not allow the `..Default::default()` usage that is pretty convenient...

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: No, this is the size of the headers map from what I see in underlying pingora's code.

We write `CONTENT-LENGTH` ourselves in `respond_with_error`

kms-connector/crates/proxy/src/core/config.rs
eudelins-zama: Would be a bit inconsistent with other fields of the config, without clear benefits to me :thinking:

kms-connector/config/proxy.toml
eudelins-zama: Fixed bd1bf11609b7259d7ffeb448f7ed06a052fdd04a

kms-connector/crates/proxy/src/bin/proxy.rs
eudelins-zama: Small improvement 80ec0c9974b56685576bc1dcaf742aa677229168

kms-connector/crates/proxy/src/core/auth.rs
eudelins-zama: Test 3240805e096628b3dbd7bc1b9ec6a8dec1fc2338

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: Fixed 5a1a8a80184ce17f688a6693d19c6fa88a9a398b

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: Fixed 03cd1b4dbc0ffae23c31d18f26e3c80a26d39405

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: Fixed 03cd1b4dbc0ffae23c31d18f26e3c80a26d39405

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: Fixed ce8f2eb530332c6d599984c5c7ec7a49b3169b7b

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: Fixed ce8f2eb530332c6d599984c5c7ec7a49b3169b7b

kms-connector/crates/proxy/src/core/http_proxy.rs
dvdplm: Maybe we can legitimately end up here when the endpoint(s) restart? Either way I kind of agree with you and most likely the client will retry if they care regardless of `Retry-After`.

kms-connector/crates/proxy/src/core/auth.rs
dvdplm: Fair enough.

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: > Maybe we can legitimately end up here when the endpoint(s) restart?

You mean shutting down the proxy entirely in this case?
That would be the goal of the healthcheck endpoint implementation (not done yet) I guess :slightly_smiling_face:

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -21,10 +20,12 @@ use pingora::{
...
 use tracing::{error, info};
 
-/// The `X-Forwarded-*` headers set by the proxy (any inbound value is dropped first).
+// The `X-Forwarded-*` headers set by the proxy (any inbound value is dropped first).
 const X_FORWARDED_FOR: HeaderName = HeaderName::from_static("x-forwarded-for");
 const X_FORWARDED_PROTO: HeaderName = HeaderName::from_static("x-forwarded-proto");
-/// Hop-by-hop headers without a constant in the `http` crate.
+// Client-supplied forwarding header the proxy never sets but must not forward either.
+const X_FORWARDED_HOST: HeaderName = HeaderName::from_static("x-forwarded-host");
+// Hop-by-hop headers without a constant in the `http` crate.
```
dvdplm, line 28: What is a "hop-by-hop" header? And do we care that it's not in the http crate?

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -317,46 +318,29 @@ fn default_error_code(e: &Error) -> u16 {
...
-    ]
-    .into_iter()
-    .chain(connection_headers)
-    .chain(forwarded_headers)
-    .collect();
-
-    for name in to_remove {
-        req.remove_header(&name);
+    for name in &STRIPPED_HEADERS {
+        req.remove_header(name);
```
dvdplm, line 343: much much nicer. ty

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -99,9 +98,6 @@ impl ProxyHttp for Proxy {
         }
     }
 
-    // No downstream module: in particular no response compression, bodies must stay untouched.
-    fn init_downstream_modules(&self, _modules: &mut HttpModules) {}
```
dvdplm, line 103: nice

kms-connector/crates/proxy/src/core/http_proxy.rs
```diff
@@ -67,9 +69,14 @@ impl Proxy {
         &self,
         session: &mut Session,
         status: StatusCode,
+        allow: Option<&Method>,
```
dvdplm, line 72: What was this for again? Or is this a new thing since I looked last perhaps?

kms-connector/crates/proxy/tests/integration_tests.rs
```diff
@@ -0,0 +1,480 @@
...
+        .unwrap()
+}
+
+async fn wait_for_listener(addr: SocketAddr) {
+    let deadline = Instant::now() + Duration::from_secs(10);
+    while tokio::net::TcpStream::connect(addr).await.is_err() {
+        assert!(Instant::now() < deadline, "{addr} did not come up");
+        tokio::time::sleep(Duration::from_millis(50)).await;
+    }
+}
```
dvdplm, line 480: I like your test tooling

kms-connector/config/proxy.toml
dvdplm: Fair enough!

kms-connector/crates/proxy/src/core/config.rs
dvdplm: I don't love that we do not have compile-time enforcment that non-test code must not use `Default::default`. I think safe defaults should be the norm. I'm letting this slide  here because it's a "Stage 1" implementation but yeah, not ideal.

kms-connector/crates/proxy/src/core/http_proxy.rs
dvdplm: Benchmarks will tell if this  is costly or not, and we can re-assess if it's an actual problem.

kms-connector/crates/proxy/src/core/http_proxy.rs
dvdplm: > You mean shutting down the proxy entirely in this case?

No, the other way around: the proxy stays up inviting clients to retry, or perhaps sleeping for a few seconds and then retries to check if the configured  endpoints are now available. 

But I don't think you need that for the time being (if ever). Was just an idea.

kms-connector/crates/proxy/src/core/http_proxy.rs
dvdplm: Could be. Benchmarks will tell.

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: If you send a wrong method (let's say `PUT`) to a given endpoint, for ex the `v1/version` that only accept `GET`, you're supposed to return `405 Method Not Allowed` + set the `Allow` header with the list of methods you accept for that endpoint

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: "Hop-by-hop" header is a real term (see [RFC 9110 §7.6.1](https://datatracker.ietf.org/doc/html/rfc9110#name-connection)) in the HTTP world, not an AI slop jargon :innocent: .

AFAIU for a proxy, this means "the header meant just for the connection between the client and the proxy that must not be forwarded to the server behind the proxy"

kms-connector/crates/proxy/src/core/http_proxy.rs
dvdplm: I see. Hmm. I guess in an adversarial setting this raises the question about how much meta-data we want to reveal. I'd probably err on the side of caution and not tell the outside world what we accept.

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: > And do we care that it's not in the http crate?

Keep-Alive and Proxy-Connection might not be in `http` crate, but still cited in the RFC above, so I guess the `http` crate is not 100% exhaustive.
Bot insist on keeping them stripped, which I makes sense to me and I guess can't hurt...

dvdplm: TIL!

kms-connector/crates/proxy/src/core/config.rs
eudelins-zama: Tracking issue: https://github.com/zama-ai/fhevm-internal/issues/2031

kms-connector/crates/proxy/src/core/http_proxy.rs
dvdplm: I wonder what it's using the number of headers for? Seems pretty useless to a server, at least without also knowing the length of each header.

kms-connector/crates/proxy/src/core/http_proxy.rs
eudelins-zama: Hmm this is the expected behavior from [web standard](https://developer.mozilla.org/fr/docs/Web/HTTP/Reference/Status/405), so I think it's fine to let the client know about this.

Tbf, a super-simple scripts bruteforcing every methods would reveal this anyway...

dvdplm (approved):

eudelins-zama:
@mergifyio queue

nboisde (approved):

eudelins-zama:
@mergifyio refresh

PR #3914, opened 2026-09-11
fix(coprocessor): read the stack role live instead of the startup copy

PanGan21:
## Summary

Services resolved `gcs_mode` once at startup and kept the bool on long lived structs. `StackMode` flipps at cutover, but the copies did not, so an ex-green service that became live still acted as green at the next proposal.

### Changes

- host-listener: `Database` holds the shared `StackMode` and reads it for the synthetic-ops gate, the write guards and the pool.
- tfhe-worker and zkproof-worker: run the stack-version listener like the other services and read the shared `StackMode` in the activation gates, write guards and GCS-only queries.
- upgrade-controller: gates its handlers and reconcile on the live mode, so an ex-green controller no longer rejects the next release's proposal or re-creates the dropped GCS schema on a timeout.

Related: zama-ai/fhevm-internal#1942


coprocessor/fhevm-engine/tfhe-worker/src/tfhe_worker.rs
```diff
@@ -2205,6 +2207,30 @@ async fn run_tfhe_worker_inner(
...
+        None,
+    )
+    .await?;
+    {
+        let listener_mode = stack_mode.clone();
+        tokio::spawn(async move {
+            if let Err(err) = run_stack_version_listener(
+                listener_pool,
+                listener_mode,
+                tokio_util::sync::CancellationToken::new(),
```
goshawk-3: Shouldn't we pass here the tfhe-worker cancellation token?

PanGan21:
@goshawk-3 I am wondering if we need the same for consensus-detector. What should the consensus-detector do after the cutover? 
Currently there’s no explicit runtime blue/green role distinction in the consensus detector and I feel this might be problematic for the next B/G release

Eikix (commented):
[AUTOMATED]

[P2] Recover the stack role when the cutover notification is missed.

The new call sites rely on `run_stack_version_listener`, but that shared helper only reconciles after a notification: it does not read the role after registering LISTEN, and a failed reconciliation is logged without retry.

If cutover commits between the startup role read and LISTEN registration, the service keeps its old Green role. A failed role read while handling the notification has the same outcome. This can preserve the stale-role behavior this PR fixes during consecutive upgrades. This is a code-backed recovery gap, not a reproduced rollout failure; the helper predates this PR.

Suggested fix: register LISTEN, immediately reconcile the role, and periodically reconcile to recover missed notifications or failed reads. Add tests for cutover before listener startup and recovery after a failed reconciliation, without emitting another notification. The added test repeatedly emits notifications, so it does not cover these cases.

Reviewed head: 147f8979a2d67b9de5839241e38995ae4d17ebe6. Relevant helper: coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs:154-180.

PanGan21:
> [AUTOMATED]
> 
> [P2] Recover the stack role when the cutover notification is missed.
> 
> The new call sites rely on `run_stack_version_listener`, but that shared helper only reconciles after a notification: it does not read the role after registering LISTEN, and a failed reconciliation is logged without retry.
> 
> If cutover commits between the startup role read and LISTEN registration, the service keeps its old Green role. A failed role read while handling the notification has the same outcome. This can preserve the stale-role behavior this PR fixes during consecutive upgrades. This is a code-backed recovery gap, not a reproduced rollout failure; the helper predates this PR.
> 
> Suggested fix: register LISTEN, immediately reconcile the role, and periodically reconcile to recover missed notifications or failed reads. Add tests for cutover before listener startup and recovery after a failed reconciliation, without emitting another notification. The added test repeatedly emits notifications, so it does not cover these cases.
> 
> Reviewed head: [147f897](https://github.com/zama-ai/fhevm/commit/147f8979a2d67b9de5839241e38995ae4d17ebe6). Relevant helper: coprocessor/fhevm-engine/fhevm-engine-common/src/versioning.rs:154-180.

Thanks, quite an edge case. This has been fixed now in this commit: https://github.com/zama-ai/fhevm/pull/3914/commits/9cd9d23184ffc9e9399be8ccc88876c60bb9410c

Eikix (approved):
[AUTOMATED] LGTM

PR #3916, opened 2026-09-11
fix(coprocessor): move Gateway transport to HTTP

antoniupop:
This hotfix has 3 main parts:
 - move from WSS to HTTP in tx-sender
 - update retry accounting for proof responses to avoid exhaustiung the budget on infra failures.
 - sanitise error logs some pre-existing but which the http migration makes worse (this can be trimmed down if it's too broad, but feels wrong to leave this out)

### DIFF
--- a/charts/coprocessor/Chart.yaml
+++ b/charts/coprocessor/Chart.yaml
@@ -1,6 +1,6 @@
 name: coprocessor
 description: A helm chart to distribute and deploy Zama fhevm Co-Processor services
-version: 0.12.1
+version: 0.12.2
 apiVersion: v2
 keywords:
   - fhevm
--- a/charts/coprocessor/templates/coprocessor-tx-sender-deployment.yaml
+++ b/charts/coprocessor/templates/coprocessor-tx-sender-deployment.yaml
@@ -150,7 +150,7 @@ spec:
               {{- end }}
             {{- end }}
             {{- end }}
-            {{- with .Values.commonConfig.gatewayUrl }}
+            {{- with (.Values.txSender.config.gatewayUrl | default .Values.commonConfig.gatewayUrl) }}
             - name: GATEWAY_URL
               {{- if .value }}
               value: {{ .value | quote }}
--- a/.github/workflows/test-suite-e2e-tests.yml
+++ b/.github/workflows/test-suite-e2e-tests.yml
@@ -324,6 +324,23 @@ jobs:
           fi
           ./fhevm-cli up "${args[@]}" --dry-run
 
+      - name: Pull database image with retries
+        shell: bash
+        run: |
+          # Pre-pull the image from the compose file so a transient registry
+          # failure does not abort stack startup after other services start.
+          image="$(awk '$1 == "image:" { print $2; exit }' test-suite/fhevm/docker-compose/database-docker-compose.yml)"
+          for attempt in 1 2 3; do
+            if docker pull "$image"; then
+              break
+            fi
+            if [ "$attempt" -eq 3 ]; then
+              echo "Database image pull failed after three attempts" >&2
+              exit 1
+            fi
+            sleep "$((attempt * 10))"
+          done
+
       - name: Boot fhevm Stack
         working-directory: test-suite/fhevm
         run: |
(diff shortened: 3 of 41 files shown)


rudy-6-4 (approved):
LGTM

Eikix (commented):
[AUTOMATED]

Code review of the transaction-sender HTTP transport, proof retry accounting, diagnostics and endpoint wiring. These files were reviewed at `02b4e877` and are unchanged at `9be9730e`. This assessment is based on source inspection, including the locked Alloy/reqwest implementations; it makes no assessment of the author's test execution or the later CI/audit-policy changes.

No blocking finding in these scoped changes. The HTTP provider retains the existing submission flow, bounds requests, and disables redirects and automatic transport retries. Recognized infrastructure failures preserve proof eligibility while returning an error to the existing backoff loop. Existing contract/configuration-error handling takes precedence. Diagnostic formatting preserves the typed errors used for recovery, and the sender-specific HTTP endpoint remains separate from the listener's WebSocket endpoint.

Optional follow-up: reuse the infrastructure-error classifier for ciphertext submissions' existing unlimited-retry path. `add_ciphertext.rs` still uses Alloy 1.1.2's narrower `is_retry_err()`, so an HTTP connection failure during gas estimation consumes `txn_limited_retries_count`. With a smaller configured `--add-ciphertexts-max-retries`, repeated failures can make an unsent digest ineligible after Gateway recovery. The default is `i32::MAX`, so this is non-blocking for the minimal hotfix under that configuration.

Eikix (approved):
[AUTOMATED]

Approved following the code review and re-review. Reviewed the runtime at `0cfb3c81`; the current head `d4087082` adds rollout documentation only.

No blocking finding in the reviewed HTTP transport, proof retry preservation, scheduling, or diagnostic changes. The cooldown/polling interaction can add recovery latency but is accepted as a non-blocking tradeoff for this minimal hotfix. The earlier ciphertext retry-classification suggestion remains optional under the default retry limit.

This approval reflects source review, not an independent assessment of test execution or deployment validation.

nboisde (commented):
Not sure why we are changing relayer tests here ?
Regarding tx manager since its just a provider change, nothing to add there.
On the copro pure internal logic for queries, not able to comment there.

antoniupop:
> Not sure why we are changing relayer tests here ?
> Regarding tx manager since its just a provider change, nothing to add there.
> On the copro pure internal logic for queries, not able to comment there.

No relayer runtime behavior is changed - the tesst were added to check SDK/client retries actually recover after a sender outage. They confirm that the relayer readiness window ends in a terminal 503, the SDK does not automatically resubmit, and an explicit fresh request can succeed after recovery.
I can move the relayer/SDK tests out if you prefer.

rudy-6-4 (approved):

antoniupop:
@mergifyio queue

charts/coprocessor/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: coprocessor
 description: A helm chart to distribute and deploy Zama fhevm Co-Processor services
-version: 0.12.1
+version: 0.12.2
```
tawadaa, line 3: We are currently at `0.13.8` https://github.com/zama-zws/gitops/blob/main/environments/zws-dev/fhevm-coproc/gw-coprocessor-1.yaml

I think there will be a problem with chart release, as it's only released when merged to main

charts/coprocessor/templates/coprocessor-tx-sender-deployment.yaml
```diff
@@ -150,7 +150,7 @@ spec:
               {{- end }}
             {{- end }}
             {{- end }}
-            {{- with .Values.commonConfig.gatewayUrl }}
+            {{- with (.Values.txSender.config.gatewayUrl | default .Values.commonConfig.gatewayUrl) }}
```
tawadaa, line 153: why don't we simply change the value here instead of changing the chart
https://github.com/zama-zws/gitops/blob/main/values/gw-blockchain/gw-coprocessor/values-zws-dev.yaml#L6

tawadaa (approved):

isaacdecoded (approved):

paulo-zama (approved):

PR #3917, opened 2026-09-11
feat(host-contracts): scalar operand handling improvement

jatZama:
Ports the[ 0.13 PR from release-0.13.x-scalar](https://github.com/zama-ai/fhevm/pull/3844) to `release/0.14.x`.
Extends scalar range checks to fheMulDiv and type validation to its library overloads. Bumps the executor to v0.6.0 with reinitializeV6() (initializer version 7), and updates tests, generated artifacts, and the Solidity package to 0.14.1.


host-contracts/test/fhevmExecutor/fhevmExecutor.t.sol
```diff
@@ -470,7 +476,34 @@ contract FHEVMExecutorTest is SupportedTypesConstants, Test {
         assertEq(fhevmExecutor.getInputVerifierAddress(), inputVerifierAdd);
         assertEq(fhevmExecutor.getACLAddress(), aclAdd);
         assertEq(fhevmExecutor.getHCULimitAddress(), hcuLimitAdd);
-        assertEq(fhevmExecutor.getVersion(), string(abi.encodePacked("FHEVMExecutor v0.5.0")));
+        assertEq(fhevmExecutor.getVersion(), string(abi.encodePacked("FHEVMExecutor v0.6.0")));
+    }
+
+    function test_UpgradeFromV5PreservesStateAndConsumesReinitializer() public {
```
obatirou, line 482: nit: why preserve state?
I am not sure why test the counter is preserved when doing an upgrade. Seems this part of the test could be dropped

host-contracts/test/fhevmExecutor/fhevmExecutor.t.sol
```diff
@@ -508,7 +541,7 @@ contract FHEVMExecutorTest is SupportedTypesConstants, Test {
         address sender = address(123);
 
         bytes32 lhs = _generateMockHandle(FheType(fheType));
-        bytes32 rhs = _generateMockHandle(FheType(fheType));
+        bytes32 rhs = scalarBool ? bytes32(uint256(1)) : _generateMockHandle(FheType(fheType));
```
obatirou, line 544: nit: 1 is always used as scalar in those fuzz tests.

host-contracts/test/fhevmExecutor/fhevmExecutor.t.sol
jatZama: The state must be preserved during upgrade unless reinitializer touches said storage variable, which is not the case for the counter.

host-contracts/test/fhevmExecutor/fhevmExecutor.t.sol
jatZama: yes this is because it fits in all types, even for ebool, for simplicity.

host-contracts/test/fhevmExecutor/fhevmExecutor.t.sol
jatZama: btw this is not something new from this PR, same test existed in the original 0.13 PR

host-contracts/test/fhevmExecutor/fhevmExecutor.t.sol
jatZama: Same here, we had the exact same test in the original 0.13 PR btw

obatirou (approved):

tawadaa (approved):

PR #3922, opened 2026-09-11
fix(coprocessor): move Gateway transport to HTTP

antoniupop:
Forward port to main of #3916


PanGan21 (approved):

antoniupop:
@mergifyio queue

nboisde (approved):
Same remark for v13, I would move ts tests from the relayer folder, but we can move thoses campaign scripts later.

PR #3924, opened 2026-09-12
chore(coprocessor): upgrade to tfhe-rs 1.8.1

PanGan21:
### Summary

Upgrades the coprocessor from tfhe-rs 1.6.3 to 1.8.1.

### Breaking change

- Shifting by an amount equal to or larger than the type's width now returns `0`. Before, the amount was wrapped around e.g. `x << 70` on a 64-bit value behaved like `x << 6`. 
- Rotations are unchanged.

The e2e operator tests confirm this. With the old expectations still in place, 11 of 85 failed, and every failure named `shl` or `shr`:

[run with old expectations →](https://github.com/zama-ai/fhevm/actions/runs/34685466958) 74 passing, 11 failing

```
shl(euint8, 255) scalar rhs: got 0, want 128
shr(euint8, 255) scalar rhs: got 0, want 1
```

Setting `OVERSHIFT_RETURNS_ZERO = true` in `shiftSemantics.ts` updates those expectations, and everything passes:

[run after the flip →](https://github.com/zama-ai/fhevm/actions/runs/34692511046) 85 passing

Related: https://github.com/zama-ai/fhevm-internal/issues/1990


coprocessor/fhevm-engine/fhevm-engine-common/src/keys.rs
```diff
@@ -105,6 +106,7 @@ impl FhevmKeys {
             None, // noise squashing compression key excluded
             re_randomization_keyswitching_key,
             None, // oprf key excluded
+            transciphering_key,
```
nsarlin-zama, line 109: Not really sure about how this function is used, but if you exclude the oprf key you might want to exclude the transciphering key too? Since at the moment the transciphering key is "just" an oprf key

PanGan21: You're right, I should follow whatever we do with the oprf key.
While we're here though, should we still be excluding the oprf key at all? What is your view on that?

nsarlin-zama: I'm not exactly sure what this function is about to be fair, is it for testing purpose? 

New keys generated with tfhe-rs will have the dedicated oprf key but AFAIK when the prod keys were generated it didn't have a dedicated key and used the compute key instead. So if you want to mimic exactly prod behavior you might want to exclude it. But after the next keygen is done it will have to be re-enabled I guess.

PanGan21: Yes i believe it is test/dev only.
Do you maybe have any information when we will have to support a dedicated key?
cc @antoniupop  and @dd23 on this maybe you know the planning on when we will move to a dedicated OPRF key

PanGan21: I excluded the transciphering_key as suggested here: https://github.com/zama-ai/fhevm/pull/3924/commits/2159435ba8e59b5707e8b0427c0154b0d9c04f6c

PanGan21: I think regarding the previous discussion about OPRF key, what we do in production is just pass it through. When a key arrives from the KMS we keep whatever OPRF key it carries, and same when the workers load it from the DB:

https://github.com/zama-ai/fhevm/blob/fa40f7af7e907465dfd5ec66e08515341ef2e463/coprocessor/fhevm-engine/host-listener/src/kms_generation/sks_key.rs#L164
https://github.com/zama-ai/fhevm/blob/fa40f7af7e907465dfd5ec66e08515341ef2e463/coprocessor/fhevm-engine/fhevm-engine-common/src/db_keys.rs#L380

Today KMS keys don't have one, so tfhe-rs falls back to the compute key. Once the migration produces keys with a dedicated OPRF key, we'd pick it up automatically without any code change.

Is that the behaviour we would expect?

nsarlin-zama: I think it will be enabled during the next keygen, along with the transciphering key

PanGan21: Then the code seems that it handles it correct right? At least for production usage

antoniupop: My understanding is that we will generate it in the key update for compressed xof keyset and we'll start using it when we switch (in a blue-green upgrade process). Maybe I'm misunderstanding, but it should probably not be excluded since it's only going to be picked up when it's available, and that would be when we actually start using the updated keyset.

antoniupop: Starting to use this key is a consensus hazard, so it needs to be done in a BG.

PanGan21: This file is only test file which excludes the OPRF key, it generates the local keys for dev stacks, it's not used in any deployment. Maybe we should keep the local setup until prod matches local again.
On the production path I think we already do what you describe. We copy the OPRF key untouched, so the dedicated key gets picked up automatically when the new keyset is switched in KMS.
What do you think?

antoniupop: I think the test needs to be in sync with where we expect prod to go, so if we'll be activating the key in v0.15.0->v0.15.1 then we should have it active in the tests now. If it's not, then we can't pick up potential issues post key update.

PanGan21: Ok this has been changed now in this commit for both keys: https://github.com/zama-ai/fhevm/pull/3924/commits/4eecab72ef905e81a0458f74bef5f545a04f1800

Worth noting that in order to catch issues before the migration we'd need a test that generates a key with the dedicated OPRF key and runs FHE.rand through it which is probably something to be done within the guild working on the migration itself and not in the upgrade pr.

antoniupop (approved):

geoxel (approved):

PR #3926, opened 2026-09-14
ci(test-suite): accept 0.14 and 0.15 ActivateKey shapes in the preview-env keygen wait

Eikix:
## Why

`preview-env-deploy` lets a dispatch pin every image to an older release through `overrides`, for example `host_contracts_version=v0.14.1`. The deploy-time keygen job still runs main's wait snippet, which filters logs on the 0.15 `ActivateKey(keyId, existingKeyId, urls, digests)` signature. A 0.14 contract emits `ActivateKey(keyId, urls, digests)`, so the wait never matches, the job times out after 15 minutes, and the deploy stops before the relayer and test-suite, even though the key was activated.

## What

Match either `ActivateKey` topic and decode with the matching interface. `keyId` and `kmsNodeStorageUrls` exist in both shapes, so the code after the wait is unchanged. `ActivateCrs` has the same signature in 0.14 and 0.15 and needs nothing.

## Checked

Ran the extracted snippet against a fake provider returning an encoded log of each shape on ethers 6.14.3. Both decode and write the same `addresses/.env.keymaterial` as before. Not yet run on a live pinned launch.


fegmorte (approved):
LGTM

PR #3931, opened 2026-09-14
fix(sdk): accept relayer's new no_attestation_consensus 500 label

geoxel:
zama-ai/fhevm#3650 adds an off-chain Coprocessor attestation consensus check to the relayer's readiness path (RFC-023) and wires up two error labels on the async decrypt status endpoints: no_attestation_consensus (500) and gateway_not_reachable (503).

gateway_not_reachable was already accepted by the SDK's RelayerApiError503 type/guard. no_attestation_consensus was not: RelayerApiError500 only recognized `internal_server_error`, so a genuine 500 response carrying the new label would fail the SDK's own schema assertion and surface as a generic
RelayerResponseInvalidBodyError instead of a proper RelayerResponseApiError with the relayer's actual label/message.

Widen RelayerApiError500's label union and the corresponding runtime guards (RelayerApiError500, RelayerResponseFailed's dispatcher and its unsupported-label list) to accept both labels.


0xalexbel (approved):

geoxel:
@mergifyio queue

PR #2023, opened 2026-02-24
fix(library-solidity): fix npm package link

poppyseedDev:

PR #3934, opened 2026-09-14
refactor(sdk): prepare Solana SDK for maintainer handover

Eikix:
The Solana SDK previously exposed EVM-only capabilities, required consumers to wire account reads separately, and returned raw public-decryption claims without local authentication. This change gives it an explicit native client surface, reuses shared SDK infrastructure, and migrates the demo and CLI to the published consumer API.

- Bind one supplied `@solana/kit` RPC to each client and reuse it for SDK-owned reads, including permit revocation state.
- Expose only supported options and capabilities. Align `encryptValue(s)`, `generateZkProof`, and `decryptValue(s)` with the EVM API and shared result types.
- Compose proof generation/submission into encryption. Authenticate typed public decryption against fresh host configuration and the requested live KMS context, while retaining raw certificates for on-chain verification.
- Reuse shared HTTP/auth/configuration and canonical signing helpers. Normalize cancellation, retain bounded private-decrypt deadlines, and reduce the main Solana runtime exports from 95 to 45.
- Remove private SDK imports and fabricated EVM proof contexts from Solana consumers. Preserve Solana wire/authorization behavior and custom host-program configuration.
- Repair the exercised build/bootstrap path: defer SDK imports until provisioning, include required IDLs in the runner image, support the exact cached KMS image when remote inspection is unavailable, and align database bootstrap with main.

Validation of production implementation `127e9f6b8`:

- Full SDK production build, codegen consistency, formatting, lint, types, ESM/CJS and WASM packaging passed. 1,378 SDK unit tests, 250 demo tests and 475 CLI/harness tests passed; demo and CLI type checks passed.
- Installed-tarball consumer passed strict TypeScript and runtime import checks.
- Fresh live demo smoke passed 17 assertions, including authenticated typed public decryption and exact private payout. Seven further live scenarios passed: token burn/redeem/disclosure, current and historical decrypt, two-holder transfer, 32 dependent FHE steps, wallet delegation and real Squads 2-of-3 delegation.
- Burner-wallet browser acceptance covered public deposit, private redeposit, yield accrual, partial/full redemption, reveal/hide, invalid amounts, refresh, disconnect/reconnect and copy controls. Final observed state was zero shares and 1,177 confidential USDC. All ten metrics targets were up; inspected logs and a correlated 15-span KMS trace supported the successful flows.
- Independent correctness and maintainability reviews completed; confirmed findings were fixed and rechecked.
- Follow-up `0175ece90` only clarifies the EIP-712 signing-domain name in a test. The unchanged dead-surface guard, 26 public-decryption tests, formatting and SDK type checking pass.

CI is pending on the rebased head. This PR is open for maintainer review. External wallet integrations and a live threshold KMS deployment are not claimed by the burner-wallet/local central-KMS acceptance.

Private decryption still requires explicit trusted party-id/epoch configuration. Public timeout follows the EVM HTTP/poll budget; private decryption retains an overall deadline. The complete parity audit is preserved in a PR comment.

Tracks zama-ai/fhevm-internal#2024 and zama-ai/fhevm-internal#2013. Targets `feature/solana`; zama-ai/fhevm-internal#2023's linker redesign is excluded.

Review follow-up (`da168eb8f`): same-program generated PDA defaults now honor custom deployments for randNonce and KmsContext as well as hostConfig. Regression tests preserve argument-derived seeds. Client factories pass chain explicitly, the redundant public decorator extension and hidden core chain property are removed, and the host barrel reuses the curated generated instruction index. Rust-produced permit bytes and malformed Borsh boolean encodings are pinned by regression tests.

The low-level permit/user-decrypt export removals are intentional breaking changes within the Solana PoC. No compatibility aliases are retained; consumers use the supported client actions.

Follow-up validation: full SDK production build and 1,383 unit tests pass; 250 demo tests, demo/CLI type checks, and two custom-program PDA tests and the dead-surface guard pass. Browser/live acceptance was not repeated for this follow-up. CI must complete on the new head.

Rebased onto `feature/solana` after #3954 merged. The Solana SDK source tree is unchanged by the rebase; main's shared-core fixes are retained. Duplicate transport and pinned-SBF-toolchain fixes are already in the base. On rebased head `c4fe23009`, all 1,384 SDK tests, TypeScript checking, and 158 focused Solana/compatibility/render tests pass. Solana E2E passed on the pre-rebase head; the rebased CI run is pending. No fresh browser acceptance is claimed for the rebase.


Eikix:
## EVM / Solana SDK parity audit and disposition

This records the full 41-item public-surface and decryption comparison against `2512edc2f758c76d00df3ffbb8e91967f9140c2c`, and its disposition in the implemented SDK cleanup. Implementation is in `127e9f6b8`; test-only correction `0175ece90` names the EIP-712 signing domain explicitly to satisfy the retired-ACL glossary guard. Issue #2023 is excluded.

The recommendations consolidate into six changes:
1. Supply one native Kit RPC per client and reuse it throughout SDK-owned reads.
2. Expose only supported capabilities, options and truthful types.
3. Match EVM method names, typed results and shared lifecycle/error conventions.
4. Compose encryption/proof submission and authenticate typed public decryption.
5. Reuse configuration, transport and cryptographic helpers; retain protocol-specific boundaries where necessary.
6. Curate package exports and migrate the demo/CLI to the public consumer paths.

The original audit described the absence of a native Solana client and the lack of locally authenticated public decryption as acceptable defaults. That conclusion was revisited: neither is a Solana protocol restriction. The implementation now binds one native `@solana/kit` RPC to each SDK client and provides authenticated typed public decryption, while keeping raw certificate production available for on-chain consumers.

### Public surface and lifecycle

| Audited area | Status | Result or concrete justification |
|---|---|---|
| Package entrypoints | Retained — design | `@fhevm/sdk/solana` is the consumer surface; `solana/host` exposes the generated host subset for instruction integration. Separate generated codecs avoid making the main entrypoint an instruction catalogue. |
| Chain identity | Retained — protocol | The host chain id remains an exact `bigint`; RFC-021 Solana ids exceed JavaScript's safe integer range. Host/program identities remain bytes32 rather than EVM addresses. |
| Native client | Fixed | All client factories receive one native Kit `rpc`. Client-owned encrypted-store, delegation, permit-invalidation, and public-decryption account reads reuse it. No new provider interface or RPC implementation is introduced. |
| Public client type honesty | Fixed | An explicit Solana surface exposes `chain`, `rpc`, lifecycle and supported actions. Shared core remains private; unsupported EVM `ethereum`, `protocolVersion`, `runtime`, inert `options`, and EVM-shaped `extend` are not public capabilities. |
| `ready` and version lifecycle | Fixed / retained — protocol | `ready` and `init` retain shared core initialization. They no longer promise an EVM protocol context. Encryption initializes the deployment-compatible TFHE artifact; public certificate production needs no decrypt WASM initialization. |
| Runtime configuration | Retained — design | Solana retains its own lazy adapter singleton and shared WASM/auth configuration model. Adapter separation prevents runtime extensions/configuration from leaking between Solana and EVM. |
| Config idempotency semantics | Fixed | Auth equality uses the existing shared comparator. Logger callback equality has one shared implementation used by Solana, viem and ethers; equivalent configuration no longer depends on which adapter received it. |
| Factory-specific option types | Fixed | Encrypt accepts the useful injected encryption key. Public/base/decrypt factories no longer advertise irrelevant encryption or version options. |
| Module-version overrides | Fixed / retained — deployment constraint | Solana rejects unsupported runtime module-version overrides and omits them from its public type. The deployed verifier determines compatible artifacts; accepting an ignored override was misleading. |
| `batchRpcCalls` | Fixed | Removed from the Solana option surface. Kit RPC ownership is explicit; an EVM batching flag is not repurposed as a Solana setting. |
| Injected FHE encryption key | Fixed | The encrypt factory validates that injected key metadata names the configured relayer URL, so a mismatched key cannot silently populate a different cache entry while the client fetches another key. |
| Encryption API granularity | Fixed / retained — protocol | `generateZkProof` names local proof construction, `submitInputProof` remains the advanced submission operation, and `encryptValues` combines them, with `encryptValue` as its scalar counterpart. The convenience result uses shared `EncryptedValue[]` plus a submitted `SolanaInputProof`, rather than presenting the local ZK object as the submitted proof. Solana instruction attestation fields differ from an EVM ABI proof byte string. |
| Encrypt identity types | Fixed | The documented bytes32 contract/user identities are typed as `Bytes32Hex`; runtime builder checks remain. |
| Decrypt factory split | Fixed | The private decrypt factory requires trust and adds private actions to public decryption. Public-only callers use `createFhevmPublicDecryptClient`; the two factories no longer construct identical surfaces through an optional-trust overload. |
| Combined client factory | Retained — design | No combined factory is added. Separate capability factories share the runtime/cache and accept the same native RPC. There is no demonstrated consumer requirement worth adding another factory and composition contract. |
| Public decrypt result | Fixed / retained — protocol | `decryptPublicValue(s)` authenticates KMS signatures using host configuration/context read through the client's RPC and returns shared `TypedValue` results. `publicDecryptCertificate` remains explicitly an untrusted claim for on-chain consumers; those consumers still need host verification and public-history/MMR evidence. |
| User decrypt key/session model | Retained — protocol and design | `signPermit` generates the Solana transport pair and returns the signed permit and pair together in a reusable session. An EVM transport-key generator would produce incompatible material. Session ownership prevents accidental key/permit mismatch. |
| Trust/config defaults | Remaining integration dependency | Private decryption still requires explicit KMS party-id/address mapping, context/epoch routing, FHE parameter and Gateway domain. Public on-chain signer addresses do not establish the private response party-id mapping or epoch source. Canonical sources must be agreed with their owners; the SDK does not invent defaults. This dependency does not prevent public verification from using the available host config and KMS-context accounts. |
| Key cache lifecycle | Retained — concrete development requirement | Scoped relayer-URL cache invalidation remains for same-URL validator/deployment restarts. It reuses the existing cache and avoids exposing its entire implementation. |
| Cleartext clients | Retained — design | No Solana cleartext entrypoint is added. EVM cleartext factories depend on an EVM mock contract/relayer model; an equivalent Solana backend is not part of this change. A factory without that behavior would be another false capability. |
| Main-barrel breadth | Fixed | The primary barrel uses curated consumer exports instead of wildcard exports of retry classifiers, clocks, raw transport/session orchestration and request internals. Protocol-facing instruction/proof helpers remain where consumers need them. |

### Decryption behavior

| Audited concern | Status | Result or concrete justification |
|---|---|---|
| User-decrypt entry points | Fixed / retained — design | `decryptValue` and `decryptValues` return shared `TypedValue` results. The batch entries already support different allowed keys, so a separate `decryptValuesFromPairs` alias is not needed to express heterogeneous authorization. |
| Permit creation and key ownership | Retained — protocol | Solana uses sRFC-38 wallet signing and bytes32 identities; EVM uses its EIP-712 permit forms. The generated transport key remains committed to the signed Solana permit and owned by its session. |
| Permit scopes and delegation | Retained — protocol | Solana permit scopes are `(program, scope)` pairs, and entries may specify an allowed key/delegator. The host authorizes store/program relationships rather than EVM contract/address ACL tuples. |
| Permit validity and revocation inputs | Fixed / retained — protocol | `signPermit` reads the canonical signer invalidation account with the configured RPC/program, then normalizes its start against the watermark. The normal action no longer assumes caller-supplied or zero watermark. A missing account or an empty nonexecutable System-owned prefunded PDA means no previous revocation, matching the host and Connector. Malformed or wrongly owned records fail validation. |
| Permit caching | Retained — design | No hidden permit cache is added. Callers retain and reuse the returned session, as EVM callers retain signed authorization and their transport pair. |
| Request validation | Retained — protocol | Existing checks enforce nonempty requests, count/bit budgets, bytes32 identities, chain binding, order and permit/key relationships before submission. Public typed batching also validates its aggregate handle budget and chain. Different authorization evidence remains host-specific. |
| Submission loop | Retained — deployed service behavior | Private decryption keeps bounded unchanged resubmissions for retryable Connector readiness/indexing failures or empty completed jobs. Each attempt uses the shared `RelayerAsyncRequest`; this is not a second HTTP polling implementation. |
| Timeout budget | Retained — explicit design | Public `options.timeout` follows the EVM HTTP/poll request budget; it does not bound the account reads or later local verification. Private decryption deliberately retains an overall operation deadline across submissions, retry backoff and the final verification check, passing remaining time to each HTTP attempt. The stronger private budget is a design choice, not a protocol requirement. |
| Cancellation and progress | Fixed | Private aborts outside polling now use `RelayerAbortError` and the abort progress policy, including retry backoff/final checks. Public typed decryption passes cancellation into Kit account reads and checks it around those reads, so cancellation is not restricted to HTTP. This does not turn public timeout into an overall operation deadline. |
| Retry limits | Retained — design | Shared request-level fetch/poll limits remain. Private outer resubmission also retains its validated positive finite attempt budget; removing it would allow readiness failures to resubmit without an operation-level bound. |
| Auth and transport options | Retained / fixed | Runtime auth is merged before per-call overrides. Existing headers, fetch retry, signal, timeout and progress options continue through shared transport. Private transport now receives the runtime logger as well. |
| Relayer errors | Fixed / retained — design | Shared typed relayer errors remain. `SolanaUserDecryptRunError` still records outer retry exhaustion and the last rejection; that is distinct from an individual HTTP failure. Abort error shape no longer depends on whether it happens during inner polling or outer backoff. |
| User response authentication | Remaining integration dependency | The Solana WASM response path continues to authenticate/reconstruct against explicit trusted signer party ids and Gateway domain, binding the request and checking threshold/count/order/type. Automatic private trust discovery awaits the canonical party-id and epoch sources described above. |
| Signed request binding | Retained — protocol | Solana reuses the exact signed permit bytes/signature and binds the program, chain, KMS route, handles and transport key in its canonical transcript. EVM's V1/V2 transcript is different but serves the same authentication purpose. |
| Public-decrypt result and trust boundary | Fixed | Typed methods verify against RPC-read host configuration and a live, non-destroyed requested KMS context. They verify distinct registered signer threshold and requested handle/cleartext/context/store/domain binding before returning a typed value. Raw certificate claims remain distinct from authenticated SDK results. |
| Public-decrypt batching | Fixed / retained — current verifier shape | `decryptPublicValues` provides ordered typed results with aggregate budget checks. It uses bounded concurrency over single-handle certificates because the current Solana verifier consumes one handle/store/context certificate per invocation; it does not invent a new batch certificate wire format. |
| Public-decrypt response validation | Fixed / retained — protocol | Existing encoding and requested-extra-data checks remain. Typed decryption additionally verifies KMS signatures locally. Host/on-chain consumption still verifies its own certificate and history proof; client-side verification does not replace transaction validation. |
| Public-decrypt URL and logging | Fixed | The certificate action uses shared auth-aware URL validation/endpoint construction and passes the runtime logger to `RelayerAsyncRequest`. |
| Public-decrypt readiness retry | Retained — deployed endpoint behavior | The Solana adapter keeps readiness-timeout retry for its relayer endpoint. This is a service integration policy and does not alter signature trust or claim verification. |
| Initialization/version selection | Fixed / retained — protocol | Unsupported public version overrides and EVM protocol-version members are removed. Solana retains its compatible versioned WASM path; public signature verification uses crypto without requiring EVM provider initialization. |

### Follow-through on review recommendations

- The initially identified `encryptValues` result mismatch is addressed: `encryptedValues` uses the common type, and `inputProof` now means the submitted Solana attestation. Its structured representation remains different from EVM ABI bytes because Solana instruction consumers need those fields.
- Public-decrypt RPC cancellation and the demo's store-reader bypasses are addressed in the implementation. Demo `getEncryptedStore` and public-proof polling take the configured SDK client's `fetchEncryptedStore` action instead of independently binding RPC plus a hard-coded host id.
- The pure canonical public-decrypt digest has moved alongside `createKmsPublicDecryptEip712` in shared core KMS code. Solana-specific account and signature acceptance rules remain local.
- The review's earlier option of a public Solana extension interface is deliberately not implemented. No inspected consumer needs it; retaining core `extend` would leak EVM contracts, and adding a separate public extension system would expand this cleanup unnecessarily.
- Singular `encryptValue` and `decryptValue` wrappers are implemented. A separate `decryptValuesFromPairs` alias is unnecessary because the existing batch entries already express heterogeneous authorization.
- Canonical automatic private trust/epoch discovery remains an integration dependency, not a hidden follow-up implied by this cleanup. No #2023 work is included.

### Verification

Production implementation `127e9f6b8`:
- Full SDK production build passed, including code generation consistency, formatting, lint, types, ESM/CJS and WASM packaging.
- SDK: 1,378 unit tests across 74 files passed.
- Demo: 250 tests across 42 files and type checking passed. CLI: 475 tests across 36 files and type checking passed.
- Packed SDK installed into a separate consumer; strict TypeScript consumer and ESM/CJS imports checked. Main runtime exports reduced from 95 to 45.
- Burner-wallet browser acceptance: public deposit, encrypted join, dispatch, settlement, claim, private redeposit, partial and full redemption, hide/reveal, invalid inputs, refresh, copy controls and disconnect. Final observed state: zero shares and 1,177 confidential USDC after the yield scenario.
- Fresh live demo smoke: 1 test / 17 assertions passed, including typed public decryption against real HostConfig/KmsContext and exact private payout recovery.
- Seven further live scenarios passed: token burn/public decrypt/redeem/disclose; current and historical FHE decryption; two-holder transfer; 32 dependent FHE steps with concurrent unrelated work; wallet delegation/revocation; 2-of-3 Squads delegation/revocation.
- Service evidence: all 10 Prometheus scrape targets up; inspected KMS/Copro/Relayer logs had no ERROR entries in the sampled window. A Jaeger request trace contained 15 spans across three services covering validation, submission, polling and response forwarding.
- Independent correctness and maintainability reviews were followed by fixes for encryption result shape, final public-decryption account snapshot, RPC cancellation, demo RPC bypasses, and prefunded permit-invalidation PDAs.

The final test-only naming correction passed 26 public-decryption tests, formatting and SDK type checking. CI status is reported separately below; local acceptance is not a claim of green CI. Browser acceptance used the burner wallet, not external wallet providers. Service logs and traces are sampled evidence, not a proof that all possible failures are absent.

### Decisions retained for maintainer review

Private trust discovery still needs a canonical party-id mapping and epoch source; explicit trusted configuration is required today. Public timeout keeps the EVM HTTP/poll budget while private decryption retains an overall deadline. Solana permits, MMR evidence and structured instruction attestations remain protocol-specific. These differences are explicit above, rather than being hidden behind a generic cross-chain provider.

Latest head: [`0175ece90`](https://github.com/zama-ai/fhevm/commit/0175ece90af318fcde49a16b7d326f1f361ecf49). The unchanged dead-surface guard passes locally. CI has restarted and is pending; a native follow-up is active to inspect failures and finish current-head verification.

PR #3937, opened 2026-09-14
fix(listener): harden database password, chain_id configuration, block computer error handling

maxnovawind:
## What

Three small robustness fixes to `listener_core` configuration handling.

**Database passwords with special characters now work.** The listener only
accepted credentials embedded in `APP_DATABASE__DB_URL`. sqlx percent-*decodes*
the password it parses out of a connection string but never encodes one, so the
caller had to pre-encode it — and getting that wrong failed in two ways: `/`, `?`
and `#` made the URL unparseable (surfacing as a cryptic "invalid port number" at
connect time), while any literal `%XX` sequence was silently decoded into a
*different* password. An optional `APP_DATABASE__DB_PASSWORD` now overrides the
URL password and is passed to the driver byte-for-byte, skipping URL parsing.

**Out-of-range `chain_id` is rejected at config load.** `chain_id` was only
checked for being non-zero. It doubles as a PostgreSQL advisory lock key, and the
cleaner, finality and final-cleaner flows derive their keys by XOR-ing it with
salts that all carry bit 62. Bounding it to 2^60 guarantees a salted key can
never collide with another chain's raw fetch/reorg key on a shared database, and
never goes negative as an `i64`.

**The retention floor is a named constant.** The `999` minimum for
`cleaner.blocks_to_keep` was duplicated across the check, its error message and
two tests. It is now `MIN_BLOCKS_TO_KEEP`, so those cannot drift apart.

## Compatibility

No behaviour change for existing deployments. `db_password` defaults to `None`,
leaving the `db_url` path byte-identical; every config in the repo uses a
`chain_id` far below the new bound; and the retention floor keeps the same value.

## Test plan

- `cargo test -p listener_core --lib` — 111 passing (was 109).
- New coverage: password override vs. a correctly pre-encoded URL (round-trip
  equality), raw special characters still failing at parse, `db_url` untouched
  when no override is set, and the `chain_id` boundary at 2^60 / 2^60 + 1.
- `cargo clippy --all-targets` and `cargo fmt --check` clean, with and without
  the `iam-auth` feature.

The IAM auth path is unaffected — it parses `db_url` independently and replaces
the password with a short-lived token.


tawadaa (dismissed):

nboisde (approved):
LGTM !

rudy-6-4 (approved):

maxnovawind:
@mergify queue

PR #3939, opened 2026-09-14
ci(charts): deploy the Solana host and listener into a preview environment

Eikix:
## What changes

Second half of the Solana preview environment, after #3841 (the deployer image and the Solana-side changes) and #3996 (`host wipe`, which resets the preview's program-owned accounts on Solana devnet), both merged. This PR only touches the preview-env workflow and scripts, the coprocessor chart, and their checks. The deployer is consumed as an image with a documented environment contract; nothing here needs Solana context beyond what is written below.

### Coprocessor chart

`solanaHostListener` (default off) adds one `solana_host_listener` Deployment plus a ClusterIP Service per coprocessor release, mirroring the one-Deployment-per-EVM-chain pattern.

- One replica with `Recreate`: the listener's durable checkpoint is a singleton in that coprocessor's database, and two writers would race on it.
- Readiness probes `/healthz`, which covers database and HTTP health only. A stream that stops delivering blocks is not visible to Kubernetes; the listener reconnects itself after 30 seconds of silence, and checkpoint progress is the operator's health signal. The preview script gates on it.
- Configuration comes through `env` (RPC and gRPC URLs, gRPC token, proof API key, program id), all expected from Secrets and the host addresses ConfigMap.

### Preview workflow

Two new `preview-env-deploy.yml` inputs:

- `deploy_solana`: turns the Solana steps on.
- `deploy_example_programs`: also deploy confidential-token, demo-vault and confidential-batcher.

When on, `ci/preview-env/solana-host/deploy-preview.sh` runs after the EVM bring-up:

1. Install the `sync-secrets` chart with `values-solana-rpc.yaml` and `values-solana-deployer.yaml`, the way `chain_mode=testnets` gets its RPC URLs, so the External Secrets Operator writes `solana-rpc` and `solana-deployer` into the namespace from AWS Secrets Manager. Generate `solana-proof-api`, the leaf-proof bearer token, for this preview; only the listeners and connectors read it.
2. Wait until each coprocessor database holds the canonical (chain 12345) keys.
3. Install the Solana host through the `contracts` chart (`host deploy` Job) with the KMS and coprocessor thresholds derived from the party counts.
4. Register the Solana host chain on the Gateway (`addHostChains` overlay).
5. Per coprocessor: register the database (`coprocessor register` Job), upgrade the release in place with the listener overlay, restart the zkproof worker (it loads `host_chains` at startup), wait for the listener rollout.
6. Upgrade each KMS connector in place with the proof endpoints, and patch the relayer's second host chain.
7. Gate on listener checkpoint progress, then optionally deploy the example programs.

Topology validation rejects Solana together with Polygon, blockchain-dev and blue-green, and requires `build_images=true` or an explicit `overrides.solana_programs_version`, because no workflow publishes the `solana-programs` image under a base-commit tag. Launches share one concurrency group because they share the preview's Solana devnet program addresses.

## What to check

- The listener template against the existing per-chain host-listener template: same env handling for database auth and TLS, same pod defaults.
- `deploy-preview.sh` reads and rewrites existing release values with `helm get values` so party-specific settings survive the in-place upgrades. Anything it should carry forward and does not is a bug.
- `test_charts.py` renders the chart and each overlay in `charts-helm-checks`.

## Limits

- **Shared Solana devnet identity.** The preview deploys onto the public Solana devnet cluster with its own preview-only program keypairs. Zama devnet will deploy its own Solana host onto the same cluster with separate keypairs, so the two never share addresses, and the concurrency group serializes preview launches only. Program addresses, `HostConfig` and the KMS context live on Solana devnet, not in the namespace. Destroying a preview does not reset them, and a fresh preview generates a new Gateway and committee that cannot be silently rebound. `solana/deploy/README.md` has the operator steps for compatible and breaking changes.
- **One experiment at a time** per set of program addresses. CI serializes launches; manual deployments are the operator's responsibility.
- **AWS entries not yet created.** `zws-dev/external-solana-rpcs` (Helius devnet RPC, gRPC URL and token) and `zws-dev/preview-solana-host-keypairs` (deployer, zama_host, confidential_token keypair JSONs). Both exist in the 1Password vault "Preview Env"; the values files name the AWS properties, and can follow other names if infra prefers.
- **KMS image.** The preview defaults to KMS `v0.14.1`, which has no Solana user-decryption path. A Solana dispatch needs `kms_core_version` and `kms_repo_ref` overrides pointing at the kms `solana` branch.
- **Chain id.** The overlays and script write `2^63 | 12345`, the id the code mints today. DD-052 assigns Solana devnet `0x01ce59db5080fc2c`; when the fhevm port of kms#856 lands, these four sites change with it.
- **Not yet run live.** Chart checks, shellcheck and actionlint pass. The first real dispatch with `deploy_solana=true`, followed by one compute-and-decrypt smoke test, is the remaining acceptance step.


Eikix:
Review of ce28de088 (2 commits, base `solana/preview-host-deploy` at 7257f1c4c). Chart checks, shellcheck and the rest of CI are green. Nothing here has run live, so the review is a static pass over the wiring plus a render of the chart paths the preview actually uses.

### Verified against the tree

- **Workflow wiring.** `CONTRACTS_CHART`, `COPROCESSOR_CHART` and `KMS_CONNECTOR_CHART` come from `resolve-charts.sh` via `GITHUB_ENV`, and `NAMESPACE`, `TAGS_JSON`, `NB_COPROCESSOR`, `NB_KMS_CORE`, `COMMON_CHART_VERSION` are job or workflow env, so `set -u` in `deploy-preview.sh` is satisfied. `build-solana-programs` is in `resolve-tags`' `needs` and that job already runs under `!cancelled()`, so a skipped build does not skip resolution. The `solana_programs` image entry matches the reusable workflow's `build_result` output and `fhevm/solana-programs` name. The concurrency expression falls back to the PR number on `pull_request`, where `inputs.*` is empty.
- **Listener template.** The env block is the same sequence as `coprocessor-host-listener-deployment.yaml` (user, endpoint, password under the `password` auth mode, `databaseUrl`, IAM flag and region, SSL root cert env and mount). `--database-url=$(DATABASE_URL)` resolves in the preview because `values-coprocessor-e2e.yaml` defines `databaseUrl` from `$(DATABASE_USER)`, `$(DATABASE_PASSWORD)` and `$(DATABASE_ENDPOINT)`, which render before it. `coprocessor-<i>` is the IRSA ServiceAccount every other worker in the release uses. The five flags the container passes exist on `solana_host_listener.rs`, both env-read secrets match, and `/healthz` is routed in `http_server.rs`.
- **Address and config plumbing.** `host deploy` writes `ZAMA_HOST_ADDRESS=` to `.env.solana`; the contracts chart strips `_ADDRESS` and lowercases, so `solana-host-addresses/zama_host.address` is the right key for the listener, connector and relayer. The connector's `$(SOLANA_RPC_URL)`, `$(SOLANA_HOST_PROGRAM_ID)` and `$(SOLANA_PROOF_API_KEY)` placeholders expand because `kmsConnectorKmsWorker.env` renders at line 57 of the kms-worker template and `KMS_CONNECTOR_HOST_CHAINS` at line 105; `test_charts.py` pins that order. The relayer accepts a base58 `ACL_ADDRESS` for a Solana chain id (`settings.rs`), and index 1 is free because Polygon is rejected in the topology check.
- **Restart scope.** `HostChainsCache::load` is only used by `zkproof-worker/src/verifier.rs`, so restarting just the zkproof worker after registration is correct. `registerSolanaCoprocessorSql` is idempotent and refuses a chain already registered under another program.
- **In-place upgrades.** Bootstrap sets everything through `-f`, `--set`, `--set-string` and `--set-json`, all of which `helm get values` returns, and no script patches the coprocessor, connector or relayer releases with `kubectl`, so nothing is reverted by the re-upgrade. Thresholds match: `kms_t` is the corruption threshold `host deploy` expects in `KMS_THRESHOLD`, `coproc_threshold` the majority it expects in `COPROCESSOR_THRESHOLD`.

### Worth changing before the first live dispatch

1. **The deployer keypair lands in a namespace where dev groups have `admin`.** `deploy-preview.sh` copies `solana-deployer` (devnet upgrade authority plus `zama_host.json`) into the preview namespace, and the "Grant dev-access groups RBAC" step binds `coprocessor-dev-access` and `kms-dev-access` to the `admin` ClusterRole there. Anyone in those groups can read the keypair that owns the shared program addresses every Solana preview depends on. It is devnet, but the PR's own Limits section says those addresses cannot be silently rebound, so a leaked or misused key costs every user of the preview. Either run the two deploy Jobs in the secrets namespace and only copy `solana-rpc` and `solana-proof-api`, or say explicitly in Limits and the README that the deployer key is readable by all preview-namespace admins and is treated as disposable.
2. **`psql` through `kubectl exec` without `PGPASSWORD`.** `assert-gcs-dry-run.sh` wraps the same call as `psql_party` with `env PGPASSWORD=zama`. The new loops rely on local-socket trust auth in the Postgres image. Reusing the helper (or lifting it into `lib.sh`) removes that assumption and the two copy-pasted polling loops.
3. **Topology guard misses `build_test_suite_only=true`.** With Solana on and that flag set, `build-solana-programs` is skipped and no override is required, so the run fails later in `resolve-tags` with "no published image" instead of at validation. One more clause in the guard, or reuse the `should_build_images` output the build jobs already gate on.
4. **The listener test renders only the `databaseUrl` literal path.** `test_listener_shares_database_and_keeps_proofs_private` passes `commonConfig.databaseUrl=postgresql://db/coprocessor`, so the user/endpoint/password expansion the preview uses is never rendered. Rendering once with `ci/preview-env/coprocessor/values-coprocessor-e2e.yaml` and asserting `DATABASE_USER`, `DATABASE_ENDPOINT` and `DATABASE_PASSWORD` precede `DATABASE_URL` would cover the real configuration.

### Nits

- `pip install pyyaml` is unpinned while every action in that job is pinned by SHA.
- The wait loops give keys 10 minutes and checkpoint progress 5 minutes with no log line per attempt; a short progress echo would make a stuck run diagnosable from the job log.
- The chart default image `ghcr.io/...` differs from the overlay's `hub.zama.org/ghcr/...`; consistent with the other overlays, just noting the override is load-bearing.

The acceptance step in the PR body (one dispatch with `solana_secrets_namespace`, then a compute-and-decrypt run) is the right remaining gate; items 2 and 4 are the ones most likely to surface there.

fegmorte (approved):
LGTM ! Thanks a lot

PR #3940, opened 2026-09-14
chore(kms-connector): proxy healthcheck

eudelins-zama:
Healthcheck implementation for kms-connector's proxy


kms-connector/crates/proxy/src/bin/proxy.rs
```diff
@@ -39,7 +46,17 @@ fn run() -> anyhow::Result<()> {
...
 
-            Proxy::from_config(config)?.run()?;
+            // The monitoring server runs on the tokio runtime, next to Pingora's own.
+            let monitoring_endpoint = config.monitoring_endpoint;
+            let (proxy, state) = Proxy::from_config(config)?;
+            let cancel_token = CancellationToken::new();
+            let monitoring_server_task =
+                start_monitoring_server(monitoring_endpoint, state, cancel_token.clone());
+
+            proxy.run()?;
```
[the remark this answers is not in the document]
eudelins-zama: This is a shared pattern among all connector's services.
Makes sense to abort on port conflict or such, but worth a dedicated PR to update other services IMO.
Non-blocking as a k8s would shut the pod in case it is not able to query the monitoring server of the service anyway.

kms-connector/crates/proxy/src/bin/proxy.rs
```diff
@@ -22,8 +28,9 @@ fn run() -> anyhow::Result<()> {
         Subcommands::Validate { config } => {
             Config::from_env_and_file(Some(config))?;
         }
-        Subcommands::Health { endpoint: _ } => {
-            todo!("Proxy healthcheck is not implemented yet")
+        Subcommands::Health { endpoint } => {
+            let runtime = tokio::runtime::Runtime::new()?;
+            runtime.block_on(query_healthcheck_endpoint::<HealthStatus>(endpoint))?;
```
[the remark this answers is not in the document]
eudelins-zama: Same as below, looks true, but is a shared pattern worth fixing in a follow-up.
Currently, only "real" use case for this `<connector_service> health` commands are Dockerfiles `HEALTHCHECK` setup.
We always use the `--endpoint` there, so should be non-blocking.

Ex: [Dockerfile](https://github.com/zama-ai/fhevm/blob/main/kms-connector/crates/gw-listener/Dockerfile#L76-L77)

kms-connector/crates/proxy/src/bin/proxy.rs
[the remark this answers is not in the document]
dvdplm: Fair. File an issue?

kms-connector/crates/proxy/src/core/proxy.rs
```diff
@@ -10,22 +13,24 @@ use pingora::{
...
 
 /// How long in-flight connections are given to finish once the graceful shutdown deadline (
 /// [`Config::shutdown_grace_period`]) has passed, before Pingora forcibly closes them.
 const GRACEFUL_SHUTDOWN_TIMEOUT_SECONDS: u64 = 5;
 
 impl Proxy {
-    pub fn from_config(config: Config) -> anyhow::Result<Self> {
-        let endpoint_balancer =
+    /// Creates the `Proxy` from its configuration, and the `State` used to monitor it.
+    pub fn from_config(config: Config) -> anyhow::Result<(Self, State)> {
```
dvdplm: nit: given how inflated the word "state" is, perhaps reading `health::State` is clearer? I.e. change the use clause above to `use crate::monitoring::health::{self, endpoint_health_check}` or similar?

kms-connector/crates/proxy/src/core/proxy.rs
```diff
@@ -37,12 +42,20 @@ impl Proxy {
...
+            Arc::clone(&endpoint_balancer),
+            config.healthcheck_timeout,
+        )?;
+        let proxy = Self {
             verifier: ApiKeyVerifier::new(config.api_key_digest),
             config,
             endpoint_balancer,
-        })
+        };
+        Ok((proxy, state))
```
dvdplm, line 61: Interesting that these are distinct. Can you discuss why the `State` is not a part of the `Proxy`? Not wrong keeping them distinct mind you.

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    /// Performs the healthcheck of the proxy's TLS listener by querying the `v1/version` route
+    /// through the loopback interface.
+    ///
+    /// No credentials are sent, so the proxy is expected to answer `401`: what is checked here is
+    /// that the listener accepts TLS connections and answers HTTP requests, not the routing.
+    async fn tls_listener_healthcheck(&self) -> Result<(), String> {
+        let url = &self.proxy_version_url;
+        let request = self.loopback_client.get(url.clone()).send();
+        match tokio::time::timeout(self.healthcheck_timeout, request).await {
```
maksymsur, line 65: Why do we have two identical timeouts? See client `.timeout` at line 34 above and this outer tokio timeout. reqwest arms its deadline inside `send()` and tokio polls the inner future first, so the "timed out" arm at line 75 may be unreachable and a hung listener reads exactly like a refused one (`error sending request for url (...)`). Maybe it's worth dropping the client-level timeout as `endpoint/src/monitoring/health.rs:101` does, or add an `e.is_timeout()` arm?

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+use std::{
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    sync::Arc,
+    time::Duration,
+};
+
+/// The struct used to monitor the state of the `Proxy`.
+#[derive(Clone)]
+pub struct State {
+    proxy_version_url: Url,
```
dvdplm, line 18: What does the version have to do with the endpoint health? I would have thought that one endpoint service would server all versions at the same time? Maybe that's the case and this is just a naming thing?

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+use kms_connector_api::VERSION_ROUTE;
+use pingora::lb::{Backend, LoadBalancer, health_check::HttpHealthCheck, selection::RoundRobin};
+use serde::{Deserialize, Serialize};
+use std::{
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    sync::Arc,
+    time::Duration,
+};
+
+/// The struct used to monitor the state of the `Proxy`.
```
dvdplm:
```suggestion
/// The struct used to monitor the health of the `Proxy`.
```

yeah?

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+#[derive(Clone)]
+pub struct State {
+    proxy_version_url: Url,
+    endpoint_balancer: Arc<LoadBalancer<RoundRobin>>,
+    healthcheck_timeout: Duration,
+    /// The client used to reach the proxy's own TLS listener through the loopback interface.
+    loopback_client: reqwest::Client,
```
dvdplm, line 22: Do we need TLS on the loopback interface? EDIT: oh it's for the self-check, got it.

kms-connector/crates/proxy/src/core/proxy.rs
```diff
@@ -37,12 +42,20 @@ impl Proxy {
                 .map(|b| b.addr.to_string())
                 .collect::<Vec<_>>()
         );
+        endpoint_balancer.set_health_check(Box::new(endpoint_health_check(&config)?));
```
maksymsur, line 46: maybe I miss smth... but the health check is attached but never scheduled (`health_check_frequency` is `None`, balancer not registered as a background service), so it only runs "classically" when someone polls `/healthz`. Eviction then depends on the prober's cadence, and one unreachable endpoint can make `/healthz` to block for ~3–6s or whatever time it takes depending on the runtime env. 

Isn't Pingora's intended shape - [quick start — Peer health checks](https://github.com/cloudflare/pingora/blob/0.9.0/docs/quick_start.md#peer-health-checks), [load_balancer.rs](https://github.com/cloudflare/pingora/blob/0.9.0/pingora-proxy/examples/load_balancer.rs#L74-L93)? 

Shall we set [health_check_frequency](https://docs.rs/pingora-load-balancing/0.9.0/pingora_load_balancing/struct.LoadBalancer.html#structfield.health_check_frequency) + `parallel_health_check`, wrap with [background_service](https://docs.rs/pingora-core/0.9.0/pingora_core/services/background/fn.background_service.html) and `add_service` it; `/healthz` then just reads `backends.ready(b)` ?

maksymsur (commented):
LGTM, but look into comments. If you think they may be reasonably dismissed - the PR may be approved

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+        // The proxy's certificate is issued for its public hostname, not for the loopback
+        // address used here, so its verification is disabled for this client only.
+        let loopback_client = reqwest::Client::builder()
+            .timeout(healthcheck_timeout)
+            .tls_danger_accept_invalid_certs(true)
+            .build()
+            .context("Failed to build the healthcheck HTTP client")?;
+
+        Ok(Self {
+            proxy_version_url: version_url(bind_address),
```
dvdplm, line 39: hmm, this `version` naming is confusing me. I think we can just remove it and just have `proxy_url`? Or will this become a list of URLs  once we have more than one version?

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+    ///
+    /// Splits them into the healthy and unhealthy ones.
+    async fn endpoints_healthcheck(&self) -> (Vec<String>, Vec<String>) {
+        let backends = self.endpoint_balancer.backends();
+        backends.run_health_check(true).await;
+        let all = backends.get_backend();
+        let (healthy, unhealthy) = all.iter().partition(|b| backends.ready(b));
+        let addresses =
+            |backends: Vec<&Backend>| backends.into_iter().map(|b| b.addr.to_string()).collect();
+        (addresses(healthy), addresses(unhealthy))
```
dvdplm, line 54: Do we care about the healthy ones? Like, why not just return the list of unhealthy endpoints (and if it's empty => all is good)?

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+            errors.push(format!(
+                "Unhealthy endpoints: {}",
+                unhealthy_endpoints.join(", ")
+            ));
+        }
+        if healthy_endpoints.is_empty() {
+            errors.push("No healthy endpoint available".to_string());
+        }
+
+        let healthy = tls_listener_reachable && !healthy_endpoints.is_empty();
```
dvdplm, line 94: Isn't this the  same as `tls_listener_reachable && unhealthy_endpoints.is_empty()`? That would avoid the double-negative my brain struggles with...

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+        let status_code = if healthy {
+            StatusCode::OK
+        } else {
+            StatusCode::SERVICE_UNAVAILABLE
```
dvdplm, line 98: mmmh, what if `tls_listener_reachable` is true and there's at least one healthy endpoint? Would that mean that the service is *degraded* but not wholly unavailable?

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub tls_listener_reachable: bool,
+    /// The endpoints currently receiving requests.
+    pub healthy_endpoints: Vec<String>,
+    /// The endpoints that failed their healthcheck. Non-empty is degraded, but not unhealthy.
+    pub unhealthy_endpoints: Vec<String>,
+    /// Details about any issues encountered during healthcheck.
+    pub details: String,
+}
+
+/// The healthcheck run against each endpoint: a `GET v1/version` must be answered with `200`.
```
dvdplm, line 133: It could be a `HEAD` instead of `GET` so we can respond with just the headers (faster)?

Also: the route reads weird to me: `/v1/version` –– it's circular! If I request `v1` I already know the version! Can't we expose a root `/version` instead? (responds to `GET` with the latest version supported? responds to `HEAD` with just a 200 OK?)

kms-connector/crates/proxy/src/core/proxy.rs
eudelins-zama: Indeed, this might not be super intuitive.

But basically, I did that on purpose.
In prod, in our k8s setup, we are querying the `healthz` endpoint of the service periodically (restarting the service if not healthy for some time).
IMO, each call to this endpoint should trigger the underlying checks the service make (DB healthcheck, or endpoints healthcheck in the case of the proxy).
If we go for the default Pingora health check, it would have its own healthcheck period, independant from the k8s one. Meaning if we want to update the period of the healthcheck, we now to update both the k8s settings and pingora settings.

Makes sense?

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: Easier to know how many remaining healthy ones we have in case of degraded endpoints I'd say?

kms-connector/crates/proxy/src/core/proxy.rs
eudelins-zama: Because `Proxy` is moved into Pingora by `run(self)` and lives on Pingora's runtimes, while `State` lives in the monitoring server task (tokio runtime)

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: Nope, they're not equivalent. With one healthy + one unhealthy endpoint, current logic returns healthy, while `unhealthy_endpoints.is_empty()`  would return unhealthy

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: This means the service is healthy.

For now AFAIK, health is binary, either healthy or not

kms-connector/crates/proxy/src/core/proxy.rs
maksymsur: Yes.. but only if we look at this issue from 1 angle. If we look widely - the difference from the DB check in other services is that this pass doesn't just report - it decides which endpoints get traffic. So with k8s running `liveness` + `readiness` against `/healthz`, two passes run concurrently every period and can overwrite each other's result, and when one endpoint is down `/healthz` waits its 3 s connect timeout before answering "healthy" - longer than k8s's default `1s` probe timeout, so a proxy that's serving fine via the other endpoint gets restarted. If our k8s is configured to account to this risk - than Ok.

But I'd let Pingora's built-in timer own the probes and have `/healthz` return the stored result instantly; the k8s period then only answers "is the proxy alive". Non-blocking if you prefer to keep it - but then let's cap the pass with a timeout and document the required `timeoutSeconds`.

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: `HEAD` is not a method we would accept on that endpoint.

> Also: the route reads weird to me: /v1/version –– it's circular! If I request v1 I already know the version! Can't we expose a root /version instead? (responds to GET with the latest version supported? responds to HEAD with just a 200 OK?)

Indeed this looks weird. Not in the scope of this PR IMO, Idk how this endpoint would be used by Relayer in practice

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: To check the health of the `Proxy` listener, we query its `v1/version` endpoint of the proxy. That's what lives in `proxy_version_url`.

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: This is just the url to the version endpoint of the proxy. Not sure how this could be more descriptive? :thinking: 

Using `proxy_url` would move the problem to the `tls_listener_healthcheck` I'd say...

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: Indeed! Fixed 0bc744f44e0a214b3877959e9c0ae2453a915298

kms-connector/crates/proxy/src/core/proxy.rs
eudelins-zama: Fixed 1b3d898836f36712f28a34db6d4334f8da0342b6

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: Fixed 1b3d898836f36712f28a34db6d4334f8da0342b6

kms-connector/crates/proxy/src/monitoring/health.rs
```diff
@@ -0,0 +1,182 @@
...
+pub fn endpoint_health_check(config: &Config) -> anyhow::Result<HttpHealthCheck> {
+    // The endpoints answer whatever the `Host` header says, so it is left empty
+    let mut health_check = HttpHealthCheck::new("", false);
+    let version_uri = VERSION_ROUTE
+        .parse()
+        .map_err(|e| anyhow!("Failed to build the endpoint healthcheck request: {e}"))?;
+    health_check.req.set_uri(version_uri);
+    health_check.peer_template.options.connection_timeout = Some(config.endpoint_connect_timeout);
+    health_check.peer_template.options.read_timeout = Some(config.healthcheck_timeout);
```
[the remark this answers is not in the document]
eudelins-zama: Fixed 2e48d4e2f2a0715548424a4a88e12b5420556bc1

kms-connector/crates/proxy/src/bin/proxy.rs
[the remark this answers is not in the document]
eudelins-zama: Done https://github.com/zama-ai/fhevm-internal/issues/2044

kms-connector/crates/proxy/src/core/proxy.rs
eudelins-zama: > the difference from the DB check in other services is that this pass doesn't just report - it decides which endpoints get traffic

This is true indeed. I'll change this and go for the pingora approach then

eudelins-zama: Fixed 43fa56c59a50621fd081514d536c16cea4e1775e

maksymsur: Cool, thnx!

maksymsur (approved):
thanks for addressing comments! approving

kms-connector/crates/proxy/src/core/config.rs
```diff
@@ -30,6 +33,12 @@ pub struct Config {
     /// The timeout to establish a TCP connection to an endpoint.
     #[serde(with = "humantime_serde", default = "default_endpoint_connect_timeout")]
     pub endpoint_connect_timeout: Duration,
+    /// How often the endpoints are probed in the background to filter which ones receive requests.
+    #[serde(
+        with = "humantime_serde",
+        default = "default_endpoint_healthcheck_frequency"
+    )]
+    pub endpoint_healthcheck_frequency: Duration,
```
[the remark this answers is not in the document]
eudelins-zama: Indeed, might be good to handle.

Added it in the follow-up [issue](https://github.com/zama-ai/fhevm-internal/issues/2044), as it could be used in many places across connector's services, not just the proxy.

kms-connector/crates/proxy/src/monitoring/health.rs
dvdplm: You know the number of endpoints you have, so you only need to know how many are unhealthy. Or how many are healthy. You don't need both, that's my point.

kms-connector/crates/proxy/src/monitoring/health.rs
dvdplm: ok, I see.

kms-connector/crates/proxy/src/monitoring/health.rs
dvdplm: Ok, good to know and a fair choice imo. Worth noting somewhere? (comment, doc)

kms-connector/crates/proxy/src/monitoring/health.rs
dvdplm: hmm, I think this code – brand new and without any legacy – should make the Right Choices™ and then let the Relayer or whoever adapt. I see two disjoint problems here:

1. Using the version endpoint to mean "I'm healthy"
2. the odd nesting of `v1` + `version` in the same breath

I'm fine with the first "quirk", but the second is worth fixing imo.

eudelins-zama: 1. What would you suggest instead? Seems like a reasonable way to me to check the health of the proxy's TLS listener
2. This is out of the scope of the PR, because the `v1/version` is part of the shared `kms-connector-api`, and used in many places, it's not healthcheck related, so I don't think we should fix that in this PR

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: If we're introducing auto-scaling etc., might not be that trivial to infere from unhealthy ones...
It does not hurt to include the healthy ones right?

kms-connector/crates/proxy/src/monitoring/health.rs
eudelins-zama: It is documented on Notion I'd say

kms-connector/crates/proxy/src/monitoring/health.rs
dvdplm: An auto-scaling solution will still need to inform the proxy about added/removed endpoints though right? I mean, the proxy must always have this information I think?

eudelins-zama: Yes but not the client querying this endpoint, thus why I want to keep it in the response

kms-connector/crates/proxy/src/monitoring/health.rs
dvdplm: > What would you suggest instead?

Just `/healtcheck` or `/heartbeat` or `/alive` or something that conveys the meaning

> This is out of the scope of the PR, because the v1/version is part of the shared kms-connector-api

But the endpoints are brand new code, we can add whatever urls we want I thought?

dvdplm: Clarified in a huddle. Resolving.

kms-connector/crates/proxy/src/monitoring/health.rs
dvdplm: Clarified in a huddle. Resolving.

dvdplm (approved):

eudelins-zama:
@mergifyio queue

PR #3941, opened 2026-09-14
ci(test-suite): poll GHCR before orchestrated e2e

fegmorte:
## Summary

Orchestrated e2e was failing with `manifest unknown` for two different reasons. Docker-build jobs can report success before the short-SHA tag is pullable (registry delay). More often, skipped components used a **baseline tag that is unavailable** after a failed publish/retag on the base SHA.

Preview already checks GHCR in `resolve-tags`. `create-e2e-tests-input` now validates **every lock-selected image**, including baselines, before stack startup.

## Changes

* Extract the GHCR read client to `ci/preview-env/scripts/ghcr-registry.cjs` (preview resolver unchanged in behavior)
* `resolve-baseline` uses `GHCR_READ_TOKEN` and `REQUIRE_PUBLISHED_IMAGE_CHECK=true`: missing package-read fails instead of an unverified lock; empty required-package tag sets fail instead of unverified pins; ancestor fallback still applies when the requested sha is unpublished
* After recording successful builds, poll each built `repo:shortSha` until the manifest exists or 120s elapses (404 is "not yet")
* For skipped builds, assert each lock-selected baseline tag exists; a 404 fails immediately (`baseline tag is unavailable`)
* Do not retry compose on `manifest unknown` / registry `not found` (those are missing images, not delay)
* `create-e2e-tests-input` requires `resolve-baseline` success so a credential failure is not reported as a missing lock artifact
* Resolve diagnostics are kept in the job log (no longer swallowed by `tail -n 1`)

Closes zama-ai/fhevm-internal#2028


ci/preview-env/scripts/resolve-tags.cjs
```diff
@@ -17,8 +17,7 @@
 
 // CI pushes to ghcr.io/zama-ai/*; the cluster pulls the same artifacts through
 // the hub.zama.org/ghcr proxy-cache, so GHCR is where existence is checked.
-const GHCR_HOST = 'ghcr.io';
-const GHCR_OWNER = 'zama-ai';
+const { registryClient } = require('./ghcr-registry.cjs');
```
fegmorte, line 20: [AUTOMATED] This `require('./ghcr-registry.cjs')` will fail in `preview-env-deploy.yml`'s `resolve-tags` job. That job sparse-checkouts only `ci/preview-env/scripts/resolve-tags.cjs` (`sparse-checkout-cone-mode: false`), so Node cannot load the extracted sibling module and every preview deploy that uses this job will error with `MODULE_NOT_FOUND`.

Add `ci/preview-env/scripts/ghcr-registry.cjs` to that sparse-checkout list. Orchestrated e2e is unaffected because it does a full checkout.

Confidence: 97/100

fegmorte (commented):
[AUTOMATED]

## Code Review

Checked for bugs, guideline compliance, error handling, and protocol compatibility.

No compatibility-sensitive changes.

One blocking issue: `resolve-tags.cjs` now requires `./ghcr-registry.cjs`, but `preview-env-deploy.yml` still sparse-checkouts only `resolve-tags.cjs`. Preview deploys that run the `resolve-tags` job will fail with `MODULE_NOT_FOUND`.

GitHub does not allow requesting changes on your own PR; treat this as blocking until the sparse-checkout is updated.

ci/preview-env/scripts/resolve-tags.cjs
fegmorte: [AUTOMATED] Fixed in 6544e0688: `preview-env-deploy.yml` now sparse-checkouts both `resolve-tags.cjs` and `ghcr-registry.cjs`.

Eikix:
**Human-directed AI review**

The cited failures point to missing **baseline images after failed publishing jobs**, rather than delayed visibility of successfully built head images:

- [Gateway publishing prerequisite failed](https://github.com/zama-ai/fhevm/actions/runs/33656469122/job/100336151735): `git ls-remote` could not authenticate, so retagging was skipped.
- [Coprocessor image build failed](https://github.com/zama-ai/fhevm/actions/runs/34216270708/job/102028929602): Chainguard returned `Rate exceeded`. Its baseline tag was still missing when [E2E failed](https://github.com/zama-ai/fhevm/actions/runs/34224954172/job/102058622208).

`waitBuiltGhcrTags` only checks successful builds at the head SHA, so it misses these cases.

Please validate **all images selected by the resolved lock**, including baseline tags. Give baseline resolution package-read credentials, retain its diagnostics, and use the existing permitted ancestor fallback or fail early with the missing `image:tag`. Avoid treating every `manifest unknown` as transient.

A regression test should cover **successful head builds + a missing baseline image**, asserting a verified fallback or failure before stack startup.

fegmorte:
Addressed in 6da839091:

- `resolve-baseline` now uses `GHCR_READ_TOKEN` plus `REQUIRE_PUBLISHED_IMAGE_CHECK=true`, so a missing package-read scope fails instead of emitting an unverified lock. Existing ancestor fallback still applies when the requested sha is unpublished.
- `create-e2e-tests-input` verifies **lock-selected** tags for skipped (baseline) images and still polls only successfully built head tags. A baseline 404 fails immediately with `image:tag`.
- Compose no longer retries `manifest unknown` / registry `not found`; those are treated as missing images, not transient delay.
- Regression: successful head overlay + missing baseline asserts a verified ancestor fallback or a failure before stack start.

Eikix (commented):
**Human-directed AI review — suggested pre-merge checks**

Reviewed at `4ffbfd3d`. The new baseline resolution and registry validation address the failure path we identified. No changes requested; a few invitations to close the loop:

- **Exercise the current head before merging.** The [current orchestration run](https://github.com/zama-ai/fhevm/actions/runs/34953565311) was skipped; the [successful dispatch](https://github.com/zama-ai/fhevm/actions/runs/34873259308) predates the fix. Please dispatch this head and confirm authenticated package lookup, baseline resolution, artifact download, and `assertLockGhcrTags` succeed. Absence of `published-image-check=skipped` alone is insufficient: empty package-tag sets can still produce unverified pins.
- Consider requiring `resolve-baseline` success before downloading its artifact, so an upstream credential failure does not also produce a misleading missing-lock error.
- Consider rejecting empty/unverified required-package results in strict mode. The later registry assertion protects startup, but cannot recover the intended ancestor fallback.
- Refresh the PR body: it still describes propagation delays and `manifest unknown` retries. Also prefer “baseline tag is unavailable” over “never published”: a 404 alone cannot establish publication history.

The main remaining confidence check is a successful execution of the new workflow path with its actual CI credentials.

fegmorte:
Follow-up for the pre-merge checks, landed in 7d3e0df4a:

The dispatch on 4ffbfd3d already ran authenticated package lookup: [resolve-baseline succeeded](https://github.com/zama-ai/fhevm/actions/runs/34958924042/job/104347586440) with `GH_TOKEN` + `REQUIRE_PUBLISHED_IMAGE_CHECK=true`, and the lock artifact is a verified pin (`requested-sha=c2f416be…`, no `published-image-check=skipped` / `unverified:` sources). Optional unpublished images were omitted. That dispatch is still in progress on the previous head for the rest of the path (builds → `assertLockGhcrTags`).

On this head:
- `create-e2e-tests-input` now requires `resolve-baseline` success before downloading the lock
- strict mode rejects empty required-package tag sets (no unverified pins); local CLI still keeps them
- unverified/`published-image-check=skipped` lock sources fail before stack start
- resolve stdout is no longer swallowed by `tail -n 1`
- 404 copy is “baseline tag is unavailable”
- PR body refreshed

A second dispatch on `7d3e0df4a` is needed for the remaining workflow-path checks (`assertLockGhcrTags` with lock download + the new `if:`).

Eikix (approved):
lgtm

PR #3953, opened 2026-09-15
fix(test-suite): support the config-based kms-gen-keys CLI

eitjuh:
KMS core v0.14.1 replaced the flag-based kms-gen-keys CLI with a single --config-file argument, so the pinned stack no longer boots. Probe --help for --public-storage and use the matching form, keeping older pins working, then bump CORE_VERSION to the stable v0.14.1 tag.

Verified the generated TOMLs parse in the v0.14.1 image for both the centralized and threshold paths.


dd23:
The change was introduced here: https://github.com/zama-ai/kms/pull/675

It removes the config flags from `kms-gen-keys` and replaces it with a TOML config file. In a threshold deployment, each party needs its own config file.

Example threshold config for a single party 3:
```toml

mock_enclave = true

[keygen]

[threshold]
my_id = 3
tls_subject = "dev-kms-core-3.com"
tls_wildcard = true

[aws]
region = "us-east-1"
s3_endpoint = "http://dev-s3-mock:9000"

[public_vault.storage.s3]
bucket = "kms"
prefix = "PUB-p3"

[private_vault.storage.s3]
bucket = "kms"
prefix = "PRIV-p3"
```

See here for example, where we create configs for a set of parties: https://github.com/zama-ai/kms/pull/675/changes#diff-91d64953c8b9045350ecbeb858ea191061534b710615b6f88943394eb5181237

Usage is then `kms-gen-keys --config-file party_<i>_config.toml`, with `<i>` being the respective party id, when we deploy all parties from a central place, like in the test suite or on devnet.

tawadaa:
@dd23 @eitjuh What is the decision on this one, should we merge or discard?

test-suite/fhevm/src/resolve/presets.ts
```diff
@@ -2,7 +2,7 @@
  * Defines hardcoded companion version pins layered onto non-network targets during bundle resolution.
  */
 export const MAINLINE_COMPANIONS = {
-  CORE_VERSION: "b9087af",
+  CORE_VERSION: "v0.14.1",
```
dd23: we can bump to v0.14.2-0 now, if we want, but it will also work as is

test-suite/fhevm/docker-compose/core-docker-compose.yml
```diff
@@ -18,9 +18,18 @@ services:
...
-          # Older cores scope to signing keys via --cmd signing-keys; newer cores dropped it. Probe and inject only when supported.
-          if kms-gen-keys --help 2>&1 | grep -q -- "--cmd"; then CMD="--cmd signing-keys"; else CMD=""; fi
-          kms-gen-keys --public-storage s3 --public-s3-bucket "$$KMS_CORE__PUBLIC_VAULT__STORAGE__S3__BUCKET" --public-s3-prefix "$$KMS_CORE__PUBLIC_VAULT__STORAGE__S3__PREFIX" --aws-s3-endpoint "$$S3_ENDPOINT" --aws-region "$$S3_REGION" --private-storage file --private-file-path "$$KMS_CORE__PRIVATE_VAULT__STORAGE__FILE__PATH" $$CMD centralized
+          # The kms-gen-keys CLI changed across core images: newer cores take a single TOML
+          # --config-file (centralized = no [threshold] section; private vault is file-backed),
+          # while older cores took --public-storage/--aws-region/... flags with a `centralized`
+          # subcommand (and a --cmd signing-keys selector on the oldest). Probe --help and use the
+          # matching form so a pinned old or new CORE_VERSION both boot.
+          if kms-gen-keys --help 2>&1 | grep -q -- "--config-file"; then
```
dd23, line 30: `core-docker-compose.yml:26` probes for `--config-file`; `kms-core.ts:181` probes for `--public-storage`. Both are correct on the two known images, but a reader has to check both strings against both CLIs to trust them, and the PR description mentions only `--public-storage`. Pick one probe string for both files.

test-suite/fhevm/docker-compose/core-docker-compose.yml
```diff
@@ -18,9 +18,18 @@ services:
...
+            printf "[keygen]\noverwrite = false\nshow_existing = false\n\n[aws]\nregion = \"%s\"\ns3_endpoint = \"%s\"\n\n[public_vault.storage.s3]\nbucket = \"%s\"\nprefix = \"%s\"\n\n[private_vault.storage.file]\npath = \"%s\"\n" "$$S3_REGION" "$$S3_ENDPOINT" "$$KMS_CORE__PUBLIC_VAULT__STORAGE__S3__BUCKET" "$$KMS_CORE__PUBLIC_VAULT__STORAGE__S3__PREFIX" "$$KMS_CORE__PRIVATE_VAULT__STORAGE__FILE__PATH" > /tmp/kms-gen-keys.toml
```
mkmks:
```
cat > /tmp/kms-gen-keys.toml <<EOF

your config goes here
EOF
```
would be much more readable and less error-prone than `printf`.

test-suite/fhevm/docker-compose/core-docker-compose.yml
mkmks: If this PR is meant for the 0.14 branch, I'm not sure if there's a need to check which `kms-gen-keys` is available because it only uses config files in KMS 0.14. Is there a real need to maintain backwards compatibility in this setup at all? Parsing `--help` is very brittle.

mkmks (approved):
If the fix works, I'm approving it, but I also have some comments about maintainability (nothing too grave though).

test-suite/fhevm/docker-compose/core-docker-compose.yml
dd23: not for this PR, but maybe we should add a `--version` flag to `kms-gen-keys` to have something a little more robust.

eitjuh: Fixed

test-suite/fhevm/docker-compose/core-docker-compose.yml
eitjuh: changed it to heredoc

test-suite/fhevm/src/resolve/presets.ts
eitjuh: bumped it

mkmks (approved):

dd23 (approved):

PR #3960, opened 2026-09-15
fix(coprocessor): reject GPU image architecture mismatches

antoniupop:


coprocessor/fhevm-engine/tfhe-worker/src/lib.rs
```diff
@@ -94,6 +94,9 @@ async fn async_main_inner(
         let _ = OTEL_GUARD.set(otel_guard);
     });
 
+    #[cfg(feature = "gpu")]
+    fhevm_engine_common::gpu_arch::ensure_matching_visible_devices()?;
```
PanGan21: Agent finding

The `?` propagates the error to `start_runtime_inner`, which only logs it and returns normally, so tfhe_worker exits with status 0 despite failing to start. 
That also prevents on-failure restart policies from restarting it. The SNS and ZK-proof workers already exit with status 1 here.

PanGan21 (approved):

antoniupop:
@mergifyio queue

PR #3961, opened 2026-09-15
refactor(sdk): remove multi-version wasm support

geoxel:
### Summary

- The SDK now ships exactly one `tfhe` and one `tkms` wasm version at a time, so
  all of the machinery that used to support loading/coexisting multiple wasm
  versions simultaneously has been removed.
- Follow-on cleanup: the `dev`/`prod` build-profile split (a leftover from when
  the manifest could list different version sets per profile) is gone too,
  since both profiles now produce byte-identical output.
- Further follow-on: the legacy docker-based `localstack_v11`/`v12`/`v13` test
  chains are no longer reachable now that CI and npm scripts only target the
  latest localstack, so that plumbing is removed too.
- **Behavior change, not just cleanup:** `KMSVerifier < 0.3.0` (Protocol API
  v11) is no longer supported — `readKmsSignersContext-p.ts` now throws
  instead of falling back to the old v11 signer-context read path. Called out
  in its own commit for visibility.

### What changed

1. **`chore(sdk): drop unused tfhe/tkms wasm versions from the manifest`**
   Removes the npm-published wasm binaries/generated loaders for versions no
   longer shipped (tfhe v1.5.3, v1.6.0-dev; tkms v0.13.10, v0.13.20-0).
   Mechanical, no behavior change to the versions still supported.
2. **`refactor(sdk): replace HyperWasmSolver with single-version resolution`**
   `HyperWasmSolver-p.ts` solved for which version(s) to load across a set of
   supported versions; with one version each, that collapses to a fixed
   choice (`WasmVersions-p.ts`). `ProtocolVersionResolver-p.ts` now targets a
   single fixed protocol version instead of resolving one from on-chain
   contract versions.
3. **`refactor(sdk): adapt encrypt/decrypt call sites to single-version wasm`**
   Updates the consumers (encrypt/decrypt modules, kms, key
   (de)serialization, zk-proof handling, frozen-context resolution,
   ethers/viem client config) to the new shapes.
4. **`fix(sdk): drop support for KMSVerifier < 0.3.0 (Protocol API v11)`**
   The one real behavior-breaking change in this PR — see Summary above.
5. **`test(sdk): remove the multi-wasm Playwright test suite`**
   Deletes `test/multi-wasm/` (matrix-driven roundtrip suite exercising
   multiple simultaneously-loaded wasm versions) and its planning notes.
6. **`test(sdk): update test suites for single-version wasm`**
   browser-smoke, browser-next, browser-ui, fheTest, and infra test updates;
   renames `multiWasmHarness.ts` → `smokeHarness.ts`.
7. **`docs(sdk): update docs for single-version wasm`**
8. **`test(sdk): fix memory leakage detection tool`**
9. **`chore(sdk): remove build profiles (prod/dev)`**
   `versionsManifest.js`'s `dev`/`prod` profile split is gone — both profiles
   produced identical output once each library was down to one version.
   Updates `package.json` scripts, CI workflows (`js-sdk-dod.yml`,
   `js-sdk-publish.yml`), and `test-suite/e2e`'s SDK installer accordingly.
10. **`chore(sdk): update wasm:install to prune version dirs no longer in versionsManifest.js`**
    `npm run wasm:install` now deletes any `src/wasm/<lib>/v*` directory that
    isn't listed in the manifest, instead of leaving stale installs behind.
11. **`chore(sdk): tfhe worker pruner: allow the new __wbg___wbindgen_boolean_get_* shim`**
    Unblocks installing tfhe versions built with a newer wasm-bindgen that
    emits this (standard, benign) import shim.
12. **`chore(sdk): wasm-to-base64: record the Source path relative to the repo root`**
    The generated `.wasm.base64.js` header no longer embeds an absolute,
    machine-specific path.
13. **`test(sdk): remove legacy localstack_v11/v12/v13 test support`**
    Drops the docker-based `localstack_v11`/`v12`/`v13` chain configs,
    chain-defaults entries, `setupCommon.ts` wiring, the now-dead
    `localstack_v12` golden permit fixture, and the corresponding
    `cleartext`/`localstack-v11`/`localstack-v12`/`localstack-v13` legs from
    `js-sdk-dod.yml` (CI now only exercises `localstack-latest`). The
    `stalePermitMigration` suites are kept as-is — they synthesize old-era V1
    permits against whatever chain is running rather than depending on an
    actual v11/v12/v13 stack, so migration coverage is unaffected.

### Test plan

- [x] `npm run build` passes (clean, codegen, prettier, lint, cjs/esm/types/tests)
- [x] `npm run codegen:loaders:check` passes
- [x] `npm run wasm:install` installs/prunes correctly against the manifest
- [x] History verified bisectable: tree at each commit boundary matches the
      pre-split state; no new build regressions introduced by the reorg
- [x] `npm run test:unit`
- [x] `npm run test:browser-smoke`
- [x] CI (`js-sdk-dod.yml`) green on this branch — now `quality`, `browser`,
      `localstack-latest`


sdk/js-sdk/src/core/runtime/ProtocolVersionResolver-p.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    // Trailing slashes are normalized before matching relayer URLs.
+    expect(protocolContextForChain(makeChain(`${sepolia.fhevm.relayerUrl}/`)).pubKeyCrsVersion).toEqual({
+      version: '1.4.0-alpha.3',
+      comparator: 'eq',
+    });
+  });
+
+  it('resolves the generated PubKey/CRS version for unknown relayers', () => {
+    expect(protocolContextForChain(makeChain('http://localhost:3000')).pubKeyCrsVersion).toEqual({
+      version: '1.6.1',
```
0xalexbel, line 41: i don't understand 1.6.1

geoxel: `'1.6.1'` is the value of `GENERATED_PUB_KEY_CRS_VERSION` defined at `sdk/js-sdk/src/core/runtime/ProtocolVersionResolver-p.ts:32` : 

`const GENERATED_PUB_KEY_CRS_VERSION: PubKeyCrsVersion = '1.6.1';`

The PubKey/CRS version this SDK expects when key material is freshly generated for its target protocol line (`0.15.0`).

For known public relayers (mainnet/sepolia/dev), the code has an explicit override table (lines 58-71) mapping their relayer URL to `'1.4.0-alpha.3'`, since those relayers were snapshotted serving that version at release time. For any other relayer (like the test's `http://localhost:3000`, which isn't in that table), it falls back to `GENERATED_PUB_KEY_CRS_VERSION`, i.e. `'1.6.1'`. The test at line 41 is just asserting that fallback behavior for an "unknown" relayer.

0xalexbel (approved):

geoxel:
@mergifyio queue

PR #3963, opened 2026-09-15
ci(test-suite): manual Blue/Green QA rounds on live chains

PanGan21:
### Summary

Manual, QA-driven Blue/Green rounds in the preview env, on live chains (Sepolia + Amoy).

**Runbook:** [`ci/preview-env/BLUE-GREEN-QA.md`](https://github.com/zama-ai/fhevm/blob/panos/blue-green-preview-env-qa/ci/preview-env/BLUE-GREEN-QA.md)

### Changes
- **Manual mode.** `--blue-green` without `--tests` now deploys Blue only and does not propose. The operator drives Green, the proposal and the checks by hand.
- **Two-chain rounds.** The upgrade spans Sepolia and Amoy; the single-chain guard is gone.
- **New scripts** for preview-env and Blue Green testing`:
  - `bg-traffic.sh` — one ERC-20 per chain, continuous transfers/mints/decryptions, verified against
    tracked balances
  - `bg-contracts.sh` — host + gateway contract upgrade, driven by `REINITIALIZER_VERSION`
  - `bg-stack.sh` — relayer / kms-connector / test-suite to the target release
  - `bg-green.sh` — Green schema migration and fleet start
  - `bg-propose.sh` — the on-chain proposal, from a laptop
  - `bg-checkpoints.sh` — PASS/FAIL assertions per phase
  - `bg-reset.sh` — back to Blue so the same env can run again

### Round flow

Each step is run by hand; the scripts do one thing each.

| # | Step | Command | What it proves |
| --- | --- | --- | --- |
| 1 | Deploy Blue only, contracts pinned to the previous release | `preview-env launch --blue-green --testnets` (no `--tests`) | the old stack comes up and completes keygen |
| 2 | Deploy the token and start traffic | `bg-traffic.sh setup` then `start` | one ERC-20 per chain, continuous transfers / mints / decryptions |
| 3 | Snapshot on the old release | `bg-traffic.sh stop` then `verify` | every balance decrypts before anything changes |
| 4 | Upgrade contracts, then relayer / kms-connector / test-suite | `bg-contracts.sh upgrade`, `bg-stack.sh upgrade` | the proxies move and the services follow, **before** the cutover |
| 5 | Snapshot again | `bg-traffic.sh verify` | the contract upgrade did not break the running coprocessor |
| 6 | Start Green beside Blue | `bg-green.sh migrate` then `start` | Green migrates its own schema while Blue keeps serving |
| 7 | Resume traffic and leave it running | `bg-traffic.sh start` | the window is exercised under load |
| 8 | Propose | `bg-propose.sh send` | one tx on the gateway carrying a window per host chain |
| 9 | Cutover | `bg-checkpoints.sh cutover` then `post` | Green live, Blue paused, Green schema dropped |
| 10 | Final verify | `bg-traffic.sh verify` | **every** handle the round wrote still decrypts under Green including the ones written after the window opened and before the flip, which are superseded within seconds and are what a cutover could damage |
| 11 | Reset and repeat | `bg-reset.sh` | the same env runs another round without redeploying (~25 min instead of ~1 h) |

The runbook also covers the same round with **no traffic at all**, which is the only way to test that the synthetic work injected at the window open is what carries the consensus when the chains are quiet.


haroldsphinx (dismissed):

fegmorte (dismissed):
Awesome work Panos 🙏

fegmorte (approved):

PR #3965, opened 2026-09-15
fix(coprocessor): isolate host listener poller state

goshawk-3:
Isolates per-stack table writes during Blue-Green dry-run that were falling through the gcs,public search_path into shared public rows, letting the lagging green dry-run stack corrupt or rewind live blue state.

-  host_listener_poller_state: duplicated into gcs and merged GCS-wins at cutover (like gw_listener_last_block), so green's lagging poller cursor can no longer blind-overwrite blue's live cursor.
-  host_chain_consumer_blocks, bridge_handle_events, delegate_user_decrypt, drift_revert_signal: duplicated into gcs for isolation only (dropped at cutover, not merged), since these are host-listener/gw-listener writes blue already re-derives independently.


PanGan21 (approved):

goshawk-3:
@mergifyio queue

PR #3969, opened 2026-09-15
chore(host-contracts): replace internal function by uint8 type guard

jatZama:


host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -691,7 +686,7 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
         if ((1 << uint8(resultType)) & supportedTypes == 0) revert UnsupportedType();
 
         uint256 maxSize = (resultType == FheType.Uint64 || resultType == FheType.Uint128)
-            ? FHE_COLLECTION_WIDE_MAX_SIZE
+            ? uint256(FHE_COLLECTION_WIDE_MAX_SIZE)
             : FHE_COLLECTION_NARROW_MAX_SIZE;
```
obatirou: only one of the 2 is casted
same below

obatirou: if not mistaken the cast can be dropped?

host-contracts/contracts/FHEVMExecutor.sol
```diff
@@ -1043,21 +1038,9 @@ contract FHEVMExecutor is UUPSUpgradeableEmptyProxy, FHEEvents, ACLOwnable {
     ///      rotating a single minted operand through every slot.
     function _consumeOperand(bytes32 ct, uint256 position) internal view virtual returns (uint256 shiftedBit) {
         if (!ACL.isAllowed(ct, msg.sender)) revert ACLNotAllowed(ct, msg.sender);
-        if (position >= 256) revert BoundaryBitPositionOverflow(position);
```
antoniupop, line 1046: _consumeOperand still takes an arbitrary uint256 and is internal virtual. A future caller or derived implementation could pass >= 256 
EVM SHL then yields zero, silently dropping the boundary bit and reopening the aliasing issue this backstop protects against.

Could we keep BoundaryBitPositionOverflow, the runtime check, and its test?

host-contracts/contracts/FHEVMExecutor.sol
jatZama: It's because only one is needed to cast the `uint8` type to `uint256` here. It cannot be dropped otherwise it would fail at compilation.

host-contracts/contracts/FHEVMExecutor.sol
jatZama: Why ? It adds gas cost and increases the bytecode size of the contract. I prefer to remove it, I already put this warning in the natspecs. An alternative would be to pass a uint8 position argument instead of a uint256.

jatZama: I "fixed" the issue you pointed by using the alternative method I suggested in 4094a412a797ad79caff4a417b8bf6d67d5c6c5a i.e passing a `uint8` position argument in `_consumeOperand`. Wdyt?

antoniupop (dismissed):

host-contracts/contracts/FHEVMExecutor.sol
obatirou: I do not understand: 
`FHE_COLLECTION_WIDE_MAX_SIZE` and `FHE_COLLECTION_NARROW_MAX_SIZE` are both `uint8` and are the 2 possible results of the ternary.
Why would only one need a cast to `uint256`?

obatirou: In local I removed the cast and ran `forge test --match-path 'test/fhevmExecutor/fhevmExecutor.t.sol'` and it works.
Anyway I might miss something and this might not be worth taking time for this.

jatZama: Ah you were right, this is not needed for correctness/compilation, my bad. This is what codex told me:
```
Yes—both casts can be dropped without changing behavior. Solidity implicitly widens the uint8 conditional result to the uint256 maxSize variable.
I kept them as a small compiler optimization. Rechecking the current code with Solidity 0.8.24 and the project's settings:
- With casts: 20,315 bytes of runtime bytecode.
- Without casts: 20,321 bytes, due to two additional masks.
- ABI unchanged; both compile successfully.
So they're optional for correctness. Removing them favors simpler source at a tiny bytecode/gas cost. No files changed.
```

So my answer: not a strong opinion , but maybe we should keep them to save few bytes of bytecode, since we are near contract size limit of 21 KB.

obatirou: Ok good for me then either: both are casted or neither are casted.
Will let you decide on that.

jatZama: Ok I casted both collection constants in this commit 930526abf89906a6322615ab8272e34d210a198b btw notice this is just for readability, it does not impact bytecode size (still at 20,315 bytes).

obatirou: Great, thanks!

obatirou (dismissed):

obatirou (dismissed):

obatirou (approved):

isaacdecoded (approved):

0xalexbel (approved):

PR #3972, opened 2026-09-16
fix(coprocessor): materialize the source ciphertext on BridgeHandle

PanGan21:
## Summary

Bridging a transient-only allowed handle could leave the destination waiting indefinitely for a source ciphertext that was never materialized.

- Mark valid bridged source handles for computation and enqueue SnS.
- Queue SnS after recording computations, so handles created and bridged in the same transaction are linked to the correct source block.
- Add coprocessor regression tests for scheduling, block tracking, and invalid destination chain IDs.
- Add an e2e test that computes and bridges a handle without persistent ACL permission, then decrypts it on the destination.

## Validation

- Added the e2e regression test first: it [failed before the coprocessor fix](https://github.com/zama-ai/fhevm/actions/runs/34889331975/job/104127766681) and [passed after the fix](https://github.com/zama-ai/fhevm/actions/runs/34980261172)


coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
```diff
@@ -2443,6 +2443,41 @@ impl Database {
         Ok(inserted)
     }
 
+    /// Enqueues SnS for bridged handles: `send` accepts a transient allowance,
```
dartdart26: I think it would be good to explain what `send` is in this context, because it is not easy to understand if the reader doesn't know the bridging contracts and LZ specifics.

coprocessor/fhevm-engine/Cargo.lock
dartdart26, line 1: Is the change in this file absolutely needed? Maybe we can do it as part of another PR?

dartdart26 (approved):
Looks good, thanks!

coprocessor/fhevm-engine/Cargo.lock
PanGan21: I made it here in order to finish faster, but can be made as another pr ofc.

coprocessor/fhevm-engine/host-listener/src/database/tfhe_event_propagate.rs
PanGan21: Yes you are right. LZ specifics are not relevant to coprocessor

PanGan21: The comment has been improved and doesn't include LZ specifics anymore

coprocessor/fhevm-engine/Cargo.lock
PanGan21: This change has been removed from this pr and merged separately: https://github.com/zama-ai/fhevm/pull/3985

paulo-zama (approved):

haroldsphinx (approved):

PR #3973, opened 2026-09-16
test(consensus): add deferred manual fault and recovery campaigns

antoniupop:


Eikix:
### Automated review

Reviewed head `c203589e583ace983ba91650b450639d75d61075`. One confirmed test-coverage finding; no production defect demonstrated. Checked existing PR comments and reviews: this finding has not already been posted.

**[P2] The operand-order assertion can pass without proving that order affects the rerandomization seed.**

In [the rerandomization test](https://github.com/zama-ai/fhevm/blob/c203589e583ace983ba91650b450639d75d61075/coprocessor/fhevm-engine/scheduler/src/dfg/scheduler.rs#L1052-L1062), swapping encrypted `[17, 29]` to `[29, 17]` necessarily changes the output list because the plaintext order changes. Comparing the whole list therefore does not prove that operand order affects the seed: the assertion would still pass if the seed calculation incorrectly treated the inputs as unordered.

Could this check use three inputs, swap the first two, and compare only the unchanged third input's output ciphertext? That would isolate the effect of order on the seed. Similarly, when replacing the second input with a fresh encryption of 29, compare the unchanged first input's output rather than the whole list.

This is a suggested correction to the existing test, not a request for another campaign. The gap was checked against the code; the cryptographic mutation was not executed. No new live campaigns were run for this review loop.

Eikix (approved):

antoniupop:
@mergifyio queue

eudelins-zama:
@mergifyio dequeue

eudelins-zama:
@mergifyio queue

PR #3975, opened 2026-09-16
feat(relayer): http minimal first version with endpoints and aggregator

nboisde:


relayer-http/crates/relayer-http/src/settings.rs
```diff
@@ -0,0 +1,262 @@
...
+/// No `deny_unknown_fields` at this level: every `APP_*` variable of the pod lands here as a key, and a typo in
+/// `kms_aggregator` still fails as a missing field. The nested structs are strict.
+#[derive(Debug, Clone, Deserialize)]
+pub struct Settings {
+    pub name: String,
+    #[serde(default)]
+    pub log: LogConfig,
+    pub http: HttpConfig,
+    pub kms_aggregator: KmsAggregatorConfig,
+}
```
geoxel, line 21: Just a thought: the file is named config.yaml, the struct is named Settings, and inside, we have Config's.
What about naming this struct RelayerConfig instead ? And maybe renaming the file to config.rs or relayer_config.rs ?

relayer-http/crates/relayer-http/src/settings.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub endpoint: SocketAddr,
+    /// Largest accepted request body; bigger ones are answered `400 malformed`.
+    #[serde(default = "default_max_body_bytes")]
+    pub max_body_bytes: usize,
+    /// Host chain ids a ciphertext handle may carry (bytes 22..30 of the handle).
+    pub supported_chain_ids: Vec<u64>,
+}
+
+fn default_max_body_bytes() -> usize {
+    2 << 20
```
geoxel, line 37: Why ? And why is it 2^21 in the default config.yaml file ?

relayer-http/crates/relayer-http/src/kms_aggregator/client.rs
```diff
@@ -0,0 +1,624 @@
...
+use bytes::{Bytes, BytesMut};
+use kms_connector_api::{ErrorCode, ErrorResponse};
+use reqwest::header::{AUTHORIZATION, CONTENT_TYPE, HeaderValue};
+use serde::de::DeserializeOwned;
+use url::Url;
+
+use super::config::{AuthConfig, ConfigError, EndpointConfig, KmsAggregatorConfig};
+
+/// Response bodies above this are dropped (a connector answer is a few KiB).
+pub const MAX_RESPONSE_BYTES: usize = 4 << 20;
```
geoxel, line 16: Strange to read "a connector answer is a few KiB" and below a max size of 4 MiB :-)

relayer-http/crates/relayer-http/src/kms_aggregator/client.rs
```diff
@@ -0,0 +1,624 @@
...
+            .connect_timeout(CONNECT_TIMEOUT.min(cfg.call.timeout))
+            // Safety net for a socket that stops answering; the aggregator's deadline is the decision.
+            .timeout(cfg.call.timeout)
+            // Never forward the key to another host.
+            .redirect(reqwest::redirect::Policy::none())
+            .https_only(https_only)
+            .no_proxy()
+            .pool_idle_timeout(Duration::from_secs(10))
+            .tcp_keepalive(Duration::from_secs(15))
+            .user_agent(concat!("zama-relayer-http/", env!("CARGO_PKG_VERSION")))
```
geoxel, line 178: Nice to send the version

relayer-http/crates/relayer-http/src/kms_aggregator/client.rs
```diff
@@ -0,0 +1,624 @@
...
+            request = request.header(AUTHORIZATION, auth.clone());
+        }
+        let mut response = request.send().await.map_err(transport)?;
+        let status = response.status().as_u16();
+        let too_large = || AttemptError::Body("response too large".to_owned());
+        if response
+            .content_length()
+            .is_some_and(|len| len > MAX_RESPONSE_BYTES as u64)
+        {
+            return Err(too_large());
```
geoxel, line 220: We could log `len` here or put it in `too_large` return value ?

relayer-http/crates/relayer-http/src/kms_aggregator/client.rs
```diff
@@ -0,0 +1,624 @@
...
+        }
+        let mut buf = BytesMut::new();
+        while let Some(chunk) = response.chunk().await.map_err(transport)? {
+            if buf.len() + chunk.len() > MAX_RESPONSE_BYTES {
+                return Err(too_large());
```
geoxel, line 225: Same here.

relayer-http/crates/relayer-http/src/kms_aggregator/client.rs
```diff
@@ -0,0 +1,624 @@
...
+    use alloy::primitives::B256;
+    use kms_connector_api::UserDecryptionResponse;
+    use wiremock::matchers::{body_bytes, header, method, path};
+    use wiremock::{Mock, MockServer, ResponseTemplate};
+
+    use super::*;
+    use crate::kms_aggregator::config::tests::valid;
+    use crate::kms_aggregator::mock::{self, Fixed};
+
+    const ALL_CODES: [ErrorCode; 14] = [
```
geoxel, line 272: We could add ErrorCode::UnsupportedAttestationType too.

relayer-http/crates/relayer-http/src/kms_aggregator/call.rs
```diff
@@ -0,0 +1,370 @@
...
+    async fn attempt_loop<R: DeserializeOwned>(
+        &self,
+        endpoint: &Endpoint,
+        route: &str,
+        body: Bytes,
+        deadline: Instant,
+        attempts: &mut u32,
+    ) -> Result<R, CallError> {
+        loop {
+            let Ok(permit) = self.permits.acquire().await else {
```
geoxel, line 117: Claude review (correctness / availability) 1 / 3 : 

The call semaphore holds a permit across the whole HTTP attempt. With shipped defaults (13 endpoints, `max_concurrent_calls: 64`, 5s deadline) only ~4 aggregations fit; hung nodes park permits for the full deadline, so a 5th concurrent request can burn its entire deadline inside `permits.acquire()` and return `504 timeout` with `counted: 0, attempts: 0` — having contacted no node — while `/healthz` still reports `ready`.

relayer-http/crates/relayer-http/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+    // Serves until `shutdown` is cancelled, then drains the in-flight requests (bounded by call.timeout).
+    let mut server = tokio::spawn(endpoint::serve(app));
+    tokio::select! {
+        result = &mut server => {
+            error!(result = ?result, "server stopped unexpectedly");
+            return ExitCode::FAILURE;
+        }
+        () = wait_for_signal() => {
+            shutdown.cancel();
+            if let Err(e) = server.await {
```
geoxel, line 52: Claude review (correctness / availability) 2 / 3 : 

`if let Err(e) = server.await` only matches the `JoinError`; the inner `io::Result` from `endpoint::serve` is dropped, so a drain-time server failure exits `SUCCESS` and looks like a clean stop.

relayer-http/crates/relayer-http/src/kms_aggregator/client.rs
```diff
@@ -0,0 +1,624 @@
...
+            Self::Body(_) | Self::Transport(_) => false,
+        }
+    }
+
+    /// The connector's own table (`ErrorCode::retryable`); unknown codes follow the body's flag.
+    pub fn is_retryable(&self) -> bool {
+        match self {
+            Self::Api { error, .. } if error.code == ErrorCode::Unknown => error.retryable,
+            Self::Api { error, .. } => error.code.retryable(),
+            Self::Status(status) => *status == 408 || (500..600).contains(status),
```
geoxel, line 114: Claude review (correctness / availability) 3 / 3 : 

`is_retryable` covers `408` and `500..600` but not `429`. A rate limit answered by an ingress with a non-connector body is terminal after one attempt and surfaces as `upstream_transient` (502) instead of `rate_limited` (429) — inconsistent with the same status carrying a connector body, which does get retried.

relayer-http/crates/relayer-http/src/endpoint/error.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// The request is wrong: unreadable or oversized body, invalid JSON, unknown field, wrong content type, or a
+    /// validation rule. `message` names the field.
+    pub fn malformed(request_id: &str, message: impl Into<String>) -> Self {
+        Self::new(StatusCode::BAD_REQUEST, "malformed", request_id, message)
+    }
+
+    /// The aggregator's verdict. A dominant connector error is answered with the connector's own status and code.
+    pub fn aggregation(request_id: &str, error: AggregationError) -> Self {
+        match error {
+            AggregationError::Timeout {
```
geoxel, line 53: Claude review (diagnostics / hardening):

The `Timeout` arm discards `rejected` and `dominant` (unlike `ThresholdNotReached`), so a failure actually caused by 5 rejections reports only "6 of 9 responses". `AggregationError::dominant()` is unreachable for that variant.

relayer-http/crates/relayer-http/src/endpoint/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+        .layer(DefaultBodyLimit::max(app.http.max_body_bytes))
+        .with_state(app)
+}
+
+/// Binds `http.endpoint` and serves until the shutdown token is cancelled, then drains the in-flight requests
+/// (bounded by `call.timeout`: every running aggregation ends with `Cancelled`).
+pub async fn serve(app: App) -> std::io::Result<()> {
+    let listener = TcpListener::bind(app.http.endpoint).await?;
+    let shutdown = app.shutdown.clone();
+    axum::serve(listener, router(app))
```
geoxel, line 38: Claude review (diagnostics / hardening):

`axum::serve` has `DefaultBodyLimit` but no header/body-read or idle timeout, so slow-loris connections live indefinitely; the aggregator deadline never applies because the handler is never entered.

relayer-http/crates/relayer-http/src/endpoint/docs.md
```diff
@@ -0,0 +1,197 @@
...
+| `POST` | `/v4/exp/public-decrypt` | public decryption: the plaintexts and the KMS signatures over them |
+| `GET` | `/liveness` | the process runs |
+| `GET` | `/healthz` | the pod is ready to serve (503 while shutting down) |
+
+Configuration (`http:` in `config/config.yaml`): `endpoint` (bind address), `max_body_bytes` (default 2 MiB),
+`supported_chain_ids` (host chain ids a handle may carry). One port for everything; no version endpoint yet.
+
+## 2. Request payloads
+
+The shapes are the current relayer's, so the SDK keeps working. Hex fields are typed: a handle is 32 bytes, an
```
geoxel, line 24: Claude review (docs):

§2 still claims "the shapes are the current relayer's, so the SDK keeps working", but `9fba04ee7` renamed `attestedPayload`→`payload` and dropped `version`/`type`. With `deny_unknown_fields` the old body now gets `400 malformed`, and the example further down already shows the new shape — the section contradicts itself.

relayer-http/crates/relayer-http/src/kms_aggregator/client.rs
nboisde: Changing the comment since its wrong indeed -> around 1MiB.

nboisde: Changing also formats for also better readability.

relayer-http/crates/relayer-http/src/config.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub endpoint: SocketAddr,
+    /// Largest accepted request body; bigger ones are answered `400 malformed`.
+    #[serde(default = "default_max_body_bytes")]
+    pub max_body_bytes: usize,
+    /// Host chain ids a ciphertext handle may carry (bytes 22..30 of the handle).
+    pub supported_chain_ids: Vec<u64>,
+}
+
+fn default_max_body_bytes() -> usize {
+    2 << 20
```
nboisde: Allowing more space, but indeed need to recalculate properly the max payload size, changing this. 
Exploration will be required as well for further flows.

relayer-http/crates/relayer-http/src/kms_aggregator/call.rs
nboisde: I agree, the default value must be raised.

relayer-http/crates/relayer-http/src/endpoint/error.rs
nboisde: The example is does not work in this case, but I will change the message to forward the context.

PR #3977, opened 2026-09-16
chore(contracts): create the protocol-contracts directory and claim ownership

leventdem:
Creates `protocol-contracts/` and its CODEOWNERS entry, ahead of the packages that will fill it.

Part of [`planning-blockchain#1314`](https://github.com/zama-ai/planning-blockchain/issues/1314) — `staking`, `governance`, `safe`, `token`, `feesBurner` and `pauserSetWrapper` are moving here from `zama-ai/protocol-apps`.

## Why this goes first, separately

**CODEOWNERS is evaluated against the base branch.** Without this entry on `main`, each of the move PRs would route for review to `@zama-ai/fhevm-devs` — a team that owns none of this code and has no context on it. With it, they route to `@zama-ai/fhevm-protocol-contracts`.

## What is in it

Two files:

- `protocol-contracts/README.md` — which packages are coming and which two stay with Protocol Apps.
- `.github/CODEOWNERS` — one line, `/protocol-contracts/ @zama-ai/fhevm-protocol-contracts`.

## Notes for the reviewer

- The directory is **created** here rather than forward-declared: every other path in `.github/CODEOWNERS` currently resolves to something that exists, and this keeps that true.
- `@zama-ai/fhevm-protocol-contracts` already has push access to this repository.
- No build, test or workflow is touched. The only executable change is review routing.


protocol-contracts/README.md
```diff
@@ -0,0 +1,49 @@
...
+| `feesBurner` | Receives protocol fees and burns them | to move |
+| `pauserSetWrapper` | Wraps the pauser set so governance can administer it | to move |
+
+The Solana OFT programs are tracked separately, since they cannot live inside an EVM package —
+[`planning-blockchain#1352`](https://github.com/zama-ai/planning-blockchain/issues/1352).
+
+`confidential-wrapper` and `confidential-token-wrappers-registry` are **not** moving; they stay with
+Protocol Apps.
+
+## Conventions
```
isaacdecoded: Not sure we want to define these conventions in this PR. Maybe better to discuss and define them after migrating the projects, so we have a better picture of everything.

protocol-contracts/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Each package is self-contained and follows `gateway-contracts` rather than the layout it had in
+`protocol-apps`:
+
+- **One package per directory**, each with its own `package.json`, lockfile, `hardhat.config.ts` and
+  `foundry.toml`.
+- **`selectors.txt` is per package**, as in `gateway-contracts/selectors.txt`. In `protocol-apps` a
```
isaacdecoded: If we decide to keep the conventions, careful here since we talked about keeping a single selectors file for these protocol-contracts.

protocol-contracts/README.md
leventdem: Yes I removed this section for now

protocol-contracts/README.md
leventdem: let's discuss it with the team first, I agree

isaacdecoded (dismissed):
LGTM!

isaacdecoded (approved):

PR #3978, opened 2026-09-16
chore(test-suite): proxy connector e2e

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/2045

Successful e2e tests [run](https://github.com/zama-ai/fhevm/actions/runs/35217740664)

### Summary

Adds the `kms-connector-proxy` to the e2e stack and routes the existing Connector HTTP suites through it over TLS, with an API key, instead of talking to `kms-connector[-i]-endpoint` directly.

### Stack

A `kms-connector-proxy` service is added per KMS party (`kms-connector[-i]-proxy`), configured to forward requests to `kms-connector[-i]-endpoint`.
`CONNECTOR_PROXY_VERSION` is added to optionally configure the image of the proxy.
The compatibility gate becomes `supportsConnectorHttp`: the endpoint and the proxy are now gated together as a pair, so a bundle missing either one drops both services and the suites skip.

### TLS and auth

A self-signed test certificate is checked in under `test-suite/fhevm/static/config/kms-connector-proxy`, with a README documenting how it was generated. It is its own trust anchor and its SANs cover `kms-connector-proxy` plus `kms-connector-{2..32}-proxy`, so one pair serves any party count the harness supports. The e2e container trusts it through `NODE_EXTRA_CA_CERTS`; the thin connector client now sends `Authorization: Bearer $KMS_CONNECTOR_API_KEY` on every call, and the proxy checks it against the `KMS_CONNECTOR_API_KEY_DIGEST` in the connector env. Both are test-only values, checked in like the rest of `templates/env`.

### CI

The orchestrator and the e2e workflow pass a `proxy` entry in `connector-versions`, and the failure log snapshot includes the proxy container.


test-suite/fhevm/docker-compose/kms-connector-docker-compose.yml
```diff
@@ -62,5 +62,17 @@ services:
       kms-connector-db-migration:
         condition: service_completed_successfully
 
+  kms-connector-proxy:
```
nboisde, line 65: Is there a way somehow to include an exposed port, and compatible with kms threshold scenario.
It will help us to streamline tests of the relayer and ease development process. 
Also could a bit "mimic" exposition.

nboisde (approved):
Thanks @eudelins-zama ! LGTM regading proxy integration. 
If we could add port exposure for proxy in this PR or a following one would be amazing.

test-suite/fhevm/docker-compose/kms-connector-docker-compose.yml
eudelins-zama: AFAIK, the relayer runs inside the compose network, and the proxy is reachable there in threshold today!
The current 'simple' HTTP client we use instead of Relayer currently hits `https://kms-connector[-i]-proxy:8443` per party
So AFAIK, no host port is needed, just the URLs in the relayer config. Publishing ports to the host is a different thing (per-party port allocation, trusting self-signed cert etc.), and I'd rather not add it unless you specifically want to run the relayer on the host. Is that the case?
Either way I think it fits better in the relayer e2e PR where the exact need is clear. Wdyt?

Happy to huddle about it if needed!

eudelins-zama: Discussed in huddle :slightly_smiling_face:

eudelins-zama:
@mergifyio queue

PR #3979, opened 2026-09-16
refactor(relayer): parse v3 user-decrypt as one tagged envelope

Eikix:
## Summary
- Parse `/v3/user-decrypt` as one internally tagged `UserDecryptV3RequestJson` instead of peeking `attestationType` and deserializing two sibling structs.
- OpenAPI now documents the POST body as a oneOf of the EIP-712 and Solana sRFC-38 arms.
- Connector comments, README naming, and `HOST_CONFIG_SEED` now follow the Solana field names already on the ABI/DB (`solanaRequest`), and `ed25519-dalek` is a relayer test dependency only.

## Test plan
- [x] Relayer unit tests for the tagged envelope, unknown tag, Solana payload `deny_unknown_fields`, and `validate_v3_attestation_type` allowlist
- [x] Relayer `solana_envelope_fixture` and `solana_v0_teardown`
- [x] Relayer HTTP tests: unknown/missing `attestationType`, Solana sRFC-38 accept/reject, unknown extra envelope field, EVM direct-handle accept
- [x] kms-worker `check_user_decryption_request_v2` unit tests (comments-only connector change)
- [ ] Relayer CI `user-decrypt` and `solana` groups on this PR


Eikix:
Review of 47e2a0eac (2 commits) against feature/solana, with a clean-code and lean-code lens. CI green except the two pending jobs (solana-e2e scenarios, kms-connector tests).

### Verdict

The core change is a real simplification and should land. Before this PR the `attestationType` tag lived in five places: the handler's probe struct, a `String` field on each of the two envelope structs, the `validate_v3_attestation_type` allowlist, and the `match` inside the EVM `TryFrom`. After it, the tag lives in one place, the serde `tag` on `UserDecryptV3RequestJson`, and an unknown tag cannot reach either arm by construction. The handler drops from two parse paths to one. That is the lean shape.

Three things pull against that: a string-matching heuristic added to the generic JSON parser to preserve an error body only a relayer test pins, a `deny_unknown_fields` attribute that does nothing where it sits, and documentation that records a retired name. Details below.

### Verified

- `UserDecryptV3RequestJson` needs the hand-written `Validate` impl: validator_derive 0.20.1 only walks `Data::Struct`, so a derive on the enum would validate nothing.
- `HOST_CONFIG_SEED` re-export is a true dedup: `zama_solana_acl::HOST_CONFIG_SEED` is `b"host-config"` and the zama-host program already re-exports it the same way.
- `validate_handles_and_extract_chain_id` has one caller, the V2 EVM check, so the rewritten doc comment is accurate.
- `publish_user_decryption_v3` exists in gw-listener as the README now claims.
- `openapi.yml` is generated (`openapi-export` bin, `make openapi-check`), so the oneOf block is tool output, not hand-maintained.
- The `ed25519-dalek` move to dev-dependencies compiles (relayer build jobs passed).
- `attested_user_decrypt_rejects_unknown_attestation_type` really tests the tag: internally tagged enums read the tag before the buffered content, so `"attestedPayload": {}` never gets a chance to fail first.

### Should fix

1. **`parsing.rs` now knows about v3.** `is_v3_attestation_variant_error` imports the two v3 tag constants into the generic JSON-error mapper and string-matches serde's "unknown variant ... expected ..." text to rebuild the previous 400 body (field `attestationType`, message "Unsupported attestationType; expected one of: [...]"). The only consumer of that exact text is `user_decrypt_v3_test.rs:393`; the SDK never reads the message. Two leaner options, pick one:
   - Let serde's message stand. It already names the allowed tags. Update the test expectation and drop `is_v3_attestation_variant_error` and the two imports. The field name falls back to `request`, which is the trade.
   - Keep the exact body but put the knowledge in the v3 layer: add a `#[serde(other)] Unknown` unit variant to the enum (serde supports `other` on internally tagged enums) and have `TryFrom<UserDecryptV3RequestJson>` return the "Unsupported attestationType" error. That is typed code where the tag is defined, not a regex over a library's error string in a module every endpoint shares.
   Either way the generic parser stops carrying endpoint-specific constants.

2. **`#[serde(deny_unknown_fields)]` on the enum is a no-op.** serde_derive applies it in `de/struct_.rs` and `de/enum_adjacently.rs` only. For an internally tagged enum with newtype variants, the inner structs' own `deny_unknown_fields` is what rejects a stray envelope key, and the doc comment two lines up already says so. Remove the attribute so the code does not claim a guarantee it is not providing; `v3_rejects_unknown_extra_field_on_envelope` keeps passing.

3. **Rename leftovers.** `relayer/Cargo.toml:77` still says "Canonical `hostPayload` serialization". The PR's stated goal is that comments follow `solanaRequest`; this one was missed.

### Lean nits

- The `info!` calls in both `TryFrom` impls ("Converting X to UserDecryptRequest") are progress narration. The Solana one now logs `attestation_type` as a compile-time constant, which carries no information. Drop both or move to `debug!`.
- The kms-connector README table documents `hostPayload` as a "retired name". A name that no longer exists in code does not belong in the reader-facing README; the SQL migration comment already records the history. Drop that row, and consider whether the table belongs in `docs/architecture.md` next to the rest of the connector data model rather than in the top-level README between configuration and Support.
- The second commit only tidies the first (removes the validator the first commit orphaned). Squash before merge.

### No action

- Comment-only edits in the connector ("Solana V2" to "Solana", `hostPayload` to `solanaRequest`) are consistent and correct.
- `is_field_specific_serde_error` is a good extraction; adding "unknown variant" there is right for every endpoint, not just v3.
- The fixture and teardown test updates are mechanical and keep the same assertions.

Eikix:
Follow-up review at b673594af (the two commits since 47e2a0eac). Seven CI jobs were still pending at review time, none failing.

Commits reviewed: a9353421f (drop the v3 sniffing from the generic parser) and b673594af (`Unknown` variant on the envelope enum).

### Earlier items

| # | Item | State |
|---|------|-------|
| 1 | v3 constants leaked into `parsing.rs` | RESOLVED. `is_v3_attestation_variant_error` and both imports are gone. The parser now returns serde's own text for any `unknown variant` and falls back to field `request`, endpoint-agnostic. The exact 400 body is kept by a `#[serde(other)] Unknown` variant whose `Validate` impl emits `attestation_type` + "Unsupported attestationType; expected one of: [...]"; `to_camel_case` in the v2 error mapper turns the key into `attestationType`, so `v3_rejects_unknown_attestation_type` keeps its unchanged expectation. |
| 2 | No-op `deny_unknown_fields` on the enum | RESOLVED. Attribute removed; the inner structs keep theirs. |
| 3 | `hostPayload` leftover in `relayer/Cargo.toml` | RESOLVED. |
| nit | Narration `info!` in both `TryFrom` impls | RESOLVED, both removed. |
| nit | README "retired name" row | RESOLVED, row dropped. Table stays in the top-level README. |
| nit | Squash | OPEN. Four commits; the last three all refine the first. |

### New at head

1. **The hand-written OpenAPI schema is the price of the `Unknown` variant, and it is mostly justified.** utoipa 5.5 has no way to skip an enum variant in the derive, so hiding `Unknown` from the spec needs a manual `PartialSchema`. `openapi.yml` is byte-identical to the previous head, so the manual builder reproduces the derive's oneOf exactly. Two trims would make it cheaper to own:
   - `impl ToSchema for UserDecryptV3RequestJson {}` is enough. Both default methods do the right thing (`name()` returns the type name, `schemas()` is a no-op), and the two inner types are already listed in `components(schemas(...))` in `openapi.rs`, so the twelve-line `schemas()` body registers them a second time for no effect.
   - `Ref::from_schema_name("AttestedUserDecryptRequestJson")` and the Solana literal should be `Ref::from_schema_name(AttestedUserDecryptRequestJson::name())`. A rename then fails to compile instead of silently producing a dangling `$ref`.
   - The description string duplicates the doc comment word for word. Keep one; the doc comment is the one a reader finds.
2. **No workflow runs `make openapi-check`.** The Makefile target exists but nothing in `.github/workflows` invokes it, so the manual schema drifting from the derive output would not be caught by CI. Not introduced by this PR, but this PR is the first place a hand-written schema can drift. Worth a follow-up issue or a one-line addition to the relayer lint job.
3. **`TryFrom<UserDecryptV3RequestJson>` has an unreachable `Unknown` arm.** `parse_and_validate` runs `validate()` before the conversion, so `Unknown` never reaches `try_from` on the HTTP path. The arm is needed for exhaustiveness and correctly reuses `unsupported_attestation_type_message`, so there is nothing to change; a one-line comment saying validation rejects it first would stop the next reader from hunting for the path.
4. **The `solana_envelope_fixture` `Unknown` branch** maps an unknown tag to `RejectedBy::JsonShape`, which matches the layer the fixture names. Fine.

### Verdict

The head is leaner than the previous head: the parser is generic again, the enum owns the tag and the rejection, and the OpenAPI output is unchanged. The remaining asks are small: drop the redundant `schemas()` body and the string-literal refs, squash, and decide who owns the missing `openapi-check` in CI.

Eikix:
Hard requirement before merge: no hand-written OpenAPI schema. `openapi.yml` is generated from the utoipa derives, and the moment one component is built by hand (`impl PartialSchema for UserDecryptV3RequestJson`, 67318e99a) the spec has two sources of truth and nothing in CI checks the second one. Every v3 type must derive `ToSchema`.

The manual impl exists only to hide the `#[serde(other)] Unknown` variant from the spec, and that variant exists only to keep the exact 400 body (`attestationType` / "Unsupported attestationType; expected one of: [...]"), which nothing outside `user_decrypt_v3_test.rs` depends on. So the lean shape is the one a9353421f already had, minus the extra variant:

- Drop `Unknown`, `unsupported_attestation_type_message`, the `PartialSchema`/`ToSchema` impls and the `tagged_openapi_arm` helper. `#[derive(ToSchema)]` on the enum produces the same oneOf.
- Let serde reject an unknown tag. `parsing.rs` already maps `unknown variant` to a 400 with serde's own message, which names the two allowed tags. Update `v3_rejects_unknown_attestation_type` to expect field `request` and that message.
- If the exact `attestationType` field in the 400 is considered part of the API contract, the remap belongs in the v3 handler, after `parse_and_validate` returns, not in the generic parser and not in the type: a few lines that only the v3 handler owns, with no effect on the schema.

Then regenerate `openapi.yml` and confirm `make openapi-check` passes. Adding that target to the relayer clippy job is a separate two-line PR; happy to open it.

PR #3986, opened 2026-09-16
fix(sdk): harden hexToBytes and fix abortableSleep listener leak

geoxel:
## Description
Two small correctness fixes in `sdk/js-sdk/src/core/base/`, ported from changes reviewed in #3934.

### `hexToBytes` (bytes.ts)
- Throws on non-hexadecimal characters instead of silently coercing them to `0` via `parseInt` returning `NaN`.
- Strips the `0x` prefix case-insensitively, fixing incorrect parsing of uppercase-prefixed input (e.g. `"0Xff"` was previously mis-parsed as `[0, 255]` instead of `[255]`).

### `abortableSleep` (timeout.ts)
- The `abort` event listener was only removed when the signal actually aborted (via `{ once: true }`). When the sleep resolved normally instead, the listener stayed attached to the `AbortSignal` indefinitely.
- Now explicitly removes the listener when the timer fires and the promise resolves, preventing a listener leak on long-lived or reused `AbortController` instances.


0xalexbel (approved):

geoxel:
@mergifyio queue

PR #3990, opened 2026-09-16
feat(kms-connector): multi attestation user decrypt

eudelins-zama:
Adapt to tech-spec [update](https://github.com/zama-ai/tech-spec/pull/517).

We changed the API to support multiple attestation type, so integrating non-EVM chains like Solana is easier in the future.


kms-connector/crates/api/src/types.rs
```diff
@@ -48,20 +48,28 @@ sol! {
...
-        /// The user's EIP-712 signature.
-        bytes signature;
         bytes extraData;
     }
+
+    /// `POST v1/user-decrypt` request body: the user's signature over the protocol inputs of a
+    /// user decryption, tagged with the scheme it was produced with.
+    #[derive(Debug, PartialEq, Serialize, Deserialize)]
+    #[serde(deny_unknown_fields)]
+    struct UserDecryptionRequest {
```
Eikix, line 67: [AUTOMATED] Crate header still says v1 shapes do not change; this wrap is an in-place v1 change.

kms-connector/crates/endpoint/src/core/validation.rs
```diff
@@ -32,6 +34,25 @@ pub enum ValidationError {
...
+
+fn supported_attestation_types() -> String {
+    AttestationType::supported().collect::<Vec<_>>().join(", ")
+}
+
+impl ValidationError {
+    /// The error code this failure is answered with.
+    pub fn code(&self) -> ErrorCode {
+        match self {
+            Self::UnsupportedAttestationType(_) => ErrorCode::UnsupportedAttestationType,
```
Eikix: [AUTOMATED] Module doc still says every failure is `malformed`.

kms-connector/crates/api/src/types.rs
eudelins-zama: Yep, that is an exception as we're not in prod yet

kms-connector/crates/endpoint/src/core/validation.rs
eudelins-zama: `UnsupportedAttestationType` maps to `400 malformed` as stated in the module doc, in terms of HTTP return code, so I think it's still accurate

eudelins-zama: Would need to be `400 Bad Request` if we want to be precise, not malformed indeed

eudelins-zama: Fixed

kms-connector/crates/api/src/types.rs
```diff
@@ -48,20 +48,28 @@ sol! {
...
+        string attestationType;
+        UserDecryptionPayload payload;
+        /// The user's signature, in the scheme named by `attestationType`.
+        bytes signature;
```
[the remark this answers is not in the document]
eudelins-zama: Yep, that is an exception as we're not in prod yet

kms-connector/crates/api/src/attestation.rs
```diff
@@ -0,0 +1,19 @@
+//! The `attestationType` discriminant of `v1/user-decrypt`.
+
+use strum::{Display, EnumIter, EnumString, IntoEnumIterator, IntoStaticStr};
+
+/// The signature schemes a `v1/user-decrypt` `signature` can be produced with.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Display, EnumString, EnumIter, IntoStaticStr)]
+pub enum AttestationType {
+    /// Unified EIP-712 user decryption per RFC 016: the signature is the user's EIP-712 (or
+    /// ERC-1271) signature over the `payload`.
```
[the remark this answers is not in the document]
eudelins-zama: Fixed b101aa60a4361541961589bce0c0a6045569dcd1

test-suite/e2e/test/sdk/connector/connectorHttp.ts
```diff
@@ -49,16 +49,22 @@ export interface RequestValidity {
   durationSeconds: number;
 }
 
-export interface UserDecryptionRequest {
+/** The RFC 016 fields the user signs over. */
```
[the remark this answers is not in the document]
eudelins-zama: Fixed b101aa60a4361541961589bce0c0a6045569dcd1

kms-connector/crates/api/src/types.rs
```diff
@@ -48,20 +48,28 @@ sol! {
         uint64 durationSeconds;
     }
 
-    /// `POST v1/user-decrypt` request body: the protocol inputs of a user decryption
-    /// (unified shape per RFC 016).
+    /// The protocol inputs of a user decryption, i.e. the `payload` of a [`UserDecryptionRequest`].
     #[derive(Debug, PartialEq, Serialize, Deserialize)]
     #[serde(deny_unknown_fields)]
-    struct UserDecryptionRequest {
+    struct UserDecryptionPayload {
```
jot2re, line 54: I know it is not part of the RFC and would complicate compatibility but should the request `AttestationType` be part of the authenticated payload to prevent any potential future downgrade attacks?

test-suite/e2e/test/sdk/connector/connectorHttp.ts
```diff
@@ -49,16 +49,22 @@ export interface RequestValidity {
   durationSeconds: number;
 }
 
-export interface UserDecryptionRequest {
+/** The RFC 016 protocol inputs. All fields except `handles` are signed over. */
+export interface UserDecryptionPayload {
```
jot2re, line 57: Since `handles` are not signed should it be part of the outer type and not the payload type?

kms-connector/crates/endpoint/src/core/http/user_decrypt.rs
```diff
@@ -29,7 +29,7 @@ pub async fn user_decrypt(
 ) -> Result<HttpResponse, ErrorResponse> {
     let request = body.into_inner();
     validate_user_decryption(&request, &state.config)
-        .map_err(|e| ErrorResponse::new(ErrorCode::Malformed, e.to_string(), None))?;
+        .map_err(|e| ErrorResponse::new(e.code(), e.to_string(), None))?;
```
jot2re, line 32: Small nitpick from Claude: the public decryption still hardcodes `Malformed` error instead of deriving it from the validation result.

kms-connector/crates/endpoint/src/core/validation.rs
```diff
@@ -46,15 +74,22 @@ pub fn validate_user_decryption(
     request: &UserDecryptionRequest,
     config: &Config,
 ) -> Result<(), ValidationError> {
-    validate_handles(request.handles.iter().map(|h| &h.handle), config)?;
-    if request.allowedContracts.len() > config.max_allowed_contracts {
+    let _attestation_type = request
+        .attestationType
+        .parse::<AttestationType>()
+        .map_err(|_| {
+            ValidationError::UnsupportedAttestationType(request.attestationType.clone())
```
jot2re: Claude mentioned that this might be an issue in case of malicious calls since we now clone the whole content, so this would make a DOS by out-of-memory easier

jot2re (commented):
Overall looks good to me. Just some minor comments and a one that is a bit bigger, but probably something we don't want to change at this point.

test-suite/e2e/test/sdk/connector/connectorHttp.ts
eudelins-zama: I don't think so.
There is a tricky nuance, because `decryption_id` is computed over the entire `UserDecryptionPayload`, so we must keep handles there.
For signature, this is dictated by [Decryption.sol](https://github.com/zama-ai/fhevm/blob/main/gateway-contracts/contracts/Decryption.sol#L50-L69) (not in our scope) and handles are excluded from this `UserDecryptRequestVerification`.
The comment "All fields except `handles` are signed over." was not there in the first place and was flagged by Copilot as inaccurate (which is true indeed), thus why it was added.
But I understand it can add such confusion...

kms-connector/crates/api/src/types.rs
eudelins-zama: - for the `decryption_id` computation, this is done over the entire `UserDecryptionRequest`, so already used
- for the `UserDecryptRequestVerification` signature this is not doable without breaking existing `relayer-sdk` and Gateway Decryption verification
  - this should not be that bad as long as we don't mess up when introducing new schemes (attacks would require the new scheme use the same EIP-712 struct and domain with weaker kms-connector checks AFAIU)

kms-connector/crates/endpoint/src/core/http/user_decrypt.rs
eudelins-zama: Fixed 338399b22a0871afa39dcbc3aeefa96ac3c852e5

kms-connector/crates/endpoint/src/core/validation.rs
eudelins-zama: Fixed 338399b22a0871afa39dcbc3aeefa96ac3c852e5

We no longer embed the `attestationType` in the response sent to the client.
The client should know what it has sent.

kms-connector/crates/api/src/types.rs
jot2re: Yeah, I know it is not simple to update. I just wanted to mention it for completeness. I agree it is not a show-stopper

test-suite/e2e/test/sdk/connector/connectorHttp.ts
jot2re: Can you add a clarifying comment at the struct definition what is signed and what is used for deriving the `decryption_id`? I think that would be helpful :)

jot2re (dismissed):
LGTM! Maybe just add the an extra comment to the `UserDecryptionPayload` struct like discussed in-line

test-suite/e2e/test/sdk/connector/connectorHttp.ts
eudelins-zama: Done 5c53f49e8a5cd3252d52ab6afdd1d2dd95316171 !

jot2re (approved):
LGTM!

eudelins-zama:
@mergifyio queue

PR #3995, opened 2026-09-16
chore(sdk): upgrade tfhe wasm to v1.8.1 and kms_lib wasm to v0.15.0-0

geoxel:
## Summary

Upgrades the SDK's WASM dependencies on top of the single-version WASM loading introduced in `devex/geoxel/remove-multi-wasm`:

- `tfhe` wasm → **v1.8.1** (replacing v1.6.2)
- `kms_lib` wasm → **v0.15.0-0** (replacing v0.14.0-1)

### Supporting changes

- **`install-tfhe.sh` / `install-from-manifest.mjs`**: tfhe-rs 1.7.0 split the client-only wasm build into its own npm package (`tfhe-client`); versions before that keep shipping under `tfhe`. Both scripts now pick the right package name based on the requested version.
- **`build-tfhe.ts`**: fixes the default type-check import — `WasmAssetLoadMode` lives in `core/types/wasmAssets.ts`, not `coreFhevmRuntime.ts`.
- **`browser-smoke`** (`smoke-cdn.ts`, `smoke-csp-block.ts`, `smoke-wasm.ts`): wasm paths now derive from `CANONICAL_WASM_VERSIONS` instead of being hardcoded, so they track the manifest automatically.


0xalexbel (dismissed):

geoxel:
@mergifyio queue

0xalexbel (approved):

PR #3996, opened 2026-09-16
feat(solana): close every program-owned account for the preview host

Eikix:
DD-051 requires every acquire and destroy of the preview environment to leave zero accounts owned by the preview host program `DPq5y89RDZPq9NcMh9X1NgjBWgYmSXg3QoipSBV3ZMzQ`. Only the owner program can zero an account's lamports or reassign it, and zama-host had no instruction that does so. This PR adds that instruction and the deployer command that drives it. Wiring the command into the preview-env workflows follows in #3939.

## Program

`close_owned_accounts` is compiled only behind a new `admin-sweep` feature, which the `preview-env` profile enables. Localnet and future production builds do not contain the instruction.

- The signer must be the program's upgrade authority, checked against `ProgramData` like `initialize_host_config`. `HostConfig.admin` is not used because a half-initialized deployment has no config.
- Targets are passed as remaining accounts and stay untyped: an account left by an older layout would fail to deserialize, and that leftover is what the wipe must delete.
- An account the program does not own is skipped rather than rejected, so a stale `getProgramAccounts` page cannot fail a batch.
- Each owned account is closed the standard way: lamports moved to the signer, data resized to zero, owner assigned to the System program.

## Deployer

`host wipe` first checks that the program is deployed and that the deployer is its upgrade authority, so a wrong profile or cluster fails instead of printing a clean sweep. It then lists the program's accounts with `getProgramAccounts` (`dataSlice` length 0), sends the instruction in batches of 25 writable accounts, and re-lists with a `minContextSlot` taken after the last confirmation, failing if any owned account remains. In the deployer image it needs only `SOLANA_RPC_URL` and the deployer keypair.

The instruction is absent from the vendored IDL by design, so the deployer builds it by hand with a pinned discriminator instead of a Codama client. A Rust test pins the discriminator and the account order and flags against the Anchor-generated `instruction::CloseOwnedAccounts` and `accounts::CloseOwnedAccounts`.

## Tests

Mollusk runs against `target/deploy` artifacts built on default features, which do not contain the instruction. `check-zama-host-idl.sh` now also builds zama-host with `admin-sweep` as `zama_host_admin_sweep.so`. Only the two sweep tests load it; every other suite still runs against the artifact that ships. The vendored IDL is unchanged.

- Closes a HostConfig and an arbitrary owned account named twice, refunds the signer once per account, leaves a foreign program's account untouched.
- Rejects a signer who is not the upgrade authority, even when that signer is `HostConfig.admin`, and rejects a finalized program.

## Follow-ups (deliberately out of scope)

Accounts owned by the shared demo programs (`confidential_token`, `demo_vault`, `confidential_batcher`) are not closed; DD-051 still needs a decision there (same instruction per program, or throwaway program keypairs per preview).

The wipe needs the instruction in the bytecode already on chain. `DPq5y89…` is not deployed today, so the first preview deploy carries it, and every later cycle can wipe before it upgrades.

Nothing exercises `host wipe` end to end: localnet builds lack the feature, so the e2e deployment test cannot run it. The deployer path is covered by the wire-format pin only; the first real run is the preview-env trial.

## Validation

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` on `zama-host` and `zama-solana-runtime-tests`, `cargo clippy -p zama-host --features decode`.
- `bash scripts/check-zama-host-idl.sh`: both artifacts built, vendored IDLs and ABI goldens unchanged.
- `cargo test --workspace` in `solana/`: all suites pass, including the three new tests.
- `bun run check` in `solana/deploy`.
- `bash scripts/dead-surface-check.sh`.

Cost worth a conscious yes: the IDL gate now compiles `zama_host` twice (two feature sets). A plain `anchor build` leaves `zama_host_admin_sweep.so` stale, in which case the two sweep tests fail on a missing-instruction error until the script is rerun.


Eikix (commented):
Review of head `99febb34a`, two passes: one on lean/quality, one independent on correctness and security with recomputation outside the crate. No correctness or security defect found. CI is green at head, including the two-artifact build, the new Mollusk tests and `solana-e2e/scenarios`.

**Verified clean**

- The gate cannot be spoofed. `Account<'info, ProgramData>` enforces the loader owner and the bincode variant, and the `address` constraint fixes the account to `get_program_data_address(&crate::ID)`. A finalized program has `upgrade_authority_address == None` and never matches. Both rejections are covered by Mollusk tests. The instruction grants the upgrade authority nothing it could not already do by replacing the program.
- Foreign accounts are skipped before any borrow, and the refund test asserts the foreign account byte-for-byte. The `ProgramData` account, the program account and the admin all fall into the same skip because none is owned by the program.
- The close pattern is right and in the right order. Agave refuses an owner change unless the data is already zeroed, so `resize(0)` must precede `assign`, and the resize is permitted because it is checked against the pre-instruction owner. Zero lamports, zero data and the system owner leave nothing to revive.
- Duplicate targets are safe because the SBF entrypoint's duplicate marker makes the second `AccountInfo` share the first one's `owner` pointer, so the second pass reads `system_program` and skips. The test with `owned` listed twice and an exact lamport-sum assertion proves no double refund, and Mollusk drives the real loader serializer, so this is the real aliasing, not a simulation.
- Discriminator recomputed independently: `sha256("global:close_owned_accounts")[..8]` is `[36, 68, 214, 114, 46, 227, 146, 228]`, matching `wipe.ts` and the Rust wire-format pin. `PROGRAM_DATA_AUTHORITY_OFFSET = 4 + 8` matches bincode's u32 tag, u64 slot and 1-byte option tag, so the authority is read at 13.
- Feature gating: `#[cfg(feature = "admin-sweep")]` sits on the `#[program]` handler and on the module and re-export. The default build compiles only if Anchor propagated the `cfg` into the dispatch arm. The IDL and ABI goldens are unchanged because the sweep build uses `--no-idl` and only the default build writes `target/idl`. `build-programs.sh` passes `preview-env` only for that profile, so the localnet image carries neither feature.
- `minContextSlot` verification is fail-closed: the slot is read after the last confirmation, and a lagging node returns `MinContextSlotNotReached`, which throws, rather than an empty list.
- Message size at 25 targets recomputed by hand: 28 unique keys give about 1038 bytes, under the 1232 limit.
- No bare "devnet" for the Solana cluster in the added text.

**Recommended changes**

1. **Assert the instruction is absent from the shipped artifact.** Every new test loads `zama_host_admin_sweep.so`. Nothing loads the default `zama_host.so` and confirms it rejects the discriminator. That absence is the point of the feature gate, and today it rests on "the default build compiles". One Mollusk test against the default artifact expecting Anchor's `InstructionFallbackNotFound` closes it and becomes the second half of the wire-format pin: the same bytes asserted present in one artifact and absent in the other. `host_mollusk.rs` already loads that artifact, so no new harness.
2. **Exercise the 25-target transaction once.** `TARGETS_PER_TRANSACTION = 25` is tested nowhere; the Mollusk tests pass 4 remaining accounts. Compute cost is estimated at roughly 100 CU per target on top of the `ProgramData` deserialize, far under the default 200k, but estimated only. One Mollusk test with 25 program-owned targets proves the loop and prints the CU number.
3. **Slice the `ProgramData` fetch.** `assertUpgradeAuthority` in `wipe.ts` fetches the whole account, which is the full program bytecode, to read 45 bytes. Add `dataSlice: { offset: 0, length: 45 }`; the offsets are relative to the slice start and stay valid.
4. **Doc comment on `assertUpgradeAuthority` overstates its coverage.** It says a wrong profile or cluster must not print a clean sweep. That holds only when the resolved program is undeployed or has a different authority. A deployed program with the same authority and zero owned accounts prints a clean sweep without proving the instruction exists in its bytecode. Benign, but the comment should say what it covers.

**Nits and observations**

- `HostConfigAdminMismatch` is raised for an upgrade-authority mismatch that the module doc says is not `HostConfig.admin`. It mirrors `initialize_host_config.rs`, so consistency is right here; the rename is a separate cleanup.
- The TS discriminator and account order are pinned in Rust with a "keep in sync" comment but nothing mechanical links them. Editing `wipe.ts` fails no test. Optional to fix.
- README and DD-051 prescribe wipe before upgrade, which needs the deployed `.so` to already carry `admin-sweep`. The preview program id does not exist on Solana devnet yet and only the `preview-env` profile writes it, so every build it will hold has the feature. Targets are untyped, so upgrade-then-wipe also works and has no such precondition.
- `wipe.ts` has no executable coverage: `solana/deploy` has no tests and the localnet program cannot carry the instruction. Everything above about the CLI is from reading it. The first live `host wipe` against the preview host is the acceptance step.

PR #4013, opened 2026-09-17
feat(host-contracts): add BNB and Hoodi to the upgrade windows

PanGan21:
### Summary

Adds the new host chains to the coprocessor-upgrade proposal tooling, which builds one dry-run window per chain from that chain's block time.

- Hoodi and BNB testnet on devnet and testnet, BNB mainnet on mainnet only.
- Block times measured over 2000 blocks: BNB 0.45s, Hoodi 13.2s against a nominal 12s.
- Runbook: every operator must index every chain in the proposal, or the attempt expires instead of cutting over.

Closes https://github.com/zama-ai/fhevm-internal/issues/2049


isaacdecoded (approved):

host-contracts/tasks/utils/environments.ts
```diff
@@ -85,6 +115,7 @@ export const ENVIRONMENTS: Record<string, EnvironmentDef> = {
     chains: [
       { chainId: 1, label: 'ethereum', fallbackBlockTimeSeconds: 12, rpcUrlEnv: 'ETHEREUM_MAINNET_RPC_URL' },
       { chainId: 137, label: 'polygon', fallbackBlockTimeSeconds: 2, rpcUrlEnv: 'POLYGON_MAINNET_RPC_URL' },
+      { chainId: 56, label: 'bnb', fallbackBlockTimeSeconds: 0.45, rpcUrlEnv: 'BNB_MAINNET_RPC_URL' },
```
obatirou, line 118: Note that hardhat.config.ts do not have a BSC mainnet entry yet.
Not sure if you want to add it now or in a new PR
should this be a fix in 0.14 ?

obatirou (approved):

host-contracts/tasks/utils/environments.ts
PanGan21: For now this is not needed on hardhat.config.ts but let's verify if we want to have this at all here until it is supported

PR #4019, opened 2026-09-17
feat(solana): encode host chain ids with type byte 0x01

Eikix:
## Summary

Solana host `chain_id` is now type byte `0x01` plus a 56-bit cluster tag. The old SHA-256 + bit-63 marker is gone. EVM stays type byte `0x00`. Anything else is refused.

This is the fhevm half of kms#856. Handle bytes 22–29 already carry a `u64` big-endian chain id. The high byte is the type. Localnet is `solana_host_chain_id(12345)` = `0x0100000000003039`. Public clusters take the first seven bytes of genesis.

Helpers live in the owners that already decided host vs EVM (host constants, coprocessor `ChainId`, relayer event, connector config, JS `utilsSolana.ts`). No new crate.

## Pins

- kms-grpc / bc2wrap: `93af722c8d1011fb5fe19a16e27dd08dab32b993` (kms#856)
- Solana TKMS WASM: `v0.15.0-0-solana.93af722c`
- Linker vectors copied from that same KMS commit
- e2e `CORE_VERSION=93af722` so the kms-core image matches the client after 856 is on `solana`

EVM tkms `0.14.0-1` is unchanged.

## Out of this PR

- Dual-read of bit 63
- EVM handle minting
- Gateway contract changes
- Coprocessor BIGINT schema rewrite

## Test plan

- [x] `zama-solana-permit` suite (type-byte derivation, regenerated permit vectors)
- [x] Connector auth vectors regenerated; `rejects_chain_id_inconsistent_with_chain_kind`
- [x] `fhevm-engine-common` lib tests, including Solana zk aux layout
- [x] Coprocessor pre-commit `cargo check` + clippy
- [ ] Relayer / SDK / host program CI on this PR
- [ ] e2e after the kms-core image exists at `93af722`


Eikix (commented):
Review of head `b92fa037f`, two passes (correctness and lean/maintainability). The encoding itself checks out: all predicates are equality tests, the fixtures and vectors recompute, the IDL golden is in sync, and every `SOLANA_CHAIN_TYPE_BIT`, negative id and `9223372036854788153` is gone. The items below are what stands between this head and a mergeable one. The Cursor CLI review loop on this branch already flagged the "not Solana ⇒ EVM" dispatch and is closing it; it is listed under item 4 for completeness, not re-raised.

**Blocking**

1. **The SDK now sends an odd-length hex chain id, and nothing tests the string it actually sends.** `uintToHex0x` is `0x${n.toString(16)}` with no padding. Under bit 63 every Solana id had a non-zero high nibble, so this always produced 16 nibbles. With type byte `0x01` the high nibble is always zero, so every Solana cluster id now serialises as `0x100000000003039` (15 nibbles) at `submitInputProof.ts:64` and `fetchCoprocessorSignatures.ts:28`. `submitInputProof.test.ts:121` was hand-edited to the padded `0x0100000000003039`, which is the failing assertion in `js-sdk-tests/test` and `dod quality`. The relayer parses with `u64::from_str_radix` and survives odd length, but its fixture `SOLANA_CHAIN_ID_HEX` at `event.rs:1315` is also the padded form, so neither side exercises the real wire string, and any consumer that byte-decodes the field breaks. Fix: pad Solana ids to 16 nibbles at the call sites (a `u64ToHex0x` with `padStart(16, '0')`), keep the test expectation, and give the relayer test both forms with the unpadded one pinned as "what the SDK sent before this fix".

2. **Every pin points at the squash-merged PR head, not at `solana`.** kms#856 merged by squash as `6b36473c7`; `93af722c8` is not an ancestor of `solana`. This head pins it in `kms-connector/Cargo.toml` (`kms-grpc`, `bc2wrap` revs), `KMS_BUILT_FROM` (`kms_branch=elias/solana-chain-type-byte`, `kms_head=93af722c8…`, `tkms_wasm_version=v0.15.0-0-solana.93af722c`) and `test-suite/fhevm/solana-images.env` (`CORE_VERSION=93af722`). The e2e failure is exactly this: `ghcr.io/zama-ai/kms/core-service:93af722: manifest unknown`. The image will only ever exist for the merge commit. Move the revs to `6b36473c7`, regenerate the WASM set and `KMS_BUILT_FROM` from it, and set `CORE_VERSION` to the published tag.

3. **`cargo fmt --check` fails in all three Rust workspaces, and fmt runs before the tests.** Relayer (`settings.rs:2310`, `event.rs:1329`, `acl_checker.rs:1115`), connector (`decryption.rs:1025`, `kms_worker.rs:613`, `solana/deployment.rs:24,143`, `tests/solana_deployment_identity.rs:21`) and the Solana workspace (`zama-solana-permit/tests/vectors.rs:744`). Because the fmt step gates them, `cargo test --workspace`, the `decode` feature tests, clippy, the connector tests and the relayer tests have not run at this head. Do not read "five of six failures are cosmetic" as "the Rust side is green". One spot to watch once they run: `decryption.rs:1029` changed `resolve([7;32], chain_id | SOLANA_CHAIN_TYPE_BIT)` to `resolve([7;32], chain_id)` with `chain_id` extracted from the test handle, so the handle fixtures there must now embed `0x01` or `resolve` returns `ChainTypeByteInvalid`.

**Lean and abstractions**

4. **The eight-item encoding block is pasted into four Rust files and two TS files.** Same `EVM_CHAIN_TYPE`, `SOLANA_CHAIN_TYPE`, `CHAIN_TYPE_SHIFT`, `CLUSTER_TAG_MASK` and four helpers in `zama-host/src/constants.rs`, `fhevm-engine-common/src/chain_id.rs`, `kms-worker/src/core/config.rs` and `relayer/src/core/event.rs`. The relayer and the connector already depend on `zama-solana-acl`, and `zama-host` already re-exports `HOST_CONFIG_SEED` from it "so the program and off-chain readers cannot drift". Define the block once in `zama-solana-acl`, re-export from `zama-host`, import in the connector and relayer. The coprocessor common crate has no Solana dependency and is the one justified copy. The copies have already drifted in surface: the relayer exports `is_evm_host_chain_id` and `chain_type_byte` that nothing there calls. In TS, `utilsSolana.ts` exports `isSolanaHostChainId` but `solana/index.ts` does not, so the demo dapp, the deployer (`bootstrap.ts:100`) and the test suite each hand-roll `((x >> 56n) & 0xffn) !== 0x01n` seven times although both link `@fhevm/sdk` from source. Export it and use it. Separately, dispatch in `zkproof-worker/src/auxiliary.rs:25`, `transaction-sender/src/ops/verify_proof.rs:452`, `buildInputProofMetaData-p.ts` and the relayer's `InputProofRequest::try_from` is still binary (`if solana else evm`), so a `0x02` byte takes the EVM path where kms now refuses it. The predicates themselves are all correct. This is the item the Cursor loop is already fixing with an explicit reject arm.

5. **Compat code kept after its reason disappeared.** `SOLANA_HOST_CHAIN_ID_I64 = SOLANA_HOST_CHAIN_ID` in `solana/deploy/src/constants.ts:4` is an alias of itself, threaded through `coprocessor.ts` and `test-suite/fhevm/src/solana/addresses.ts`, and `deploy.ts:326` now prints the same number twice. `solanaHostChainIdI64` in `test-suite/fhevm/src/generate/solana.ts:43-49` still two's-complements "so a leftover bit-63 id round-trips"; no code path can produce one. `ChainId::from_canonical_u64` keeps its bit-63 bitcast rationale. Their tests are now tautologies: `addresses.test.ts:25-26` asserts `f(x) == f(x)` on one binding and `solana.test.ts:56-57` asserts an identity function is an identity function. Both are in `solana-tests/test-suite-fhevm`, which passed. Delete the alias, the helper, the tests and the stale doc comment.

6. **`solana_host_chain_id` silently truncates.** All copies do `(0x01 << 56) | (tag & CLUSTER_TAG_MASK)`. `kms-connector/tests/solana_support/mod.rs:53` calls it with the full `0x0123_4567_89ab_cdef` and gets the right answer only because the discarded byte equals the type byte. A `debug_assert!(cluster_tag <= CLUSTER_TAG_MASK)` makes the intent visible. In `solana_deployment_identity.rs:50` the test is still named `…without_the_chain_kind_bit…` and hard-codes `0x0123_4567_89ab_cdef & 0x00ff_ffff_ffff_ffff` instead of `CHAIN_ID & CLUSTER_TAG_MASK`.

7. **Stale prose.** `DESIGN_DECISIONS.md:430` and `EVM_PARITY.md:21` still say "high-bit chain id" / "high-bit Solana chain classification"; `DESIGN_DECISIONS.md:2546` still says "what that predicate must become"; "chain-kind" survives in `solana/deployment.rs:34`, `tests/evm_path_unchanged.rs:43` and `tests/solana_deployment_identity.rs:46,182`. The `20260605120000_relax_chain_id_checks_for_solana_host.sql` comment is historical and must stay, but it is worth one line noting that the `>= 0` CHECK it dropped is now looser than the type-byte design needs.

**Verified clean**

- `solana_linker_v1.json` digest `f93384ed…` equals the `.sha256` sidecar and the kms file at `93af722c8` byte for byte; `handle-chain-type-byte` and `handle-without-the-solana-type-byte` present, old ids absent.
- Public ids recomputed from genesis hashes: devnet `0x01ce59db5080fc2c`, mainnet-beta `0x0145296998a6f8e2`, testnet `0x013a132ece10305e`; PoC id `0x0100000000003039` = 72057594037940281 everywhere.
- Permit fixtures regenerated to the real rule: `permit_v1.json` genesis `4539cf79…` gives `0x014539cf79f66704` = 91542930698168068, which is the fixture `CHAIN_ID` and the `Chain id:` line in the signed text. The fhevm permit set is no longer a stand-in. The `connector_auth_v1.json` mismatch is the documented synthetic pair, unchanged from before.
- `InvalidChainTypeBit` → `InvalidChainTypeByte` is a rename in place at code 6057 in `errors.rs` and the IDL golden; `check-zama-host-idl.sh` passed.
- WASM revendor: only the `93af722c` file set remains, all import sites updated, transcripts differ only in `host_chain_id`, the embedded handle bytes and signatures. Item 2 means it has to be redone once more from the merge commit.
- No `Number()`/`parseInt` on a Solana id in the SDK, deployer or test suite. No bare `devnet`/`testnet`/`mainnet` in added lines.
- kms-connector is the one component that fails closed on both sides (`validate_host_chain_configs` cross-checked against `chain_kind`). The host program rejects a bad byte in `initialize_host_config`.

**CI at `b92fa037f`:** 67 pass, 6 fail, 1 pending. `js-sdk-tests` and `dod quality` fail on item 1 (one test of 1385). `relayer-test`, `kms-connector-tests` and `solana-tests/build-and-test` fail on item 3. `solana-e2e/scenarios` fails on item 2. None is a flake.

Eikix (commented):
Re-review at head `22c12414b` (three commits on top of the reviewed `b92fa037f`: `67509aa64` "refuse unknown host chain type bytes", `c785578ce` "pad chain-id hex and pin kms squash merge", `22c12414b` "type-check the u64ToHex0x padding test"). Items 1 to 3 and 5 of the first review are addressed; item 2 needs one action in the kms repo before the e2e can pass; items 4, 6 and part of 7 are untouched.

**Blocking**

1. **Fixed in `22c12414b`, CI pending.** At `c785578ce` the new `u64ToHex0x` test called both hex helpers with a raw `bigint` where the branded `Uint` is required (`error TS2345`), so `npm run lint` failed before any test ran and took `js-sdk-tests/test` and the five `dod` jobs with it. `22c12414b` wraps the arguments in `asUint(...)` like the rest of the file. The production change itself is right: `submitInputProof.ts` now sends `u64ToHex0x(inputProof.chainId)`, the relayer gained `solana_input_proof_accepts_unpadded_type_byte_hex` pinning the 15-nibble form and `input_proof_request_refuses_unknown_chain_type_byte`. `fetchCoprocessorSignatures.ts:28` still uses `uintToHex0x`; that path carries EVM ids only, so it is not a bug, but the two sites should agree.

2. **The kms-core image for the new pin does not exist, and nothing will build it automatically.** The pins moved to `6b36473c7` (`kms-grpc`/`bc2wrap` revs, `KMS_BUILT_FROM` now `kms_branch=solana`, `CORE_VERSION=6b36473`), which is correct: it is the `solana` branch head and its tree (`0392b6669…`) is identical to `93af722c8`, so the renamed WASM set is byte-identical and honest. But `solana-e2e/scenarios` fails with `cannot inspect ghcr.io/zama-ai/kms/core-service:6b36473: manifest unknown`. In kms, `main.yml` runs on push only for `main` and `release/*`, and `docker-build.yml` runs on release or `workflow_dispatch`. The three previous `solana` images (`79de4926d`, `fa749daab`, `0ead5f740`) were all built by manual dispatch, and the merge commit has zero workflow runs. Someone with kms Actions rights has to run `gh workflow run docker-build.yml --repo zama-ai/kms --ref 6b36473c7bc2720fa0d4d3d735c3e6b213931e3b`, then re-run the e2e job. Worth a line in the PR body so the next revendor does not rediscover this.

**Addressed since `b92fa037f`**

- `cargo fmt` clean in all three workspaces. The Rust jobs are now actually running (connector, relayer, Solana workspace, coprocessor clippy and tests all in progress at review time); no verdict yet.
- Unknown type bytes are refused at every dispatch site: `ChainId::is_evm_host` added and used in `zkproof-worker/auxiliary.rs`, `verifier.rs`, `transaction-sender/verify_proof.rs` (drops the proof with an error log), relayer `Settings::validate_host_chains`, `HostAclChecker::new`, both relayer `TryFrom` request parsers, SDK `buildInputProofMetaData` and `ZkProofBuilder.build()`, connector `validate_host_chain_configs` (two new `0x02` cases), and `initialize_host_config` (`rejects_unknown_host_chain_type_byte`). Each has a `0x02` test. This closes item 5 and matches kms fail-closed behaviour.
- The connector fixture failure predicted in item 3 was real: `rand_solana_handle()` now embeds `solana_host_chain_id(12345)` at bytes 22..30 for the three Solana-backend dispatch tests.
- `SOLANA_HOST_CHAIN_ID_I64`, `solanaHostChainIdI64` and both tautological tests are deleted; `deploy.ts` prints the id once. `from_canonical_u64` doc now says type-byte ids do not need the bitcast path.
- `high-bit` wording fixed in `DESIGN_DECISIONS.md:430` and `EVM_PARITY.md:21`; `vectors.rs` module doc no longer names `zama-solana-chain-id-v1`.

**Still open**

3. **Item 4, one definition.** The eight-item block is still pasted into `zama-host/src/constants.rs`, `fhevm-engine-common/src/chain_id.rs`, `kms-worker/src/core/config.rs` and `relayer/src/core/event.rs`, and the new head added `EVM_CHAIN_TYPE` and `isEvmHostChainId` to the TS copy too. `solana/index.ts:109` still exports only `defineFhevmSolanaChain`, and the seven inline `(x >> 56n) & 0xffn` predicates in the demo dapp, deployer and test suite are unchanged. Move the Rust block into `zama-solana-acl` (relayer and connector already depend on it; `zama-host` already re-exports `HOST_CONFIG_SEED` from it), export the TS predicate from the Solana index and use it. This can be a follow-up if the branch needs the revendor landed first, but it should be filed before merge.

4. **Item 6.** `solana_deployment_identity.rs:51` still hard-codes `0x0123_4567_89ab_cdef & 0x00ff_ffff_ffff_ffff` instead of `CHAIN_ID & CLUSTER_TAG_MASK` (the test was renamed, the body was not). No `debug_assert!(cluster_tag <= CLUSTER_TAG_MASK)` in `solana_host_chain_id`, so `solana_support/mod.rs:53` still passes a full 64-bit value that happens to round-trip.

5. **Item 7 leftovers.** "chain-kind" survives in `kms-worker/src/core/solana/deployment.rs:32`, `tests/solana_deployment_identity.rs:182` and `relayer/src/host/acl_checker.rs:126`; `DESIGN_DECISIONS.md:2547` still says "what that predicate must become".

**CI at `c785578ce`** (the last completed head): 6 failures, all explained: `js-sdk-tests/test` and five `dod` jobs on item 1, `solana-e2e/scenarios` on item 2. The Rust test and clippy jobs, relayer integration, `solana-tests/build-and-test`, deployer images and `dod localstack` were still running. `22c12414b` re-triggers everything; the e2e will fail again until the kms image is dispatched.

PR #4038, opened 2026-09-17
fix(listener): use TLS for https Yellowstone endpoints

Eikix:
## Why

In the first live Solana preview (run 35234087525, namespace `fhevm-ci-eikix-g6pgrxh`) the Solana host listener never received a block. It logged, every 2 seconds:

```
gRPC subscription dropped; reconnecting inclusively  error="subscribe"
```

The listener builds tonic with only the `server` feature and calls `Channel::from_shared(url).connect()` with no TLS config. Against the hosted Yellowstone endpoint (`https://laserstream-devnet-ewr.helius-rpc.com`) that opens a plaintext HTTP/2 connection, which the server closes at the first request. The e2e stack's local geyser plugin is plaintext, so CI never exercises this path.

Evidence: `grpcurl` from a laptop with the same token and the same block + sysvar filters streams blocks; `grpcurl -plaintext` to the same host:443 gets an immediate EOF, matching the listener's behaviour.

## What

- Enable tonic's `tls-ring` and `tls-webpki-roots` features for the host listener and configure TLS on the endpoint when the URL scheme is `https`. Plaintext `http://` endpoints (e2e) are unchanged.
- Log the full error chain (`{err:#}`) in the two reconnect paths. `%err` printed only the outermost context, which hid the transport error behind the word "subscribe".

## Verification

- `cargo check -p host-listener --features solana-grpc,solana-reconstruct` passes.
- `cargo test -p host-listener --features solana-grpc,solana-reconstruct --lib solana_grpc`: 30 passed.
- Live check against Helius follows in the next preview dispatch; this PR body will be updated with the result.


Eikix (commented):
[AUTOMATED]

Independent review of a9cf335aa…4d9c0af43 (TLS for https Yellowstone). Three passes: open-ended, maintainability, design/integration. Findings were checked against tonic 0.14.6 and this tree.

**Accepted.** `Channel::from_shared` leaves `tls` unset. Enabling `tls-ring` / `tls-webpki-roots` is required so HTTPS fails closed instead of opening plaintext HTTP/2; `tls_config(with_webpki_roots)` is required so the handshake actually happens. `http://` (local e2e geyser) stays plaintext. `{err:#}` prints the anyhow chain that the first preview collapsed to `error="subscribe"`. webpki-roots is the right store for the distroless image.

**Changed after review (`4d9c0af43`).** TLS is keyed off the parsed URI scheme (`https`), not `starts_with("https://")`. That is the URI tonic dials.

**Rejected.** Switching the constructor to `Endpoint::new`: it is `#[doc(hidden)]` with an upstream FIXME. Extracting a helper plus a constructor-success unit test: tonic's `tls` field is crate-private, so `is_ok()` would not distinguish the old prefix check from the scheme bind. `tls-aws-lc` instead of `tls-ring`: optional alignment with workspace rustls, not required for this bug.

**Still unverified.** CI only hits local plaintext geyser. Live ingest from `https://laserstream-devnet-ewr.helius-rpc.com` (a confirmed slot, not reconnect spam) is still the proof this transport patch works.

PR #4039, opened 2026-09-17
fix(relayer): keep host chain ids from env as strings

Eikix:
## Why

In the Solana preview environment (run 35234087525, namespace `fhevm-ci-eikix-g6pgrxh`) the relayer configures both host chains through `APP_HOST_CHAINS__<i>__*` env vars: the Anvil host at index 0 and the Solana host at index 1 with chain id `9223372036854788153` (RFC-021, above `i64::MAX`). After the Solana step the new relayer pod crash-looped with:

```
invalid type: floating point `9223372036854788000.0`, expected a u64 as a number or a decimal string for key `host_chains.1.chain_id`
```

`Settings::new` uses the config crate's `try_parsing(true)` so numeric fields inside `Vec` entries deserialize from env. That parser tries bool, then i64, then f64: an id above `i64::MAX` becomes a lossy float, which `deserialize_u64_from_str_or_num` correctly rejects. The docker-compose e2e stack does not hit this because it configures the Solana host chain in YAML, where the id is a quoted string.

## What

Partition the `APP_*` environment: `APP_HOST_CHAINS__<i>__CHAIN_ID` values are fed to a second `Environment` source without `try_parsing`, so they reach the existing string path of `deserialize_u64_from_str_or_num`. Everything else keeps the current behaviour.

Adds `test_host_chain_id_above_i64_max_from_env`, which sets the preview's six env vars and loads the test YAML through `Settings::new`.

## Verification

- Without the `Settings::new` change the new test fails with exactly the error above.
- `cargo test -p fhevm-relayer --lib config::settings`: 50 passed.
- Live check in the next preview dispatch; this PR body will be updated with the result.


Eikix (commented):
[AUTOMATED]

Independent review of a9cf335aa…52ff267a5 (relayer host chain id from env). Three passes: open-ended, maintainability, design/integration. Findings were checked against config 0.15.19 `try_parsing` and `Settings::new`.

**Accepted.** Kubernetes env is already an unquoted decimal string. YAML quoting cannot own this. `try_parsing` is bool → i64 → f64, so `9223372036854788153` becomes `9223372036854788000.0`; `deserialize_u64_from_str_or_num` has no `visit_f64`, which is the preview crash. Global `try_parsing(false)` is not the fix: tagged `KeyUrlConfig` u64s still need it. The new `Settings::new` test hits the production function with preview-shaped `APP_HOST_CHAINS__0/1__*` vars.

**Changed after review (`52ff267a5`).** Process env with `try_parsing(true)` is kept. Only `APP_HOST_CHAINS__*__CHAIN_ID` is overlaid as an unparsed string source. The snapshot of every `APP_*` key into two maps is gone. `cargo fmt --package fhevm-relayer` is clean. `cargo test -p fhevm-relayer --lib config::settings`: 50 passed, including the large-id env test and `test_settings_from_yaml_with_env_overrides`.

**Still unverified.** Live Helm upgrade in a preview namespace. Relayer CI `make fmt` on this head.

Eikix:
Not needed after [4019](https://github.com/zama-ai/fhevm/pull/4019)

PR #4040, opened 2026-09-17
docs(solana): record Solana preview findings from the first live runs

Eikix:
## Why

First live runs of the Solana preview environment (`preview-env-deploy.yml` with `deploy_solana=true`, Anvil, 1 coprocessor, 4 enclave KMS parties, kms `0ead5f7`) on 2026-09-17: runs 35234087525 (namespace `fhevm-ci-eikix-g6pgrxh`) and 35244420014 (`fhevm-ci-eikix-g6vm8j2`). This PR carries the documentation corrections and lists everything found, fixed or still open. Companion fixes already merged: #4027 (endpoint Dockerfile), #4034 (tx-sender gateway URL), #4037 (deploy sign attempts). Open drafts: #4038 (listener TLS), #4039 (relayer env chain id).

## What this PR changes

- `ci/preview-env/README.md`: `kms_repo_ref` must be a full commit SHA or a ref name; a short SHA makes `actions/checkout` fail.
- `solana/deploy/README.md`: run `host wipe` before destroying a preview (the destroy workflow does not); the SDK/demo/test-suite consumer graph is pinned to the localnet program ids, so the e2e scenarios and demo cannot target a preview; the transfer scenarios need `deploy_example_programs=true`.

## What worked (live, namespace g6vm8j2)

- Whole EVM stack, keygen + CRS, relayer with two host chains (`12345`, `9223372036854788153`).
- Solana host deploy after a wipe: `initialize_host_config` and `define_kms_context` on a fresh HostConfig; wipe → destroy → recreate flow is sound (`host wipe` closed 3/3 then 6/6 program-owned accounts).
- Host listener with TLS (#4038): continuous ingestion of sealed devnet blocks from Helius LaserStream, checkpoint advancing, Solana transactions decoded into computations and encrypted-state leaves.
- Through the SDK against devnet: SPL mint, confidential mint, confidential token accounts, wrap of 1000 (with the local program-id substitution described below).

## Findings

1. **Handle derivation mismatch on devnet (open, blocks every FHE flow).** The on-chain program and the listener derive different handles for the same events. Alice's initial balance handle on chain was `036e79…`, the listener derived `911fdf…`; supply `c45537…` vs `770f9e…`; post-wrap `3e0eff…` vs `799384…`. The wrap's operation graph therefore depends on handles the coprocessor never computed, tfhe-worker logs `some inputs are missing to execute the dependence chain` and nothing materializes. Chain id is identical (same handle suffix `…8000000000003039 05 00`), so the difference is in `previous_bank_hash` or `unix_timestamp`. A 25 s capture of the LaserStream sysvar stream (296 updates) is self-consistent (Clock inner slot equals tagged slot, SlotHashes newest entry equals slot-1), and no derivation code changed between the program build and the listener build. Needs a controlled reproduction with an offline recomputation of one handle from captured sysvar data. Until fixed, the connector ↔ coprocessor leaf-proof path and user decrypt cannot be exercised on the preview.
2. **AWS `zws-dev/preview-solana-host-keypairs` holds public keys.** All three properties are 44-character base58 public keys; the deployer expects keypair JSON arrays (the 1Password item's three attachments, all named `json`). Every dispatch fails at `solana-host-deploy` with `invalid Solana keypair JSON` until the AWS values are replaced. Worked around in-namespace both times.
3. **Consumer graph pinned to localnet program ids.** 34 generated files across `sdk/js-sdk`, `solana/demo-dapp` and `test-suite/fhevm` inline `6Atbv…` and `pS2gM…` (constants, PDA defaults, instruction account defaults). Nothing selects the `preview-env` profile. The smoke test only ran after substituting the ids locally (reverted, not committed).
4. **`deploy_example_programs=false` leaves `confidential_token` undeployed**, which every confidential-transfer scenario needs. Deployed by hand for the test: 882 KB, 256 s, 4.49 SOL of rent. The three demo programs together (1.9 MB) need roughly 9.6 SOL; the deployer had 5–6 SOL and the Helius devnet faucet gives 1 SOL per request before rate limiting.
5. **Destroy workflow does not wipe the Solana host**, so a recreate on the same program addresses finds the previous HostConfig. Documented here; a wipe step in `preview-env-destroy.yml` gated on the namespace having `solana-deployer` would remove the manual step.
6. **`solana program deploy` on devnet needs many re-sign attempts** (fixed in #4037). Each failed attempt strands a buffer holding the program rent (3.67 SOL); recover with `solana program close <buffer>`.
7. **Listener error logging hid the transport error** (`error="subscribe"`); #4038 logs the full chain.
8. **Relayer env parsing of ids above i64::MAX** (#4039).
9. **Coprocessor tx-sender was still configured with a ws:// gateway URL** (#4034); the same latent bug exists in the main preview values.
10. **Early CrashLoops on every dispatch**: coprocessor workers and connector gateway listeners start before the DB-migration Job finishes (`relation "drift_revert_signal" does not exist`) and restart 2–3 times. Harmless but noisy.
11. **Infra**: during run 35244420014 a node replacement moved `postgres-coprocessor-1-0` and `postgres-relayer-0` (Multi-Attach volume warnings); the coprocessor registration Job ran during the move and failed. The Solana step has no retry for that; resumed by hand.
12. **`deploy-preview.sh` hardcodes the Solana chain id** (`9223372036854788153`) in the relayer env patch; a single source (HostConfig or the profile JSON) would avoid drift.
13. **Kyverno policy warnings** on the one-off pods/Jobs created by hand (no resource limits in some ad-hoc manifests); the chart-managed workloads are fine.
14. **Operational**: while masking RPC URLs during debugging, the Helius gRPC `x-token` was printed once to a local terminal. Rotation recommended.

## Verification

Docs-only diff. The live evidence is in the session runbook (not committed).


Eikix (commented):
[AUTOMATED]

Independent review of a9cf335aa…703516c71 (docs-only preview findings). Three passes: open-ended, maintainability, design/integration. Findings were checked against destroy/deploy workflows, bootstrap rebind, listener crate features, and the 101 dispatch page.

**Docs that were wrong in the tree and are now aligned (`703516c71`).** Wipe belongs before the next `host deploy`. Destroy and deploy's namespace reset only helm-uninstall and delete the namespace; they do not close Solana accounts. The numbered steps no longer say destroy-then-wipe. A Job inside the namespace is not the only remaining path — AWS / 1Password keypairs survive teardown. The `kms_repo_ref` short-SHA trap is on the 101 dispatch example operators copy, not only the internals table. `deploy_example_programs=true` sits next to the launch flag (transfer scenarios need `confidential_token`). The SDK/generated-client inventory was removed from the deployer README; sibling #4041 owns the SDK chain field.

**Finding 1 owner (PR body, not implemented here).** The handle mismatch is compiled `crate::ID`, not sysvar entropy. On-chain preview is `DPq5y89…`. host-listener links `zama-host` without `preview-env`, so reconstruction hashes with `6Atbv…`. `--program-id` is the watch/PDA key only; `finish_computed_handle` then overwrites bytes 21–31, which is why the suffix can match while the hash does not. Recorded under Validation.

**Left as named follow-up, not this PR.** Wiring wipe into destroy/deploy (DD-051). Building the listener with `preview-env`. Replacing the AWS pubkey-vs-keypair-JSON item. Profile-aware demo/test-suite clients. The 14 live-run items belong in the PR body as a session log, not as a second operator procedure.

Issue #2055, opened 2026-03-06
feat(coprocessor): add direct log-trace correlation without JSON rewrite pass


Eikix:
## Outcome

Make coprocessor JSON logs directly correlate to their OTel traces in the observability stack.

In practice, logs emitted while an OTel span is active should carry the trace/span identifiers needed by Grafana/Loki/Tempo-style backends to jump from a log line to the exact trace.

## Constraints

- Do not re-parse and re-serialize every JSON log line after formatting.
- Keep the current log shape and tracing behavior materially unchanged for operators.
- Keep the implementation small and maintainable.

## Acceptance criteria

- JSON logs emitted inside an active OTel span include the trace/span identifiers expected by the backend.
- The implementation avoids the current parse-JSON -> mutate -> serialize-JSON-again pattern.
- The solution is implemented in the tracing/log formatting layer, not by manually threading trace IDs through call sites.
- Tests cover the active-span case and the no-active-span case.

## Notes

A good implementation likely uses a proper `tracing_subscriber` formatter/field-formatting approach instead of post-processing serialized JSON, but the issue owner should validate that against high-quality Rust codebases/crates that already solve this cleanly.

PR #4041, opened 2026-09-18
refactor(solana): name the host program in the SDK chain definition

Eikix:
Closes zama-ai/fhevm-internal#2055

## Why

`FhevmSolanaChain.fhevm.verifyingProgramId` was optional, and every SDK action that needed the zama-host program id fell back to `ZAMA_HOST_PROGRAM_ADDRESS` from the generated localnet client. The public-decrypt client, `createSolanaFheTransaction`, the permit-revocation helpers and the user-decryption-delegation builders therefore targeted the localnet deployment unless each caller overrode them. The first preview-env runs surfaced this (#4040, finding 3). The EVM SDK has no such default: the ACL address is a required part of each chain definition.

## What

- `FhevmSolanaChain`: `fhevm.verifyingProgramId?` becomes `fhevm.programs.host.address` (required, canonical bytes32), mirroring `fhevm.contracts.acl.address` on the EVM chain. `verifyingProgramId` remains the permit wire field and is derived from it.
- SDK actions drop the generated-constant fallbacks. `solanaHostProgram(chain)` is the one conversion to a base58 `Address`. `createSolanaFheTransaction`, `solanaPermitInvalidationAddress`, `buildRevokePermitsInstruction`, `fetchSolanaPermitInvalidation`, the delegation builders and `fetchSolanaUserDecryptionDelegation` take `programAddress` explicitly.
- Demo dapp: chain literals use the new field; the FHE transaction helper gets `config.programs.host`. Inside the vault module, which is compiled against the generated `confidential_token` client, the fixed host id that module already asserts is passed through.
- `test-suite/fhevm`: same renames; the public-decrypt runner now receives the host program (`PD_VERIFYING_PROGRAM_ID`, fed from the vertical config) instead of relying on the SDK default; provisioning helpers pass the generated id they already import.
- Demo config: `aclProgram` (bytes32) and `programs.host` (base58) name the same program in two spellings; `parseDemoConfig` now refuses a config where they differ, since the SDK chain and the instruction builders would otherwise target different programs.
- `solanaHostProgram(chain)` is exported from `@fhevm/sdk/solana`; the README example uses it.

Operational note: anything invoking the `solana-public-decrypt` profile directly must now set `PD_VERIFYING_PROGRAM_ID` (bytes32 hex of the host program). The only in-repo caller, the FHE vertical, sets it.

Not in this PR: making the demo vault module and the test-suite provisioning helpers profile-aware. They inline the localnet ids of `zama_host` and `confidential_token` through the generated clients; that is the follow-up named in the issue.

## Verification

- `sdk/js-sdk`: `tsc -b --noEmit`, eslint, prettier, vitest (`src/solana`, `src/core/chains`, `src/core/coprocessor`: 37 files, 506 tests) pass.
- `solana/demo-dapp`: `tsc -b`, oxlint, vitest (42 files, 250 tests) pass.
- Two independent review passes (subagents) enumerated every public export that builds an instruction, derives a PDA or reads a host-owned account and found no remaining defaulted program id. Their findings are folded in: demo config consistency, `solanaHostProgram` export and docs (817efe8f7); `buildVerifyPublicDecryptInstruction` requiring the program id (f8f908062); a fixture type error that broke `tsc -b` in CI (4a3e82db8); and a public-decrypt test that fails if the bundled id ever comes back (last commit).
- Not taken from review: a runtime guard in `defineFhevmSolanaChain` for untyped callers. The EVM `defineFhevmChain` has none either and the lint rules reject a guard on a typed value; the type carries the requirement.
- `test-suite/fhevm`: `tsc --noEmit`, `bun test src` (516 tests) and `bun test demo` (75 tests) pass.
- Not run: the e2e scenarios and the demo against a live stack. The edits there are the field rename plus passing an address already in scope.


Eikix (commented):
[AUTOMATED]

Review loop of `feature/solana`…`f8f908062` (three independent passes: open-ended, maintainability, design/integration). Validated against the tree. Not posted until this comment.

The chain change matches the EVM ACL pattern. `fhevm.programs.host.address` is required (canonical bytes32). Permit `verifyingProgramId` is derived from it. Named SDK actions no longer fall back to generated localnet `6AtbvED1…`. Nested `programs.host.address` is the right shape; do not flatten it, and do not store a base58 `Address` on the chain type.

**Findings that were real**

1. `buildVerifyPublicDecryptInstruction` still left `programAddress` optional. Omitting it forwarded `undefined` into Codama, which substitutes `ZAMA_HOST_PROGRAM_ADDRESS`. Same class of bug this PR closes for `createSolanaFheTransaction`, revoke, and delegation. The helper is a public `@fhevm/sdk/solana` export. Fixed in `f8f908062`: `programAddress` is required; the test passes a non-localnet id.

2. `solanaHostProgram` was the documented conversion but was not exported from `@fhevm/sdk/solana`. Callers re-decoded bytes32 by hand. Exported in `817efe8f7`. `solana/README.md` now shows `createSolanaFheTransaction({ payer, programAddress: solanaHostProgram(chain) })`.

3. Demo boot JSON has two spellings of one deployment (`aclProgram` bytes32 vs `programs.host` base58). They can disagree. `817efe8f7` refuses a config where they name different programs.

4. Public-decrypt fixture pinned localnet as a pasted hex literal beside an imported `ZAMA_HOST_PROGRAM_ADDRESS`. `f8f908062` derives the chain bytes32 from that Address.

**Left out of scope (named in #2055)**

Demo vault module and test-suite provisioning still compile against the generated `confidential_token` client and pass its localnet ids as explicit arguments. That is not an SDK default coming back. Generated Codama `?? ZAMA_HOST_PROGRAM_ADDRESS` under `internal/generated` was left alone.

**Verification**

Ran: vitest on the three touched SDK files after `f8f908062` (62 tests). Not run in this loop: full SDK `tsc`/eslint, e2e, live demo stack. Preview “two Zamas, same `chain_id`” is not exercised here.

Eikix (commented):
[AUTOMATED]

Independent review of a9cf335aa…778415f95 (`fhevm.programs.host.address` required on the Solana chain). Three passes: open-ended, maintainability, design/integration. Head moved after the first loop on this PR; this is the current range, including `778415f95`.

**Accepted.** The chain field is the EVM ACL analogue: required bytes32, kit `Address` only at helper boundaries, `solanaHostProgram` the conversion. Generated Codama `?? ZAMA_HOST_PROGRAM_ADDRESS` stays under `internal/generated`. Vault/provisioning still compiling against generated `confidential_token` is the named non-goal. Nested `programs.host` is the right shape; do not flatten.

**Changed after review (`778415f95`).** `runSolanaPublicDecrypt` types `verifyingProgramId` as `Bytes32Hex` and drops the cast. Delegated-user-decrypt takes `solanaHostProgram` from the chain it already builds. The demo parser mismatch check has a unit test.

**Rejected / left.** Collapsing demo `aclProgram` off the wire format is a schema change, not this PR. The test-suite demo parser still round-trips a synthetic mismatch; seed on localnet writes matching values, and making that parser profile-aware is the same leftover as vault/provisioning. `fhe-vertical.ts` EncryptedStore reads still pass the generated localnet host id while decrypt uses `verifyingProgramId` — same leftover class, not a silent SDK-action fallback.

**Still unverified.** demo-dapp vitest in this worktree (needs a built SDK). Live preview / two Zamas on one cluster. Generated `@fhevm/sdk/solana/host` omit-still-defaults.

PR #4043, opened 2026-09-18
fix(solana): derive handles from the followed program id, not the compiled one

Eikix:
## Problem

Every computed handle (`computed_*_handle` in `zama-host/src/state/mod.rs`) hashes the host program id into its keccak preimage, and that id was read from `crate::ID`. The preview-env deployer builds zama-host with `--features preview-env`, which swaps `declare_id!` to `DPq5y89RDZPq9NcMh9X1NgjBWgYmSXg3QoipSBV3ZMzQ`, while the host-listener image builds the same crate with the default id `6AtbvED1rfX68aCT1tYgU1aeu4kFksPDxZG9gtB1Fgtu`. The listener therefore reconstructed handles the program never emitted, and nothing linked the two: `config.program_id` only filtered instructions.

Confirmed empirically before the change: with an identical context, the default build derives `1b6abdbb…30390500` and the preview-env build `410304c5…30390500`.

## Change

- `HandleDerivationContext` gains a `program_id: Pubkey` field. The program passes `crate::ID`; an off-chain reconstruction passes the id of the deployment it follows. `computed_rand_handle` / `computed_rand_bounded_handle` take the program id the same way.
- `reconstruct_context` in the host-listener uses the configured `program_id` and fails loudly (instead of silently skipping) when it is not a valid pubkey.
- Listener tests now run against a program id that is not the compiled-in one, so a regression to `crate::ID` would fail them. New unit test `handle_derivation_binds_the_program_id`.

PDA derivations keep using `crate::ID`; they are the program's own accounts.

Not covered here: `solana/crates/zama-fhe/src/execution.rs:390` also uses `zama_host::ID`. Worth a look with the same lens.

Related: #1985.

## Verification

- `cargo test -p zama-host` (default and `--features preview-env`): 110 passed.
- `cargo test -p host-listener --features solana-grpc,solana-reconstruct --lib solana`: 62 passed.
- `cargo check --tests -p tfhe-worker`, clippy `-D warnings` and `cargo fmt --check` in both workspaces.
- `zama-solana-runtime-tests`: transient/operator suites pass; `batcher_mollusk` fails locally only because `confidential_batcher.so` is not built here (unrelated to this change).
- Not run: an end-to-end preview-env deploy with the listener following the preview program id.


Eikix (commented):
[AUTOMATED]

Independent review of 66df8f70b…ecdb38431 (handles hash the followed program id). Three passes: open-ended, maintainability, design/integration. Range reviewed at `12ca821e9`; `a48178f61` (reconstruction tests on a non-compiled id) and `ecdb38431` landed after.

**Accepted.** Computed handles already bind program id (DD-043 / DD-051). Preview-env compiles zama-host as `DPq5y89…` while the listener crate keeps `6AtbvED1…`; `config.program_id` only filtered instructions. `HandleDerivationContext.program_id` is the smallest split: on-chain still passes `crate::ID`, reconstruction passes the deployment it follows. PDA helpers staying on `crate::ID` is correct. `zama-fhe` `select_returned_handle` vs `zama_host::ID` is CPI return-data origin, not this bug. Helm already passes `--program-id=$(SOLANA_HOST_PROGRAM_ID)`.

**Rejected.** Folding `computed_rand_handle` into the full slot context: rand handles do not bind `previous_bank_hash` / timestamp (DD-043). A shared preimage-suffix helper across the seven eval functions: every constructor already requires `program_id` at compile time; that is not this PR's choke point. Storing a `Pubkey` on `SolanaGrpcListenerConfig` and deleting the `Result<Option>` parse: Yellowstone instruction ids are already base58 strings; the extra parse is fail-closed for non-CLI builders. Open-ended did not rank it.

**Changed after review (`ecdb38431`).** Listener module docs, INVARIANTS #28/#33, and `SolanaGrpcListenerConfig.program_id` now say reconstruction hashes the followed `--program-id`, not `declare_id!`. `handle_derivation_binds_the_program_id` covers trivial encrypt and both rand regimes. `cargo test -p zama-host handle_derivation_binds_the_program_id` passed.

**Already on the branch (`a48178f61`).** Listener and reconstruct tests follow `7DYCAhqw…`, not `zama_host::ID`. `ReconstructContext` rustdoc already names `--program-id`.

**Still unverified.** End-to-end preview-env ingest with the listener following `DPq5y89…`.

PR #4046, opened 2026-09-18
chore(kms-connector): proxy and endpoint helm charts

eudelins-zama:
Closes https://github.com/zama-ai/fhevm-internal/issues/2046

Tested in the follow-up preview-env [PR](https://github.com/zama-ai/fhevm/pull/4062).

### What

Adds the two new services to the `kms-connector` Helm chart: the **endpoint** and the **proxy**.

Each gets a Deployment, a ClusterIP Service and an optional ServiceMonitor, with the same `env`, `tracing`, `probes`, `resources` and scheduling knobs as the existing components.

### Host chains refactor

The per-chain `commonConfig` fields (`ethereumUrl`, `polygonUrl`, `bnbTestnetUrl` and their chain id / ACL overrides) are replaced by a single `commonConfig.hostChains` list shared by every component. Each entry names a preset (`ethereum`, `polygon`, `bnb_testnet`) and can override `chainId` and `aclAddress`. The kms-worker and endpoint consume the whole list, while gw-listener and tx-sender pick the entry named `ethereum` through a new helper. This is a breaking change for existing values files, hence the chart bump to `2.0.0`.

### Secrets and exposure

The proxy expects two secrets that the chart does not create: a `kubernetes.io/tls` secret referenced by `kmsConnectorProxy.tls.secretName`, and the SHA-256 digest of the relayer API key, read from `kmsConnectorProxy.apiKeyDigest.secret` or set inline with `apiKeyDigest.value`. Exposing the proxy outside the cluster is left to the deployment..

### Misc

Monitoring listen addresses are now derived from `ports` so they stay consistent with the Service and probes.


charts/kms-connector/templates/kms-connector-proxy-deployment.yaml
```diff
@@ -0,0 +1,119 @@
...
+                  fieldPath: metadata.name
+          {{- end }}
+          ports:
+            {{- range $portName, $portValue := .Values.kmsConnectorProxy.ports }}
+            - name: {{ $portName }}
+              containerPort: {{ $portValue }}
+              protocol: TCP
+            {{- end }}
+          resources:
+            requests:
```
maxnovawind: We can use `with` pattern  : `with .Values.kmsConnectorProxy.resources`

charts/kms-connector/values.yaml
```diff
@@ -32,32 +34,24 @@ commonConfig:
     decryption: ""
     gatewayConfig: ""
 
-  # Ethereum chain RPC node endpoint (HTTP)
-  ethereumUrl: "http://ethereum-node:8545"
-  # Ethereum chain identifier (override the network preset when set)
-  ethereumChainId: ""
-  # Ethereum smart contract addresses (override the network preset when set)
+  # Host chains, shared by every component (ACL checks in kms-worker, accepted
```
maxnovawind: I think we need to ensure that we support retrocompatibility configuration by including a deprecated field to prevent any misconfiguration on the partner side. WDYT ?

charts/kms-connector/values.yaml
```diff
@@ -362,6 +347,153 @@ kmsConnectorTxSender:
...
+    limits:
+      cpu: 1000m
+      memory: 512Mi
+
+  # Health check probes
+  probes:
+    liveness:
+      enabled: true
+      httpGet:
+        path: /liveness
```
maxnovawind, line 484: Are the liveness/readiness paths correct?

charts/kms-connector/templates/kms-connector-proxy-deployment.yaml
```diff
@@ -0,0 +1,119 @@
...
+              value: {{ printf "0.0.0.0:%v" .Values.kmsConnectorProxy.ports.monitoring | quote }}
+            - name: KMS_CONNECTOR_PROXY_BIND_ADDRESS
+              value: {{ printf "0.0.0.0:%v" .Values.kmsConnectorProxy.ports.https | quote }}
+          {{- with .Values.commonConfig.env }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- with .Values.kmsConnectorProxy.env }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+            - name: KMS_CONNECTOR_TLS_CONFIG__CERT_PATH
```
maxnovawind, line 74: Can we support both HTTP and HTTPS protocols? https should be optional normally

charts/kms-connector/values.yaml
eudelins-zama: Yes they are. It's other ("old") services liveness path that are wrong IMO.
We should be using `/liveness` for liveness, and `/healthz` for readiness.

I'll fix for old services.

charts/kms-connector/values.yaml
eudelins-zama: I don't think this is a good idea tbh, but you tell me as you're more experienced on partners' setup.

My point is that we're going to need to update partners' config anyway to support these new services. At that moment, better to make sure their config is well up-to-date, instead of complexifying the charts (already very unreadable IMO) themselves just to support an outdated config format...

What Claude suggest as middle ground would be to have a "hard-fail guard" failing the templating explicitly if the partners still use a deprecated field. I think this would complexify the charts without much benefit, because we would already clearly see that the config is wrong with crash loop errors or whatever I think.

Wdyt :slightly_smiling_face: ?

charts/kms-connector/templates/kms-connector-proxy-deployment.yaml
eudelins-zama: Nope we currently don't. I'm not sure to understand why we would like to support plain HTTP tbh :thinking: 

Anyway, if we go this way, we would need a follow-up PR to update both the code and the charts, I suggest we keep it out of scope of this PR and discuss the HTTP support internally first :slightly_smiling_face:

charts/kms-connector/templates/kms-connector-proxy-deployment.yaml
eudelins-zama: Fixed af96a3d59a8d251e76f305ad83059f4d3c8d2ef4

Applied to all connector services for consistency

charts/kms-connector/values.yaml
eudelins-zama: Fixed 71ebafd1f3bcfe796daa08f4f8061276f111a82a

eudelins-zama: EDIT: re-read definition of k8s liveness, and now think that `/healthz` is the right choice for both liveness and readiness probes.
It seems our current `/liveness` is a bit useless, but out of scope for this PR...
Reverting

eudelins-zama: Amend done 3e4cb773d4362b9971ed99a0910970a0ea3763ca

Re-used the values from devnet config find in gitops files fyi

charts/kms-connector/values.yaml
```diff
@@ -32,32 +34,24 @@ commonConfig:
...
-  ethereumUrl: "http://ethereum-node:8545"
-  # Ethereum chain identifier (override the network preset when set)
-  ethereumChainId: ""
-  # Ethereum smart contract addresses (override the network preset when set)
+  # Host chains, shared by every component (ACL checks in kms-worker, accepted
+  # handles in endpoint). `name` selects the preset entry (ethereum, polygon,
+  # bnb_testnet); chainId and aclAddress override the preset when set. An
+  # "ethereum" entry is required by gw-listener and tx-sender.
+  hostChains:
+    - name: ethereum
```
piizama: Following offline discussion, we agreed on those changes:

* change hostChain list to mapping:
```
hostChains:
  ethereum:
    url: **MUST BE SET BY OPERATOR**
    aclAddress
  polygon:
  [...]
```
* Make sure the  hostChains.ethereum.url can point to a secret variable (eg. `$(SECRET_URL)`)
* Use presets to validate that URL are set by the chart users for networks which are deployed

piizama (changes requested):

charts/kms-connector/values.yaml
piizama: Following our offline discussion, we accept the chart breaking change for this time.

charts/kms-connector/templates/kms-connector-proxy-deployment.yaml
eudelins-zama: Discussed internally. Out of scope for now

charts/kms-connector/values.yaml
eudelins-zama: Done in a377083ba1562a2d7e16fc83edbfdf81e433e606

charts/kms-connector/values.yaml
piizama: @eudelins-zama please set the liveness check to `/liveness` and not `/healthz`, it's very dangerous to use the wrong endpoint for this check as it can cause cascading failures. See this Claude explanation ⬇️:

>  ### why `livenessProbe` should not point at `/healthz`                                                                                                                                                                                                                                                                                                                                                                                                                     
>                                                                                                                                                                                                                                          
>   `/healthz` aggregates external dependencies: Postgres, Gateway RPC, Ethereum RPC, KMS Core gRPC shards (see `crates/tx-sender/src/monitoring/health.rs`, `crates/kms-worker/src/monitoring/health.rs`). None of those is fixed by restarting the connector.                                                                                                                                                                                                              
>                                                                                                                                                                                                                                          
>   **Risk:** a shared dependency blip (RPC node, Postgres) makes `/healthz` return 503 on every replica at once. With `periodSeconds: 10` and the default `failureThreshold: 3`, ~30s of upstream downtime gets the whole fleet SIGKILLed, then CrashLoopBackOff. Restarts add reconnect load on the struggling dependency, drop in-flight work, and wipe the logs you need to diagnose it. The probe amplifies the outage instead of containing it.                              
>

piizama: A basic /liveness endpoint is not useless, it will be able to restart the pod in case the process gets stuck. However, our gitops also needs to be updated to correct this.

eudelins-zama: Fair. Fixed e5ed86aca1ac18034e3c312a0ab49a7fb50e516d :slightly_smiling_face:

maxnovawind (approved):

piizama (approved):

dvdplm (approved):

nboisde (approved):

PR #4052, opened 2026-09-18
refactor(solana): read program ids from per-environment config files

Eikix:
## Why

Each Solana program bakes its id in with `declare_id!`, and the preview program uses different ids than localnet. Until now that was a `preview-env` cargo feature in four crates, mirrored by hand in `Anchor.toml`, the deployer's `program-profile.ts` and `profiles/preview-env/program-ids.json`. Any crate linking `zama-host` inherited whichever feature was on (the cause of #4043), and every future Zama (zama-devnet, zama-testnet, mainnet) would add `#[cfg]` lines to four crates.

## Change

Program ids and build features become per-environment config: `solana/environments/<name>.json`, selected at build time by `PROGRAM_ENVIRONMENT`.

- `crates/program-environment`: a 20-line build-script helper. Each program's `build.rs` calls `declare_program_id("<program>")`; it reads the environment file and writes the `declare_id!` line to `OUT_DIR`, which `lib.rs` includes.
- `solana/.cargo/config.toml` pins `PROGRAM_ENVIRONMENT=localnet` with `force = true`, so a stray shell export cannot change what a build under `solana/` compiles. `scripts/build-programs.sh` is the only thing that overrides it, via `--config` forwarded through anchor's second `--`, and a non-localnet build prints a cargo warning. The pin does not reach builds of `zama-host` from another workspace (coprocessor images); no off-chain code reads the compiled id, so those images' always-localnet id is not load-bearing.
- `preview-env` feature removed from the four programs. `admin-sweep` stays: it is a capability, and `preview-env.json` lists it under `features.zama_host`.
- Deployer: `program-profile.ts` becomes `environment.ts` and reads the same files; `SOLANA_PROGRAM_PROFILE` becomes `SOLANA_ENVIRONMENT`, `PROGRAM_PROFILE` becomes `PROGRAM_ENVIRONMENT` (Dockerfile, compose, CI matrix, preview values). Preview keypairs fetched by `fetch-preview-env-keypairs.sh` land in `solana/deploy/keypairs/preview-env/` (gitignored).
- Deployer: `deploy --allow-upgrade` deploys absent programs and upgrades ones whose bytecode differs when the deployer holds the upgrade authority. Plain `deploy` keeps refusing differing bytecode. The first preview run from this branch stopped on exactly that refusal, for the host and again for the demos (confidential-token existed with older bytecode while demo-vault did not exist, a state neither `deploy` nor `upgrade` handled).
- Preview wipe wired: the solana-host Job runs `host wipe` before `host deploy --allow-upgrade` (DD-051 had it as follow-up). Destroy-side wipe stays follow-up.
- `Anchor.toml` loses the reference-only `[programs.devnet]` table; `[programs.localnet]` stays because Anchor reads it.
- Coprocessor Dockerfiles that copy `solana/programs` now also copy `solana/environments`; the three coprocessor CI workflows watch that path.
- Docs: DD-053 (this decision and the rejected alternatives, index row), DD-051 wipe wording, open decision on off-chain leaf-history growth (fhevm-internal#2060), `solana/deploy/README.md`, `solana/environments/README.md`.

Adding an environment is a new JSON file, its entry in `SOLANA_ENVIRONMENTS` (static import, one-file bundle) and deployer keypairs. No Rust change. A wrong id in the file flows consistently into the `.so` and the deployer and stops at deploy time, where the deployer refuses a program keypair whose pubkey differs from `declare_id!`.

## Verification

- `bash scripts/build-programs.sh preview-env zama_host confidential_token demo_vault confidential_batcher`: each generated `program_id.rs` carries the preview id; the localnet rebuild afterwards restores the localnet ids.
- Stray export is inert: `PROGRAM_ENVIRONMENT=preview-env cargo test -p zama-host` under `solana/` compiles the localnet id (the pin wins). Only the two `--config` flags select preview, and that build prints the warning.
- host-listener `cargo check` from the coprocessor workspace (build script resolves the file from another workspace).
- Deployer `tsc`; test-suite `deploy-programs.test.ts` (including `--allow-upgrade` deploy, upgrade and wrong-authority cases), `environment.test.ts`, `demo/lifecycle.test.ts` (68 pass); `ci/preview-env/solana-host/test_charts.py` (9 OK); dead-surface sweep clean; workspace clippy and fmt.
- Live: preview-env-deploy run 35337871926 from dffbc5cf7 built all images and deployed the stack. The solana-host Job failed on the `deploy` refusal; the same release re-run by hand with `host upgrade` uploaded the bytecode and initialized HostConfig and the KMS context on devnet. Gateway registered chain id `72057594037940281`, the listener ingested sealed blocks over Yellowstone and its checkpoint advanced, connectors and relayer took the Solana host chain. The demos Job hit the same refusal. A redispatch from this head validates the wipe and `--allow-upgrade` path end to end.


Eikix:
Reviewed at `afec13b2d` (base `feature/solana`). Three independent passes: open-ended, maintainability, design/integration. Findings checked against the tree and the failed `solana-tests/deployer-image (preview-env)` log.

The environments-file approach is the right one. `admin-sweep` stays a capability, not an environment identity.

## Accepted

1. **Preview build passed `admin-sweep` to every program.** CI: `confidential-token` does not declare that feature. Fixed in `97aaad3` (`features.zama_host` only).
2. **Laptop keypair fetch aborted after a successful write.** `set -u` still expanded `${PROFILE_DIR}`. Fixed in `afec13b`.
3. **DD-053 described handle reconstruction as if #4043 had landed.** The listener `--program-id` filters instructions and looks up HostConfig; keccak still hashes the compiled `zama-host` id. Wording updated in `afec13b`.
4. **`cluster` and `deployer` in the JSON had no readers.** Removed so the files match “program ids and features only.”
5. **Tests did not lock the per-program feature map**, and `lifecycle.test.ts` still parsed `declare_id!("`. Both updated. `bun test` on those cases: 4 pass.

## Rejected

Delete `crates/program-environment` and have each `build.rs` take an env var. That would drop the ten coprocessor `COPY solana/environments` lines. Leaner, but a second design: the JSON files are the source of truth this PR set out to add.

## Unverified

Preview Docker rebuild after `97aaad3`, a live `deploy_solana` run, and #4043 (handle `program_id`) still open. Localnet e2e was not rerun.

Eikix (commented):
Review of dffbc5cf7 against `feature/solana` (a555960ec, which already contains #4005). The mechanism is sound and the preview flow works end to end; the findings are one verification gap, one doc claim that is too strong, and hygiene left over from the rename. Nothing blocks.

## Verified on this head

- `bash scripts/build-programs.sh preview-env zama_host` produces a `zama_host.so` that contains the preview id bytes and not the localnet ones, and `close_owned_accounts` is compiled in (the `admin-sweep` feature came from `preview-env.json`).
- The command-line `--config env.PROGRAM_ENVIRONMENT...` override beats the `force = true` pin in `.cargo/config.toml`, for the build script and for the test binary. Alternating localnet and preview builds in one target dir re-run `build.rs` each time (`rerun-if-env-changed` honours `[env]`).
- `cargo-build-sbf` consumes `--features` itself and forwards everything after `--` to `cargo`, so the double `--` in `build-programs.sh` is right, including when the environment lists no features.
- Every Docker context that compiles `zama-host` copies `solana/environments`; `.dockerignore` does not exclude it. Off-chain crates no longer read the compiled `zama_host::ID` outside a test comment, so the coprocessor images' always-localnet id is not load-bearing.
- `PROGRAM_PROFILE`, `SOLANA_PROGRAM_PROFILE`, `program-profile`, `profiles/preview-env` and the `preview-env` cargo feature are gone repo-wide.

## Findings

1. **The `cargo test` verification in the description is vacuous since the pin commit.** With `PROGRAM_ENVIRONMENT=preview-env cargo test -p zama-host` run from `solana/`, `[env] force = true` replaces the export for both the build script and the test binary: the build printed no "compiling the preview-env program id" warning and the test compared localnet with localnet. The same applies to `PROGRAM_ENVIRONMENT=nope`. The test only exercises preview when run with the two `--config` flags (that run passes and prints the warning). Suggest correcting the Verification section, and either renaming `id_is_the_selected_environments_zama_host` to what it proves (the `programs.<key>` lookup in `build.rs`) or dropping it: `check-zama-host-idl.sh` already byte-compares the committed IDL, whose `address` is `declare_id!`, against a fresh build. The `serde_json` dev-dependency on the program crate goes with it.

2. **DD-053 overstates the pin.** "a shell export cannot change what a build compiles" (DESIGN_DECISIONS.md:1975) holds only for cargo invoked under `solana/`. Cargo discovers config from the invocation's cwd, so `solana/.cargo/config.toml` does not apply when `zama-host` is built as a path dependency from `coprocessor/fhevm-engine`; an exported `PROGRAM_ENVIRONMENT` changes the id compiled into those images. Low impact today (see above), but the sentence should say "in this workspace", as `environments/README.md` and the config comment already do. The `program-environment` lib.rs doc comment ("Reads `<PROGRAM_ENVIRONMENT>.json`, `localnet` when unset") needs the same caveat, since it describes the one path a reader will try first.

3. **DD-053 has no row in the DD index table**; every other entry has one.

4. **"Adding an environment is one JSON file plus deployer keypairs"** also needs an entry in `SOLANA_ENVIRONMENTS` and a static import in `environment.ts` (the bundle is one file, so static imports are the right call). `environment.test.ts` iterates that tuple, not the directory. Documentation nit in the PR body and DD-053.

5. **`crates/program-environment/Cargo.toml:5-6` gives a reason that does not hold.** The coprocessor Docker contexts do copy `solana/Cargo.toml`, and `zama-host` itself uses `authors.workspace = true` in those same images. Either inherit like the sibling crates or state the real reason. `[lib] name = "program_environment"` is also the default and can go.

6. **Coprocessor CI path filters miss the new build input.** `coprocessor-docker-build.yml:171-173`, `coprocessor-cargo-clippy.yml:37-39` and `coprocessor-cargo-tests.yml:41-45` list `solana/programs/zama-host/**` and crates but not `solana/environments/**`. Low impact for the same reason as 2.

7. **Stale wording the rename missed**: `deploy/src/wipe.ts:2,32` ("preview-env profile", "a wrong profile"), `scripts/check-zama-host-idl.sh:19`, `runtime-tests/tests/host_admin_mollusk.rs:440`, `programs/zama-host/src/state/mod.rs:561` ("per cargo profile"), `docs/DESIGN_DECISIONS.md:1887`, and `deploy/README.md:150` ("the listener image needs no `preview-env` feature", which no longer exists).

8. **Preview ids lost their static cross-check.** Deleting `[programs.devnet]` and `program-profile.test.ts` leaves `preview-env.json` as the single source, checked only for "distinct from localnet". A typo now flows consistently into the `.so` and the deployer and surfaces at deploy time through `deploy-programs.ts:79-84`, which refuses a keypair pubkey that does not match `declare_id!`. That is the right guard and worth one sentence in the description.

Smaller, optional: `environment.test.ts:71-74` repeats the exact `toEqual` on line 69; the bring-up KMS context assertion sits under `describe('solana environments')` unrelated to it; the `.cargo/config.toml` assertion is whitespace-exact. `lifecycle.test.ts:135-140` parses `localnet.json` inside the loop. `environments/README.md` could say the file covers 4 of the 7 programs (`encrypted-counter`, `dep-chain`, `delegator-vault` keep literal ids). `build-programs.sh:17` could derive its program allowlist from the file's `programs` keys.

## Not run

A preview-env deploy. The deployer image and the e2e paths were checked statically only.

PR #4061, opened 2026-09-18
refactor(solana): one program id per program on every cluster

Eikix:
## Problem and behavior

Localnet and preview used different Solana program IDs, so generated clients could target the wrong deployment. This PR gives the four host/demo programs one identity across localnet and preview. Localnet loads upgradeable programs at genesis; preview verifies or upgrades those same identities in place.

The preview lifecycle now retains private recovery keys before funding, recovers available SOL and account rent before reset/destroy, and supports wiping/rebootstrapping Solana state without rebuilding the warm stack.

## Changes

- Remove the four committed localnet program keypairs and the separate localnet ID configuration. Keep the two explicitly scoped E2E specimen keypairs. Regenerate IDLs, clients and cross-language PDA fixtures for the shared IDs.
- Preserve explicit host upgrade support and authority/bytecode checks. The automated host-upgrade scenario remains localnet-only.
- Add preview-only cleanup instructions guarded by upgrade authority, persisted recovery inventory/transaction receipts, upload-buffer recovery, and public fee/refund accounting. Preserve executable/programdata accounts for reuse. Deactivate eligible lookup tables before waiting, allowing their cooldowns to overlap.
- Hold preview ownership through reset/rebootstrap. Add the reset-only workflow option, preserve existing image pins, retain locks while remote jobs are active, and permit interrupted-reset retries.
- Save test/demo wallet keys privately and mirror them before funding. Restrict ordinary test cleanup to that run's wallets. Devnet faucets require recoverable wallets; external wallets self-fund. Unfinished application state remains for coordinated reset rather than being destroyed by another test run.
- Run one demo operator behind an authenticated same-origin proxy. Remove permissive relayer CORS; preserve wallet/position state across reloads. Explorer links and RPC display/copy no longer expose RPC credentials.
- Integrate the shared confidential-token client and transient-store API from #4068/#4071, including the added cancellation/refund scenario.

## Dependency and merge order

#4068 and #4071 are merged. This PR is rebased onto `feature/solana` at `8a62174d0ea3cd261d78679a209e7e299d4fb3c8`. The rebased application head `d23f2c8e6` passed CI; the final workflow-only correction uses the PR base branch for automatic teardown checkout.

The distinct-ID question for persistent Zama deployments remains tracked in zama-ai/fhevm-internal#2055.

## Validation

On the prepared/rebased code:

- Test-suite `bun test src demo`: **610 passed**.
- SDK Solana suite: **453 passed**; SDK ESLint and TypeScript checks passed.
- Demo suite: **247 passed**, including Explorer credential checks and localnet/devnet wallet-funding policy coverage. Demo typecheck and lint passed.
- Preview orchestration Python suite: **16 passed**.
- Rebuilt Solana programs and IDL/schema/golden checks passed. Pinned runtime suite: **705 passed, 1 ignored**.
- Final teardown workflow passes actionlint; preview orchestration tests pass.
- Independent correctness and maintainability review of the combined changes completed; the external-wallet deposit caller found in review was fixed and rechecked. The Explorer credential fix received a separate bounded review.

Live evidence from the preceding preview code, not a deployment of this rebased head:

- Confidential transfer, token wrap/burn/redeem, deposit/settle/claim, cancellation/refund and browser lifecycle passed on the four existing devnet programs.
- [In-place reset](https://github.com/zama-ai/fhevm/actions/runs/35593361428) succeeded; a fresh confidential transfer passed afterward.
- A fresh browser deposit on the warm preview completed shield through claim in about 25 seconds and user-decrypted to exactly 1 cShares: [settlement](https://explorer.solana.com/tx/4bsZ1SfRbfE56VYWcXTpfxNZQ9hEHgc5qb4LfUbu2JLUVtA5n6wwYE5LQ8VzUZxVNvrepBwNK8n2FH4csqCQcmgu?cluster=devnet), [claim](https://explorer.solana.com/tx/4pcHZLr7Wi2edLEAiBrQGoW4fEHPckSqpcvNEZJN2NDDbTvvuL1PHd33A3rUDN7U7j8sEkU4txvQ4bbFV65hLstz?cluster=devnet). Both independently verified finalized and successful on public devnet RPC.

## Limits

The live experiment uses Test FHE parameters and mock USDC. Counter/dependency-chain scenarios were explicitly excluded from the four-program scope. Restart checks do not establish interruption recovery during an in-flight settlement. The newest rebase/recovery fixes still require CI and deployment validation; no claim of current-head live acceptance is made.

Reset accounting distinguishes reusable program rent, non-closeable classic SPL mint rent, and eight legacy token accounts with unavailable signing keys. The experiment added no new missing-signer accounts. Normal transaction fees are spent; cumulative recovered SOL includes recycled test funding and is not profit.


Eikix (commented):
## fhevm-review-loop

Three independent passes (open-ended, maintainability, design/integration) over `2cc812e435b7110275230798a0fbca8bed97b526...cd860cd0e6741ee900767c4277f096739e603690`. Findings below were checked against that tree. No code changes from this review.

The one-id-per-program move, genesis `--upgradeable-program`, deployer ignoring `target/deploy/<program>-keypair.json`, operator split, and on-chain `reclaim_batch_authority` / `close_join_record` guards are sound. No blockers.

### Accepted

**1. Well-known demo keys are funded and used on public Solana devnet.** `solana/scripts/demo/demo-keypairs/README.md` still says those keypairs are never funded on a public cluster. `demo/seed.ts` then `fundSol`s keeper / mint-authority / alice / bob from `DEMO_KEYPAIRS`, and `operator-server.ts` signs live keeper and mint-authority txs from the same files when `SOLANA_E2E_SOURCE=devnet`. Those secrets are in git. Anyone can impersonate the keeper and take the SOL this path puts on those accounts. Either stop seeding committed keys on a public cluster (per-namespace keys, same pattern as program keypairs), or treat them as public demo actors and drop the README claim.

**2. Demo faucet still drips localnet amounts, with no cap.** `planDemoFunding` tops a new burner to 5 SOL / 1000 USDC. `parsePositiveNumber` only checks `> 0`. Harness funding on devnet is 0.5 / 0.2 SOL. The page now mints a burner per browser, so each `localStorage` identity can pull 5 SOL from the deployer. Cap `sol` / `amount` (and make `planDemoFunding` network-aware).

**3. Host-upgrade scenario is not gated to the local validator.** `e2e/scenarios/fhe-vertical.scenario.test.ts` rebuilds `zama_host` and calls `deployHostProgram({ upgrade: true })` against `env.rpcUrl` with no localnet check. A full scenario run with `preview.env` sourced upgrades the shared preview host in place. Skip when `env.network !== "localnet"`.

**4. Operator encryption-key fetch always sends `x-api-key: local`.** `encryptionKeyMaterial.ts` hardcodes it. Deposit-arc, two-holder transfer, and SDK encrypt use `ZAMA_FHEVM_API_KEY ?? "local"`. If the preview relayer is not the local key, `GET /demo-encryption-key` 503s and the page cannot encrypt.

**5. After one program id, `chain_id` is the remaining cluster discriminator and is still the localnet sentinel on preview.** DD-052 says preview on Solana devnet uses the `solana-devnet` row (`0x01ce59db5080fc2c`). Deploy, Helm, harness, and seed all pin `72057594037940281` (`0x0100000000003039`). The stack is internally consistent, so this is not a silent decrypt bug today. Before this change, different program ids still separated localnet from preview. They no longer do. Either record the PoC sentinel as the preview row in DD-052, or switch preview to the published solana-devnet integer (HostConfig wipe + Gateway/coprocessor re-registration).

**6. Docs still describe the two-id world.** `.github/workflows/solana-e2e.yml` says demo programs deploy from `demo-keypairs/`. `EVM_PARITY.md` says the keeper cranks `close_join_record` with the settle lookup tables; that instruction is user-signed, the operator cranks reclaim only, and the dapp never closes a join record (deposit-arc smoke does).

**7. `close_join_record` is missing from the interactive path.** Mollusk and deposit-arc exercise it. After a successful claim the burner session should send `buildCloseJoinRecordInstruction`. Otherwise preview users keep paying join-record rent.

**8. Demo funding confirmation is a 10s poll.** `ensureDemoFunding` waits 40 × 250ms. Enough on the test validator, tight on public RPC. Confirm the signature `fundSol` returns, or use a network-dependent deadline.

### Rejected

- **Collapse the one-row `SOLANA_ENVIRONMENTS` table.** DD-053 keeps a file per Zama and the Rust `PROGRAM_ENVIRONMENT` pin. A one-element TypeScript mirror is that extension point with `localnet` deleted, not a second id set.
- **Merge the two demo-config parsers.** The seed artifact is roots plus endpoints. The dapp parser adds URL policy, `aclProgram` ↔ `programs.host`, and boot id. That is consumer-side validation of one JSON, not duplicate schema debt this PR introduced.
- **Treat Vite’s loopback port fallbacks as a layout leak.** `vite.config.ts` documents why it cannot import Bun-typed `layout.ts`. The bind is `127.0.0.1` regardless of `DEMO_DAPP_URL`.

### Unverified

- Preview/devnet scenarios from this exact head (PR says that was not run here).
- Whether the preview relayer accepts API key `local`.
- First public-cluster deploy of an absent program (Helm-mounted keypair, program missing). `solana-deployment.test.ts` starts from genesis-loaded bytecode; the mocked deployer test always supplies a keypair.

Eikix (commented):
Independent review loop of `c41d33cac` vs `feature/solana` `@ b87931e`, including the unique delta `4f4211588..c41d33cac`. Three perspectives (open-ended, maintainability, design/integration), then findings checked against the diff. No patches applied.

This head includes #4071 only through `4f4211588`. It does **not** contain `6e28b033` (`TransientStoreNotOpened`, `programError.ts`, demo translation). Merge #4071 first, then rebase this PR onto the merged `feature/solana` head. A rebase that keeps this PR's host/SDK files drops 6080.

## Accepted

**High — current head still has the generic unopened-journal path.** `fhe_execute.rs` still uses `AccountLoader` for `transient_store`. There is no `TransientStoreNotOpened` in `errors.rs`, no `sdk/js-sdk/src/solana/programError.ts`, and `sendTransaction.ts` has no translator. Forgotten append on this head is still Anchor 3007. After #4071, the mollusk unopened-store test must keep 6080, not `AccountOwnedByWrongProgram`. 4071's teardown still lists `demoServerPlugin.ts`; this PR deletes that file — keep this tree list.

**Medium — leftover localnet identity in a golden.** `test-suite/fhevm/src/solana/spl.test.ts` pins `TOKEN_PROGRAM = pS2gMMq6PNZKpjxiANeoN5XxJgwaFsUR6xaJkpUHcDg` as “the deployed confidential-token program id”. That is the deleted localnet id from `environments/localnet.json`. Live/preview id is `FAWs7E52LZmXR5YzFy4aXanfBjNtXV2qooQVtkmBa3cL`. Vault-authority / escrow goldens would not fail if live PDA derivation broke.

**Medium — silent local chain-id fallback.** `solana/deploy/src/cli.ts` uses `process.env.SOLANA_HOST_CHAIN_ID ?? SOLANA_HOST_CHAIN_ID`. The constant is localnet `72057594037940281`. Helm sets preview `130140237723663404`, so the happy path is correct. A Job that drops the env bootstraps HostConfig / coprocessor `host_chains` with the local id on the shared public program ids. Same class of silent identity split this PR removed for program ids.

**Medium — `personas.fund` on live clusters is ungated.** Demo operator wraps `requireRecoverableWallet` on devnet. Harness `personas.fund` always `context.fundSol`. `deposit-arc.scenario.test.ts` still funds keeper/alice that way. Public-cluster funding policy is not one path.

**Medium — destroy/reset recovery image can prefer a stale recovery Job.** Deploy sets `SOLANA_RECOVERY_IMAGE`. Destroy and `reset-preview.sh` do not. `recover.sh` then `max((is_recovery_job, timestamp, image))`, so any `solana-recovery-*` Job beats a later `solana-host` deploy Job. After a bytecode upgrade, teardown can run an older recovery image.

**Low — preview helper vs component settings.** `preview-env-solana-e2e.sh` copies the leaf-proof key from secret `solana-proof-api` and still writes `DEMO_RELAYER_API_KEY="${DEMO_RELAYER_API_KEY:-local}"`. Vite sends `x-api-key: … ?? "local"`. Relayer OpenAPI marks the key optional on testnet, so this is settings drift, not a proven encrypt failure. Wallet-capability tests still default to localnet and never call `assertWalletAccountCapabilities(..., 'devnet')`.

## Rejected

- Collapsing `SOLANA_ENVIRONMENTS`. One file on purpose; a further Zama is a further file (DD-051/053).
- Merging the two `parseDemoConfig` functions. Dapp vs seed have different fields.
- Treating `admin-sweep` on the unified id as a leftover dual path. It is gated on upgrade authority; that is how preview wipe works with one id.
- Rewriting preview recovery into one control plane, or forcing one send helper / one funding table, as a merge blocker. That is this PR's job at a larger scope than the accepted issues.
- File-length / 1k-line explosions. `batcher_mollusk.rs` was already over 1k.

## Looks aligned

One `declare_id!` from `preview-env.json`; validator `--upgradeable-program` at those ids; localnet.json and the four committed program keypairs are gone. Solana preview registers `130140237723663404` on Gateway/relayer/connector and copies coprocessor keys from EVM `12345`. Permissive relayer CORS removed; demo relayer is same-origin. Burner recovery is registered before faucet; external wallets self-fund on devnet. Explorer links do not put RPC credentials in the URL. Automated host-upgrade stays localnet-only. Wrap/accounts names are gone in the unique delta.

## Unverified

No `solana-e2e` / SDK / demo / mollusk run of this head. Live evidence in the PR body is from a previous preview (previous demo program ids: `gjgLTv4…` / `A2akndg4…`, now `DKuvkbkqo…` / `DAVFjzkMr…`). Concurrent preview ownership on the shared devnet programs was not exercised. Leftover rent on the old demo program ids may still sit on public devnet; recovery only inventories current compiled ids.

